PR #18018, opened 2015-10-22
[v4] remove border-radius for first and last cards in group

KOWLOR:
In a group, cards don't have radius.
- The **right radius** for the **first** card in a group has been removed.
- The **left radius** for the **last** card in a group has been removed.

PR #18097, opened 2015-10-28
Fixed border radius for cards in group

AndersMad:
Using the same syntax as the img. Sample page only works because the first item is an img - the card it self needs border radius.


KOWLOR:
I've already made a PR for that :) [#18018](https://github.com/twbs/bootstrap/pull/18018)

mdo:
Already addressed tonight, thanks!

PR #18159, opened 2015-11-04
v4 - clearfix nav-pills

zalog:


cvrebert:
Do you have an example of the problem that this would solve?

zalog:
Sure.
Because there is no clearing on that list, we can not apply `m-b` or `bg-*`.

Before:


After:


`.nav-tabs` already has clearfix :)

mdo:
Already in `v4-dev`.

PR #18189, opened 2015-11-06
Switch to a numerical color variant naming system based on lightness of variants

michaseel:
Inspired by Font-Weights (200, 400, 600...) and the Material Design Color System (https://www.google.com/design/spec/style/color.html#color-color-palette):
Use numbers to define different shades of gray. This is much more consistent and extendable than light, lighter, lightest, lightest-light.
Because sometimes Designers want just more than 5 shades of gray (or primary colors)

### DIFF
--- a/scss/_utilities-background.scss
+++ b/scss/_utilities-background.scss
@@ -28,9 +28,9 @@
 
 @include bg-variant('.bg-danger', $brand-danger);
 
-@include bg-variant('.bg-gray-100', $brand-gray-100);
-@include bg-variant('.bg-gray-300', $brand-gray-300);
-@include bg-variant('.bg-gray-500', $brand-gray-500);
-@include bg-variant('.bg-gray',     $brand-gray);
-@include bg-variant('.bg-gray-700', $brand-gray-700);
-@include bg-variant('.bg-gray-900', $brand-gray-900);
+@include bg-variant('.bg-gray-100', $gray-100);
+@include bg-variant('.bg-gray-300', $gray-300);
+@include bg-variant('.bg-gray-500', $gray-500);
+@include bg-variant('.bg-gray',     $gray);
+@include bg-variant('.bg-gray-700', $gray-700);
+@include bg-variant('.bg-gray-900', $gray-900);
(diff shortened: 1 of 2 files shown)


oleersoy:
It might be a good idea to use variables for the percentages as well.  For example (PostCSS Implementation):

``` css
:root {
  --percentage-lightest: 93%;
  --percentage-lightness: 72%;
  --percentage-light: 46%;
  --percentage-median: 33%;
  --percentage-dark: 20%;
  --percentage-darker: 13%;
  --percentage-darkest: 5%;
}

:root {
  --color-gray-base: #000;

  --color-gray-lightest: color(var(--color-gray-base) lightness(var(--percentage-lightest)));
  --color-gray-lighter:  color(var(--color-gray-base) lightness(var(--percentage-lighter)));
  --color-gray-light:    color(var(--color-gray-base) lightness(var(--percentage-light)));
  --color-gray:          color(var(--color-gray-base) lightness(var(--percentage-media)));
  --color-gray-dark:     color(var(--color-gray-base) lightness(var(--percentage-dark)));
  --color-gray-darker:   color(var(--color-gray-base) lightness(var(--percentage-darker)));
  --color-gray-darkest:  color(var(--color-gray-base) lightness(var(--percentage-darkest)));
}
```

mdo:
Agreed on something different than the current solution, but unsure why you've added the primary variables here as well.

michaseel:
I would like to have a central place (in variables.scss) where all color shades are defined. Currently there are a lot of different lighten() and darken() operations spread across the sourcecode. That makes the control of particular color-shades very hard.

mdo:
Alternate approach coming in #18462 with full colors included. Needs a ton of work, but figured I'd build on this some how.

oleersoy:
FYI - I took a stab at providing a PostCSS based implementation here:
https://github.com/superfly-css/superfly-css-variables-colors/blob/master/src/main/css/index.css

The [superfly-css-utilities-colors]() module uses the variables to provide a spectrum for both the notification and gray colors.

PR #18217, opened 2015-11-11
Add top border to card header if the header isn't the first child

defsdoor:
This allows a card to have multiple headers.

I chose to do this rather than add a sub-header class.


mdo:
I could see this backfiring where some random hidden element would cause this to glitch out. Thanks, but we'll pass for now on making this default behavior.

defsdoor:
I considered that when I submitted this but I saw it being identical in constraint of use as the classes that add borders on last-child.  For example add an hidden element after a footer and it will screw the borders.

PR #18221, opened 2015-11-11
.media-right regardless of dom order

andheiberg:
The docs show the use of .media-right requiring reordering .media-body and .media-right in the dom.

As we now have flex-box we can just use order so the dom order doesn't matter.

In the case of having css turned off having the image come first is often better.


mdo:
I don't think this works for us since the `order` is a literal integer value (e.g., if someone set `2` on their `.media-body`, this would be bunk).

Issue #17001, opened 2015-08-18
Border color for input group on focus highlighted unevenly on right


jplehmann:
I'm having the same problem which can be seen in BS documentation:
http://getbootstrap.com/components/#input-groups-buttons

Click on the right example "Search for...".  Notice how the blue border/outline on the right side is covered up.  The left example doesn't really have this problem.

![screen shot 2015-08-17 at 8 48 00 pm](https://cloud.githubusercontent.com/assets/460964/9320568/44b4222e-4521-11e5-8750-45e0cdd78d4a.png)

The blue comes from .form-control:focus { border-color, outline, webkit-box-shadow, box-shadow }  at least on my page.

My site is: http://www.morphmarket.com and you'll see the same issue on its search input and the middle button too.  I messed with Chrome inspector for a while, but this is beyond me -- and when I saw the docs had the same problem, I figured I could create an issue.   Thanks!

kkirsche:
PR submitted @jplehmann — thanks for reporting this. @mdo build failed due to a travis issue it seems not the code change itself. JSBin of the fix is in the PR. Just a simple z-index change.

mdo:
Fixed in #18253.

PR #17007, opened 2015-08-18
Use z-index 3 to show focus state on all sides of input

kkirsche:
Fixes #17001

JSBin with fix: http://jsbin.com/gifexipibu/edit?html,css,output

PR #18253, opened 2015-11-15
Ensure proper z-indexing in input-groups

mdo:
Alternate fix to #17007 that limits the `z-index: 3` to only be on `:focus`.


mdo:
Derp, fixes #17001.

PR #18255, opened 2015-11-15
Block level cards

### LINKED ISSUE #18170: v4 - Make .card class display: block
Cards using an anchor element are rendering as inline as the `display` property is inherited, eg.

```
<a class="card card-block" href="#">
  <h4 class="card-text">The parent is inline</h4>
  <p class="card-text">This is some text</p>
</a>
```

Adding `display: block;` to `.card` class would fix this.

mdo:
Fixes #18170.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -4,6 +4,7 @@
 
 .card {
   position: relative;
+  display: block;
   margin-bottom: $card-spacer-y;
   background-color: $card-bg;
   border: $card-border-width solid $card-border-color;


bassjobsen:
@mdo is there a reason that `.card-columns > .card` still got `display: inline-block`? I seems that the cards do not arrange well in Google Chrome (Version 52.0.2743.82 (64-bit)) when it's set to `inline-block`.

mdo:
Looking back at it, we can probably remove that part. I think I was operating under the assumption I needed the elements to be `inline-block` for our use of CSS columns.

Issue #18171, opened 2015-11-05
v4: should carousel captions be hidden by default for small viewports?


bassjobsen:
See: https://github.com/twbs/bootstrap/blob/4d17a04a2766315299de7e6f5d0894bb9ed25228/scss/_carousel.scss#L199
Should i remove the comment or implement it?

mdo:
Yeah, they should be hidden.

mdo:
This can be optionally handled with display utilities. Example coming with #21298.

PR #18258, opened 2015-11-15
hide captions for for smaller viewports

bassjobsen:
should close https://github.com/twbs/bootstrap/issues/18171

### DIFF
--- a/docs/components/carousel.md
+++ b/docs/components/carousel.md
@@ -56,7 +56,7 @@ The `.active` class needs to be added to one of the slides. Otherwise, the carou
 
 ### Optional captions
 
-Add captions to your slides easily with the `.carousel-caption` element within any `.carousel-item`. Place just about any optional HTML within there and it will be automatically aligned and formatted.
+Add captions to your slides easily with the `.carousel-caption` element within any `.carousel-item`. Place just about any optional HTML within there and it will be automatically aligned and formatted. Captions are hidden by default for smaller viewports.
 
 <div class="bd-example">
   <div id="carousel-example-captions" class="carousel slide" data-ride="carousel">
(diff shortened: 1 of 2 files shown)


bassjobsen:
possibly also consider to apply `sr-only()` to hide the caption? But reset it will make the code more complex

patrickhlauke:
Maybe I'm missing something here, but: if captions convey information, why should they be hidden, ever?

Turning them into screen-reader only text on small screen size will, paradoxically, make the information accessible to screen-reader users while not making it accessible to any other user group. Why? Are non-AT-users not supposed to get the info?

cvrebert:
There is a general problem with the caption text getting cut off if there's insufficient room for it.
Example from #18140: Compare:
small
Versus:
supersmall
(The only difference being the viewport width)
Note how the heading in the caption got cut off in the latter screenshot.

However, this isn't particular to XS screens per se, so I agree that this PR's hiding scheme isn't a great solution.

@mdo Could you give your reasoning for https://github.com/twbs/bootstrap/issues/18171#issuecomment-156784143 ?

bassjobsen:
@patrickhlauke 
>non-AT-users not supposed to get the info?
Well non-AT-users can see the images of the carousel. But indeed i agree this make only sense when the information is only addtional to the images.
we shall wait for @mdo :)

mdo:
My thought there was just how crazy they look with the captions overlaying things. Looking at the screenshots again, it'd be nice if the captions simply moved to the bottom of the images instead.

iatek:
A workaround I've found is using CSS 3 `object-fit:cover` (no IE9, edge, etc..) which makes the img behave more like a background image so that it can be scaled properly along with the caption.

http://www.codeply.com/go/EaMgFMhff2

mdo:
This can be optionally handled with display utilities. Example coming with #21298.

PR #18299, opened 2015-11-20
Doc fix #18277

### LINKED ISSUE #18277: Mismatch between documentation & implementation of collapse.js
According to the documentation of Collpase.js:
> The data-target attribute accepts a CSS selector to apply the collapse to.

However in the Collapse.js code:
```javascript
// COLLAPSE PUBLIC CLASS DEFINITION
  // ================================

  var Collapse = function (element, options) {
    this.$element      = $(element)
    this.options       = $.extend({}, Collapse.DEFAULTS, options)
    this.$trigger      = $('[data-toggle="collapse"][href="#' + element.id + '"],' +
                           '[data-toggle="collapse"][data-target="#' + element.id + '"]')
```

As you can see the this.$trigger will only be set to the triggering element only if it has in the data-target an ID, and not any CSS selector. 

This might lead to some unexpected behavior, which the "collapsed" class is not being applied to the button.

See the following example
```
<button class="btn" data-target="#tab1" data-toggle="collapse" type="button">button1</button>
<button class="btn" data-target=".tab2" data-toggle="collapse" type="button">button2</button>
```
when clicking on button1, it will get the class "collpased" which will allow me to style it in different states for collapsed and expand, but the other button will not get this class. 

I suggest other fix the documentation or change the code.

phmasek:
Clarified the documentation that if another CSS selector than an `id` is used, the `class=collapsed` is not added to the trigger element.

### DIFF
--- a/docs/_includes/js/collapse.html
+++ b/docs/_includes/js/collapse.html
@@ -180,7 +180,7 @@
   <p>These classes can be found in <code>component-animations.less</code>.</p>
 
   <h3>Via data attributes</h3>
-  <p>Just add <code>data-toggle="collapse"</code> and a <code>data-target</code> to the element to automatically assign control of a collapsible element. The <code>data-target</code> attribute accepts a CSS selector to apply the collapse to. Be sure to add the class <code>collapse</code> to the collapsible element. If you'd like it to default open, add the additional class <code>in</code>.</p>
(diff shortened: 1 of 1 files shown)


mdo:

<3,
@mdo and team

PR #18346, opened 2015-11-25
Adding some missing CSS prefixes

etiennemargraff:
I used an online static tool to check what was missing and changed the source for **Less** files (in the mixins when needed) to make the test pass.

I included missing **-moz**, **-ms** or **-webkit** prefixes.

Here are two samples before / after including links to the test tool so you can see what I changed:
- **current** bootstrap version : http://bootstrapmeulta.azurewebsites.net/index.current.bootstrap.html 
- bootstrap.css generated **after my changes** : http://bootstrapmeulta.azurewebsites.net/index.html

These are **light changes** and it helps to reflect standards. :)

If needed I will be happy to talk about this here and make some other changes.

Note : I created this in the master branch so it can be reflected in any other branches.


mrmlnc:
Bootstrap uses Autoprefixer.

> **Contributing guidline**
>
> Do not edit bootstrap.css, bootstrap-theme.css, or bootstrap.js directly!

meulta:
Ok thanks for the feedback.
I will do a PR for Normalize directly for the fixes in this file.

I did not edit CSS, I just changed the .less files and let the grunt compiler generate the css files.

While bootstrap is using autoprefixer, some prefixes are still missing in the .less files in the bootstrap repo. Are you really sure I cannot provide this changes?

mrmlnc:
Once these prefixes do not exist in the Normalize and Bootstrap, then maybe they should not be here?

For example, your first change:
```css
//
// 1. Avoid the WebKit bug in Android 4.0.* where (2) destroys native `audio`
//    and `video` controls.
// 2. Correct inability to style clickable `input` types in iOS.
// 3. Improve usability and consistency of cursor style between image-type
//    `input` and others.
//

button,
html input[type="button"], // 1
input[type="reset"],
input[type="submit"] {
  -webkit-appearance: button; // 2
  -moz-appearance: button;              <--- this
  appearance: button;                       <--- and this
  cursor: pointer; // 3
}
```

In comments to Normalize said that item 2 is necessary for:

> 2. Correct inability to style clickable `input` types in **iOS**.

You add this property for all browsers, why?

meulta:
Thank you a lot for taking the time to answer me :)

In my (naive?) opinion, it is always a good idea to add all the vendor prefixes when they exist to be sure to address the biggest range of browsers. Why adding it with one prefix only ?

Thanks

mrmlnc:
When you add all prefixes — you repair that has not broken. For this reason was created Autoprefixer, which is based on [Caniuse](https://github.com/Fyrd/caniuse).

PR #18356, opened 2015-11-26
Card mixin

lesleh:
I'm aware that currently cards are a lot simpler than buttons right now
but this may change in future and it feels cleaner to be able to create
new card styles without needing to know the internals of how they're
implemented.


mdo:
Thanks!

PR #18401, opened 2015-12-02
Proposed better phrasing in the "images" section

tpoulos:
Previously, this read "Bootstrap doesn’t apply this automatically as it causes complications to other image formats." 

In this sentence, the number of the subject and the verb don't agree, and the word "this" is unclear.


patrickhlauke:
Looking good, thanks.

PR #18417, opened 2015-12-03
Multiple button addons in input groups

Caster:
_Let me start by saying that I am not sure if this qualifies as a bug fix or a feature request, but I feel like it is more a feature request, hence I am opening this against the `v4-dev` branch.
I can make this work for `master` too, if you like._

**Changes I made**
_(as screenshots, because images say more than a 1000 words and all that)_

**Before**
before
before-focus

**After**
after
after-focus
after-dropdown-1
after-dropdown-2

**Aim of change**
In an `.input-group`, it is now possible to add multiple button addons after the input with dropdown menus. These have to be placed in separate `.input-group-btn` elements because otherwise the JavaScript breaks. Focus state now highlights the complete group of controls too, instead of only the input. Note that the highlighting only works if the input is the first element in the `.input-group`, because of CSS limitations (I want [`:focus-within`](https://drafts.csswg.org/selectors-4/#focus-within-pseudo)).

**Usecase**
In a project I implemented this for, I have a search box where the user can change a couple of settings for the search. There are three settings that are all of the type 'select from a couple of options'. My approach (that I feel is intuitive, correct me if I'm wrong) was to add three dropdown selects as addons. When that was not styled the way I expected it to be, I first tried to find out what I was doing wrong, then whipped up this, then improved it before creating this pull request.

**Remarks**
I am aware of the following issues:
- The highlighting does not work when there are addons before the input. I purposefully disabled the highlighting of other controls then, because I am unable to select the siblings before the `:focus`ed input with pure CSS. An implementation using JS is of course possible, but I think that that is not something you guys would be happy with, right? This can be fixed once (if) the CSS4 [`:focus-within`](https://drafts.csswg.org/selectors-4/#focus-within-pseudo) selector is supported.
- The left border of buttons is not visible when hovering them. The current implementation has exactly this issue, but for right borders. As soon as a button is focused or active, the border becomes visible. The current implementation will still not show the right border when the button is focused or active.

Finally, I feel that this pull request has an added value because it gets rid of a `margin-left: -1px` hack. This is in place currently to get rid of "double borders". I fix this by removing the left border of elements and adding it back on `:focus` (and `:active`) only. At that point the `z-index` is increased as well, as in the current implementation.

_Side note:_ are you considering to introduce variables for the border width as well, or will that remain "hard-coded" to a single pixel?

**How can I test this?**
I extended the Bootstrap documentation examples and used those for testing.

Link to [development server](http://localhost:9001/components/input-group/#button-addons) _(for your convenience :-))_ | [current documentation](http://v4-alpha.getbootstrap.com/components/input-group/#button-addons).

**To conclude**
It goes without saying that I am open to any and all suggestions, critique and feedback. I can update this pull request as required. Thanks in advance for considering this change.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -35,7 +35,9 @@
   }
 
   // Customize the `:focus` state to imitate native WebKit styles.
-  @include form-control-focus();
+  &:focus {
+    @include form-control-focus();
+  }
 
   // Placeholder
   &::placeholder {
@@ -58,6 +60,12 @@
 
   &:disabled {
     cursor: $cursor-disabled;
+  }
+
+  // Also highlight buttons and dropdown toggles after the focused input in an input group.
+  .input-group > &:first-child:focus ~ .input-group-btn > .btn,
+  .input-group > &:first-child:focus ~ .input-group-btn > .dropdown-toggle {
+    @include form-control-focus();
   }
 }
 
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -61,12 +61,6 @@
   &:disabled {
     cursor: $cursor-disabled;
   }
-
-  // Also highlight buttons and dropdown toggles after the focused input in an input group.
-  .input-group > &:first-child:focus ~ .input-group-btn > .btn,
-  .input-group > &:first-child:focus ~ .input-group-btn > .dropdown-toggle {
-    @include form-control-focus();
-  }
 }
 
 
--- a/docs/components/input-group.md
+++ b/docs/components/input-group.md
@@ -110,13 +110,22 @@ Buttons in input groups are a bit different and require one extra level of nesti
 </div>
 <br>
 <div class="row">
-  <div class="col-lg-offset-3 col-lg-6">
+  <div class="col-lg-6">
     <div class="input-group">
       <span class="input-group-btn">
         <button class="btn btn-secondary" type="button">Hate it</button>
       </span>
       <input type="text" class="form-control" placeholder="Product name">
       <span class="input-group-btn">
+        <button class="btn btn-secondary" type="button">Love it</button>
+      </span>
+    </div>
+  </div>
+  <div class="col-lg-6">
+    <div class="input-group">
+      <input type="text" class="form-control" placeholder="Product name">
+      <span class="input-group-btn">
+        <button class="btn btn-secondary" type="button">Hate it</button>
         <button class="btn btn-secondary" type="button">Love it</button>
       </span>
     </div>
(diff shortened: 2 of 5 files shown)


cvrebert:
> Side note: are you considering to introduce variables for the border width as well, or will that remain "hard-coded" to a single pixel?

As of #18150, none of the borders themselves should be hardcoded anymore. But `_button-group.scss` has some hardcoded `-1px` values for some `margin-*` properties; presumably those are now broken and should be changed.

scss/_forms.scss
```diff
@@ -59,6 +61,12 @@
   &:disabled {
     cursor: $cursor-disabled;
   }
+
```
cvrebert: This should probably go in `_input-group.scss`

cvrebert:
> Focus state now highlights the complete group of controls too, instead of only the input.

Why? To me, that just seems confusing.

Caster:
- Button groups are unaffected by my changes. I am now using `$input-border-width` for the border widths in `_input-group.scss`, since ac5be69cfc462586b9bf2c14993e59cd5b206a35.
  - Moved the styling to `_input-group.scss` in 41859b186001b3abfb6ffd0eb26f748f9430b224.
  - My reasoning behind highlighting the entire group is that it feels like a single form element to me. I found it confusing that not the entire group was highlighted. This is a very debatable thing and I am aware of that, it was my intention to start a small discussion on that. I can take it out easily :-)

cvrebert:
> Button groups are unaffected by my changes. I am now using `$input-border-width` for the border widths in `_input-group.scss`

Right, derp, pretend I said `_input-group.scss` instead of `_button-group.scss`.

---

CC: @patrickhlauke for an accessibility perspective regarding the focus highlighting

patrickhlauke:
> The highlighting does not work when there are addons before the input. I purposefully disabled the highlighting of other controls then, because I am unable to select the siblings before the :focused input with pure CSS. An implementation using JS is of course possible, but I think that that is not something you guys would be happy with, right? This can be fixed once (if) the CSS4 `:focus-within` selector is supported.

This part does concern me, as - unless we strongly advise against putting addons before the input, it will break essential functionality for users who rely on focus highlighting.
As for highlighting the whole group as a unit, I don't have many concerns here. It may be unfamiliar when first encountered, but should not confuse long term.

cvrebert:
> An implementation using JS is of course possible, but I think that that is not something you guys would be happy with, right?

Indeed.

Caster:
> This part does concern me, as - unless we strongly advise against putting addons before the input - it will break essential functionality for users who rely on focus highlighting.

Let me rephrase: I only highlight the input itself in case there are addons before the input, so the input is *always* highlighted. In case there are no addons before the input, then the addons are highlighted too.

**Visual explanation of possible situations**
focus-1
focus-2
focus-3
focus-4

patrickhlauke:
But once the addons themselves are focused (moving to them with TAB/SHIFT+TAB) they do get their normal highlighting, right? If that's the case then yeh, that's fine.

Caster:
Yes, that is indeed the case.

focus-5
focus-6
focus-7
focus-8
focus-9
focus-10

scss/_forms.scss
```diff
@@ -35,7 +35,9 @@
   }
 
   // Customize the `:focus` state to imitate native WebKit styles.
-  @include form-control-focus();
+  &:focus {
```
mdo: The `&:focus` is part of the mixin.

mdo:
Started to review this, but realized it's going down the wrong path. You're calling the `form-control-focus` mixin in a ton of places, repeating the shared styles that all textual inputs freely get with it on `.form-control`. Beyond that, the selectors don't seem to scale out in a good way given your Remarks section from the opening comment—those behavioral changes block anything from changing here.

Thanks for the hard work, but we won't be merging this. If you're looking to try again, something with `first-of-type` might be more helpful and lead to few changed lines.

Caster:
Thank you for you feedback. Just for completeness, and to do better in the future, I'd like to reply to the remarks and ask a couple of questions - please do not feel obliged to answer, I'd just appreciate it a ton if you (or anyone really) would.

  - The `form-control-focus` mixin was used in only one spot, I used it in one other spot as well, to highlight addons after a focused input. First of all: I realize that this behavioral change may not be something you want, so I could take it out. But say we would want this, how could you do it with fewer `form-control-focus` calls? It would be awesome if the mixin could still include the `&:focus` as well, I was not able to keep it because of the different selector I wanted to use it in (where the `:focus` part is halfway the selector, opposed to at the end).
  - So I am feeling sort of stupid now, but I fail to see which styles I have been repeating for textual inputs. Could you point out one or two, so that I can prevent this from happening in the future?
  - Is the behavioral change with the focus highlighting not desired at all, or just not in its current form? And would making it configurable with a variable make it better, or even worse (because there would be more configuration options hence more complex et cetera)?
  - Thank you for the `first-of-type` suggestion. Unfortunately I fail to see where it could help, but I will give it some more thought and wait for an *aha! moment* :-)

Thank you again for all the feedback and in general for working on Bootstrap. I love every single bit of it.

PR #18431, opened 2015-12-05
V4 dev docs library updates

XhmikosR:
/CC @cvrebert @hnrch02


cvrebert:
The Holder.js update seems to break in Chrome:
holder-broke

cvrebert:
The updated Anchor.js and Clipboard.js seem to work fine.

XhmikosR:
Weird, I tested Firefox, IE 10 and worked fine there. I'll skip the holder update and merge.

XhmikosR:
@imsky: any idea why this happens?

On a side, note this happened with Firefox the second time, so it's like random. v2.8.2 works fine in all cases, it seems.

imsky:
i'll look into why this is happening

PR #18462, opened 2015-12-08
Color palettes

mdo:
Motivated by #18189.

This implements a basic color palette using color functions for each of our base colors, including gray. It's intended to give us and users more flexibility and granularity in their color choices and increase the level of customization throughout the project. As with the original mentioned PR, this is **super heavily** pulled from Google's design docs—big fan of that stuff.

Here's how it looks right now.

screen shot 2015-12-07 at 11 17 03 pm
screen shot 2015-12-07 at 11 21 00 pm


cvrebert:
Are we going to define `$brand-*` in terms of this palette?

mdo:
> Are we going to define `$brand-*` in terms of this palette?

Yeah, still have to work out the relationship there.

More todos:

- [ ] Figure out how to handle `$brand-*` vars
- [ ] Sass the shit out of this
- [ ] Maybe use Jekyll data collections for spitting out the HTML?
- [ ] Peep color contrast of the `500`s
- [ ] Update gray scale to use the `500` as a base

lachlanjc:
@mdo Do we want these classes available for anyone to use on their Bootstrap site, or just show them inside the docs?

cvrebert:
Given the number of colors, exposing them as classes seems like it'd be excessive/bloaty.

lachlanjc:
👍

mdo:
Hitting a wall on this unfortunately. Could use some help if anyone is up for flexing some Sass map fu :). Ideally we'd have Sass dynamically output all those `-100`-`-900` variables, but Sass doesn't do dynamic variable mapping. I tried all day to get some kind of mashing of Sass maps together, but never got exactly what I wanted.

That's the part I can't get locked down and my brain is fried after futzing with it for so long.

lachlanjc:
I don't have a ton of experience with Sass functions, but could we use an `@each` to set the variables? Then use another each for setting the class names?

sergej2705:
How about implementing the `theme-color`-function like

```SCSS
@function theme-color($key: 'primary', $illumination: 500) {
  $bias: 500;
  $illumination-factor: 0.08%;
  
  $color-map: map-merge($colors, $theme-color);
  $act-color: map-get($color-map, $key);
  
  $illumination-biased: $illumination - $bias;
  $illumination-change: abs($illumination-biased * $illumination-factor);
  
  @if $illumination >= 0 and $illumination < $bias {
    @return lighten($act-color, $illumination-change);
  }
  @else if $illumination > $bias and $illumination < ($bias * 2) {
    @return darken($act-color, $illumination-change);
  }
  @else { // if illumination is 500 or out of range return the unchanged color
    @return $act-color;
  }
}
```

You would call then `theme-color('foreground', 200)` instead of `$foreground-200`.

There are then nearly infinite possibilities:

```SCSS
theme-color('success', 700);    // $success-700
theme-color('secondary', 500);  // $secondary-500
theme-color('foreground', 200); // $foreground-200
theme-color('info', 900);       // $info-900

theme-color();                  // $primary-500
theme-color('primary');         // $primary-500
theme-color('primary', 500);    // $primary-500

// and other colors that were not possible before
theme-color('orange', 200);
theme-color('teal', 500);
theme-color('blue', 900);
```

mdo:
Overhauled things a bit this last week or so.

- Shades now done via `mix` instead of `darken`/`lighten` resulting in less faded out colors
- New theme color function for handling the shading
- Updated most instances of light grays to ensure components are feeling properly colored

screenshot 2016-04-10 22 45 56

patrickhlauke:
i'm guessing this will impact https://github.com/twbs/bootstrap/pull/19693 - can we coordinate these two (assuming you're ok with the changes in that one)?

mdo:
> i'm guessing this will impact #19693 - can we coordinate these two (assuming you're ok with the changes in that one)?

Yup, definitely will. Accessibility is important, but we need to be careful with those changing colors. They quickly change from feeling like a family of colors to a bunch of colors we just picked randomly. That green and orange get bad real quick with those changes.

patrickhlauke:
subjective, but i actually find them more cohesive in https://github.com/twbs/bootstrap/pull/19693 ... they have more consistent brightness compared to the overly washed-out default green and orange

patrickhlauke:
If you get a chance, I know the whole of #19693 will need rebasing once this lands, but: leaving aside the palette itself, mind looking over all other proposed contrast tweaks there? Most should be less controversial than the theme color changes (mostly revolve around picking a slightly darker grey to start with for certain things, and tweak the secondary button in normal and outline form).

mdo:
Cherry-picked a couple commits from your PR @patrickhlauke. Going to punt on changing theme colors and the button styles just yet though. Really want to get this merged and lock some other things up and I'm not happy with where those new colors ended up in that PR.

patrickhlauke:
sure, happy to sit tight once yours has merged...but forewarning you that regardless of wcag contrast ratios, subjectively i find some of the color combos quite washed out and weak even for "normal" users (eg the breadcrumb styles and the secondary outline button, which has various issues from others filed against it).

patrickhlauke:
btw, speaking of color contrast etc, I do like what Dropbox did in their documentation/style guide with regards to calling out the colors that ARE in fact WCAG 2.0 compliant. see http://dropbox.github.io/scooter/functions.html#colors (which incorporates my recent PR https://github.com/dropbox/scooter/pull/31)

mdo:
@cvrebert Any idea what's up with Travis? I see some PRs passing, but CI seems to be super flakey and looking at the history of even `v4-dev`, I don't see obvious changes that could be the culprit.

cvrebert:
@mdo Sauce Labs was suffering from an incident; see https://github.com/js-cookie/js-cookie/pull/179
It's supposedly been resolved now, so the build should succeed if you retry.

patrickhlauke:
dropping this in as related https://medium.com/eightshapes-llc/color-in-design-systems-a1c80f65fa3#.3pkskno6n

mdo:
Added color variables in #21378. Going to punt on the rest until after stable v4.

PR #18521, opened 2015-12-10
add $input-font-weight variable for .form-control

andrewmnlv:


mdo:
Passing on this—we don't make use of any special weight for form controls by default. Thanks though!

Issue #18527, opened 2015-12-10
contextual, striped, animated progressbar not working in IE and Edge


canfiax:
There is no styling on the progressbar in Edge on Windows 10

patrickhlauke:
http://getbootstrap.com/components/#progress (and, if you were referring to v4 alpha, http://v4-alpha.getbootstrap.com/components/progress/) look just as they should in Edge/Win 10.

patrickhlauke:
![capture2](https://cloud.githubusercontent.com/assets/895831/11728053/7787a92e-9f7f-11e5-9c2b-3c03cd399175.PNG)
![capture](https://cloud.githubusercontent.com/assets/895831/11728054/778bf48e-9f7f-11e5-978f-1099796e5e5c.PNG)

canfiax:
V4 Alpha. They don't. They don't adopt correct colors. They are all blue

patrickhlauke:
Fixed - https://github.com/twbs/bootstrap/commit/70247e36f64295c61e0f867606793da5a26eebeb

PR #18531, opened 2015-12-10
contextual progressbar for IE10+/Edge, clarify striped/animated lack of support

patrickhlauke:
closes https://github.com/twbs/bootstrap/issues/18527

### DIFF
--- a/scss/_progress.scss
+++ b/scss/_progress.scss
@@ -19,12 +19,19 @@
   margin-bottom: $spacer-y;
 }
 .progress[value] {
-  // IE10 uses `color` to set the bar background-color
-  color: #0074d9;
+  // Set overall background
+  background-color: #eee;
   // Remove Firefox and Opera border
   border: 0;
   // Reset the default appearance
   appearance: none;
+  // Set overall border radius
+  @include border-radius($border-radius);
+}
+.progress[value]::-ms-fill {
+  background-color: #0074d9;
+  // Remove right-hand border of value bar from IE10+/Edge
+  border: 0;
 }
 .progress[value]::-webkit-progress-bar {
   background-color: #eee;
@@ -108,6 +115,10 @@
   background-size: $spacer-y $spacer-y;
 }
 .progress-striped[value]::-moz-progress-bar {
+  @include gradient-striped();
+  background-size: $spacer-y $spacer-y;
+}
+.progress-striped[value]::-ms-fill {
   @include gradient-striped();
   background-size: $spacer-y $spacer-y;
 }
--- a/scss/mixins/_progress.scss
+++ b/scss/mixins/_progress.scss
@@ -9,6 +9,11 @@
     background-color: $color;
   }
 
+  // IE10+, Microsoft Edge
+  &[value]::-ms-fill {
+    background-color: $color;
+  }
+
   // IE9
   @media screen and (min-width:0\0) {
     .progress-bar {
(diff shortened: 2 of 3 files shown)


patrickhlauke:
Before (left) and after (right)

capture

cvrebert:
Nitpick: There's trailing whitespace in the docs changes.

cvrebert:
What about using [`::-ms-fill`](https://msdn.microsoft.com/en-us/library/windows/apps/hh465757.aspx) ?

patrickhlauke:
first i've heard of `::ms-fill`. will have a play and see...

patrickhlauke:
`::-ms-fill` did indeed do the trick. however, despite trying variations, it appears that it doesn't support CSS3 animations

capture

scss/mixins/_progress.scss
```diff
@@ -9,6 +9,11 @@
     background-color: $color;
   }
 
+  // IE10+, Microsoft Edge
+  &[value] {
```
cvrebert: Use `::-ms-fill` here too?

patrickhlauke: it feels "cleaner" to me not to - as it follows the common advice that's floating around about IE using color rather than background-color, and the basic example from https://msdn.microsoft.com/en-us/library/windows/apps/jj651676.aspx 

have to admit though, as soon as i pushed this new version, i started pondering that myself. using `::-ms-fill` would at least be consistent with the use of pseudos for other engines...

patrickhlauke: also, looking at `scss/_progress.scss` we currently have

```
.progress[value] {
  // IE10 uses `color` to set the bar background-color
  color: #0074d9;
  ...
```
so it feels a bit more consistent to still stick with non-`::-ms-fill` one here. having said all that, i don't feel particularly strongly about it here either way...

patrickhlauke: having said all that, i found a need to start using `::-ms-fill` anyway in `scss/_progress.scss` - to suppress an unsightly right-hand border - which then led me down a whole different avenue anyway. long story short: yes, using `::-ms-fill` for this now (and now IE10+/Edge also get the rounder borders like webkit)

patrickhlauke:
capture

scss/_progress.scss
```diff
@@ -19,12 +19,20 @@
...
 }
 .progress[value] {
-  // IE10 uses `color` to set the bar background-color
-  color: #0074d9;
   // Remove Firefox and Opera border
   border: 0;
   // Reset the default appearance
   appearance: none;
+  // Set overall border radius
+  border-radius: $border-radius;
```
cvrebert: Should use `@include border-radius($border-radius);` ?

scss/_progress.scss
```diff
@@ -19,12 +19,20 @@
...
+  // Set overall background
+  background-color: #eee;
+}
+.progress[value]::-ms-fill {
+  background-color: #0074d9;
+  // Remove right-hand border of value bar from IE10+/Edge
+  border: 0;
+  border-radius: $border-radius;
 }
```
patrickhlauke: ah, i see this is not exactly how the webkit one is handled - only adds the rounding to the left AND right if it's a 100% value. i'll get onto this tomorrow...

patrickhlauke:
actually...seems it's much simpler to achieve this correctly here. result: progress bar is vertical/straight, but overall progress corners are rounded, just as with webkit.

capture

scss/_progress.scss
patrickhlauke: done

PR #18579, opened 2015-12-15
Fix vertical alignment issue on floated label in list group item

### LINKED ISSUE #18574: Floated .label text is vertically misaligned with other .list-group text
The problem is that the text line and number are not on the same line.
monosnap 2015-12-15 04-53-20
And if we make it less in width it looks ugly.
monosnap 2015-12-15 04-59-11
From http://v4-alpha.getbootstrap.com/components/list-group/#labels

gaga1227:
Fix #18574
Adding a position adjustment to compensate vertical positioning difference due to different font-size and line-height values.

### DIFF
--- a/scss/_labels.scss
+++ b/scss/_labels.scss
@@ -27,6 +27,17 @@
   top: -1px;
 }
 
+// Fix vertical alignment on floated labels in list group item
+@each $breakpoint in map-keys($grid-breakpoints) {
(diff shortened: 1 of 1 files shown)


mdo:
I'm unsure exactly what the fix should be here, but I don't think this is the right one just yet. Going to close this out and leave the issue open until I can find a better solution that doesn't require extra positioning on top of a float utility as used today. Thanks though!

Issue #18510, opened 2015-12-09
Not all .container widths are divisible by 12


arechsteiner:
I've been working on a Photoshop starter template for Bootstrap 4 with column guides, and I've noticed  that not all column widths are divisible by twelve, which results in oddly sized column widths.

In alpha 2 this only affects "Large":

`940px - 360px (gutter) / 12 = 48.33333...px`

As a result, some columns are rendered with `48px` width and others with `49px` width. 

This makes working with graphic tools harder and more cumbersome. For example, Photoshop has a "Guide Layout" functionality that would work well *if* columns were evenly sized:

![2015-12-09 17_50_20-untitled-1 100 rgb_8](https://cloud.githubusercontent.com/assets/1199486/11691938/7db7a0d4-9e9d-11e5-9bdf-d65e60f73088.png)

Is there a good reason for this that I am unaware of? I initially assumed this was a trade-off related to using rem, but now that this was ditched, I don't quite understand why the containers are not made so they divide evenly.

PS: I've published my [Photoshop template here](http://hackerthemes.com/blog/bootstrap-4-psd-template/), if you want to take a look. It includes test renderings that I've made in Firefox in order to get the column widths exactly right.

bassjobsen:
see also: http://benstewart.net/2014/06/bootstrap-grid-calculation-explained/ or set the width of your large container to 936px

arechsteiner:
Thanks @bassjobsen. I've looked at the SCSS and I think I understand why it's done with %-values. However:

> or set the width of your large container to 936px 

My point is, why isn't this (or 948px) set by default? I'm convinced that 99% of Bootstrap end-users will not customize Bootstrap through recompilation of the SCSS source, so why not have sane defaults that do not result in half-pixels in the column width?

PR #18600, opened 2015-12-18
Changed .container max-width for large from 940px to 948px

arechsteiner:
As discussed here:

https://github.com/twbs/bootstrap/issues/13120 and https://github.com/twbs/bootstrap/issues/18510

The two options that make sense are either 948px or 936px. Bootstrap 3 uses 940px. I think it's better to err on the side of having a few pixels more instead of less.

Then all containers will be divisible by 12 by default, which does away with uneven column widths.

### DIFF
--- a/docs/layout/grid.md
+++ b/docs/layout/grid.md
@@ -95,7 +95,7 @@ See how aspects of the Bootstrap grid system work across multiple devices with a
         <td>None (auto)</td>
         <td>576px</td>
         <td>720px</td>
-        <td>940px</td>
+        <td>948px</td>
         <td>1140px</td>
       </tr>
       <tr>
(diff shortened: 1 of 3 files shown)


mdo:
Thanks, but I'm going to leave it as-is for now. Might take another look before finishing v4, but we've run 940px for a few years now :).

arechsteiner:
@mdo: Thanks for considering my pull request. Please consider your comments [here](https://github.com/twbs/bootstrap/issues/13120#issuecomment-38090127) and [here](https://github.com/twbs/bootstrap/issues/13120#issuecomment-38318860) from two years ago.

This is your only chance for another few laps around the sun.

PR #18621, opened 2015-12-21
Use $gray-lightest in _variables.scss

simonewebdesign:
Since we have two very similar colors, `#f5f5f5` and `#f7f7f9`, I'd propose to just use the first one -- so this PR replaces every occurrence of `#f7f7f9` with `$gray-lightest`.

Also, since we are not using `$gray-lightest` anywhere apart from [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/utilities/_background.scss#L13), I guess it would make sense to either remove it, or start using it somewhere else.

Thoughts?

### DIFF
--- a/docs/assets/scss/_component-examples.scss
+++ b/docs/assets/scss/_component-examples.scss
@@ -79,7 +79,7 @@
   position: relative;
   padding: 1rem;
   margin: 1rem -1rem;
-  border: solid #f7f7f9;
+  border: solid $gray-lightest;
   border-width: .2rem 0 0;
   @include clearfix();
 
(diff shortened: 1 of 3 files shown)


simonewebdesign:
ping

ollym:
@simonewebdesign 100% agree with this.

Also similarly `#e5e5e5` is used all over the place, which is very similar to `$gray-lighter: #eceeef !default`.

I'd suggest updating your pull request to set `$card-border-color`, `$modal-header-border-color`, `$dropdown-divider-bg`, `$card-border-color` to `$gray-lighter`

simonewebdesign:
Hey @ollym, thanks for your input and sorry for the long silence!

I just did as you suggested, apart from `$card-border-color` which changed to `rgba(0,0,0,.125)` in the meantime, so I left it as it is.

mdo:
Superseded by #21206. Thanks!

PR #18654, opened 2015-12-23
Docs: reintroduce `aria-describedby` advice for help text

patrickhlauke:
Based on v3 http://getbootstrap.com/css/#forms-help-text, adapted to make examples more "real life"

### DIFF
--- a/docs/components/forms.md
+++ b/docs/components/forms.md
@@ -517,19 +517,33 @@ Wrap inputs in grid columns, or any custom parent element, to easily enforce des
 
 No official help text classes exist in Bootstrap 4 (previously we had `.help-block` in v3), but thanks to our utility classes like `.text-muted`, you can create much more flexible help text as you need it.
 
+{% callout warning %}
+#### Associating help text with form controls
+
+Help text should be explicitly associated with the form control it relates to using the `aria-describedby` attribute. This will ensure that assistive technologies – such as screen readers – will announce this help text when the user focuses or enters the control.
+{% endcallout %}
+
 Inline text can use any typical inline HTML element (be it a `<small>`, `<span>`, or something else).
 
 {% example html %}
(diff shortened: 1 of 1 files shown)


patrickhlauke:
capture

docs/components/forms.md
```diff
@@ -517,19 +517,31 @@ Wrap inputs in grid columns, or any custom parent element, to easily enforce des
...
+</form>
 {% endexample %}
 
 Block help text—for below inputs or for longer lines of help text—can be easily achieved with a `<p>`.
 
 {% example html %}
-<p class="text-muted">
-  A block of help text that breaks onto a new line and may extend beyond one line.
+<label for="inputPassword2">Password</label>
+<input type="password" id="inputPassword2" class="form-control" aria-describedby="passwordHelpBlock">
```
cvrebert:
```
Running "htmllint:src" (htmllint) task
_gh_pages/components/forms/index.html [L1679:C103] Duplicate ID “inputPassword2”.
```

I'd recommend using `inputPassword4` and `inputPassword5` as the IDs in these new examples instead.

patrickhlauke: ah, good catch (related, I really need to sort out my eslint which keeps giving me weird errors in unrelated files - js/src/scrollspy.js - as that's what's been keeping me from running local tests)

cvrebert:
Seems like these should use `.form-group` for consistency with the other form examples.

patrickhlauke:
Added a `<div class="form-group">` to the first example, as that also had the full `<form>` around it. For the second one, I'd say it can remain without the `<div>`, as it's only a fragment and not a complete form (similar to some of the more focused previous examples like radio buttons etc)

Also spotted two instances of help text in the main form controls example which lacked `aria-describedby`...rectified now

PR #18662, opened 2015-12-24
Docs: accessibility fix-up of collapsible content navbar, change site-wide main navbar

patrickhlauke:
Adds the missing `aria-controls`, `aria-expanded` and - because of the use of a single unicode character on the toggle, with no actual content that can be read by assistive tech - `aria-label` to the navbar collapsible content examples.

Moves the first collapsible content _after_ the toggle for accessibility/focus and reading order issues.

Modifies the site-wide main navigation to follow the collapsible content navbar pattern from the actual documentation.

Note: because v4's current styles for collapsible navbars are a bit...broken, this also introduces the same brokenness to the site-wide navigation. Even more reason to tackle this soon...

### DIFF
--- a/docs/_includes/nav-home.html
+++ b/docs/_includes/nav-home.html
@@ -25,7 +25,7 @@
   {% endcomment %}
 
   <div class="clearfix">
-    <button class="navbar-toggler pull-xs-right hidden-sm-up" type="button" data-toggle="collapse" data-target="#bd-main-nav">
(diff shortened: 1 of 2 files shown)


patrickhlauke:
This is what the main site nav will now look like on small screen when expanded...
capture

but this is consistent with the - currently broken - look of collapsible navbars, as seen on http://v4-alpha.getbootstrap.com/components/navbar/#collapsible-content

capture2

PR #18673, opened 2015-12-25
introduce the plain-focus-active mixin and use  the plain-focus-active and hover-focus mixins

bassjobsen:


mdo:
Man, that `plain` keyword choked me up a bit there. We should document in the CSS or our docs what that means. For a few mins I read it as "the plain-focus and plain-active" states and was super confused, but it's plain, focus, and active.

mdo:
Going to punt on this for code legibility and general staleness. Thanks though!

PR #18689, opened 2015-12-26
fix broken layout for flexbox

### LINKED ISSUE #18688: documentation layout broken when compiling Bootstrap with Flexbox support
On pages of the documentation the `.bd-content` columns is too wide due to the `div.highlight` takes full width and did not get `overflow-x: scroll`.
The page look like that shown in the figure shown beneath (on both FF and chrome):

brokendocs

According to http://stackoverflow.com/questions/12022288/how-to-keep-a-flex-item-from-overflowing-due-to-its-text fixing is possible with:

I found that the following code fixes the issue for both FF and Chrome:

    .bd-content {
       min-width: 0px;
     }

For other browsers the following code is possible needed:

    .bd-content {
      flex-shrink: 1;
      min-width: 0px;
    }

    .bd-sidebar {
      flex-grow: 1;
    }
    
Or even:

    .bd-content {
      flex-shrink: 1;
      flex-order: 1;
      min-width: 0px;
    }

    .bd-sidebar {
      flex-order: 2; 
      flex-grow: 1;
    }

bassjobsen:
possible closes #18688, but should be test for other browser than FF and Chrome

### DIFF
--- a/docs/assets/scss/_content.scss
+++ b/docs/assets/scss/_content.scss
@@ -3,6 +3,7 @@
 //
 
 .bd-content {
+  min-width: 0px;
   > table {
     display: block;
     width: 100%;
--- a/docs/assets/scss/_content.scss
+++ b/docs/assets/scss/_content.scss
@@ -3,7 +3,7 @@
 //
 
 .bd-content {
-  min-width: 0px;
+  min-width: 0;
   > table {
     display: block;
     width: 100%;


bassjobsen:
close in favor of https://github.com/twbs/bootstrap/pull/17756
But i wonder if setting `min-width` or `max-width` do not do the same? (trigger recalcution of the widths)

Issue #17586, opened 2015-09-13
.container, .container-fluid should be conditional upon $enable-grid-classes


mmjaeger:
I'm actually wondering why you guys didn't put .container and .container-fluid not into the $enable-grid-classes condition - if I'm using the container mixin I don't need those container classes be created?

mdo:
Fixed in #18695.

PR #18695, opened 2015-12-26
.container, .container-fluid should be conditional upon $enable-grid-…

bassjobsen:
…classes

see: https://github.com/twbs/bootstrap/issues/17586

Notice that all code can also wrapped in a single `@if $enable-grid-classes {}` now.

Code can also be wrapped inside a mixin `grid-classes()` with the following code in bootstrap.scss:
    @import "grid";
    @if $enable-grid-classes {
      @include grid-classes;
    }

And finally notice that the `@import "grid";` is not needed at all when you do not need the precompiled grid classes. Removing the import from `bootstrap.scss` is as simple as setting    `$enable-grid-classes` to `false`.

### DIFF
--- a/scss/_grid.scss
+++ b/scss/_grid.scss
@@ -2,22 +2,23 @@
 //
 // Set the container width, and override it for fixed navbars in media queries.
 
-.container {
-  @include make-container();
-  @include make-container-max-widths();
-
+@if $enable-grid-classes {
+  .container {
+    @include make-container();
+    @include make-container-max-widths();
+  }
 }
-
 
 // Fluid container
 //
 // Utilizes the mixin meant for fixed width containers, but without any defined
 // width for fluid, full width layouts.
 
-.container-fluid {
-  @include make-container();
+@if $enable-grid-classes {
+  .container-fluid {
+    @include make-container();
+  }
 }
-
 
 // Row
 //


cvrebert:
MSTM. CC: @mdo

bassjobsen:
what does mstm means?

cvrebert:
"Makes sense to me"

bassjobsen:
thanks! MSTM :)

2015-12-27 16:11 GMT+01:00 Chris Rebert <notifications@github.com>:

mdo:
Snap, thanks!

dalabarge:
I just had a use case where I do not rely on the grid classes but do use the grid mixins and expected the container classes to exist. With an update to this commit, it broke. To turn on the grid would bloat my CSS output with a lot of unused classes just so I can get the container classes enabled. I do use the grid, just not the classes. I could foresee also not even needing the grid but perhaps needing the containment of the container so that I get more predictable layouts at defined breaking points.

Not sure this inclusion really make sense as the container doesn't necessarily connect to the grid so making them dependent on each other seems unnecessary coupling. Another config option seems in order howbeit then we are getting into config overhead land. Maybe `$enable-container-classes` or perhaps separate the grid classes from the concept of grid entirely so that classes are included with `$enable-grid-classes` (turns on all `.col-*` and `.row` classes) while `$enable-grid` turns on all related classes and mixins such as `.container` and `.container-fluid`. You could then have `$enable-grid == true` and `$enable-grid-classes == false` to achieve what this PR permits while still giving my use case a simple solution.

PR #18699, opened 2015-12-27
Add pointer-events:none to disabled pagination/pager links

patrickhlauke:
as currently this was missed out (compared to disabled `a.btn`)


patrickhlauke:
x-ref also https://github.com/twbs/bootstrap/pull/18700

PR #18707, opened 2015-12-27
update the dashboard examples

bassjobsen:
Please notice that many fixes are required for the issue already described at https://github.com/twbs/bootstrap/pull/17756.

There is also an issue for the sidebar with `position: fixed` in a flexbox item having now `width`, see also /\* see http://stackoverflow.com/questions/34344623/fixed-position-element-inheriting-width-of-flex-item */
Solved for by:

```
width: calc(100% * (3/12));
@media (min-width: 768px) {
  .nav-sidebar {
    width: calc(100% * (2/12)); 
  }
}  
```

Code only tested on FF and Chrome

### DIFF
--- a/docs/examples/dashboard/dashboard.css
+++ b/docs/examples/dashboard/dashboard.css
@@ -7,7 +7,6 @@ body {
   padding-top: 50px;
 }
 
-
 /*
  * Global add-ons
  */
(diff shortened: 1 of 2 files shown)


bassjobsen:
close in favor of https://github.com/twbs/bootstrap/pull/18709

bassjobsen:
see: https://github.com/twbs/bootstrap/issues/17816

PR #18709, opened 2015-12-27
Compile examples CSS with Sass using BS variables and mixins

bassjobsen:


mdo:
Thanks, but the examples are intended to be 100% usable without compilation of any kind. Most of the examples are also not that complex and wouldn't benefit from Sass honestly.

bassjobsen:
Hi @mdo, i respect your choice and understand your motivation, but i do not agree with it. I think compiling the examples do not have any drawbacks. You can still use the compiled code as before, On the other hand it is a test for the code (mixins),  build process and people can use the examples to find out how to extend and modify the code leveraging Sass.
dashboard.scss uses the media query mixins, the hover mixin and the `$enable-flex` var, also see the comments of https://github.com/twbs/bootstrap/pull/18707

PR #18716, opened 2015-12-29
Add max-width to file browser; fixes #18651

### LINKED ISSUE #18651: File browser too wide on iOS/iPhone
Not sure if it's easily fixable, but on iOS 9.2 on my iPhone 5c the file browser http://v4-alpha.getbootstrap.com/components/forms/#file-browser is wider than the viewport.

samhh:
Although visually the first added property fixes it, with only the first property Chrome was still displaying the element as leaking out of its container - this is what the second added property is for.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -175,11 +175,13 @@
 .file {
   position: relative;
   display: inline-block;
+  max-width: 100%;
   height: 2.5rem;
   cursor: pointer;
 }
 .file input {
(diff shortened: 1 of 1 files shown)


patrickhlauke:
Great stuff. confirming that this addresses the issue perfectly in iOS, with no adverse side effects in other tested browsers.

SamHH:
Hooray for my first PR in a big project! :-D

patrickhlauke:
:clap:

PR #18741, opened 2016-01-01
fix justified navbar

bassjobsen:
should close https://github.com/twbs/bootstrap/issues/17723, see also https://github.com/twbs/bootstrap/pull/18653


mdo:
Thanks!

PR #18782, opened 2016-01-06
v4: Drop pager component

mdo:
This drops the entire pager component, nullifying #18629 in the process.

I've been debating this for awhile, but essentially the pager component is a slightly customized version of an outline button with custom float classes. That hasn't sat right with me because it feels woefully unnecessary by comparison to other components. Rather than encourage it's usage into v4, I'm dropping it entirely.
- [x] Remove Sass styles
- [x] Remove docs
- [x] Add migration docs notes
- [x] Remove from examples


mdo:
Here's how the pagination styles look in the blog example with outline buttons and a customized `border-radius` applied:

screen shot 2016-01-06 at 2 56 39 pm

PR #18804, opened 2016-01-08
Scrollspy selecting the last element when at the bottom of the page

### LINKED ISSUE #17739: ScrollSpy no longer selects last element when at the bottom of the page
This was added in v2.0.3 but appears broken with v4-alpha and ScrollSpy attached to the `<body>`.  The last item is not selected when you scroll to the bottom of the page. 

```
var maxScroll = this._config.offset + scrollHeight - this._scrollElement.offsetHeight;
```

The issue appears to be that `maxScroll` is `Nan` because `this._scrollElement.offsetHeight` is `undefined` 

```
this._scrollElement = element.tagName === 'BODY' ? window : element;
```

As I'm using the `<body>` element `this._scrollElement` is `window`

gurisko:
Fixes #17739

### DIFF
--- a/js/src/scrollspy.js
+++ b/js/src/scrollspy.js
@@ -195,12 +195,17 @@ const ScrollSpy = (($) => {
       )
     }
 
+    _getOffsetHeight() {
+      return this._scrollElement === window ?
+          window.innerHeight : this._scrollElement.offsetHeight
+    }
+
     _process() {
       let scrollTop    = this._getScrollTop() + this._config.offset
       let scrollHeight = this._getScrollHeight()
       let maxScroll    = this._config.offset
         + scrollHeight
-        - this._scrollElement.offsetHeight
+        - this._getOffsetHeight()
 
       if (this._scrollHeight !== scrollHeight) {
         this.refresh()
@@ -212,6 +217,7 @@ const ScrollSpy = (($) => {
         if (this._activeTarget !== target) {
           this._activate(target)
         }
+        return
       }
 
       if (this._activeTarget && scrollTop < this._offsets[0]) {
--- a/js/tests/visual/scrollspy.html
+++ b/js/tests/visual/scrollspy.html
@@ -28,6 +28,7 @@
             <a class="dropdown-item" href="#three" tabindex="-1">three</a>
           </div>
         </li>
+        <li class="nav-item"><a class="nav-link" href="#four">@four</a></li>
       </ul>
     </div>
   </nav>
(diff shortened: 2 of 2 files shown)


cvrebert:
Please add a relevant JS unit test or visual test for your change; see https://github.com/twbs/bootstrap/tree/v4-dev/js/tests for instructions.

js/src/scrollspy.js
```diff
@@ -212,6 +217,7 @@ const ScrollSpy = (($) => {
         if (this._activeTarget !== target) {
           this._activate(target)
         }
+        return
```
cvrebert: Why?

gurisko: Because it could happen that `scrollTop` is lower than `offset` of the last target despite it is the end of the page and it would overwrite `this._activeTarget` to the last but one target (line 235).

cvrebert: Ah, sorry, I was on mobile and it looked like this was already the end of the function (obviously it's not).

gurisko:
@cvrebert is the visual test enough for now?

js/tests/visual/scrollspy.html
```diff
@@ -28,6 +28,7 @@
             <a class="dropdown-item" href="#three" tabindex="-1">three</a>
           </div>
         </li>
+        <li class="nav-item"><a class="nav-link" href="#four">@four</a></li>
```
cvrebert: Change this to something like "Selected when page scrolled to bottom", so as to explicitly give the test expectation.

timfish:
What's left to do before this can be merged in?

gurisko:
All comments should be implemented.

@timfish Sorry, I missed the notifications. I believe nothing is left to do before the merge anymore.

cvrebert:
So, this has the side-effect of making the unselectability of middle items on the last screen noticeable. Which also feels weird from a user perspective. Here's an example:
kobayashi-maru

Notice how "2nd to last" is unreachable.

I'm not sure how we ought to handle that.
CC: @patrickhlauke @XhmikosR @hnrch02 @mdo for thoughts

timfish:
Still slightly strange behaviour but not sure how else it should work and doesn't actually appear to be a regression from v3:
https://github.com/twbs/bootstrap/issues/15729

hnrch02:
That's a though UX decision—I'm not sure about this either. But this PR restores the original behavior from v2 and v3 so I'd say 👍 .

mdo:
Super strange, but this is now in `v4-dev` and didn't get auto-closed. Might be something to do with the fork being deleted (as it says `unknown repository` up top).

Relevant commits on our end are 39d7861f34e01933ebee4f6b05e3ed24bca7f116 and d328c911cbc74e6958ee747ebf17a7481a32d831.

PR #18830, opened 2016-01-10
Add prefix to carousel classes

krzysztof-grzybek:
Add carousel- prefix to Carousel's .next,.prev,.left,.right classes #17021


mdo:
Dupe of #18161.

PR #18838, opened 2016-01-11
refactor pagination in accordance with the Navs

bassjobsen:


mdo:
This was nullified by Alpha 6. Thanks though!

PR #18871, opened 2016-01-13
Added missing file promised in _variables.scss

draecko:
_variables.scss mentions _custom.scss for custom variables, but is not included in the file structure or build list. This is a quick solution.


wolfy1339:
Duplicate of #18398

PR #18872, opened 2016-01-13
Remove <meta name="keywords"> from docs

coliff:
Remove meta keywords - AFAIK no search engines use them for search ranking at all. There's no need for them to be included.


Adrian-Q:
www animowane-gify com-gify-rozne-buzki-65

cvrebert:
Merged as https://github.com/twbs/bootstrap/commit/562d2239c0b8ef56dae04bd44eac0c419def6467. Thanks.

Issue #16673, opened 2015-06-19
IE11 radio button and checkboxes using button plugin don't work correctly when navigating with keyboard


mynameisstephen:
Steps to reproduce
1. In IE11 tab to a radio button group.
2. Press right or left to select a different radio button
3. Tab out of the button group
4. Unhide the underlining input tags.  The selected radio button will differ from the selected input

The left and right keys are triggering the input tags to change however the display is not being updated.

kkirsche:
Could you post a [JSBin](http://jsbin.com/) demonstrating the isolated issue?

mynameisstephen:
http://jsbin.com/mogoqusuje/1/edit?html,css,output

The bootstrap site also exhibits the issue.
http://getbootstrap.com/javascript/#buttons-checkbox-radio

mdo:
Does this happen in every browser?

animafact:
NAVER - http://www.naver.com/
--------------------------------------------

animafact@naver.com 님께 보내신 메일 <Re: [bootstrap] IE11 radio button groups display differ from underlining input state (#16673)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

kkirsche:
@animafact please don't post ads / links to third party sites unrelated to the issue.

@mdo I believe so, or at least seems to in Chrome of OS X — Assuming that I understood correctly that left and right moves the active class and prevents the user from selecting the focused element (via space bar) but does not change the actual active radio.

mynameisstephen:
@kkirsche @mdo Hmm Firefox seems to be the only browser working correctly.  Looks like Chrome and IE differ in behavior. IE will change the radio button on left and right keys but won't update the display while Chrome will update the display but won't update the radio buttons.

kkirsche:
My guess is that this is being caused somewhere in https://github.com/twbs/bootstrap/blob/master/js/button.js#L56-L72

Removing the JS solves the value not moving but stops the active class from moving with the value. I'd need to dig in a little deeper to identify the exact cause in this

kkirsche:
Odd... in a further stripped JSBin where I only load the button JS, this seems to prove opposite. Overlap with another JS component potentially? — http://jsbin.com/tojiro/edit?html,css,js,output

mynameisstephen:
[...]

PR #18900, opened 2016-01-14
Button toggling - trigger change event on input

vlastikcz:
Bootstrap’s .button styles can be applied to other elements, such as labels, to provide checkbox or radio style button toggling.

And when the checkbox or radio state is changed, there should be triggered
the change event. Currently, the change event is triggered on the Button, which is not correct. Only input fields do support the change event.

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -90,7 +90,7 @@ const Button = (($) => {
 
           if (triggerChangeEvent) {
             input.checked = !$(this._element).hasClass(ClassName.ACTIVE)
-            $(this._element).trigger('change')
+            $(input).trigger('change')
           }
         }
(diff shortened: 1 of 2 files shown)


wolfy1339:
Please exclude dist file changes

vlastikcz:
@wolfy1339 done

vlastikcz:
are there any other issues with the Pull Request?

mdo:
I believe this fixes #16673.

AlexVentura:
Hey there!
Has the issue being fixed? Because I still having it on **Bootstrap v4.1.1**. This is [my example](https://codepen.io/alexventura/pen/gKqwXr) for reference.

PR #18914, opened 2016-01-16
Stop oversized images in .card-img-* from overflowing the card's boundaries

simeon:
Images in cards using card-img-top or card-img-bottom ~~via src="" (instead of data-src="")~~ would overflow past the bounds of the card if the image width was too large.  Now they will at most be equal to the width of the card.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -168,11 +168,11 @@
 // Card image caps
 .card-img-top {
   @include border-radius($card-border-radius-inner $card-border-radius-inner 0 0);
-  max-width: 100%;
+  @include img-fluid(block);
 }
 .card-img-bottom {
   @include border-radius(0 0 $card-border-radius-inner $card-border-radius-inner);
-  max-width: 100%;
+  @include img-fluid(block);
 }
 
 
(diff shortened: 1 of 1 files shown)


cvrebert:
FYI, `data-src` is a [Holder.js](https://github.com/imsky/holder)-ism. The fact that `data-src` was ever shown in the docs code snippets was a bug (#18558).
So you're really just talking about oversized images in general.

cvrebert:
Seems like it might be better to `@include img-fluid` instead, but I'm not too familiar with the Cards code.

simeon:
@cvrebert I looked into it and you're right. Using @include img-fluid(block); would do the same job with a better "grammer".  I've updated my commit to reflect that.

cvrebert:
Have you tested this out when Flexbox is enabled (`$enable-flex: true;`)?

simeon:
@cvrebert  Yup, I just recompiled with Flexbox enabled and it's all good.

cvrebert:
CC: @mdo for final review

mkroeders:
Have this issue also, so to counter it I did an .img-fluid on the img. . Which did not solve it for me, a width: 100% was needed for me.

I checked when you use $enable-flex: true; the width: 100% is needed otherwise the .img-fluid is enough.

mdo:
This was addressed with the use of utilities in #20667, but I didn't check out the flex variation. If that's still a problem, can one of y'all open an issue with a test case that confirms that behavior? <3

PR #18919, opened 2016-01-17
Add breakpoints for .pull-right on .dropdown-menu

Metrakit:
Added .breakpoints for .pull-right on .dropdown-menu.

I have added breakpoints to this class for can use the .dropdown-menu at right like on the v3.3.
This feature answer the #18550 issue.

Usage Example : 

``` html
<ul class="nav pull-md-right">
  <li class="dropdown">
    <a href="#" class="dropdown-toggle" data-toggle="dropdown">
      Menu Right
    </a>
    <ul class="dropdown-menu">
      <li><a href="#">link</a></li>
    </ul>
  </li>
</ul>
```


wolfy1339:
Please exclude the changes in dist files

mdo:
Probably not going to go this route with our utilities. I don't think we should be doing this kind of behavior in v4 given our class-based, fewest parent selectors as possible approach. Plus, if you're moving around your dropdown menus, I feel like that's a sign of a different kind of problem in your UI. Thanks though!

Metrakit:
Ok, good to know. Thanks.

PR #18948, opened 2016-01-20
V4 dev border radius

danmartin:
I altered border radius mixin for my in house team. We do a lot of sass work and I wanted a single mixin to reference that would also reduce markup... in v4-dev it only reduced the bootstrap.css codebase by 27 lines. My team uses it a lot in development so it the code reduction is greater.

I thought I would hand it over to the repository to see if it had any merit.


wolfy1339:
Please exclude changes in the dist files

danmartin:
My apologies I overwrote the branch. I will re-submit the pull request minus the dist files.

PR #18959, opened 2016-01-20
Google

googlelegacy:
GoogleLegacy was here!


kkirsche:
Stop posting that you were here please. At least give some form of useful message

PR #19006, opened 2016-01-22
Add variables for popover

xcatliu:
Add two variables `popover-padding-y` and `popover-padding-x`


mdo:
Almost thinking we nuke that inner padding entirely to avoid setting the extra styles. Let me noodle on that for a bit.

mdo:
Alternate fix coming in #19889. Thanks!

PR #19007, opened 2016-01-22
Add media type "all" to media queries to fix them for IE9

DaBs:
Writing @media (max-width: something) will break in IE9, as the media type is not defined. This is easily fixed by simply rewriting all media queries with e.g. @media all and (max-width: something).

### DIFF
--- a/less/dropdowns.less
+++ b/less/dropdowns.less
@@ -202,7 +202,7 @@
 //
 // Reiterate per navbar.less and the modified component alignment there.
 
-@media (min-width: @grid-float-breakpoint) {
+@media all and (min-width: @grid-float-breakpoint) {
   .navbar-right {
     .dropdown-menu {
       .dropdown-menu-right();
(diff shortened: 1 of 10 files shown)


cvrebert:
Given how long Bootstrap 3 has been out, it seems rather unlikely that such a major bug has gone unnoticed all this time.

cvrebert:
The existing media queries seem to be working fine on IE9.
Testcase: http://getbootstrap.com/css/#responsive-utilities-tests

Extra-small viewport:
xs

Small viewport:
sm

Medium viewport:
md

@DaBs Do you have a counterexample to offer?

DaBs:
Yes, it did seem strange to me when I reported it too, albeit the proposed fix completely eliminated the problem. Interestingly enough, I cannot reproduce the bug as of today, albeit adding "all" to the media queries fixed my issues yesterday. Closing pull request as this does not seem reproducible.

PR #19013, opened 2016-01-23
Use $nav-item-inline-spacer for navbar brand/items

Mevrael:


mdo:
That variable is for `.nav-inline`, not the navbar. Thanks though!

Mevrael:
@mdo Yes, variable is for inline nav items and navbar-nav is also inline nav by it's nature.

You can manage spacing via variable between items only in `.nav-inline` but can't for `nav navbar-nav` and adding `.nav-inline` won't help.

I though't about making new variable but do we really need to seperate same presentation logic. When I first viewed source I tried to override $nav-item-inline-spacer in my sass but only hours later I figured out that 1em spacing in navbar items is hardcoded.

Later, if will be necessary, new variable for navbar-nav can be added. It is the same as telling that $border-radius is not for navbar and you should create $navbar-border-radius.

Anyway, I need this for my new project and don't want to keep a bit modified copy of original sass files for too long. If you wan't new var $navbar-item-inline-spacer I can fix PR.

PR #19018, opened 2016-01-25
Update print styles to v5.3.0 of html5 boilerplate.

XhmikosR:
/CC @cvrebert @mdo

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -2,14 +2,21 @@
 
 // ==========================================================================
 // Print styles.
-// Inlined to avoid the additional HTTP request: h5bp.com/r
+// Inlined to avoid the additional HTTP request:
+// http://www.phpied.com/delay-loading-your-print-css/
 // ==========================================================================
 
 @media print {
   *,
   *::before,
-  *::after {
+  *::after,
+  *::first-letter,
+  *::first-line {
+    // Bootstrap specific; comment out `color` and `background`
+    //color: #000 !important; // Black prints faster:
+                              //   http://www.sanbeiji.com/archives/953
     text-shadow: none !important;
+    //background: transparent !important;
     box-shadow: none !important;
   }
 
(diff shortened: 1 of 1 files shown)


scss/_print.scss
```diff
@@ -18,18 +24,37 @@
...
+
+  a[href^="#"]::after,
+  a[href^="javascript:"]::after {
+    content: "";
+  }
+
   pre,
   blockquote {
-    border: $border-width solid #999;
+    border: 1px solid #999;
```
XhmikosR: This was removed so that we are on par with upstream and since it wasn't marked as custom. I can put it back and add a comment indicating it's custom to fit our needs.

scss/_print.scss
```diff
@@ -2,14 +2,20 @@
...
 // ==========================================================================
 
 @media print {
   *,
   *::before,
-  *::after {
+  *::after,
+  *::first-letter,
+  *::first-line {
+    color: #000 !important; // Black prints faster:
```
cvrebert: This was deliberately removed and should not be re-added

scss/_print.scss
```diff
@@ -18,18 +24,37 @@
     text-decoration: underline;
   }
 
+  a[href]::after {
```
cvrebert: This was deliberately removed and should not be re-added

scss/_print.scss
```diff
@@ -18,18 +24,37 @@
...
+    content: " (" attr(href) ")";
+  }
+
   abbr[title]::after {
     content: " (" attr(title) ")";
   }
 
+  //
```
cvrebert: Ditto

XhmikosR: All right, I wonder how we could mark those as such for future reference.

How about the border variable?

cvrebert: I suppose we could comment them out and add additional comments explaining why.

scss/_print.scss
```diff
@@ -18,18 +24,37 @@
...
+
+  a[href^="#"]::after,
+  a[href^="javascript:"]::after {
+    content: "";
+  }
+
   pre,
   blockquote {
-    border: $border-width solid #999;
```
cvrebert: \*shrug* I don't know the context for this one off the top of my head. Seems like a reasonable deviation from upstream though.

XhmikosR:
@cvrebert: can you check now?

cvrebert:
Since we're commenting out `color`, seems like we should comment out `background` too (so that we don't accidentally make any text unreadable due to lack of contrast).

cvrebert:
Any opinion on
https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css
vs.
https://github.com/h5bp/html5-boilerplate/blob/5.3.0/src/css/main.css
for the link in line 1?

XhmikosR:
@cvrebert: fixed.

About the link, I just feel that putting the tag is too much. Personally, I mention the version in the commit message.

cvrebert:
Fair. I have no further objections then.

PR #19022, opened 2016-01-25
Jumbotron sass: replace background-color with background

Mevrael:
background-color: limits variable only to colors but jumbotron often is used with bg image.

background-color: X have same effect as background: X, however, now image can be set as jumbotron bg using variable.


mdo:
We prefer explicit properties over the shorthand. I see what this would provide, but rather than add variables and change the code here, I'd suggest simply writing new CSS to add the background image.

PR #19029, opened 2016-01-26
Update grunt-html

XhmikosR:
- [x] We'll probably need to tweak Travis to use OpenJDK 8 before this is merged
- [ ] And backport this to v3

/CC @cvrebert @patrickhlauke

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: false
+sudo: required
+dist: trusty
 language: node_js
 git:
   depth: 10
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -327,8 +327,7 @@ module.exports = function (grunt) {
           'The “datetime-local” input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.',
           'The “month” input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.',
           'The “time” input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.',
-          'The “week” input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.',
-          'Attribute “integrity” not allowed on element “script” at this point.' // Until https://github.com/jzaefferer/grunt-html/issues/86 gets fixed
+          'The “week” input type is not supported in all browsers. Please be sure to test, and consider using a polyfill.'
         ]
       },
       src: ['_gh_pages/**/*.html', 'js/tests/visual/*.html']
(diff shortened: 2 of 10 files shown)


XhmikosR:
From a quick look, I had to switch to `java` for the travis project. I'm experimenting with other solutions but they seem to be slower.

If anyone has any ideas please let me know.

cvrebert:
If only https://github.com/travis-ci/travis-ci/issues/4090 was implemented.. :disappointed:

XhmikosR:
So, should I switch to `java` in the meantime? It also seems unlikely that Travis will update to OpenJDK 8 as the default for now so we either switch to `java` or require sudo and install it ourselves; the latter would be a lot slower, I suspect.

cvrebert:
Yes, I agree. Guess we'll have to invoke `nvm` ourselves, which won't be too hard.

XhmikosR:
All right, I'll get to it this evening.

XhmikosR:
@cvrebert: Seems to fail in `nvm`. Any ideas?

cvrebert:
@XhmikosR Try this matrix:
```yaml
matrix:
  - NODE_VERSION=5 TWBS_TEST=core
  - NODE_VERSION=4 TWBS_TEST=core
  - NODE_VERSION=5 TWBS_TEST=validate-html
  - NODE_VERSION=5 TWBS_TEST=sauce-js-unit
```

XhmikosR:
Thanks @cvrebert, that seems to fix it the first build.

I thought I'd need a `&&` or something but apparently not.

Now, there are issues due to caching and the different node.js versions...

cvrebert:
Looks like we'd need https://github.com/travis-ci/travis-ci/issues/4393

cvrebert:
Looks like we may have to stick with `language: node_js` and accept the slowness of installing JDK8 ourselves.

XhmikosR:
All right I'll give it a go later unless you beat me to it. In that case
just commit on this branch.

XhmikosR:
@cvrebert: this doesn't seem to work either yet :/ Maybe I'm missing something...

cvrebert:
Seems to be working, but OpenJDK would probably be simpler and faster to install than Oracle Java.

XhmikosR:
@cvrebert: so we only need `openjdk-8-jre`?

XhmikosR:
@cvrebert: I tried `openjdk-8-jre` but doesn't seem to work.

XhmikosR:
@cvrebert: can you check again now? It seems using the Trusty image works fine.

cvrebert:
LGTM

XhmikosR:
All right, I'll clean up the branch and merge. Can you backport it to v3?

cvrebert:
Ok.

Issue #18726, opened 2015-12-30
Changing card-header from div to h3 adds margin-bottom


Studio384:
The card header adds .5em margin to the bottom when switched from div to h3, that shouldn't happen.

mdo:
Fixed by #19048.

PR #19048, opened 2016-01-28
Have same margin for <hN> as a <div> on .card-headers

mkroeders:
Related #18726

When you use a `<hN>` as `.card-header` there is `margin-bottom: .5rem` added. The changes set the margin-bottom to 0

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -71,7 +71,7 @@
 .card-header {
   @include clearfix;
   padding: $card-spacer-y $card-spacer-x;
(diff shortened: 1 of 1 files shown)


mdo:
Merged this for now, but I may come back to revert this and similar changes in favor of margin utilities that can easily remove the `margin-bottom` for you.

jamesharvard:
Hey, @mkroeders & @mdo, given the [default appearance](https://v4-alpha.getbootstrap.com/components/card/#header-and-footer) seems to require zero margin, might it be preferable to just set all margins with `margin: 0;`?

PR #19122, opened 2016-02-08
Use bootstrap-toc plugin for sidebar navigation in docs

afeld:
I was looking to replicate the sidebar navigation from http://getbootstrap.com (as [are](http://stackoverflow.com/questions/23303035/how-to-recreate-menu-like-the-right-nav-menu-at-getbootstrap-com) [others](http://stackoverflow.com/questions/23458567/easy-way-navbar-bootstrap?lq=1)), but was disappointed to discover that it wasn't a standalone Component. Decided to create one myself:

http://afeld.github.io/bootstrap-toc/

I figured the ultimate test would be seeing if the plugin actually works for http://getbootstrap.com – this pull request is that test!

Previously, the sidebar navigation for every page (seems to have been) constructed by hand, which means each file needs to be kept up-to-date with the headings on the respective page – with the plugin, that all happens automatically. Switching to this plugin results in (nearly) identical markup and identical functionality of the side nav bar, with a lot less clutter in the repository.

To verify the results, I ran the resulting markup from each page before and after the switch to the plugin through `diff`, so there shouldn't be anything missing. I did discover that some headings from the `main` area were previously missing from the side nav, so this PR also fixes those issues.


docs/_includes/js/overview.html
```diff
@@ -76,11 +76,9 @@ <h2 id="js-version-nums">Version numbers</h2>
...
+  <h2 id="js-disabled" data-toc-text="When JavaScript is disabled">No special fallbacks when JavaScript is disabled</h2>
   <p>Bootstrap's plugins don't fall back particularly gracefully when JavaScript is disabled. If you care about the user experience in this case, use <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noscript"><code>&lt;noscript&gt;</code></a> to explain the situation (and how to re-enable JavaScript) to your users, and/or add your own custom fallbacks.</p>
 
-  <div class="bs-callout bs-callout-warning" id="callout-third-party-libs">
-    <h4>Third-party libraries</h4>
-    <p><strong>Bootstrap does not officially support third-party JavaScript libraries</strong> like Prototype or jQuery UI. Despite <code>.noConflict</code> and namespaced events, there may be compatibility problems that you need to fix on your own.</p>
-  </div>
+  <h2 id="callout-third-party-libs">Third-party libraries</h2>
+  <p><strong>Bootstrap does not officially support third-party JavaScript libraries</strong> like Prototype or jQuery UI. Despite <code>.noConflict</code> and namespaced events, there may be compatibility problems that you need to fix on your own.</p>
```
afeld: This is one of the couple of places I intentionally changed to make the heading the same as the others on the page, so that it would show up in the sidebar.

**Before**

screen shot 2016-02-08 at 4 19 35 am

**After**

screen shot 2016-02-08 at 4 20 14 am

There are other "warning" sections on the page that aren't displayed as callouts (such as "no special fallbacks" info above), so I went the easy route rather than trying to support the existing markup. Let me know if this is a deal-breaker.

docs/getting-started.html
```diff
@@ -16,9 +16,9 @@
 {% include getting-started/disabling-responsiveness.html %}
 
 <!-- Cross link to new migration page -->
-<div class="bs-callout bs-callout-info" id="migration">
-  <h4>Migrating from v2.x to v3.x</h4>
-  <p>Looking to migrate from an older version of Bootstrap to v3.x? Check out <a href="../migration">our migration guide</a>.</p>
+<div class="bs-docs-section">
+  <h1 class="page-header" id="migration" data-toc-text="Migrating from 2.x to 3.0">Migrating from v2.x to v3.x</h1>
+  <p class="lead">Looking to migrate from an older version of Bootstrap to v3.x? Check out <a href="../migration">our migration guide</a>.</p>
```
afeld: Ditto my comment [above](#r52146168).

**Before**

screen shot 2016-02-08 at 4 18 53 am

**After**

screen shot 2016-02-08 at 4 24 42 am

cvrebert:
I don't really trust rawgit as a CDN...
Also, since v3 is in maintenance mode, we're generally trying to avoid making major changes to anything, for the sake of stability.
Regardless, neat plugin!

afeld:
> I don't really trust rawgit as a CDN

Fair enough – that was just the easiest thing that would work. Happy to publish it somewhere else or add it to this repository. Suggestions?

> since v3 is in maintenance mode, we're generally trying to avoid making major changes to anything

Well, it does fix some missing sidebar links, but I hear you.

The [v4 docs](http://v4-alpha.getbootstrap.com/getting-started/introduction/) don't do the affixing or use ScrollSpy – were those intentionally excluded, or just something that hasn't been gotten to yet? Happy to do a separate pull request to `v4-dev` if that would be preferred. Or, you know, if it's a "thanks but no thanks", that's fine too.

sad puppy

> neat plugin!

Thanks!

mdo:

<3,
@mdo and team

PR #17965, opened 2015-10-18
v4 - .c-input must have some validation styles

zalog:
```
<label class="c-input c-checkbox has-error">
    <input type="checkbox">
    <span class="c-indicator"></span>
    Check this custom checkbox
</label>
```

PR #19139, opened 2016-02-09
v4: Validation states on custom forms

mdo:
Alternate fix to #17965. Updates to use the new class name and adds some basic docs for custom checkbox/radio validation. Nothing added for custom selects or file inputs—yet.

/cc @zalog


zalog:
Yeah, that's better! Thanks!

PR #19157, opened 2016-02-11
v4: Rename .label class to .tag

ameya-pandilwar:
Renamed the .label class to .tag so as to avoid naming conflict between Bootstrap's '.label' class name and the '<label>' HTML tag. Fix for #19094


simonewebdesign:
:+1: thanks for this!

Issue #19199, opened 2016-02-15
Outlined buttons have the same appearance on :active and :focus states


TiGR:
Slightly modified example from the docs:

```html
<div class="btn-group" data-toggle="buttons">
  <label class="btn btn-primary-outline">
    <input type="checkbox" checked="" autocomplete="off"> Checkbox 1
  </label>
  <label class="btn btn-primary">
    <input type="checkbox" autocomplete="off"> Checkbox 2
  </label>
</div>
```

Once you clicked the button you can't say what is its state actually until you move your mouse pointer away (or on mobile devices - tap outside of the buttons area). So, if user have tapped some checkbox he can't really say what state is it in.

Compare this behavior with not outilined buttons.

patrickhlauke:
I'm assuming you mean in v4 ?

TiGR:
Yep, there are no outlined buttons in v3

patrickhlauke:
btw note that in your example code `btn-primary-outline` should be `btn-outline-primary`

TiGR:
My code was based on docs, it seems that outline buttons docs need to be updated.

patrickhlauke:
actually, ignore me...i seem to have gotten myself mighty confused while hacking away with devtools. you and the documentation are right...

patrickhlauke:
need to stop looking at the out-of-sync v4-alpha site, as it seems the naming was swapped at some point. i'll make a note to check/update the docs if needed.

patrickhlauke:
just to confirm that the docs in the latest version here are up-to-date. it's just http://v4-alpha.getbootstrap.com that is quite a few commits behind.

TiGR:
@patrickhlauke I've checked the actual docs. See pull request.

PR #16154, opened 2015-03-25
Darker style focus/hover on active buttons

patrickhlauke:
Currently, hovering with mouse or setting focus on a button which is
active has same styling as on a non-active button. This results in
problems for keyboard users, who set focus on a toggle and activate it,
but cannot visually see that their action had any effect. Ditto for
mouse users hovering over a toggle and clicking it. This adds an
explicit additional style for focus/hover on active buttons.
Note that this does not address issues of browser focus remaining on a
button after a mouse click (e.g. #13971), as this will likely require
extra JavaScript to fix.

PR #19201, opened 2016-02-15
Add extra styles for focused/hovered active outline buttons

patrickhlauke:
Ports the same approach from https://github.com/twbs/bootstrap/commit/1f153b640de55a4fff1aca3cf7331e9b7e3a408f#diff-edb9cd89b66120ef946a81de31f6bb06 (originally https://github.com/twbs/bootstrap/pull/16154 in v3) to outline buttons as well

Fixes https://github.com/twbs/bootstrap/issues/19199

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -64,19 +64,33 @@
   background-color: transparent;
   border-color: $color;
 
+  @include hover {
+    color: #fff;
+    background-color: $color;
+        border-color: $color;
+  }
+
   &:focus,
-  &.focus,
+  &.focus {
+    color: #fff;
+    background-color: $color;
+        border-color: $color;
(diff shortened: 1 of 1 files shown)


patrickhlauke:
before (left) and after (right) comparison (using a modified bit in our docs) - note before that an active outline button style was the same when hovered with mouse or focused with keyboard (minus the very faint outline in the latter case). this patch should make it a lot clearer.

> cap

patrickhlauke:
@cvrebert any idea why @houndci-bot flagged those spaces, when just a few lines previously (especially in the current mixin for `button-variant`) it's using the exact same spacing?

cvrebert:
@patrickhlauke Not really, although we don't align `-color` like that in many other places in the code.

PR #17266, opened 2015-08-24
Appropriate use of fieldset/div, legends, accessibility advice

patrickhlauke:
Use more generic `<div>` for single form controls, use `<fieldset>` and
`<legend>` for radio button groups and checkboxes (where appropriate).
Include accessibility advice from v3
Fixes #17248

PR #19141, opened 2016-02-09
v4: Forms cleanup

mdo:
This branch aims to overhaul a decent amount of forms styles and documentation with the hopes of getting it near final. It's still WIP and has a few more todos (noted below).

#### Quick screenshot

screen shot 2016-02-09 at 1 06 30 am

#### Changes

- #17498, #18130: Removes horizontal padding on `.form-control-label` so it avoids overriding grid-based form layouts and renders correctly in stacked forms.
- #18515: Add `.form-control-{sm/lg}` classes to match small and large input classes, namely for vertical alignment of label and input in horizontal form groups.
- #18702: Updates form validation docs to include basic examples for situations in which you'd want to use each state.
- #18703: Drops `.text-help` for `.form-control-feedback`, designed specifically to only display temporary form validation feedback text _in addition to_ normal help text.
- #18704: Updates form validation docs to include examples of validation feedback text alongside always-present form field help text.

#### Todos
- [x] #17248, #17266, #19277: Fieldsets vs divs in form groups
- [x] #18390: Consider renaming .radio and .checkbox 
- [x] #18573: Consider renaming .form-control-label for it's more contextual use cases (e.g., only horizontal/grid forms and not necessary with radios/checkboxes as shown in #18389).
- ~~#19004: Color contrast and color blindness on validation colors~~ _Punting on this for another time_
- [x] #19139: Validation states on custom forms (needs select menus and file inputs)—maybe cherry-pick that PR into this one?
- [x] #19771: Form control sizing not applying to `<select>`s.

PR #19277, opened 2016-02-19
Forms fixes: fieldset/legend, new .form-control-legend, accessibility advice

patrickhlauke:
- Add new `.form-control-legend` class (to style `<legend>` like `<label>` for inline etc forms)
- More appropriate use of `<fieldset>` and `<legend>`
- Added accessibility advice for hidden/no label alternatives

Supercedes https://github.com/twbs/bootstrap/pull/17266

### DIFF
--- a/docs/components/forms.md
+++ b/docs/components/forms.md
@@ -19,16 +19,16 @@ Remember, since Bootstrap utilizes the HTML5 doctype, **all inputs must have a `
 
 {% example html %}
 <form>
-  <fieldset class="form-group">
+  <div class="form-group">
     <label for="exampleInputEmail1">Email address</label>
     <input type="email" class="form-control" id="exampleInputEmail1" aria-describedby="emailHelp" placeholder="Enter email">
     <small id="emailHelp" class="text-muted">We'll never share your email with anyone else.</small>
-  </fieldset>
-  <fieldset class="form-group">
+  </div>
+  <div class="form-group">
     <label for="exampleInputPassword1">Password</label>
     <input type="password" class="form-control" id="exampleInputPassword1" placeholder="Password">
-  </fieldset>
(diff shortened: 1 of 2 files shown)


patrickhlauke:
Note that even with this new pull request, the issue of having `<legend>` in the grid example is not completely solved.

> capture

However, the issue here is not so much the new `.form-control-legend`, but rather the fact that the right-hand side of the grid, for the radio button and checkbox cases, lacks the same top padding of the previous two rows (email, password). So, it's something that should be addressed there, rather than trying to force a specific vertical padding on the `<legend>` there, I'd say.

@mdo I would propose to merge this, and then address the vertical inconsistency in https://github.com/twbs/bootstrap/pull/19141 ?

patrickhlauke:
@mdo thoughts on this?

docs/components/forms.md
```diff
@@ -47,34 +47,37 @@ Remember, since Bootstrap utilizes the HTML5 doctype, **all inputs must have a `
...
-    </label>
-  </div>
-  <div class="radio disabled">
-    <label>
-      <input type="radio" name="optionsRadios" id="optionsRadios3" value="option3" disabled>
-      Option three is disabled
-    </label>
   </div>
+  <fieldset class="form-group">
+    <legend>Radio buttons</legend>
```
cvrebert: What does this legend look like visually?

cvrebert:
LGTM, modulo adding a TODO regarding the legend alignment.

patrickhlauke:
For completeness (as @cvrebert asked separately), this is how the `<legend>` on https://github.com/twbs/bootstrap/pull/19277/files#diff-582bb2b553026528f5f065d75a757c33R61 looks like after this.

> capture

PR #19281, opened 2016-02-21
FIX: Part of Popover is outside boundaries of viewport

### LINKED ISSUE #19279: Part of Popover is outside boundaries of viewport
bootstrap-1

it should look like this:

bootstrap-2

martinfojtik:
Fixes #19279.

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -278,6 +278,15 @@ const Tooltip = (($) => {
 
         $(this.element).trigger(this.constructor.Event.INSERTED)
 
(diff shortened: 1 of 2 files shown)


jrtnq514:
In order for this to adjust the popover arrow wouldn't you have to change the popover arrow back to utilizing the `.popover-arrow` class instead of the pseudo elements it was changed to in this [commit](https://github.com/twbs/bootstrap/commit/e7f79b32f22d7f39138a7003272b54ceb3b12426#diff-4145fbc2788b1617e73eeebdc4e38e60)?

Johann-S:
Sorry but we removed Tether for Popper see : #22444

Issue #19315, opened 2016-02-25
Add utility class for word-wrap/word-break


adrien-be:
Relevant pull request: https://github.com/twbs/bootstrap/pull/19316

Very long words can break out of its containing tag. 

One can force these long words to break to the next line(s) if necessary.

See an example on https://jsfiddle.net/AdrienBe/n16vqz5p/

Resources:
http://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/
http://stackoverflow.com/questions/320184/how-to-prevent-long-words-from-breaking-my-div
http://caniuse.com/#search=word-wrap
http://stackoverflow.com/a/22811590/759452
http://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap

eddyerburgh:
I suggest adding these to body in normalize : 

```
body {
  // Stops words overflowing 
  -ms-word-wrap: break-word;
  word-wrap: break-word; 
}
```

vanillajonathan:
https://getbootstrap.com/docs/4.0/utilities/text/

There is the `.text-nowrap` class and the `.text-truncate` class. There is no class to force long lines uninterrupted words to break into new lines though.

Perhaps there ought to be a `.text-wrap` or `.break-word` class.

```css
.text-wrap {
  word-wrap: break-word; 
}
```

PR #19316, opened 2016-02-25
new feature: break long words to next line

0xadri:
See relevant feature request: https://github.com/twbs/bootstrap/issues/19315

### DIFF
--- a/less/utilities.less
+++ b/less/utilities.less
@@ -53,3 +53,16 @@
 .affix {
   position: fixed;
 }
+
+
+// Break long words to next line
+// -------------------------
+
+.break-long-words {
+  word-break: break-word;
+  word-wrap: break-word;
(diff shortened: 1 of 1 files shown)


adrien-be:
Just realized my PR was on Bootstrap 3 ... :/

cvrebert:
My note and comment from #19313 apply here as well.

IdanCo:
@adrien-be , love the utility class, but why not do it globally?

in #18927 they specifically mentioned that `We deliberately do NOT reset font-size or word-wrap.`. What am i missing here? does any one **wants** titles or paragraphs to exceed container's width?

adrien-be:
Because I'd rather let everyone opt-in.
And if one wants to opt-in globally, you can do this easily using
sass mixins.
First world problem I suppose ^^

PR #19331, opened 2016-02-27
Travis: use node.js 5.6

XhmikosR:
/CC @cvrebert

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ git:
   depth: 10
 node_js:
   - "4"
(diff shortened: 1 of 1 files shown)


cvrebert:
Or we could downgrade to node.js v5.6 in Travis CI.

XhmikosR:
5.7.1 should be out around 1st March so we could wait.

XhmikosR:
Switched to 5.6. Just remember to change this after 5.7.x is out with a fix.

PR #19376, opened 2016-03-03
update dependencies

z11h:
all the dependencies have been updated :)


XhmikosR:
Thanks but I already have a [branch](https://github.com/twbs/bootstrap/tree/v3-deps) in the works for this.

PR #19408, opened 2016-03-04
Introduction for using table-reflow.

### LINKED ISSUE #18683: Docs: responsive tables reflow needs some explanatory text
http://v4-alpha.getbootstrap.com/content/tables/#reflow

Needs some general introduction of what reflow actually does, as well as some strong-ish warning that this is only applicable to a certain type / arrangement of content (where the first "column" is actually the `<thead>`, and the other columns contain data...so effectively, a traditional table turned on its side). If devs blindly take this table and stuff their "regular" table content in there, it may lead to nonsensical structures.

Pvanhesteren:
This should resolve #18683 (https://github.com/twbs/bootstrap/issues/18683)

### DIFF
--- a/docs/content/tables.md
+++ b/docs/content/tables.md
@@ -564,6 +564,16 @@ For more information, read [this Stack Overflow answer](https://stackoverflow.co
 
 
 ### Reflow
+Turn traditional tables on their side by using `table-reflow`. When using reflow, the table header becomes the first column of the table, the first row within the table body becomes the second column, the second row becomes the third column, etc.
+
+{% callout warning %}
+#### Content order and complex tables
+
+Beware that the `table-reflow` style changes the visual order of content. Make sure that you only apply this style to well-formed and simple data tables (and in particular, don't use this for layout tables) with appropriate `<th>` table header cells for each row and column.
+
+In addition, this class will not work correctly for tables with cells that span multiple rows or columns (using `rowspan` or `colspan` attributes).
+
+{% endcallout %}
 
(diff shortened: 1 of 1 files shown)


patrickhlauke:
I'd suggest rewording the first sentence slightly to

```
Turn traditional tables on their side by using `table-reflow`.
```

Having thought about it a bit further, I think the impact of people applying this incorrectly is minimal, provided that they do apply it to properly formed data tables (rather than to lay out content, using tables for layout), and only those without any cells spanning multiple rows or columns. So perhaps the last sentence could be removed, and instead add a callout along the lines of

```
{% callout warning %}
#### Content order and complex tables
 
Beware that the `table-reflow` style changes the visual order of content. Make sure that you only apply this style to well-formed and simple data tables (and in particular, don't use this for layout tables) with appropriate `<th>` table header cells for each row and column.

In addition, this class will not work correctly for tables with cells that span multiple rows or columns (using `rowspan` or `colspan` attributes).
{% endcallout %}
```

Pvanhesteren:
@patrickhlauke, thanks for the quick response. I'll make the changes tomorrow! :+1:

florisverburg:
@patrickhlauke, everything is improved according to your suggestions.

patrickhlauke:
Lovely. One last request: could you squash this into a single commit?

florisverburg:
@patrickhlauke, done!

patrickhlauke:
Perfect, thank you

PR #19414, opened 2016-03-05
Add variables for controlling button active/hover darken amount

### LINKED ISSUE #19353: Ability to choose how much darker the button is on active
For the moment the button gets darken by 10% when active, for us the shade difference between active and non-active is too small. That'd be nice to be able to be able to change that.

kevinkucharczyk:
fixes #19353


sqal:
:+1: or maybe just allow user to take full control of the bg color, border-color, and text color for hover state? and not just the amount of darken ? I personally always needed this kinda control of the buttons, like since BS 3..

provenic:
I personally need buttons which are lightened on hover, not darkened. Why not to give full control of button hover, focus and active states through variables? Like
    background-color:: $btn-hover-bg;
    border-color :$btn-hover-border;
etc.

mdo:
Closing per @sqal's comment. This is the second or third PR I've seen deal with this. Opening a new issue to track and see what'd be more doable/helpful. Thanks though!

PR #19420, opened 2016-03-06
Title attribute should never be an empty string

majornista:
After storing the original `title` attribute, the `title` attribute should not be replaced with an empty string, which would result in an accessibility violation on elements, like `input`, that may use the title for a label. This seems to have been improperly translated from the source material: https://github.com/jaz303/tipsy/blob/master/src/javascripts/jquery.tipsy.js#L94.

### DIFF
--- a/js/tests/unit/tooltip.js
+++ b/js/tests/unit/tooltip.js
@@ -40,7 +40,7 @@ $(function () {
   QUnit.test('should empty title attribute', function (assert) {
     assert.expect(1)
     var $trigger = $('<a href="#" rel="tooltip" title="Another tooltip"/>').bootstrapTooltip()
-    assert.strictEqual($trigger.attr('title'), '', 'title attribute was emptied')
+    assert.strictEqual($trigger.attr('title'), undefined, 'title attribute was removed')
   })
 
   QUnit.test('should add data attribute for referencing original title', function (assert) {
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -343,7 +343,7 @@
   Tooltip.prototype.fixTitle = function () {
     var $e = this.$element
     if ($e.attr('title') || typeof $e.attr('data-original-title') != 'string') {
-      $e.attr('data-original-title', $e.attr('title') || '').attr('title', '')
+      $e.attr('data-original-title', $e.attr('title') || '').removeAttr('title')
     }
   }
 


twbs-savage:

Commit: cf638bd2349c1e945684ad2bc041a0bb1fda25af
Build details: https://travis-ci.org/twbs-savage/bootstrap/builds/114085461

twbs-savage:

Commit: 18f0c772f5cb1ceea716d1836878dcf4f6a98581
Build details: https://travis-ci.org/twbs-savage/bootstrap/builds/114087000

twbs-savage:

Commit: d24890878008ddc6dd2b713e310e544d2932922c
Build details: https://travis-ci.org/twbs-savage/bootstrap/builds/114088655

patrickhlauke:
> would result in an accessibility violation on elements, like input, that may use the title for a label.

do you have a test case, and an actual combination of browser/assistive technology which is known to cause a problem?

from cursory testing (just using NVDA and JAWS and Firefox), and from looking at the code itself that's generated, everything should actually work fine since despite the `title` being set to empty, the tooltip itself is referenced with `aria-describedby` and thus the text which formerly was in the `title` is announced once the control receives focus.

it may arguably be cleaner though to remove the `title` attribute altogether, rather than blanking it. @cvrebert thoughts?

patrickhlauke:
(though this is currently tagged/aimed at v3, this could probably be ported to v4)

majornista:
@patrickhlauke It came to my attention because it was flagged as an issue by an accessibility validation tool, either aXe or HTML_CodeSniffer. Now, in practice it probably wouldn't cause harm because the tooltip is referenced with `aria-describedby`, but it would be better not to throw false positives. Also, I think the explanation in this case would lie with the developer who originally ported code that removes the `title` attribute entirely and replaced it with code that adds `title=''`.

patrickhlauke:
> it would be better not to throw false positives

do these tools still flag false positives after your proposed PR? also, it would be good to flag the fact that the tools have false positives to the tools themselves.

as for referring back to the original tipsy, note that the code there has not been touched in 3 1/2 years, and it currently has issues of its own - so it's not a completely reliable source of best practice.

majornista:
I think the place in the code I've been remediating where validators identified this as an issue was an icon `button` element that used a `title` attribute and the bootstrap tooltip class to display the label as a tooltip. I've since refactored to use a span of "hidden accessible" text as the label instead of the `title` attribute, but found it odd to still find an empty `title` attribute in the DOM.

I only referred to tipsy because I was doing some forensics to try to understand the rationale for including `title=""` after caching the existing title in a data attribute. Now it could be that for some performance reason, the person who ported the code chose to change the value of the attribute rather than remove it, but we'll probably never know.

Also, the Tooltip class only adds `aria-describedby` after showing the tooltip element, so with `title=""` we have an attribute that could serve as a label for an element with an empty string for a value.

majornista:
I'm also happy to make the change in the v4 branch.

patrickhlauke:
> Also, the Tooltip class only adds aria-describedby after showing the tooltip element, so with title="" we have an attribute that could serve as a label for an element with an empty string for a value.

but with your proposed fix, the `title` attribute would be removed altogether...so i don't quite follow what the functional difference here would be (an empty string vs no attribute at all)?

majornista:
@patrickhlauke Would you prefer that had I not wasted your time by identifying this as an accessibility issue, given that there is probably little functional difference between removing the title attribute and setting it to an empty string? I guess my question is why one would prefer setting the title attribute to an empty string as opposed to removing it, and without clear justification for it, why would one repeat the pattern in v4?

patrickhlauke:
>Would you prefer that had I not wasted your time by identifying this as an accessibility issue, given that there is probably little functional difference between removing the title attribute and setting it to an empty string?

this isn't about wasting my or anybody else's time. i'm just asking what difference it makes, from an accessibility point of view, whether there's an empty `title` or no `title`. if none, then i'm just wondering if this is not simply personal preference (not having the attribute, versus having the attribute but empty).

MichielBijl:
As far as I know, there is no accessibility issue here. From the [HTML Accessibility API Mappings (AAM)](http://w3c.github.io/aria/html-aam/html-aam.html#input-type-text-input-type-password-input-type-search-input-type-tel-input-type-url-and-textarea-element):

> 1. If the control has an aria-label or an aria-labelledby attribute the accessible name is to be calculated using the algorithm defined in Accessible Name and Description: Computation and API Mappings 1.1.
> 2. Otherwise use the associated label element(s) accessible name(s) - if more than one label is associated; concatenate by DOM order, delimited by spaces.
> 3. Otherwise use the placeholder attribute
> 4. Otherwise use the title attribute

So in any accessible name calculation, the `title` attribute is well behind `aria-label` or `aria-labelledby`.

That said, I do agree it would be cleaner to remove the title attribute.

majornista:
>  if none, then i'm just wondering if this is not simply personal preference (not having the attribute, versus having the attribute but empty).

It may have been to some extent personal preference, but now after rereading [3.2.3.2 The title attribute](https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#the-title-attribute), I think I have a better understanding of why `title=""` was used.

> If this attribute is omitted from an element, then it implies that the title attribute of the nearest ancestor HTML element with a title attribute set is also relevant to this element. **Setting the attribute overrides this, explicitly stating that the advisory information of any ancestors is not relevant to this element. Setting the attribute to the empty string indicates that the element has no advisory information.**

So setting the `title` attribute to an empty string explicitly overrides any `title` attribute on an ancestor element, which prevents an ancestor's native tooltip from displaying at the same time as the Bootstrap Tooltip. 

I stand corrected and humbly withdraw this PR.

MichielBijl:
Just for archival purposes, HTML spec linked above is an old working draft. Same sections in the [HTML 5.1 Editor's Draft](http://w3c.github.io/html/dom.html#the-title-attribute) and the [HTML 5 Recommendation](https://www.w3.org/TR/html5/dom.html#the-title-attribute).

PR #19541, opened 2016-03-14
Update tooltip.js

tracker1:
Don't reference `Tether` via attachment to `window`, with the update one can import bootstrap providing the dependencies in webpack with:

``` js
new webpack.ProvidePlugin({
  $: 'jquery',
  jQuery: 'jquery',
  Tether: 'tether',
});
```

Then inside one's own bootstrap/globals, `import 'bootstrap';` will simply work, and $/jQuery can be used from there.

I had wanted to do this, but also expose jQuery, Tether, etc when in development build in my code, but if I provide `window.Tether`, I can't then expose it to the outside...

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -16,7 +16,7 @@ const Tooltip = (($) => {
    * Check for Tether dependency
    * Tether - http://github.hubspot.com/tether/
    */
-  if ('undefined' === typeof Tether) {
+  if (typeof Tether === 'undefined') {
     throw new Error('Bootstrap tooltips require Tether (http://github.hubspot.com/tether/)')
   }
 
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -16,7 +16,7 @@ const Tooltip = (($) => {
    * Check for Tether dependency
    * Tether - http://github.hubspot.com/tether/
    */
-  if (window.Tether === undefined) {
+  if ('undefined' === typeof Tether) {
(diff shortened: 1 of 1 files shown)


vsn4ik:
http://jsperf.com/is-undefined

js/src/tooltip.js
```diff
@@ -16,7 +16,7 @@ const Tooltip = (($) => {
    * Check for Tether dependency
    * Tether - http://github.hubspot.com/tether/
    */
-  if (window.Tether === undefined) {
+  if (typeof Tether === 'undefined') {
```
Johann-S: Why not just : ``if (Tether === undefined)`` ?
Your JSPerf prove it's faster than using ``typeof``

tracker1: @Johann-S  That will throw an exception if Tether isn't defined...

tracker1:
@vsn4ik in your example `undef` is defined, it's value is initially set to `undefined` so that the test passes... if you try to run a truly undefined/undeclared value (no var declaration)...

```
if (iHaveNotBeenReferencedOrDefinedAnywhere == undefined) console.log('you win');
```

You will see...

```
Uncaught ReferenceError iHaveNotBeenReferencedOrDefinedAnywhere is not defined
```

Since `Tether` is supposed to be declared in another library, and will be undeclared if it isn't included, then doing `if (Tether === undefined) ...` will throw an uncaught exception.

mdo:
@bardiharborow @Johann-S Do we need this? I'm unclear if this helps us at all.

bardiharborow:
I'd say close in favor of #20745.

PR #19548, opened 2016-03-15
Use $card-border-radius-inner for card-img border-radius

safrazik:


mdo:
This was already done at some point, thanks though!

mdo:
See #20667.

PR #19582, opened 2016-03-22
Remove nonsense text

elisbyberi:
Removed "fade-in" at line 621!


mdo:
Not nonsense: http://sass-lang.com/documentation/Sass/Script/Functions.html#fade_in-instance_method

PR #19618, opened 2016-03-25
Revert "Use vanilla npm shrinkwrap instead of uber/npm-shrinkwrap"

ghost:


cvrebert:
Going to have you blocked from the repository now. We've warned you repeatedly.

Issue #18650, opened 2015-12-23
Default outline on links and buttons on mobile


patrickhlauke:
In Microsoft Edge / Windows 10 Mobile, the browser applies the default dotted outline to links - while Safari/iOS and Chrome/Android don't. This can lead to some unsightly effects after tapping pagination links, for instance.

![outline-win10mob1](https://cloud.githubusercontent.com/assets/895831/11979077/45bfdd86-a988-11e5-8a53-fd59fa8ff868.png)

![outline-win10mob2](https://cloud.githubusercontent.com/assets/895831/11979081/4960face-a988-11e5-94ca-f432752dfb55.png)

![outline-win10mob3](https://cloud.githubusercontent.com/assets/895831/11979084/4d401634-a988-11e5-8181-db77cc343a1e.png)

Related, outline is also shown for buttons both on Microsoft Edge / Windows 10 Mobile and Chrome / Android.

![outline-win10mob-button](https://cloud.githubusercontent.com/assets/895831/11979110/7e8fce46-a988-11e5-998e-62a48b7f13aa.png)

![outline-android-button1](https://cloud.githubusercontent.com/assets/895831/11979165/fa857898-a988-11e5-86f9-610a553ae808.png)

![outline-android-button2](https://cloud.githubusercontent.com/assets/895831/11979167/fe690efc-a988-11e5-9352-58cddc567f81.png)

On iOS, links and buttons do not receive any standard outline, and the same for links in Chrome / Android.

As both pagination links and buttons provide some fairly clear change in style when focused already, wondering if we could/should explicitly suppress outline on them. May be a mild accessibility-related issue, though.

bassjobsen:
about the outline for the buttons. This outline is set by the `tab-focus()` mixin in the `scss/mixins/_tab-focus.scss` file. It look like that shown beneath:

    @mixin tab-focus() {
      // Default
      outline: thin dotted;
      // WebKit
      outline: 5px auto -webkit-focus-ring-color;
      outline-offset: -2px;
    }

The  `outline: 5px auto -webkit-focus-ring-color;` makes the outlines look different as explained at http://stackoverflow.com/a/11929801/1596547

If we do need an outline for the focus and active states i can not find a reason to `outline: 5px auto -webkit-focus-ring-color;` and not only `outline: thin dotted;`. The `outline: 5px auto -webkit-focus-ring-color;` looks different on diffent browsers.

The reason to use a outline will be explained in `scss_reboot.sccs`:

[...]

Issue #17006, opened 2015-08-18
Dotted border issue in firefox


Hemanth089:
When click happens in all elements a dotted border is added in firefox browser? Can you add necessary css to fix this?

kkirsche:
What version of Firefox, what operating system and version, and would you be able to post a isolated example on JSBin

Hemanth089:
I'm using Firefox 39.0.3. In JSBin it is working correctly. But in my local it shows the dotted border. I'm using bootstrap  v3.3.5.

Hemanth089:
Check out the output in JsBin example in firefox 39.0.3 browser 

http://output.jsbin.com/cokiyavogo

patrickhlauke:
This is standard behavior for focused elements in Firefox and Internet Explorer. Chrome tries to do some trickery in not applying outlines on focusable elements as a result of a mouse interaction, but it used to behave exactly the same way. In fact, the `outline` property is explicitly defined for focused links and `.btn` elements in the CSS. In short, this is the intended behavior.

patrickhlauke:
We may look at working around this some more in Bootstrap v4 (off the top of my head, a fix for this is to NOT specifiy an explicit `a:focus, .btn:focus { outline:dotted; }` and to leave it up to the browser to do its own standard outline, which it can then optimize for mouse use), but for v3, we're not likely to make any changes to this I'm afraid.

PR #19693, opened 2016-04-08
Color contrast tweaks for V4

patrickhlauke:
A series of (hopefully non-controversial) tweaks to some common color contrast issues that V4 inherited from V3.

Some of these tweaks aim for the full 4.5:1 color contrast ratio recommended by WCAG 2.0, while others (notably the success, info and warning tweaks) go for the slightly more relaxed 3:1 (which is really for large text, so still not ideal in all situations...but a slightly more achievable goal, particularly for vibrant colors).

Not all possible combinations have been tried out, but most tweaks seem to move towards a more positive result judging purely by the documentation examples.

PR #19708, opened 2016-04-11
Remove tab-focus mixin and add surgical `outline: none` where appropriate

patrickhlauke:
Explicitly setting a particular outline (e.g. the default `outline: thin dotted`) overrides heuristics in some modern browsers (such as Firefox) that automagically suppress `outline` as a result of mouse/tap interaction (while retaining it for keyboard interaction).

This removes the generic `tab-focus` mixin (mostly leaving outline handling up to the browser and its own heuristics), but also adds explicit `outline: none` in cases where it's deemed "safe", i.e. where sufficiently clear focus indication is provided (`.btn` elements, pagination links)

Fixes https://github.com/twbs/bootstrap/issues/18650 / https://github.com/twbs/bootstrap/issues/17006 and the Firefox focus part of https://github.com/twbs/bootstrap/issues/19398

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -22,7 +22,7 @@
   &.active {
     &:focus,
     &.focus {
-      @include tab-focus();
+      outline: none;
     }
   }
 
--- a/scss/_mixins.scss
+++ b/scss/_mixins.scss
@@ -23,7 +23,6 @@
 @import "mixins/resize";
 @import "mixins/screen-reader";
 @import "mixins/size";
-@import "mixins/tab-focus";
 @import "mixins/reset-text";
 @import "mixins/text-emphasis";
 @import "mixins/text-hide";
--- a/scss/_nav.scss
+++ b/scss/_nav.scss
@@ -71,6 +71,10 @@
       border-color: $nav-tabs-link-hover-border-color $nav-tabs-link-hover-border-color $nav-tabs-border-color;
     }
 
+    &:focus {
+      outline-offset: -2px;
+    }
+
     &.disabled {
       @include plain-hover-focus {
         color: $nav-disabled-link-color;
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -56,6 +56,7 @@
     color: $pagination-hover-color;
     background-color: $pagination-hover-bg;
     border-color: $pagination-hover-border;
+    outline: none;
   }
 }
 
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -165,10 +165,6 @@ a {
     color: $link-hover-color;
     text-decoration: $link-hover-decoration;
   }
-
-  &:focus {
-    @include tab-focus();
-  }
 }
 
 // And undo these styles for placeholder links/named anchors (without href).
@@ -183,10 +179,6 @@ a:not([href]) {
   @include hover-focus {
     color: inherit;
     text-decoration: none;
-  }
-
-  &:focus {
-    outline: none;
   }
 }
 
--- a/scss/bootstrap-reboot.scss
+++ b/scss/bootstrap-reboot.scss
@@ -5,7 +5,6 @@
 @import "custom";
 @import "variables";
 @import "mixins/hover";
-@import "mixins/tab-focus";
 
 @import "normalize";
 @import "reboot";
--- a/scss/mixins/_tab-focus.scss
+++ b/scss/mixins/_tab-focus.scss
@@ -1,9 +1,0 @@
-// WebKit-style focus
-
-@mixin tab-focus() {
-  // Default
-  outline: thin dotted;
-  // WebKit
-  outline: 5px auto -webkit-focus-ring-color;
-  outline-offset: -2px;
-}


cvrebert:
Removing the `outline-offset` makes the outline look clunkier in WebKit/Blink:
without

Original:
with

cvrebert:
I believe `.tab-focus()` was originally used for the `:focus` styles for the Tabs widget (hence the name), and then it got applied more broadly to other widgets.

patrickhlauke:
> Removing the outline-offset makes the outline look clunkier in WebKit/Blink

but as I explicitly set an `outline: none` on those `.btn`s, you shouldn't be seeing any outline, clunky or otherwise...

cvrebert:
(First introduced in 7e72eb5063470a3728d612acd2cc0368aabc8664)

patrickhlauke:
(as an aside, i really wish we had https://github.com/ten1seven/what-input natively, in order to explicitly style outline but only for keyboard users...maybe worth considering adding this tiny library?)

patrickhlauke:
> I believe .tab-focus() was originally used for the :focus styles for the Tabs widget (hence the name), and then it got applied more broadly to other widgets.

as tabs don't exist in v4 anymore, and the cases where `tab-focus` was now used in v4 can actually do without an outline being forced, it seems to me that removing the mixin and adding surgical outline suppression seems not to cause any major problem. @mdo ?

cvrebert:
Ah, I see. You think the background+border+text color change alone is sufficiently clear & noticeable accessibility-wise (or will be after #19693)?
(I always thought the translucent blue outline was a nice touch, visually.)

cvrebert:
> as tabs don't exist in v4 anymore

Erm, yeah they do: http://v4-alpha.getbootstrap.com/components/navs/#tabs

patrickhlauke:
> Ah, I see. You think the background+border+text color change alone is sufficiently clear & noticeable accessibility-wise (or will be after #19693)?
(I always thought the translucent blue outline was a nice touch, visually.)

Yes, I'd say it's a sufficiently strong visual indication of focus. Agree that having an additional blue outline is a nice touch (note that it was never blue/translucent in Firefox/IE/Edge, so just a webkit/blink specific "niceness"), but as the approach causes unwanted outlines in Firefox and others following a mouse or tap interaction (e.g.  #18650 / #17006 ) I think - unless we can explicitly add those styles purely for keyboard interaction, using something like https://github.com/ten1seven/what-input - it's a small loss of niceness to gain a more consistent experience.

patrickhlauke:
> Erm, yeah they do: http://v4-alpha.getbootstrap.com/components/navs/#tabs

doh, for some reason I was looking for tabs in the list of components...need to check how this change affects them

patrickhlauke:
The `outline-offset` was indeed needed to get the tabs to look a bit nicer. However, as the negative offset also resulted in a very crowded-looking outline on regular links in the content, I opted to reintroduce it only for that specific case of `.nav-tabs .nav-link`

patrickhlauke:
This still leaves potentially unsightly outlines as a result of mouse/tap on tabs...but this is definitely the point where they can't be suppressed without completely breaking them for sighted keyboard users in absence of a more discriminating "only add outline as result of keyboard" logic.

cvrebert:
Is https://github.com/twbs/bootstrap/pull/19708/files#diff-7f9eec82312835f6155c8de839bf0e49R185 now unnecessary?

patrickhlauke:
hmm...probably yes. will check in the morning and force-push a change if so

patrickhlauke:
> Is https://github.com/twbs/bootstrap/pull/19708/files#diff-7f9eec82312835f6155c8de839bf0e49R185 now unnecessary?

It does indeed seem redundant now, so I've removed it and force-pushed the change.

patrickhlauke:
@mdo thoughts on all this before we push any further? (also quite interested to know if anything like https://github.com/ten1seven/what-input could be considered, which would make some of this a bit easier at least while browsers don't provide any of this logic natively and their heuristics for trying to do it are flaky, as in the case of Firefox)

patrickhlauke:
ping @mdo

patrickhlauke:
@mdo ?

mdo:
Was this mixin added because of our tabs component, or the tab key (because "keyboard focus")? Seems odd to only adjust a focus state like that for (variation) of a component.

mdo:
For the `what-input` bit, let's open a new issue and ask @fat.

cvrebert:
@mdo You were the one who created the mixin (see https://github.com/twbs/bootstrap/pull/19708#issuecomment-208103728 ), so if you don't know, then nobody does.

mdo:
Yeah, pretty sure that was it then. Just a not great name here; it's not limited to tabs.

patrickhlauke:
i've lost track of this...so @mdo are you ok with the PR, or what other discussion needs to happen to move this forward?

patrickhlauke:
ping @mdo

patrickhlauke:
@mdo ... any chance of a response on this? been trying to pin you down in various places ...

mdo:
If I understand it right, this PR doesn't make sense to ship as-is.

The `tab-focus` mixin wasn't created for our tabs component or any other navigation element. It was named for hitting the Tab key on your keyboard (I know, I know). It's a global change we've applied for a very long time to improve the display of those outlines.

patrickhlauke:
@mdo ok I was just trying to ascertain what the purpose of the mixin was. If it IS purely about setting focus styles, then see my rationale for the change right at the start:

> Explicitly setting a particular outline (e.g. the default `outline: thin dotted`) overrides heuristics in some modern browsers (such as Firefox) that automagically suppress `outline` as a result of mouse/tap interaction (while retaining it for keyboard interaction).

patrickhlauke:
though yes at this point, 5 months later, I'd need to go back and check the actual effect this would have visually on our current codebase's output...

mdo:
I have a local copy of this branch that has conflicts resolved with the latest `v4-dev`. Are we still wanting to change this stuff up, or did anyone have any new thoughts? Only thing on my mind is that the outline colored right in these instances (easily fixed, I think) and we're missing some focus styles on inputs (likely stuck behind the `$enable-shadows` var). All fixable issues though.

patrickhlauke:
I'll make some time in the coming days to revisit this (I'll be honest, it's been so long since I proposed this that I need to refamiliarise myself with what was actually going on / what I intended to do here). But the general gist was that setting outline to none and then reenabling it specifically has side effects in browsers (it overrides the heuristic that browsers like Firefox have, which doesn't display outline at all as a result of a mouse click / tap, and only shows it as result of keyboard interaction). Also need to evaluate if using the newly spec'd `:focusring` selector (and perhaps polyfilling it for older browsers) can help https://github.com/w3c/csswg-drafts/pull/709

mdo:
Think this has all been addressed across a few PRs now.

PR #19721, opened 2016-04-12
Strikethrough-when-checked and remove-when-checked checkboxes

ghost:
I've created two new possible classes check-hide and check-strike that can be used in lists with check-boxes such as To-do lists or statistics questions.

The check-hide class hides the checkbox and it's parent.
The check-strike class on the other hand changes the line-decoration of the label(sibling) to line-through.

I've put the code under js/src/checkbox.js. Someone needs to use Grunt to regenerate the dist folder, a thing that I haven't done. I did test the code in bootstrap.js and compiled it in bootstrap.min.js and it worked perfectly(I didn't left it there after I read the guidelines so I put the code in js/src/checkbox.)


js/src/checkbox.js
```diff
@@ -0,0 +1,12 @@
+ // Checkbox classes 
++function ($) {
+$(document).ready(function() {
+         $(".check-hide input:checked").parent().remove()
+         $(".check-strike input:checked").next().css( "text-decoration", "line-through" )
+});
+$(document).on('click',function() {
+         $(".check-hide input:checked").parent().fadeOut('slow')
```
cvrebert: Bootstrap doesn't use jQuery animations (such as `fadeOut`), only CSS animations.

cvrebert: Also, this should look at the target of the click event rather than performing several global DOM queries.

cvrebert:
I'm not sure this is sufficiently common or non-trivial to need to be part of core Bootstrap.
CC: @mdo

cvrebert:
(Also, if we were to add this, it would need unit tests and documentation.)

patrickhlauke:
Feels more like a candidate for a 3rd party plugin to me, personally

HansiCollaku:
@cvrebert @patrickhlauke I understand perfectly what you mean. In any-case it might be good to get this script rewritten, documented and added to Bootstrap or it might not. I personally would like this feature I don't know about others so. And if not, it can be turned into a simple plugin which I think that this and many other kinds of 'little scripts'  like this can be turned into simple plugins that make developers lives less tedious. On the other hand adding them to Bootstrap with time can lead to a greater use and a unification of code, so that a lot more features, even minimalistic features like this and others could be accessible through Bootstrap with no need to download other script and link to them, which might turn out better.

mdo:
Punting as a won't fix/add. Thanks though!

PR #19745, opened 2016-04-15
Darken middle border of certain split button dropdowns

MatthewRayfield:
Addresses #17180

image


mdo:
Thanks, but punting as a won't fix for now.

Issue #19317, opened 2016-02-25
Contextual classes don't work properly with .table-inverse


segrax:
Not sure if this *should* work properly, but applying 'table-danger' / 'table-success' / others on a row in a 'table-inverse'... renders it unreadable

can be tested by adding any one of the classes to a `<tr>` on

http://v4-alpha.getbootstrap.com/content/tables/#inverse-table

mdo:
Coming in #19863.

PR #19755, opened 2016-04-17
Add contextual backgrounds for .table-inverse

MatthewRayfield:
Addresses #19317

image


mdo:
Dry-er fix coming in #19863. Thanks though!

PR #19756, opened 2016-04-17
remove unnecessary `/` in reference url

Rplus:
https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.2/scss%2Fmixins%2F_screen-reader.scss#L3

current reference URL of `@mixin sr-only` is 404

> http://a11yproject.com/posts/how-to-hide-content/

Just remove unnecessary `/` in the end.
http://a11yproject.com/posts/how-to-hide-content


patrickhlauke:
good catch, thanks

PR #19832, opened 2016-05-01
fix the offcanvas example

bassjobsen:
prevent that the sidebar is not hidden on the small (sm) grid


XhmikosR:
@bassjobsen: can you fetch and rebase?

mdo:
Pulling this into #21407. Thanks!

PR #19828, opened 2016-04-30
added variables to fix card-outline-*

indymx:
added

 border-style: $card-border-style;
 border-width: $card-border-width;

 to @mixin card-outline-variant

 added

 $card-border-style:   solid !default;

 to _variables.scss

 so card-outline-\* would work

@mdo this fixes https://github.com/twbs/bootstrap/issues/19097

PR #19861, opened 2016-05-08
Jank card borders were too janky

### LINKED ISSUE #19097: .card-outline-* classes no longer work

mdo:
- Drops the experiment I had going for `box-shadow`-powered borders
- Reinstates regular `border` using existing variables

Fixes  #19097 and #19143. Nullifies #19828.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -9,8 +9,7 @@
   background-color: $card-bg;
   // border: $card-border-width solid $card-border-color;
   @include border-radius($card-border-radius);
-  // Doesn't use mixin so that cards always have a "border"
-  box-shadow: inset 0 0 0 $card-border-width $card-border-color;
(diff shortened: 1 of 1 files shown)


Starsam80:
This only partially fixes #19143, as there is still box-shadows on lines [76](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_card.scss#L76) and [89](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_card.scss#L89) (`.card-header` and `.card-footer`), also causing them to [look a bit wierd](http://i.imgur.com/KdLM7A9.png).

mdo:
@Starsam80 Followed up on that with #19885.

PR #19862, opened 2016-05-08
Update Normalize to v4.0.0

mdo:
Supersedes #19639.

Had to address a number of changes that caused changes in our code, including:
- Table's no longer collapse borders
- Code blocks no longer automatically overflow

Also updated our build config so that our linter ignores all of Normalize's syntax problems.

/cc @marktopper


mdo:
I imagine we'll find more bugs related to this down the line, but getting this merged in the mean time.

PR #19864, opened 2016-05-09
Dropdown: remove dependency to role="menu", role="listbox" a and li elements => fix keyboard navigation

pvdlg:
Per migration documentation in https://github.com/twbs/bootstrap/commit/4d8d8bdab4b0e2931fcf92ce30d772b05a989434:

> - Rebuilt dropdown styles and markup to provide easy, built-in support for `<a>` and `<button>` based dropdown items.
> - Dropdown items now require `.dropdown-item`.

Per https://github.com/twbs/bootstrap/issues/19034#issuecomment-179749958 role="menu" should have been dropped.

This PR:
- Remove the dependency to `role="menu"` or `role="listbox"` and rely on the class `dropdown-menu` instead
- Remove the dependency to the `a` element for the dropdown-item and rely on the class `dropdown-item` instead
- Remove the necessity to wrap dropdown-item in `li` elements

All this problems affected only the keyboard navigation in the dropdown.
With this modification the javascript now matches the markup in the documentation.
The keyboard navigation works again now in [dropdowns](http://v4-alpha.getbootstrap.com/components/dropdowns/).

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -44,11 +44,9 @@ const Dropdown = (($) => {
     BACKDROP      : '.dropdown-backdrop',
     DATA_TOGGLE   : '[data-toggle="dropdown"]',
     FORM_CHILD    : '.dropdown form',
-    ROLE_MENU     : '[role="menu"]',
-    ROLE_LISTBOX  : '[role="listbox"]',
+    MENU          : '.dropdown-menu',
     NAVBAR_NAV    : '.navbar-nav',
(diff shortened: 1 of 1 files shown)


vanduynslagerp:
Opened #21535 that fix the conflict.
For some reason I couldn't update this PR as I might have delete the source branch.

jump020305:
@all can you help me with this one http://stackoverflow.com/questions/41645884/accessibility-for-select-fields-in-ie11

PR #19869, opened 2016-05-09
V4: Allow for background options on a card

mkroeders:
`$card-bg` implies to me I can use the background options available, ie images. But currently it only allows colors

In a standard `<div class="card"></div>`, there are no other `background` attributes so this change should not break any other thing.


mdo:
We can change the variable if enough folks are at odds with it, but we won't be changing to a shorthand notation where we don't need it. For now, we'll stick with `background-color`. Thanks though!

PR #19874, opened 2016-05-10
Make named anchor/placeholder link style reset more specific

patrickhlauke:
This avoids applying the reset to named anchors/placeholder links (links without an `href`) that have explicitly been made keyboard-focusable (using `tabindex`). This is not fool-proof - it's not easy/straightforward to check for the actual `tabindex` value itself, to ensure it's positive, not will this apply if a link has been "blessed" with `tabindex` via JS. However, this should catch most common uses (and gives a reasonably valid way around the issue for developers who, for whatever reason, DO want to use links without `href` - as side effect, it forces best practice of at least ensuring these links can also be focused with the keyboard)

(Partially) addresses concerns raised in https://github.com/twbs/bootstrap/pull/19411#issuecomment-218106243

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -171,12 +171,13 @@ a {
   }
 }
 
-// And undo these styles for placeholder links/named anchors (without href).
+// And undo these styles for placeholder links/named anchors (without href)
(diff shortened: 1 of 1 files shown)


mdo:
Seems fine by me!

Issue #19883, opened 2016-05-11
width scrolling problem with .card-deck


rubenscd:
I have a problem with the card-deck, he is breaking the width of the size of the body / html, creating horizontal scrolling, following image and test html

![card-deck-size-problem](https://cloud.githubusercontent.com/assets/40513/15190558/e19b6f92-1786-11e6-8d5c-ee17a8d82ecc.PNG)

`
<!DOCTYPE html>
<html lang="pt-br">
  <head>
    <title>test nivo</title>
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.2/css/bootstrap.min.css">
  </head>
  <body>
  <div class="container-fluid">
    <div class="card-deck-wrapper">
      <div class="card-deck">
        <div class="card">
          <div class="card-block">
            Card 1
          </div>
        </div>
        <div class="card">
          <div class="card-block">
            Card 2
          </div>
        </div>
        <div class="card">
          <div class="card-block">
            Card 3
          </div>
        </div>
        <div class="card">
          <div class="card-block">
            Card 4
          </div>
        </div>
      </div>
    </div>
  </div>
  <script src="https://code.jquery.com/jquery-2.2.3.min.js"></script>
  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.2/js/bootstrap.min.js"></script>
  </body>
</html>
`

RyanZim:
Please use a JSFiddle next time. I created one using your code (https://jsfiddle.net/Lhuq3frk/) and am unable to reproduce in Firefox 46.0.1 on Windows 7. What browser & OS are you using?

rubenscd:
@RyanZim Windows 10 - Google Chrome 50.0.2661.94 m (64-bit) - and Mozilla Firefox 46.0.1 also

RyanZim:
@rubenscd Is that tested with the JSFiddle or with your own code?

rubenscd:
@ryanzim both

RyanZim:
OK, still can't reproduce. Any insight @cvrebert?

mdo:
This is a mismatch of grid and card deck values in our CSS. Need to rework those to match the grid for something like this.

mathmatrix828:
+1 on this issue. Ran into it when creating dynamic content via PHP, adding in a row causes all kinds of formatting problems as well

ve3:
[...]

PR #19886, opened 2016-05-12
Fix width scrolling problem with .card-deck

ghost:
Updated value of space between cards to 0.9375rem which is 15px

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -218,7 +218,7 @@
   }
 } @else {
   @include media-breakpoint-up(sm) {
-    $space-between-cards: (2 * $card-deck-margin);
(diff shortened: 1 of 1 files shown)


cvrebert:
> Fix width scrolling problem

@anupjon What scrolling problem are you talking about?

> Updated value of space between cards to 0.9375rem which is 15px

Why 15px specifically?

anupjon:
https://github.com/twbs/bootstrap/issues/19883

ckugot:
This fixes the issue on Mac Chrome as well.

mdo:
Closing out since I think I want to do this with the grid instead of extra classes; plus this user is no more given the `@ghost` account.

PR #19887, opened 2016-05-12
Alphabetized carousel options & added "ride" to it

ryanburnett:
This PR adds "ride" documentation to the carousel options at:
http://v4-alpha.getbootstrap.com/components/carousel/

The "ride" values are:
- **"carousel"**: autoplays the carousel on load
- **false**: autoplays the carousel after the user manually cycles the first item

Additionally, the options table was in a seemingly random order so I alphabatized it. If this order was intentional please let me know and I'll restore the original order and add the "ride" entry at the bottom of the table.


mdo:
Dope! Thanks <3

PR #19935, opened 2016-05-19
Remove forced default focus outline

### LINKED ISSUE #19933: Mac OS X Firefox showing “double” focus indicator: blue glow plus dotted outline
Bootstrap is adding the dotted outline unnecessarily on focused radio buttons and checkboxes. No other major Mac browser does this.

screen shot 2016-05-18 at 5 34 30 pm

This may be a dupe of #19398.

patrickhlauke:
the current `outline: thin dotted` that is forced as "default" is not in fact default in Firefox/OS X; forcing it leads to unsightly dotted outline that's non-standard for the browser/platform.

removing it seems to have no adverse effect in other all situations where this mixin is used (buttons, scaffolding for general link styles, and the case of forms). attempted to use `outline: initial` at first, but this seems somehow broken in Firefox (ends up removing standard Firefox dotted outline on links).

This is mostly a bug-fix, rather than a new feature (hence targetting it for v3)

Closes https://github.com/twbs/bootstrap/issues/19933

### DIFF
--- a/less/mixins/tab-focus.less
+++ b/less/mixins/tab-focus.less
@@ -1,9 +1,9 @@
 // WebKit-style focus
 
 .tab-focus() {
-  // Default
-  outline: thin dotted;
-  // WebKit
+  // WebKit-specific - other browser will keep their default outline style
+  // (initially tried to also force default via `outline: initial`
+  // but this seems to erroneously remove the outline in Firefox altogether)
   outline: 5px auto -webkit-focus-ring-color;
   outline-offset: -2px;
 }


cvrebert:
Any adverse effect on IE or Win Firefox?

patrickhlauke:
will post some screenshots in a bit, but generally: no adverse effect (though subtly different, checkboxes, tabs - in navs - etc all still have a good focus indication)

patrickhlauke:
some before/after shots.

for IE11, removing that explicit outline effectively brings the outline back to its natural look (i.e. links, buttons and checkboxes do have that particular outline normally)

> ie11

for Firefox/Windows, the notable change is that `<button>` elements now have no forced outline - however, this is in line with general Firefox behavior once a particular `:focus` style has been set (FX won't show default dotted outline - though it would show an inner dotted outline around the actual text, which is suppressed by normalize.css). I'd argue it's still distinctive enough, though could be stronger...

> fx

patrickhlauke:
@mdo @cvrebert thoughts on this?

cvrebert:
Ported to v4 in #20275.

PR #19941, opened 2016-05-20
Add mixin to build button with params

Hightrack:
This pull request adds a new mixin for building a button with the following params:
- Line height
- Padding y
- Padding x
- Font size
- Radius

In case you want to create a different type of basic button (a rounded button for instance) this is more scalable because you can create a new CSS class for this new button with that mixin.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -5,47 +5,7 @@
 //
 
 .btn {
-  display: inline-block;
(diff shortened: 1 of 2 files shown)


mdo:
Probably not going to do this for v4—if we do one mixin-based abstraction, we'll likely be asked to do it for many other components. Thanks though!

PR #20017, opened 2016-06-01
Collapsable → Collapsible

dereckson:
Typo in panels.less code comment.


dereckson:
The typo doesn't exist in v4-dev, only in master (v3).

Issue #17371, opened 2015-08-27
Dropdown items in nav tabs can get stuck in .active mode


rebot:
bug bootstrap alpha v4 - http://v4-alpha.getbootstrap.com/components/navs/#javascript-behavior When you press a button/tab in the dropdown, try another tab and go back, you will see that the button/tab under the dropdown hasn't been deactivated. I was implementing this in a project and came across this bug.

EDIT: It's worse.. You can't even 'reactivate' the tab. So it's stuck on 'active' mode and won't change your content in the tab view

crehard:
![screen shot 2015-09-04 at 15 10 56](https://cloud.githubusercontent.com/assets/1673893/9684340/2cf95e16-5317-11e5-98a9-3b8a92e24659.png)

JoyceBabu:
Until the code is fixed, the issue can be worked around by removing the `active` class from `shown` event

	$('.nav-tabs').on('shown.bs.tab', 'a', function (e) {
		if (e.relatedTarget) {
			$(e.relatedTarget).removeClass('active');
		}
	})

JD-Robbs:
Thanks for the workaround! :+1:

HallofFamer:
Any updates on this issue? Still not yet fixed?

rebot:
There is a workaround suggested, but this problem is not yet fixed and was added to milestone v4.0.0-alpha.3... Hopefully, they fix it!

HallofFamer:
Note JoyceBabu's solution does not work unless you wrap it in $(document).ready(), like this:
 
    $( document ).ready(function() {
        $('.nav-tabs').on('shown.bs.tab', 'a', function (e) {
            console.log(e.relatedTarget);
            if (e.relatedTarget) {
                $(e.relatedTarget).removeClass('active');
            }
        });    
    });

JoyceBabu:
@HallofFamer - Depends on where you place the code. It is almost always better to place the script tags at the end of the page, before the closing `</html>` tag. In that case, you don't need to wrap the code in `$(document).ready` or `$()` (unless it is a dynamic element created during dom ready).

bpierson:
I was able to solve this with a combination of a workaround and changing the selector on line 2554.

Workaround:
```
$(document).ready(function () {
    $('.nav').on('shown.bs.tab', 'a', function (e) {
        console.log(e.relatedTarget);
        if (e.relatedTarget) {
            $(e.relatedTarget).removeClass('active in');
        }
    });
});
```
[...]

PR #20039, opened 2016-06-04
Fixing bug described here:  Multiple Tabs can become .active at once #20027

Manuellama:
Fixing bug described here:

Multiple Tabs can become .active at once #20027


maxbeatty:
Related to #17371

mdo:
Merged #17642 awhile back, so this should be fixed. Thanks though!

Issue #19813, opened 2016-04-27
data-target with multiple target doesn't work


stassinari:
When trying to use multiple targets with the data-target attribute, it seems like only the first one is triggered and the remaining ones are discarded.
This happens with either a class targeting multiple elements, or with multiple IDs as attribute value.

I've created two pens on CodePen to reproduce the behaviour.
The HTML is identical, but the configuration points to different Bootstrap versions:

v4 with bug: http://codepen.io/stassinari-andiamo/pen/jqvgPW
v3 working: http://codepen.io/stassinari-andiamo/pen/dMqxYj

There's a (closed) issue raised for version 3.2.0 (#18342), where another user finds this behaviour in v4. There's a snippet of code in there, pointing to the actual JS code responsible for the handling.

I hope this feature will be kept (unlike @cvrebert), because being able to target multiple elements without the use of JavaScript is a nice feature, IMO.

RyanZim:
> being able to target multiple elements without the use of JavaScript is a nice feature, IMO.

I agree, however, it may be wise to only accept IDs instead of arbitrary selectors (to avoid unwanted conflicts). How about this:
```
<a data-toggle="collapse" data-target="#c1 #c2">Click me</a>

<p id="c1" class="collapse collapsible">First paragraph</p>
<p id="c2" class="collapse collapsible">Second paragraph</p>
```

stassinari:
I understand that accepting only IDs is safer, but there are cases (like in the page I'm working on now) where filling a page with dozens of IDs doesn't really make sense.
What I did is create a custom data attribute mimicking the data-target I'd like to use (the one with a class), and then reimplement the collapsing logic in my own JS.

stassinari:
I'd like to point out that, regardless of the class dispute, this is still an issue, because multiple IDs are not currently working with the alpha version.
I've updated the Pens to reflect this.

jitendra8799:
<a data-toggle="collapse" data-target="#c1,#c2">Click me</a>

mdo:
I feel like this is an edge case we don't want to explicitly support. Closing out for now, but happy to revisit if enough folks would like to see this changed. Thanks!

kutenai:
Ack! I just found this thread. Am I to understand this won't be supported in v4? This is a desirable feature if I can put my 2 cents in..

alexcasalboni:
+1

plusz:
it losk that there are also issues when using class selector instead of id
data-target='.someselector'

[...]

PR #20097, opened 2016-06-08
Multiple data-targets for collapse plugin

arnetkachev:
data-target with multiple target doesn't work #19813


mdo:
Closing per https://github.com/twbs/bootstrap/issues/19813#issuecomment-256835766.

PR #20100, opened 2016-06-09
add mixin/custom.scss file to be able to overide easily bootstrap mixins

ExFlo:
An empty custom file is available to override bootstrap's variables  without modifying source files. This pull request intend to allow user to be able to override mixin with a similar file.

### DIFF
--- a/scss/_mixins.scss
+++ b/scss/_mixins.scss
@@ -52,3 +52,7 @@
 @import "mixins/grid-framework";
 @import "mixins/grid";
 @import "mixins/pulls";
+
+// Custom (let this import at the end of the file).
+@import "mixins/custom";
+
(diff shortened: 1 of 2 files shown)


tomlutzenberger:
Why dont't you create this file by yourself?
To add and commit an empty file is IMHO pointless.

ExFlo:
Well, this kind of empty file was added also for variables (#17322) and feature was documented. But with this file you are only able to override variables and CSS class but not mixin. I think we can add the same "feature" for mixin.

tomlutzenberger:
Still don't see the point to implement this.
I always made my own style.less or *.scss and included all the libraries i needed plus my custom files.
Initial work of maybe 15min.

takuhii:
I can kinda see what exflo is getting at, but I also do what Tom is doing
and create my own _custom.scss as and when I need to

mdo:
The tendency to override the mixins seems far less common than the variables. Passing for now, though I'd be open to revisiting after v4.0.

PR #20130, opened 2016-06-16
Contextual background-inverse text color fix

jeradrutnam:
utility classes bootstrap


Francismori7:
What about `.bg-faded`?

jeradrutnam:
`.bg-faded` has a light grey background color. So text with dark grey is more readable for it. But then again it depends on where we use it.

PR #20167, opened 2016-06-24
custom-control spacing

zalog:
If we don't have `.custom-control-indicator` remains some extra padding:
http://codepen.io/zalog/pen/oLBzBq


mdo:
The padding is there to ensure that the content wraps properly when checkbox labels pass more than one line. I don't think moving it to padding on the indicator will work since those indicators are positioned absolutely.

PR #20188, opened 2016-06-27
Added breakpoint classes for margin and padding spacing

el-angel:
Currently spacing classes (padding and margin) can only be applied without breakpoints. It's currently not possible to add spacing only at certain breakpoints (xl, lg, md, sm, xs), I added the breakpoints wrapper function so it becomes possible.

It's now possible to use 
`{spacing-type}-{side}-{breakpoint}-{size}` (i.e. `m-t-xs-2`)

### DIFF
--- a/scss/utilities/_spacing.scss
+++ b/scss/utilities/_spacing.scss
@@ -9,25 +9,29 @@
   margin-left:  auto !important;
 }
 
-@each $prop, $abbrev in (margin: m, padding: p) {
-  @each $size, $lengths in $spacers {
-    $length-x:   map-get($lengths, x);
-    $length-y:   map-get($lengths, y);
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-down($breakpoint) {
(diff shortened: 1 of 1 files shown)


Snowbell92:
+ for this! I always had to write a bunch of margin-padding queries.

leggomuhgreggo:
+1

Stadly:
+1

Stadly:
I think it would be better to use `media-breakpoint-up` instead of `media-breakpoint-down` to make the margin and padding classes work the same way as the grid classes.

Alternatively, it could work like the hidden classes, with `{spacing-type}-{side}-{breakpoint}-{size}` specifying a single breakpoint, while `{spacing-type}-{side}-{breakpoint}-{size}-up` and `{spacing-type}-{side}-{breakpoint}-{size}-down` specifies a breakpoint and up/down. This would lead to a lot of classes, and would not be the best solution, in my opinion.

mdo:
This will be handled by #20926 if it lands. Thanks though!

PR #20194, opened 2016-06-28
Change broad helper to specific helper (a tiny docs edit)

frostbitten:
...since it targets just one example. It could be problematic if there were to be more usage of the bg-\* classes within the docs.

I noticed this when building a set of custom docs (extended colors) and the additional items didn't look right. I figure, what is the point of that css? I think my css selector answers that question directly


mdo:
Thanks, but I'll leave them as a specific override for now instead of the attribute selector there.

PR #20203, opened 2016-06-29
Switch to jQuery slim version.

XhmikosR:
We don't use the extra features the normal build has.

See https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/

/CC @cvrebert @hnrch02 @dmethvin for thoughts.


cvrebert:
👍

hnrch02:
SGTM. Maybe we should add a note in the docs about the slim version?

XhmikosR:
Personally, I'd like to know what are the plans for the slim version, i.e. do they plan to keep releasing it or not. Apart from that, we should probably mention that the slim version works fine with Bootstrap v4 at this point.

dmethvin:
Usually once we add a build we don't take it away. If we did stop making it I would expect that to not happen until the next major version. 

I think it's great that you're using the slim build. Most users will end up using a full build but this makes it possible for them to use slim or their own custom build.

calvinf:
This makes sense as long as the Bootstrap v4 website makes it easy and clear that either jQuery 3.0 or 3.0 slim builds can be used.

XhmikosR:
Well I'm no good with writing docs. I'd rather merge this and address the
doc changes later with someone else's help for the docs.

lorvent:
Personally i feel its better to use regular version.

even though you don't need functions like `$.ajax` those who uses bootstrap with server side languages needs them.

and if we download bootstrap through bower, if you ship slim version, people need to include regular version again.

and difference in size is just 6kb

hnrch02:
@lorvent This only touches the jQuery files used in the docs/tests.

lorvent:
oops...sorry about that

auvipy:
what about jquery 3.1?

XhmikosR:
Make a separate issue for that please.

bardiharborow:
Just a heads up that this PR changes [_config.yaml](https://github.com/twbs/bootstrap/blob/v4-dev/_config.yml), which cascades to the example on [docs/getting-started/introduction.md](https://github.com/twbs/bootstrap/blob/v4-dev/docs/getting-started/introduction.md). I think changing that without clarifying further might be inadvisable.

mdo:
Should we try to get this into Alpha 5? Seeing some conflicts preventing this from being merged.

XhmikosR:
This is obsolete and I need to update it. I'll do it tomorrow probably.

mdo:
Replaced by #21466.

PR #20215, opened 2016-07-01
Fix small grammatical mistakes

aalaap:
- Linked two closely related ideas "We don't help..." and "We just link to them."
- Replaced "just" with "only" to indicate a singular item, i.e. "We only link to them."
- Added "Have you" before the question to make it proper, i.e. "Have you finished...?" vs. just "Finished...?"

### DIFF
--- a/docs/_includes/getting-started/translations.html
+++ b/docs/_includes/getting-started/translations.html
@@ -7,6 +7,6 @@
       <li><a href="{{ language.url }}" hreflang="{{ language.code }}">{{ language.description }} ({{ language.name }})</a></li>
     {% endfor %}
   </ul>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
As the language used there is informal and in a colloquial tone, I don't really see those as mistakes.

aalaap:
I'm only trying to improve things, but I do agree with you.

XhmikosR:
The changes look good to me, as far as I'm concerned.

mdo:
Going to leave them as-is given the overall tone I'm after with most of our docs prose. Thanks though!

aalaap:
@mdo That's totally fine. Thanks for looking into it!

PR #20245, opened 2016-07-06
Update accessibility.html

ctrlaltvikas:
Some content updated with a few changes and some mistakes removed.


patrickhlauke:
Not really mistakes, more "opinionated" changes I'd say (e.g. "lots" instead of "many" is simply more informal..."many" isn't wrong; "comes before" is correct, "comes prior" is incorrect - should really just be "prior" then, but that's unnecessarily formal). And "skip to main content" is fairly standard wording - see https://www.google.co.uk/search?&q="skip%20to%20main%20content" (though yes, "to THE main content" would make a more complete sentence). As such, thank you, but I think we'll pass on these changes for now.

PR #20253, opened 2016-07-07
Docs: Optimize *.png with zopflipng -m

scop:
Before: 24352 bytes
After:  24236 bytes


XhmikosR:
I have already optimized the images myself with zopfli so I find it strange that you have different output.

What was the options you used?

XhmikosR:
So, unless I can reproduce the same results, I'm against this going in @cvrebert.

scop:
As said in the commit message, this was done with the `-m` option. zopflipng is the one that ships with zopfli 1.0.1.

XhmikosR:
You mention it in the commit message, not the PR.

Anyway, my zopfli version is older and I don't have the -m switch.

scop:
The title of the PR does mention `-m` too.

XhmikosR:
Right.

XhmikosR:
All right, I managed to get better compression with `zopflipng -m --lossy_transparent --iterations=1000`.

XhmikosR:
Thanks for bring this into my attention.

I created a branch with better results.

PR #20260, opened 2016-07-08
v4 hamburger icon fix

patrickhlauke:
closes https://github.com/twbs/bootstrap/issues/20234

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -116,9 +116,12 @@
     text-decoration: none;
   }
 
+  // Hamburger menu icon
   &::before {
     font-weight: 500;
-    content: "\2261"; // see http://unicode.johnholtripley.co.uk/2261/
+    content: "\2261"; // IDENTICAL TO (U+2261)
+    // wider overall support (see http://unicode.johnholtripley.co.uk/2261/)
+    // compared to previously used TRIGRAM FOR HEAVEN (U+2630)
   }
 
 }
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -115,6 +115,12 @@
   @include hover-focus {
     text-decoration: none;
   }
+
+  &::before {
+    font-weight: 500;
+    content: "\2261"; // see http://unicode.johnholtripley.co.uk/2261/
+  }
+
 }
 
 // scss-lint:disable ImportantRule
--- a/docs/_includes/nav-home.html
+++ b/docs/_includes/nav-home.html
@@ -27,9 +27,7 @@
 
     <nav>
       <div class="clearfix">
-        <button class="navbar-toggler pull-xs-right hidden-sm-up" type="button" data-toggle="collapse" data-target="#bd-main-nav" aria-controls="bd-main-nav" aria-expanded="false" aria-label="Toggle navigation">
-          &#9776;
-        </button>
+        <button class="navbar-toggler pull-xs-right hidden-sm-up" type="button" data-toggle="collapse" data-target="#bd-main-nav" aria-controls="bd-main-nav" aria-expanded="false" aria-label="Toggle navigation"></button>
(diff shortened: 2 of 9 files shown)


patrickhlauke:
comparison before/after screenshot (admittedly the `\2261` glyph is a bit anemic - hence the bump in `font-weight` - and not quite as wide as I'd like...)

> hamburglar

patrickhlauke:
@mdo @cvrebert thoughts?

scss/_navbar.scss
```diff
@@ -115,6 +115,12 @@
   @include hover-focus {
     text-decoration: none;
   }
+
+  &::before {
+    font-weight: 500;
+    content: "\2261"; // see http://unicode.johnholtripley.co.uk/2261/
```
cvrebert: Maybe comment it as `IDENTICAL TO (U+2261)` similar to https://github.com/twbs/bootstrap/blob/1347cb0152c992f8b71e185fa7ef1794fc94d931/scss/_carousel.scss#L145?

scss/_navbar.scss
```diff
@@ -115,6 +115,12 @@
   @include hover-focus {
     text-decoration: none;
   }
+
+  &::before {
```
cvrebert: A comment explaining what this is for (e.g. `// Hamburger menu icon`) would be good.

cvrebert:
I'd also suggest adding a comment explaining why we can't use TRIGRAM FOR HEAVEN (U+2630).
Documenting these sorts of design decisions / historical context is best-practice.

scss/_navbar.scss
```diff
@@ -116,9 +116,15 @@
...
   }
 
+  // Hamburger menu icon
   &::before {
     font-weight: 500;
-    content: "\2261"; // see http://unicode.johnholtripley.co.uk/2261/
+    content: "\2261"; // IDENTICAL TO (U+2261)
+    // wider overall support (see http://unicode.johnholtripley.co.uk/2261/)
+    // compared to previously used TRIGRAM FOR HEAVEN (U+2630)
+    // using generated content, rather than adding this in markup,
```
cvrebert: Erm, where did we establish that using generated content made any difference?

patrickhlauke: Good call, i got confused

cvrebert:
LGTM technically speaking. I can't speak as to the visual design; @mdo ought to weigh in.

mdo:
Aesthetically that looks a bit off right now. I'd be curious to see an exploration that use an SVG background image like we do with our validation icons for comparison.

patrickhlauke:
rough sketch of how an SVG approach may look like https://jsfiddle.net/patrick_h_lauke/nu5ejdy3/2/

cvrebert:
A survey of implementation options: https://css-tricks.com/three-line-menu-navicon/

patrickhlauke:
tricks using borders, shadows etc just feel...extremely hacky and old-school to me. my personal preference would be either an SVG or unicode approach (I guess SVG would give us more flexibility in design of course, so not adverse to changing this to match my rough fiddle https://jsfiddle.net/patrick_h_lauke/nu5ejdy3/3/ ... @mdo thoughts?)

XhmikosR:
I agree with @mdo that visually the new one doesn't look so good.

The SVG one does look a lot better IMO.

patrickhlauke:
@mdo @cvrebert @XhmikosR should i update this PR to take the approach i tried in my jsfiddle instead then? I'm tending towards that one too myself

XhmikosR:
Personally I think yes. But wait for the others to confirm too.

cvrebert:
Sounds like we're onboard with SVG. I'd say go for it Patrick.

patrickhlauke:
closing in favor of successor https://github.com/twbs/bootstrap/pull/20329 which uses SVG

PR #20310, opened 2016-07-16
[WIP] Transition from Grunt to Gulp

wolfy1339:
As said in #17021 in the super ambitious nice to haves, I have transitioned from Grunt to Gulp.
**Some things still need to be done and not everything works.**
This is a base for the transition from Grunt to Gulp.
I have done my best to try to create all the tasks, it wan't that easy trying to find what each task does.
Also note that `grunt-saucelabs` does not have a gulp equivalent, so something will need to be thought out as an alternative.
**Another note, the hound config needs updating.**
If anyone has any improvements, please post them here.


Johann-S:
It seems it's a question
> [ ] Move to Gulp from Grunt?

but I want to know what's the reasons to move from Grunt to Gulp ?

neilhem:
@Johann-S because of it has been mentioned here https://github.com/twbs/bootstrap/pull/17021
> We also have some super ambitious nice to haves: Move to Gulp from Grunt?

cvrebert:
TBH, I'd rather move to npm scripts instead of gulp.

cvrebert:
We're going with a different route: #20332

wolfy1339:
Ok, That's fine.

On Jul 20, 2016, at 15:43, Chris Rebert <notifications@github.com<mailto:notifications@github.com>> wrote:


We're going with a different route: #20332<https://github.com/twbs/bootstrap/pull/20332>

-

PR #17222, opened 2015-08-22
Replace PNG with SVG and compress SVGs in custom form controls

thekondrashov:
#### Replace PNG with SVG

|  | Preview (scale 500%) |
| --- | --- |
| before / before (pixelated) / after | Preview Preview Preview |
#### Compress SVGs

Well compressed with SVGO, but I'm sure it is possible to compress more:
- form-icon-success
- form-icon-warning
- form-icon-error
- input:checked icon

(re)drawn by me (with maximum compression, is impossible to compress better :muscle:):
- c-select
- input:indeterminate icon
- input:checked icon

What do you think about this hack?

``` css
input:checked ~ .c-indicator {
  background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-4 -4 8 8'%3E%3Ccircle fill='%23fff' r='3'/%3E%3C/svg%3E");
}
```

Looks dirty, but still smaller. Read more: https://codepen.io/Tigt/blog/optimizing-svgs-in-data-uris

PR #20329, opened 2016-07-19
V4 hamburger fix2

patrickhlauke:
successor to https://github.com/twbs/bootstrap/pull/20260

closes https://github.com/twbs/bootstrap/issues/20234

for comparison: current v4-dev (top) using the unicode character, and the CSS SVG background image (bottom)

> hamburglar

### DIFF
--- a/docs/_includes/nav-home.html
+++ b/docs/_includes/nav-home.html
@@ -27,9 +27,7 @@
 
     <nav>
       <div class="clearfix">
-        <button class="navbar-toggler pull-xs-right hidden-sm-up" type="button" data-toggle="collapse" data-target="#bd-main-nav" aria-controls="bd-main-nav" aria-expanded="false" aria-label="Toggle navigation">
-          &#9776;
-        </button>
+        <button class="navbar-toggler pull-xs-right hidden-sm-up" type="button" data-toggle="collapse" data-target="#bd-main-nav" aria-controls="bd-main-nav" aria-expanded="false" aria-label="Toggle navigation"></button>
         <a class="navbar-brand hidden-sm-up" href="{{ site.baseurl }}/">
           Bootstrap
         </a>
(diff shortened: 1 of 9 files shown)


patrickhlauke:
for comparison: current v4-dev (top) using the unicode character, and the CSS SVG background image (bottom)

> hamburglar

patrickhlauke:
prototype fiddle to test out https://jsfiddle.net/nu5ejdy3/7/ - only issue seems to be Edge and IE11 rendering the svg a bit too...harshly. looks good in chrome/firefox though

patrickhlauke:
rendering of the SVG icon in chrome, firefox, edge and IE11 on Win10

> hamburglar2

cvrebert:
I believe the conclusion from #17222 was that SVGs shouldn't be base64-ed, since it increases their size.

XhmikosR:
Agreed with @cvrebert, but just to verify, @patrickhlauke: isn't the base64 version larger?

patrickhlauke:
fair enough, i'll admit to not knowing all the ins and outs of data URIs. optimized the source SVG and url-encoded it ... https://jsfiddle.net/nu5ejdy3/8/

patrickhlauke:
force-pushed the change to encoding

patrickhlauke:
@mdo @cvrebert any more thoughts on this?

mdo:
Haven't had a chance to review much else with trying to wrap up the Alpha 3 stuff. Will peep when I can after that gets released.

patrickhlauke:
ping @mdo

patrickhlauke:
alternatively, there's a neat set of icons https://github.com/danklammer/bytesize-icons that we could look at using (also for other things like the X close icons)

bardiharborow:
@patrickhlauke, aren't X close icons just `&times;` these days?

PR #20393, opened 2016-07-28
Update _navbar.scss to fix navbar items in wrong position when in state "collapse in", with any navbar-toggleable-*

lvmajor:
Little modification to fix the alignment problem with the nav items and the navbar-brand when viewing collapsed menu.

It's my first try at PR's so please don't be rude. If I made mistakes, I will be happy to learn from them. 

Reference issue: #20392

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -222,6 +222,7 @@
       .navbar-nav .nav-item {
         float: none;
         margin-left: 0;
+        clear: both;
       }
     }
     @include media-breakpoint-up(sm) {
@@ -235,6 +236,7 @@
       .navbar-nav .nav-item {
         float: none;
         margin-left: 0;
+        clear: both;
       }
     }
     @include media-breakpoint-up(md) {
@@ -248,6 +250,7 @@
       .navbar-nav .nav-item {
         float: none;
         margin-left: 0;
+        clear: both;
       }
     }
     @include media-breakpoint-up(lg) {


scss/_navbar.scss
```diff
@@ -248,6 +250,7 @@
       .navbar-nav .nav-item {
         float: none;
         margin-left: 0;
+        clear:both;
```
os1r1s110: Fixed typo.

wolfy1339:
Thanks, for wanting to contribute to Bootstrap!
There are a few issues with your Pull Request:
- The title should be changed to something that describes your pull request in one line
- You should squash your commits together and change the commit message too, by doing so:
If you do not know how to do the second one, that is fine. A project contributor can do it when they merge this pull request

os1r1s110:
@wolfy1339  First thanks for kindly pointing out what was wrong with my PR. 

Secondly, I changed the title but I don't know if it is exactly the right way to describe it, see referenced issue to know exactly what problem I am trying to fix here. English is not my natal language so I may have not described it correctly.  

For the squashing, I don't know what it is yet but I'm looking in some tutorials to get a better idea and do it myself if I can.

os1r1s110:
Ok so if I got it right, the squash is a kind of rebase to only get 1 final commit to push to dev branch is that correct? 

I'm trying to do it on my fork but it doesn't work... Could I simply backtrack my branch, remove the two commits and just pull the dev branch and create a new commit that combines both previous ones, which would give the same result as a squash if I got it right...?

wolfy1339:
You got it right.
What part is not working? It doesn't help if you don't tell me where you're having issues.
You could do that, and it would give the same results, Yes.
To do that you would simply do:
```
git reset --hard HEAD~2
modify the file
git commit
git push -f
```

os1r1s110:
Thanks for answering quickly. I am doing it via Sourcetree as I'm not really familiar with Git cli. 

In sourcetree I have an option to rebase children interactively so I tried to do that but in the gui, nothing was appearing as possible commits to rebase. I'm trying to sort it out. Else I might have to do it cli way but I think I'm trashing my branch right now... :( I did a "manual" rebase and it kinda worked but it created a whole new commit which is not what I wanted in the first place...

os1r1s110:
Oh I think it worked... but now how do I change the PR to point to that new commit, the two previous ones don't even exist anymore if I made it right and understood what squashing was for... :)

wolfy1339:
Everything looks good from here

os1r1s110:
There is still lots of bugs though especially when working with flex navbar... but that will be for another day/PR :)

mdo:
This was tackled a while back in d506bd8e50b23118e3c380511f94e29881867721.

PR #20394, opened 2016-07-28
Fix package.json

dangreen:
For example, `resolve` npm module can't resolve path to bootstrap.


vsn4ik:
https://github.com/twbs/bootstrap/blob/v4-dev/package.json#L92

niftylettuce:
thanks @dangreen ❤️

PR #20397, opened 2016-07-28
Updates .list-group-flush to not create a double border when placed inside a .card

### LINKED ISSUE #20395: .list-group-flush within .card causes double border
Saw this locally when I upgraded to alpha 3, confirmed in the [public documentation](http://v4-alpha.getbootstrap.com/components/card/#content-types):

image

Seeing this in Chrome 51, Firefox 46, and Safari 9.1.2.

judetucker:
Resolves #20395

### DIFF
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -70,6 +70,7 @@
 
 .list-group-flush {
   .list-group-item {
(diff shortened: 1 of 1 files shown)


mdo:
Fixed without the extra setting of the top and bottom borders in #20873.

PR #20413, opened 2016-07-30
Remove unused $size parameter

allanchau:
$size is not currently being used in the make-col-ready() mixin


iamravenous:
`$columns` either, also with `$gutter` in `make-col()` mixin.
They aren't needed.

PR #20419, opened 2016-07-30
Updated example screenshots

esemeniuc:
Updated screenshots to reflect v4 alpha 3. Screenshots are also smaller and higher quality PNGs (1.3MB vs 0.7MB now)


mdo:
Thanks, but I'm partial to showing our screenshots in San Francisco over Segoe on Windows :grin:. I'll get these updated when they're all finalized. <3

Issue #20388, opened 2016-07-28
Fork Normalize


mdo:
Reading through #20384, and looking at the history of Normalize, I think it might be time to explore forking the project and combining it with our own Reboot. There are a few styles in Reboot that override Normalize anyway, so at some point I'd like to take a deeper look at this.

tomlutzenberger:
So this means that normalize.css will no longer just be included as kind of a "component" but actually merged with the current code base?

kkirsche:
Would you want to keep a proper fork of it in a separate repo or would you prefer to incorporate it inline within this repo? My thought would be a full fork in a separate repo which you then incorporate in, this would allow you to pull in any upstream changes easier (as I see it) while maintaining the overrides.

mdo:
Going to punt on this for v4. Not toooooo stoked for keeping up with this just yet seeing how much Normalize is already doing for us.

PR #20428, opened 2016-08-01
Upgrade to normalize.css v4.2.0.

### LINKED ISSUE #20384: <select multiple> overflows
Operating system and version: Ubuntu 14.04
Browser and version: Chrome 52.0.2743.82

Options overflow on multiple selects if there are at least 5 options.
http://codepen.io/anon/pen/GqGdbw
https://s3.amazonaws.com/upload.screenshot.co/33f65bef97

If am not mistaken it is caused by the overflow: visible; declaration set by normalize.css on selects. Possible solution is to upgrade it to the latest version, the declaration was removed in this commit: https://github.com/necolas/normalize.css/commit/e4a7827199b61e1cdcabe807044b3a2384bba924.

bardiharborow:
Fixes #20384. We probably need to merge this fix in even if we do end up going with #20388.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -3,7 +3,6 @@ fail_on_violations: true
 scss:
   config_file: scss/.scss-lint.yml
   enabled: true
-  ignore_file: scss/_normalize.scss
 
 javascript:
   enabled: false
--- a/scss/.scss-lint.yml
+++ b/scss/.scss-lint.yml
@@ -3,6 +3,9 @@ scss_files:
   - "**/*.scss"
   - "docs/assets/scss/**/*.scss"
 
+exclude:
+  - "scss/_normalize.scss"
+
 plugin_directories: ['.scss-linters']
 
 # List of gem names to load custom linters from (make sure they are already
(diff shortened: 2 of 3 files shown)


bardiharborow:
Simply copy-pasting the library caused Hound to explode, so I went and fixed the comment style. It's still complaining about vendor prefixes and property ordering, but most of those issues were present last time around anyway and it may not even be possible to fix them.

wolfy1339:
This proves there is a problem in the configuration file with Hound (`exclude_file` option in `.hound.yml ` doesn't exist according to Hound docs) and SCSSLint as `_normalize.scss` isn't ignored for some rules

bardiharborow:
Okay, so `ignore_file` is only for `jshint`, not `scss`. For `scss` you need an `exclude` in `scss/.scss-lint.yml`. I've pushed a new commit which fixes this.

wolfy1339:
Keep it as-is, we don't want comments in our outputted CSS (except for License/Copyright headers)

vanillajonathan:
Might also need to update Normalize to 5.0 to resolve issue #21060

PR #20434, opened 2016-08-02
Fix dropdown example in docs (Fix #20410)

### LINKED ISSUE #20410: Dropdown examples in docs are broken

pvdlg:
Fix #20410


judetucker:
This is a dupe of #20411

vanduynslagerp:
Indeed, I should have search more.
Although I also removed the part of the doc that recommend to use data-target for links as it seems it's irrelevant in Bootstrap v4 (The JS find the parent `.dropdown` element without requiring a data-target or an href).

PR #20441, opened 2016-08-02
Visibility mixin

cortopy:
The [docs](http://v4-alpha.getbootstrap.com/components/utilities/#invisible-content) mention an invisible @mixin which is not present in the source scss files.

This adds a visibility mixins file, and references it in the utilities file.

### DIFF
--- a/_visibility.scss
+++ b/_visibility.scss
@@ -1,0 +1,5 @@
+// Visibility
+
+@mixin invisible {
+  visibility: hidden !important;
+}
--- a/scss/_mixins.scss
+++ b/scss/_mixins.scss
@@ -28,6 +28,7 @@
 @import "mixins/text-emphasis";
 @import "mixins/text-hide";
 @import "mixins/text-truncate";
+@import "mixins/visibility";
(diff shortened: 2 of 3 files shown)


_visibility.scss
```diff
@@ -0,0 +1,5 @@
+// Visibility
+
+@mixin invisible {
+  visibility: hidden !important;
```
cortopy: This !important was in the original file

cortopy:
@tomlutzenberger this mixin is mentioned in the docs: http://v4-alpha.getbootstrap.com/components/utilities/#invisible-content

The way I'm using it is by creating my own class with the three mixins for an accessible invisible class.  I thought the docs made sense, as you may or may not want to hide for screen readers too. Without the @invisible mixin my class doesn't work on absolute positioned elements

```sass
.hidden
  @include invisible
  @include sr-only
  @include sr-only-focusable
```

This may not be a good idea for some reason but then the docs need amending

mdo:
Thanks!

PR #20288, opened 2016-07-12
Update doc assets

XhmikosR:
/CC @cvrebert

PR #20446, opened 2016-08-03
Fix Hound warnings for minified files.

bardiharborow:
Fixes warnings in #20288. See [docs](http://eslint.org/docs/user-guide/configuring).


XhmikosR:
@cvrebert: can you confirm this works?

PR #20447, opened 2016-08-03
Make .card class inside .card-columns display: block too

bassjobsen:
See: https://github.com/twbs/bootstrap/pull/18255#issuecomment-237034763


mdo:
This is being reverted in #20665.

PR #20454, opened 2016-08-03
Use card-spacer-y for vertical padding

719media:
.card-block should use $card-spacer-y for vertical padding, just like .card-header and .card-footer, right?


mdo:
The blocks are meant to have the larger padding. I'll have to update the variables to clear up the name and meaning so that's clearer.

PR #20457, opened 2016-08-04
cache selectors

dciccale:
isn't better to cache selectors where possible?

for instance  `$this = $(this)` and `$target = $(target)` instead of calling jQuery every time to wrap the element.

what about `$(this._element)` in the `_showBackdrop` method? wouldn't be good to set `this._$element = $(this._element)` in the Modal constructor?


dciccale:
I better open an issue to discuss this before working on it (:

PR #20464, opened 2016-08-04
adds a .custom-select-lg and makes .custom-select-sm use proper vars

### LINKED ISSUE #20316: .c-select-sm and .c-select-lg missing
If you have `.c-select form-control form-control-lg` it does not use proper padding and so it is not a `lg`, it's tiny.

judetucker:
This PR does two things.
1. Adds a `.custom-select-lg` and corresponding variables to Resolve #20316 
2. Makes the pre-existing `.custom-select-sm` class use the pre-existing `$custom-select-sm-padding-y` and `$custom-select-sm-font-size` which were added to the `_variables.scss` in https://github.com/twbs/bootstrap/commit/6b318ef176f894ed1ed8f81dd892319701ff0dcf#diff-d8ee409a461718bfb6240710c8c73382R373 but never referenced anywhere.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -184,14 +184,19 @@
 }
 
 .custom-select-sm {
-  padding-top: $custom-select-padding-y;
-  padding-bottom: $custom-select-padding-y;
+  padding-top: $custom-select-sm-padding-y;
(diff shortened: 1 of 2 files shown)


mdo:
This was addressed in #24699.

PR #20472, opened 2016-08-05
Add auto to `m-` spacer utility

schoenburg:
`margin-*: auto` is an invaluable tool when using flexbox positioning,
and with Bootstrap 4 supporting flexbox, it would be good
to have it built in as a base utility to avoid the need for messy
one-off classes added by end users.

This adds auto as a spacer value to use with the spacing utility.
`.m-t-auto` is an example of the class being used. With this new class,
the explicit .m-x-auto class is no longer needed as it is generated
like the others.

Documentation updated with note adding `auto`


schoenburg:
Not sure why this is failing on Travis, all that changed in the last commit was some doc text, and it passed previously.

schoenburg:
Closing and reopening to kick off Travis again.

mdo:
`auto` is excluded from the Sass map because it's independent of the spacer scale. Folks don't really gain anything by this, but you do run the risk of losing the utility if you override the `!default` values without explicitly setting it again. Thanks though!

PR #20475, opened 2016-08-06
Updated docs with extra auto resize example

califrench:
I updated the doc with an example of a flex layout of 3 columns where the center column is specified as a `.col-xs-5` to show that other columns will resize no matter what the width of the center column.
Experienced bootstrap developers will expect to have to use an even number for the center column width when in reality any size will work.
I believe this is quite useful information and doesn't add too much to the docs.
Another option would be to make that initial example a `.col-xs-5` instead of a `.col-xs-6` to highlight this fact in the first place.


mdo:
I dig it.

PR #20501, opened 2016-08-11
Popover inside modal

### LINKED ISSUE #20093: Issues with popover inside modal
On the docs page for [modals](http://v4-alpha.getbootstrap.com/components/modal/), there are some weird issues with the modal. 

To reproduce the issues, go to that docs page and scroll to [Live Demo](http://v4-alpha.getbootstrap.com/components/modal/#live-demo) section. Click on **Launch demo modal**.

This modal has a popover inside it which triggers upon clicking. Now.
1. Upon triggering the popover, the page in background is scrolled to top. I would expect it to just show the popover and the background page to stay where it is. I agree that it goes back to the correct scroll position after the modal is closed and focuses on the modal trigger but I wish it didn't scroll the entire page when the popover was opened inside it.
2. If I close the modal with the popover open, the modal get's destroyed but not the popover and it weirdly sticks on the top left corner of the screen as seen in the following screenshot.

image

nschomberg:
This fix allows popovers inside modals to be hidden when a modal is dismissed. This resolves #20093.

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -304,7 +304,7 @@ const Tooltip = (($) => {
             this._leave(null, this)
           }
 
-          if ($(this.element).parents('.modal').length > 0){
+          if ($(this.element).parents('.modal').length > 0) {
             $(tip).appendTo($(this.element).parents('.modal').get(0))
           }
         }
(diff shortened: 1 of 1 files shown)


mdo:
Closing for #21227.

PR #20534, opened 2016-08-17
Prevents overflow visible on <select multiple>

nanocom:
I had a bug when upgrading from alpha.2 to alpha.3 with a select multiple tag : the options where displayed with forced overflow visible, which renders all options even out of the select tag.
capture du 2016-08-17 11-16-28

### DIFF
--- a/scss/_normalize.scss
+++ b/scss/_normalize.scss
@@ -274,7 +274,7 @@ optgroup {
 
 button,
 input, // 1
-select { // 2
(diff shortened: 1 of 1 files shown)


nanocom:
Issue already tracked: https://github.com/twbs/bootstrap/issues/20384

nanocom:
So my pull is useless regarding the discussion on the issue.

PR #20537, opened 2016-08-18
Remove some strong language in docs/forms.md

enzanki-ars:
There may be children/students using these examples...


patrickhlauke:
Agree, that feels unnecessary.

Issue #18490, opened 2015-12-09
Add responsive spacing utility classes


ozgureffe:
It would be nice to have responsive spacing utility classes to manage gutters easily. 

like `p-xs-a-2` where `xs` is breakpoint and `2` is a kind of multiplation factor for `$spacer` value.

It would be also nice to keep abbreviations like `md, lg` for only breakpoints to reduce the confusion.

And classes like `.m-xs-auto`, `.m-xl-auto` would be great for responsively centering block elements.

davidjb:
The current v4 spacing utilities are very useful for laying out a page, but are ineffective in a responsive layout as they're a fixed size.  For example, `.m-a-3` is useful at `lg` or `xl` breakpoints (where the margin is, by default, 48px (3*1rem)).  However, on smaller viewports, especially `xs`, the margin now uses a fair proportion of the screen (eg 30% at 320px wide or ~18% at 543px).

Introducing a breakpoint-and-up based system would make it consistent to the other utilities like text alignment and floats (eg `.text-md-center` centres for `md` and up), but would add complexity.

That said, this feels akin to the responsive typography discussion (eg #17095); perhaps at least add documentation for the best practice for making spacing utilities responsive.

mdo:
Not going to happen. It'd multiply the number of classes we have for that stuff by five, one set for each tier. These are utilities, they're not meant to serve every use case.

guylepage3:
Understandable @mdo, what do you recommend as a replacement?

woto:
May be?

```html 
<div class="m-t-1 hidden-sm-up"></div>
```

sergej2705:
@mdo: A lot of people are asking for responsive spacing utility classes. Shouldn't it be considered to add these classes? IMO there are a lot of use-cases for it.

If you worry about the file-size there could be a `$enable-responsive-spacing`-variable to switch it on if needed.

leggomuhgreggo:
What @sergej2705 said.

pi-mont:
Pretty big oversight here. The padding/margin classes should adjust based on viewport width.

mdo:
@pi-mont There's no oversight, I said no to them. I'm still no to them until we at least get the rest of v4 in shipping order, too.

pi-mont:
I understand you said no, the oversight was in thinking that only one set of padding and margin classes would suit all viewports.

tinchou:
This is one of the first search results for this feature, so I thought I'd share my solution here.

[...]

PR #20570, opened 2016-08-23
Add responsive spacing utility classes

jamby:
Had this come up with my team. Designer wanted to have some spacing for mobile be different from other devices (desktop/tablet), so I added this onto the spacing utility and seems to work well!

More than happy to make any changes that people think are necessary. (For instance, wasn't sure if since I made this into a mixin if I should move it to the mixin folder now)

### DIFF
--- a/scss/utilities/_spacing.scss
+++ b/scss/utilities/_spacing.scss
@@ -1,43 +1,55 @@
-// Width
+@mixin spacing-utility($screens-size) {
+  // Width
 
-.w-100 { width: 100% !important; }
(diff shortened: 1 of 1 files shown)


razh:
This was brought up previously and turned-down (for good reason): #18490 

In any case, this should follow the convention used for defining the other responsive utilities:

```scss
    // ...
  }
}
```

Example:
https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.3/scss/utilities/_pulls.scss

PR #20574, opened 2016-08-25
Updated wrong numbering within columns

zosdev:


mdo:
The labels are meant to convey this is _one_ of the _total_ columns, not the order of them.

zohebainapore:
ok thanks, updated the same document at another place to convey the standard meaning throughout. Look for "2 of 3 (wider)"

PR #20304, opened 2016-07-14
scrollspy: fix wrong activation of all nested links

mr-july:
if the scrollspy applied to hierarchical navigation, then too many nav-links are activated.
Example markup:

```
      <ul id="toc" class="nav nav-pills nav-stacked">
        <li>
          <a href="#v1" class="nav-link">Volume 1</a>
          <ul class="nav">
            <li>
              <a href="#ch-1-1" class="nav-link">Chapter 1.1</a>
            </li>
            <li>
              <a href="#ch-1-2" class="nav-link">Chapter 1.2</a>
            </li>
            <li>
              <a href="#ch-1-3" class="nav-link">Chapter 1.3</a>
            </li>
          </ul>
        </li>
        <li>
          <a href="#v2" class="nav-link">Volume 2</a>
          <ul class="nav">
            <li>
              <a href="#ch-2-1" class="nav-link">Chapter 2.1</a>
            </li>
            <li>
              <a href="#ch-2-2" class="nav-link">Chapter 2.2</a>
            </li>
          </ul>
        </li>
      </ul>
```

on activation of "Chapter 1.2" all siblings ("Chapter 1.1" and "Chapter 1.3") are also activated, because the selector

```
$('.nav-link[href="#ch-1-2"]').parents('li').find('.nav-links')
```

matches all ".nav-link"s under "#v1", so the correct selector should be

```
$('.nav-link[href="#ch-1-2"]').parents('li').find('> .nav-links')
                                                   ^
                                      only direct children of li
```

The illustration of this problem can be found on [JSFiddle](https://jsfiddle.net/8bd158ra/2/), and the solution (just patched version of bootstrap.min.js) is [here](https://jsfiddle.net/8bd158ra/3/).

PR #20587, opened 2016-08-26
Change Scrollspy to add active class to nav item, nav link's parent

wkurosawa:
This PR change Scrollspy to add `active` class to link's parent `nav-item` , like v3 used to do, instead of adding to `nav-link`

### DIFF
--- a/js/src/scrollspy.js
+++ b/js/src/scrollspy.js
@@ -45,6 +45,7 @@ const ScrollSpy = (($) => {
   const ClassName = {
     DROPDOWN_ITEM : 'dropdown-item',
     DROPDOWN_MENU : 'dropdown-menu',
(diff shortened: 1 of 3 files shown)


bardiharborow:
You've added an extraneous `lSpy` at the end of `js/src/scrollspy.js`. I have not had a chance to test this, but take care not to break the [non list-based nav example](https://v4-alpha.getbootstrap.com/components/navbar/#nav).

bardiharborow:
Related: #17770, #18163, #18804, #20304.

mdo:
This does the same as #18163, which isn't what we're after. See my comment there for more info: https://github.com/twbs/bootstrap/pull/18163#issuecomment-256838134. Thanks though!

PR #20602, opened 2016-08-31
Fixed incorrect instructions: "Press ⌘ to copy"

tjkohli:
Changed to "Press ⌘C to copy"

### DIFF
--- a/docs/assets/js/src/application.js
+++ b/docs/assets/js/src/application.js
@@ -73,7 +73,7 @@
     })
 
     clipboard.on('error', function (e) {
-      var fallbackMsg = /Mac/i.test(navigator.userAgent) ? 'Press \u2318 to copy' : 'Press Ctrl-C to copy'
+      var fallbackMsg = /Mac/i.test(navigator.userAgent) ? 'Press \u2318C to copy' : 'Press Ctrl-C to copy'
 
       $(e.trigger)
         .attr('title', fallbackMsg)
--- a/docs/assets/js/src/application.js
+++ b/docs/assets/js/src/application.js
@@ -73,8 +73,8 @@
     })
 
     clipboard.on('error', function (e) {
-      const modifierKey = /Mac/i.test(navigator.userAgent) ? '\u2318' : 'Ctrl-';
-      let fallbackMsg = 'Press ' + modifierKey + 'C to copy';
+      const modifierKey = /Mac/i.test(navigator.userAgent) ? '\u2318' : 'Ctrl-'
+      let fallbackMsg = 'Press ' + modifierKey + 'C to copy'
 
       $(e.trigger)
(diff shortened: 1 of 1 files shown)


wolfy1339:
According to http://www.fileformat.info/info/unicode/char/2318C/index.htm this PR would be wrong

tjkohli:
@wolfy1339 Have you tried the code in your browser? The last C is the letter "C" and not part of the character. u2318 renders the command icon (⌘)

wolfy1339:
Yes I have, but It can be mis-interpreted as another Unicode glyph. So it would be better to use string concatenation to avoid any potential bugs when using any other font than the ones in default Bootstrap

Kovah:
Why not simply using `\u2318 + C` = ⌘ + C?

tjkohli:
@Kovah I do not recommend that as Apple stylizes shortcuts without a '+' sign, and adding a '+' could possibly confuse users. There is no '+' action between the ⌘ and the C, so ⌘C makes the most sense.

tjkohli:
I updated it to concatenate the two pieces. Should this work now?

tomlutzenberger:
I'd recommend `'Press \u2318 + C to copy'`(remove those single-quotes).

tjkohli:
@tomlutzenberger I don't recommend that, at least for Mac users, who this PR applies to.

> I do not recommend that as Apple stylizes shortcuts without a '+' sign, and adding a '+' could possibly confuse users. There is no '+' action between the ⌘ and the C, so ⌘C makes the most sense.

screen shot 2016-09-08 at 10 32 54 am

tomlutzenberger:
Hm okay...
In that case I would do the following:
```javascript
      const modifierKey = /Mac/i.test(navigator.userAgent) ? '\u2318' : 'Ctrl-';
      let fallbackMsg = 'Press ' + modifierKey + 'C to copy';
```

That way you dodge unnecessary string concatenation AND prevent duplicate code/text.
Note that I also changed `var` to `let`.

tomlutzenberger:
Oh yes, I forgot. They have this "no-semicolon-on-line-endings"-Rule.
After removing those, it should be fine @tjkohli

tjkohli:
@tomlutzenberger Done. Thanks dude!

bardiharborow (approved):

tomlutzenberger:
Why do the checks still fail?
Is there a possibility to restart them?

//cc @cvrebert

bardiharborow:
@tomlutzenberger just `git commit --amend` the last commit and `git push -f` the branch again.

bardiharborow:
@tjkohli could you rebase and squash these commits, and force push the branch again. This will also trigger Travis to rebuild this PR.

mdo:
Merged this with #21236, so closing. Thanks!

Issue #17163, opened 2015-08-20
Update travis to run on container-based infrastructure


kkirsche:
Current travis build (due to the use of sudo) runs on the legacy infrastructure. It should probably be updated to speed up builds, etc.

http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

PR #20605, opened 2016-09-01
Move Travis builds to container-based infrastructure.

bardiharborow:
Testing fix for #17163. So far this is green on Node v6, but there are issues with `htmllint` needing Java 8 and Node v4 trying to install `fsevents` (a clearly Mac OS X related package.json).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,16 @@
-sudo: required
-dist: trusty
+sudo: false
 language: node_js
 git:
-  depth: 10
+  depth: 1
 node_js:
   - "4"
   - "6"
 before_install:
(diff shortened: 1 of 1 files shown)


wolfy1339:
They require sudo in order to use the trusty ubuntu version. So this is probably invalid.
And, that issue was already fixed by 4e103218874d9d6a43a34837d2727f2e1b624a94 (in v4-dev), so that issue should be closed.

bardiharborow:
@wolfy1339, 4e10321 is not in v4-dev. However if [the Trusty enviroment](https://docs.travis-ci.com/user/ci-environment/) is necessary, then we will have to wait for the containerised enviroment to be updated before moving. Closing temporarily because the changes are broken and I need to do more testing anyway.

PR #20637, opened 2016-09-06
Add card header/footer background color consistency

### LINKED ISSUE #20614: Background color in card variant header/footer
Headers and footers in card variants are practically indistinguishable from the card body since the background color is the same, where in the standard card a slightly darker background gives them their own visual semantic.

screen shot 2016-09-02 at 10 47 34

Is this intended? In my opinion headers and footers in variants should follow the default visual behavior.

kevinruscoe:
Standard card headers/footers are 4% darker than the body. This PR
mimics that % to the background variants to add consistency. This resolves #20614


dist/css/bootstrap.css
```diff
@@ -3987,7 +3987,7 @@ input[type="button"].btn-block {
 
 .card-header {
   padding: 0.75rem 1.25rem;
-  background-color: #f5f5f5;
+  background-color: whitesmoke;
```
wolfy1339: Use hex codes, don't use color names

kevdotbadger: sass compiled that to colournames, not me. Anywho that's in the compiled css, not in the scss.

wolfy1339: Well that's interesting. It doesn't do it for other colors that have names. Also, please remove the dist files as per CONTRIBUTING.md in order to reduce the risks of merge conflicts

kevdotbadger: I'll close this and give it another go. It might had been a version sass that added colournames, instead of hex.

PR #20641, opened 2016-09-06
Update _buttons.scss

### LINKED ISSUE #20609: Change button outline to work with lighter colours
The mixin **button-outline-variant** currently has hardcoded values to make the colour of the text white. When using it with lighter colours (for example yellow on a dark themed website) the white is lost to the button background on hover / focus / active.

Instead I think there are 2 solutions to this:

1) Change the mixin to accept a second parameter, _$text_. This will work in a very similar way to **button-variant**.

2) Change the mixin to do a quick **@if (lightness($color) > 50) {** check and set it to fff or 000 as necessary.

kevinruscoe:
The hover/active color is defaulted to white. If the user adds a lightly coloured button (such as yellow), then the white text gets lost. This conditionally checks the color passed to return either white, or the $body-color. Fixes #20609

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -65,14 +65,22 @@
   border-color: $color;
 
   @include hover {
-    color: #fff;
+    @if (lightness($color) > 50) {
+      color: #fff;
+    }@else{
+      color: $body-color;
+    }
     background-color: $color;
         border-color: $color;
   }
 
   &:focus,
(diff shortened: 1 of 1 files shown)


bardiharborow (changes requested):
Is there any reason you aren't putting spaces around the `@else`?

mdo:
Punting on this after merging in #20734 just now. Thanks though!

PR #20646, opened 2016-09-06
Fix docs for heading sizes

TiGR:
Sizes here were from v3. Fixed these to correspond with v4 (for real).

Should rem sizes be also added to docs?

### DIFF
--- a/docs/content/typography.md
+++ b/docs/content/typography.md
@@ -31,27 +31,27 @@ All HTML headings, `<h1>` through `<h6>`, are available.
     <tbody>
       <tr>
         <td><h1>h1. Bootstrap heading</h1></td>
-        <td class="type-info">Semibold 36px</td>
+        <td class="type-info">Semibold 40px</td>
       </tr>
       <tr>
         <td><h2>h2. Bootstrap heading</h2></td>
-        <td class="type-info">Semibold 30px</td>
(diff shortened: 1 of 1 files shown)


schoenburg:
I was looking at this discrepancy today too and I was trying to figure out which was wrong, the docs or the code base? Seems like the sizes in the code are a bit big, but if those are correct than this is definitely a great fix. 

Also agree it's a good idea to update with the rem values as well since that is how they are coded.

outdooricon:
yes, the rems are great!

TiGR:
Added rem sizes, not sure about if this is intended behavior (I mean the sizes). Anyway, its up to maintainers to decide.

schoenburg:
The rems are based off of an assumed `16px` value for most browsers, and is at this time actually explicitly set to that in the styles, so some of these values are a bit off.

`h1` and `h2` are correct.

`h3, .h3` is `1.75rem` so should be `28px`

`h4, .h4` is `1.5rem` so should be `24px`

`h5, .h5` is `1.25rem` so should be `20px`

`h6, .h6` is 1rem so should be `16px`

TiGR:
Not sure how I was able to mess this up - I remember calculating these numbers, but somehow ending up writing line-heights instead of these...

mdo:
Thanks!

PR #20648, opened 2016-09-06
hover link

kevenjesus:
mouse hover nav-link


mdo:
This seems unnecessary—anchors automatically get the proper cursor by default in browsers. Thanks though!

PR #20649, opened 2016-09-06
Fixes 20614

### LINKED ISSUE #20614: Background color in card variant header/footer

kevinruscoe:
Standard card headers/footers are 4% darker than the body. This PR
mimics that % to the background variants to add consistency. This
resolves #20614

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -70,7 +70,7 @@
 .card-header {
   @include clearfix;
   padding: $card-spacer-y $card-spacer-x;
-  background-color: $card-cap-bg;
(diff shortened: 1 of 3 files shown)


mdo:
Going to punt on this for now and see what the general reaction will be to the component when we release it. We can also revive this PR and implement this in subsequent versions, but for now I'm more keen on the transparent backgrounds.

PR #20651, opened 2016-09-06
Remove classes at one time

J2TEAM:
Based on [jQuery API Document](https://api.jquery.com/removeclass/), we can remove multiple classes at one time by using space-separated classes.


bardiharborow:
There is no need to file three separate PRs.

J2TeaM:
Thank you, @bardiharborow! I've merged others into this one.

PR #20652, opened 2016-09-06
Use .attr() method and cache array length

J2TEAM:


bardiharborow:
This change causes the Travis build to fail with: `dropdowns - should set aria-expanded="true" on target when dropdown menu is shown`. Clearly `.attr` is not working the same way as `setAttribute`.

J2TeaM:
@bardiharborow I think the problem is not `.attr()`, but `.focus()`. Based on Travis logs and https://github.com/twbs/bootstrap/blob/master/CONTRIBUTING.md#js

mdo:
Conflicts aside, is this good to go, @bardiharborow?

Johann-S:
Why change ``setAttribute`` to JQuery ``attr`` method ? 
``setAttribute`` is a native method and will be more effective than using ``attr``

bardiharborow:
@Johann-S, that's more an architectural question. jQuery certainly doesn't provide any extra browser support here, since [caniuse tells us that](http://caniuse.com/#search=setAttribute):

> These DOM properties are supported in effectively all browsers (since IE6+, Firefox 2+, Chrome 1+ etc)

@mdo all good as far as I can see.

Johann-S:
It seems the Bootstrap core team prefer to keep more vanilla JavaScript for exemple : https://github.com/twbs/bootstrap/pull/19259

bardiharborow:
@Johann-S if that is the policy then we'll probably want to pass on this.

tomlutzenberger:
@bardiharborow  What about the second change? Does it improve performance in any way?

Johann-S:
For me it doesn't improve performance because ``length`` is an array property which is already stored

mdo:
Passing then—thank you all for reviewing and discussing! <3

And just in case, just holler if I need to revisit this at all. Will happily reopen if need be.

PR #20664, opened 2016-09-07
Fix cards cut of when in columns

### LINKED ISSUE #20654: V4 alpha 4 card-columns cuts off cards on top and bottom
With the release of V4 alpha 4, **card-columns** class truncates the top of the middle and right cards when the cards are of varying heights. The same behavior occurs on the bottom edge, although it appears to be the third column that is shown completely, with 1 and 2 being cut off.

card_cols_cutoff

iamphill:
This changes the cards to use `display: inline-block` when in columns to stop them
wrapping when in columns and therefore stopping them getting cut off.

screen shot 2016-09-07 at 09 43 40

Fixes #20654

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -314,6 +314,7 @@
     column-gap: $card-columns-sm-up-column-gap;
 
     .card {
+      display: inline-block;
       width: 100%; // Don't let them exceed the column width
     }
   }


mdo:
Thanks! Fixed it in #20665 with a comment so we don't remove it down the line <3.

johanromero1121:
I already added display:inline-block.
but it's still not working...

image

PR #20665, opened 2016-09-07
v4: Fix card columns

### LINKED ISSUE #20654: V4 alpha 4 card-columns cuts off cards on top and bottom

mdo:
Fixes #20654 by reverting #20447. Cards must be `inline-block` in columns, otherwise they can get split across columns.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -314,6 +314,7 @@
     column-gap: $card-columns-sm-up-column-gap;
 
     .card {
+      display: inline-block; // Don't let them vertically span multiple columns
       width: 100%; // Don't let them exceed the column width
     }
   }


schoenburg:
FYI, looks like #20664 is the same fix as this, minus the helper text. One should probably be closed out.

johanromero1121:
still not working in my side

mdo:
Card columns are buggy as hell, and you've commented on nearly three year old issue. If you need rock solid layout here, we recommend our grid system, Masonry, or another manual approach. Sorry, we had much higher hopes for card columns, but the CSS support across browsers just isn't there.

PR #20677, opened 2016-09-08
Use color variables consistently

Kovah:
Replaces some raw colors with the appropriate color variables where it makes sense.
Examples:
- `#eee` will be replaced by `$gray-lighter` (#eceeef)
- `#333` will be replaced by `$gray-dark` (#373a3c)

The two colors `#fff` and `#ddd` are not replaced by a variable. In my opinion it would make sense to add a new variable for `#ddd` because it's widely used but I wanted to ask first. Also the naming of the variable needs to be discussed.
# 

Reference: #14840

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -328,11 +328,11 @@ $input-border-radius-lg:         $border-radius-lg !default;
 $input-border-radius-sm:         $border-radius-sm !default;
 
 $input-bg-focus:                 $input-bg;
-$input-border-focus:             #66afe9 !default;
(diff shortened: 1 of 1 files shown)


mdo:
Merged this manually given the conflicts. Only thing I took out was the `$code-color` change since that isn't meant to be red but a pink-ish color. Thanks!

Kovah:
Great! 🎉 

Any idea how the many `#ddd`-like colors will be handled?

mdo:
No clue honestly! 😆

Kovah:
Okay haha! I will take a look at it, maybe I find a suitable solution for this. 😃

PR #20678, opened 2016-09-08
Docs: Break out Utilities into separate section and optimize its pages

Kovah:
Utilities are now a separate section in the docs with the following pages:
- Clearfix
- Close icon
- Colors
- Display property
- Image replacement
- Invisible content
- Responsive helpers
- Screenreaders
- Sizing and positioning
- Spacing
- Typography

May not be the best consolidation of the utilities right now but it's a good starting point for further optimizations.


mdo:
Ooooo, I really like this. Let me run through this and see how I feel about the overall order. Thanks for tackling this!

mdo:
Love it. Thanks so much!

Kovah:
Thank you! Will try to improve some more parts of the docs.
Also thought about starting with the Learn & Extend or Approach mentioned in the v4 todo list. Do you already have any ides for that?

mdo:
I don't yet I'm afraid, and I don't want to lead you down something you'd spin your wheels on too much :grin:. Maybe jump on some other docs issues without PRs? I'll try to outline some stuff on those pages in issues.

PR #20683, opened 2016-09-09
Fix #20548

### LINKED ISSUE #20548: v4: mixins/_breakpoints.scss does not compile with custom $grid-breakpoints
In Bootstrap 4 when I use custom grid tiers as described in the documentation at http://v4-alpha.getbootstrap.com/layout/grid/#grid-tiers, bootstrap fails to compile to CSS.

The problem can be reproduced with the following minimal `_custom.scss`:

```scss
// _custom.scss
//
// Use custom grid tiers as described in documentation:
// http://v4-alpha.getbootstrap.com/layout/grid/#grid-tiers.
// This breaks compilation.

$grid-breakpoints: (
  sm: 480px,
  md: 768px,
  lg: 1024px
);
```

The error messages from a Rails 5 app using the `bootstrap` gem version `v4.0.0.alpha3.1` are:

```
Invalid null operation: "null lt 4".
bootstrap (4.0.0.alpha3.1) assets/stylesheets/bootstrap/mixins/_breakpoints.scss:19:in `media-breakpoint-down'
bootstrap (4.0.0.alpha3.1) assets/stylesheets/bootstrap/_navbar.scss:221
bootstrap (4.0.0.alpha3.1) assets/stylesheets/_bootstrap.scss:34
```

The error messages from a fresh checkout of the `v4-dev` git branch, running `grunt sass` after following the setup steps in http://v4-alpha.getbootstrap.com/getting-started/build-tools are:

```
Running "sass:core" (sass) task
>> Error: Undefined operation: "null lt 3".
>>         on line 19 of scss/mixins/_breakpoints.scss
>> >>   @return if($n < length($breakpoint-names), nth($breakpoint-names, $n + 1), n
>>    -------------^
Warning:  Use --force to continue.

Aborted due to warnings.
```

The following functions are called, starting at line 221 of `_navbar.scss`, before the error occurs:

- `@include media-breakpoint-down(xs)`
- `breakpoint-max(xs)`
- `breakpoint-next(xs)`:

    ```scss
    @function breakpoint-next($name, $breakpoints: $grid-breakpoints, $breakpoint-names: map-keys($breakpoints)) {
        $n: index($breakpoint-names, $name);
        @return if($n < length($breakpoint-names), nth($breakpoint-names, $n + 1), null);
    }
    ```

    Here the `null` comparison happens since `xs` is not in `$grid-breakpoints` and thus `$n` is `null`.

I tried to come up with a proposal for a possible solution but there seems no semantically correct way to handle the case of a non-existing breakpoint in `breakpoint-next` since a `null` return value already has the meaning that the largest breakpoint was reached.

Should the code calling the `media-breakpoint-*` mixins test for the existence of the breakpoint it passes in as an argument first? This seems a little bit ugly since it would add many conditionals to the calling code. I found 33 usages of the mixins:

- `_card.scss`  (4 usages)
- `_carousel.scss`  (1 usage found)
- `_forms.scss`  (1 usage found)
- `_jumbotron.scss`  (1 usage found)
- `_modal.scss`  (2 usages found)
- `_navbar.scss`  (10 usages found)
- `_breakpoints.scss`  (4 usages found)
- `_grid-framework.scss`  (1 usage found)
- `_grid.scss`  (1 usage found)
- `_flex.scss`  (4 usages found)
- `_pulls.scss`  (1 usage found)
- `_text.scss`  (1 usage found)
- `_visibility.scss`  (2 usages found)

Depending on the difficulty of this issue I could try to work on it, but might need a little bit of mentorship. Let me know what you think.

pvdlg:
Fixes #20548 
Add a null check in breakpoint-next.
As a result, the content of `@include media-breakpoint-down(XX)` or `@include media-breakpoint-up(XX)` will be generated only if the breakpoint exist in `$grid-breakpoints`.

### DIFF
--- a/scss/mixins/_breakpoints.scss
+++ b/scss/mixins/_breakpoints.scss
@@ -16,7 +16,7 @@
 //    md
 @function breakpoint-next($name, $breakpoints: $grid-breakpoints, $breakpoint-names: map-keys($breakpoints)) {
   $n: index($breakpoint-names, $name);
-  @return if($n < length($breakpoint-names), nth($breakpoint-names, $n + 1), null);
+  @return if($n, if($n < length($breakpoint-names), nth($breakpoint-names, $n + 1), null), null);
 }
 
 // Minimum breakpoint width. Null for the smallest (first) breakpoint.


mdo:
I tested this out locally with a small addition, using `xxl` as a breakpoint (since it doesn't exist). The downside is this still compiles the CSS, just without the media query. I'm unsure if I want to add that since it seems like it hides an error or bug in someone's code. Thinking I'll pass for now, but I'm happy to revisit if there's something I'm missing.

---

Example Sass change:

```scss
@include media-breakpoint-up(xl) {
  .element {
    display: block;
  }
}
```

CSS added:

```diff
➜ ~/work/bootstrap (vanduynslagerp-fix-20548) git diff dist/css/bootstrap.css
diff --git a/dist/css/bootstrap.css b/dist/css/bootstrap.css
index 792fa8f..3677c7d 100644
--- a/dist/css/bootstrap.css
+++ b/dist/css/bootstrap.css
@@ -854,6 +854,12 @@ pre code {
   }
 }
 
+.element {
+  display: block;
+}
+
```

vanduynslagerp:
Understood.

2 solutions I can think of.

1. Add an error message in case the breakpoint doesn't exists.

2. Create the following functions:

- get-smallest-breakpoint()

- get-largest-breakpoint()

This way instead of `@include media-breakpoint-up(xl)` yu would do `@include media-breakpoint-up(get-largest-breakpoint)`. This way the media query will be generated for the largest breakpoint in `$grid-breakpoints` no matter how it is named. Similar idea for the smallest breakpoint.
We could even create a `get-breakpoint(index)` function  that would allow to create a media query for the nth breakpoint.
The compilation would fail is the `$grid-breakpoints` doesn't contains the requested breakpoint.

A quick search in bootstrap code base shows that only xs (smallest one) breakpoint is used by name. The documentation scss also use sm (the second smallest one).

Happy to do a PR with either solution if you are interested. Let me know !

PR #20684, opened 2016-09-09
v4: The Utilities Update

### LINKED ISSUE #20266: Rename .img-rounded to .rounded

mdo:
Addressing a ton of things in one massive PR around our utilities. We've already broken each group of utilities out into new pages with #20068, and now I'm cleaning things up to make utilities more granular, easier to use, and more flexible to a multitude of use cases.
- **Changed syntax of spacing utilities** to help avoid collision with Microformats (see #20185). This doesn't completely resolve the problem, however, because I want to move from `.pa-*`/`.ma-*` to just `.p-*` and `.m-*` to avoid a weird and unclear abbreviation of "all" there.
- **Revamps the text and background color utilities.** Background utilities no longer set a `color`; you now apply that with a separate utility for more control. Also new is a `.text-white` for those common cases where you just want white text on dark background.
- **Renames `.img-rounded` an `.img-circle` to `.rounded` and `.rounded-circle`.** Also, add new utilities for rounding a complete side of an element (e.g., `.rounded-top` or `.rounded-right`). (Pulls in commits from #20363 to fix #20266.)
- **Removes `display: block;` from `.img-fluid`.** Responsive image behavior is not dependent on `display: block;`, so we can safely remove it on our end. Should you need block level, you can easily override that in the source or with a utility class.

**Still to do:**
- [x] Rename the `.pull-*` classes to reflect their CSS property, `.float-*`.
- [ ] ~~Update all instances of background utilities to ensure color contrast.~~ _Punting._
- [ ] ~~Consider reorganizing these new utility docs a bit more (e.g., combine display and responsive display, combine floats and clearfix, or something to that effect).~~ _Punting._
- [ ] Update all URLs that link to and within the new utility docs.
- [ ] Audit and make consistent use of `!important`.
- [ ] ~~Add vertical alignment utilities~~ _Punting to 4.1 ideas in #20866_

Am I missing anything else major?


bardiharborow:
You'll need to remove `Components/Utilities` out of `docs/_data/nav.yml` now that it is in it's own section.

razh:
How about the naming scheme for the flexbox utility classes?

For `flex-direction` values other than the `row` default (i.e., `row-reverse`, `columm`, and `column-reverse`), the `-top`, `-bottom`, `-left`, and `-right` suffixes don't reflect the rendered layout:

```
┌──────────────────────────────────────────────────────┐
│ .flex-items-xs-left with flex-direction: row-reverse │
│                     ┌───────┐  ┌───────┐  ┌───────┐  │
│                     │   3   │  │   2   │  │   1   │  │
│                     └───────┘  └───────┘  └───────┘  │
└──────────────────────────────────────────────────────┘
```

I believe this is also an issue with RTL languages.

danawoodman:
`.rounded-circle` seems a bit redundant does it?

scss/utilities/_spacing.scss
```diff
@@ -14,18 +14,18 @@
     $length-x:   map-get($lengths, x);
     $length-y:   map-get($lengths, y);
 
-    .#{$abbrev}-a-#{$size} { #{$prop}:        $length-y $length-x !important; } // a = All sides
-    .#{$abbrev}-t-#{$size} { #{$prop}-top:    $length-y !important; }
-    .#{$abbrev}-r-#{$size} { #{$prop}-right:  $length-x !important; }
-    .#{$abbrev}-b-#{$size} { #{$prop}-bottom: $length-y !important; }
-    .#{$abbrev}-l-#{$size} { #{$prop}-left:   $length-x !important; }
+    .#{$abbrev}-#{$size} { #{$prop}:        $length-y $length-x !important; } // a = All sides
```
raspberryman: Was this supposed to have the -a-?

mdo: Nope. Think of it like the CSS properties—`margin-left` goes with `m-l` while `margin` goes with `m`.

PR #20690, opened 2016-09-09
Add .tag-outline-* classes

mhedrick:
This is a very simple PR - just emulating the button outline classes with labels.  My team is using the `.btn-outline-*` for buttons as well as just simply displaying data which is semantically strange to me.  It will allow for style matching when sites are using the buttons with outlines.

screen shot 2016-09-09 at 3 09 57 pm

### DIFF
--- a/docs/components/tag.md
+++ b/docs/components/tag.md
@@ -32,6 +32,19 @@ Add any of the below mentioned modifier classes to change the appearance of a ta
 <span class="tag tag-danger">Danger</span>
 {% endexample %}
 
+## Outline tags
+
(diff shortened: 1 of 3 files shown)


mdo:
Thanks, but we'll pass for now <3.

TheSharpieOne:
You passed for now and now has passed 😉 , with v4 coming closer and a milestone for v4.1 **ideas**, I ask you to please reconsider, at least for [v4.1 ideas](https://github.com/twbs/bootstrap/milestone/37)
Obviously this PR needs some work (`tag-outline-*` -> `badge-*-outline`), but at least consider the possibility of these very nice looking components. If you need an updated/valid PR before adding it to the list of ideas, I can make that happen.

PR #20715, opened 2016-09-13
replace some profanities

VarelloThemes:
bearing in mind the budding younger programmers


bardiharborow:
I have mixed feelings about the incessant political correctness of this.

VarelloThemes:
me too, it sounds mad right? but I have actually written course material, using bootstrap files, which is consumed by learners as young as 10 years of age!

It's unlikely, but I think it is maybe a standard we should maintain for just this weird and extreme IRL edge case. Just a thought :baby:

PR #20725, opened 2016-09-15
Add container to tooltip options

### LINKED ISSUE #17755: Tooltip/Popover container option is not implemented
V4 documentation shows the container option exists but it is not actually implemented in the code.
http://v4-alpha.getbootstrap.com/components/tooltips/#options
https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L28

Also found 2 unit tests which attempt to use the option even though it is not implemented.
https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/tooltip.js#L331
https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/tooltip.js#L661

andyexeter:
Fixes #17755 by adding the already documented `container` option for tooltips and popovers.

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -47,7 +47,8 @@ const Tooltip = (($) => {
     selector    : false,
     placement   : 'top',
     offset      : '0 0',
-    constraints : []
+    constraints : [],
+    container   : false
   }
 
   const DefaultType = {
@@ -60,7 +61,8 @@ const Tooltip = (($) => {
     selector    : '(string|boolean)',
     placement   : '(string|function)',
     offset      : 'string',
-    constraints : 'array'
+    constraints : 'array',
+    container   : '(string|boolean)'
   }
 
   const AttachmentMap = {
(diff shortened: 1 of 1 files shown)


andyexeter:
Looks like this is failing because an element is being passed as an option in one of the tests. The documentation states that the `container` option type should be a `string` or `boolean`. Should I update the test or update the docs/code so an element is accepted?

andyexeter:
Should be noted here as well as in #17755 that the `container` option will only work for elements which are `position: static` - otherwise Tether moves the popover/tooltip into the `body`. Should I update the docs to mention this?

mdo:
@bardiharborow @Johann-S Any chance one of y'all could review this? <3

bardiharborow:
I've made a few changes in 0ded703, including fixing the build error and switching to `const` to be compliant with the new linting rules.

(`git checkout v4-dev && git fetch https://github.com/bardiharborow/bootstrap.git container-tooltip-option:container-tooltip-option -f && git merge container-tooltip-option` should work.)

mdo:
Added in 87a0749f891e64a38b4be1806e59cffb250501ce.

PR #20731, opened 2016-09-16
jspm updates package information

peebeebee:
Doing a  `jspm install npm:bootstrap` did not give back the dist-folder. so no css is included, and no bootstrap.js (only source js files)

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "LICENSE"
   ],
   "jspm": {
-    "main": "js/bootstrap",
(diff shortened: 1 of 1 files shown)


peebeebee:
Seems like `jspm install github:bootstrap` does give back files differently. 
Is this expected behaviour?

peebeebee:
Okay, I read https://github.com/jspm/registry/issues/552, and it became clear it's not my choice how to install bootstrap. Bootstrap should be installed from github.

Don't know how you end up with different files though.

PR #20734, opened 2016-09-16
Remove hard-coded value in button-outline-variant

### LINKED ISSUE #20609: Change button outline to work with lighter colours

beausmith:
Fixes #20609

PR #20641, another solution to this issue, hard-codes use of $body-color, which may not be the desired color for the text when hovering.
### Before

before

```
+button-outline-variant(white)
```
### After

after

```
+button-outline-variant(white, $color-brand-teal)
```

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -58,21 +58,21 @@
   }
 }
 
-@mixin button-outline-variant($color) {
+@mixin button-outline-variant($color, $color-hover: #fff) {
   color: $color;
   background-image: none;
   background-color: transparent;
   border-color: $color;
 
   @include hover {
-    color: #fff;
+    color: $color-hover;
     background-color: $color;
         border-color: $color;
   }
 
   &:focus,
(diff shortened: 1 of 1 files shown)


intellix:
It seems that this isn't ever used. It always defaults to white which is the same as being hardcoded: https://github.com/twbs/bootstrap/blob/v4-dev/scss/_buttons.scss#L76

beausmith:
@intellix - I posted my answer to your question on your PR: https://github.com/twbs/bootstrap/pull/22164#issuecomment-285975613

PR #20745, opened 2016-09-18
Refs #17201: Use Tether as a non-window global variable, last try

karolyi:


karolyi:
this could be a temporary fix until you get the ES6 imports right.

karolyi:
bump

karolyi:
manually ref-ing #17201

karolyi:
bump

bardiharborow:
Please explain the Webpack workaround and why you need the try-catch block instead of `const _Tether = Tether || window.Tether`.

karolyi:
if you use `x = bla || window.bla` you'll get a `ReferenceError` in browsers when the variable doesn't exist in `window`. hence the try-catch

the webpack workaround:
https://github.com/karolyi/forum-django/blob/development/frontend/webpack/config.base.js#L64-L71

karolyi:
bump

bardiharborow:
@karolyi why would `const _Tether = typeof Tether === "undefined" ? window.Tether : Tether;` not work? And could you inline that into the function call so that we don't pollute the global namespace, i.e. `})(jQuery, typeof Tether === "undefined" ? window.Tether : Tether)`.

karolyi:
fair enough. I had no other idea to check if Tether is available in the global namespace (as in `window` or `global` in case of nodejs)

I'll change it and commit.

If it works, would you be so kind to merge it?

bardiharborow:
Closing under the assumption that #19541 fixes this. Ping me if there are any further problems.

Johann-S:
You can close issue @bardiharborow 😲 well done 👍

PR #20775, opened 2016-09-23
Make equal-width flexbox extend grid-column

vinpac:
Fixed equal-width flexbox columns, so now they will act like normal columns as they break.

### DIFF
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -37,9 +37,9 @@
       // Provide basic `.col-{bp}` classes for equal-width flexbox columns
       @if $enable-flex {
         .col-#{$breakpoint} {
-          width: auto !important;
           flex-basis: 0;
           flex-grow: 1;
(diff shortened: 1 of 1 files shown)


vinpac:
A `col-fill` would be a great feature too, but I'm looking for more real world examples before adding it.

```
.col-fill {
  flex: 1 1 auto;
}
```

mdo:
What does this fix exactly? I don't see the benefit of this without a JS Bin or demo showing the before and after.

vinpac:
These equal-width flexbox classes aren't breaking as normal columns do. When they were supposed to break, they just don't break, making the breakpoints useless on them. It happens because they were not extending %grid-column.

Here's an example: https://jsbin.com/baqoriqagu/edit?html,css,output

Just change the `$equal-width-flex-extend-grid-column` to see the difference.

mdo:
Gotcha. Okay, I took a look and cleaned some things up in #20829. Thanks!

PR #20779, opened 2016-09-24
Fix multi select box overflow

indymx:
This fixes multiple select box overflow issue. 

Also found an extra text-decoration: underline in abbr[title]

### DIFF
--- a/scss/_custom.scss
+++ b/scss/_custom.scss
@@ -2,10 +2,3 @@
 //
 // Copy variables from `_variables.scss` to this file to override default values
 // without modifying source files.
-@mixin card-outline-variant
-{
-  border-style:$card-border-style;
-  border-width:$card-border-width;
-}
-
-$card-border-style: solid !default;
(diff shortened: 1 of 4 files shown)


mdo:
Is there a corresponding change in Normalize.css for this? We pull in this entire CSS file from them, so changing it is kind of a no-no as it's a dependency :).

indymx:
There is not yet, but I can do that. I was not aware that you were pulling from them.

indymx:
I just checked their source, there is not "select"  in the offending section. 

from their source:
`button,
input { /* 1 */
  overflow: visible;
}`

from _normalize.scss:

`button,
input, // 1
select { // 2
  overflow: visible;
}`

bardiharborow:
This is already fixed in #20428 over two months ago. Just waiting for merge on that PR.

indymx:
Any time frame for that merge?

mdo:
Fixed in #20428.

PR #20795, opened 2016-09-27
[Fix #19849] Tabs are opened even if disabled.

### LINKED ISSUE #19849: Click on disabled tab still shows corresponding pane

coorasse:
Fixes #19849 and adds tests for disabled case and active case (previously not tested).


coorasse:
If I fix hound issues the build on Travis fails because of ES6 arrow functions.

coorasse:
Is this PR still needed?

bardiharborow (approved):

Johann-S (approved):

PR #20816, opened 2016-10-01
Strip $ from copied shell commands.

### LINKED ISSUE #20800: "$" shell prompt should be excluded when using the docs' clipboard Copy buttons
Copying from the Docs shouldn't copy `$` for `npm install bootstrap@4.0.0-alpha.4` installation instructions.

bardiharborow:
Fixes #20800.

### DIFF
--- a/docs/assets/scss/_syntax.scss
+++ b/docs/assets/scss/_syntax.scss
@@ -61,3 +61,6 @@
 .css .o,
 .css .o + .nt,
 .css .nt + .nt { color: #999; }
+
+.language-bash::before { color: #009; content: "$ "; user-select: none; }
+.language-powershell::before { color: #009; content: "PM> "; user-select: none; }
--- a/docs/getting-started/contents.md
+++ b/docs/getting-started/contents.md
@@ -13,7 +13,7 @@ Once downloaded, unzip the compressed folder and you'll see something like this:
 
 <!-- NOTE: This info is intentionally duplicated in the README. Copy any changes made here over to the README too. -->
 
-{% highlight bash %}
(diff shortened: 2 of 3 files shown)


mdo:
What if we instead used CSS to render the `$` at the beginning of the line with `::before`?

patrickhlauke:
Alternatively, use `user-select:none` on the `<span>` around the $ ?

patrickhlauke:
i.e. add this to _syntax.scss (assuming that class isn't used for other stuff)
```
.gp {
    color: #009;
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

bardiharborow:
@patrickhlauke, this is about the copy button, rather than manually selecting the text, however that suggestion will work if we go the `::before` path.

mdo, that would be a better solution, I was just taking the easy way out. I suspect I'll end up adding a `shell` flag of some kind to `highlight` that looks something like `.highlight-shell::before { content: '$', user-select: none; }`. I'll probably get around to this tomorrow.

mdo:
> I suspect I'll end up adding a shell flag of some kind 

I think there's a class in there for the language that you can already utilize?

bardiharborow:
@mdo, oh, yeah, duh. I've pushed a commit. However this causes problems on multi-line blocks like the [NuGet section](https://v4-alpha.getbootstrap.com/getting-started/download/#nuget), because I can't `::before` each line. Not sure where to go from here.

bardiharborow:
@mdo, I've just split the NuGet section into two parts. It's probably not the ideal solution though.

PR #20819, opened 2016-10-01
Fixed progress bar color variables

mvdriel:
I think there are two issues with the current implementation of the progress bar:
- the global `$brand-primary` color doesn't effect the color of the default progress bar (it's color is still the default blue)
- the naming of the color variables for the progress bar is confusing, because their names end on '-bg' (for example `$progress-bar-success-bg`) although they change the (foreground) color of the progress bar

### DIFF
--- a/scss/_progress.scss
+++ b/scss/_progress.scss
@@ -132,14 +132,14 @@ base::-moz-progress-bar, // Absurd-but-syntactically-valid selector to make thes
 //
 
 .progress-success {
-  @include progress-variant($progress-bar-success-bg);
+  @include progress-variant($progress-bar-success-color);
 }
 .progress-info {
-  @include progress-variant($progress-bar-info-bg);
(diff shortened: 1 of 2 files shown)


wolfy1339:
This is invalid, the variables you're referring to don't exist. Unless you meant the color of the placeholder text ontop of the progress bar

mvdriel:
@wolfy1339 Sorry, description was missing. I renamed some variables to make their name more meaningful.

wolfy1339 (changes requested):
Please change those variables in _variables.scss

mvdriel:
@wolfy1339 I did rename those variables in _variables.scss and in _progress.scss.

wolfy1339 (approved):

mdo:
Not too into the variable renaming here, but I'll take a look at the other proposed change here of tying this to the brand primary color.

PR #20821, opened 2016-10-02
Update print styles borrowed from h5bp

### LINKED ISSUE #20814: *::first-line selector in print styles crashes IE11 print / print preview
There seems to be a bug in printing with Internet Explorer.  It's not a bug in my code, because the website of Bootstrap v4 also doesn't work. My system run with W10 + IE11. Does anyone have the same problem or a solution?

patrickhlauke:
Fixes #20814 - print/print preview crash in IE11 (caused by `*:first-line`
selector)

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -14,7 +14,10 @@
     *::before,
     *::after,
     *::first-letter,
-    *::first-line {
+    p::first-line,
+    div::first-line,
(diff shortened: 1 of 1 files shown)


patrickhlauke:
@twbs/team note there seem to be conflicting requirements in the checks...hound wants "::" for pseudo selectors, while travis wants single ":"

PR #20702, opened 2016-09-12
v4: Try out Jekyll seo plugin 

mdo:
Fixes #18437.

PR #20825, opened 2016-10-02
v4: Social meta tags

### LINKED ISSUE #18437: Add social related tags in docs
First, we should try https://github.com/benbalter/jekyll-seo-tag, see how good it works for our case, otherwise I'll do it manually.

Would be good to have it for v3 too, but I don't think I'll bother with it personally.

mdo:
This supersedes #20702, but still fixes #18437.

I found the Jekyll SEO plugin stuff wasn't working right, and having a quick glance at other sites, I realized things were feeling just a tad complicated. Rather than rely on another plugin, this handles it directly in a partial.

I've added `description` frontmatter for all the pages and included that where appropriate. Also in this PR is two quick images for the Twitter cards (square for most pages, big one for home page).


XhmikosR:
Why didn't you use jekyll-social-tag?

mdo:
Couldn't get the images to work, didn't feel I needed all those meta tags, and wanted more fine grained control over the type of cards shown.

XhmikosR:
Well, the plugin aims to make things simpler and more correct. Why reinvent the wheel if we can use something that works already? That being said, I've had issues with the plugin too. But, everyone will benefit if you could report your issues upstream, so that eventually we can use it.

bardiharborow:
@XhmikosR, I looked at the template the plugin uses and it's simply not flexible enough. Even getting `<title>`s to work correctly took a bit of a hack when I tried it.

Given [this section](https://github.com/jekyll/jekyll-seo-tag#what-it-doesnt-do) in the readme, I'm not sure if there's much we can do about it:
> Jekyll SEO tag isn't designed to accommodate every possible use case. It should work for most site out of the box and without a laundry list of configuration options that serve only to confuse most users.

PR #20851, opened 2016-10-06
Adds size="n" to <select multiple>

dherges:
Keeps height in chrome

screen shot 2016-10-06 at 19 33 14

screen shot 2016-10-06 at 19 33 36

### DIFF
--- a/docs/components/forms.md
+++ b/docs/components/forms.md
@@ -41,7 +41,7 @@ Remember, since Bootstrap utilizes the HTML5 doctype, **all inputs must have a `
   </div>
   <div class="form-group">
     <label for="exampleSelect2">Example multiple select</label>
-    <select multiple class="form-control" id="exampleSelect2">
+    <select multiple size="5" class="form-control" id="exampleSelect2">
       <option>1</option>
       <option>2</option>
       <option>3</option>


cvrebert:
I would be more inclined to call that either a bug in our CSS or a bug in Chrome.

dherges:
It's a bit unclear. From a quick googling / stack overflow, in CSS the only way to seems to be set a fixed height.

``<select>`` spec says:
> [..] represents the number of rows in the list that should be visible at one time. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#Attributes

cvrebert:
I mean, the focus outline overflowing the bounds of the select like in your screenshot absolutely should not happen.

dherges:
Sure, if I start to add more ``<option>`` elements without the ``size`` attr, it begins to look like that:

screen shot 2016-10-07 at 10 08 11

Figured out that an ``overflow-y: auto`` on the ``<select>`` fixes that broken display:

screen shot 2016-10-07 at 10 11 38

dherges:
This should've been fixed in #20284, see discussion in #20779

cvrebert:
Closing as duplicate.

PR #20863, opened 2016-10-08
Fix typo in navbar docs

erikgall:
## Currently

`.navbar-nav` for a _full-height_ lightweight navigation (including support for dropdowns)
## Should be

`.navbar-nav` for a _full-width_ lightweight navigation (including support for dropdowns)


mdo:
It's correct as-is. This refers to the styling of the navbar elements stretching from the top of the navbar to the bottom.

PR #20867, opened 2016-10-09
Drop the $small-font-size variable for the newer $font-size-sm one

### LINKED ISSUE #20859: font-size-sm and small-font-size in v4
In v4 there are two variables that are named similar and act similar but are different– font-size-sm and small-font-size.

```scss
$small-font-size: 80% !default;
```
```scss
$font-size-sm:   .875rem !default;
```

This was causing some confusion on my team when someone removed `font-size: $font-size-sm` from an element's css in favor of a `.small` class in the html. So couple questions:

1. Is this a bug or as designed?
2. If it is as designed, is there a desire to change something to make it less confusing. 

One change that would allow .small to continue to change things relevant to it's parent but keep text the same as font-size-sm when the parent has a normal height would be to change small-font-size to .875%? Not sure, though.

Anyways, thanks again for all the work on BS4. We've been using it in prod since Jan. and really appreciate all the effort that's gone into making it great.

mdo:
Originally `$small-font-size` was added in v3 to set the `font-size` of the `<small>` element. Now that we have newer, global type variables, we can use those instead for these kind of things.

Fixes #20859

### DIFF
--- a/scss/_type.scss
+++ b/scss/_type.scss
@@ -60,7 +60,7 @@ hr {
 
 small,
 .small {
-  font-size: $small-font-size;
+  font-size: $font-size-sm;
   font-weight: normal;
 }
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -205,8 +205,6 @@ $headings-color:         inherit !default;
 $lead-font-size:   1.25rem !default;
 $lead-font-weight: 300 !default;
 
-$small-font-size: 80% !default;
-
 $text-muted: $gray-light !default;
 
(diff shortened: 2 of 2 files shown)


bradly:
@mdo So this changes the way <small> works inside an h* element. Here is the example in the docs: http://v4-alpha.getbootstrap.com/content/typography/#customizing-headings . What do you think about setting `h1, h2, h3, h4, h5, h6 > small { font-size: 80% }? I can create a pull request if you like this approach. Thanks.

mdo:
Oh snap, I forgot about that part. I'll revisit.

PR #20898, opened 2016-10-13
Fixed #20897 by changing btn-default to btn-secondary in docs for V4

### LINKED ISSUE #20897: Modal documentation for V4 is referencing btn-default instead of btn-secondary
The new class for default buttons is `btn-secondary` instead of `btn-default`.

The modal documentation page for V4 still uses `btn-default` in the "Using the grid system" example.

The fix is to just switch default for secondary.

roooodcastro:
The modal documentation page for V4 was using `btn-default` in the "Using the grid system" example.

I changed it to the new correct class of `btn-secondary`.

### DIFF
--- a/docs/components/modal.md
+++ b/docs/components/modal.md
@@ -298,7 +298,7 @@ To take advantage of the Bootstrap grid system within a modal, just nest `.conta
         </div>
       </div>
       <div class="modal-footer">
-        <button type="button" class="btn btn-default" data-dismiss="modal">Close</button>
+        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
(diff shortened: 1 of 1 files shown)


simon04:
For the reference: the renaming happened in https://github.com/twbs/bootstrap/commit/f399d21ef5e20bd736d5cf2946acb982378cd2f4.

To get the `btn-default` class back for legacy code:
```scss
@import "node_modules/bootstrap/scss/variables";
@import "node_modules/bootstrap/scss/mixins";

.btn-default {
  @include button-variant($btn-secondary-color, $btn-secondary-bg, $btn-secondary-border);
}
```

PR #20907, opened 2016-10-16
Carousel: Only prevents default for ARROW_LEFT and ARROW_RIGHT keys

matheusazzi:
Fixes 2 bugs:
1. All keydowns were being prevented. Because of that the user wasn't able to navigate in the whole page using ARROW_UP/ARROW_DOWN.
2. Even when is an input or textarea the keydowns were being prevented. Because of that the user wasn't able to type any text on these elements.

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -231,17 +231,17 @@ const Carousel = (($) => {
     }
 
     _keydown(event) {
-      event.preventDefault()
-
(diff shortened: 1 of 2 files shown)


Johann-S:
Second bug fix already done : https://github.com/twbs/bootstrap/pull/19058

bardiharborow (approved):

PR #20909, opened 2016-10-16
Proper label for input "remember me" checkbox

SamuelMarks:
So you can now e.g.: click the text to check the checkbox


bardiharborow:
I fairly sure it was intentionally like that, possibly to create a very slightly bigger click surface.

patrickhlauke:
The approach taken in the current example is a valid, proper "implicit label" https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly - and the label text is already clickable to check the checkbox, unless i'm missing something?

PR #20916, opened 2016-10-16
Update brand.md

jilvin:
Hey there.

The branding logo of bootstrap was not properly linked and was showing a 404 error. I made a simple correction by changing the src value. It will work as such, whenever assets folder is one folder up the directory from the about folder.

If other cases needs to be solved, other methods shall be used.

### DIFF
--- a/docs/about/brand.md
+++ b/docs/about/brand.md
@@ -40,7 +40,7 @@ Download the Bootstrap mark in one of three styles, each available as an SVG fil
     <img class="svg" src="../assets/brand/bootstrap-outline.svg" alt="Bootstrap" width="144" height="144">
   </div>
   <div class="bd-brand-item inverse">
-    <img class="svg" src="..}/assets/brand/bootstrap-punchout.svg" alt="Bootstrap" width="144" height="144">
+    <img class="svg" src="../assets/brand/bootstrap-punchout.svg" alt="Bootstrap" width="144" height="144">
   </div>
 </div>
 
--- a/docs/about/brand.md
+++ b/docs/about/brand.md
@@ -34,13 +34,13 @@ Download the Bootstrap mark in one of three styles, each available as an SVG fil
 
 <div class="bd-brand-logos">
   <div class="bd-brand-item">
-    <img class="svg" src="{{ site.baseurl }}/assets/brand/bootstrap-solid.svg" alt="Bootstrap" width="144" height="144">
+    <img class="svg" src="../assets/brand/bootstrap-solid.svg" alt="Bootstrap" width="144" height="144">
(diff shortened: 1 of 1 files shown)


docs/about/brand.md
```diff
@@ -34,13 +34,13 @@ Download the Bootstrap mark in one of three styles, each available as an SVG fil
...
-    <img class="svg" src="{{ site.baseurl }}/assets/brand/bootstrap-solid.svg" alt="Bootstrap" width="144" height="144">
+    <img class="svg" src="../assets/brand/bootstrap-solid.svg" alt="Bootstrap" width="144" height="144">
   </div>
   <div class="bd-brand-item inverse">
-    <img class="svg" src="{{ site.baseurl }}/assets/brand/bootstrap-outline.svg" alt="Bootstrap" width="144" height="144">
+    <img class="svg" src="../assets/brand/bootstrap-outline.svg" alt="Bootstrap" width="144" height="144">
   </div>
   <div class="bd-brand-item inverse">
-    <img class="svg" src="{{ site.baseurl }}/assets/brand/bootstrap-punchout.svg" alt="Bootstrap" width="144" height="144">
+    <img class="svg" src="..}/assets/brand/bootstrap-punchout.svg" alt="Bootstrap" width="144" height="144">
```
bardiharborow: I highly doubt you intended to write `..}`.

bardiharborow (changes requested):
`{{ site.baseurl }}` is specifically designed to sort out this sort of issue. Removing it breaks things like preview.twbapps.com. Under which circumstances exactly does this cause a problem?

docs/about/brand.md
jilvin: screenshot from 2016-10-16 21-48-57
screenshot from 2016-10-16 21-49-23

jilvin: Alternative text "Bootstrap" is being shown instead of svg being loaded. Moreover, when the alt text is clicked, it redirects to a 404 error as shown in the screenshots attached above.

jilvin:
screenshot from 2016-10-16 21-48-57
screenshot from 2016-10-16 21-49-23


Alternative text "Bootstrap" is being shown instead of svg being loaded. Moreover, when the alt text is clicked, it redirects to a 404 error as shown in the screenshots attached above.

docs/about/brand.md
jilvin: Yeah. That was a mistake.

patrickhlauke:
Of course it gives a 404 when viewed this way directly on github. The whole point is that the docs need to be processed by a preprocessor that generates the HTML version/distribution

PR #20917, opened 2016-10-17
Add CDNJS version badge in readme

PeterDaveHello:
This will add a badge that shows the lib version on CDNJS and link to its page on CDNJS!

Thanks!


mdo:
We don't mention or point folks to CDNJS anywhere else in our docs, so I don't see a reason to include this here. Feel free to make your case again and I can reconsider, but otherwise thanks— we'll pass for now <3.

PeterDaveHello:
Thanks @mdo, I didn't come up with other cases yet, or is it possible to add CDNJS in the doc?

PR #20918, opened 2016-10-17
Revert "Drop the $small-font-size variable for the newer $font-size-sm one"

mdo:
Reverts twbs/bootstrap#20867


mdo:
We actually do want to keep the difference between these two variables; one is a relative percentage size for use in all typography and the other is a specific font-size as part of a type scale.

PR #20926, opened 2016-10-18
Make spacer utils responsive

mdo:
This makes the spacer utilities responsive. Instead of `.mt-0` for example, we have `.mt-sm-0`. Bit longer of a class name, but more versatile.


mdo:
Biggest question for this—should we keep the non-responsive classes in addition to the responsive ones?

mdo:
Second biggest question is the scale—I'm thinking we need more levels to the spacers. Something like .25rem, .5rem, 1rem, 1.5rem, and 3rem.

Loupiot:
A small suggestion for levels spacers :

```
p-0 => 0
p-01 => $spacer-x * .1
p-02 => $spacer-x * .2
p-03 => $spacer-x * .3
p-04 => $spacer-x * .4
p-05 => $spacer-x * .5
p-1 => $spacer-x 
p-2 => $spacer-x * 2
p-3 => $spacer-x * 3
p-4 => $spacer-x * 4
p-5 => $spacer-x * 5
```

Stadly:
+1

mdo:
This PR now includes the same approach as #20934:

- The `xs` tier no longer has a breakpoint abbreviation in the class name. So instead of `.mt-xs-0`, it's just `.mt-0`. Super easy transition from what's there currently. This works out well because none of these classes are bound by media queries given they apply to `min-width: 0` and up.

- The rest of the default tiers—`sm`, `md`, `lg`, and `xl`—all include their abbreviations. So for a large breakpoint, you'd use `.mt-lg-0`.

PR #20938, opened 2016-10-19
Docs: Sass Variables Table of Contents

JPeer264:
The documentation of `table of contents` were not fully completed. I also added numbers to have a better overview.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -6,16 +6,41 @@
 
 // Table of Contents
 //
-// Colors
-// Options
-// Spacing
-// Body
-// Links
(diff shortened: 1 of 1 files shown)


bardiharborow:
I would suggest against numbers, as it will create a lot of churn when things are changed.

JPeer264:
That's a good point, but it also creates a bit of structure.

mdo:
I can see this being helpful. Agreed that the numbers could get whacky, but it's not likely we'll change these once v4 really is finalized. It does highlight some reordering I'd like to do though after merging :). I'll take a look and update it after <3.

Issue #20956, opened 2016-10-20
SVG data URLs with unencoded # characters break in IE11 / Firefox


fdaunis:
Up-down arrows on custom select control no visible on Firefox and IE11. On Chrome works fine.

on Bootstrap 4 alpha 4 works ok, on alpha 5 you have a error on css.
![chome](https://cloud.githubusercontent.com/assets/1379676/19553600/27a12d62-96b5-11e6-93d1-a05716cab481.jpg)
![firefox](https://cloud.githubusercontent.com/assets/1379676/19553599/279f5b68-96b5-11e6-9cc1-26fcd9e4156b.jpg)
![ie11](https://cloud.githubusercontent.com/assets/1379676/19553601/27a2f64c-96b5-11e6-98b8-05231e0b6d20.jpg)

OK alpha4:
data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 4 5'%3E%3Cpath fill='%23333' d='M2 0L0 2h4zm0 5L0 3h4z'/%3E%3C/svg%3E

bootstrap_4.0.0-alpha4.css line 3456

ERROR alpha5:
data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 4 5'%3E%3Cpath fill='#333' d='M2 0L0 2h4zm0 5L0 3h4z'/%3E%3C/svg%3E

bootstrap_4.0.0-alpha5.css line 3521

xcartmods:
Same goes for custom checkboxes/radios. SVG icons display fine in Chrome 54, but not in Firefox 49.
http://v4-alpha.getbootstrap.com/components/forms/#checkboxes-and-radios-1

Johann-S:
Me too for form validation I cannot see SVG Icon on Firefox 48 and 49 but it's fine on Chrome but you should open a new issue IMO @xcartmods

patrickhlauke:
Confirmed that there is an issue with data URLs which in the latest alpha have a clear `#` rather than an encoded `%23`, which results in IE11/Firefox crapping out on them.

xcartmods:
I use a <base href=""> tag in my site, so inline SVG's don't tend to work for me anyway.
So I use the images instead. Here they are if anyone needs them also...
[bs_svg_icons.zip](https://github.com/twbs/bootstrap/files/541713/bs_svg_icons.zip)

PR #20962, opened 2016-10-20
Fix unescaped `#` in SVG data URLs

patrickhlauke:
Uses a custom string replacement function to ensure that `#` characters are correctly URL encoded to `%23`, to avoid Firefox/IE11 choking on the SVG data URLs.

Closes https://github.com/twbs/bootstrap/issues/20956

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -397,15 +397,15 @@ $custom-control-active-indicator-bg:         #84c6ff !default;
 $custom-control-active-indicator-box-shadow: none !default;
 
 $custom-checkbox-radius: $border-radius !default;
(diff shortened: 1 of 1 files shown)


patrickhlauke:
@mdo / @cvrebert would appreciate thoughts/review (because of the custom Sass function I had to add)

jonathanhefner:
Would an easier approach be a SASS function that only escapes hex colors?  Such a function would only check the first string character, and would only be applied to the color, instead of the entire URL.

patrickhlauke:
@mdo @cvrebert as we're getting various duplicate issues filed now, I'm going ahead with merging this...

PR #20963, opened 2016-10-20
Add support for column class which automatically sizes flexbox columns

alanmoo:
Creates classes like `.col-xs-auto`, `.col-sm-auto`, etc.

These allow a column to automatically be sized based on its content at certain viewports, using the same cascade of the rest of the column sizing classes. They're useful for elements that you might want stacked on smaller screens, but normally lay out nicely on a horizontal at larger screens, without the need to be fixed to a specific grid at those sizes.

### DIFF
--- a/docs/layout/flexbox-grid.md
+++ b/docs/layout/flexbox-grid.md
@@ -91,6 +91,37 @@ Auto-layout for flexbox grid columns also means you can set the width of one col
 {% endexample %}
 </div>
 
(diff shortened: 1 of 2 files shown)


alanmoo:
See http://codepen.io/alanmoo/pen/pEQXNW for demo.

alanmoo:
I've added documentation to this as well.

cc @mdo as we talked about this in Slack the other day.

mdo:
This is super cool to have. Tweaked some docs and got this merged, thanks!

PR #20966, opened 2016-10-20
Fix documentation for all-side spacing classes.

bardiharborow:
Fixes #20965. I debated the wording, but this is the cleanest I can think of for now.

PR #20989, opened 2016-10-24
fix 'all sides' utility class documentation

browner12:
if i'm not mistaken, it looks like the new syntax is to omit the `a` value.

ex: `m-1` vs `ma-1`


mdo:
Closing since #20966 was merged.

PR #20994, opened 2016-10-25
Fixing ruby gem version to 4.0.0.alpha5

ramaboo:


mdo:
Thanks, I'll need to update our shipping docs to manually adjust this for next time.

PR #21003, opened 2016-10-25
Remove custom QUnit-PhantomJS bridge.

bardiharborow:
The file is effectively the same as [the default](https://github.com/gruntjs/grunt-contrib-qunit/blob/master/phantomjs/bridge.js), except for a `.bind()` polyfill, which is [unnecessary](http://caniuse.com/#feat=es5) now that we have dropped IE8.


mdo:
Can I get a review from @cvrebert or @Johann-S?

Johann-S:
If you remove the bridge QUnit result will be : 

image

Instead of : 

image

Which is less verbose than with the bridge.

But yes it's interesting to remove the ``bind`` polyfill but I'm not sure about the Phantom Bridge which allow us to have more details on tests results.

bardiharborow:
Hmm. I missed that. Not sure whether being slightly more verbose warrants the extra file, especially given how quiet the other build commands are already.

A failing test looks like this. Seems like there is still plenty of verboseness when something goes wrong.

Johann-S (approved):

Johann-S (commented):
~~Instead of delete this file why not convert it to ES6 @bardiharborow ?~~

bardiharborow:
Punting on this for right now due to concerns by @Johann-S. I'll revisit later.

Johann-S:
Sorry I thought you transformed all our JS tests in ES6 but it's not the case. But yes I'm not sure of the interest of this PR except the removal of the ``bind`` polyfill

PR #21016, opened 2016-10-26
Add spacer mixin, refactor utility class generation

lukasbesch:
I found myself trying to extend the spacing utility classes within media queries. As this is not supported by Sass I decided to write a mixin so you can reuse defined spacers in code.


lukasbesch:
Now after creating this PR I found #20926. I could combine them two if this is desirable.

mdo:
Punting on this for the responsive variants we're adding in #20926.

PR #21019, opened 2016-10-26
Rename .in to .active to fix collapse

Starsam80:
aa11f002182877fe545933d6ac6009cd5363072c is what most likely broke this. My screen recorder broke so I can't show any gifs to show what the bug was, but anything that relied on collapse stopped working.

### DIFF
--- a/scss/_animation.scss
+++ b/scss/_animation.scss
@@ -2,26 +2,26 @@
   opacity: 0;
   transition: opacity .15s linear;
 
-  &.in {
+  &.active {
     opacity: 1;
   }
 }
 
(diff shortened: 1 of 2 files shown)


mdo:
/cc @Johann-S 

Thanks for the fix, @Starsam80! I'll take a look at this shortly :D.

Starsam80:
Here's what the bug looks like: 


It also affects dismissible alerts, collapsed navbars, and modals. There might be more, I just haven't checked.

Johann-S:
Nice catch 👍 I'll double my last PR

PR #21024, opened 2016-10-27
Docfix: Change label from Default to Secondary, and reorder dropdowns

eigan:
This pull request does two thing:
1) Change typo on dropdowns. They are secondary dropdowns, not default.
2) Change order to be same as buttons. Primary dropdown before secondary.


mdo:
Thanks!

PR #21028, opened 2016-10-27
Docs correction to change .pull to .float on Card docs

judetucker:
Fixing a reference to the old `.pull-*-*` in the Card docs located here: http://v4-alpha.getbootstrap.com/components/card/#header-nav

screen shot 2016-10-27 at 10 19 06 am


mdo:
Nice, thanks!

PR #21035, opened 2016-10-27
cleaned up the docs to complete the change from .tag to .badge

judetucker:
Cleaned up the docs a bit more to fix all of the old references to `.tag` and change them to `.badge`


mdo:
Thanks!

PR #21036, opened 2016-10-27
Rename `.active` to `.show`

### LINKED ISSUE #21021: Follow up #20982 w/ active/open nav fixes
In #20982 we renamed the `.open` to `.active` for consistent JS class names. That messes with the dropdown states though for navs and button groups. Going to have to change that class name, or find a new one to replace everything from #20982.

/cc @Johann-S @Starsam80

Starsam80:
Fixes #21021.

Also fixes tabs by changing `.in` to `.show`

### DIFF
--- a/docs/components/navs.md
+++ b/docs/components/navs.md
@@ -321,7 +321,7 @@ Use the tab JavaScript plugin—include it individually or through the compiled 
     </li>
   </ul>
   <div class="tab-content" id="myTabContent">
-    <div role="tabpanel" class="tab-pane fade in active" id="home" aria-labelledBy="home-tab">
+    <div role="tabpanel" class="tab-pane fade show active" id="home" aria-labelledBy="home-tab">
       <p>Raw denim you probably haven't heard of them jean shorts Austin. Nesciunt tofu stumptown aliqua, retro synth master cleanse. Mustache cliche tempor, williamsburg carles vegan helvetica. Reprehenderit butcher retro keffiyeh dreamcatcher synth. Cosby sweater eu banh mi, qui irure terry richardson ex squid. Aliquip placeat salvia cillum iphone. Seitan aliquip quis cardigan american apparel, butcher voluptate nisi qui.</p>
     </div>
     <div class="tab-pane fade" id="profile" role="tabpanel" aria-labelledBy="profile-tab">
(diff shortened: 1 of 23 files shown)


mdo:
I think this will run into similar issues as the `.show`/`.hide` business from jQuery? Was that a thing?

Starsam80:
Just tried it, and all it does is add `style="display: none;"`. The closest issue I could find is https://github.com/twbs/bootstrap/issues/9881 and that was dealing with the `.hidden` class, not `.show`

Starsam80:
Ah, it looks like the jQuery `.show()` & `.hide()` thing affects the `[hidden]` attribute, not any of the classes. https://v4-alpha.getbootstrap.com/content/reboot/#html5-hidden-attribute

docs/components/navs.md
```diff
@@ -302,11 +302,11 @@ $('#myTab li:eq(2) a').tab('show') // Select third tab (0-indexed)
 
 ### Fade effect
 
-To make tabs fade in, add `.fade` to each `.tab-pane`. The first tab pane must also have `.in` to make the initial content visible.
+To make tabs fade in, add `.fade` to each `.tab-pane`. The first tab pane must also have `.show` to make the initial content visible.
 
 {% highlight html %}
 <div class="tab-content">
-  <div class="tab-pane fade in active" id="home" role="tabpanel">...</div>
+  <div class="tab-pane fade show active" id="home" role="tabpanel">...</div>
```
vsn4ik: `show` and `active`?

Starsam80: Think of `.show` as it "showing" the fade animation, while `.active` is the one that is currently selected.

Mevrael:
Right now in dev brunch I see that fade animation has now `fade active` instead of `in`. However, that wouldn't allow to add animation to tabs, for example, which already has active state.

So what class name is going to be used instead of `fade in`?

Same question about `active` for tabs, navs, etc.

P.S. talking about selected (active) states, may be, it is better to use `[aria-selected]` instead? @mdo

Starsam80:
> So what class name is going to be used instead of `fade in`?

`fade show`

> Same question about `active` for tabs, navs, etc.

As far as I know, the `active` class will stay the same for tabs, navs, navbars, etc.

Starsam80:
I change 1 line and hound blows up. This really needs to be fixed.

bardiharborow:
@Starsam80 that is being worked on.

TheBuzzer67:
I think Collapse should be .in, popover .show  and collapse .active if we want a certain logic and it will correct the conflict!!

Starsam80:
@TheBuzzer67 Wait you mean the "Conflicting files"? I can fix that anytime, but when I do, hound will probably blow up again.

PR #21044, opened 2016-10-30
Adds responsive spacing

dimartinofrancesco:
The Responsive spacing is really useful.
For example: In few scenarios you want put margin/padding bottom when boxes are side by side but you don't want space (0px) when the same boxes are one below the others (like It's common in XS views).


mdo:
Thanks, but this is a duplicate of #20962 :).

PR #21046, opened 2016-10-30
Fixes #20975: navbar-text on navbar-dark and bg-inverse is invisible

### LINKED ISSUE #20975: navbar-text on navbar-dark and bg-inverse is invisible
`nav-link` change colors based on theme but `navbar-text` doesn't.

also I find `nav-link` and `navbar-text` not consistent. shouldn't be `nav-link` and `nav-text` ?

full example:
```html
<nav class="navbar navbar-dark bg-inverse navbar-full">
    <span class="navbar-text">gdsafdsa</span> <!-- invisible because color = bg-color -->
</nav>
```

abvalentine:
Made navbar-text color on navbar-dark equal to non-hovered link color.

navbar-dark

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -226,6 +226,10 @@
   .navbar-divider {
     background-color: rgba(0,0,0,.075);
   }
+  
+  .navbar-text {
+    color: $navbar-light-color;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


mdo:
Thanks!

PR #21047, opened 2016-10-30
Fix whitespace

Starsam80:
Caused by https://github.com/twbs/bootstrap/commit/e11e6ec9133e8f7eb5c06829233fd12403947e55#diff-cd8ec88dc7823a5c78f7111e37a78adbR29
Also fixes Travis CI and grunt (scss-lint): https://travis-ci.org/twbs/bootstrap/builds

### DIFF
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -26,7 +26,7 @@
   } @else {
     display: block;
   }
-  
+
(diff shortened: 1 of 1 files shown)


Starsam80:
That was quick, this got fixed already but not pushed.

PR #21065, opened 2016-11-01
Added note about async tooltip('destroy')

asher-dev:
Added a note to the `destroy` blurb similar to the ones on `hide`, `show`, and `toggle`.

The tooltip method `tooltip('destroy')` returns before removing the tooltip element from the DOM, which can lead to some unexpected behaviour (particularly when trying to change the tooltip on an element by removing it and then adding a new one).

Since that behaviour is indicated on the other methods, a reader could take its absence here to mean that the `destroy` method is not asynchronous.

### DIFF
--- a/docs/_includes/js/tooltips.html
+++ b/docs/_includes/js/tooltips.html
@@ -232,7 +232,7 @@ $('#example').tooltip(options)
   {% highlight js %}$('#element').tooltip('toggle'){% endhighlight %}
 
   <h4><code>.tooltip('destroy')</code></h4>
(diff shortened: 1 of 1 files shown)


bardiharborow:
Is this intentionally targeting `v3-dev`?

CosmicFluke:
I'm a first-time contributor and I just wanted to update the v3 docs after coming across this inconsistency. Should I change the PR to target `master`?

Johann-S:
Yes the branch to target V3 is ``master``

wolfy1339:
I believe they changed it to v3-dev and master is left as-is until a new release of v3 where they will push in the changes until v4 is released

mdo:
Closing as a won't fix given this is for v3.

PR #21085, opened 2016-11-04
v4 - Add methods for determining if elements are shown, visible, expanded

Johann-S:
I made this PR for this feature request in #17021 :

> Add methods for determining if elements are shown. visible, expanded, etc #15573

- [x] Tooltips
- [ ] Collapse
- [ ] Modal
- [ ] Documentation

They are a lots of work to do but if I can have some feedbacks from maybe @fat, @bardiharborow and @mdo It would be great 😄 because I'm not sure if I chose the best way to do this

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -123,12 +123,13 @@ const Tooltip = (($) => {
     constructor(element, config) {
 
       // private
-      this._isEnabled        = true
-      this._timeout          = 0
-      this._hoverState       = ''
-      this._activeTrigger    = {}
-      this._isTransitioning  = false
-      this._tether           = null
+      this._isEnabled         = true
+      this._timeout           = 0
+      this._hoverState        = ''
+      this._activeTrigger     = {}
+      this._tether            = null
+      this._isTransitioning   = false
+      this._isShown           = false
 
       // protected
       this.element = element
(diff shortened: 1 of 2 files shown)


mdo:
Unsure how much feedback I can give you on this one :). I'm in favor of the added methods to help folks build a bit easier though. This also probably doesn't need to be done immediately for v4 and could be punted to v4.1 to help prioritize things.

Johann-S:
Yeah sure I agree plus it's a big change. My concerne is about the way I chose to do this 
https://github.com/Johann-S/bootstrap/blob/c055840713d2a718f8c2f017287ad13f06a18974/js/src/tooltip.js#L619-L636

Johann-S:
I have another solution we limit ``isShown`` method for just one target so we can get rid of the exception 
like this : 
```javascript
      if (config === 'isShown' && this.length > 0) {
        if (this.length > 1) {
          throw new Error('isShown can be used only for one tooltip')
        }
        let data   = $(this).data(DATA_KEY)
        if (!data) {
          const _config = typeof config === 'object' ? config : null
          data = new Tooltip(this, _config)
          $(this).data(DATA_KEY, data)
        }
        return data[config]()
      }

Johann-S:
I close this PR, for now it's to early to work on that feature since we use jQuery

PR #21086, opened 2016-11-04
Feature/omit xs from grid

ETNOL:
Everything changed in the diff LGTM. Hard to judge if there was anything else that *should* have been updated and was missed. I checked for any "push-xs-*", "pull-xs-*", "offset-xs-*", and "col-xs-*" classes.  They all appear to be correctly updated to omit "xs" now.

### DIFF
--- a/docs/_layouts/docs.html
+++ b/docs/_layouts/docs.html
@@ -21,10 +21,10 @@
 
     <div class="container">
       <div class="row">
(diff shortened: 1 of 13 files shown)


wolfy1339:
Please exclude `dist` files

ETNOL:
@wolfy1339  Done

mdo:
This is huge, thanks so much! <3

PR #21093, opened 2016-11-06
Close #21090 - Fix aria-pressed attribute for buttons in container with data-attribute="buttons"

### LINKED ISSUE #21090: Buttons in container with data-toggle="buttons" don't get aria-pressed attribute
Toggle buttons need to have a relevant `aria-pressed="true"`/`aria-pressed="false"` attribute set to programmatically indicate their activation state. While this currently works fine for singles toggles http://v4-alpha.getbootstrap.com/components/buttons/#toggle-states, it doesn't for buttons that are inside a container with `data-toggle="buttons"` ... here, `active` gets correctly set/unset, but no `aria-pressed` attribute is added/updated accordingly.

```
<div class="btn-group" data-toggle="buttons">
<button class="btn btn-secondary" type="button">One</button> 
<button class="btn btn-secondary" type="button">Two</button>
</div>
```

Test case: https://jsfiddle.net/11o6oxcg/2/

Johann-S:
Fixes #21090 

I think I should /cc @patrickhlauke for this PR

### DIFF
--- a/js/tests/unit/button.js
+++ b/js/tests/unit/button.js
@@ -60,6 +60,17 @@ $(function () {
     assert.strictEqual($btn.attr('aria-pressed'), 'true', 'btn aria-pressed state is true')
   })
 
+  QUnit.test('should toggle aria-pressed on buttons with container', function (assert) {
+    assert.expect(1)
+    var groupHTML = '<div class="btn-group" data-toggle="buttons">' +
+        '<button id="btn1" class="btn btn-secondary" type="button">One</button>' +
+        '<button class="btn btn-secondary" type="button">Two</button>' +
+      '</div>'
+    $('#qunit-fixture').append(groupHTML)
+    $('#btn1').bootstrapButton('toggle')
+    assert.strictEqual($('#btn1').attr('aria-pressed'), 'true')
+  })
+
   QUnit.test('should toggle aria-pressed when btn children are clicked', function (assert) {
     assert.expect(2)
     var $btn = $('<button class="btn" data-toggle="button" aria-pressed="false">redux</button>')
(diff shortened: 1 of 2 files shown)


Johann-S:
@mdo and @bardiharborow if you have some feedbacks 👍

patrickhlauke:
sorry, been otherwise engaged. will review this shortly...

js/tests/unit/button.js
```diff
@@ -60,6 +60,19 @@ $(function () {
     assert.strictEqual($btn.attr('aria-pressed'), 'true', 'btn aria-pressed state is true')
   })
 
+  QUnit.test('should toggle aria-pressed on buttons with container', function (assert) {
```
patrickhlauke: @mdo is this just a problem of configuration? running the test locally worked just fine with no errors

patrickhlauke (approved):
Apologies for the delay in reviewing this. Tested the fix, and it works perfectly. Barring the issue with hound (which i suspect is a dud), this is good to merge

js/tests/unit/button.js
Johann-S: @bardiharborow should we add ``qunit`` in ``.eslintrc.json`` to avoid this ? 
see : http://eslint.org/docs/user-guide/configuring#specifying-environments

bardiharborow: @patrickhlauke @Johann-S this is due to Hound not supporting the cascading ESLint configs we use. Hound was recently set to ignore our test files, until a workaround can be found. Travis lints everything correctly.

bardiharborow (approved):

js/tests/unit/button.js
patrickhlauke: ok, in that case i'll go ahead and merge. thanks for clarifying, @bardiharborow

PR #21097, opened 2016-11-07
Social buttons

jakemitchellxyz:
I added some simple button variations since social media buttons are such a common thing, I figured we should include the option for them. I asked on Slack and most people agreed. I modified the documentation as well, but to see how they were implemented, you can refer to https://github.com/MeestorHok/bootstrap-social-buttons#usage

Let me know if there is anything that needs to change. Thanks!


wolfy1339:
Please exclude dist files

MeestorHok:
@wolfy1339 how do I do that, I thought I did

MeestorHok:
Ok, should be good to go!

Starsam80:
It was said back in #2422 that social buttons wouldn't be added (it's been 3 years though). Also, there is no Font Awesome on the documentation site so your examples don't work.

mdo:
Thanks, but we won't be adding these.

PR #21099, opened 2016-11-08
Fix broken and redirected links.

bardiharborow:
Reran the [W3C Link Checker](https://validator.w3.org/checklink), fixed broken and moved links. Adjusted the navigation generation code to include a trailing slash (to avoid server side redirects) in one case where it did not.


mdo:
<3

PR #21121, opened 2016-11-12
Update visual tests.

bardiharborow:
The visual tests have been reformatted to conform with the style guide (and to an extent, my own personal aesthetic), and outdated components have been updated. The tests now match the relevant example in the docs as closely as possible.


mdo:
<3

PR #21128, opened 2016-11-14
Remove extra media queries for setting gutters

yordis:
Because we use `media-breakpoint-up` for setup the gutters we could validate if the previous gutter is the same of the current one don't add any media query.

I am wondering how many rules we could remove following the same concept 🤔 Every time we use `media-breakpoint-up` based on some map configuration we could validate the previous one like `make-container-max-widths` as well


yordis:
@mdo if I remove my repo the PR goes away?

mdo:
No longer relevant with #22940.

PR #21145, opened 2016-11-16
Add lint check that all variables are !default

arthur801031:
#20717


brynmute:
True!

XhmikosR:
We moved away from scss-lint. If you come up with another solution feel free to CC me.

PR #21155, opened 2016-11-17
 Made input[type="color"] square instead of rectangular

wolfy1339:
This supersedes #20103


mdo:
Looking at this again, I'm not super happy with this approach. It feels like this is reimplementing vast majority of the `.form-control`, but missing all it's states. Plus it doesn't work in IE (11 down), doesn't work in iOS Safari, looks like a regular textual input in desktop Safari, and looks kinda jank in FF50.

Going to pass and mark this as a future potential feature for now.

PR #21161, opened 2016-11-18
Invalid visible items selector for dropdowns prevents arrow keys from working

FMCorz:
The keys up/down do not work with Bootstrap 4 dropdowns. The selectors in Javascript require nodes which are not present in BS4 markup. The attached patch seem to solve the problem I was facing.


michaeldrotar:
Submitted pull request to fix: https://github.com/twbs/bootstrap/pull/21167

Chrome and Firefox both trigger a native `click` event on spacebar `keyup` for `<button>` elements. Chrome suppresses the event if `preventDefault` is called from `keydown`, but Firefox does not.

vanduynslagerp:
I think it's a duplicate of #19864

FMCorz:
You are right. Thanks! Closing this one. Cheers!

Issue #21159, opened 2016-11-17
Can't open Dropdown via space bar in Firefox


maufarinelli:
In v4, using Firefox.  Hitting spacebar to open the dropdown menu, it opens and quickly closes

steps to reproduce:
- focused on the dropdown components by tab
- press "spacebar" one time
- the menu is display, but then hide again.

Expected behavior: menu could be shown when press spacebar
Actual result: the menu does not open correct. Actually it opens and close quickly

#20303 v3.

michaeldrotar:
Submitted pull request: https://github.com/twbs/bootstrap/pull/21167

Firefox and Chrome both fire a native `click` event on spacebar `keyup`, but Chrome suppresses it if `preventDefault` is called from spacebar `keydown`

Tikam02:
![dropdown](https://cloud.githubusercontent.com/assets/10548085/21154329/0ea1e766-c194-11e6-8e98-9cf9fc054811.png)
 here is my html code:it works
![image](https://cloud.githubusercontent.com/assets/10548085/21154377/2e64ac6e-c194-11e6-9a56-2fe295f14c7b.png)
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.css" rel="stylesheet">
  </head>
  <body>    
    <div class="container">

      <div class="row">
        <div class="col-xs-5 col-xs-offset-3">
          <div class="form-group">
            <label for="test" class="control-label">Select</label>
            <select name="test" id="test" class="form-control">
              <option value="1">one</option>
              <option value="2">two</option>
              <option value="3">three</option>
            </select>
          </div>

          <div class="dropdown">
            <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
              Dropdown
              <span class="caret"></span>
            </button>
            <ul class="dropdown-menu" aria-labelledby="dropdownMenu1">
              <li><a href="#">Action</a></li>
              <li><a href="#">Another action</a></li>
              <li><a href="#">Something else here</a></li>
              <li role="separator" class="divider"></li>
              <li><a href="#">Separated link</a></li>
            </ul>
          </div>
        </div>
      </div>

[...]

PR #21167, opened 2016-11-19
Suppress spacebar keyup for dropdowns

michaeldrotar:
Firefox and Chrome both fire click events when the spacebar is released on a `<button>`; however, Chrome suppresses this event if `preventDefault` is called from the `keydown` handler and Firefox still fires it so it must be explicitly prevented from the `keyup` as well to get them both working the same.

Alternatively, could change the dropdown to work with the `keyup` event instead of `keydown` but I took this approach as it seemed less destructive with existing code.

Fix for issue https://github.com/twbs/bootstrap/issues/21159


patrickhlauke:
Am currently on the road with no access to a properly set-up dev machine. Will review this next week once I'm hopefully back home. Sorry for the delay...

patrickhlauke:
Superseded by https://github.com/twbs/bootstrap/pull/21535

PR #21171, opened 2016-11-21
Fix incorrect select height calculation

### LINKED ISSUE #20977: <select> element height differs from other form inputs when using custom line height
When compiling Bootstrap with line-height other than the default value, the select element's height slightly differs from other elements and buttons.

select element with default line height:
default line height

select element with 1.75 line height:
1.75 line height

Thanks!

schliflo:
This fixes a bug where the height of the select element was calculated incorrectly when using a $line-height-base with a value other that 1.5 (which is the current default).

Fixes #20977.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -67,7 +67,7 @@
 select.form-control {
   &:not([size]):not([multiple]) {
     $select-border-width: ($border-width * 2);
-    height: calc(#{$input-height} - #{$select-border-width});
+    height: calc(#{$input-height} + #{$select-border-width});
   }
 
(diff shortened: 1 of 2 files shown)


bardiharborow:
The relevant issue is #20977.

mdo:
As a heads up, I'm undoing this change in https://github.com/twbs/bootstrap/pull/21407/commits/8e455ff56105e210cd313786c23532b7b81d01c9. I think we had it right before, but missed out on the `line-height` that was shared. Should've fixed that instead of just changing `calc`.

PR #20982, opened 2016-10-24
Fix #19344 - Use a single class name for opened/expanded/shown state of widgets

Johann-S:
Use only `.active` when a widget is opened/expanded/shown

TODO : 
- [x] Alert
- [x] Collapse
- [x] Dropdown
- [x] Modal
- [x] Popover
- [x] Tooltips
- [x] Documentation
- [x] Visual tests

~~PR in progress~~

Close : #19344

PR #21190, opened 2016-11-24
Change remaining .in references to .active.

bardiharborow:
Looks like we missed a few. Refs #20982.


Starsam80:
Tabs are currently broken, actually. Since we use `.active` for both showing and fading now, it breaks it. Changing a small typo in the docs now seems a bit of a waste. I've fixed this in #21036, but that hasn't been merged yet.

bardiharborow:
@Starsam80 you have a valid point. Didn't notice your PR. Closing out. <3

PR #21196, opened 2016-11-25
missing jquery dependency on npm (package.json)

renoirsousa:
Hi guys!

When you install bootstrap by npm install, the jquery does not install like a dependency of bootstrap, like bower installs. So i put the dependency on package.json to solve this issue. 

If my PR has anything wrong, please tell me.


mdo:
Ah, this is for v3. We're not making any changes to that version of the code right now. Thanks though!

PR #21199, opened 2016-11-25
v4: Miscellaneous color changes

### LINKED ISSUE #20952: $gray-light color inaccessible on white background
Hi. Currently the `$gray-light`’s contrast ratio on top of a white background doesn’t meet the WCAG2 minimum for body copy. Current value is:

``` scss
$gray-light: #818a91 !default;
```

The contrast ratio for `#818a91` on `#ffffff` is 3.51:1 (tested with http://webaim.org/resources/contrastchecker/), and WCAG 2 (AA) calls for a minimum contrast of 4.5:1 for body copy.

It seems that this color is used to signify:
- Muted text
- Disabled elements

For muted text, a darker value such as `$gray` might be better?
For disabled elements, maybe shooting for the highest accessible value would be alright? In my tests, both `#767676` and `#71777f` are just at or above 4.5:1 (on a `#ffffff` background).

Or, going for a more generic solution, maybe just lowering `$gray-light` in the `#707070` range (e.g. `#6a7279`, keeping the slightly blue hue)?

mdo:
This branch addresses a handful of color changes and contrast improvements. It doesn't solve everything from #19693, but I'm okay with that since the `$brand-` color changes aren't very good looking to me. We'll deal with that later.

- Darkens the close icon colors via `opacity` changes.
- Fixes #20952 by darkening the `$gray-light` value.
- In concert with the above fix, darkened `$gray` and `$gray-dark` to keep the color scheme intact.
- Darken the opacity of text in active list group items.

/cc @patrickhlauke

### DIFF
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -5,13 +5,13 @@
   line-height: 1;
   color: $close-color;
   text-shadow: $close-text-shadow;
-  opacity: .2;
+  opacity: .5;
 
   @include hover-focus {
     color: $close-color;
     text-decoration: none;
     cursor: pointer;
-    opacity: .5;
+    opacity: .75;
   }
 }
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -92,9 +92,9 @@
 //
 // Grayscale and brand colors for use across Bootstrap.
 
-$gray-dark:                 #373a3c !default;
-$gray:                      #55595c !default;
-$gray-light:                #818a91 !default;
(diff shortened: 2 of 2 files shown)


Francismori7:
Sadly I did prefer the brand colors in #19693, they were more vibrant to the eye on my screens.

mdo:
@Francismori7 I'm not saying we won't have changes to those colors; just not those exact values at this particular moment in time :).

Francismori7:
@mdo don't worry about it!

PR #21202, opened 2016-11-26
Update bootstrap-grid.scss to include box-sizing and responsive changes

### LINKED ISSUE #17982: Including "bootstrap-grid.scss" only leads to somewhat broken grid
I am not sure if this is by design or not. But I wanted to use "BS4 grid only" for my app. I assumed that should be possible, since I see "Grid Only" download option in alpha-docs.

So, when I import bootstrap fully like that: `@import 'bower_components/bootstrap/scss/bootstrap';` in my project I see the following grid:

When I change it to use grid only `@import 'bower_components/bootstrap/scss/bootstrap-grid';` is see this:

This is **not** browser specific.

I figured out that it's because "reboot.scss" is not included, so I added:

```
@import 'bower_components/bootstrap/scss/reboot';
@import 'bower_components/bootstrap/scss/bootstrap-grid';
```

That didn't work, since `reboot.scss` uses mixins and variables inside without importing them. So I had to manually go and figure it out and the following code fixes that:

```
@import 'bower_components/bootstrap/scss/mixins/hover';
@import 'bower_components/bootstrap/scss/mixins/tab-focus';
@import 'bower_components/bootstrap/scss/variables';
@import 'bower_components/bootstrap/scss/reboot';
@import 'bower_components/bootstrap/scss/bootstrap-grid';
```
## 

To sum up, I think it should be obvious default way to use "just a grid" or "just a grid with reboot" (if by design grid should not include "reboot"). That can be solved with explicit imports added to `_reboot.scss`; So the following code would just work:

```
@import 'bower_components/bootstrap/scss/reboot';
@import 'bower_components/bootstrap/scss/bootstrap-grid';
```

If, by design, grid _should_ include "reboot" then also explicit import of "reboot" in a `bootstrap-grid.scss`.

What do you think? Should I send a PR?

mdo:
This fixes #17982 I think—judging by the issue's screenshots, it was missing some box-sizing. Alongside that, since this is a responsive grid, I figured we'd need those responsive changes from Reboot as well.

/cc @Restuta

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -3,6 +3,25 @@
 // Includes relevant variables and mixins for the regular (non-flexbox) grid
 // system, as well as the generated predefined classes (e.g., `.col-4-sm`).
 
+//
+// Box sizing, responsive, and more
+//
+
+@at-root {
+  @-ms-viewport { width: device-width; }
+}
+
+html {
+  box-sizing: border-box;
+  -ms-overflow-style: scrollbar;
+}
+
+*,
+*::before,
+*::after {
+  box-sizing: inherit;
+}
+
 
 //
 // Variables


mdo:
Looks like it works: https://jsbin.com/bicamu/edit?output

Restuta:
@mdo thanks, would be happy to test it, is it published to bower?

mdo:
You can point any Bower package to a particular SHA or branch I think. We don't have this release published anywhere else yet.

PR #21211, opened 2016-11-26
Add .no-gutters option to remove gutters from rows

### LINKED ISSUE #19107: Add a "row without gutters" class
There are many instances when u want the grid to don't have the gutters in it .
As it just take small 8-9 lines of code then why not 

```
.row.no-gutters {
  margin-right: 0;
  margin-left: 0;
}
.row.no-gutters > [class^="col-"],
.row.no-gutters > [class*=" col-"] {
  padding-right: 0;
  padding-left: 0;
} 
```

The SCSS/Less Version of the same 

```
.row.no-gutters {
  margin-right: 0;
  margin-left: 0;

  & > [class^="col-"],
  & > [class*=" col-"] {
    padding-right: 0;
    padding-left: 0;
  }
}
```

The scss-lint version

```
.row.no-gutters {
  margin-right: 0;
  margin-left: 0;

> [class^="col-"],
> [class*=" col-"] {
    padding-right: 0;
    padding-left: 0;
  }
}
```

Will be great if this can be added in v4 ?

mdo:
This branch adds `.no-gutters` to the predefined grid classes (meaning it can be removed if you disable the grid classes). It includes docs for how it's built in the source Sass and how to use it with a quick demo. This was one of the biggest feature requests for v4, so getting it in there with relatively few lines of code added was a no brainer.

Fixes #19107.

screen shot 2016-11-26 at 11 08 51 am

### DIFF
--- a/docs/layout/grid.md
+++ b/docs/layout/grid.md
@@ -22,7 +22,7 @@ At a high level, here's how the grid system works:
 - Content should be placed within columns, and only columns may be immediate children of rows.
 - Column classes indicate the number of columns you'd like to use out of the possible 12 per row. So if you want three equal-width columns, you'd use `.col-sm-4`.
 - Column `width`s are set in percentages, so they're always fluid and sized relative to their parent element.
-- Columns have horizontal `padding` to create the gutters between individual columns.
+- Columns have horizontal `padding` to create the gutters between individual columns, however, you can remove the `margin` from rows and `padding` from columns with `.no-gutters` on the `.row`.
 - There are five grid tiers, one for each [responsive breakpoint]({{ site.baseurl }}/layout/overview/#responsive-breakpoints): extra small, small, medium, large, and extra large.
 - Grid tiers are based on minimum widths, meaning they apply to that one tier and all those above it (e.g., `.col-sm-4` applies to small, medium, large, and extra large devices).
 - You can use predefined grid classes or Sass mixins for more semantic markup.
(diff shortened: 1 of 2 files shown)


Igcorreia:
Please consider adding it to the container itself:

Please consider adding:

```
.no-gutter.row,
.no-gutter.container,
.no-gutter.container-fluid{
margin-left: 0;
margin-right: 0;
padding-left: 0;
padding-right: 0;
}

.no-gutter>[class^="col-"]{
padding-left: 0;
padding-right: 0;
}
```

mdo:
I'd advise you not to use the container if you want 100% width. The containers are there for a reason—horizontally pad the contents of the page and sometimes center it. Since you don't need it for `.no-gutters`, remove it outright instead of overriding it.

Igcorreia:
@mdo so should be .row directly? :) It didn't crossed my mind. It did the Job! Don't consider my suggestion. If this goes forward we only need in the docs how to do a 100% responsive section.

IamManchanda:
**containers** are not for 100% width ... instead fixed width as per device size ( media query ) . Just there to take a fixed and neat width ( center aligned ).
**container-fluid** also has a padding right and left of 15px each . Its there for full screen, specially used written material ( content ) .
**sections** are there for 100% width of the screen. specially used in the backgrounds

`.row.no-gutters` is enough i guess . no need for including container, container-fluid or section just nest them around as per requirement. For eg..

1. If you are looking for a center aligned fixed width, **nest the row within a container** 
2. If you are looking for a full screen but with a tiny left and right padding, **nest the row within a container**
3. If you are looking for a 100% full screen with no padding, **nest the row within a section**

Igcorreia:
@IamManchanda thanks, is clear now. Youe description should go on the DOCs in my opinion.

IamManchanda:
ohk then go ahead and make a Pull request @Igcorreia ... i will surely won't mind ... lol

gdevany:
@IamManchanda Thanks for the explanation of section/row used for 100% width.  You ended my hair pulling trying to fix the 'phone view' bottom scrolling issue by nesting row within section instead of nesting row within container (at width: 100%).

IamManchanda:
@gdevany i am happy i saved your hairs :smile:

bryankang:
@mdo Apologies, but I still don't understand the reasoning behind keeping the container's padding. 

An extremely common scenario is a navbar that takes advantage of bootstrap's `max-width` depending on the screen size. This obviously leads me to conveniently use a container. 

However, when the screen size allows the navbar edges to touch, the left logo link and the right hamburger menu cannot reach out to the very edges (without additional styles). This may work for some designs where the logo and hamburger menu don't have to be in their exact corners, but this isn't always the case (e.g., [stripe.com](https://stripe.com/)). In addition, the logo and menu links cannot take advantage of extra space on the edges because of the container's padding. 

I may be wrong here, but isn't the main reason why containers have paddings in the first place because columns can act like containers when nested? This makes for a more consistent pattern to design and code with. For the same reason, I believe that if columns don't have padding, the container shouldn't either.

PR #21218, opened 2016-11-27
Rename `.flex-xs-*` and `.flex-items-xs-*` to `.flex-*` and `.flex-items-*`

Starsam80:
Same approach as #21217, but this time I changed the formatting of `_flex.scss` a little bit so it was more easier to add `@if {...} @else {...}`.

The only remaining `xs` classes after this PR are:
- `.col-xs`
- `.col-xs-auto`
- `.navbar-toggleable-xs`
- `.hidden-xs-up`
- `.hidden-xs-down`

I plan on doing something about `.col-xs(-auto)` without affecting [this](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_grid.scss#L38) (`.col` wouldn't match) but that takes a bit more work. `.hidden-xs-(up|down)` seem like they have a naming convention going on, so changing one may break it. I'll leave that for someone else to decide.

### DIFF
--- a/docs/assets/scss/_component-examples.scss
+++ b/docs/assets/scss/_component-examples.scss
@@ -19,9 +19,9 @@
     }
   }
 
-  .flex-items-xs-top,
-  .flex-items-xs-middle,
-  .flex-items-xs-bottom {
+  .flex-items-top,
+  .flex-items-middle,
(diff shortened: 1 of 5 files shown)


mdo:
Awesome, thanks <3

PR #21220, opened 2016-11-27
Remove redundant media queries

Starsam80:
These media queries for min-width were unnecessary and produced stuff like
```css
@media (min-width: 576px) and (min-width: 576px) {
  /* Some CSS here */
}
```

Removing these queries fixes this.

P.S.: Lately, I've been noticing a lot of stuff weird/wrong with the code, so expect some PRs from me. I would like to do a giant PR with all of my fixes but that would be annoying to review, so I'm splitting it up into several different ones. With this method, it would also be easier to reject some changes by just closing the PR.

### DIFF
--- a/scss/utilities/_display.scss
+++ b/scss/utilities/_display.scss
@@ -8,15 +8,13 @@
     $min: breakpoint-min($breakpoint, $grid-breakpoints);
 
     @if $min {
-      @media (min-width: $min) {
-        .d-#{$breakpoint}-none { display: none !important; }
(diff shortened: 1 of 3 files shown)


mdo:
> P.S.: Lately, I've been noticing a lot of stuff weird/wrong with the code, so expect some PRs from me. I would like to do a giant PR with all of my fixes but that would be annoying to review, so I'm splitting it up into several different ones. With this method, it would also be easier to reject some changes by just closing the PR.

Holy crap yes pls to all of that. 😻 Super appreciative of all these PRs from you lately <3.

PR #21222, opened 2016-11-27
Rename `.col-xs` to `.col` + some other cleanup

Starsam80:
I changed lots of little stuff but the main points are
- Use `breakpoint-min` instead of a counter: This cleans up the code quite a bit and makes it more readable.
- Add `> .col,` to the `.no-gutters` class. Without this, `> [class*="col-"]` wouldn't match `.col`, and changing it to something like `> [class*="col"]` would make it match `.collapse` and more.
- Rename `.col-xs` and `.col-xs-auto` to `.col` and `.col-auto`
- Update some documentation to use new classes that weren't changed before
- Fixed a minor bug where even though `$enable-flex` was false, it would still produce parts of flex classes.

### DIFF
--- a/docs/assets/scss/_featured-sites.scss
+++ b/docs/assets/scss/_featured-sites.scss
@@ -2,7 +2,7 @@
   margin-right: -1px;
   margin-left: -1px;
 }
-.bd-featured-sites .col-xs-6 {
+.bd-featured-sites .col-6 {
   padding: 1px;
 }
 .bd-featured-sites .img-fluid {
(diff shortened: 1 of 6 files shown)


mdo:
<3<3<3

einfallstoll:
Wouldn't it make more sense to come up with a table like this:

| Name | Tiny* | Extra small | Small | Medium | Large |
| --- | --- | --- | --- | --- | --- |
| Size | <576px | ≥576px | ≥768px | ≥992px | ≥1200px |
| Prefix | `.col-` | `.col-xs-` | `.col-sm-` | `.col-md-` | `.col-lg-` |

\* I actually don't have a better name to come up with. (Does it even make sense to be that small? I mean my smartphone is *huge*)

This would be backward-compatible while still implementing this "tiny `.col-`"-class.

PR #21226, opened 2016-11-28
Rename `.navbar-toggleable-xs` to `.navbar-toggleable` + more cleanup

Starsam80:
The first commit is purely just whitespace fixes and can be dropped if needed:
- Align all the CSS properties
- Remove outdated comment

The second commit does several things:
- Remove `-xs` from `.navbar-toggleable-xs`: While my solution _works_, it isn't the prettiest. From my experimentation I couldn't find a better way without copying and pasting the whole thing.
- Remove some duplicate code
- Move the `.navbar-toggleable` chunk down to the bottom of the file, so it can override some CSS properties like `margin-left` without needing `!important`

The third commit is pretty simple:
- Rename `.navbar-toggleable-xs` to `.navbar-toggleable`
- Rename `.float-xs-(left|right)` to `.float-(left|right)`: This fixes several bugs in the docs and examples.

This PR ended up a lot bigger than I would have wanted, but I find it's really hard to separate commits so I can't undo some of the changes I did. Oh well.

### DIFF
--- a/docs/_includes/nav-home.html
+++ b/docs/_includes/nav-home.html
@@ -1,7 +1,7 @@
 <header class="navbar navbar-light navbar-static-top bd-navbar">
   <div class="container">
     {% comment %}
-    <nav class="nav navbar-nav float-xs-right">
+    <nav class="nav navbar-nav float-right">
       <div class="nav-item dropdown">
         <a class="nav-item nav-link dropdown-toggle" href="#" id="bd-versions" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
           v{{ site.current_version }}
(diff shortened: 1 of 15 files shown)


mdo:
> The first commit is purely just whitespace fixes

Awesome, thanks for jumping on this. I pushed a change to Hound earlier that silenced the errors around this and was about to tackle it tonight.

> While my solution works, it isn't the prettiest. From my experimentation I couldn't find a better way without copying and pasting the whole thing.

While not the prettiest, this is super succinct. Mad props for finding a solution.

> This PR ended up a lot bigger than I would have wanted, but I find it's really hard to separate commits so I can't undo some of the changes I did. Oh well.

Nah, this is amazing! You did a great job separating things and then writing up everything here in the PR. Made reviewing this super easy. Seriously, can't say thanks enough for all this work :).

PR #21236, opened 2016-11-28
Replacement for #20602

mdo:
This branch updates #20602 with latest from `v4-dev` so I can get tests passing and squash on merge.

### DIFF
--- a/docs/assets/js/src/application.js
+++ b/docs/assets/js/src/application.js
@@ -73,7 +73,7 @@
     })
 
     clipboard.on('error', function (e) {
-      var fallbackMsg = /Mac/i.test(navigator.userAgent) ? 'Press \u2318 to copy' : 'Press Ctrl-C to copy'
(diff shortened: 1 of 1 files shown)


docs/assets/js/src/application.js
```diff
@@ -73,7 +73,8 @@
     })
 
     clipboard.on('error', function (e) {
-      var fallbackMsg = /Mac/i.test(navigator.userAgent) ? 'Press \u2318 to copy' : 'Press Ctrl-C to copy'
+      const modifierKey = /Mac/i.test(navigator.userAgent) ? '\u2318' : 'Ctrl-'
+      let fallbackMsg = 'Press ' + modifierKey + 'C to copy'
```
wolfy1339: This should be using ES5 and not ES6. `let` and `const` should be `var`. Then it will match the rest of the file and support all supported browsers.

wolfy1339 (changes requested):

PR #21242, opened 2016-11-28
Inverse card improvements

### LINKED ISSUE #21164: Docs: specify .card-inverse usage
Please specify in the Card documentation: (starting from) http://v4-alpha.getbootstrap.com/components/card/#image-overlays
that ``.card-inverse`` only works for the following classes (as stated in ``bootstrap-flex.css``):
``.card-header``
``.card-footer``
``.card-title``
``.card-blockquote``
``.card-link``
``.card-text``
``.card-subtitle``
``.blockquote-footer``

mdo:
Fixes #21164.

Updates the docs to better articulate how the `.card-inverse` works, and also improve the class itself to apply a `color` to the entirely of the `.card-inverse` so it works with more content.

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -314,7 +314,7 @@ Turn an image into a card background and overlay your card's text. Depending on 
 
 ## Inverted text
 
-Cards include a class for quickly toggling **the text color**. By default, cards use dark text and assume a light background. **Add `.card-inverse` for white text** and specify the `background-color` and `border-color` to go with it.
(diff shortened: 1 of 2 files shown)


markzolotoy:
I am trying using card-inverse and dont see anything on it. Moving a mouse around trying to select something will reveal texts. Is that something that is not working at this point?

mdo:
Please review the latest docs and code in `v4-dev` for how to use the inverse card styles.

PR #21243, opened 2016-11-28
minor update to the .card-img class

mannyhenri:
when testing this on my own the image don't take the full width but also has some extra space (padding on the .card itself that could be removed to make the card better). Your site documented it the way it should look like but the CSS doesn't work towards that.


mdo:
These kinda things should be relegated to utilities I think. Going to pass on adding this for all card images for the time being. If it's a bigger issue after stable release, we can revisit. Thanks though!

PR #21253, opened 2016-11-30
Remove unnecessary .btn-outline and .btn-outline-inverse on docs

ysds:


patrickhlauke:
Unless you can provide some good reason for this...

tomlutzenberger:
@patrickhlauke He's right. The `.btn-outline` has no effect, it's not even defined (missing type, e.g. `primary`). See https://github.com/twbs/bootstrap/blob/v4-dev/scss/_buttons.scss#L82

The "outline" look is defined by the `docs.css` (bootstrap branding). Therefore, these classes can be removed in the documentation.

patrickhlauke:
Thanks for clarifying @tomlutzenberger . yeah, i can see the issue now. looks good to me to merge, thanks @ysds

PR #21268, opened 2016-12-01
Navbar icon bars do not appear in Windows High Contrast theme

joe-watkins:
Navbar `.icon-bar` elements do not appear in High Contrast theme in Windows. Adding a transparent border to the element makes them visible.

**Image example of bug in High Contrast theme:**
Screenshot of Navbar component with icon bars not visible in High Contrast theme

**Image example of fix in High Contrast theme:**
Screenshot of Navbar component with icon bars visible in High Contrast theme after fix

Example CSS:
``` CSS
  .icon-bar {
    ...
    border: 1px solid transparent;
  }
```


wolfy1339:
v3 is no longer maintained. If you would like you can submit this again against v4-dev.
A pointer for next time, don't run `grunt dist`, only edit the `.less` (v3) or `.scss (v4) `files

joe-watkins:
@wolfy1339 Great :) I'll work to get that into V4 if it persists there. So no need to compile or build after `.scss` edit?

patrickhlauke:
closing in favor of https://github.com/twbs/bootstrap/issues/21269

(the issue is broader in v4 due to use of SVG backgrounds. simply adding a transparent border won't fix the problem there, likely only solution will be to rethink the approach for critical cases like the hamburger menu)

PR #21270, opened 2016-12-01
Overwrite margin-bottom on form-control-static

gijsbotje:
In the documentation the `.form-control-static` class has been used on a `p` element.
On the regular vertical form the margin bottom of the `p` element gets overwritten by a `.mb-0` class.
In the inline form example this class hasn't been applied, therefore the `p` element gets a `margin-bottom`.

To prevent this behavior we can add a `margin-bottom` of `0`.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -346,7 +346,7 @@ select.form-control-lg {
     // Make static controls behave like regular ones
     .form-control-static {
       display: inline-block;
-      margin-bottom: 0px;
+      margin-bottom: 0;
     }
 
     .input-group {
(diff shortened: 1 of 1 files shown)


afuno:
This is not necessary. Use it: `mb-0`.

gijsbotje:
in bootstrap 3 it was also set to `margin-bottom: 0;`
I don't see why you wouldn't prevent this behavior

afuno:
@gijsbotje I think that in some situations this indentation is needed.
This controversial situation :)

mdo:
In this case we probably do want this on the class itself, as opposed to an override via utility. Reason being we *need* this override for the component to function correctly.

PR #21292, opened 2016-12-04
Remove `.float-left` from card header nav docs

Starsam80:
The docs say that we have to include a float utility class for proper alignment, but this must have changed recently. In my testing, removing `.float-left` not only [fixed the header nav when flex was enabled](http://i.imgur.com/w4cGPUB.png) (we were floating in a element without clearfix because of [this line](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_card.scss#L72)), but also made no difference when flex was disabled.

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -226,12 +226,12 @@ Card headers can be styled by adding `.card-header` to `<h*>` elements.
 
 ## Header nav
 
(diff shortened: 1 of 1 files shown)


mdo:
Note that right aligned navs in cards will need a float utility, and in those instances you'll still need the clearfix on the header itself.

Issue #21130, opened 2016-11-14
Change jQuery CDN from Google to jQuery.com


coliff:
I'd like to see the jQuery CDN links change to the ones hosted at jQuery.com. HTML5BoilerPlate changed from Google's Hosted Libraries to jQuery's CDN in January 2016 and now say this:

> The jQuery CDN version was chosen over other potential candidates (like Google's Hosted Libraries) because it's fast (comparable or faster than Google by some measures) and, (unlike Google's CDN) is available to China's hundreds of millions of internet users. For many years we chose the Google Hosted version over the jQuery CDN because it was available over HTTPS (the jQuery CDN was not,) and it offered a better chance of hitting the cache lottery owing to the popularity of the Google CDN. The first issue is no longer valid and the second is far outweighed by being able to serve jQuery to Chinese users

REF: https://github.com/h5bp/html5-boilerplate/blob/5.3.0/dist/doc/html.md#jquery-cdn-for-jquery
.

IamManchanda:
100%

pgadige:
I'd want to work on this issue. Could you please assign it to me?

Johann-S:
@pgadige no need to assign this task to you, you only have to create a PR 👍

pgadige:
Should I also include attributes `integrity` and `crossorigin` inside the `<script>` tag to allow browsers to ensure that resources hosted on third-party servers are not tampered with? It was recommended at [jQuery CDN](https://code.jquery.com/)

coliff:
I think so. It's also recommended by Bootstrap's CDN too.
REF: https://www.maxcdn.com/blog/bootstrapcdn-subresource-integrity/

PR #21306, opened 2016-12-06
Update jQuery CDN to use jQuery.com instead of Google

coliff:
Fixes: #21130


bardiharborow:
This uses a SHA265 hash instead of SHA384 (which seems to be the established pattern here). It should be:

```
sha384-3ceskX3iaEnIogmQchP8opvBy3Mi7Ce34nWjpBIwVTHfGYWQS9jwHDVRnpKKHJg7
```

mdo:
Fixed that in a separate commit, thanks @bardiharborow.

PR #21307, opened 2016-12-06
Use github-flavored markdown code block delimiters

bantic:
Use triple-backtick to denote codeblocks in the examples section of `spacing.md`.

The `spacing.md` file appears to use liquid syntax highlighting directives (e.g. `{% highlight scss %}`). This PR changes them to a format that displays better in github.


bantic:
Nevermind, I misunderstood how these markdown files were used. I see that they are compiled to generate the official docs, rather than meant to be read in github.

Issue #21278, opened 2016-12-02
Y-margin of .custom-controls can't be adjusted.


omnibrain:
I tried to increase the *y-spacing* of the [stacked radio buttons](http://v4-alpha.getbootstrap.com/components/forms/#stacked) but noticed this is currently not adjustable. 

There is a variable ```$custom-control-spacer-y``` [in the _variables.scss](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L432) but it isn't used anywhere. I suppose this variable is supposed to defined the margin between the stacked ```.custom-control```s, right?

PR #21314, opened 2016-12-07
Use $custom-control-spacer-y variable for margin-top of custom-controls (#21278)

omnibrain:
This merge request fixes issue #21278


mdo:
This should go on the bottom margin instead so it can override the default `<label>` margins. Alternate fix coming in #21315.

PR #21318, opened 2016-12-08
Change $input-border-color to HEX color value

kvlsrg:
This PR fixes two problems with input groups.

First is hover change of middle (not rounded) border color of input because of input & button borders overlay:

search-hover

Second one is that now input color is not compatible with dark background:

hex-border-color-before

after this PR:

hex-border-color-after

Sorry if I didn't get why now `$input-border-color` has RGBA value. But it will be great to fix this issues without changing z-indexes logic or any other major changes. Thanks!

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -396,7 +396,7 @@ $input-bg:                       $white !default;
 $input-bg-disabled:              $gray-lighter !default;
 
 $input-color:                    $gray !default;
-$input-border-color:             rgba($black,.15) !default;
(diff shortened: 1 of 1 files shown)


bastienmoulia:
Exemple: http://jsbin.com/sagipexaha/1/edit?html,css,output

mdo:
I'm actually liking #22612 more than this PR. Will revisit after Beta 1 ships though, so punting this to the next release cycle.

mdo:
Going with #22612 in Beta 2.

Issue #21266, opened 2016-12-01
Add variables for transitions


tomlutzenberger:
In order to standardize transitions it would be useful to add variables for transitions.
When looking at **[transition usage](https://github.com/twbs/bootstrap/search?l=SCSS&q=%40include+transition&utf8=%E2%9C%93)**, I see that in most cases `ease-in-out` is used as timing-function and 0.2s / 0.6s as duration. These values could be transfered into variables.

Additionally, the mixin could be extended to provide those variables as default values if no parameters are given. So using `@include transition();` could end up in `transition: all 0.2s ease-in-out;` after compiling.

I would be happy to make those changes and open a PR 😃

tomlutzenberger:
Any opinions on this, @mdo ?

virgofx:
Also -- Can we add a variable for the input focus transition (for border and box shadow) in the [13. Forms] section. It's a pain to override the 0.15s in external CSS.

guylepage3:
+1

An example of variables that we've implemented are;

```
$transitions-primary:        opacity .15s linear !default;
$transitions-secondary:      all .2s ease-in-out !default;
```

We are entering a new area of UX dominance where interactions are becoming more and more useful in a designer's tool belt. I feel this is a basic but decent start in for the Bootstrap framework.

mdo:
Fixed by #21345.

PR #21334, opened 2016-12-14
Add variables for transitions

tomlutzenberger:
- The transition shorthand was split up into the single properties to ensure flexibility
- Durations were unified to reduce the amount of different values (`0.2s` => `0.15s`)
- Duration naming scheme differs in general, because time shouldn't be named like `sm` or `lg`
- The two most used timing functions and transition properties were implemented as variables, because other current use cases depend on the behaviour of a specific component, e.g. `height` on `.collapsing`
- The base transition shorthand `$transition-base` (`all .3s ease-in-out`) is used if no parameters were passed to the mixin

See #21266

### DIFF
--- a/docs/assets/scss/_anchor.scss
+++ b/docs/assets/scss/_anchor.scss
@@ -10,7 +10,7 @@
 
 *:hover > .anchorjs-link {
   opacity: .75;
-  transition: color .16s linear;
+  transition: color $transition-duration-short $transition-timing-secondary;
 }
 
 *:hover > .anchorjs-link:hover,
--- a/scss/_animation.scss
+++ b/scss/_animation.scss
@@ -1,7 +1,7 @@
 .fade {
   opacity: 0;
 
-  @include transition(opacity .15s linear);
+  @include transition(opacity $transition-duration-short $transition-timing-secondary);
 
   &.active {
     opacity: 1;
@@ -32,5 +32,5 @@ tbody {
   height: 0;
   overflow: hidden;
 
-  @include transition(height .35s ease);
+  @include transition(height $transition-duration-base ease);
 }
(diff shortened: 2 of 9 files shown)


Starsam80:
I was going to do this, but I never had enough time. With that being said, I have a few questions:
- Why do we need `$transition-property-primary` and `$transition-property-secondary`? Even if they are the most used, changing the variable is likely going to break everything, making it pointless as a variable. You even noted that classes like `.collapsing` wouldn't be able to use these variables, so without much documentation, some people would get confused.
- Why not have per-component transition variables? For example: `$modal-fade-transition: transform .3s ease-out` & `include transition($modal-fade-transition)`. This is because maybe the user doesn't want to affect _every_ transition but only a single one. With this approach it would also be able to add another transition (if there was ever a reason to). You can also combine this with base transition variables: `$modal-fade-transition: transform $transition-duration-base $transition-timing-primary`

tomlutzenberger:
Hey @Starsam80, thanks for the feedback.
I agree, what you describe would be actually a better way to go.
Per-component variables are something I haven't thought about. Smart.

But... then the base variables need to be defined before the components. This either means to put them into another existing section OR make a new one and change the all numbers.

Starsam80:
I think the numbers should be changed. See https://github.com/twbs/bootstrap/pull/20938#issuecomment-256523928

guylepage3:
As I stated here.. https://github.com/twbs/bootstrap/issues/21266

I feel we don't need to be so detailed with these variables. I feel we should have them, yes. But to add too many parameters and variables is adding unnecessary complexity. Let's keep it simple. I feel that two variables is plenty for the time being.

```
$transitions-primary:        opacity .15s linear !default;
$transitions-secondary:      all .2s ease-in-out !default;
```

Also, I feel that the naming convention for these variables should match `btn` names, reducing complexity.

tomlutzenberger:
I'll will work on this tonight, thanks 😃

PR #21338, opened 2016-12-14
fix(readme): locked composer to a proper version

strayobject:
Without a version lock, users get current stable


bardiharborow:
We have the slightly different `composer require twbs/bootstrap:4.0.0-alpha.5` over at [getting-started/download/](https://v4-alpha.getbootstrap.com/getting-started/download/).

PR #21343, opened 2016-12-14
Allow flex on per-mixin-use basis

eatonphil:
I would like to be able to use flex rows in some places but traditional rows in others. Adding a flex variable here lets you disable $enable-flex (the default) but have a custom row class that makes use of the same mixin but with flex support.

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -39,8 +39,8 @@
   }
 }
 
-@mixin make-row($gutters: $grid-gutter-widths) {
-  @if $enable-flex {
+@mixin make-row($gutters: $grid-gutter-widths, $flex: $enable-flex) {
+  @if $flex {
     display: flex;
     flex-wrap: wrap;
   } @else {


bardiharborow:
Hmm. Could you explain your use-case for this a little more?

eatonphil:
@bardiharborow sure. I'd like to default to traditional rows/columns. But for a certain class I'd like to use flex columns. Here's a class I recently wrote:

```
.row-justify {
    @extend .row;
    display: flex;
    flex-wrap: wrap;

    justify-content: space-between;
}
```

However, I'm just hardcoding part of the make-grid mixin there and I'd rather not (in case that mixin changes, etc.).

With my change I could do this instead:

```
.row-justify {
    @include make-grid($grid-gutter-width-base, true);
    justify-content: space-between;
}
```

wolfy1339:
This is invalid now that mdo decided to go flexbox only

mdo:
No longer going to be necessary once #21389 lands—we're going all in on flexbox. Thanks though!

PR #21345, opened 2016-12-14
Add variables for transitions

tomlutzenberger:
Revision of #21334

See also #21266

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -15,7 +15,7 @@
   user-select: none;
   border: $input-btn-border-width solid transparent;
   @include button-size($btn-padding-y, $btn-padding-x, $font-size-base, $btn-border-radius);
-  @include transition($transition-base);
+  @include transition($btn-transition);
 
   &,
   &:active,
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -15,7 +15,7 @@
   user-select: none;
   border: $input-btn-border-width solid transparent;
   @include button-size($btn-padding-y, $btn-padding-x, $font-size-base, $btn-border-radius);
-  @include transition(all .2s ease-in-out);
+  @include transition($transition-base);
 
   &,
   &:active,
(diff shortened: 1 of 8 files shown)


scss/_buttons.scss
```diff
@@ -15,7 +15,7 @@
   user-select: none;
   border: $input-btn-border-width solid transparent;
   @include button-size($btn-padding-y, $btn-padding-x, $font-size-base, $btn-border-radius);
-  @include transition(all .2s ease-in-out);
+  @include transition($transition-base);
```
Starsam80: This should be its own variable: `$btn-transition`

scss/_images.scss
```diff
@@ -16,7 +16,7 @@
   background-color: $thumbnail-bg;
   border: $thumbnail-border-width solid $thumbnail-border-color;
   @include border-radius($thumbnail-border-radius);
-  @include transition(all .2s ease-in-out);
+  @include transition($transition-base);
```
Starsam80: This should be its own variable: `$thumbnail-transition`

Starsam80 (changes requested):
You didn't have to close the old PR, `git push --force` does exist. If you fix the 2 notes I made, LGTM.

First time doing a GitHub review. No idea how this is going to work

guylepage3:
@tomlutzenberger 
I agree with @Starsam80, you shouldn't keep moving this around from issue to issue to PR to PR. It makes it very difficult for others contributing to follow along.

Again as stated way back in issue https://github.com/twbs/bootstrap/issues/21266 & https://github.com/twbs/bootstrap/pull/21334
> I feel that the naming convention for these variables should match btn names, reducing complexity.

The reason being is that this could limit the number of variables to a standard set. Bootstrap is not created to offer "all" solutions but a couple. It's just a framework. Not a full huge behemoth framework that does everything. IMO.

tomlutzenberger:
Yes, I have to admit that I haven't thought of the forced push.
I'll fix those 2 things.

Starsam80 (approved):

PR #21347, opened 2016-12-15
Add example of <nav>-based JS tabs to the docs

ghost:


bardiharborow:
The build fails with `Bad value “tablist” for attribute “role” on element “nav”`. Also is there a reason you are PRing mdo's commits?

patrickhlauke:
you should not override the native navigation role of `nav` with `tablist`. otherwise you're removing the information that this is a (primary) navigation block. nest some further container inside `nav` instead. and yes, a clean PR would be appreciated

patrickhlauke:
Also yes, this PR just references a commit from @mdo ... so quite pointless?

PR #21356, opened 2016-12-16
Add 'system-ui' to font stack.

### LINKED ISSUE #21354: Use new "system-ui" generic font family in Chrome.
Per [csswg CSS Fonts 4](https://drafts.csswg.org/css-fonts-4/#valdef-font-family-system-ui), Chrome is shipping a [new `system-ui` generic font family](https://www.chromestatus.com/feature/5640395337760768) in Chrome 56, which will replace `BlinkMacSystemFont`. [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1226042) and [Safari](https://bugs.webkit.org/show_bug.cgi?id=151493) support is under development. We should consider adding it to our font stack and possibly removing the old name once the change hits stable.

bardiharborow:
Fixes #21354.

### DIFF
--- a/docs/content/reboot.md
+++ b/docs/content/reboot.md
@@ -41,7 +41,9 @@ The default web fonts (Helvetica Neue, Helvetica, and Arial) have been dropped i
 $font-family-sans-serif:
   // Safari for OS X and iOS (San Francisco)
   -apple-system,
-  // Chrome for OS X (San Francisco)
+  // Chrome >= 56 for OS X (San Francisco), Windows, Linux and Android
+  system-ui,
+  // Chrome < 56 for OS X (San Francisco)
   BlinkMacSystemFont,
   // Windows
   "Segoe UI",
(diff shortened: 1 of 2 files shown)


klimashkin:
@bardiharborow Why `system-ui` is not first in list, why it should be after `-apple-system`?

wolfy1339:
@kilmaskin This was reverted in #22377 because​ it was causing issues

GrimLink:
@mdo I don't understand why system-ui i has been removed from the stack.
Instead move the "Segoe UI" before the value in the stack instead.

Maybe like this:
`$font-family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, Roboto, "Helvetica Neue", Arial, sans-serif !default;`
This way we can still have the advantage of the system-ui without having the Windows bug.

mdo:
@GrimLink Read #22328 for the background on why it was reverted outright. We didn't test, as far as I know, changing the font stack's order, but if you have a demo of it fixing the problems listed in the referenced issue, please open a new issue to discuss.

PR #21368, opened 2016-12-19
Use brand color for progress bar

szastupov:
This patch changes default progress bar color to brand primary. Contextual alternatives already respect branding.


mdo:
I recall being against this at some point, but I'm indifferent to it now—don't really see a reason to keep it separate from this variable, so thanks!

PR #19978, opened 2016-05-28
Dropdown: close menu when focusing outside element

pvdlg:
Currently the behavior of dropdown menu is inconsistent for mouse and keyboard users:
- when menu is open, clicking outside, close it
- when menu is open, focusing outside (tab key), doesn't close it

The changes in this PR close the dropdown menu when focusing on an element outside the dropdown menu.
It's handled in a similar fashion it's currently done for click events (except that focusin events inside the menu don't close it contrary to click events).

PR #21375, opened 2016-12-20
Dropdown: close menu when focusing outside element

pvdlg:
Re-open #19978 following a mistake while resolving conflicts.

Currently the behavior of dropdown menu is inconsistent for mouse and keyboard users:
- when menu is open, clicking outside, close it
- when menu is open, focusing outside (tab key), doesn't close it

The changes in this PR close the dropdown menu when focusing on an element outside the dropdown menu.
It's handled in a similar fashion it's currently done for click events (except that focusin events inside the menu don't close it contrary to click events).


vanduynslagerp:
All the Hound violation are on the unit test file even though my modifications follow the same pattern as the other unit test.
Should I fix these violation ? Or should Hound configuration be modified to ignore the qunit directory ?

bardiharborow:
If you rebase now, the Hound issue should be fixed.

mdo:
Looking at these tests I think we need to redo those examples with the new dropdown markup. Going to make an issue for that later.

vanduynslagerp:
I updated the unit test to the new dropdown markup

PR #21388, opened 2016-12-21
Prevent nested carousel items from being updated

GodLesZ:
As it's noted in/since #14441 nested carousel's are not supported.
Anyways this fix will prevent inherit carousel item's from being updated from a parent carousel.

### DIFF
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -115,7 +115,8 @@
   }
 
   Carousel.prototype.slide = function (type, next) {
-    var $active   = this.$element.children('.carousel-inner').children('.item.active')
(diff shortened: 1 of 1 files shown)


mdo:
Why change the JS to account for this behavior if we don't intend to support it at all?

mdo:
Also, just noticed this is against v3—and we're not actively developing that right now. Closing out for that reason, though I thank you for the suggestion and your time! <3

GodLesZ:
@mdo because Bootstrap is known as kindly and working out-of-the-box. So this would be just kind ;)

Issue #20506, opened 2016-08-11
V4 input-group inside btn-toolbar requires fixed width


bpierson:
Possibly this is the expected behavior, but when adding an `input-group` to a `btn-toolbar` (which appears to be common usage, since it is specifically accounted for in bootstrap.css), an `input-group` will always go to full width and wrap to the next line, unless it is assigned a fixed width.

As I said, this might be expected behavior, but is not documented (that I could find).

[Fiddle](https://jsfiddle.net/53qwqdob/)

```
<div class="btn-toolbar" role="toolbar">
    <div class="btn-group" role="group">
        <button class="btn-secondary btn" type="button">
            <span>One</span>
        </button>
        <button class="btn-secondary btn" type="button">
            <span>Two</span>
        </button>
    </div>
    <!-- ADD style="width: 500px;" or some other fixed width to fix wrapping. width: auto does not work. -->
    <div class="input-group">
        <span class="input-group-btn">
            <button class="btn btn-info" title="clear search" type="button">
                <i class="fa-fw fa-times fa"></i>
            </button>
        </span>
        <input class="form-control" spellcheck="false" type="text" />
        <div class="input-group-btn" role="group">
            <button class="btn btn-info" type="button">
                <i class="fa-fw fa-search fa"></i>
            </button>
            <button aria-haspopup="true" class="btn-info dropdown-toggle btn" data-toggle="dropdown" type="button"></button>
            <div class="dropdown-menu-right dropdown-menu" role="menu">
                <a class="dropdown-item" href="#">Help on Search</a>
                <div class="dropdown-divider"></div>
                <div class="dropdown-header">STORED SEARCHES</div>
                <a class="dropdown-item" href="#">This and That</a>
                <a class="dropdown-item" href="#">Last Week</a>
                <a class="dropdown-item" href="#">Closed Only</a>
            </div>
        </div>
    </div>
</div>
```

PR #21394, opened 2016-12-22
Flexbox button group

### LINKED ISSUE #20530: Issue with .btn-group in td.text-nowrap
Please check this demo to reproduce the issue: http://www.bootply.com/wWM2a9R8Mt#

The unexpected horizontal overflow seems to be caused by the `float: left` of the buttons in the `btn-group`. For me it was unexpected that `btn-group` contains floating elements. ~~If floating is really needed, it would be good to include the `clearfix` in `btn-group`s.~~ It seems like adding the `clearfix` class to `btn-group` doesn't solve the issue in itself.

How could I solve this table cell layout issue?

Is there any way to avoid floats in `btn-gruop`s in future versions? I mean for example `input-group`s are implemented without floats, so `btn-group`s might have an alternative implmentation too.

mdo:
This PR starts to update the button group with flexbox styles instead of floats. I'm not done here though; I got hung up on getting the vertical variation to work right. Any guidance on getting these things to size correctly would be amazing.

screen shot 2016-12-21 at 11 30 16 pm

---

Also happens to fix #20530.


Starsam80:
I don't know if you found a solution for the vertical variation yet, but adding this makes it work:
```scss
.btn-group-vertical {
  display: inline-flex;

  .btn,
  .btn-group {
    width: 100%;
  }
}
```

mdo:
@Starsam80 😭 I tried `inline-flex` but without the `width: 100%` in there.

mdo:
Pushed up changes with @Starsam80's fix, fixed #20506 while I was in there, and then added some more docs examples to illustrate how to use input groups in toolbars as a follow-up.

PR #21402, opened 2016-12-23
New width and height utils

mdo:
Adding 25%, 50%, and 75% to the mix for later use with cards. Mainly doing this for cards (PR coming next).


mdo:
For those following along at home, here's how these look now in the docs btw.

screen shot 2016-12-22 at 3 42 46 pm

Issue #20408, opened 2016-07-29
Right padding for .media-left increases element height in IE 10 with flexbox enabled


heikojansen:
Using Flexbox enabled v4 Alpha 3 CSS and this HTML:

``` html
<!Doctype html>
<html>
  <head>
    <title>IE 10 media object flexbox problem</title>
    <link rel="stylesheet" href="bootstrap.css" />
    <style type="text/css">
      body { background-color: #ffa; max-width: 500px; margin: 0 auto; }
      .media { background-color: #aff; }
      .media-left { background-color: #faa; }
    </style>
  </head>
  <body>
    <div class="media">
      <a href="#" class="media-left">
        <img alt="64x64" class="media-object" style="width: 64px; height: 64px; background-color:#999" src="">
      </a>
      <div class="media-body">
        <h4 class="media-heading">Media heading</h4>
        Text needs to be short to expose the problem.
      </div>
    </div>
  </body>
</html>
```

I get this in Firefox 47.0.1 and also in IE11:
![ff47](https://cloud.githubusercontent.com/assets/295487/17246671/dfbad880-558d-11e6-8ef1-e57dbe765dcc.png)

But this in IE 10 (Virtual Machine with IE 10 on WIN 7 downloaded from modern.ie):
![ie10](https://cloud.githubusercontent.com/assets/295487/17246760/70561e68-558e-11e6-9b8a-c36b8c8485b6.png)

mdo:
You're running into [Flexbug #12](https://github.com/philipwalton/flexbugs#12-inline-elements-are-not-treated-as-flex-items) where inline elements aren't counted as flex items in IE10-11. Set an explicit non-inline display (like `block`, `inline-block`, or `flex`) to fix it.

I'm adding a note to the docs in #21405 to help folks in the future.

PR #21405, opened 2016-12-23
v4: Flex media redux

mdo:
This PR overhauls the `.media` component by reducing it to only two required classes, `.media` and `.media-body`. Everything else can be accomplished with utility classes for padding, margin, and alignment. This also overhauls the docs to improve copy, add more examples, and improve existing examples with the new markup and utils usage guidelines.

While working on this, I took a look at #20408. Turns out that's a known bug with inline flex items not being counted in IE10-11. I've added a docs warning for that.

_As an aside, I'll likely be tackling some flex utils name changes and docs updates in a follow-up PR. We're missing some key flex utils and the class names could use a small change for consistency with our other property-value utils (e.g., `d-flex`)._

### DIFF
--- a/docs/layout/media-object.md
+++ b/docs/layout/media-object.md
@@ -5,50 +5,53 @@ description: Documentation and examples for Bootstrap's media object to construc
 group: layout
 ---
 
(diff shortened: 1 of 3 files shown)


Quy:
Reminder to remove unused media variables.

PR #21407, opened 2016-12-23
v4: Various flexbox fixes and updates

mdo:
Starting a catchall branch of sorts to clean some stuff up around the move to flexbox.

### Changes so far

- Fixed inline forms in responsive navbars. The use of `my-` utilities on the buttons here caused the inputs to grow to match the computed height of the button *plus* those margins, so things looked broken. I've changed the utilities here and fixed it all up.

- Updated the three remaining examples to use the new navbar—starter template, jumbotron, and offcanvas are now all up to speed.

- Speaking of the offcanvas example, that's also been fixed. I pulled in the changes from #19832.

- Huge flexbox utilities overhaul. Replaced all existing `.flex-*` utilities that had custom names with new utilities named after their `property: value` pair (e.g., `.flex-items-between` is now `.justify-content-between`). Also added brand new docs to showcase all these new utils.

- Rewrote modal docs while aiming to fix some double borders in modal examples.

---

#### Flexbox bugs
- [x] Inline forms in navbar examples across the board are fubared
- [x] Forgot to update to flexbox navbar in starter template, jumbotron, and offcanvas examples.
- [x] Fix offcanvas menu breakage
- [ ] Inline form buttons of mixed sizes are resized to be equal height
- [ ] Inline form checkbox text is too close to the control
- [x] Custom selects are wrong height
- [ ] Static control example for inline form lacks padding utils
- [x] Inline help text example is poorly aligned

#### Flexbox todos
- ~~Remake form checks/radios with flexbox instead of abs position~~ nevermind i'm an idiot, can't do this because of our markup
- ~~Remake dismissible alert with flexbox instead of abs position~~ would require extra markup to do, skipping
- ~~Redo modal positioning with flexbox (hopefully we can vertically center and keep the scroll)~~ nevermind, going to do this later; too many other changes to make alongside this.
- [x] Remake navs with flexbox instead of floats
- [ ] Remake pagination with flexbox instead of floats
- [x] Blow out flex utils, document them all
- [ ] Modal with flexbox—see #21425

#### Docs bugs
- [x] Modal demos have nested example markup

### DIFF
--- a/docs/_data/breakpoints.yml
+++ b/docs/_data/breakpoints.yml
@@ -1,0 +1,29 @@
+- breakpoint: xs
+  abbr: ""
+  name: Extra small
+  min-width: 0px
+  container: ""
+
+- breakpoint: sm
+  abbr: -sm
+  name: Small
+  min-width: 576px
+  container: 540px
+
+- breakpoint: md
+  abbr: -md
+  name: Medium
+  min-width: 768px
+  container: 720px
+
+- breakpoint: lg
+  abbr: -lg
+  name: Large
+  min-width: 992px
+  container: 960px
+
+- breakpoint: xl
+  abbr: -xl
+  name: Extra large
+  min-width: 1200px
+  container: 1140px
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -56,6 +56,7 @@
     - title: Clearfix
     - title: Close icon
     - title: Colors
+    - title: Flexbox
     - title: Display property
     - title: Image replacement
     - title: Invisible content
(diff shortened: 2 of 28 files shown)


mdo:
Working on adding more flexbox utilities and documenting them well. This might be overly verbose, so please let me know if it's seeming a bit over the top. Each property will get a blurb and examples. Some will have tables, too.

screen shot 2016-12-23 at 11 48 08 am

For responsive versions, I've started by listing them all out like so. Too much? Just right?

screen shot 2016-12-23 at 11 48 13 am

mdo:
Playing with it more, this feels a bit better and less repetitive.

screen shot 2016-12-23 at 11 55 23 am

mdo:
This branch is running away from me a bit, so going to call it as-is once this next build passes and then move on to another branch.

mdo:
Will tackle remaining todos in separate PRs, btw.

PR #21418, opened 2016-12-23
U can't use command from readme

cisuminas:
meteor add twbs:bootstrap@=4.0.0-alpha.5 does not work.. So it is better to write the command that works, or remove meteor section completly?

see https://github.com/twbs/bootstrap/issues/20389


bardiharborow:
That command should work. I'll look into it.

bardiharborow:
Fixed by #21420. :)

PR #21419, opened 2016-12-24
Higher priority for .btn-primary

Yahasana:
`.btn-primary` should has higher priority than `.btn-secondary`

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -59,11 +59,11 @@ fieldset[disabled] a.btn {
 // Alternate buttons
 //
 
-.btn-primary {
-  @include button-variant($btn-primary-color, $btn-primary-bg, $btn-primary-border);
-}
 .btn-secondary {
   @include button-variant($btn-secondary-color, $btn-secondary-bg, $btn-secondary-border);
+}
+.btn-primary {
+  @include button-variant($btn-primary-color, $btn-primary-bg, $btn-primary-border);
 }
 .btn-info {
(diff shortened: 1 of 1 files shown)


Yahasana:
it'll be easier to change the secondary state to primary by just toggle `btn-primary` class through js. now i have to remove the `btn-secondary` then add `btn-primary`

Starsam80:
@Yahasana Did you try using [`$.toggleClass('btn-primary btn-secondary')`](https://api.jquery.com/toggleclass/)?

Yahasana:
```html
<button class="btn btn-secondary">default</button>
```
```js
$el.toggleClass("btn-primary") ==> 
    <button class="btn btn-secondary btn-primary">primary</button>
```
again return to default state
```js
$el.toggleClass("btn-primary") ==> 
    <button class="btn btn-secondary">default</button>
```
```js
$el.toggleClass('btn-primary btn-secondary') ==>
    <button class="btn">incorrect button</button>

Yahasana:
BTW, in BS 3.x `btn-primary` has higher priority than `btn-default`. why don't keep this rule in v4

Yahasana:
@Starsam80 after a second thought, you are right. `$el.toggleClass('btn-primary btn-secondary')` works. but `$el.toggleClass('btn-primary')` is 💯 clean

mdo:
> why don't keep this rule in v4

These classes are individual modifier classes that aren't intended to be stacked on top of one another. Use one or the either. There was no rule for this in v3 either; just kind of happenstance that one was before the other.

PR #21422, opened 2016-12-24
Update toc numbering

Quy:


mdo:
😭 Thank you. I'm going to straight up delete these numbers next.

PR #21426, opened 2016-12-25
v4: Mixing auto margins with flexbox

mdo:
screen shot 2016-12-24 at 6 11 21 pm

Turns out auto margins and flexbox make an amazing team. I discovered this while working on #21425 and really wanted to put something together for it. Using `margin-{direction}: auto` on a flex item will automatically place it on the opposite end of the flex items (when mixed with `justify-content` or `align-items`, at least). This makes for some awesome layout opportunities that are perfect for grids, modal footers, cards, and more.

### DIFF
--- a/docs/utilities/flexbox.md
+++ b/docs/utilities/flexbox.md
@@ -59,69 +59,6 @@ Responsive variations also exist for `.flex-row` and `.flex-column`.
 - `.flex{{ bp.abbr }}-row`
 - `.flex{{ bp.abbr }}-column`{% endfor %}
 
-## Auto margins
-
-Flexbox can do some pretty awesome things when you mix `justify-content` with `margin-right: auto` or `margin-left: auto` on a particular flex item. For example, we can move all flex items to the right, but keep one on the left like so.
-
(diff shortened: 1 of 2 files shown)


Starsam80:
Correct me if i'm wrong, but `margin-top: auto` and `margin-bottom: auto` always evaluate to 0, so having classes for these seem pointless. If I am wrong, then you should add a `.my-auto` class (`margin-top: auto; margin-bottom: auto;`)

mdo:
Hmm, doesn't seem to be the case with flexbox. Here's how the `align-items` and top/bottom utils looks in the docs example I've created.

screen shot 2016-12-24 at 6 36 47 pm

Starsam80:
Huh, never thought about it being in a flex container. Most of the sources I saw online were before flexbox, so that's where I must have got the idea that `margin-{top,bottom}: auto` didn't work.

For completeness, I think you should also add the `.my-auto` class I mentioned before and a `.m-auto { margin: auto; }` class

Starsam80 (approved):

Issue #21415, opened 2016-12-23
disabled btn-link button has a different style on hover


kosl90:
- Operating system and version : Windows 7
- Browser and version : Google Chrome 55.0.2883.87 m (64-bit)
- Reduced test cases and potential fixes using [JS Bin](https://jsbin.com)
https://jsbin.com/yetixexohi/edit?html,output

potential fixes with:

.btn-link:disabled {
    color: #818a91;
    text-decoration: none;
}

mdo:
Dupe of #21235.

mdo:
(Not an exact dupe, but counting it as similar given the disabled stuff. Revisiting that now...)

kosl90:
I had looked into that, you guys did a great job again :+1:

Issue #19189, opened 2016-02-13
disabled <button> and <a> tags styling should not be affected by :active state


patchee500:
Generally it is quite hard to force the `:active` state on a `<button>` that has the `disabled` attribute or an `<a>` with class `disabled`. If you however manage to do so (e.g. in dev tools) the buttons appearance will change. It will look like the button is being clicked.
You can also achieve this behavior in Firefox by trying to drag a `<button>` with `disabled` attribute.
Here's a fiddle: https://jsfiddle.net/ekswzwgz/7/

mdo:
Merged #21439 to address this.

PR #21439, opened 2016-12-26
v4: Simplify button states

mdo:
This PR drastically simplifies our button states to address #19189, #21235, and #21415. My hope is we can simplify the selectors, reduce the number of selectors overall, and make it easier to customize our buttons. Here's a rundown of what's changed and why.

- **Moved active state after the disabled state.** This way, disabled *active* buttons automatically inherit the proper changes to visual styles and simply modifies the opacity. Previously we had to reset the `background-color` and `border-color` across all states.

- **Unchains the focus state from active state.** There's no need to add extra selectors to the original focus state within `&, &:active, &.active { ... }`—we just need to set the focus styles.

- **Removes all hover, focus, and disabled styles from the active states.** Now that active is after disabled in Sass, we can easily inherit the proper styles. Similarly, hover and focus will inherit automatically as well. This does remove the ability to see a hover on active, but I'm okay with that as it's more akin to native behaviors.

See the new CSS in action at https://output.jsbin.com/homiyos.

/cc @cvrebert @patrickhlauke

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -11,7 +11,6 @@
   text-align: center;
   white-space: nowrap;
   vertical-align: middle;
-  cursor: pointer;
   user-select: none;
   border: $input-btn-border-width solid transparent;
   @include button-size($btn-padding-y, $btn-padding-x, $font-size-base, $btn-border-radius);
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -17,15 +17,12 @@
   @include button-size($btn-padding-y, $btn-padding-x, $font-size-base, $btn-border-radius);
   @include transition($btn-transition);
 
-  &,
-  &:active,
-  &.active {
-    &:focus,
-    &.focus {
-      @include tab-focus();
-    }
+  &:focus,
+  &.focus {
+    @include tab-focus();
   }
 
+  // Share hover and focus styles
   @include hover-focus {
     text-decoration: none;
   }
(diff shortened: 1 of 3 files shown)


patrickhlauke:
not had a chance to test this out live, but just looking at the description cold:

> Unchains the focus state from active state

> Removes all hover, focus, and disabled styles from the active states

is it still possible to see a difference between a focused active and focused inactive buttons? if not, then i'd strongly suggest to reconsider this

mdo:
Instead of a different `background-color` and `border-color`, focused buttons have a new "outline" via `box-shadow`. I wish I could use `outline`, but most browsers don't round it to match `border-radius` and it causes the `background-color` to bleed out passed the rounded corners. 😢 

So, instead we have `box-shadow: 0 0 0 2px rgba($border, .5);`, which looks pretty good I think. Originally I was planning on having a pixel of negative space around the button (like an `outline-offset`, which no one supports), but that's difficult without setting another opaque shadow (e.g., `#fff`). As such, we have this custom shadow.

The benefit of these new selectors, @patrickhlauke, is that the `:focus` styles are automatically inherited into focused active buttons.

screen shot 2016-12-26 at 10 10 29 pm

patrickhlauke:
@mdo on that screenshot, the "Focus state" and "Active focus state" are indistinguishable. assuming it's a mistake in the screenshot, right?

subjectively, i'd say the focus outline (shadow) could do with being slightly more pronounced to make it more obvious.

on a positive note, removing the hover state at least reduces the number of iterations, so that's definitely a good thing.

mdo:
@patrickhlauke Gah, fixed the screenshot and demo. I had left in the active bg and what not on the regular `:focus` state for the button variant mixin. Fixed now.

I'll also say that the focus outline _is_ more apparent when tabbing through, as opposed to everything being highlighted here in the demo. That said, I'll play with making the colors more vibrant here.

md5:
I just noticed the change introduced by caef376 as part of this PR, namely to remove the `cursor: pointer` from buttons.

While I understand the logic behind this change and I support the removal of that style, it introduced an inconsistency between `button.btn` and `a.btn` in that the former now has an arrow cursor and the latter has a hand cursor. It seems to me that the intent of applying `.btn` to an `<a>` tag is to make it look, act, and feel like a button, so I would think the right fix would be to set `cursor: default` on `.btn` instead of removing the `cursor` property entirely.

I can open a separate issue or provide a PR if that makes sense, but I wanted to start the discussion in the context where the original change was made.

md5:
I've opened https://github.com/twbs/bootstrap/pull/22693 to make the change I'm suggesting.

PR #21442, opened 2016-12-27
v4: Move `_custom.scss` after vars and mixins

### LINKED ISSUE #21409: Correct import order of _custom.scss and _variables.scss?
[Docs explain:](https://v4-alpha.getbootstrap.com/getting-started/options/#customizing-variables) 

> Copy and paste relevant lines from `_variables.scss` into the custom file

Currently `bootstrap.scss` imports `_custom.scss` **before** `_variables.scss` which seems to make any variable declaration within `_variables.scss` unavailable for use within `_custom.scss`.

    @import "custom"; // (contains customization for: $lead-font-size)
    @import "variables";

    Error: Undefined variable: "$lead-font-size".

Whereas when I swap the order of these two files it compiles without errors.

    @import "variables";
    @import "custom"; // (contains customization for: $lead-font-size)

mdo:
Allows reuse of vars and mixins, while also allowing customization of vars. Fixes #21409.

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -27,7 +27,6 @@ html {
 // Variables
 //
 
-@import "custom";
 @import "variables";
 
 //
@@ -38,5 +37,7 @@ html {
 @import "mixins/breakpoints";
 @import "mixins/grid-framework";
 @import "mixins/grid";
+
+@import "custom";
 
 @import "grid";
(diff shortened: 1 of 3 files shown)


maimairel:
I'm not sure if sass works this way, but after moving _custom.scss after _variables.scss, some variables inside _custom.scss are ignored. I have `$gray-dark` set to #000000 inside _custom.scss, but the compiled CSS have it still its default value. What doesn't make sense is, `$grid-breakpoints` and `$container-max-widths` does work while the other custom values are ignored.

PR #21443, opened 2016-12-27
Don't show `.bd-*` classes to the user

Starsam80:
I think screenshots show this change the best:
Before: Before
After: After

`.bd-*` classes aren't in the regular bootstrap css, so when people copy these class names, they aren't going to do anything.

A small bug (but can be used as a feature) I noticed is when other class names appear after the `bd-*` class and they disappear as well, so something like `<div class="d-flex bd-highlight p-2">` would display as `<div class="d-flex">`

### DIFF
--- a/docs/_plugins/highlight_alt.rb
+++ b/docs/_plugins/highlight_alt.rb
@@ -63,7 +63,12 @@ eos
       end
 
       def remove_example_classes(code)
+        # Find `bd-` classes and remove them from the highlighted code. Because of how this regex works, it will also
+        # remove classes that are after the `bd-` class. While this is a bug, I left it because it can be helpful too.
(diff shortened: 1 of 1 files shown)


mdo:
Ohhhhh shit what, I was thinking about this earlier. Some of those flexbox examples use a bit of mixed techniques of manually showing the example and code, others use the before style shown above. Dope that we can use that "bug" to hide those classes as needed.

Starsam80:
I've added comments and another regex to my latest commit. Again, I think screenshots are needed:
Before: image
After: image
I did rearrange the order of classes for that screenshot, but all I can say is that it looks a lot nicer.

mdo:
<3 thanks!

PR #21453, opened 2016-12-28
deprecated syntax in docs

silvermanj7:
Changed offsetting commands for the grid. The previous commands were frustratingly useless. A stackOverflow search revealed the real commands. It is now updated for others.


mdo:
The current syntax in our docs is correct.

PR #21464, opened 2016-12-29
Make container `position: relative`

mdo:
Allows for better/easier positioning of elements within (e.g., a toggler in a navbar). Alternative is to add a relative positioning utility, but I think this is more intuitive.


tomlutzenberger:
Just as a reminder:

That change makes it necessary to place elements, which should be absolute to e.g. the body or main wrapper need to be placed **outside** of the container. Sometimes this might not be possible due to the logic of a framework or system.

PR #21466, opened 2016-12-29
v4: Switch to jQuery Slim

mdo:
This supersedes #20203, changing our jQuery version to the latest slim version while updating our examples to match the new CDN and what not.


Starsam80:
I think `docs/assets/js/vendor/jquery.min.js` and `js/tests/vendor/jquery.min.js` need to be updated as well

bardiharborow:
`js/tests/vendor/jquery.min.js` is only referenced in `js/tests/index.html`, so I suppose you could use `../../docs/assets/js/vendor/jquery.min.js`. If you do this, you should do the same to Tether.

bardiharborow:
You're missing [js/tests/index.html#L9-L10](https://github.com/twbs/bootstrap/blob/a3a5b21307d2a630d3e39c2619f90c19d7b0b626/js/tests/index.html#L9-L10], and then LGTM.

PR #21467, opened 2016-12-29
Close #21249 : Add a Unit test for #21227

### LINKED ISSUE #21249: Add a Unit test for #21227
https://github.com/twbs/bootstrap/pull/21227

it's important to add units tests for the futur of Bootstrap

/cc @mdo @bardiharborow

Johann-S:
Add a unit test for #21227 

Should I add same test for Tooltip ?

Fixes #21249 

/cc @bardiharborow @mdo

### DIFF
--- a/js/tests/unit/popover.js
+++ b/js/tests/unit/popover.js
@@ -332,4 +332,36 @@ $(function () {
       done()
     }
   })
+
+  QUnit.test('should hide popovers when their containing modal is closed', function (assert) {
+    assert.expect(1)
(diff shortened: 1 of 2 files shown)


mdo:
> Should I add same test for Tooltip ?

Probably? Since popovers are built on tooltips, would a single tooltip unit test cover both though?

Johann-S:
It seems we can find aproximatively same tests in Tooltips and Popover so I'll add a test for tooltip

Issue #20631, opened 2016-09-05
Getting to v4


mdo:
Long story short, v4 has taken far too long and I know we're struggling to find enough time to make significant dents across the open issues and PRs. To help get through it all, I'm calling it on v3 and outlining the big steps remaining here so folks know what's in store.

Here's how we get from a string of alphas to a finalized v4 release:

- [ ] **Stop all work on v3—today.** The open issues, split dev setup, and more holds us back from focusing entirely on v4. I'll close all remaining `v3` issues and milestones. Any new changes to v3 will be sporadic and highly irregular.

- [ ] **Switch v4-dev to the default branch.** Building on step one, this makes `v4-dev` the code you see when you visit the repo. I'll cut a `v3-dev` branch and leave `master` as-is until v4 ships.

- [ ] **Move the docs sites.** The biggest pain will be moving our docs pages around (specifically, moving `v4-alpha.getbootstrap.com` to `getbootstrap.com`). The current v3 docs will be moved to a sub-directory like our other hosted versions.

- [ ] **Ship some betas.** After the alpha releases take care of the major breaking changes, we'll ship some beta releases to work out the final kinks before RC and final releases.

There's a ton this list misses, but these are the broad strokes that get us to v4. Holler with any questions or comments as I push us through this.

/cc @twbs/team

cvrebert:
> The current v3 docs will be moved to a sub-directory like our other hosted versions.

@mdo But this shouldn't break any links, right? We just need to keep things under `/<version number>/` directories for v4 going forward.

AdamHess:
A large part of the community thinks this is premature and a terrible idea until bootstrap 4 is released. 

I personally have not started working with bootstrap 4 explicitly because it is still in Alpha. I as many others would still benefit from fixes, tweaks or features until bootstrap 4 is completely released. 

see more discussion: 
https://www.reddit.com/r/web_design/comments/51cfo4/bootstrap_v3_is_no_longer_maintained/

julianlam:
@AdamHess I feel a large part of the negative sentiment in that reddit submission was due to a selective bias present in the title: "Bootstrap v3 is *No Longer Maintained*", like @mdo and @cvrebert decided to set everybody adrift and wave goodbye, when this obviously isn't the case.

As @mdo has pointed out, commits to v3 will still occur, just irregularly and likely only when needed.

It's all really in how you look at it... *No Longer Maintained*, vs *Declared Stable*

[...]

PR #21483, opened 2016-12-31
v4: Rip out IE compatibility mode meta tags

### LINKED ISSUE #21476: Remove IE-EDGE from starter template as not required now ( Flexbox only, IE9 Drop )
There is no need for the below code so please remove this from starter template 
`<meta http-equiv="X-UA-Compatible" content="IE=Edge">`

Source: http://stackoverflow.com/a/6771584 
( Same link that @mdo posted in his codeguide )

Here are the important quotes to consider from this comment 

> Depending upon what Microsoft browsers you support you may not need to continue using the X-UA-Compatible tag. If you need to support IE 9 or IE 8, then I would recommend using the tag. If you only support the latest browsers (IE 11 and/or Edge) then I would consider dropping this tag altogether. 

ie-edge

mdo:
Fixes #21476.

Updates the docs, examples, and JS tests to remove the relevant `<meta>` tag. Also updates our docs to remove mention of some irrelevant IE8-9 business. This info will still be there in v3's docs, so we can update it later to link to things as needed.

### DIFF
--- a/docs/getting-started/browsers-devices.md
+++ b/docs/getting-started/browsers-devices.md
@@ -105,70 +105,15 @@ Similarly, the latest versions of most desktop browsers are supported.
 
 For Firefox, in addition to the latest normal stable release, we also support the latest [Extended Support Release (ESR)](https://www.mozilla.org/en-US/firefox/organizations/faq/) version of Firefox.
 
-Unofficially, Bootstrap should look and behave well enough in Chromium and Chrome for Linux, Firefox for Linux, and Internet Explorer 9 and below, though they are not officially supported.
+Unofficially, Bootstrap should look and behave well enough in Chromium and Chrome for Linux, Firefox for Linux, and Internet Explorer 9, though they are not officially supported.
 
 For a list of some of the browser bugs that Bootstrap has to grapple with, see our [Wall of browser bugs]({{ site.baseurl }}/browser-bugs/).
 
 ## Internet Explorer
 
-Internet Explorer 10+ is supported, however, IE9 down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported.
+Internet Explorer 10+ is supported; IE9 down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported in IE10, or require prefixed properties for full functionality. Visit [Can I use...](http://caniuse.com/) for details on browser support of CSS3 and HTML5 features.
 
-<div class="table-responsive">
-  <table class="table table-bordered table-striped">
-    <thead>
-      <tr>
-        <th scope="col">Feature</th>
-        <th scope="col">Internet Explorer 9</th>
-        <th scope="col">Internet Explorer 10</th>
-      </tr>
-    </thead>
-    <tbody>
-      <tr>
-        <th scope="row"><a href="https://developer.mozilla.org/en-US/docs/Web/CSS/transition"><code>transition</code></a></th>
-        <td class="text-danger">Not supported</td>
-        <td class="text-success">Supported</td>
-      </tr>
-      <tr>
-        <th scope="row"><a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#attr-placeholder"><code>placeholder</code></a></th>
-        <td class="text-danger">Not supported</td>
-        <td class="text-success">Supported</td>
-      </tr>
-      <tr>
-        <th scope="row"><a href="https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes">Flexbox</a></th>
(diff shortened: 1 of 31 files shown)


IamManchanda:
super awesome @mdo

Quy:
`introduction.md` needs to be updated.

bardiharborow (approved):

IamManchanda:
Exactly @Quy .... i didn't notice. 
That page was the ticket infact so i am surprised that is not commited ? 

cc /@mdo @bardiharborow  
Would humbly request you to plz commit that also .... cheers :+1:

IamManchanda (approved):

IamManchanda (approved):

Starsam80:
I've been wondering this for a while now (and this probably isn't the place to talk about it, I just got reminded of it because of the change adding the note about v3), but v3 status is "no more features, [only critical bug fixes and documentation improvements](https://github.com/twbs/bootstrap/issues/20631#issuecomment-246120195)", but every v3 issue and pull request I've seen has been closed. Just how critical are we talking here? Why not merge #21341 which seems like a fairly big problem to some people? And why not merge pull requests like #21065 that just add a documentation improvement?

mdo:
@Starsam80 I'm labelling all those issues as they get closed. Anything I can come back to when v4 is released, I'll come back to. Otherwise, just talking about mental load and balancing work, life, and OSS, I can only focus on one major code version at a time. Trying to do both was no good for anyone for far too long. 😞

Quy:
...and this `Forcing Internet Explorer to use its latest rendering mode (read more)` ;)

IamManchanda:
@Starsam80 sorry but please support mark on this .... dont worry we all will get back to v3 rather sooner than later .... at this time v4 should be the focus ... These coming months, we should work hard for this v4 release first and then think about anything else like v3.

mdo:
@IamManchanda Thanks, but I appreciate @Starsam80 pushing back on that stuff. I did make some copy changes there and he's right that it's not all that accurate in how we're "maintaining" v3.

IamManchanda:
Yup won't fix is the not the right choice , even i agree .... @mdo @Starsam80 

Would be better to reopen those important issues or something like a temporary lock can also be a better choice for some of these issues

Starsam80:
Or maybe a new label to show that they will be reviewed later. Maybe something like `punted`?

IamManchanda:
Yup :+1: @Starsam80 .... but i guess `reviewed later` is also a good choice of the label..... rest on mark to decide

IamManchanda:
hmmmn @mdo seems on hurry ..... from 20 open to just 15 open for the milestone in the the last 15-20 mins ..... awesome .... so we can presume alpha6 in coming week i guess .... yuppieee

Studio384:
Why are these tags being removed? IE10 and IE11 still need those to make sure that they are using their respective latest rendering engine, no?

docs/getting-started/browsers-devices.md
```diff
@@ -105,70 +105,15 @@ Similarly, the latest versions of most desktop browsers are supported.
...
 
-Unofficially, Bootstrap should look and behave well enough in Chromium and Chrome for Linux, Firefox for Linux, and Internet Explorer 9 and below, though they are not officially supported.
+Unofficially, Bootstrap should look and behave well enough in Chromium and Chrome for Linux, Firefox for Linux, and Internet Explorer 9, though they are not officially supported.
 
 For a list of some of the browser bugs that Bootstrap has to grapple with, see our [Wall of browser bugs]({{ site.baseurl }}/browser-bugs/).
 
 ## Internet Explorer
 
-Internet Explorer 10+ is supported, however, IE9 down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported.
+Internet Explorer 10+ is supported; IE9 down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported in IE10, or require prefixed properties for full functionality. Visit [Can I use...](http://caniuse.com/) for details on browser support of CSS3 and HTML5 features.
```
Studio384: Not sure, but shouldn't it say "IE9 and down is not"?

patrickhlauke:
>  IE10 and IE11 still need those to make sure that they are using their respective latest rendering engine, no?

no. HTML5 doctype has the exact same effect, without needing vendor-specific meta elements

Herst:
Sorry for bumping this old PR but the following statement is not entirely correct:

> no. HTML5 doctype has the exact same effect, without needing vendor-specific meta elements

This only applies to pages which aren't in the "Intranet" zone, see https://www.leapinggorilla.com/Blog/Read/1016/ie-ate-my-css---disabling-compatability-mode

What is true though is that the other reason for setting the meta tag in the past, the UI option to switch a page to compatibility mode is hidden per default (at least in IE11).

PR #21487, opened 2016-12-31
v4: Remove tab-focus mixin

mdo:
Last remaining instance of this mixin was in Reboot (after the great button refactor of #21439 and more). Removing all that now and just letting `:focus` states do their thang.

Handles the last bits of #19708 I think.

/cc @patrickhlauke


patrickhlauke:
LGTM...will grab the latest changes and have a quick look-see to make sure nothing was missed

PR #21489, opened 2017-01-01
Move node-sass to npm script and drop Ruby Sass.

bardiharborow:
This PR moves node-sass to a NPM script.

Per #20332, I've also dropped Ruby Sass support:

> Compile Sass files with node-sass **(no more Ruby options)**


hnrch02:
Do we know if libsass finally achieved feature parity with sass? A comparison in output for the unminified file would be handy, I think. (Unless this has already been examined and I'm OOTL.)

bardiharborow:
Bootstrap [already uses libsass by default](https://v4-alpha.getbootstrap.com/getting-started/build-tools/#switching-sass-compilers), so this doesn't change that. The only [differences](http://pastebin.com/7kGeVhBX) in the output between libsass and Ruby Sass are to do with how whitespace is treated (I had to use `--ignore-blank-lines` to avoid diffing hundreds of added lines.) As for whether this will disrupt anyone, I posted a heads up on Slack and haven't heard any complaints.

mdo (approved):
Let's do it! Looks like tests are failing for some linting issues, otherwise looks good to me!

bardiharborow:
@hnrch02 did you have any further concerns or shall we LGTM?

Johann-S (approved):

hnrch02 (approved):
Seems like I was just out of the loop on ruby Sass 😄  LGTM!

PR #21491, opened 2017-01-01
Update more copyright years to 2017

cquanu:


hnrch02:
@mdo Do we need to update all of those?

mdo:
Strictly speaking, no, we never do, but folks see it as a sign of staleness if there's not an up to date copyright year.

PR #21496, opened 2017-01-01
Remove duplication in `flexbox.md`

Starsam80:
First commit removes a lot of duplication by using loops. In fact, I can reduce the duplication even more by using `{% cycle %}` but then it starts looking like code-golf, so I'm not going to do that.

The second commit just puts the `bd-highlight` class in front of the `p-2` class, so when it parses, it will hide the `p-2` class. A lot of this is pointless because it'll get replaced with `...` later.

The third commit merges the 2 functions we had before (`remove_holderjs` and `remove_example_classes`) and puts it into a new function called `cleanup_highlighted_code`. The TODO comment was when I realized [that regex does not parse HTML](http://stackoverflow.com/a/1732454) and should be replaced by something better.
The 2 new features that I've also added in this commit are:
- The `data-bd-hide` attribute, which hides the attributes that follow it.
  - `<div class="d-flex" data-bd-hide style="height: 100px;">` =>  `<div class="d-flex">`
- The `data-bd-remove` attribute, which replaces the entire element with `...`
  - `<div data-bd-remove class="d-flex">` => `...`

Please note that these are not going to work 100% of the time, because of the way HTML and regex work.

The fourth commit uses the stuff that I added in the previous commit. I plan to do this to every single file, but I first need to fix the TODO.

For people that want to see what this changes, I've made a build of the docs here: https://bootstrap-tests.starsam80.net/more-bd/utilities/flexbox/

### DIFF
--- a/docs/_plugins/highlight_alt.rb
+++ b/docs/_plugins/highlight_alt.rb
@@ -58,25 +58,28 @@ eos
         "<div class=\"bd-example\" data-example-id=\"#{@options[:id]}\">\n#{output}\n</div>"
       end
 
-      def remove_holderjs(code)
-        code = code.gsub(/data-src="holder.js.+?"/, 'src="..."')
-      end
+      def cleanup_highlighted_code(code)
+        # TODO: We should actually parse the HTML and remove things that way instead of using regex
 
-      def remove_example_classes(code)
-        # Find `bd-` classes and remove them from the highlighted code. Because of how this regex works, it will also
-        # remove classes that are after the `bd-` class. While this is a bug, I left it because it can be helpful too.
-        # To fix the bug, replace `(?=")` with `(?=("|\ ))`.
-        code = code.gsub(/(?!class=".)\ *?bd-.+?(?=")/, "")
+        # Find holder.js images and replace the `data-src` with `src="..."`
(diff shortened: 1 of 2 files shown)


mdo:
Hmm, I'm kinda torn on this. While I understand the looping here to reduce the repeated code bits, I kinda prefer the static examples. It's easier to copy-paste from the source code, ensures readability while viewing files here on GitHub, and is more approachable to beginners.

Starsam80:
@mdo I don't personally care about if you want to merge this or not, it's just that when I look at the markup files, I tend to scroll a lot just to get to the bottom of the document and this fixes that. It also makes the markup files more 1:1 with the actual docs that get produced (if you ignore the loops). I noticed files like `components/modal.md` are kinda split with the actual content that they are showing and the highlighted code. Using the modal as the example, the 2 buttons show up as `Close` and `Save changes` in the shown example but `Save changes` and `Close` in the highlighted code. After I fix the TODO, we could use a single `{% example %}` tag to fix these kind of problems.

TL;DR: I don't care, do whatever you want with this pull request.

mdo:
Going to punt on this. Sorry for keeping it open so long with no movement.

Starsam80:
No worries. I totally forgot about this PR myself.

PR #21539, opened 2017-01-04
Fix link (a element) buttons color when focused

### LINKED ISSUE #21625: <a> button :focus styles not overridden correctly.
On any desktop browser, do the following:

1. Go to http://v4-alpha.getbootstrap.com/components/popovers/#dismiss-on-next-click
2. Click on the red _Dismissible popover_ button to bring up the popover.
3. Finally, move your cursor off of the button.

The text of the button changes to blue; which in my opinion, looks and feels a little weird.

pvdlg:
This has been introduced in https://github.com/twbs/bootstrap/commit/a9bee8b6c8c31455dfba029434fad88b862c3846

Reboot set the focused color for `a` elements.
The button-variant mixin has to set the proper variant color in order for link button to display properly when focused.

Before:

After:

Fixes #21625.

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -12,7 +12,6 @@
   border-color: $border;
   @include box-shadow($btn-box-shadow);
 
-  // Hover and focus styles are shared
   @include hover {
     color: $color;
     background-color: $active-background;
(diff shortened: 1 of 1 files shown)


mdo:
Hmm, I think we should probably change Reboot to avoid causing these complications rather than try to patch it in our mixins "downstream." Perhaps we change the `hover-focus` mixin to just `hover` for those link styles there?

vanduynslagerp:
I imagine we still want to have the basic links with `color: $link-hover-color;` both when focused and hovered. It would be weird to have a basic link with the default link color (`$link-color`) when focused...

Maybe we can change `a` to `a:not(.btn)` at [_reboot.scss#L163](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L163). But it feels a bit weird to reference the `.btn` class in reboot.

PR #21390, opened 2016-12-21
v4: Flexbox navbar

mdo:
Following up on #21389, this PR is a WIP for adding the flexbox navbar. It started off as a late night experiment again with `display: table`—which was pretty awesome mind you—and got me thinking heavily about flex options. Given we're going all flex now, this approach gets even easier.

Still a ton to do, but wanted to open this sooner than later.

---

### Done
- Fixes #18875 and fixes #18875 with move to flexbox-based approach.
- Fixes #20937 by closing out all the below issues.
- Moves `.navbar-toggleable-*` from the `.collapse` to the `.navbar`, making it even easier to toggle content as needed.
- Fixes #21364 by updating all examples to use new navbars.
- Revamp navbar docs to push responsive implementation from the get-go instead of at the end. Also adds new examples.
- Fixes #21361 by changing how togglers works. Alignment has changed and the icon is now a separate element, `.navbar-toggler-icon`, within `.navbar-toggler`.
- Adds left and right absolutely positioned toggler variations.
- Fixes #20766 with ability to center navbar links (with example in docs)
- Fixes #18115: Renamed `.navbar-dark` to `.navbar-inverse` to match nomenclature of `.bg-inverse` and `.card-inverse`. `.navbar-light` is still the default class.

### Todo
- [x] Fixes #21012 Better toggler styling and alignment
- [x] Fixes #18053: Container adding extra padding for mobile sizes
- [x] Fixes #20116: `.navbar-text` is supported and documented properly.
- [x] Migration docs

PR #21540, opened 2017-01-04
Adds Sidebar Drawer Option On Modal

judetucker:
Per this comment https://github.com/twbs/bootstrap/pull/21390#issuecomment-268703161 regarding adding a drawer to the framework, I thought I'd offer up the solution that we have been using in prod on V4 for a while. 

This attaches the Drawer to the `.modal` rather than the `.collapse` to take advantage of the modal backdrop. It can be used as an alternative look to the regular modal dialog and as an option for those looking for an off-canvas nav. 

drawer

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -11,6 +11,10 @@
   .modal {
     overflow-x: hidden;
     overflow-y: auto;
+
+    &.drawer {
+      overflow-y: hidden;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


mdo:
Whoa, this sounds awesome.

Johann-S:
Great work but I think it's bit strange to not create a new component. Why build this feature on Modal component instead of creating a new one ?

mdo:
To me, it's either the modal, the navbar, or the collapse. I'd also take an independent plugin that simply toggles a class somewhere though 😆 (like a more basic "collapse").

judetucker:
Honestly, I tacked it onto Modal because it didn't add any new JS to the code and it provided the modal-backdrop, which helps draw the focus to the drawer. (andd because my JS isn't the strongest 🤐 )

giulianovelli:
IMHO this should its own component and not tied to other components (modal, navbar and so on). These drawers have become really commonplace and it's probably the biggest bootstrap omission right now. While the effect shown in this page it's cool it's also very different from how drawers usually behave in apps and websites (pushing the content while opening of just covering it without light effects)

LauLogisch:
Great idea!
I think the default page should slide to the other side so this can be also used on mobile devices. 
Dragging the page to the right making the navigation visible or with a button.

XhmikosR:
@mdo: I rebased this and cleaned it up, minus a couple of linting issues.

XhmikosR:
But it seems we have #27035 for the same functionality.

Johann-S:
The same question remain I don't think we should tweak our Modal to do a Drawer, we should build a new component which will be called Drawer

XhmikosR:
Well, if you can refactor the code to be shared, sure. Otherwise there will be a lot of duplication.

XhmikosR:
But anyway I feel like #27035 is a little better.

XhmikosR:
Closing since this will be a new component.

viktorsmari:
Can you reference this new component here @XhmikosR ?

XhmikosR:
There's nothing to reference yet.

PR #21541, opened 2017-01-04
Update nuspec to support jquery v1.9.1 - v3.x

supergibbs:
@DavidDeSloovere emailed about the latest NuGet package requiring jquery v3.0.0.1, this will update the nuspec to match the bower.json file and require jquery to be at least v1.9.1 and less than v4.0.0

Also updates 
- the version to plain 4.0.0 (it's set on publishing to match the package.json file)
- the `tags` to match `keywords` of package.json

### DIFF
--- a/nuget/bootstrap.nuspec
+++ b/nuget/bootstrap.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>bootstrap</id>
-    <version>4.0.0-alpha.3</version>
+    <version>4.0.0</version>
     <title>Bootstrap CSS</title>
     <authors>The Bootstrap Authors, Twitter Inc.</authors>
     <owners>bootstrap</owners>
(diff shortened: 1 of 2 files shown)


nuget/bootstrap.sass.nuspec
```diff
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>bootstrap.sass</id>
-    <version>4.0.0-alpha.3</version>
```
bardiharborow: Why did you remove the `-alpha.3` rather than updating it?

supergibbs: less churn, the nuget publishing script sets the version to match the project.json file so now there is one less thing to update for version bumps.

bardiharborow (approved):

nuget/bootstrap.nuspec
```diff
@@ -16,9 +16,9 @@
     <copyright>Copyright 2017</copyright>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <dependencies>
-      <dependency id="jQuery" version="3.0.0.1" />
+      <dependency id="jQuery" version="[1.9.1,3)" />
```
tillkruss: Is this correct? Square opening bracket and round closing parenthesis?

fchiumeo: in [nuget packages](https://docs.nuget.org/ndocs/create-packages/dependency-versions#version-ranges) yes

supergibbs: Yes, but I did mess up values. Just updated to `[1.9.1,4)` which means greater or equal to v1.9.1 but less than v4

supergibbs:
Squashed 3 commits into one final one

PR #21545, opened 2017-01-05
Fixed property order, and same line issues

matrixfox:
Suppressed scss lint minor issues inside of the mixins folder.


scss/mixins/_hover.scss
```diff
@@ -8,16 +8,15 @@
 //    }
 //  }
 //  @else {
-    &:hover { @content }
+  &:hover { @content }
```
wolfy1339: I disagree with this change die to the comment in the code.

matrixfox: Line should be indented 2 spaces, but was indented 3 spaces at line 11 col 1.

mdo: This change should be removed; it's purposefully aligned with the commented out parts above and below it.

mdo (changes requested):

PR #21558, opened 2017-01-06
Fix media object alignment examples.

### LINKED ISSUE #21556: Media object examples do not show alignments
Under the Alignment section, the media object examples are missing the flex markup to display the different alignments.

bardiharborow:
Fixes #21556.


bardiharborow:
I'm going to go ahead and make a judgement call that merging this so late in the game and avoiding the dozens of complaints we will inevitably get if we ship this broken, is worth the added risk. Fingers crossed this doesn't blow up in my face. Or maybe I'm just being paranoid.

PR #21569, opened 2017-01-06
Fix IE11 table layout issue with flexbox

willyboy:
When applying flexbox to a table in IE, the table cells have strange widths. This is because IE11 sets TD to display: table-cell even when the parent has display: flex. The solution is to manual set the TDs to display: block. There are also wrapping issues that mean table and tbody also need display: block. That bug limits our ability to be more specific by saying "tr.row > td".

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -27,7 +27,7 @@ html {
 .container {
   table,
   tbody,
-  td, 
+  td,
   th {
     display: block;
   }
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -22,6 +22,16 @@ html {
   box-sizing: inherit;
 }
 
+//http://stackoverflow.com/questions/35451371/flex-css-property-in-table-doesnt-work-on-ie-and-firefox
+//Not sure what the best way to do this is since someone may want a table outside of the flex grid
+.container {
+  table,
+  tbody,
+  td, 
+  th {
+    display: block;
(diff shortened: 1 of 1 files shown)


mdo:
Why would we need to change the `display` on tables to `flex`?

willyboy:
It doesn't change the display on tables to flex. It changes the default for tables inside a container to display:block. It's not a great solution. Maybe a better one would be to have a "flex-table" class. But here's the stack overflow that talks about the issue: http://stackoverflow.com/questions/35451371/flex-css-property-in-table-doesnt-work-on-ie-and-firefox

You can see the issue by viewing this pen in IE11 and edge: http://codepen.io/willyboy/pen/apOEGw

Notice that the columns don't expand to fit the row until you uncomment "display:block"

mdo:
I don't think this makes sense to add to Bootstrap. The generic application of changing the `display` for table elements is bound to cause issues. Plus, this isn't really needed in the core at all—no tables have `display: flex` on them right now. If you were to start changing the display on tables, you will need to account for the rows and cells.

willyboy:
That contradicts the code in bootstrap v3 (see below) and also means that the bootstrap flex grid system is broken for all tables in IE (when it works for tables in v3). I can change the implementation to be something similar to what you have in v3 but to say, "It's okay for our grid to not work in even the latest version of IE" seems flat out wrong for such a widely used framework.

table td[class*=col-], table th[class*=col-] {
    position: static;
    display: table-cell;
    float: none;
}

PR #21578, opened 2017-01-06
Fix backdrop for dropdown menu on mobile

### LINKED ISSUE #21572: Dropdowns don't close when tapping outside the menu on mobile
iOS 10.2 Safari & Chrome

Tried the dropdowns page in hosted docs. https://v4-alpha.getbootstrap.com/components/dropdowns/

pvdlg:
Fixes #21572

Fix the backdrop for dropdown menu on mobile:
- Create backdrop only if the menu is actually open (do not create it if
the show event is prevented)
- Drop the backdrop only when the corresponding menu is closed (do not
remove if there is no menu to close or if the hide event is prevented)
- Drop only the backdrop associated with the dropdown that is being closed


mdo:
Can you apply these changes to the `js/src/` file instead?

vanduynslagerp:
Ooops my bad. Done !

PR #21589, opened 2017-01-07
JS module transition: allow to customize transition duration in sass and custom css

pvdlg:
The JS modules using transitions (modal, alert, popover, collapse, carousel) currently emulate the transitionEnd event with a duration defined as a static value in each module.

If the transitions duration set in css are different than the statics values in each modules (`TRANSITION_DURATION`) it creates a visual glitch.
The `.show` class is set after the duration set in each module, disregarding the value set in the css `transition` property.
For example the collapsing module define a static duration of 600ms. If we set in scss `$transition-collapse:    height 1s ease`:
- The 1s transition will be initiated and the collapsible area start expanding
- After 600ms the function `transitionEndEmulator` trigger the transitionEnd event (even though the transition is not complete yet
- The `collapsing` class will be removed from the collapsible area and it will suddenly appear expanded fully without the transition having completed

Also if we set in scss `$transition-collapse:    none`, a 600ms transition will still be emulated and the `.show` class will be set after 600ms while it should be set right away.

This PR add the function `transitionEmulator` to emulate emulate the transition with the duration set in css and to handle the case of no transition or a 0 duration.

As a result:
- we can change the transition duration of modal, alert, popover, collapse and carousel components with sass variables
- we can define specific custom classes with different transition duration, i.e. we can have a collapsible area that expand in 350ms and another one in 700ms
- we can set the transition to none with sass variable (#18156 can be solve by setting `$transition-collapse:    none`)
- we don't have a risk to have discrepancy between the js module static duration value and the sass variable that would cause a visual glitch and the static duration in each module are now useless

In addition the function `transitionEmulator` handle more common code that were previously repeated in each module:
- determine if there is a transition to apply (previously the presence of a class was checked in each module which was failing to determine a transition set by a custom class)
- Order properly the transition start and the completion callback depending on the presence of a transition or not (i.e. if transition: setTimeout to call complete callback then start the transition / if no transition: start the class change and then call the complete callback without timeout)
- As a result it simply the code related to transition in each module and would make it easier to avoid mistakes in future modifications in each modules

Finally the PR updates the unit to test each module behavior with and without transitions


twbs-rorschach:
Hi @vanduynslagerp!

Thanks for wanting to contribute to Bootstrap by sending this pull request!
Unfortunately, your pull request seems to have some problems:
* [Changes must be made to the original Sass or Less source code file(s), not just the compiled CSS file(s).](https://github.com/twbs/rorschach/blob/master/docs/css.md)

You'll need to **fix these mistakes** and revise your pull request before we can proceed further.
Once you've fixed these problems, you can either ask the maintainers to re-open this pull request, or you can create a new pull request.
Thanks!

(*Please note that this is a [fully automated](https://github.com/twbs/rorschach) comment.*)

bardiharborow:
@vanduynslagerp sorry about the closer bot, but you'll need to not check in changes to the `dist/` and `docs/dist/` folders.

vanduynslagerp:
Thanks for re-opening @bardiharborow.
It's done.

js/src/alert.js
```diff
@@ -107,20 +106,6 @@ const Alert = (($) => {
       return closeEvent
     }
 
-    _removeElement(element) {
```
Johann-S: Why did you remove this method ?

js/src/alert.js
```diff
@@ -107,20 +106,6 @@ const Alert = (($) => {
...
-          !$(element).hasClass(ClassName.FADE)) {
-        this._destroyElement(element)
-        return
-      }
-
-      $(element)
-        .one(Util.TRANSITION_END, (event) => this._destroyElement(element, event))
-        .emulateTransitionEnd(TRANSITION_DURATION)
-    }
-
```
vanduynslagerp: @Johann-S With the simplification due to`transition`, `_removeElement(element)` became a one liner and as it's called only once it seemed better to remove it and the one liner directly

Johann-S: Yes I understand after the post of my comment 👍

bardiharborow:
Given that [`transitionend`](http://caniuse.com/#search=transitionend) is now supported by all our supported browsers, would we be able to drop this whole issue and just make the JS listen for `transitionend` directly?

vanduynslagerp:
This is what I was thinking to do in the first place, but during my tests I realized that the when you start a transition you are not necessarily guaranteed to receive a `transitionend`.

One of these cases is when you interrupt a transition (start a show animation and start the hide animation before the show has completed).
I also found several other cases especially with tooltips and collapse when they are triggered several time quite fast.

If we don't receive the event that leave the component in inconsistent state (either visually or in a way that prevent the component to work again).

The only way I found to guarantee the end event will be received and processed is to keep the `setTimeout` that was originally there.

vanduynslagerp:
@bardiharborow, actually, I can completely drop the `transitionEnd` emulation as it's not required anymore. I'll update the PR tonight.

mdo:
This PR got a little messed up it looks like. Wanna clean it up or start from scratch @vanduynslagerp?

mdo:
Closing as out of date.

Issue #21587, opened 2017-01-07
[v4 css] form-control-sm styling


Mottie:
When comparing form elements with a "form-control" versus a "form-control-sm" class in Bootstrap v4.0.0-alpha.6, there are some notable differences &ndash; I'm using Windows 10 and Chrome version 55.0.2883.87 m (64-bit), in case that matters.

[demo](http://codepen.io/Mottie/pen/YNXxeL) (updated!)

![2017-01-06 19_38_24-bootstrap v4 0 0-alpha 6 issue](https://cloud.githubusercontent.com/assets/136959/21738259/d2b94ff2-d447-11e6-8390-ab2bd13bd887.png)

* <del>`select.form-control-sm` &ndash; border too dark</del>
* `input.form-control-sm`&ndash; mismatched height <del>& inset border</del>

-----

Update: sorry, I didn't realize that both classes needed to be applied.... the `input` height is still slightly off.

sergiopantoja:
Same on macOS Sierra and Chrome version 55.0.2883.95 (64-bit).

![image](https://cloud.githubusercontent.com/assets/2530982/21737938/5eca2a30-d44c-11e6-952d-b8323d21adb6.png)

sergiopantoja:
@Mottie 

So it looks like `.form-control-sm` needs to be used with `.form-control`. For example:
```html
<input class="form-control form-control-sm" type="text">
```
```html
<select class="form-control form-control-sm">
  <option>Small select</option>
</select>
```

Can you try this and see if you still encounter the same issue?

Mottie:
Ahh, you're right!... but the input height is still not matching perfectly... here is a zoomed in screenshot:

![2017-01-06 19_38_24-bootstrap v4 0 0-alpha 6 issue](https://cloud.githubusercontent.com/assets/136959/21738259/d2b94ff2-d447-11e6-8390-ab2bd13bd887.png)

mdo:
Fixed by #22287.

PR #20874, opened 2016-10-10
Match the height of select.form-control with input.form-control

mdo:
Uses a local variable and some calc love to counteract the border-width (times 2) from the height of the select. Fixes #17194 and nullifies #19967.

Demo of the new CSS in effect with inputs, selects, and custom selects for good measure: https://jsbin.com/yitijo/.

PR #21592, opened 2017-01-07
Match height of select.form-control-sm, input.form-control-sm, and .btn-sm

sergiopantoja:
This commit shares line-height across these form elements, and adds the same height calculation to `select.form-control-sm` that we have for `select.form-control`.

For issue #21587 

Relevant commits:

https://github.com/twbs/bootstrap/commit/8e455ff56105e210cd313786c23532b7b81d01c9
https://github.com/twbs/bootstrap/pull/20874/commits/0c467e7b29e2dd601e51b2af161c0546a79d432c
https://github.com/twbs/bootstrap/commit/6993db1840561e89dd772302d5b3d567bc92f033

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -137,11 +137,10 @@ fieldset[disabled] a.btn {
 //
 
 .btn-lg {
-  // line-height: ensure even-numbered height of button next to large input
   @include button-size($btn-padding-y-lg, $btn-padding-x-lg, $font-size-lg, $btn-border-radius-lg);
 }
 .btn-sm {
-  // line-height: ensure proper height of button next to small input
+  line-height: $line-height-sm; // ensure proper height of button next to small input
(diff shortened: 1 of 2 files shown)


sergiopantoja:
Before (current HEAD of v4-dev): https://output.jsbin.com/faqigodumu

After: https://output.jsbin.com/yowuzuvuze

Feel free to edit as needed or let me know if anything should be changed!

NicoLB:
I feel like this is a constant source of issues and would like to discuss it further with you guys.

It seems the main issue is all `.btn-{size}` share a similar `$btn-line-height` where `.form-control-{size}` all have a different one.

On the other hand we already have a `$input-btn-border-width`, so why not have `$input-btn-line-height` shared by `.btn` and `.form-control`?

We could even push a bit further and replace `$btn-padding-y-{size}` and `$input-padding-y-{size}` variables by `$input-btn-padding-y-{size}` making it impossible not to have matching heights.

What do you guys think about this solution?

mdo:
Superseded by #22287 after diving into your latest comment here. Thanks for getting this started!

Issue #21595, opened 2017-01-07
$font-family-sans-serif and $font-family-base are ignored?


gthomas-appfolio:
When customizing alpha.6 via _custom.scss, if I set values for `$font-family-sans-serif` and `$font-family-base`:

    $font-family-base: sans-serif !default;
    $font-family-sans-serif: sans-serif !default;

this is set for `html`, but seems to be ignored for `body`, `.popover`, and `.tooltip`, in the built css.  Instead the font-family for these is the default `$font-family-sans-serif` from _variables.scss:

    font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;

This used to work in previous alphas.  Anything I'm missing above?

Jas-n:
Same here, looks like it's related to the moving of the @import "custom"; line in scss/bootstrap.scss (Commit: https://github.com/twbs/bootstrap/commit/44825032370e37f7cec410356e4a1b1cd9999307).

I tried changing it back and the custom variables work again.

gthomas-appfolio:
Ah thanks for noticing, I'll temporarily use this as workaround

mdo:
Closing since #21603 was merged.

Issue #21583, opened 2017-01-06
Alpha 6 ignoring _custom.scss


bluesurfer:
I am trying to customize Bootstrap by editing `_custom.scss` like this:
```
$blue: red !default
```
But, when compiling with `grunt dist-css`, the new version seem to ignore `_custom.scss` - nothing changes. On the other hand editing directly `_variables.scss` works. Am I missing something new or is this an actual bug?

maimairel:
This is caused by #21442 I also have the same issue

mdo:
Drop the `!default`. https://jsbin.com/ficofo/edit?html,css,output

maimairel:
@mdo Yes, I don't have the `!default` on the custom values. What I don't understand is while `$grid-breakpoints` and `$container-max-widths` do work the other custom values are ignored.

szastupov:
I can confirm the issue, here is my _custom.scss:

```
$brand-primary:             #f0941c;
$brand-inverse:             #0f6be3;
$brand-danger:              #e15325;
$body-color:                #505455;
```

and all those variables are ignored.

The reasons is that values from _variables.scss are used in mixins *before* including _custom.scss.

Here is your modified example to demo the behaviour: https://jsbin.com/qutoniseci/edit?html,css,output

PR #21603, opened 2017-01-08
_custom.scss should be the first include

desheikh:
Custom include should be at the top of the file. This prevents issues where for example reassigning $brand-primary would not affect any of the other variables using $brand-primary.

### DIFF
--- a/scss/bootstrap.scss
+++ b/scss/bootstrap.scss
@@ -6,9 +6,9 @@
  */
 
 // Core variables and mixins
+@import "custom";
 @import "variables";
 @import "mixins";
-@import "custom";
 
 // Reset and dependencies
 @import "normalize";


bardiharborow:
This reverts #21442. Ref #21595. @mdo any idea what's going on here?

szastupov:
Ref #21583

mdo:
@bardiharborow The problem I was trying to solve before is that folks cannot use existing variables and mixins in the `_custom.scss` file if it comes first in the import order. My hope was moving it to come after the variables and mixins would solve things, but now folks can't update the value of variables that are reassigned in our `_variables.scss` file (e.g., `$link-color` using `$brand-primary`).

So, we're kinda screwed either way. I don't think we can have both in Sass. I think I'd rather change it back though.

Thoughts?

vanduynslagerp:
The solution can be to import `_custom.scss` both before and after `_variables.scss`
Not sure if it's very "clean" but it works

judetucker:
+1 for change it back. I think it's more common for people to use it for overriding the bootstrap default vars

Starsam80:
This PR doesn't change the files `bootstrap-grid.scss` and `bootstrap-reboot.scss`

PR #21606, opened 2017-01-08
fix issue #5349 : Chrome/Firefox Zoom creates a gap of 1px between to…

hjondi:
In Chrome and Firefox if you zoom in, in certain zoom levels there would be a gap of 1px between the box and the arrow of the tooltip, this might also apply to popovers.

zooming is pretty common on high resolution screens. 

This will fix the issue #5349

### DIFF
--- a/scss/_tooltip.scss
+++ b/scss/_tooltip.scss
@@ -19,7 +19,7 @@
     margin-top: -$tooltip-margin;
 
     .tooltip-inner::before {
-      bottom: 0;
+      bottom: 1px;
       left: 50%;
       margin-left: -$tooltip-arrow-width;
       content: "";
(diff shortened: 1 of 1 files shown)


feinstein:
I just want to link here my [stackoverflow question regarding this](http://stackoverflow.com/questions/42151106/how-to-prevent-bootstrap-tootip-from-falling-apart) for anyone looking for a fix still on Bootstrap 3.3.7

mdo:
Punting on this given the referenced PR is several years old and we don't account for zooming bugs really anywhere else.

hjondi:
@mdo but this issue is being reported by many people why not just fix it ?
http://stackoverflow.com/questions/42151106/how-to-prevent-bootstrap-tootip-from-falling-apart

Issue #21607, opened 2017-01-08
v4.0.0-alpha.6 - Activating tooltips breaks modals (stuck "transitioning")


METACEO:
Operating systems and browsers:
- Windows; Firefox 50.1.0, Internet Explorer 11, Tor Browser 6.0.8 (Firefox 45.6.0)
- Android 5.0.1 (Galaxy S4); Firefox 40.0.3, Chrome 45.0.2454.84, Stock Browser 2.1.34.19

Launch [the demo modal with tooltips](https://v4-alpha.getbootstrap.com/components/modal/#tooltips-and-popovers) in the component's documentation, activate a tooltip, then try and exit the modal - nothing expected appears to happen. [I have reproduced this outside the documentation](https://jsfiddle.net/6yy600a8/) ~~while I have not yet tried to reproduce this outside the documentation~~, across all of my devices, this seems to be a show-stopping bug - freezing the user's interface and preventing any other interaction.

I'm logging two different errors in my console:
- closing the modal logs `Error: Modal is transitioning`
- activating a tooltip logs `Error: Tooltip is transitioning`.

METACEO:
I just [reproduced this in](https://jsfiddle.net/6yy600a8/) JSFiddle.

JD-Robbs:
Similar for me. Perhaps related: #21600

Paladinium:
Temporary workaround is to remove the 'fade' class on modals.

Johann-S:
It's due to those lines : https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L473-L476

Because we check if a tooltip is wrapped into a modal but not if the tooltip is visible or not and called ``hide`` twice instead of one call

ghiscoding:
Nice to know that I'm not the only one seeing this problem. However in my case I have a page with multiple elements and Tooltips (no modal in play) and after doing a few mouse hovering, I start seeing the 

`Error: Tooltip is transitioning`. 

@Johann-S will your code change also fix when using Tooltip only (not in Modal)?

Johann-S:
@ghiscoding not sure of that but if you can produce a reduce test case I can check that

ghiscoding:
[...]

PR #21613, opened 2017-01-08
Fix hide tooltip into a modal

Johann-S:
Add a check to not call hide if a tooltip is already in a transitionning mode

Do you think I have to add unit test for this ?

/CC : @bardiharborow 

Close : #21607

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -472,7 +472,11 @@ const Tooltip = (($) => {
 
         $(this.element).closest('.modal').on(
           'hide.bs.modal',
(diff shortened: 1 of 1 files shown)


bardiharborow:
Ref #21227. I wouldn't bother with test cases for now. I have to look at rewriting part of the tooltip plugin first to avoid these issues altogether.

Johann-S:
@bardiharborow IMO we should merge this one before the release of the first beta and let you rework the tooltip plugin later if you are ok

Johann-S:
Close due to : #21743

Issue #21328, opened 2016-12-12
Dropdown: jQuery throws with empty ID selector


gpakosz:
Hello,

While following the documentation, I've stumbled upon Dropdown examples: http://getbootstrap.com/javascript/#via-data-attributes-1

```
<div class="dropdown">
  <a id="dLabel" data-target="#" href="http://example.com" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
    Dropdown trigger
    <span class="caret"></span>
  </a>

  <ul class="dropdown-menu" aria-labelledby="dLabel">
    ...
  </ul>
</div>
```

When trying it with jQuery `v2.1.3`, everything works fine and [clicking the anchor displays the dropdown](http://codepen.io/anon/pen/vyzRez).

However, when using jQuery `v3.1.0` [clicking the anchor follows the link](http://codepen.io/anon/pen/mOGzLB).

bardiharborow:
[Support for v3 has ceased](https://github.com/twbs/bootstrap/issues/20631). Please let us know if your issue affects v4.

gpakosz:
Hmm with `v4` it seems markup is totally different anyways

mdo:
If this isn't affecting v4, we'll close it out as a won't fix.

gpakosz:
This is what I'm experiencing with `v4`:

- when using a link/anchor as a toggler with `data-target="#"` as per instructed in `v3` documentation: http://codepen.io/anon/pen/BQqBzw

- when using a link/anchor as a toggler **without** `data-target="#"`: http://codepen.io/anon/pen/oYavBb

I appreciate that `v4` dropdowns documentation doesn't mention `data-target="#"` while `v3` documentation used to. However, I would say people migrating from `v3` to `v4` will face that situation as the migration guide doesn't instruct to remove `data-target="#"` in that case.

In that respect, can you please enlighten me about:

- what's the current or past role of `data-target="#"` exactly?
- what are the tradeoffs between using `<button>` vs `<a>` for dropdown togglers?

Johann-S:
As it said in V3 documentation : 
> To keep URLs intact with link buttons, use the data-target attribute instead of href="#".

And the difference between ``<button>`` and ``<a>`` depends on your needs, if you want to toggle the dropdown from a classic link or from a button

gpakosz:
Should one understand that "keeping the URLs intact" mean clicking the link toggler won't toggle the dropdown?

[...]

Issue #18641, opened 2015-12-23
Regex in Util.getSelectorFromElement is too restrictive


ditiem:
In this line (v4-alpha code)

tab.js: 90:       let selector  = Util.getSelectorFromElement(this._element)

decided which pane will be selected when clicking on the "tab". Inside the getSelectorFromElement, if not data-target is provided, the href attribute is inspected. In order to decide if the selector is valid, a regular expression is used.

I generated a "fake uuid" in my code and found out:

/^#[a-z]/i.test("#8828a412-2185-4d9c-9073-81272784c4a1")
returns false

/^#[a-z]/i.test("#e58329c1-e759-4661-9ae9-c33a12b8c0b6")
returns true

(the reason is that in the first selector the first character is a number rather than a letter.)

So sometimes the tab was selected and sometimes not!

An expression like /^#[a-z1-9]/i should work, but I suspect there will be more cases to take into consideration.

ditiem:
Checked the w3 spec (I was suspecting about the _ character). See the source in http://www.w3.org/TR/CSS21/grammar.html

nmchar		[_a-z0-9-]|{nonascii}|{escape}
name		{nmchar}+
"#"{name}		{return HASH;}

So I think at least the expression mentioned in the previous comment should be:
/^#[_a-z0-9-]/i

Notice (I realized after I hit "Send comment") that we are only checking for the first character. I do not know if this was intended.

ktmud:
There are a lot of weird cases for these IDs, and it is promised all will be fixed in v4?
https://github.com/twbs/bootstrap/pull/14789#issuecomment-60138926

vanillajonathan:
I've also encountered this issue with the regular expression being too restrictive. It could not handle my numerical id, i.e. `#123`.

joezcool02:
PR for this issue [Here](https://github.com/twbs/bootstrap/pull/22128)

mdo:
Fixed by #21615.

PR #21615, opened 2017-01-08
Fix getSelectorFromElement when # is a selector

Johann-S:
Check if ``data-target`` contain ``#`` to avoid JQuery error

Close : #21328

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -112,13 +112,16 @@ const Util = (($) => {
 
     getSelectorFromElement(element) {
       let selector = element.getAttribute('data-target')
-
       if (!selector || selector === '#') {
         selector = element.getAttribute('href') || ''
-        selector = /^#[a-z]/i.test(selector) ? selector : null
       }
 
-      return selector
+      try {
+        const $selector = $(selector)
+        return $selector.length > 0 ? selector : null
+      } catch (error) {
+        return null
+      }
     },
 
     reflow(element) {
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -113,7 +113,7 @@ const Util = (($) => {
     getSelectorFromElement(element) {
       let selector = element.getAttribute('data-target')
 
-      if (!selector) {
+      if (!selector || selector === '#') {
         selector = element.getAttribute('href') || ''
         selector = /^#[a-z]/i.test(selector) ? selector : null
       }


vanduynslagerp:
See this PR #21654 that handle the problem differently and cover more (all) possible cases of selector.

Johann-S:
Yes I forget one case but I'll update my PR thank you @vanduynslagerp

vanduynslagerp:
What I meant is that #21654 solve #21328 and other issues, plus it brings other benefits described in the PR. If #21654 is merged this one (#21615) becomes irrelevant as the modified code would been removed.

Johann-S:
I didn't change how the core of Bootstrap works but I added what you did in your PR by removing the regex because as you said it fix #18641 too 👍

vanduynslagerp:
Really sorry to insist...but here is a few remarks.

Only the dropdown and the alert component expect to `data-target="#"` to be valid. This is because the modules handle this cases by selecting the parent of the trigger in this case. See [dropdown.js#L212](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/dropdown.js#L212) and [alert.js#L97](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/alert.js#L97)
So as you modification is on Util.js you impact all the other modules using `getSelectorFromElement `.

Also consider other possible wrong value such `data-target="test"`. The previous code would return `null` because "test" do not match the regex. With you modification it would set `selector` to "test" and then execute `$('test')` not throw an exception and then return `selector` with the value "test".

Finally you call Jquery on the value `selector` and then return `selector` to call JQuery again in each module again on the same `selector`. That's not really efficient as you have JQuery parsing the dom twice.

For all these reasons and several other I decided to rewrite the function `getSelectorFromElement ` to return the JQuery with the list of matching elements and use that result in each module.
That also allow me to handle different cases differently in each module (for example in dropdown when neither the value of `data-target` nor `href` is a selector that returns elements I can determine the target is the parent. In in collapse, in the same situation I just skip the action as it's impossible to determine the target).

The point I was trying to make is that the best way to handle this is to get rid of the regex and use the try catch. This way we can convert all cases and we don't have to add more `if else`s or a more complicated regex every time we have an issue reported of type "`data-target="whatever"` doesn't work and should do this or that".
And doing also require the other modifications I made in #21654.

The point of my PR was in the first place to fix #18641 and improve the mechanism by making more robust. I realize later on after posting the PR, that it would also fixe #21328 so I mentioned it here.

vanduynslagerp:
I also added a unit test in #21654 for the issue described in #21328

Johann-S:
Yes you're right on this : 
> Also consider other possible wrong value such data-target="test". The previous code would return null because "test" do not match the regex. With you modification it would set selector to "test" and then execute $('test') not throw an exception and then return selector with the value "test".

So I'll update a bit my PR thank you 👍 

In your PR the same problem is present because if the selector is set to ``test`` you return
``$('test')`` and after in each modules you check : 

```javascript
const target = targets ? targets[0] : null
```
But if targets is equal to ``$('test')`` it will return ``targets[0]`` which is undefined

vanduynslagerp:
> So I'll update a bit my PR thank you 👍

My point was that after I made my PR (without seeing yours before, sorry...) I realized that it also fixes the same problem and covers more ground. There is no point in having both #21654 and this one. Anyway I don't want to insist more.

> But if targets is equal to $('test') it will return targets[0] which is undefined

Yes that's on purpose. If there is no matching element that can be used as target then the variable `target` will be `undefined` or `null` and handled according by each module.
If you take the dropdown module as an example it's handled here: [dropdown R206](https://github.com/twbs/bootstrap/pull/21654/files#diff-bfe9dc603f82b0c51ba7430c1fe4c558R206). If target is `null`, `undefined` or an empty JQuery then we use `element.parentNode`. This what fix the case explained in #21328.
In collapse for example we need to have a defined target and we cannot figure it out by ourselves (like using `element.parentNode` in dropdown as the collapse markup is different). So if `target` is `null` or `undefined` [here](`element.parentNode`) then we skip the action as we don't have anything to collapse

Johann-S:
I add on my last commit, a check if the selector exist in the DOM or not and with that we don't have to edit all plugins because it works as you described.

Your dropdown unit test is very usefull 👍

vanduynslagerp:
The more modifications you make, the more your PR is looking like mine.
I spent a day working on #21654 and going through all this cases, details and test. So once you will go through all of them, I imagine the code will be similar to mine.
So I'm not really what we are doing here....
I just wanted to make the point, in my first comment, that I believe #21654 solves #21328 and other reported and non-reported issues in a more thorough way.

Johann-S:
I finished my PR I wont add anything. My PR only avoid the issue when ``#`` is used in selector that's all⁣​

mdo:
Lemme know if we still need to do #22128 as well.

Johann-S:
#22128 is useless due to the merge of this PR 👍 thank you @mdo

PR #21622, opened 2017-01-09
Hard padding for dropdown-item in dropdown.scss make it impossible to custom by scss var

Lausselloic:
Create a new var in _variables.scss named $dropdown-item-padding-y keeping the fixed value 3px
Replace the hard coded value in _dropdown.scss by this new var to allow custom values


scss/_variables.scss
```diff
@@ -579,6 +579,7 @@ $dropdown-link-active-bg:        $component-active-bg !default;
 $dropdown-link-disabled-color:   $gray-light !default;
 
 $dropdown-item-padding-x:        1.5rem !default;
+$dropdown-item-padding-x:        3px !default;
```
vsn4ik: `...-y`

Lausselloic:
Looking forward 

.dropdown-menu use a $dropdown-padding-y to ensure security padding inside the menu, but the same security value is also used for .dropdown-header, which mix dropdown-padding-y and dropdown-item-padding-x

May following this PR, can we update the value use for dropdown-header.

So from my point of view .dropdown-menu need to use dropdown-padding-y value
.dropdown-item and .dropdown-header will use padding: $dropdown-item-padding-y $dropdown-item-padding-x;

That way outer size will be more consistant for items inside the dropdown menu

PR #21626, opened 2017-01-09
Add align self to navbar brand

### LINKED ISSUE #21621: .navbar-brand is too wide
I have noticed that .navbar-brand is too wide. When I click on the "empty" space, I'm still able to activate link in navbar brand. It's not a problem when I have only logo on my navbar, but when I want to have an hamburger icon it's difficult to click on it. You can simply fix it by adding "align-self: flex-start;" to .navbar-brand
navbar

kvlsrg:
This PR fixes navbar brand width. Problem is that after navbar collapsed, link become full width:

width

hover

So, after adding `align-self: flex-start;` to `.navbar-brand` it will have its normal width:

after

Fixes #21621.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -29,6 +29,7 @@
 
 .navbar-brand {
   display: inline-block;
+  align-self: flex-start;
   padding-top: .25rem;
   padding-bottom: .25rem;
   margin-right: $navbar-padding-x;


PascalPixel:
THIS.

mdo:
I imagine we'll have issues when folks collapse a navbar *without* a `navbar-brand`, but for those issues the utility class should suffice. Thanks!

zalog:
#22230 fixes this.
before: `flex-direction: column` - http://jsbin.com/poneza/edit?css,output
after: flex-direction default / row  http://jsbin.com/pezeneh/edit?css,output

Issue #21627, opened 2017-01-09
Changing visual order of flex items issue


Cooleronie:
Utility in question: http://v4-alpha.getbootstrap.com/utilities/flexbox/#order

First of all, the example provided for **Order** does not include the utility `order` at all. Rather, it uses the `flex-order` classes [explained here.](http://v4-alpha.getbootstrap.com/layout/grid/#flex-order)

When I try to use the `order` utility it doesn't work, whereas `flex-first` and `flex-last` work as intended.

**So this works:**

```
<div class="d-flex">
     <div>2</div>
     <div class="flex-first">1</div>
</div>
```

**This does not:**

```
<div class="d-flex">
     <div>2</div>
     <div class="order-first">1</div>
</div>
```

Yet the documentation mentions `order-first` as if you can use it the same way one can use `flex-first`.  

The reason why this is an issue is because of the **responsive variations**. In Alpha 5 you had the option to use classes such as `flex-xs-first` and `flex-md-last`, but they seem to be deprecated. 

So, knowing I am not supposed to ask questions here, how would one go about replacing the `flex-{breakpoint}-{order}` in Alpha 6? Or is the `order` utility bugged and will get a fix soon?

(Not looking for alternative custom solutions, as I expect this feature to be shipped in Bootstrap itself)

Thanks!

PR #21628, opened 2017-01-09
Use correct classes for `order` flexbox docs.

Starsam80:
Saw this when I looked at #21627. `.order-*` classes don't exist, so we shouldn't mention them.


mdo:
Ah, these should start with `.order-` instead of `.flex` given it should be `.property-value`. New PR opened to try to tackle that, though I'm unsure on the naming scheme for them (which is likely why I had `flex` in here to start.

Issue #21624, opened 2017-01-09
Placeholder is visually indistinguishable from normal text


vanillajonathan:
The placeholder is almost visually indistinguishable from normal text input.
This is confusing and dangerous. The placeholder text ought to be brighter/dimmer.

This was fine in v4.0.0-alpha.5 but is a regression in v4.0.0-alpha.6.
The color is defined by the `$input-color-placeholder` variable in `/scss/_variables.scss`.
From alpha5 to alpha5 it was changed from `#999` to `$gray-light`. 

## Screenshot
![image](https://cloud.githubusercontent.com/assets/10222521/21770448/a0996d3e-d683-11e6-9654-52ac65b7ff65.png)

mdo:
Closing per https://github.com/twbs/bootstrap/pull/21782#issuecomment-274284975.

vanillajonathan:
So is any fix planned for this issue?

Until then, I have the following code in a CSS file that I load after Bootstrap.

```css
/* Temporary code to workaround a regression in Bootstap 4.0alpha6,
    remove this code when Bootstrap is updated with a new release. */
.form-control::-webkit-input-placeholder {
  color: #999;
}

.form-control::-moz-placeholder {
  color: #999;
}

.form-control:-ms-input-placeholder {
  color: #999;
}

.form-control::placeholder {
  color: #999;
}
```

jonathanhefner:
I agree the appearance is too indistinct.  What about defaulting to italics to distinguish placeholder text?

vanillajonathan:
@jonathanhefner, italics is unconventional. The de facto standard is dimmer text. So best would be what is common and expected to avoid confusion.

patrickhlauke:
Placeholder text is one of those things that can cause different accessibility problems for different user groups, and trying to fix it for one group will lead to problems for other groups.

Color contrast is of course one of these: you'd want the contrast to be strong enough to be legible, but also you'd want the difference between placeholder text and actual user-entered value text to be different enough to make it clearly distinguishable.

Placeholder text that appears too "prominent" (e.g. using a dark-enough text colour) can easily be mistaken for actually entered text, which would confuse certain users.

And of course, placeholder text should never be used as the only way to provide a "label" or additional hints to users (specifically because after user entry, it disappears).

[...]

PR #21630, opened 2017-01-10
Redefine grayscale variables

sergiopantoja:
Redefine grayscale variables

This would fix issue #21624 (I reverted `$input-color-placeholder` to `#999` via the new `$gray` value).

This would also solve the `#ddd` issue discussed in #20677.

#### ORIGINAL

```sass
$gray-dark:                 #292b2c !default;
$gray:                      #464a4c !default;
$gray-light:                #636c72 !default;
$gray-lighter:              #eceeef !default;
$gray-lightest:             #f7f7f9 !default;
```

#### PROPOSED

```sass
$gray-darkest:              lighten(#000, 16%) !default;   // #292929
$gray-darker:               lighten(#000, 29%) !default;   // #4a4a4a
$gray-dark:                 lighten(#000, 40%) !default;   // #666666
$gray:                      lighten(#000, 60%) !default;   // #999999
$gray-light:                lighten(#000, 86.5%) !default; // #dddddd
$gray-lighter:              lighten(#000, 93.3%) !default; // #eeeeee
$gray-lightest:             lighten(#000, 96.8%) !default; // #f7f7f7
```

Most of these gray color changes seem like a close enough match to me, and the grays used for text still pass the contrast test related to issue #20952.

#### History

The grays were originally defined with `lighten()`, but this was changed in 2014 with commit https://github.com/twbs/bootstrap/commit/b50498acd4ac46eb5c0985bfc330b624bc8a6856#diff-a7693b847cd07d7aa4c819d11e5becc1

With v4, the grays were changed with this commit https://github.com/twbs/bootstrap/commit/76907b8b1ef86ea20b9041d9c7e655d6d566ff97

### DIFF
--- a/docs/assets/scss/_component-examples.scss
+++ b/docs/assets/scss/_component-examples.scss
@@ -351,5 +351,5 @@
 }
 .highlight pre code {
   font-size: inherit;
-  color: $gray-dark; // Effectively the base text color
+  color: $gray-darkest; // Effectively the base text color
 }
--- a/docs/assets/scss/_footer.scss
+++ b/docs/assets/scss/_footer.scss
@@ -11,7 +11,7 @@
 
   a {
     font-weight: 500;
-    color: $gray;
+    color: $gray-darker;
 
     &:hover {
       color: $link-color;
--- a/docs/assets/scss/_nav.scss
+++ b/docs/assets/scss/_nav.scss
@@ -10,13 +10,13 @@
       &.active,
       &:hover,
       &:focus {
-        color: $gray-dark;
+        color: $gray-darkest;
         background-color: transparent;
       }
 
       &.active {
         font-weight: 500;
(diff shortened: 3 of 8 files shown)


sergiopantoja:
Actually now that I think about it, we can also just use the hex values directly instead of using the `lighten()` to approximate values. For example:

```sass
$gray-darkest:              #292b2c !default;
$gray-darker:               #464a4c !default;
$gray-dark:                 #636c72 !default;
$gray:                      #999999 !default;
$gray-light:                #dddddd !default;
$gray-lighter:              #eceeef !default;
$gray-lightest:             #f7f7f9 !default;
```

This would keep the exact same color palette with the benefit of adding `#999` and `#ddd` into the mix. Let me know what you all think and I will adjust the PR.

mdo:
Not going to be making any changes to these values for now. Thanks though.

PR #21664, opened 2017-01-11
Use correct value for xs and sm breakpoints in documentation

amygori:
The documentation on the available responsive utility classes includes an incorrect value for the `xs` max-width and `sm` min-width. This corrects the value and brings the available classes table values in line with the values elsewhere in the documentation (c.f. [Responsive Breakpoints](http://v4-alpha.getbootstrap.com/layout/overview/#responsive-breakpoints)) and with the functionality of the classes themselves.

### DIFF
--- a/docs/layout/responsive-utilities.md
+++ b/docs/layout/responsive-utilities.md
@@ -28,11 +28,11 @@ Try to use these on a limited basis and avoid creating entirely different versio
       <th></th>
       <th>
         Extra small devices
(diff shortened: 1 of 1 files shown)


mixalbl4-127:
This fix breaks this mixins (in alpha-6 worked fine)
```  @include media-breakpoint-only(xl)
  @include media-breakpoint-only(xs)```

PR #21670, opened 2017-01-11
Allow customization of colors, background and shadow for active, hover, focus

### LINKED ISSUE #19353: Ability to choose how much darker the button is on active

### LINKED ISSUE #21037: Better hover/active customization in button mixin
See #19414.

Ideally we could do better than a specific darken/lighten value.

pvdlg:
Fixes #21037, fixes #19353 and add similar type of variables for component that were missing some.

Basically all the colors properties that are derived from others (with `lighten` and `darken`) have now their own variable and are customizables.


roooodcastro:
Shouldn't the new color variables be set to a default value that's equal to what they were before this PR, to maximize compatibility?

mdo:
This ones mad out of date. Closing. Thanks though!

PR #18400, opened 2015-12-01
Convert Collapse accordion from Panels to Cards

Johann-S:
Hi,

Fix for https://github.com/twbs/bootstrap/issues/18375

But with this PR Collapse plugin is now dependent of cards, So if you're ok with that I'll work on how to remove this dependency.

A preview from the result : 

collapse

PR #21676, opened 2017-01-12
Table hover, accent, active variables  for both table and table-inverse

### LINKED ISSUE #21002: table-inverse : need a variable for hover, accent, active state
https://github.com/twbs/bootstrap/commit/d74a897f55ff01258b81e05c94a8126a819b8601#diff-82ee1ffb694a87db828fe22dcd868aed

Following the fix, table-hover, table-stripped combined with table-inverse works, but for accessibility reason, could we have a dedicated background value for those states?

For example : 
Actual value for table 
$table-bg-accent:     $table-bg-hover:      $table-bg-active:  
Can be completed with : $table-inverse-bg-accent:     $table-inverse-bg-hover:      $table-inverse-bg-active:

pvdlg:
Fixes #21002 and add other customization possibilities for table variables.

- Add variable to configure hover, accent, active variations for `table-inverse`
- Allow to customize variations of hover background for `table-hover` (both for regular and inverse table)
- Fix an issue in `table-hover` with a hovered `tr` that had a different background color than a tr with `table-active` while they are supposed to be the same according to [contextual-classes](https://v4-alpha.getbootstrap.com/content/tables/#contextual-classes)
- Update docs to reflect the possibility of mixing `table-hover`, `table-inverse` and row variations.


scss/_tables.scss
```diff
@@ -81,6 +81,11 @@
   tbody tr {
     @include hover {
       background-color: $table-bg-hover;
+
+      > td,
+      > th {
```
Johann-S: It seems Bootstrap's team try to avoid child selector, see this comment for exemple : https://github.com/twbs/bootstrap/pull/18400#r48385183

vanduynslagerp: Yes indeed.
There was also children selector in [_table-row.scss](https://github.com/twbs/bootstrap/pull/21676/files/54123bbb734cd44cdaa152c013c12605e98d6442#diff-d25719387baa3cdfd3780f0b21e62b85) that were there before my PR. I removed them as well for consistency.
Good catch, thanks !

mdo:
Would two different mixins be better for this to avoid the weird (to me) `null` bit here? Not super happy seeing/recommended that approach for folks.

vanduynslagerp:
Sure! I made the change.
We now have `table-row-variant` and `table-inverse-row-variant` mixins.

mdo:
Sorry, forgot about this PR! Closing for #22217 and skipping the new hover styles for contextual classes. Thanks though!

PR #21678, opened 2017-01-12
css: table as block issues (corresponding to #21677)

copperdesign:
Removed 100% width as in combination with block, it will break the float in grids and alignments and always clear the preceding. Also the .table-responsive cannot be applied to the `<table>` as display: block; breaks the automatic column width resizing. It should be applied to a wrapping div for overflow and above problems to work correctly


mdo:
Superseded by #22217

PR #21679, opened 2017-01-12
Enhancement: btn-pill

copperdesign:
In the fashion of `.badge-pill`, adding `.btn-pill` to `.btn`


mdo:
Thanks, but we'll pass on this.

PR #21695, opened 2017-01-13
Add script to check for broken links in docs

pvdlg:
Fixes #18568

This relies on [broken-link-checker](https://github.com/stevenvachon/broken-link-checker)
The check can be executed via `npm run check-broken-links` or `grunt check-broken-links`.

PR #21696, opened 2017-01-13
Remove broken links

pvdlg:
Fix link to postcss.js (link to v4-dev instead of master)
Remove the link to http://boot4.com/ as the site is down

All the other links are verified (with script from #21695)

### DIFF
--- a/docs/_data/translations.yml
+++ b/docs/_data/translations.yml
@@ -6,7 +6,7 @@
 - name: Chinese
   code: zh
   description: Bootstrap 4 中文文档教程
-  url: http://boot4.com/
+  url: http://wiki.jikexueyuan.com/project/bootstrap4/
 
 - name: Japanese
   code: ja
--- a/docs/_data/translations.yml
+++ b/docs/_data/translations.yml
@@ -3,6 +3,11 @@
   description: Bootstrap 中文文档
   url: http://v4.bootcss.com/
 
+- name: Chinese
+  code: zh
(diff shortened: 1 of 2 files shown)


docs/_data/translations.yml
```diff
@@ -3,11 +3,6 @@
   description: Bootstrap 中文文档
   url: http://v4.bootcss.com/
 
-- name: Chinese
-  code: zh
-  description: Bootstrap 4 中文文档教程
-  url: http://boot4.com/
```
bardiharborow: This link isn't broken, although it does redirect to `http://wiki.jikexueyuan.com/project/bootstrap4/`.

vanduynslagerp: I obtain yhe error page  below with `http://boot4.com/`
Although `http://wiki.jikexueyuan.com/project/bootstrap4/` works so I'll change the address to that.

Johann-S: I obtain the same as @vanduynslagerp when I go to Boot4.com

vanduynslagerp: Actually after visiting `http://wiki.jikexueyuan.com/project/bootstrap4/` a first time (by accessing this url directly, the redirect on `http://boot4.com/` seems to works...
Seems weird.
In order to reproduce, visit `http://boot4.com/` in Incognito mode.

Anyway it's probably better to provide a link to `http://wiki.jikexueyuan.com/project/bootstrap4/` rather than `http://boot4.com/`

PR #21703, opened 2017-01-13
Spacer vars with a more consistent doubling up pattern

### LINKED ISSUE #21702: Spacing sizes are a bit counter-intuitive
Referring to the spacing sizes explained in:
https://v4-alpha.getbootstrap.com/utilities/spacing/

I was surprised to see that these aren't linear increments `0, 1, 2, 3, etc` as might be expected by the classnames (i.e. p-3 yields 1 spacers worth, rather than 3)

but aside from that, with the pattern they currently have, it looks like they're supposed to double up each increment but they don't - the pattern being `0, 0.25, 0.5, 1, 1.5, 3` which is inconsistent.

Perhaps there's a logical reason?
If not, I've popped in a pull request for a consistent doubling up pattern `0, 0.25, 0.5, 1, 2, 4` although (in my opinion) linear increments might be better, for which I'd happily do a PR if desirable.

https://github.com/twbs/bootstrap/pull/21703

entozoon:
Fixes #21702.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -159,12 +159,12 @@ $spacers: (
     y: $spacer-y
   ),
   4: (
-    x: ($spacer-x * 1.5),
-    y: ($spacer-y * 1.5)
+    x: ($spacer-x * 2),
+    y: ($spacer-y * 2)
   ),
   5: (
-    x: ($spacer-x * 3),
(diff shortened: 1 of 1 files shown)


mdo:
Going to leave these as-is for now, thanks though.

PR #21711, opened 2017-01-13
JS constant rework => reduce minified JS by 7.5% (6.3% gzipped)

### LINKED ISSUE #20011: v5: JS constant declaration / minification
Most of the constant in javascript modules are declared as properties of constant objects, i.e.:

```
const ClassName = {
  BACKDROP : 'dropdown-backdrop',
  DISABLED : 'disabled',
  OPEN     : 'open'
}
```

Once minified, these constants are not mangled (for example search for 'BACKDROP' in [bootstrap.min.js](https://github.com/twbs/bootstrap/blob/v4-dev/dist/js/bootstrap.min.js)) because uglify, by default doesn't mangle object keys (I didn't find a safe way to have Uglify mangling objects keys). 

Defining a variable as a constant make it a read-only reference, however defining an object as a constant make this object reference read-only, but not its keys. See [const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).
As all the variables declared this way in Bootstrap are read but never reassigned, therefore really used as constants, I would imagine that the original intent was to make the actual variable a constant, rather than its wrapping object.

Declaring these constants directly rather than as object key would:
- allow mangling of the variables declaration and their references (smaller minified file, faster parsing)
- be more consistent with the intention (as I understand it) of declaring read-only variable

Is there a reason for which the constants are declared as object keys rather than directly as a constant ?

If not, I would like to propose a PR that would change the code example above into:

```
const CN_BACKDROP = 'dropdown-backdrop'
const CN_DISABLED = 'disabled'
const CN_OPEN = 'open'
```

The `CN_` naming convention for Class name (or `S_` for selectors, `E_` for events)  would maintain the readability of the current implementation (I imagine readability might be the reason for this choice ?).

Am I missing/misunderstanding something ? Would you consider such PR ?

pvdlg:
Fixes #20011

Improve JS minification:

| Before  | 46,653b | - | 12,648b | - |
| After  | 43,121b | 7.5% | 11,851 | 6.3% |
| With #21715  | 37,526b | 19.5% | 11,278 | 10.8% |

In addition it allows the JS linter to detect unused variables. For example [this variable](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tab.js#L45) is defined but not used and the linter cannot detect it.

### DIFF
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -25,22 +25,15 @@ const Alert = (($) => {
   const JQUERY_NO_CONFLICT  = $.fn[NAME]
   const TRANSITION_DURATION = 150
 
-  const Selector = {
-    DISMISS : '[data-dismiss="alert"]'
-  }
+  const SELECTOR_DISMISS = '[data-dismiss="alert"]'
 
-  const Event = {
-    CLOSE          : `close${EVENT_KEY}`,
-    CLOSED         : `closed${EVENT_KEY}`,
-    CLICK_DATA_API : `click${EVENT_KEY}${DATA_API_KEY}`
-  }
+  const EVENT_CLOSE = `close${EVENT_KEY}`
+  const EVENT_CLOSED = `closed${EVENT_KEY}`
+  const EVENT_CLICK_DATA_API = `click${EVENT_KEY}${DATA_API_KEY}`
 
-  const ClassName = {
-    ALERT : 'alert',
-    FADE  : 'fade',
-    SHOW  : 'show'
-  }
-
+  const CLASS_NAME_ALERT = 'alert'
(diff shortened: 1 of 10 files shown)


bardiharborow:
@cvrebert, I can go through this a triple check all the declarations have been transformed correctly, but could you tender your opinion on this?

mdo:
Closing as out of date.

XhmikosR:
Sorry for closing this @pvdlg, I agree we should try and revisit the constants, so even if late thanks for the contribution :)

I'll try to experiment with it for v5. I just want to figure out what was the reason to use Objects in the first place.

PR #21715, opened 2017-01-13
JS Mangle private properties => reduce minified JS by 11.9% (4.8% gzipped)

pvdlg:
All the private attribute (variables or function) in Bootstrap JS modules starts with `_` so we can easily configure Uglify to mandle them and produce a small minified JS file.

| Before  | 46,653b | - | 12,648b | - |
| After  | 41,059b | 11.9% | 12,063b | 4.8% |

In addition, this size reductions cumulate with #21711

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "sass-docs": "node-sass --output-style expanded --source-map true --precision 6 docs/assets/scss/docs.scss docs/assets/css/docs.min.css",
     "scss-lint": "bundle exec scss-lint --config scss/.scss-lint.yml --exclude scss/_normalize.scss scss/*.scss",
     "scss-lint-docs": "bundle exec scss-lint --config scss/.scss-lint.yml --exclude docs/assets/scss/docs.scss docs/assets/scss/*.scss",
-    "uglify": "uglifyjs --compress warnings=false --mangle --comments '/^!/' --output dist/js/bootstrap.min.js dist/js/bootstrap.js",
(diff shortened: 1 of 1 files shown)


bardiharborow:
I'm all for this in principle, but there's a problem, which is that `--mangle-props` does not exist in UglifyJS v2.4.10 (the latest version on npm). I've asked the UglifyJS team why they aren't publishing new versions to npm, and I'll update this issue when I hear back.

mdo:
This needs review and conflicts resolved before we could merge it now. @Johann-S @XhmikosR can you take a look and let me know if we should proceed with this PR or something else?

XhmikosR:
@vanduynslagerp: how safe is that?

XhmikosR:
BTW @vanduynslagerp you would need to move the change in build/uglifyjs.config.json

vanduynslagerp:
> @vanduynslagerp: how safe is that?

It's safe as long as you don't prefix your public class methods and properties with `_`.
And it provides a benefit as long as you do prefix your private class methods and properties with `_`.

It was the case when I opened this PR.
I didn't contribute to Bootstrap for a while and I didn't follow the recent updates, so I can't tell about the current safety/efficiency of this modification with the current code base.

Feel free to re-use or adapt the modification or close the PR if the code has changed and doesn't follow the convention of prefixing private methods and properties with `_`.

XhmikosR:
@Johann-S: what do you think about this? If you agree we should make a new branch and test this ourselves.

Johann-S:
Yes sure, our minified file size will grow with the removal off jQuery from our dependencies so optimizations are welcome here 👍

PR #21716, opened 2017-01-13
fix segmented buttons

### LINKED ISSUE #21593: Segmented buttons broken in Alpha 6 docs
https://v4-alpha.getbootstrap.com/components/input-group/#segmented-buttons

### LINKED ISSUE #21651: <input type="date"> inside .input-group renders incorrectly in Chrome.
Bootstrap 4 alpha6 on Chrome 55 on Windows.
Renders properly in Firefox and Edge.

## Test case
https://jsbin.com/pomicuyiza/1/edit?html,output

## Screenshot
image

gijsbotje:
Fixes #21593 and fixes #21651.

changed flex-direction to row
centering is now done by align-items instead of justify-content

this way there is no need for flex property on the .btn inside .btn-group

### DIFF
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -30,8 +30,8 @@
 .input-group .form-control {
   // Vertically centers the content of the addons within the input group
   display: flex;
-  flex-direction: column;
-  justify-content: center;
+  flex-direction: row;
+  align-items: center;
 
   &:not(:first-child):not(:last-child) {
     @include border-radius(0);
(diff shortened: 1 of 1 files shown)


scss/_input-group.scss
```diff
@@ -30,8 +30,8 @@
 .input-group .form-control {
   // Vertically centers the content of the addons within the input group
   display: flex;
-  flex-direction: column;
-  justify-content: center;
+  flex-direction: row;
```
mdo: This part can be dropped; it's the implicit default value of `flex-direction`.

mdo (dismissed):

scss/_input-group.scss
gijsbotje: you're right
dropped it in a new commit

mdo (approved):
Hell yeah, you da best.

PR #21719, opened 2017-01-14
Navbar: break at correct viewport width and give option to be collapsed at no or all breakpoints

### LINKED ISSUE #21691: Navbar 'md' triggers at 992 instead of 768
I noticed that while all of the other responsive classes define medium (md) as 768 px viewport, the navbar collapses the menu at 992 px. `navbar-toggleable-md` is the problematic class, although there could be others.

### LINKED ISSUE #21706: 4.0.0-alpha.6: Navbar should have style of .navbar-toggleable-xl by default
My navbar is compact and shouldn't be toggleable. I've had to set the class to `.navbar-toggleable-xl` to make this work, which is counter-intuitive.

I don't understand why a `navbar-toggleable-*` is required when `.navbar-toggleable-xl` makes it never toggle. Navbars should have the style currently given to `navbar-toggleable-xl` by default.

pvdlg:
Fixes #21691, fixes #21706

Currently the `navbar-toggleable-*` breaks at viewport size of the next breakpoints (i.e `navbar-toggleable-md` breaks at 992px instead of 768px).

This PR fix the `navbar-toggleable-*` generation to break at the corresponding breakpoint.
In addition, it generate `navbar-toggleable` that would prevent the navbar to collapse at all viewport size and mention it in the docs.
Finally it mentions in the docs that a navbar without `navbar-toggleable-*` will make it always collapsed.

It might be useful in order to created always/never collapsed navbar (as requested in #21706)


spongessuck:
This doesn't make any sense to me. If this PR makes the toggleable classes collapse at the specified breakpoint, `navbar-toggleable-xs` should always collapse- which is also the default if no class is set? And why should a `navbar-toggleable` class make it *never* toggleable?

vanduynslagerp:
The naming of "toggleable" might be a bit confusing.

It has to be understood this way:
- The navbar is by default toggled _(which mean collapsed and visible only by opening the menu)_
- The classes `navbar-toggleable-*` determine at which breakpoint the navbar becomes expanded _(which means toggle button invisible and `nav-item` visible)_
- The navbar becomes expanded when the viewport expand to a with superior to the breakpoint defined by the class ( _`navbar-toggleable-md` means collapsed below "md" and expanded above_)

The "toggleable" in `navbar-toggleable-md` means as "stop appearing toggleable above breakpoint md".
Therefore `navbar-toggleable` means "do not appear toggleable ever".
And no `navbar-toggleable-*` means "always appear toggleable".

It's counter intuitive because everything responsive class in Bootstrap usually means "**Do** what the class says above the breakpoint" and in the case of the navbar it means "**Stop doing** what the class says above the breakpoint".

I think it would be a lot clearer if these classes were renamed `navbar-expand-*` instead. Or `navbar-visible-*` maybe? Or even `navbar-untoggleable-*`...?

@mdo any thoughts on this ?

spongessuck:
I agree that `navbar-expand-*` would be a better naming convention, but I also think the logic should be reversed (unless I'm misunderstanding how it now works). No class means it's always collapsed (how it currently is), and `navbar-expand-xs` would make it always expanded/uncollapsed/not toggleable, `navbar-expand-sm` would be collapsed on xs, but expanded on sm and up, and so on. This would keep it cognitively harmonized with the grid classes, where `col-md-6` makes the element 50% width on medium screens and higher.

vanduynslagerp:
The fact that the navbar is by default collapsed (i.e no `navbar-toggleable-*` class means always collapsed) is by design. If I remember well it was the opposite on Bootstrap v3.

I'm pretty sure I read in a PR or comment the reason of this choice but I couldn't remember. Maybe @mdo can shed some light on this.

mdo:
Phew, working through all these navbar issues. I think i'm onto something. Using this PR as a base, I've tweaked how the toggler works, improved the container behavior, etc. How do the navbars in this JS Bin behave? https://jsbin.com/zixoxah/edit?output

vanduynslagerp:
The problem solved by this PR is also solved in the JSBin. Everything looks good to me !

One remark though, that is out of the scope of this PR (and maybe out of the scope of the component): The first navbar wraps weirdly on viewport smaller than 700px.
The search button wraps under the search input while it might be more logical to keep them next to each other. Also the carret of the dropdown wraps under it.



Maybe `navbar-collapse` should be `flex-wrap: wrap` and `navbar-nav` and `form-inline` should be `flex-wrap: nowrap`. Or at least use the flex-wrap utility in a doc sample.

mdo:
Oh yeah, that's just to demo the different navbar sizes—`sm`, `md`, and `lg` respectively for those first three. The broken display is definitely intentional.

vanduynslagerp:
Ok, so works great then !

What's your though on the naming convention ? see [comment](https://github.com/twbs/bootstrap/pull/21719#issuecomment-272602692) 
I also have similar questioning in #21720.

mdo:
Working on that shortly. Consolidating all these PRs into one and trying to include as many relevant changes from folks as I can to keep the attribution. It's been a lot to track. 😅

vanduynslagerp:
> Working on that shortly. Consolidating all these PRs into one and trying to include as many relevant changes from folks as I can to keep the attribution. It's been a lot to track. 😅

As far as I'm concern the attribution doesn't matter much to me, if that can make your life easier in the merge/consolidation process.

mdo:
Boom, peep #22230.

spongessuck:
I tried the JSBin and there doesn't seem to be an option to never collapse. I tried `navbar-toggleable-xs` and it acts the same as `navbar-toggleable-sm`.

mdo:
@spongessuck I've updated the example to include a never collapse option: https://jsbin.com/zixoxah/edit?html,output. It requires `.navbar-toggleable`. This is supported and documented in the changes from #22230.

spongessuck:
I argued that requiring a class called `navbar-toggleable` to make it never toggleable makes no semantic sense and I still feel that way.

spongessuck:
But thanks for the clarification.

mdo:
@spongessuck Ah, yeah, that's given the class name itself though, right? Not the specific need for a class that says "behave this way"?

spongessuck:
Yes, exactly.

`navbar-toggleable`, to me, says 'somehow, some way, it's toggleable' when in reality, it behaves exactly opposite to that.

Even `navbar-toggleable-xs`, if you follow the convention of the others, would mean 'the navbar expands at the xs breakpoint,' meaning it never collapses. It's still weird, but the suffix gives an important clue as to what the class does.

PR #21721, opened 2017-01-14
Harmonize navbar-brand, navbar-text and nav-link height

### LINKED ISSUE #21604: Responsive navbar is 2px shorter when content is hidden.
Example: http://v4-alpha.getbootstrap.com/components/navbar/#nav, see first snippet.

At narrow screen navbar block has height 8 + 38 + 8 px (54px). At wide screen height become 8 + 40 + 8 px (56px). Reproducible in latest FF/Chromium, Ubuntu 14.04.

I think height should not change. Any quick-tips how to fix?

### LINKED ISSUE #21705: 4.0.0.alpha6: .navbar-text should have the same vertical padding as .nav-link
Vertical padding (top and bottom):
- .navbar-text shows 0.425rem
- .nav-link shows 0.5rem

This means that 2 juxtaposed elements having these classes will not be vertically aligned.

Solution: Change .navbar-text to have 0.5rem of top- and bottom-padding.

pvdlg:
Fixes #21705 and fixes #21604

Use same padding for `navbar-text` and `nav-link`.

Compute the padding for `navbar-brand` so it will be the same height as `navbar-text` and `nav-link`, in order to take in account the difference of `font-size`.


mdo:
I don't think all three of these need the same height; they can be vertically centered easily enough with flexbox (and are by default I think). I'm not even sure I think the nav and text should share it. Beyond that, generating a height and then redistributing that is clever, but feels a bit obtuse for folks who might be looking to customize that in the future.

Going to noodle on this a bit more before merging.

mdo:
This was included in #22230.

PR #21722, opened 2017-01-14
Fix container within navbar on smallest breakpoint

### LINKED ISSUE #21590: [v4 css] Collapsed navbar toggle moves to awkward position
I am using a `navbar` with only a brand image; no brand name (as shown above). You can see that the collapse button is clearly out of place, over my brand image.

**Expected**

The collapse toggle should remain right-floated regardless of viewport width.

**Also...**

Does anyone else think that the toggle collapse button should be redesigned? For instance, on dark backgrounds it should be lighter (maybe completely white). I'd also be interested in seeing the border disappear — it is always a hassle to figure out the specific CSS heirarchy to modify it.

### LINKED ISSUE #21605: Centering Logo with Alpha 6 Navbar
I'm not overly familiar with Flexbox but i've updated to Alpha 6 and of course due to the new changes, my navbar broke.

I used to use ``mx-auto`` on my nav logo and then set a width to center it on the navbar but i can't seem to find a way to do this now with A6.

I can get it to 'partially' center in that its offset by a few pixels but as soon as i add another link on either side, it moves.

Any suggestions how to do this? Or would it be possible to push a PR to add a class to do this?

Cheers

pvdlg:
Fixes #21605 and fixes #21590.

At the smallest breakpoint the `container` has no `with` set. In the case of a `container` within a `navbar`, due to `margin-right: auto` and `margin-left: auto` the container content is coalesced toward the center.
This issue happen only in Chrome (see http://jsfiddle.net/rLqtw01t/ in chrome on xs viewport with and in another browser).

This PR set `margin-right: 0` and `margin-left: 0` for `container` within `navbar` under xs viewport  width to avoid the issue.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -20,6 +20,12 @@
   display: flex;
   flex-direction: column;
   padding: $navbar-padding-y $navbar-padding-x;
+
+  // Prevent container to center conter on smallest breakpoint (due to no with set for xs)
+  > .container {
+    margin-right: 0;
+    margin-left: 0;
+  }
 }
 
 
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -21,7 +21,7 @@
   flex-direction: column;
   padding: $navbar-padding-y $navbar-padding-x;
 
-  // Prevent container to center conter on smallest breakpoint (due to no with set for xs)
+  // Prevent container to center content on smallest breakpoint (due to no with set for xs)
   > .container {
     margin-right: 0;
     margin-left: 0;
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -21,10 +21,12 @@
   flex-direction: column;
   padding: $navbar-padding-y $navbar-padding-x;
 
-  // Prevent container to center content on smallest breakpoint (due to no with set for xs)
-  > .container {
-    margin-right: 0;
-    margin-left: 0;
+  @include media-breakpoint-down(nth(map-keys($grid-breakpoints), 1)) {
+    // Prevent container to collease content on smallest breakpoint on Chrome
+    > .container {
+      margin-right: 0;
+      margin-left: 0;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


browner12:
i like this, and this bug (?) is open in some other issues as well.  One thing I proposed was instead of setting the margins, set the `width: 100%`.  I believe they both accomplish the same thing, and this is a little more descriptive of what it is trying to accomplish.

vanduynslagerp:
It does accomplish the same thing. But the problem come from [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_grid.scss#L7) where we set `margin-left: auto` and `margin-right: auto` in regular container. So I think it's better to reset the problematic attribute in our situation :)

vanduynslagerp:
Actually, revisiting this PR it seems it's related to a Chrome bug.
Check this JSFiddle in Chrome and in another brower: http://jsfiddle.net/rLqtw01t/

It works normally in Safari and Firefox not in Chrome.

bardiharborow:
@vanduynslagerp if you could write a reduced test case (i.e without Bootstrap), then I can organise to file a Chrome bug.

vanduynslagerp:
Here you go: http://jsfiddle.net/Lgn9v9zt/
Actually I think it's a bug in Safari, as Firefox behave the same as Chrome.

Safari:


Firefox:


Chrome:


Bootstrap works as expected in Safari but not in Chrome or Firefox. So I think the fix in this PR is still relevant.
It seems that a bug in Safari make Bootstrap behave as expected.

On Mac OS X 10.12.2 and I'm using:
- Chrome 56.0.2924.59 beta (64-bit)
- Safari 10.0.2 (12602.3.12.0.1)
- Firefox 50.1.0

mdo:
Ahhh, I understand the issue now.

Auto margins are basically like pushes or pulls on flex items. We use `auto` to center a default `display: block` container, but when it's within a `display: flex` parent, it becomes a flex item that's trying to push from the left and from the right, resulting in a centered element. In reality, I believe Safari has the bug and Chrome and Firefox are correct here.

vanduynslagerp:
Yes, I think the bug is in Safari too. So this PR (or similar) is still relevant.
We can still open a bug with Safari, even though it doesn't "really" impact Bootstrap (as it make something work that shouldn't)

PR #21737, opened 2017-01-15
Use cssnano instead of clean-css => reduce minified CSS by 8.6% (4.9% gzipped)

pvdlg:
| clean-css  | 150,996b | - | 20,648b | - |
| cssnano  | 137,984b | 8.6% | 19,628b | 4.9% |


Olivia101:
@vanduynslagerp According to [this benchmark](http://goalsmashers.github.io/css-minification-benchmark/) clean-css is much faster and it's just as good at CSS minification.

vanduynslagerp:
I think the benchmark use Bootstrap 3.
I didn't measure the execution time, but I'm not sure it's relevant anyway, as you wouldn't minify when you dev, only to push a new version.
In terms of file size I did the measure myself with the actual code current on the branch v4-dev.

bardiharborow:
Ref #20324.

wolfy1339:
This needs rebasing because of the new build system

mdo:
Closing given this is based on Grunt.

PR #21739, opened 2017-01-16
Rename order utilities to intended class names

mdo:
Somehow I screwed this up in Alpha 6. These utilities should be `property-value` whenever possible, so instead of `flex-first` and the like, it should be `order-first`. I've left the `-1` and `1` as `last` and `first`, but renamed `unordered` to `0`. Is that weird? I'm not sure how best to handle this one.

Closes #21628.


bardiharborow:
docs/utilities/flexbox.md#L368 will need updating for `.order-0`.

PR #21741, opened 2017-01-16
Drop Normalize, port relevant parts to Reboot

### LINKED ISSUE #21588: Inputs and buttons use base sans-serif font, not native font stack
Repro: check out the [forms documentation page](http://v4-alpha.getbootstrap.com/components/forms/) for v4 alpha. Inspect one of the input elements. Observe that it is using a base `sans-serif`, *not* the native font stack:

screen shot 2017-01-06 at 5 10 53 pm

Presumably these should get the same base font as all the other elements on the page? I'm guessing there are probably also some other styles that are not being applied to inputs or maybe other elements too.

Cheers :beers:

#### Environment

- Operating system and version (Windows, Mac OS X, Android, iOS, Win10 Mobile)
  - macOS
  - Chrome Version 55.0.2883.95 (64-bit)

### LINKED ISSUE #21652: Buttons and input elements not inheriting $font-family-base
It seems like normalize is overriding the body with font-family: sans-serif, instead of reboot's setting of $font-family-base

### LINKED ISSUE #21740: Drop Normalize.css in v4
Reversing course from #20388.

I hate doing this right after our last alpha when we should be focused on as few changes as possible, but for stability of the project moving forward, I'm thinking we need to drop Normalize.css from v4. Here's why.

- In their next major release, they [plan to remove all _opinionated_ styles](https://github.com/necolas/normalize.css/pull/649). That includes fonts, body's default margin, and more—all of which we rely on.
- The last major release, Normalize 5, removed the `font: inherit` normalization from common form and button elements, causing #21588 (and several duplicate issues, sorry!).
- There are a ton of overrides we don't need given our browser support (e.g., IE9-, Safari 6, and iOS 4-7), meaning we can remove some unused code.
- Having our resets split across files really kinda sucks. This will consolidate them into a single point of reference for everyone.

Weigh in below if you have thoughts or feels. I'm drafting up a PR to see what this looks like in practice.

mdo:
Fixes #21740.

This removes a dedicated Normalize.css from the project, merging the relevant parts into Reboot. With the upcoming removal of "opinionated" styles that we rely on, combined with the growing overrides and inconsistencies between Normalize and Reboot, this should simplify things a bit.

This is missing a good amount of attribution (most obviously from Normalize), organization of rulesets, and lacking any real testing yet.

---

*Looking at other open issues and PRs, this also addresses a few bugs.*

Closes #21657, fixes #21588, fixes #21652.


mdo:
Just noticed Hound is running twice... 🤔

PR #21743, opened 2017-01-16
Fix JS components console error "Error: <Component> is transitioning"

### LINKED ISSUE #21727: JS components throws console error "Error: <Component> is transitioning"
Following the feature request #17800, the PR #17823 has been merged.
It prompted several issues with the error message `Error: <Component> is transitioning` in the browser console of the user.

**This ticket aims at centralizing those issues, gather feedback and discuss a possible solution.**

Here is the issues found so far:
- Moving cursor over popover quickly (#21600)
- Tooltip inside modals prevent to close the modal (#21607)
- User quickly triggering a component (collapse, carousel, popover etc...) animation by clicking fast on a button fast (#21704)
- Changing carousel slide before the previous finish transitioning: #21960

The issue #21687 has also been reported but it's not directly related. See [comment](https://github.com/twbs/bootstrap/issues/21687#issuecomment-272657301)

The original feature request and PR attempt to warn the user when he call the component API repeatedly instead of waiting for and even before making the next call. 
Even though this behavior is expected and documented. See [comment](https://github.com/twbs/bootstrap/issues/17796#issuecomment-145944908)
It worth noting that the previous behavior (skip the call if it's transitioning) is quite common and most framework works this way and expect the developer to understand the notion of asynchronous.

Unfortunately, with the current implementation (#17823) it seems we have inadvertently  created more problems than we solved... In addition we have added some complexity to the code, and it seems difficult to anticipate all the potential problem that will be found.

Here is a few solutions I can think of.

#### 1 - Revert #17823 and improve the documentation by adding a clear warning on each transition-capable module page and add more examples of events handling
- (+) Remove the added complexity
- (+) Work similarly to other framework that deal with asynchronous methods
- (+) Avoid any future issues along the lines of (#21600, #21607, #21704)
- (-) Even with a big warning on each module page in the docs, we might still have tickets like #17796

#### 2 - Improve #17823 to throw an exception only when the call has been made from the API (and not from a trigger element)
- (+) Solve #17800 and avoid tickets like #17796
- (-) Keep the added complexity
- (-) Might still create unexpected issues in cases where we mix API call and element trigger
- (-) Might still create unexpected "module interaction" issues like in #21607

#### 3 - Allow a method call or a trigger event on the JS components to cancel an on-going transition and a new one
- (+) Solve #17800 in way that doesn't produce console error and provide a good user experience 
- (+) Tooltips and Popovers already work that way
- (-) Difficult to implement for modal as a call to `$().modal` start 2 successive transitions (backdrop then the modal itself. It would require to distinguish which transition is on going and react differently.
- (-) Extremely difficult to implement for the Carousel component. As the items move independently, it would require to adapt the transition speed of each item based on their position when the transition is canceled
- (-) Overall add a lot of complexity and risks for regression

**Feel free to propose and discuss other potential solutions.**

My hunch is that solution 1 (revert and document) might be better. The problem we try to solve is related to developers who do not read the docs or understand asynchronous functions. It feels like this issue is better solved by improved documentation, especially when the "code solution" create issues for the end user.
And I think the doc can be improved in that regard.
The problem I see with the "code solution" is that it will either create a constant source of bugs or become extremely complicated in order to handle every possible case. Not mentioning that it would make adding new modules/feature more complicated.

I'd love to hear the community and bootstrap team ideas/feedback/opinion.

pvdlg:
Fixes #21727

### DIFF
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -160,7 +160,7 @@ const Modal = (($) => {
         event.preventDefault()
       }
 
-      if (this._isTransitioning) {
+      if (this._isTransitioning || !this._isShown) {
         return
       }
 
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -87,7 +87,6 @@ const Modal = (($) => {
       this._isShown             = false
       this._isBodyOverflowing   = false
       this._ignoreBackdropClick = false
-      this._isTransitioning     = false
       this._originalBodyPadding = 0
       this._scrollbarWidth      = 0
     }
@@ -111,14 +110,6 @@ const Modal = (($) => {
     }
 
     show(relatedTarget) {
-      if (this._isTransitioning) {
-        throw new Error('Modal is transitioning')
-      }
-
-      if (Util.supportsTransitionEnd() &&
-        $(this._element).hasClass(ClassName.FADE)) {
-        this._isTransitioning = true
-      }
       const showEvent = $.Event(Event.SHOW, {
         relatedTarget
       })
@@ -161,17 +152,8 @@ const Modal = (($) => {
         event.preventDefault()
       }
 
-      if (this._isTransitioning) {
-        throw new Error('Modal is transitioning')
-      }
-
-      const transition = Util.supportsTransitionEnd() &&
-        $(this._element).hasClass(ClassName.FADE)
-      if (transition) {
-        this._isTransitioning = true
-      }
-
       const hideEvent = $.Event(Event.HIDE)
+
       $(this._element).trigger(hideEvent)
 
       if (!this._isShown || hideEvent.isDefaultPrevented()) {
@@ -190,7 +172,9 @@ const Modal = (($) => {
       $(this._element).off(Event.CLICK_DISMISS)
       $(this._dialog).off(Event.MOUSEDOWN_DISMISS)
 
-      if (transition) {
+      if (Util.supportsTransitionEnd() &&
+         $(this._element).hasClass(ClassName.FADE)) {
+
         $(this._element)
           .one(Util.TRANSITION_END, (event) => this._hideModal(event))
           .emulateTransitionEnd(TRANSITION_DURATION)
@@ -259,7 +243,6 @@ const Modal = (($) => {
         if (this._config.focus) {
           this._element.focus()
         }
-        this._isTransitioning = false
         $(this._element).trigger(shownEvent)
       }
 
@@ -307,8 +290,7 @@ const Modal = (($) => {
 
     _hideModal() {
       this._element.style.display = 'none'
-      this._element.setAttribute('aria-hidden', 'true')
-      this._isTransitioning = false
+      this._element.setAttribute('aria-hidden', true)
       this._showBackdrop(() => {
         $(document.body).removeClass(ClassName.OPEN)
         this._resetAdjustments()
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -110,6 +110,14 @@ const Modal = (($) => {
     }
 
     show(relatedTarget) {
+      if (this._isTransitioning) {
+        return
+      }
+
+      if (Util.supportsTransitionEnd() && $(this._element).hasClass(ClassName.FADE)) {
+        this._isTransitioning = true
+      }
+
       const showEvent = $.Event(Event.SHOW, {
         relatedTarget
       })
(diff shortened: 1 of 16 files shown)


zcrwind:
> No violations found. Woof!

Nice work.

Johann-S (approved):
Great work @vanduynslagerp thank you 👍

gabrielkoerich:
Will it be merged?

PlugIN73:
Good time to merge! ;)

jakeNiemiec:
@PlugIN73 too bad **Savage** is taking so long on that "_special separate Travis CI build_".

js/src/modal.js
```diff
@@ -162,16 +161,17 @@ const Modal = (($) => {
       }
 
       if (this._isTransitioning) {
```
jbutz: @vanduynslagerp I think you should add a `!this._isShown` check here. If you don't and `hide()` is called when the modal is already hidden `_isTransitioning` is set to `true` and never reset to `false`.

mdo:
@Johann-S @bardiharborow Is this PR good to merge over #21613? Seems this PR fixes the tooltip in modal issue (#21607), and other related transition issues as mentioned in #21727.

Johann-S:
@mdo this PR do not fix the same issue as #21613. But yes the PR fix other issues related to #21727 (but not this one #21607)

js/src/modal.js
vanduynslagerp: Good catch ! Modification done. Thanks

bsbak:
the some problem: [github](https://github.com/twbs/bootstrap/issues/21607#issuecomment-289294424)

alecpl:
Would be nice to have popover fixed in the beta release, please.

Johann-S:
Maybe @vanduynslagerp you can rebase your branch to fix conflicts

Johann-S:
It seems I cannot update your repository anymore, so can you update your branch and I'll merge because everything is fine @vanduynslagerp

vanduynslagerp:
Yes, I removed the permission to edit from maintainers in my branches.
The branch is updated.

Turqueso:
Does this apply to npm? I just downloaded it with npm and have this error.

Johann-S:
You'll have this fix on the next release of BS4 @Turqueso not Alpha6

jakeNiemiec:
@Turqueso [And here is the progress for the next release](https://github.com/twbs/bootstrap/milestone/41) just in case if you (like me) are eagerly waiting.

selfagency:
I pulled the current v4-dev branch into a project in which I'm receiving a "collapse is transitioning" error (https://github.com/holger1411/understrap-child/issues/46) and the problem persists, even with this merge.

```
child-theme.min.js?ver=0.2.2:formatted:1876 Uncaught Error: Collapse is transitioning
    at a.show (child-theme.min.js?ver=0.2.2:formatted:1876)
    at a.toggle (child-theme.min.js?ver=0.2.2:formatted:1870)
    at HTMLDivElement.<anonymous> (child-theme.min.js?ver=0.2.2:formatted:1997)
    at Function.each (jquery.js?ver=1.12.4:2)
    at jQuery.fn.init.each (jquery.js?ver=1.12.4:2)
    at jQuery.fn.init.a._jQueryInterface [as collapse] (child-theme.min.js?ver=0.2.2:formatted:1987)
    at HTMLButtonElement.<anonymous> (child-theme.min.js?ver=0.2.2:formatted:2020)
    at HTMLDocument.dispatch (jquery.js?ver=1.12.4:3)
    at HTMLDocument.r.handle (jquery.js?ver=1.12.4:3)
```

`child-theme.min.js` is a concatenated version of bootstrap 4's js modules. You can see it [here](https://gist.github.com/selfagency/9ab2bfe459b2ab8caf56e7a5231c91e6).

hernimen:
still getting this error. Any idea?

Johann-S:
This fix will be in our next release

selfagency:
I was able to resolve this issue by removing `hidden-md-down` from my `navbar-menu`

ReSpawN:
@Johann-S when is your next release scheduled. We're still using static shows instead of transitions/animations. We can merge and pull from this brand but we'd rather wait for a release from your end.

Johann-S:
We shipped a release 4 days ago see : https://github.com/twbs/bootstrap/releases/tag/v4.0.0-beta or our blog

ReSpawN:
Don't know how I missed that! I'll start implementation in the morning! Tx.

ReSpawN:
Still getting the [Component] is transitioning errors though...

emmanueltouzery:
beta fixed it for me

PR #21747, opened 2017-01-16
Fix horizontal scrollbar appearing in docs

tagliala:
On small screens (< 576px), all documentation pages containing
`.bd-example` styled tags, show the horizontal scrollbar because of a
wrong horizontal margin property.

This PR sets the horizontal margin on gutter width basis instead of rem.


tagliala:
Example: https://v4-alpha.getbootstrap.com/content/reboot/

image

Issue #20620, opened 2016-09-03
v4 list-group not work correctly with tabs


agavazov:
On V3 when I make tabs with list-group it works fine, but in V4 tabs are open once and active class does not disappear.

This example is with V3, you can see whats happen when test it with V4

```
<html>
<head>
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js"></script>
    <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" rel="stylesheet"/>
</head>
<body>

<ul class="list-group">
    <li class="list-group-item active"><a href="#panel1" data-toggle="tab">Open panel 1</a></li>
    <li class="list-group-item"><a href="#panel2" data-toggle="tab">Open panel 2</a></li>
</ul>

<hr/>

<div class="tab-content">
    <div class="tab-pane active" id="panel1">
        <div class="panel-body">Panel 1 body</div>
    </div>
    <div class="tab-pane" id="panel2">
        <div class="panel-body">Panel 2 body</div>
    </div>
</div>

</body>
</html>
```

agavazov:
You can see the same problem with this example http://www.codeply.com/go/rv3Cvr0lZ4

mdo:
Yeah, right now our tab JS only works with navigation components I think, not the list group. Unsure how complicated it'd be to add support for that, but maybe our newest JS friends @Johann-S and @bardiharborow could give us an idea of if it's doable?

Johann-S:
It's because of this line : https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tab.js#L50

If we change this selector to accept ``.list-group-item`` it will work fine

mdo:
> You can make it works like this, add `.nav-item` to `li`

Hmmm, maybe. We could also update that line of JS to accept multiple selectors, right? I could go either way tbh. If we do the multiple selectors for our JS elsewhere, then I'd say let's do that. But if we encourage mixing/matching of classes across different components, than let's do that.

Thoughts?

Johann-S:
It's a tough choice, FYI I didn't find any multiple selectors in our JS currently and tbh it's difficult to know which solution is the better one 😞

giulianovelli:
[...]

Issue #21755, opened 2017-01-17
V4 list-group with an active li, invisible link


Johann-S:
It seems when an ``li`` in ``.list-group`` with ``.active``, links aren't visible 

See : https://jsfiddle.net/m6zz8sw4/1/

Quy:
Move `class="list-group-item active"` to the `<a>` tag.

Quy:
Why not do it as instructed in the doc?

> Use <a>s or <button>s to create actionable list group items with hover, disabled, and active states by adding .list-group-item-action. We separate these pseudo-classes to ensure list groups made of non-interactive elements (like <li>s or <div>s) don’t provide a click or tap affordance.
<div class="list-group">
  <a href="#" class="list-group-item active">
    Cras justo odio
  </a>
  <a href="#" class="list-group-item list-group-item-action">Dapibus ac facilisis in</a>
  <a href="#" class="list-group-item list-group-item-action">Morbi leo risus</a>
  <a href="#" class="list-group-item list-group-item-action">Porta ac consectetur ac</a>
  <a href="#" class="list-group-item list-group-item-action disabled">Vestibulum at eros</a>
</div>

vanduynslagerp:
Here is the link referenced by @Quy : https://v4-alpha.getbootstrap.com/components/list-group/#links-and-buttons

Johann-S:
Yes it works well with your example @Quy , but according to the documentation we can use both ``<ul>`` and ``<div>``. So for me the behavior should be the same. Am I Wrong ?

Quy:
I would think so too, however, going by the examples in the documentation, use  `<ul>` for non-interactive elements and `<div>` with `a` or `button` for interactive elements.

mdo:
I think I'll close this as a won't fix for now given we try to keep those list groups as singular large click areas. If you need the inner hyperlink, `.text-white` is also available.

PR #21756, opened 2017-01-17
Allow to use Tab.js with list-group

Johann-S:
Allow Tab.js selector to accept ``list-group``

Ref : #20620

### DIFF
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -46,9 +46,9 @@ const Tab = (($) => {
     LI                    : 'li',
     DROPDOWN              : '.dropdown',
     LIST                  : 'ul:not(.dropdown-menu), ol:not(.dropdown-menu), nav:not(.dropdown-menu)',
-    FADE_CHILD            : '> .nav-item .fade, > .fade',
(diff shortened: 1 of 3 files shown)


mdo:
Hell yeah, thanks!

Johann-S:
You're welcome 😀👌

vanduynslagerp:
Is there really a point in supporting `list-group` with tabs ? As there is already [vertical navs](https://v4-alpha.getbootstrap.com/components/navs/#vertical).

As far as I can tell both achieve the same thing.

The original issue #20620 mention "an example from v3". I think it's intended in v4 to offer vertical nav for this case while in v3 list-group were the way to go.

Johann-S:
It's the result of this discussion #20620, because we think (@mdo and myself) that it's better to allow Tab.js to work with every lists provided by BS

Johann-S:
Due to this comment : https://github.com/twbs/bootstrap/issues/21755#issuecomment-274511517

Allow to use this structur of ``.list-group`` : 
```html
<div class="list-group">
  <a href="#home" class="list-group-item active" data-toggle="tab">Home</a>
  <a href="#profil" class="list-group-item list-group-item-action" data-toggle="tab">Profil</a>
</div>

PR #21760, opened 2017-01-17
Fix .jumbotron-hr to .jumbotron hr

Quy:


mdo:
If we really want to remove the class for an element selector, we have to limit it to immediate children somehow, and that gets complicated with containers, grid system, etc. I think we have to stick with the class unfortunately, or remove the change outright.

Thoughts?

Quy:
Would this apply to `alert-*' too?
```
.alert-success hr {
  border-top-color: #c1e2b3;
}

.alert-info hr {
  border-top-color: #a6d5ec;
}

.alert-warning hr {
  border-top-color: #f7ecb5;
}

.alert-danger hr {
  border-top-color: #e4b9b9;
}
```

mdo:
Ended up nuking this entirely because that variable uses `rgba()`. cf2de1fd056bb373cd56696945892b49556bd0a4

PR #21770, opened 2017-01-18
Documentation: Correct prefix for flexbox order classes

rluba:
… they are called `flex-…`, not `order-…`, as [it can be seen in the source](https://github.com/twbs/bootstrap/blob/8167682deb6146e1949cb49baf6e87d350befe3d/scss/utilities/_flex.scss#L9-L11).


Quy:
Related to #21739

bardiharborow:
Closing in favour of #21739.

rluba:
I agree. Fixing the naming inconsistency is the better way. 👍

PR #21782, opened 2017-01-19
Fix input placeholder color

### LINKED ISSUE #21624: Placeholder is visually indistinguishable from normal text

sergiopantoja:
Fixes #21624.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -426,7 +426,7 @@ $input-border-focus:             lighten($brand-primary, 25%) !default;
 $input-box-shadow-focus:         $input-box-shadow, rgba($input-border-focus, .6) !default;
 $input-color-focus:              $input-color !default;
 
-$input-color-placeholder:        $gray-light !default;
+$input-color-placeholder:        #999 !default;
 
 $input-padding-x-sm:             .5rem !default;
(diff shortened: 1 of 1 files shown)


mdo:
The placeholder color was changed because it didn't pass color contrast standards. This addresses the visual problem, but reintroduces the old contrast problem. Given that, thanks, but we'll pass for now.

patrickhlauke:
@mdo just cross-referencing this https://github.com/twbs/bootstrap/issues/21624#issuecomment-275848876

vanillajonathan:
The placeholder color used to be set to `$grey-light` which had the value `#636c72`.
Nowadays it uses the variable `$gray-600` with the value `#6c757d`.

Would it be possible to use `$gray-500` with the value `#adb5bd`?
That would improve the contrast problem. But would it fail to pass the color contrast standards?

PR #17536, opened 2015-09-09
[Fix #17400] Add tests about padding restoration

Johann-S:
Hi,

It's a fix a for https://github.com/twbs/bootstrap/issues/17400
I added several tests to check : 
- If a padding-right is applied when modal is taller than viewport
- If padding-right is removed after that

PR #21800, opened 2017-01-20
Fix padding with modal and fixed navbar 

### LINKED ISSUE #18373: Errors in Modal._setScrollbar()
See https://github.com/twbs/bootstrap/issues/18366#issuecomment-160357727

### LINKED ISSUE #20824: Modal open causes fixed content to shift
I believe there is an issue with opening modals that causes fixed content to be shifted.

Here's an example http://codepen.io/eddywashere/pen/ZpXrxQ

I'm on OSX, 10.11.5, Chrome 53 and have `System Preferences > General > Show scroll bars: always`.

I believe this is sort of tracked in the Modal section of v4 issue https://github.com/twbs/bootstrap/pull/17021, but does not contain an active issue to track it.

> Finally fix that shifting content bug by updating the modal.js to not adjust body padding, but rather the padding on a specific set of classes.

I found an issue that was closed because it was part of v3, https://github.com/twbs/bootstrap/issues/14040#issuecomment-244817197.

### LINKED ISSUE #21700: Wrong body padding-right on modal with fixed-top nav
This padding-right in `body` on modal active is wrong when using `nav` with `.fixed-top`

https://jsfiddle.net/rorc/nke1dvpz/3/

### LINKED ISSUE #21840: Modal unexpected behaviour - class modal-open removed and padding-right issue
Hi, Ive noticed a problem with launching modals, If you launch a modal from within a modal the class .modal-open is removed from the body and if you dismiss the modals padding-right 17px; isn't removed from the body, wich causes your layout to shift to the left. Workaround is a little bit of CSS. However i believe this is unexpected behaviour. It worked fine in bootstrap 3.

I have reproduced the bugs and commented it aswell:
http://codepen.io/revolutionart/pen/YNxjgw

pvdlg:
Fixes #18373, fixes #21700, fixes #20824, fixes #21840.

Now `_setScrollbar`:
- Save the current `body` `padding-right` in `this._originalBodyPadding`
- Add a `pading-right` to the `body` corresponding to the current `padding-right` + the scrollbar size
- Save the current `padding-right` of each fixed element in the element's `'padding-right'` data-attribute
- Add a `pading-right` to each fixed element corresponding to the current `padding-right` + the scrollbar size

And `_resetScrollbar`:
- Reset the `body` `padding-right` to the saved value
- Reset the `padding-right` of each each fixed element to the saved value

The PR #17536 added 2 unit tests to verify that the `padding-right` is properly added to the `body`. However those 2 tests were ineffective because:
- The function `_getScrollbarWidth` relies on the css class `.modal-scrollbar-measure` to determine the scrollbar width and therefore the value of the `padding-right` to add to the `body`. But the class was not present in the unit tests.
- QUnit doesn't add a scrollbar even when en element is higher than the viewport, the function `_getScrollbarWidth` was always returning `0px`, therefore during a QUnit test the modal plugin was always considering that there is no sidebar, so no `padding-right` was added to the `body`

Note: `_getScrollbarWidth`, is called by `_checkScrollbar` and sets the current scrollbar width in `this._scrollbarWidth`. This value is later used by `_setScrollbar` to determine the width of the padding to add to the `body` and fixed elements. **So `_setScrollbar` depends on `_getScrollbarWidth`and  `.modal-scrollbar-measure`.**

In order to solve the first issue, this PR add the Bootstrap css to the Unit test so the class `.modal-scrollbar-measure` and `_getScrollbarWidth` can work as intended during the unit tests.

To solve the second one, the tests have been modified to simulate a scrollbar by adding `10px` to **the `html` element**. Then after the modal is opened we verify that a `padding-right` of `10px` (corresponding to the width of the simulated scrollbar) has been added to **the `body` element**.
**Note: The simulated scrollbar is added to the `html` element, and after the modal is opened the test verify the `padding-right` of the `body` (`html` != `body`).**

In addition the tests now also check for the `padding-right` to be applied on the fixed elements.


Johann-S:
Closed due to the merge of : #18441

PR #21801, opened 2017-01-21
Fix alignment of progress-bar text

jgibson02:
When the height of a .progress/.progress-bar is changed, the text within it does not appear to vertically align properly. Adding vertical-align: middle to the <span> elements within the .progress-bar should fix alignment issues.

### DIFF
--- a/scss/_progress.scss
+++ b/scss/_progress.scss
@@ -20,6 +20,10 @@
   background-color: $progress-bar-bg;
 }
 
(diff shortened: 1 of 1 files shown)


mdo:
Beyond this not solving the issue for the recommended label approach for the progress component, generic selectors targeting just elements like this are discouraged. Pushing up a proper fix for this based on what we had in v3.

Issue #21798, opened 2017-01-20
Incorrect Layout/Grid Documentation


marcusmolchany:
In the `Mix and match` section, the first example does not behave like the comment would suggest. Rather than the two cols stacking in the mobile range they remain side by side in the column each taking up half of the width.

```html
<div class="bd-example-row">
{% example html %}
<!-- Stack the columns on mobile by making one full-width and the other half-width -->
<div class="row">
  <div class="col col-md-8">.col .col-md-8</div>
  <div class="col-6 col-md-4">.col-6 .col-md-4</div>
</div>

...
```

I think maybe the first col should have the classes `"col-12 col-md-8"` ?

bardiharborow:
Closing per #21806.

PR #21806, opened 2017-01-21
Update grid layout docs

marcusmolchany:
Update Mix and match section with correct responsive classes
for stacking cols.


marcusmolchany:
Closes https://github.com/twbs/bootstrap/issues/21798

PR #21812, opened 2017-01-22
Remove `cursor: pointer;`

### LINKED ISSUE #21781: Disabled custom checkboxes and radio buttons cursor looks clickable
- Operating system and version: Mac OS X 10.11
- Browser and version: Chrome 55, Firefox 50, Safari 10

Go to https://v4-alpha.getbootstrap.com/components/forms/#disabled hover over the disabled checkbox. Notice the cursor.

disabled-custom-checkbox-radio

History: #17918

mdo:
Turns out using `cursor: pointer` on actions (e.g., buttons) is incorrect. I removed the `cursor: pointer` from our buttons in Alpha 6 knowing this, but I left the rest of the properties throughout the codebase for another time. This PR aims to finish that work, removing nearly all the remaining `cursor: pointer` properties from our CSS.

I still need to figure out the `role="button"` one for iOS click-ability though.

This would also fix #21781.

---
#### References

From the [W3C spec](https://www.w3.org/TR/CSS21/ui.html#propdef-cursor):

> **pointer**
> The cursor is a pointer that indicates a link.

From [Microsoft's design guidelines](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742466%28v=vs.85%29.aspx):

> **To avoid confusion, it is imperative not to use the hand pointer for other purposes.** For example, command buttons already have a strong affordance, so they don't need a hand pointer. The hand pointer must mean "this target is a link" and nothing else.

Regarding the cursor pointer from [Apple's HIG](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/Pointers.html):

> The content is a URL link.

### DIFF
--- a/scss/_badge.scss
+++ b/scss/_badge.scss
@@ -33,7 +33,6 @@ a.badge {
   @include hover-focus {
     color: $badge-link-hover-color;
     text-decoration: none;
-    cursor: pointer;
   }
 }
 // scss-lint:enable QualifyingElement
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -131,7 +131,6 @@
     margin-right: $carousel-indicator-spacer;
     margin-left: $carousel-indicator-spacer;
     text-indent: -999px;
-    cursor: pointer;
     background-color: rgba($carousel-indicator-active-bg, .5);
 
     // Use pseudo classes to increase the hit area by 10px on top and bottom.
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -10,7 +10,6 @@
   @include hover-focus {
     color: $close-color;
     text-decoration: none;
-    cursor: pointer;
     opacity: .75;
   }
 }
@@ -23,7 +22,6 @@
 // scss-lint:disable QualifyingElement
 button.close {
   padding: 0;
-  cursor: pointer;
   background: transparent;
   border: 0;
   -webkit-appearance: none;
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -15,7 +15,6 @@
   min-height: (1rem * $line-height-base);
   padding-left: $custom-control-gutter;
   margin-right: $custom-control-spacer-x;
-  cursor: pointer;
 }
 
 .custom-control-input {
@@ -200,7 +199,6 @@
   max-width: 100%;
   height: $custom-file-height;
   margin-bottom: 0;
-  cursor: pointer;
 }
 
 .custom-file-input {
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -216,7 +216,6 @@ select.form-control-lg {
 .form-check-label {
   padding-left: $form-check-input-gutter;
   margin-bottom: 0; // Override default `<label>` bottom margin
-  cursor: pointer;
 }
 
 .form-check-input {
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -260,17 +260,6 @@ svg:not(:root) {
 }
 
 
-// iOS "clickable elements" fix for role="button"
-//
-// Fixes "clickability" issue (and more generally, the firing of events such as focus as well)
-// for traditionally non-focusable elements with role="button"
-// see https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile
-
-[role="button"] {
-  cursor: pointer;
-}
-
-
 // Avoid 300ms click delay on touch devices that support the `touch-action` CSS property.
 //
 // In particular, unlike most other browsers, IE11+Edge on Windows 10 on touch devices and IE Mobile 10-11


patrickhlauke:
what's the issue with iOS clickability? seems that a vanilla element is tappable as long as it has a click handler (and event delegation should work fine as well, see https://patrickhlauke.github.io/getting-touchy-presentation/#220)

patrickhlauke:
Random test https://jsbin.com/zarituwaba/ works fine for me on iOS

mdo:
Guess there isn't an issue then. :laughing: I was just going by the code comment there. I'll happily remove the last reference there if it indeed isn't needed.

IamManchanda:
for me, 

```scss
cursor: not-allowed; 
```
should be used instead

ryanburnett:
@mdo I think button hover should retain `cursor: pointer`. The W3C and Apple HIG both state that it's for a link, which a button does contain. For example, go to http://www.apple.com/ and mouse over the search or shopping bag icons and you'll get a pointer. Go to https://www.microsoft.com/en-us/ and mouse over the search or cart icons and you'll get a pointer. Go to https://www.w3.org/ and mouse over the "Go" button in the top right and you'll get a pointer. If the three sites you references all use pointers for buttons I think Bootstrap should too.

P.S. Keep up the great work on Bootstrap <3

patrickhlauke:
@ryanburnett and this PR does not remove the hand cursor for links styled visually as buttons. it simply removes other situations where `cursor: pointer` is forced inappropriately, and leaves it up to the browser to handle the cursor as it normally would for the appropriate elements

patrickhlauke:
> it's for a link, which a button does contain

links and buttons (actual buttons that action an in-page change, submit a form, or similar) are different beasts though.

the examples you have from apple, microsoft and w3 actually go directly against the W3C and Apple HIG, as they're also forcing `cursor:pointer` on things that are NOT links (only one that is a link in your examples is the Microsoft cart icon, as that navigates to a new page - at least in my case, with an empty cart)

patrickhlauke:
if designers/developers do want to have `cursor:pointer` for things that normally shouldn't have that cursor, they can add it themselves. but out of the box, bootstrap now follows the browser's default (e.g. if it's a `<button>` or similar, cursor doesn't change)

joppuyo:
In my opinion this change it a bit confusing, I have some controls in my application that use the `btn` class. For technical reasons, some of those use `<button>` element and others `<a>`. In Bootstrap 4, hovering over some of them shows the cursor but others show the pointer leading to an inconsistent experience.

I preferred the old way of "normalizing" the button styles regardless of whether they are a `<button>` or `<a>` so they look and feel the same.

Luckily it's only one line of code to change the behavior back to old but it would be nice if it worked out of the box.

nextgenthemes:
I don't get it. I don't care what the specs say I don't think there its "indicating better" when I hover over a button and nothing happens. If there were different cursors for buttons and links this would be a different, this way just nothing happens for buttons so for transparent buttons or whatever that is not changes their styles on hover there is no indication at all, end even with style changes I think a cursor indication is always useful, also for accessibility ...

browner12:
maybe this would be better made into a variable, so users could toggle it on and off based on personal preference?

purple-penguin7:
This change defies logic.  You're conforming to spec for the sake of conforming to spec, not because it actually serves a constructive purpose.  The irony in all of this is that the specs and guidelines aren't even followed by the groups giving them...

W3C's own website styles buttons to have a pointer.
Microsoft's own website styles buttons to have a pointer.
Apple's own website styles buttons to have a pointer.

A mass majority of your users will overwrite this style.  Is it easy?  Sure, but that's completely irrelevant.  Why make a change that 99% of your users will be forced to overwrite.  Worse yet, the small fraction of people who don't will actually provide a worse UX to their users in terms of usability.

Guess which one has a more adverse impact?

A user hovering over a button and seeing the cursor change (i.e. pretty much what every computer user has been conditioned to expect for the past decade).

- or -

A user hovering over a button, not seeing the cursor change, and then wondering if it's disabled, not clickable, etc.

The latter is a much worse UX.

When a spec breaks basic user expectations, the spec is wrong.  If W3C comes out tomorrow and states all buttons must be blue, will Bootstrap conform to that too?  I understand the importance of standards and trying to follow specs as closely as possible, but at some point basic logic and reasoning have to come into play.  When there literally isn't a single constructive reason to make this change other than W3C says the pointer means a link, the change shouldn't be made.  Users don't see a pointer and think, "Oh, this must be an anchor element that I can click to take me to another page,"  they think, "I can click or otherwise interact with this."

This isn't a constructive change.

nextgenthemes:
Totally agree, fun fact is the issue he thought this fixed https://github.com/twbs/bootstrap/issues/21781 but that was reopened because this actually fixed nothing. Just makes horrible UX.

But I have to say I checked w3.org and their search button has in fact no pointer style on hover. Maybe somewhere else on the site but it would be too funny if they would actually have
```css
button {
  cursor: pointer;
}
```
in their stylesheet.

purple-penguin7:
They actually have exactly that (well, almost, they use a class called button instead of applying it to the element itself) :p

https://www.w3.org/2008/site/css/minimum

Check out the button class that they subsequently apply to the top navbar / main "Go" button in the top right on their website.

panva:
are button-looking elements (e.g. `a.btn`) intended to still have the cursor? (alpha.6 still does)

patrickhlauke:
@panva alpha.6 is older than this PR (which targets the first beta). explicit setting of `cursor` now doesn't happen anymore for buttons etc. what you WILL see though is that browsers will use whatever their default cursor is. if you're using an `<a href="...">` link (whether styled as a `.btn` or not) browsers will show the pointer cursor by default.

panva:
@patrickhlauke that's good news! way better than stripping it, thanks for the insight.

matejlednar:
What's indicator (visual feedback for user) that element is clickable, e.g. any icon on the screen (difference between clickable icon and static background image / standard image)?

spongessuck:
Both buttons and anchors have hover styling, making it clear that they're
interactive.

matejlednar:
Hamburger icon is SVG.  There si no way how to change color via CSS. I tried to change stroke color and use encodingURI for correct string format but stroke color was always removed, don't know why.

mikebronner:
I would like to suggest that the pointer be retained on buttons that are styled and behave as links, for example `btn-link`.

lucasdavidferrero:
I think that 'cursor: pointer' should be an utility because we use it a lot in differents elements so I've opened a issue (#23709).

printercu:
@mdo links and quotes from first comment seems to be related to desktop applications. MSDN and apple.com use `cursor: pointer` for buttons in sign-in forms.

PR #21825, opened 2017-01-24
Update _custom.scss imports in other builds

### LINKED ISSUE #21682: Column padding for fixed column vs auto
I noticed that when using `col-md-3` and `col-md-auto` together, that `col-md-3` gets 15px padding, but `col-md-auto` does not. I am a new contributor so forgive me if this has been addressed already (although I did search for issues relating to this).

The issue is that in `bootstrap-grid.css`, `col-md-auto` is only mentioned once to make it a flexbox and set `width: auto;`. There are four responsive blocks that contain all columns (in all of their responsive forms) that apply `padding-left: 15px;` and `padding-right: 15px;`.

This seems to me to be a bug, not a feature. Should this be fixed?

P.S. I am more than happy to make the change and submit a pull request, but for a big project like this I presume the development team doesn't operate like that.

mdo:
- Correctly places the _custom.scss file first in the other two buids.
- Also removes the clearfix mixin from the grid build because flexbox yo.

Fixes #21682.

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -22,22 +22,15 @@ html {
   box-sizing: inherit;
 }
 
-
-//
-// Variables
-//
-
+@import "custom";
 @import "variables";
 
 //
 // Grid mixins
 //
 
-@import "mixins/clearfix";
 @import "mixins/breakpoints";
 @import "mixins/grid-framework";
 @import "mixins/grid";
-
-@import "custom";
 
 @import "grid";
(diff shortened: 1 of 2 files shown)


bardiharborow (approved):

Starsam80:
I also commented here too: https://github.com/twbs/bootstrap/pull/21603#issuecomment-271488213 but it looks like it went unnoticed. I was going to do a pull request and I totally forgot about it.

PR #21856, opened 2017-01-27
Update tooltip.js getPosition() 

### LINKED ISSUE #21855: Tooltip on svg misplacement after scrolling the page
Issue happens because:
The Element.getBoundingClientRect() method returns the size of an element and its position relative to the **viewport**.

Need to extend tooltip position with:

```
 Tooltip.prototype.getPosition = function ($element) {
    $element   = $element || this.$element

    var el     = $element[0]
    var isBody = el.tagName == 'BODY'
   
    var elRect    = el.getBoundingClientRect();
    elRect = $.extend({}, elRect, {
        top: elRect.top + window.pageYOffset, 
        left: elRect.left + window.pageXOffset
   });
... e.t.c
```

pm-Egor-Genning:
Fixes #21855.

### DIFF
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -359,7 +359,8 @@
     var el     = $element[0]
     var isBody = el.tagName == 'BODY'
 
-    var elRect    = el.getBoundingClientRect()
+    var elRect    = el.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


Johann-S:
Support for v3 has mostly ceased. Please let us know if this issue affects v4.
See : #20631

mdo:

---

_This is a saved reply._

PR #21857, opened 2017-01-27
fix distorted card images

### LINKED ISSUE #21650: Card images distorted
Seems like the Flexbox card implementation in alpha6 has produced some unwanted effects for card images – distortion occurs regardless if using .img-fluid for the .card-img-top or not. I know it has been discussed before but if the default behaviour of .card-img-top means distortion of the image I'd say .img-fluid should be applied to card images by default.

(OS X 10.12.1 Chrome 55.0.2883.95)

http://codepen.io/membla/pen/WRrvqG

gijsbotje:
Fixes #21650.

The image won't be stretched with the css in this PR.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -160,6 +160,7 @@
 // Card image
 .card-img {
   // margin: -1.325rem;
+  width: 100%;
   @include border-radius($card-border-radius-inner);
 }
 .card-img-overlay {
@@ -175,9 +176,11 @@
 
 // Card image caps
 .card-img-top {
+  max-width: 100%;
   @include border-top-radius($card-border-radius-inner);
 }
 .card-img-bottom {
+  max-width: 100%;
   @include border-bottom-radius($card-border-radius-inner);
 }
 


gijsbotje:
ping

scss/_card.scss
```diff
@@ -160,6 +160,7 @@
 // Card image
 .card-img {
   // margin: -1.325rem;
+  width: 100%;
```
oktalk: I think this should be `max-width` instead of `width`.

gijsbotje: If i change your suggestion the images get distorted again (running on macbook pro in chrome)
here an example
http://codepen.io/gijsbotje/pen/OpQgXV
left is with max-width 100%, right is width 100%

martingara:
I encountered the distorted card image issue. I applied object-fit: contain; to .card-img-top and images appear in their correct ratio.

gijsbotje:
@martingara you are correct that it vissualy looks right but the image element itself is not in the correct ratio.
object fit makes the media fit inside the element instead of using the ratio of the media

mdo:
Closing for #22288. Thanks!

Issue #21868, opened 2017-01-28
Modal handleUpdate is protected


ashleydw:
From the docs:

> If the height of a modal changes while it is open, you should call $('#myModal').data('bs.modal').handleUpdate() to readjust the modal’s position in case a scrollbar appears.

In the source the method is protected:

https://github.com/twbs/bootstrap/blob/045888fa3887f5e65658499da11c8ad737222759/js/dist/modal.js#L394

Johann-S:
You can call call ``_handleUpdate`` by the following lines : 

```javascript
$('#myModal').modal('_handleUpdate')
// or
$('#myModal').data('bs.modal')._handleUpdate()
```

We should update the documentation

PR #21877, opened 2017-01-30
Fix Modal documentation about _handleUpdate method + move to public scope

Johann-S:
Update the way to call ``_handleUpdate`` method

IMO this method should move to the public scope

Close : #21868

### DIFF
--- a/docs/components/modal.md
+++ b/docs/components/modal.md
@@ -575,6 +575,12 @@ Manually hides a modal. **Returns to the caller before the modal has actually be
 
 {% highlight js %}$('#myModal').modal('hide'){% endhighlight %}
 
+#### `.modal('handleUpdate')`
+
+Manually readjust the modal's position if the height of a modal changes while it is open (i.e. in case a scrollbar appears).
+
+{% highlight js %}$('#myModal').modal('handleUpdate'){% endhighlight %}
+
 ### Events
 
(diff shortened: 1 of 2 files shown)


vanduynslagerp:
The function `_handleUpdate ` is called [automatically on event `resize`](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L299).

Doing something that make the sidebar to appear will necessarily change the window size and therefore trigger the `resize` event and call the function automatically.

I think it's not meant to be public and the mention of the function should be removed from the doc instead.

Unless there is a scenario in which a sidebar can appear without triggering the resize event ?

Johann-S:
Yes good catch I didn't see that 👍 

I'll update my PR to remove the mention of ``_handleUpdate`` thank you

Johann-S:
~~Yes good catch I didn't see that 👍
I'll update my PR to remove the mention of _handleUpdate thank you~~

It seems @vanduynslagerp sidebar can appear without triggering resize event for example if we do this : 
``$('#firefoxModal').find('.modal-content').css('height', '900px');``

That's what the documentation said : 
> If the height of a modal changes while it is open, you should call `$('#myModal').data('bs.modal').handleUpdate()` to readjust the modal's position in case a scrollbar appears.

vanduynslagerp:
Oh yeah right! Also if the content of the modal is updated it can change the height.
I don't know why I didn't think about that...

So yeah you're right the function should be public then.

PR #21895, opened 2017-02-02
add overflow hidden to dropdown menu

sbaechler:
This fixes the issue that when a dropdown has no vertical margins (`$dropdown-padding-y: 0`), the hover state has no rounded corners.

before

after


mdo:
Then anything with a tooltip or popover would be hidden. Sorry, not the right fix for this. I'd recommend adding the utilities or custom CSS to accomplish this on your end.

Issue #21530, opened 2017-01-04
V4 issue with .container-fluid inside navbar


eddywashere:
In Bootstrap v3 and up to v4-alpha.5, `.container-fluid` could be used inside of `.navbar`. Based on flexbox changes in master, that is no longer supported. 

Here's a codepen example with `.container` and `.container-fluid` inside `.navbar`: http://codepen.io/eddywashere/pen/MJWzVg

Codepen external css is pointing to [this commit](https://github.com/twbs/bootstrap/blob/1b194c06607f6cc3d087d25d6d49e1698771f3a1/dist/css/bootstrap.css).

There is some extra css commented out in the css tab. 1) I tried adding what is used for `.container` in `_navbar.scss` and  2) I typically add a max-width to container-fluid to prevent super wide content views.

eddywashere:
I think the issue is an expectation of being able to interchange .container with .container-fluid. My mental model is that .container is for fixed sizing within breakpoints vs .container-fluid for fluid sizing through all breakpoints. It's not super extraneous because it allows finer control over the nav inside navbar and the navbar background (wrapper). [Here's an example](https://reactstrap.github.io/) where setting a max-width on the .container-fluid helps control nav content.

I think it's also worth noting that I expected a simple nav like the one [here](https://reactstrap.github.io/) to work without a `.navbar-toggleable*` class since there is nothing to toggle. Had to dig through docs to find that out.

cuiliang:
Some times we need allow user  toggle between .container wrapped mode and full width mode.

If we add `div.container` wrapper around `navbar-nav` first and then we try to change to full width mode by removing `.container` class from the wrapper div or replace it with `.container-fluid`, the `navbar-brand` and `navbar-nav` will be splitted into multiple lines.

![image](https://cloud.githubusercontent.com/assets/1972649/21850815/5b3e810a-d847-11e6-92ee-cc7848d9dc82.png)

browner12:
I am also running into a strange navbar issue when I have it in a `.container`.  my `.navbar-brand` and my `.navbar-toggler` were both sitting in the middle of the page overlapping each other. this was due to the left and right `margin` set to 'auto'.  At the different breakpoints the `.container` gets a `width`.  Should the `.container` width be '100%' by default (xs size)?

if this should be a separate issue, let me know and I can break it off.

[...]

PR #21901, opened 2017-02-02
Fix for .navbar-toggleable-* > .container-fluid

zalog:
Problems:
- childrens inside `.navbar-toggleable-* > .container-fluid` are not staying on the same line. You can see this on bigger viewports than `*`.
- when `.container|.container-fluid > .navbar-toggler-*` left/right position is relative to `.container` and not `.navbar` itself. That means that the positioning will be doubled by the padding of the `.navbar`.

We need to set `display: flex` on `.container-fluid` and reset the `.container|.container-fluid > .navbar-toggler-*` position.

Please look at this example: http://codepen.io/zalog/pen/PWRvrJ

Later edit: In my example I've used `.container-fluid.w-100` and I think we need to set `.container-fluid` 100% width by default. Please let me know if so.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -116,7 +116,7 @@
 .navbar-toggler-left {
   position: absolute;
   left: $navbar-padding-x;
-  
+
   .container > &,
   .container-fluid > & {
     left: 0;
@@ -125,7 +125,7 @@
 .navbar-toggler-right {
   position: absolute;
   right: $navbar-padding-x;
-  
+
   .container > &,
   .container-fluid > & {
     right: 0;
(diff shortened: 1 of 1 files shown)


nextgenthemes:
This partly fixes https://github.com/twbs/bootstrap/issues/21530#issuecomment-280272925

zalog:
I made some small changes and I updated the description too.

nextgenthemes:
If this is the right way to do things, still feels weird to me but I am really new to flexbox, then I think this should have the 100% width come with this patch because that is what people expect. That w-100 class should not be needed IMO.

zalog:
Yep, I said so in my description. But for now, let's wait a bit :)

PR #21906, opened 2017-02-03
Allow numbers in breakpoint mixins

jonnitto:
At the moment the breakpoint mixins only allow the breakpoint names from bootstrap. With this small change, the mixins can also handle numbers (e.g. 350px)


mdo:
Thanks, but we'll pass on this.

Issue #21843, opened 2017-01-25
.btn-outline-secondary defaults to reboot color on :focus


psorensen:
There is no explicit color set for the active state of a the btn-outline-secondary focus state. It uses the color set by reboot.

![screen shot 2017-01-25 at 9 22 06 am](https://cloud.githubusercontent.com/assets/6152801/22301707/cf620c8e-e2e0-11e6-84b2-35d7a26b9f5e.png)

mdo:
Pretty sure this was addressed with #21633.

PR #21925, opened 2017-02-06
Button focus should apply same colours as hover

sethjeffery:
Fixes https://github.com/twbs/bootstrap/issues/21843

Since this change https://github.com/twbs/bootstrap/commit/a9bee8b6c8c31455dfba029434fad88b862c3846 buttons rendered with `<a>` have been inheriting their focus colour from reboot instead of overriding it with the button style.

### Before
image
image

### After
image

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -18,8 +18,11 @@
     background-color: $active-background;
     border-color: $active-border;
   }
+
   &:focus,
   &.focus {
+    color: $color;
+
     // Avoid using mixin so we can pass custom focus shadow properly
     @if $enable-shadows {
       box-shadow: $btn-box-shadow, 0 0 0 2px rgba($border, .5);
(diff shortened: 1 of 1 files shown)


tonifisler:
@sethjeffery With this change, the `<a class="btn">` is not behaving in the same way as all the other tags when focused:

**Example with all the possible tags with `:focus`**
screen shot 2017-02-06 at 15 26 43
http://v4-alpha.getbootstrap.com/components/buttons/#button-tags

We need to not change its background to keep it coherent.

sethjeffery:
Thanks for the check. Here's a replacement commit applying only text color change.

image

sethjeffery:
Though I must admit I have no idea how your effect managed to take hold.. (Different browser?)
In my original commit, the background color was being applied without a problem.

image

For reference, the original change was just swapping `@include hover` for `@include hover-focus`.

tonifisler:
You're right, I guess I had some remaining inline changes in DevTools... 🤔 Can't reproduce now.

Either way I think this solution is better! Keeping the same behavior as before.

mdo:
Pretty sure this isn't an issue anymore. See https://github.com/twbs/bootstrap/issues/21843#issuecomment-278208914.

sethjeffery:
Sorry my bad, I was comparing against alpha6 release, rather than against the HEAD.

PR #21935, opened 2017-02-07
change list group item action cursor to pointer

bardiarastin:


patrickhlauke:
care to explain why? we're trying to avoid explicitly setting `cursor`, as that should be taken care of by correct use of most appropriate html element (e.g. a link)

bardiarastin:
we use pointer cursor to tell the user that the item is clickable , in spa javascript frameworks using a tag makes trouble with Routers so i taught that we're using this class for most often for click stuff , then it would be nice if cursor was pointer by default; 
sorry for not providing details in description .

patrickhlauke:
if your action items act as links, but you can't use actual `<a href...>` elements, then i'd suggest adding extra `cursor:pointer` styles yourself to your additional CSS.

if the action items act as buttons (triggering something on the current page, rather than navigating to a new page / part of the current page, like buttons), then generally you want to keep the cursor as default - see reasoning behind https://github.com/twbs/bootstrap/pull/21812)

so for now, i don't think we'd want to force this in bootstrap core itself.

Issue #21936, opened 2017-02-07
Vertical text alignment of select.form-control text within .input-group is not centered


bdrelling:
**Operating System**: Mac OS Sierra 10.12.3
**Browser**: Chrome 55.0.2883.95 (64-bit)
**Bootstrap**: v4.0.0-alpha.6

**Issue Description**:

When using `select.form-control` within an `.input-group`, text is vertically aligned to the top of the box rather than vertically aligned to the middle.

**Screenshot**:

![screen shot 2017-02-07 at 11 04 51 am](https://cloud.githubusercontent.com/assets/3022693/22699423/626bc1ac-ed25-11e6-9d15-323100cd1829.png)

**Reduced Test Case**:

```
<div class="input-group">
  <span class="input-group-addon">@</span>
  <input type="text" class="form-control" placeholder="Username">
</div>

<div class="input-group">
  <span class="input-group-addon">@</span>
  <select class="form-control"><option>Username</option></select>
</div>
```

**Additional Comments**:

I'm unable to test in other browsers/operating systems at this time but can amend these results later. If someone can do so in the meantime, I'd appreciate it.

Also, any typical JS Bin-type site we're using for Bootstrap 4 right now? I didn't see one in the issues I peeked at but, if it exists, I want to be aware of it moving forward to make it easier for everyone.

mdo:
Selects aren't supported in input groups given their limited styling capabilities across all browsers. There are older duplicate issues that mention this as well.

bdrelling:
Ah, I think I left it open when searching through issues. That's good to know. Can we consider adding that into the doc as a warning on the input group page?

JohannesRudolph:
@bdrelling I'm just coming across the same issue, using `.custom-select` somewhat works for me: 

```
<div class="input-group">
  <span class="input-group-addon">@</span>
  <select class="form-control custom-select"><option>Username</option></select>
</div>
```

It's not perfect but a lot better than a plain select.

PR #21937, opened 2017-02-07
Vertical align in select

MohitKS5:
regarding bug "Vertical text alignment of select.form-control text within .input-group is not centered #21936"
added auto padding and vertical align for targetted tag and class


wolfy1339:
Hi, this pull request is not correct.
You are trying to make a change in the v3 source when the issue you are linking to says tjey are using v4 and not v3. Please use the code from the v4-dev branch instead to make a new PR.
For that reason this pull request will not get accepted and is invalid

PR #21938, opened 2017-02-07
Fix data-pause documentation for carousel

gnowland:
Carousel option data-pause expects boolean “false”. “null” causes js error!


Johann-S:
That's true we have an error with this message : ``CAROUSEL: Option "pause" provided type "null" but expected type "(string|boolean)"``

PR #21939, opened 2017-02-08
Removed xs from visibility classes

simPod:
We're removing xs from class names, so I removed it for visibility utilities.

### DIFF
--- a/docs/examples/dashboard/index.html
+++ b/docs/examples/dashboard/index.html
@@ -47,7 +47,7 @@
 
     <div class="container-fluid">
       <div class="row">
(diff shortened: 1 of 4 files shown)


mdo:
We'll need to do a bit more here to avoid the `.hidden-down` business—that should just be `.hidden` because `xs` is `0`.

simPod:
You are right. What about this one?

simPod:
But I wonder that it's a bit weird and non-intuitive. I would suppose that `.hidden` would hide the element for all screensizes and we would have a class for hidding the element just for `xs`, eg. `.hidden-xs`

mdo:
This is unneeded now since #22113 merged.

PR #21940, opened 2017-02-08
Fix fallback jquery filename

mourawaldson:
This is my first PR here, not sure if this is the correct place to fix, neither if I should regenerate the docs after the fix to include on this PR, cause I noticed all examples are tracked.

Please let me know if anything.


mdo:
You're all good, thanks!

PR #21953, opened 2017-02-09
Change 'has no affect' to 'has no effect' in Modal documentation

matteason:


patrickhlauke:
Indeed, good catch. thanks

PR #21957, opened 2017-02-10
Allow navbar links padding to be configurable.

### LINKED ISSUE #21956: Navbar nav-link hardcoded padding
There is general variable for padding of the `.nav-links`. However in navbar this padding is being hardcoded when `navbar-toggleable-*` is used. Specifically look at https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L155. No matter what variable is set for `nav-links` it will always be the same because of the heightened priority of CSS rules.

Proposal to fix this. Because navbar is a special type of navigation I suggest $navbar-link-padding variable to differentiate the padding from normal navs, which have different visual flow requirements.

That or copy the entire source file which seems a bit impractical. Parameters makes a lot more sense

coder4life:
Fixes #21956

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -153,8 +153,7 @@
           flex-direction: row;
 
           .nav-link {
-            padding-right: .5rem;
-            padding-left: .5rem;
+            $navbar-link-padding;
           }
         }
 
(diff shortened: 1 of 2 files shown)


mdo:
This isn't the right approach; you're zeroing out the vertical padding while setting horizontal padding here. Generally we do `-x` and `-y` variables and apply those as needed.

PR #21958, opened 2017-02-10
Do not assume margin direction, harder to override

coder4life:
`margin-bottom` is an opinionated way of handling margin spacing. We can remove the dependency and use simple CSS shorthand to achieve the same result. Variable is much more configurable.

### DIFF
--- a/scss/_type.scss
+++ b/scss/_type.scss
@@ -4,7 +4,7 @@
 
 h1, h2, h3, h4, h5, h6,
 .h1, .h2, .h3, .h4, .h5, .h6 {
-  margin-bottom: $headings-margin-bottom;
+  margin: $headings-margin;
   font-family: $headings-font-family;
   font-weight: $headings-font-weight;
   line-height: $headings-line-height;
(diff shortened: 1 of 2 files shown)


mdo:
> `margin-bottom` is an opinionated way of handling margin spacing.

Headings only have top and bottom margin by default, in all browsers. Given we nuke the top margin in Reboot, there's no need to set anything else.

> We can remove the dependency and use simple CSS shorthand to achieve the same result.

Simple is one way to look at it; another is that you're setting multiple unnecessary values most of the time.

> Variable is much more configurable.

That is true, but not the approach I take for Bootstrap's CSS. We only set what we need, not what you might need at some point.

Thanks though!

coder4life:
I would argue, why have variables at all. I get the "what we need" but does not give flexibility regarding choice. Regarding that mentality we should not be changing variables at all.

coder4life:
To comment further as a secondary thought, there logical conditions when being able to set the top and bottom margin makes sense with inline markup tags. Especially since under certain conditions you want to take advantage of margins overlapping.

PR #21970, opened 2017-02-11
Fix input box-shadow on focus state

pvdlg:
The variable `$input-box-shadow-focus` is currently defined as `$input-box-shadow, rgba($input-border-focus, .6) !default` that generate the invalid css `box-shadow: inset 0 1px 1px rgba(0, 0, 0, 0.075), rgba(92, 179, 253, 0.6);`

This PR add the missing `h-shadow`, `v-shadow` and `blur`. I set the default variable with the same value as the one defined for Bootstrap v3 [here](https://github.com/twbs/bootstrap/blob/master/less/mixins/forms.less#L60).


mdo:
Alternate fix coming in #22395.

PR #21972, opened 2017-02-11
Form states: set form-control state based on form-group and handle sm and lg size

pvdlg:
Per [documentation](https://v4-alpha.getbootstrap.com/components/forms/#how-it-works):
> To use, add `.has-warning`, `.has-danger`, or `.has-success` to the parent element. Any `.col-form-label`, `.form-control`, or custom form element will receive the validation styles.

Contrary to what's mentioned in the documentation the classes `form-control-success`, `form-control-warning` and `form-control-danger` are currently required on the `form-control`.

This PR conforms the code with the documentation and apply the state styles on the `form-control` based on the state class set on the parent `form-group`. The behavior of `form-control` is now similar to `form-control-label` (takes it's state styles from `form-group`).

In addition this PR set the position of the state icon based on the `form-control` size. So if `form-control-sm` is used with `has-success` the icon is now sized and positioned properly.


mdo:
No longer applicable since #22395.

PR #21975, opened 2017-02-12
Update _sizing.scss

dospuntocero:
breakpoints added for responsive width and heights

### DIFF
--- a/scss/utilities/_sizing.scss
+++ b/scss/utilities/_sizing.scss
@@ -1,12 +1,15 @@
 // responsive Width and Height properties
 
 @each $prop, $abbrev in (width: w, height: h) {
-  @each $bp in map-keys($grid-breakpoints) {
-    @each $size, $length in $sizes {
-      .#{$abbrev}-#{$size}-#{$bp} { #{$prop}: $length !important; }
+	@each $bpname,$bpsize in $grid-breakpoints{
+    @media (min-width: $bpsize) {
+		  @each $size, $length in $sizes {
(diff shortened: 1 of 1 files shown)


mdo:
Thanks, but I don't see us needing this right now.

RomainLanz:
May you reconsider this?

daviian:
Spacers include breakpoints as well, so IMHO width and height should also include that capability.

oliverdorn:
I would also appreciate this!

sebastianbrosch:
Would be very useful!

Sorgrum:
Would love this as well.

hypnocyst:
I'm surprised this isn't a feature. It should be.

bluecaret:
@mdo Why not? Why is this not needed? If your going to outright ignore something at least give a reason. These are useless without this. There are a lot of times I want column like sizing but just for a single element where the padding and margins on columns are inappropriate.

mauril26:
Did I miss something.? (No comments since 2017)
How are you doing to "solve" this situations.?

blackmonkey:
I just realized that we can use the col-* classes to fix this issue, e.g.

- w-lg-25 (width: 25% only on large+ devices): col-lg-3
- w-md-50 (width: 50% only on large+ devices): col-md-6
- w-sm-75 (...): col-sm-9
- w-100 (would be like w-xs-100): col-12
- width 25% except on large+ devices, and 16% on large+ devices: col-3 col-lg-2

PR #21980, opened 2017-02-13
Re-arranged contextual rows/cells

Jas-n:
Re-arranged contextual rows/cells so that example HTML matches example result.


mdo:
Thanks!

PR #21981, opened 2017-02-13
Use class name variables in selectors

ghost:
See comments and PR changes.

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -54,8 +54,8 @@ const Dropdown = (($) => {
     ROLE_MENU     : '[role="menu"]',
     ROLE_LISTBOX  : '[role="listbox"]',
     NAVBAR_NAV    : '.navbar-nav',
-    VISIBLE_ITEMS : '[role="menu"] li:not(.disabled) a, '
-                  + '[role="listbox"] li:not(.disabled) a'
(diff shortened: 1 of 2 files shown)


Johann-S (changes requested):
Please fix this build issue : https://travis-ci.org/twbs/bootstrap/jobs/201115160#L198-L199

DavisMiculis:
Okay, seems like tests now passed. Sorry for not checking correct linting.

Johann-S:
IMO you should check other plugins to see if you can find same issues in their (to add consistency to your PR)

DavisMiculis:
Seems like `[role="menu"]` and `[role="listbox"]` also can be modified to variables.  Need some input about `VISIBLE_ITEMS`.

DavisMiculis:
I did something similiar with tabs JS, but I need input if this is even "right" or "correct" ? Just seems strange for me to keep class name in variables, but then again use class names as strings.

bardiharborow:
@DavisMiculis, yep, this is correct from my point of view. Is this ready to merge?

DavisMiculis:
I will try to go over all JS files later today. Should I try to use variables as much as possible in selectors?

bardiharborow:
@DavisMiculis, either we won't be merging this, or we'll want them all done, so I'd suggest doing everything.

Johann-S (approved):

DavisMiculis:
Guess not.

PR #21988, opened 2017-02-14
Change .navbar-toggler color

zalog:
Change the `.navbar-toggler` to default contextual `.navbar-*-color` in case we use some other custom icons like `.navbar-toggler-icon.fa.fa-search`.
Please take a look at this demo: http://codepen.io/zalog/pen/MJLZOE/ - you must resize down the viewport.

Being on mobile, I'm not sure if we need a hover-focus mixin for `.navbar-toggler`, but if we do, please let me know.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -187,8 +187,7 @@
 
 // Dark links against a light background
 .navbar-light {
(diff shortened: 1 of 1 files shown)


mdo:
<3

PR #21994, opened 2017-02-15
Select uses wrong border variable for height calculation

### LINKED ISSUE #22011: Select uses wrong border variable for height calculation
To get `select`s matching the size of text inputs, a calculation is done which includes the width of the border. However, it uses `$border-width instead` of `$input-btn-border-width`, which means if the two don't match, your `select`s have the wrong height.

This and the fix can be seen in the demonstration at http://codepen.io/andrewbelcher/pen/rjbwox.

I've got a PR to fix it - #21994. Apologies if I shouldn't have also opened an issue - wondered whether issues were the better place for reporting.

andrewbelcher:
To get `select`s matching the size of text inputs, a calculation is done which includes the width of the border. However, it uses `$border-width instead` of `$input-btn-border-width`, which means if the two don't match, your `select`s have the wrong height.

Fixes #22011.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -66,7 +66,7 @@
 
 select.form-control {
   &:not([size]):not([multiple]) {
-    $select-border-width: ($border-width * 2);
+    $select-border-width: ($input-btn-border-width * 2);
     height: calc(#{$input-height} + #{$select-border-width});
   }
 


andrewbelcher:
Fixes #22011

mdo:
#22068 was merged as an alternate fix for this. Thanks!

PR #21997, opened 2017-02-15
Translucent .input-group-addon borders

danijelGombac:
Added background-clip: padding-box like .form-control


mdo:
Superseded by #22612. Thanks though!

PR #22004, opened 2017-02-16
normalize is no longer part of BS

nextgenthemes:
normalize is no longer part of BS


nextgenthemes:
Well since 3.0 is still active ... forget this.

Johann-S:
Support for v3 has mostly ceased (see : https://github.com/twbs/bootstrap/issues/20631)
So IMO you should open your PR

PR #22005, opened 2017-02-16
Uncomment BS Specific Styles

nextgenthemes:
Any reason this lines are not uncommented?


mdo:
Looking at the blame view for that file, it's intentional. Traces back to https://github.com/twbs/bootstrap/pull/19018 where we're trying to avoid messing up folks' styles given some user feedback. I imagine we left the commented out code in there as a reminder.

Issue #22012, opened 2017-02-16
Mixins in scss/_mixins.scss make importing into CodePen hard


andrewbelcher:
To demonstrate another issue, I thought I would use CodePen to write a reduced test case. CodePen can take SCSS and you can include SCSS files which get included before the code you add.

I needed to include `_forms.scss` which depends on a couple mixins, including `mixins/_forms.scss`. This depended on the `box-shadow` and `transition` mixins at the top of `_mixins.scss`, but as that has includes, it will fail to parse if included directly.

To make this easier, perhaps those 2 mixins should be moved to a separate file that is imported before the others, e.g. `mixins/_core.scss`?

PR #22013, opened 2017-02-16
Move mixins out of main _mixins.scss

andrewbelcher:
Possible solution to #22012


mdo:
Alternate fix coming in #22124 to avoid another mixins catch-all file (which is what `mixins.scss` is right now). Thanks though!

PR #22032, opened 2017-02-20
Remove margin-bottom from last .form-group

moso:
I like to have, eg, my submit/reset buttons inside `.form-group`s to group my form elements. But since every `.form-group` has `margin-bottom: $form-group-margin-bottom`, which is set to 1rem by default in v4 (and 15px in v3), it adds unwanted pixels in form of margin to the layout. See picture below where the unwanted pixels are marked with orange.

However, with a simple `&:last-of-type` that removes the `margin-bottom` on the last `.form-group`, one can get rid of the unwanted addition of pixels.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -188,6 +188,10 @@ select.form-control-lg {
 
 .form-group {
   margin-bottom: $form-group-margin-bottom;
+
+  &:last-of-type {
+    margin-bottom: 0;
+  }
 }
 
 .form-text {


moso:
As this contains "breaking" layout changes to even the docs, as the docs have buttons outside a `.form-group`, I might have been a bit optimistic. However, to me it makes sense to group all form elements - not just labels and input fields.

You even get the unwanted added pixels on the form-groups [docs page](https://v4-alpha.getbootstrap.com/components/forms/#form-groups) illustrated by the blue area below the "Another label/Another input" grouping, and thus not aligning the form within the padding of the `.bd-example`-box:

mdo:
This can be accomplished on an opt-in basis with `.mb-0`. Thanks though!

Issue #22002, opened 2017-02-15
[Firefox] -moz-appearance property has been removed


twbs-notify:
https://www.fxsitecompat.com/en-CA/docs/2017/moz-appearance-property-has-been-removed/

bardiharborow:
Wait no, we do actually use this over at [scss/_custom-forms.scss#L149](https://github.com/twbs/bootstrap/blob/02e0fdbd52df6ad00fe2c8ba2bb10a7f0a62f021/scss/_custom-forms.scss#L149).

janekh:
According to https://www.fxsitecompat.com/en-CA/docs/2017/moz-appearance-property-has-been-removed/ this change has been postponed. Currently Firefox 58 and Chrome 65 do not support appearance  and therefore .custom-select does not work as expected.

I can see that in https://getbootstrap.com/docs/4.0/components/forms/#select-menu it's working because the styles include both -moz-appearance and -webkit-appearance

Can this change be reversed until browsers start supporting it correctly?

wolfy1339:
@janekh The source code (SCSS) is not the same as the compiled CSS, in the build process Autoprefixer adds all the required vendor prefixes. Reverting the change is not necessary

juek:
With [Typesetter CMS](https://www.typesettercms.com) we use [scss.php](https://github.com/leafo/scssphp) to compile Bootstrap's SCSS from source, togehter with cusom variables and custom scss. But we do not use autoprefixer. Given that, we add the vendor prefixes to the source SCSS in order to make custom look O.K.

According to [caniuse.com](https://caniuse.com/#feat=css-appearance), appearance without vendor prefixes will not see the light of day any time soon (as of July 2018). `appearance: none;` currently won't work in any browser.

PR #22035, opened 2017-02-20
removing moz-appearance command which is replaced by webkit-appearance

MohitKS5:
Just removing moz-appearance statements to webkit-appearance. Since moz-appearance isnt in use


Johann-S:
``-moz-appearance`` used by Firefox see :  http://caniuse.com/#feat=css-appearance

wolfy1339:
Ref: #22002

bardiharborow:
Closing in favor of #22039.

PR #22043, opened 2017-02-21
Fix small typo in `Migrating to v4`

valery1707:


patrickhlauke:
Not a typo. Buttons, plural (as it refers to changing things for all buttons). Possessive of that is `buttons'`

PR #22068, opened 2017-02-25
Fix border width calculation for custom-select

### LINKED ISSUE #22011: Select uses wrong border variable for height calculation

rutger1140:
In addition to #22011 and #21994 also a fix for the custom-select border width. Now it respects the set variable. Fixes #22011.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -135,7 +135,7 @@
 .custom-select {
   display: inline-block;
   max-width: 100%;
-  $select-border-width: ($border-width * 2);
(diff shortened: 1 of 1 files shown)


mdo:
Thanks!

PR #22070, opened 2017-02-26
Add cursor: pointer to .navbar-toggler

hugmanrique:
Currently the toggler doesn't have the pointer icon as you would expect.


mdo:
I removed the cursor pointer from buttons in another #21812. Might end up restoring them, but this wouldn't be necessary if we did.

PR #22079, opened 2017-02-27
Fix carousel "load" event bound without any namespace

futpib:
I was trying to do something like this
```js
$(window)
	.off('load.bs.carousel.data-api')
	.on('load.bs.carousel.data-api', function () {
		// something custom
	});
```
And it did not work because the `'load'` event was bound without usual bootstrap carousel data-API namespaces.

### DIFF
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -227,7 +227,7 @@
     .on('click.bs.carousel.data-api', '[data-slide]', clickHandler)
     .on('click.bs.carousel.data-api', '[data-slide-to]', clickHandler)
 
-  $(window).on('load', function () {
+  $(window).on('load.bs.carousel.data-api', function () {
     $('[data-ride="carousel"]').each(function () {
       var $carousel = $(this)
       Plugin.call($carousel, $carousel.data())


Johann-S:
It's because Carousel do not have a load event, here Carousel plugin only listen to the load Event of the Window that's all

futpib:
@Johann-S Sorry, I don't really understand your comment. Yes, it's the `'load'` event on `window`. It is bound by bootstrap carousel as part of implementation of bootstrap data attribute API. All other data attribute API related events are namespaced accordingly, except this one. I think it should be namespaced as well. PR description provides motivational example.

Johann-S:
Only Bootstrap Events (events triggered by Bootstrap's Plugins) are prefixed.
 
Anyway support for v3 has mostly ceased. See : https://github.com/twbs/bootstrap/issues/20631

bardiharborow:
The `load` event is generated by the browser, not Bootstrap. Our code simply listens for it. It should not be namespaced. If you are simply trying to override that particular function, then you'll need to modify the Bootstrap code. I can't think of a good way of dynamically overriding it.

futpib:
Well, that sucks.

Weather event is generated by the browser or not says nothing about if it should be namespaced. `load.bs.scrollspy.data-api` is also namespaced and is also generated by the browser, same goes for `click.bs.collapse.data-api` and a bunch of other events. v4 also namespaces all events, including `load` and whatever, despite them being generated by the browser.

I can thing of a perfectly good way of dynamically overriding it, it's shown in the original post (the `off` + `on` thing). Actually v3 docs advertise that you can disable data API features by doing `.off('.bs.data-api')`, so one could argue that this is a bugfix (can't find the exact page that says that right now, since http://getbootstrap.com/ is down).

Issue #22073, opened 2017-02-26
Remove 'hover' from nav-tabs-active-link-hover-color et al


janaagaard75:
I seems to me that `hover` should be removed from these three variables, since the styles are applied to the tab, regardless of it being hovered or not. The has to has be active.

Current, v4-alpha.6, names:

    $nav-tabs-active-link-hover-color
    $nav-tabs-active-link-hover-bg
    $nav-tabs-active-link-hover-border-color

Proposed corrected names:

    $nav-tabs-active-link-color
    $nav-tabs-active-link-bg
    $nav-tabs-active-link-border-color

gettinToasty:
If no one else is working on this I can do it real quick.

mdo:
Fixed by #22087.

PR #22087, opened 2017-02-28
Remove 'hover' from nav-tabs-active-link-hover-color et al

gettinToasty:
Addressing the issue:

Remove 'hover' from nav-tabs-active-link-hover-color et al #22073


mdo:
Ayyyyy thank you!

mdo:
Going to see about getting CI to pass and then should be good to merge.

PR #22095, opened 2017-03-02
Remove margin-left on caret if empty

jesperkc:
If there is no text in dropdown button caret should not have margin left.


bastienmoulia:
Good work, I would like to have this change too!

I wonder if it would not be better to remove the `margin-left: 0.3em;` in the `.dropdown-toggle::after` and add a not empty selector?

```css
.dropdown-toggle:not(:empty)::after {
    margin-left: 0.3em;
}
```

PR #22113, opened 2017-03-03
Overhaul responsive, display, and visibility utilities

### LINKED ISSUE #21762: hidden class added by Modernizr causes blank page when combined with bootstrap
Relevant: https://github.com/Modernizr/Modernizr/issues/2147

Reproduce:
 make html page with bootstrap
 add Modernizr.js compiled with the following modernizr-config.json

```
{
  "minify": true,
  "options": [
    "setClasses"
  ],
  "feature-detects": [
    "test/dom/hidden",
  ]
}
```

suggested solution:

make https://github.com/twbs/bootstrap/blob/96d8553becfdf219c40e046e1202d4f2a2723010/scss/mixins/_visibility.scss#L4

`html .hidden `  instead of just `.hidden`

### LINKED ISSUE #21965: Rename responsive utilities to visibility utilities
I think the naming is a little confusing and could be clearer on their purpose by renaming it to be more literal on what they do. Someone could be searching the docs thinking, "how do I hide this div on mobile? Ahh visibility utilities."

mdo:
Bootstrap has a bit of a utility crisis right now. We've added a metric ton of new ones to v4 during our alphas, but we didn't really get back to cleaning things up that were already there. In particular, I'm looking at those `.hidden-` and `.visible-` classes, [their documentation](http://v4-alpha.getbootstrap.com/layout/responsive-utilities/), and how to use them with the [display utilities](http://v4-alpha.getbootstrap.com/utilities/display-property/).

As it stands now, each of these families of utilities convolutes things a bit. Responsive visibility utilities don't actually toggle `visibility`, there's overlap with the display utilities, and we've gone a bit too far into providing utilities for utilities.

This PR revamps things to hopefully clear things up and reduce some duplication in our code.

- **Drops `.hidden-*-down`.** If you need to hide something on the lowest viewport and then show it at a higher one, use two utilities. Example: `<div class="d-none d-md-flex"></div>`.

- **Drops `.hidden-*-up`.** This is the same as our responsive `display: none` utility. Instead of `.hidden-xs-up` or `.hidden-md-up`, use `.d-none` and `.d-md-none`.

- **Relocates and renames all `.hidden-*` utilities.** Instead of `.visible-*`, these classes start with `.d-` and now reside in the `_display.scss` file alongside our other `display` utilities.

- **Updates the `visibility()` mixin.** Previously this mixin was only for hiding content with `visibility: hidden`, but now we have two utilities—`.invisible` and `.visible`—that make use of this mixin with an added parameter. there is presently no default value for this mixin. Moved visibility utilities docs to new file, adding docs for newly added `.visible` class.

- **Revamps the display and responsive utilities docs.** We've renamed `display-property.md` to `display.md` and moved the newly renamed print utilities docs there. In addition, the `responsive-utilities.md` file is no more. Instead, we have a primer docs file that introduces folks to common layout-related utilities.

Okay, so that's a lot, but what's the savings? **We drop 50 lines of code in each CSS bundle.** Not a lot in the grand scheme, but this is a significant update to simplify an important area of our codebase.

Still to do:

- [x] Review examples to see if classes need updating
- [x] Clean up docs CSS to remove responsive test code
- [x] Write the new primer docs for the `Utilities for layout` page
- [x] Update migration guide

Also happens to fix #21965, and also fixes #21762.

/cc @cvrebert @bardiharborow


bardiharborow (approved):

Shakir-Ullah:
Awesome

JacobLett:
I think this is a great simplification. Thank you.

realappie:
Great work! I found the visibility utilities very confusing and hard to use.

bogdaniel:
hidden-*-up / down was a great feature i think .. easy to remember and to use =D like a shortcut for d-none d-flex combination

rromanchuk:
I just updated and tried migrating to the new `d-none`  and `d-*-block` and everything exploded because i wasn't thinking about inline vs block of the element. I'm still looking for the latest doc, but it would probably help people migrating over to make sure there to explicit examples of none, inline, and block. There is a little bit of burden on the migrator to make sure they know the existing display property on the element.

stefan505:
@rromanchuk I concur, busy migrating now (waited for Beta release) and its a complete mess for me. Example docs will be very useful.

dkrutsko:
I'm going through the migration process now and the lack of the `hidden-down` classes are a bit of a pain. The problem for me comes when I have to think about which version of `.d-*` to use. Before, I didn't have to think if the original element was `flex` or `inline`, etc. Now I have to know. Let's take this for example:

```html
<ul class="navbar-nav align-items-center hidden-md-down">
```

Should it be `d-none d-md-flex` or `d-none d-md-block`, It now depends on other class names or styles defined on that element where as before, you'd make it hidden and forget about it. Either way, as much as I think this change is great for simplification, it complicates other parts of code.

rogerlos:
It's a bit of overhead, but you could use a class like `d-md` and JS could apply the correct display class. You'd have to be sure to call the JS anytime your psuedo-class was likely to be added dynamically, of course.

```javascript
function d_any( sz ) {
    sz = typeof( sz ) === 'undefined' ? 
        [ 'sm','md','lg','xl' ] : 
        $.isArray( sz ) ? sz : [ sz ];
    $.each( sz, function( i, v ) {
        var $el = $( '.d-' + v ),
            disp = $el.removeClass('d-none d-' + v ).css('display');
        if ( $el.length ) {
            $el.addClass( 'd-none d-' + v + '-' + disp );
        }
    });
}
```
If you don't pass a size, it goes through all of them.

PR #22125, opened 2017-03-05
Clean up spacers a bit more

Starsam80:
I saw this when I looked at #22123, but didn't have enough time to write a review.


mdo:
Thanks!

Issue #22025, opened 2017-02-17
Detach accordions from .card


moso:
I'm having some fun with v4 alpha 6, however, is there a specific reason why the collapse groups, aka accordions, are hooked up on `.card`s?

Ideally, I would be able to make my own collapse grouping like this (simplified, but you will get my idea):
```html
<div id="accordion">
  <h3>
    <a data-toggle="collapse" data-parent="#accordion" href="#collapseOne" aria-expanded="true" aria-controls="collapseOne">
      Item One
    </a>
  </h3>
  <div id="collapseOne" class="collapse show">
    <ul class="list-unstyled">
      <li>Item</li>
      <li>Item</li>
      <li>Item</li>
      <li>Item</li>
    </ul>
  </div>

  <h3>
    <a data-toggle="collapse" data-parent="#accordion" href="#collapseTwo" aria-expanded="false" aria-controls="collapseTwo">
      Item Two
    </a>
  </h3>
  <div id="collapseTwo" class="collapse">
    <ul class="list-unstyled">
      <li>Item</li>
      <li>Item</li>
      <li>Item</li>
      <li>Item</li>
    </ul>
  </div>

  [...]

</div>
```
This would allow us to use collapse groupings on other than cards. The same thing was an annoying issue with 3.x where they are hooked up on panels.

mdo:
I'd be open to seeing this improved for v4 if we can muster it ahead of the next beta. It'd mean changing our JS and adding some additional styles, docs, examples, etc to do it right. At the very least we could add a second set of selectors there?

In the mean time, I think you can change the selectors and classes in our JS to work for things other than the cards.

/cc @bardiharborow and @Johann-S for ideas

Johann-S:
Maybe it's possible to get rid of Collapse's selectors by adding attributs.
For exemple : 
-  ``data-children`` allow to find the specific children of the Collapsible element. (``data-children=".card"``)
- ``data-parent`` allow to specify the root element of a collapsible element and will allow to have only one child active. (``data-parent="true"``)

And we will have a Collapse plugin more flexible than now.

If you want @mdo or @bardiharborow I can make a draft

mdo:
I can dig that approach @Johann-S. Those attributes would be separate from the card CSS, but about the transitions/collapse CSS? https://github.com/twbs/bootstrap/blob/v4-dev/scss/_transitions.scss#L10-L35. Does anything need to change there, too?

Johann-S:
[...]

PR #22138, opened 2017-03-09
Detach accordion from card

Johann-S:
- [x] Add unit test

/CC : @mdo @bardiharborow 

Closes : #22025

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -57,7 +57,8 @@ const Collapse = (($) => {
 
   const Selector = {
     ACTIVES     : '.card > .show, .card > .collapsing',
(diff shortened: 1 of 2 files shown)


Johann-S:
For me everything is done here @mdo except the documentation

Johann-S:
Only JavaScript here @bardiharborow no CSS for the labels

mdo:
> Only JavaScript here @bardiharborow no CSS for the labels

Whenever something affects or is related to the CSS (in this case the `.collapse` classes), we generally apply the CSS label :). This does need the JS one, too, though!

mdo:
I'll take a stab at adding some docs examples here shortly.

PR #22143, opened 2017-03-09
Fix selectable disabled toggle radio buttons

jimlevett:
A disabled radio toggle button is not selectable
Disabled either by
-disabled class on Label container
-input with a disabled attribute


jimlevett:
Ah, wrong branch - sorry meant to raise against the master. I'll try again.

PR #22144, opened 2017-03-09
changed justified-nav example to look more like bootstrap-4

lemuelbarango:
The [Justified Nav Example Page](https://v4-alpha.getbootstrap.com/examples/justified-nav/) still has the feel of bootstrap-3. I changed it to be along the lines of bootstrap-4


lemuelbarango:
I also updated the screenshots on the example page

mdo:
Screenshots aren't the same quality as the existing ones, despite being out of date. I'll take care of that when we're closer to the beta. The justified nav can also stay the way it is now for general looks; it does need some navbar toggler fixing, though.

PR #22151, opened 2017-03-10
Update scss mixins to comply with scss-linting rules

naicko:
Updated some mixin files to comply with the linting rules.


scss/mixins/_hover.scss
```diff
@@ -8,16 +8,15 @@
 //    }
 //  }
 //  @else {
-    &:hover { @content }
```
mdo: This one doesn't need to change. Or we just remove all the commented out stuff here; it's been brought up a few times now.

PR #22154, opened 2017-03-10
Update _screen-reader.scss

ffoodd:
Small improvements that could be important:
* `clip` [is deprecated](https://www.w3.org/TR/css-masking-1/#clip-property). Adding `clip-path` as progressive enhancement; the shorter notation came from @ryuran 's [suggestion](https://twitter.com/ryuran78/status/778943389819604992);
* [J. Renée Beach warned about single pixel with interfering with screen readers vocalisation](https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe#.vcd5xlpgg) solved with `white-space`.

See [the detailed post on Hugo Giraudel's blog](http://hugogiraudel.com/2016/10/13/css-hide-and-seek/).

Also kinda related to issue #20732 :)

Please let me know if you find any trouble with this technique. Thanks a lot!

### DIFF
--- a/scss/mixins/_screen-reader.scss
+++ b/scss/mixins/_screen-reader.scss
@@ -9,7 +9,7 @@
   height: 1px;
   padding: 0;
   margin: -1px;
-  overflow: hidden;  
+  overflow: hidden;
   clip: rect(0,0,0,0);
   white-space: nowrap;
   clip-path: inset(50%);
@@ -29,7 +29,7 @@
     width: auto;
     height: auto;
     margin: 0;
-    overflow: visible;    
+    overflow: visible; 
     clip: auto;
     white-space: normal;
     clip-path: none;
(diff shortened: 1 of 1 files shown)


scss/mixins/_screen-reader.scss
```diff
@@ -10,6 +11,8 @@
   margin: -1px;
```
mdo: Can we drop the `margin: -1px`? I don't see it in Hugo's post and I know we had an issue with it on GitHub causing a horizontal scrollbar with that.

ffoodd:
Just removed the negative margin, which I forgot to remove before requesting pull  (been testing multiple versions of this helper).

I can't find any issue related to this but I'd be interested (I'll comment again if I find it).

Thanks for reviewing!

PR #22164, opened 2017-03-11
Use btn-$variant-color variables for outlined btns

intellix:
Despite now having the ability to override outline-variant color, it's not actually being used so each are still #fff

I've set $btn-primary-color to #000 but noticed it was still #fff for outlined buttons because the var isn't passed into @mixin button-outline-variant

Related: twbs#20734

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -73,22 +73,22 @@ fieldset[disabled] a.btn {
 
 // Remove all backgrounds
 .btn-outline-primary {
-  @include button-outline-variant($btn-primary-bg);
+  @include button-outline-variant($btn-primary-bg, $btn-primary-color);
 }
 .btn-outline-secondary {
-  @include button-outline-variant($btn-secondary-border);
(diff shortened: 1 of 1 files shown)


beausmith:
👍 Nice @intellix! I had forgotten about this. 

I recall that I hesitated in committing this change because it does have one visual change… and so I committed the smallest change with the goal of getting it merged, then coming back to this… but I got distracted. So, I'm glad you've made this commit!

The hover color for the secondary outline button is currently white, this commit changes it to the `$gray-dark` color. Arguably, this change of color makes the behavior more consistent to the default button normal and hover states (see below).

## Outline buttons

Default:
image

**Hover** (before change):
image

**Hover** (after change):
image

While this does contain a visual change… it makes the secondary outline button hover state more visually similar to that of the default secondary button, adding more contrast to the text color vs background color:

## Buttons

Default:
image

Hover:
image

I approve this change… however it may have minor affect on projects when they upgrade to use this code.

PR #22199, opened 2017-03-16
Update _reboot.scss

infinnie:
The font-family and line-height CSS properties, I think, should be applied to not only the body, but the form elements as well.


mdo:
This isn't necessary. There are changes in `v4-dev` to fix the input font-family issue.

Issue #21585, opened 2017-01-07
[v4 css] inverted table styling


Mottie:
When inverting a table style using the `table-inverse` class ([demo](http://codepen.io/Mottie/pen/OWVgYZ)):

1. **Striping**<br>
   With a "table-striped" class, the stripes are not discernibly different. In the css, the definition is the same as for the non-inverted row styling:

   ```css
   .table-striped tbody tr:nth-of-type(odd) {
     background-color: rgba(0, 0, 0, .05); /* $table-bg-accent; */
   }
   ```

2. Borders<br>
   The border styling of the non-inverted tables is a very subtle, using a border color of `#eceeef`. On inverted tables, we see `#fff`! It's way to bright!

   ```css
   .table-inverse td,
   .table-inverse th,
   .table-inverse thead th {
     border-color: #fff; /* $body-bg */
   }
   ```

3. Hover<br>
   Hovering isn't as bad when no stripe is applied to the table; but when a stripe is applied, the hover is only slightly more discernible than the row color ([demo](http://codepen.io/Mottie/pen/vgOJOW))!

   ```css
   .table-hover tbody tr:hover {
     background-color: rgba(0, 0, 0, .075) /* $table-bg-hover */
   }
   ```

This may not seem like a big deal to some of you young 'uns out there, but for elderly and visually impaired users, it would make a difference.

LarryAzevedo:
I agree with @Mottie.

**Regarding the striping**, maybe we should change `$table-bg-accent: rgba($black,.05)` to `$table-bg-accent: rgba($gray-light,.075)`.

**Regarding the borders**, maybe we need a new variable called `$table-inverse-border-color` and set it's default value to `lighten($table-inverse-bg, 15%)`.

**Regarding the hover**, maybe we should change `$table-bg-hover: rgba($black,.075)` to `$table-bg-hover: rgba($gray-light,.075)`.

I tested these new settings and it looked good (on two completely different monitors) but would appreciate your feedback.  I also experimented with changing `$table-inverse-bg` to different values and that worked out well too.

So to recap, I'm proposing the following changes in _variables.scss (under the `$table-inverse-color` variable):
```
$table-inverse-border-color:    lighten($table-inverse-bg, 15%) !default;

$table-bg-accent:               rgba($gray-light,.075) !default;
$table-bg-hover:                rgba($gray-light,.075) !default;
```

And the following changes in _tables.scss:
```
.table-inverse {
  color: $table-inverse-color;
[...]

PR #22217, opened 2017-03-20
v4 table updates

### LINKED ISSUE #21577: .table-responsive not expanding to its parent
I apologize for dragging a closed issue (#18612) back into the light, but I'm not sure the responsive table is behaving as it should.

As of v4.0.0-alpha.6, the problem seems to remain:
- tables with the `.table` class are 100% of the width of their parent
- tables with `.table` + `.table-responsive` are only as wide as their content
- tables with the class `.table` that are also wrapped with `<div class="table-responsive">` are %100 of the width of their parent

Here's a fiddle:

https://jsfiddle.net/45tbox4f/

Example 2 and 3 are both scroll horizontally, but the non-wrapped table doesn't expand to its parent as @aimeos pointed out. I believe the example in the docs appears to work because its content is already quite wide(?)

(Tested on the latest Safari, Chrome OSX, FF OSX, Chrome Win and FF Win)

Please feel free to berate me if I got this wrong!

mdo:
Few table updates:

- Fixes the inverse table's border color. Adds a new variable that lightens the inverse table's background color a bit instead of using `$body-bg`.
- Fixes the responsive table behavior. Instead of using a parent class again, this scopes (as the comment above it indicates) the styles to under 768px.

---

Fixes part of #21585
Fixes #21577 (/cc @pmsaue0 as credit for the fix)
Closes #21677, closes #21678, and closes #21676 as they're no longer needed.

### DIFF
--- a/docs/assets/scss/_content.scss
+++ b/docs/assets/scss/_content.scss
@@ -6,11 +6,15 @@
 
 .bd-content {
   > table {
-    display: block;
     width: 100%;
     max-width: 100%;
     margin-bottom: 1rem;
-    overflow-y: auto;
+
+    @include media-breakpoint-down(md) {
+      display: block;
+      overflow-x: auto;
+      -ms-overflow-style: -ms-autohiding-scrollbar; // See https://github.com/twbs/bootstrap/pull/10057
+    }
 
     // Cells
     > thead,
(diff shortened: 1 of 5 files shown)


mdo:
Gave up on CI lol.

PR #22229, opened 2017-03-20
Add missing flex-fill utility

pvdlg:
Add the `.flex-{bp}-fill` utility classes that are used [here](https://v4-alpha.getbootstrap.com/components/navs/#working-with-flex-utilities) but were missing


mdo:
Continuing in #25385. Thanks!

PR #22230, opened 2017-03-21
Navbar fixes

### LINKED ISSUE #21530: V4 issue with .container-fluid inside navbar

### LINKED ISSUE #21604: Responsive navbar is 2px shorter when content is hidden.

### LINKED ISSUE #21659: [css] [v4] Wrapping navbar content with a container adds extra margin to collapse button
- Browser and version: Firefox 50.0.2, Chrome 55.0.2883.87
- JS Bin: https://jsbin.com/penemipaze/edit?html,css,output https://output.jsbin.com/penemipaze

I'm wrapping the content of the navbar with a container and it's adding an extra right margin on the collapse button when aligned to the right.

The navbar-toggler-right class adds a right margin of 1rem but it seems this is not necessary when wrapping the content with a container.

### LINKED ISSUE #21691: Navbar 'md' triggers at 992 instead of 768

### LINKED ISSUE #21705: 4.0.0.alpha6: .navbar-text should have the same vertical padding as .nav-link

### LINKED ISSUE #21706: 4.0.0-alpha.6: Navbar should have style of .navbar-toggleable-xl by default

### LINKED ISSUE #21735: Navbar containers lose padding too soon
When the navbar turns toggleable the brand is 15px left of the content.

See http://jsfiddle.net/2t6e0ryh/

image

image

### LINKED ISSUE #21846: Navbar w/o navbar-brand breaks with navbar-toggler-right(-left)
When you create a navbar with only a `navbar-nav` component (no `navbar-brand`), and then use `navbar-toggler-right` or `navbar-toggler-left`, the entire navbar height collapses when the toggler is shown. 

http://www.codeply.com/go/smB7eC4qYc

A workaround I've found is to use an `navbar-text` placeholder...

```
<nav class="navbar navbar-toggleable-md navbar-light bg-faded">
  <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarToggler1" aria-controls="navbarToggler1" aria-expanded="false" aria-label="Toggle navigation">
    <span class="navbar-toggler-icon">&nbsp;</span>
  </button>
  <span class="navbar-text">&nbsp;</span>
  <div class="collapse navbar-collapse" id="navbarToggler1">
    <ul class="navbar-nav mr-auto mt-2 mt-md-0">
      <li class="nav-item active">
        <a class="nav-link" href="#">Home <span class="sr-only">(current)</span></a>
      </li>
      <li class="nav-item">
        <a class="nav-link" href="#">Link</a>
      </li>
      <li class="nav-item">
        <a class="nav-link disabled" href="#">Disabled</a>
      </li>
    </ul>
    <form class="form-inline my-2 my-lg-0">
      <input class="form-control mr-sm-2" type="text" placeholder="Search">
      <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
    </form>
  </div>
</nav>
```

### LINKED ISSUE #21956: Navbar nav-link hardcoded padding

### LINKED ISSUE #21984: .navbar-toggler not aligned same as .navbar-collapse if .navbar-brand causes whole navbar element to expand beyond "default" height
Not sure if this is intended or not but when .navbar-brand element causes the whole navbar element to expand beyond it's "default" height, the elements in the .navbar-collapse element are vertically center aligned but the .navbar-toggler is not.

links vertically aligned
navbar toggler absolute positioned top right corner

Screengrabs taken from official docs (editing the font-size of the .navbar-brand element in inspection tools)

mdo:
This PR builds on @vanduynslagerp's work in #21719 as a basis for many of [our navbar related issues](https://github.com/twbs/bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue%20milestone%3Av4.0.0-beta%20navbar%20). :sweat_smile: Basically there's a bug in a Sass mixin, some unnecessary `position: absolute`, a missing `.container-fluid`, and some inconsistent sizing of navbar sub-components. That's all getting fixed here.

- **Correct navbar collapsing breakpoints** resulting from the broken Sass mixin that was generating our `.navbar-toggleable` classes. Basically this wasn't generating mobile-first classes, so we were acting on the wrong media query values. Fixes #21691, fixes #21706, and closes #21719.

  As part of this, `.navbar-toggleable` classes have been renamed to `.navbar-expand` to match the mobile-first approach used elsewhere.

- **Redo the `.navbar-toggler` alignment by dropping the modifiers entirely.** Instead of using absolute position, we run with the DOM order and make use of `justify-content: space-between`, forcing navbar brand and toggler to be on opposite ends of the navbar. This drops `.navbar-toggler-left` and `.navbar-toggler-right` Fixes #21659, fixes #21735, fixes #21846, fixes #21984, and closes #21729.

- **Resolve issues with `.container-fluid`** not being fully supported within `.navbar`. Fixes #21530 and closes #21901.

- **Normalizes the sizing of navbar brand, text, and link elements** by pulling in the changes from #21721 (which fixes #21705, fixes #21956, and fixes #21604). *This one might still need further love, or our inputs will; the text input and buttons are `38px` vs the nav's `40px`. Any changes to that will come in another PR.*

Still to do:

- [x] Update navbar doc examples to drop toggler modifiers and reorder brand/toggler
- [x] Drop navbar toggler variants in docs


vanduynslagerp:
Awesome ! The `justify-content: space-between` for `navbar-toggler` is really clever !

mdo:
Unsure where these test failures came from for our button JS... presently looking into that before tackling the navbar docs update.

vanduynslagerp:
I think this is due to a JQuery bug in 3.2.0 (updated in this PR #22218)

The bug is fixed in [JQuery 3.2.1](http://blog.jquery.com/2017/03/20/jquery-3-2-1-now-available/):  jquery/jquery#3579

Upgrading to JQuery 3.2.1 should fix the test.

mdo:
Oof, fixed that but still have the Qunit failures with Sauce. The build from merging #21788 (https://travis-ci.org/twbs/bootstrap/jobs/212589569) is the earliest I see of the failure.

vanduynslagerp:
@mdo, it should be fixed here #22234. Although I can't run the Saucelabs tests (apparently they run only on a push and not a PR ?). So let me know if there is other Saucelabs errors and I'll fix them ASAP.

mdo:
The last thing, outside these failures, is possibly renaming this class. The class name `.navbar-toggleable-` mostly comes from legacy Bootstrap versions where we weren't exactly thinking mobile-first on everything we did. Often times the navbar of a site or app is designed for desktop first still, so folks think "okay when does this need to collapse." On the flip side, if we start with mobile, it's when does this expand. 

I could go either way, but I'm kinda inclined to leave it as-is for few changes from our last alpha to this first beta. That said, once we go beta, we're locked in. Thoughts?

zalog:
@mdo, I think `.navbar > .container-fluid` need to be always `width: 100%`.
Please take a look here: http://codepen.io/zalog/pen/RpMoqY

mdo:
@zalog Hmm, yeah, I see that now, too. Added a fluid container example to my growing js bin examples at https://jsbin.com/zixoxah/edit?output. I feel like that `width: 100%` should probably just go on `.container-fluid` and not be limited to the `.navbar`. Thoughts?

zalog:
Yep, I was thinking on that too, I can help with a PR.
Love this changes, great work!

vanduynslagerp:
> I could go either way, but I'm kinda inclined to leave it as-is for few changes from our last alpha to this first beta. That said, once we go beta, we're locked in. Thoughts?

I would personally favor a rename for consistency reasons. Thinking in a "responsive" way is really not that easy, especially for "beginners", so having every classes working the same way (i.e. "When viewport is higher than breakpoint apply this") make things a lot simpler. Plus I saw several questions recently around that recently.

Looking at the code it seems that currently the `.navbar-toggleable-` is the only class in that situation.

mdo:
Working on a new class name, `.navbar-expand`, in https://jsbin.com/rikevuj/1/edit?output. In this setup, the classes work like so:

- No modifier, just `.navbar`, is _never_ expanded
- `.navbar-expand` is _always_ expanded, from `0` to infinity (and beyond!)
- `.navbar-expand-sm` expands at `sm` and up
- `.navbar-expand-md` expands at `md` and up
- `.navbar-expand-lg` expands at `lg` and up
- `.navbar-expand-xl` expands at `xl` and up

Put another way, the navbar is always collapsed until you tell it otherwise. The first several navbars in that example link above show navbars in that order. I think this should work for folks who have been scratching their heads at this component being in _reverse_ of our general mobile-first approach everywhere else.

Would love feedback before I push up relevant changes for that.

vanduynslagerp:
Perfect !

nextgenthemes:
@mdo on your jsbin the very first navber the dropdown menu causes navbar height to expand with it. https://jsbin.com/rikevuj/1/edit?output

PR #22231, opened 2017-03-21
Revert "Update dependencies."

mdo:
This reverts commit 09fb2b9af3ee7ef0e6a23e913ac74967d4a08d24 so tests can pass again.

/cc @bardiharborow

PR #22232, opened 2017-03-21
Bump to jQuery 3.2.1

mdo:
Hotfix for jQuery when we updated to v3.2.0 in #22218. Should hopefully negate #22231.


mdo:
Fixes the qunit tests, but keeps the ever constant Sauce Labs failures :sob:

mdo:
Near as I can tell, Qunit started failing after merging #21788. Merging this PR for now to at least fix the other failures.

PR #22238, opened 2017-03-22
Visual example does not use this class

tag:
The `.bs-example` version has a visible label, and does not use the `.sr-only` class


mdo:
Dang it, just realized this was for v3.

mdo:
Just reverted this manually and force pushed `v3-dev` branch to keep it where it is.

Issue #22249, opened 2017-03-23
v4: Cannot use input of a popover inside modal


nunofilipesf:
I've tried to create a modal with a popover inside. 
This popover has html content with an _input_ and a _select_.

If i open the modal and the popover, i can use the _select_, but not the _input_. 
Also, the focus is not on the _input_.

Sample here: https://jsfiddle.net/bpx3zsyz/2/

Tested in Chrome, Edge and Firefox, always with the same behaviour

gsferreira:
:+1:

satyajit11:
remove  tabindex="-1"  of your modal it'll work.put the below line without tabindex

<div class="modal fade" id="exampleModalLong" role="dialog" aria-labelledby="exampleModalLongTitle" aria-hidden="true">

here is example fiddle
[jsfiddle](https://jsfiddle.net/satya6565/w8yh65cr/1/)

Johann-S:
IMO @satyajit11 give a correct answer (see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

I'm not sure it's a Bootstrap bug here, any advice @mdo or @bardiharborow ?

mdo:
Our docs include `tabindex="-1"`—do we need to update to fix that on our end, or is there JS to tackle?

satyajit11:
@Johann-S  when add popover code is append to body which is outside modal because we already have tabindex="-1"  so focus won't go there.I don't proper js solution but i think we need to append inside the container or remove focus from modal when we add popover

nunofilipesf:
Removing the tabindex from the modal solved my problem.

Can i close the issue or there is here a bug in Bootstrap to solve?

Johann-S:
As @mdo said we should update our documentation and remove ``tabindex="-1"`` reference.
@nunofilipesf we have to update our documentation

patrickhlauke:
removing the `tabindex=-1` on the modal is NOT the solution. or rather: once you remove this, the modal cannot be programmatically focused by javascript when it's opened, meaning that accessibility for the modal is broken (screen reader doesn't announce there's a modal, doesn't read the title of the modal).

i'd suggest the problem is in the popup and the way it's generated, rather than the modal itself.

Johann-S:
When #22263 will be merged @nunofilipesf  you'll have to update your code to the following : 
```javascript
    $(document).ready(function () {
    	var popoverContent = '<div> \
[...]

PR #22260, opened 2017-03-24
Remove tabindex in modal documentation and visual test

Johann-S:
Remove tabindex in modal documentation and visual test

/CC : @mdo and maybe @patrickhlauke for the accessibility

Close : #22249


patrickhlauke:
see https://github.com/twbs/bootstrap/issues/22249#issuecomment-289026409 i'd strongly suggest NOT to go ahead with removing the `tabindex`

Johann-S:
Thank you @patrickhlauke I'll look for another solution

Issue #21962, opened 2017-02-10
Tooltip not appends to given element


Liad91:
Bootstrap 4 alpha 6.
I'm trying to append tooltip to element by using the container option, but it's not working.
[JSFiddle](https://jsfiddle.net/Liad91/Lab4wsvu)

echubs:
Same issue here. From what I see, tooltip [gets added](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L286) to the container first, but then tether moves it back to the body [here](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L292). I found that tether scans container and all its parent nodes up to the body, if it finds an element with a **non-static position** - it moves a tooltip ([see here](https://github.com/HubSpot/tether/blob/master/src/js/tether.js#L768)). In reality tooltip always ends up in the body, cause grid classes like .col-12 use relative position.

Liad91:
I found that tether have `bodyElement` option which can be used to specify which element the tether's should attach themselves to. The problem is that tether currently makes the assumption that all of its calculations must be relative to the body, So even if we will use this option the container element will have to be same size and position respect to the body.
Maybe [Popper.js](https://github.com/FezVrasta/popper.js) is more suitable than Tether.js?

Liad91:
#22444

Johann-S:
Will be fixed by https://github.com/twbs/bootstrap/pull/22444

Johann-S:
Closed and fixed by #22444

PR #22263, opened 2017-03-24
Implement bodyElement for Tooltip and Popover

### LINKED ISSUE #22261: Implement bodyElement in Tooltips and Popover
Tether added ``bodyElement`` option in **1.4.0** (see https://github.com/HubSpot/tether/pull/204) and it would be better to use ``bodyElement`` instead of trying to do the same thing with ``container`` option 
in Tooltip (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L284-L288) which doesn't work because Tether previous version only add element to the ``<body>``

It would be usefull for #22249

Johann-S:
- [x] Implements ``bodyElement``
- [x] Update Tether to 1.4.0

With this PR, folks will be able to specify body element of their popover or tooltips but as it said here https://github.com/HubSpot/tether/pull/204 

> It would be up to the developers to ensure that whatever element is considered the bodyElement will have the correct characteristics for positioning (same size and position w/ respect to the body).

/CC @mdo , @bardiharborow 

Close #22261, #22249, #21962


FezVrasta:
This PR isn't going to fix the problem as explained [here](https://github.com/twbs/bootstrap/issues/22249#issuecomment-291962034)

Johann-S:
For now not a good solution

PR #22265, opened 2017-03-24
[Docs v3] Typo in Modal Example

LKaemmerling:
Hello,

i found an "typo" in the Docs for v3.3.7.

In the example is the class: "bs-example-modal-lg" (and "bs-example-modal-sm") used. If you copy & paste this example, it doesn't work. The right classes are "modal-lg" and "modal-sm".


LKDevelopment:
idk why the travis build fails? I doesn't change anything on the gemfile?

mdo:

---

_This is a saved reply._

LKDevelopment:
@mdo that is a doc improvement, because if you copy & paste the example actually it doesn't work.

mdo:
Yup, but only on a case by case basis. I'm basically not making any v3 changes until v4 ships. Sorry, just trying to stay focused and ship an update. <3

PR #22272, opened 2017-03-25
Collapse - do not prevent event for input

### LINKED ISSUE #21079: Collapse prevents radio button update
When using `data-toggle="collapse"` and `data-target` on a radio button, the clicked radio button is not checked. See this minimal fiddle: https://jsfiddle.net/p0xkotyr/1/. Reproducible for me in Chrome (54.0), Safari (9.1.2) and Firefox (49.0.2) on OS X 10.11.6.

I didn't find any current open issue on this, just a couple of rather old closed ones:

* https://github.com/twbs/bootstrap/issues/1060
* https://github.com/twbs/bootstrap/issues/3857

Thanks for taking a look!

### LINKED ISSUE #22049: data-toggle=collapse with a checkbox to trigger the toggle
Using this fiddle as the testcase:

[http://jsfiddle.net/jcwchriswalker/e8wwhqcx/](http://jsfiddle.net/jcwchriswalker/e8wwhqcx/)

Note that by checking on/off the checkbox will hide and show the element targeted by the data-target. However the checkbox will not show checked/unchecked state. I looked through the documentation concerning what elements support the data-toggle attribute but could not find any definitive list. It seems that the checkbox's event handler set a preventDefault() keeping the checkbox from setting on/off but showing/hiding the data-target.

Johann-S:
Stop preventing event for input in Collapse

/CC @mdo and @bardiharborow 

Fixes #22049, fixes #21079

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -363,7 +363,9 @@ const Collapse = (($) => {
    */
 
   $(document).on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-    event.preventDefault()
+    if (/input|textarea/i.test(event.target.tagName)) {
+      event.preventDefault()
+    }
 
     const target = Collapse._getTargetFromElement(this)
     const data   = $(target).data(DATA_KEY)
--- a/js/tests/unit/collapse.js
+++ b/js/tests/unit/collapse.js
@@ -513,4 +513,20 @@ $(function () {
     })
     $target.trigger($.Event('click'))
   })
+
+  QUnit.test('should not prevent event for input', function (assert) {
+    assert.expect(2)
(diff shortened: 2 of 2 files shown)


mdo:
Seems fine by me!

js/src/collapse.js
```diff
@@ -363,7 +363,9 @@ const Collapse = (($) => {
    */
 
   $(document).on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-    event.preventDefault()
+    if (!$(event.target).is('input')) {
```
Johann-S: Yes thank you @vanduynslagerp ! That's the regex I looked for but didn't found it 😄

Issue #22259, opened 2017-03-24
v4: dropdown menu text with .navbar-inverse.bg-inverse


hypeJunction:
I have a dropdown down menu in an inversed navbar and the text of the dropdown menu comes up white/transparent.

```.navbar-inverse .navbar-nav .nav-link``` this should probably be less greedy.

mdo:
Bug reports must include a live demo of the problem. Per our contributing guidelines, please create a reduced test case via [JS Bin](https://jsbinc.om) and report back with your link, Bootstrap version, and specific browser and OS details.

---

_This is a saved reply._

hypeJunction:
http://codepen.io/hypeJunction/pen/KWBwdg

oktalk:
You've improperly formatted your dropdown. In your example you need to remove the class '.nav-link' and replace it with the class `.dropdown-item`.

These questions are better suited for Stackoverflow.

hypeJunction:
I abandoned the project. Too much time investment to rewrite an entire interface. Thanks for getting back to me though

PR #22273, opened 2017-03-25
dropdown menu issue

satyajit11:
if we use dropdown in navbar position:static; making problem.issue has been reported here

https://github.com/twbs/bootstrap/issues/22259


mdo:
This isn't the right fix for this issue; the menus are static intentionally on small displays.

Issue #21146, opened 2016-11-16
different tooltips offset when hovering first time and second times


maxlibin:
When hovering a tooltip item: http://v4-alpha.getbootstrap.com/components/tooltips/
the first time offset will be different from other times, any fixes?

![untitled-2](https://cloud.githubusercontent.com/assets/1277038/20340784/30f09654-ac1e-11e6-9154-433e911af823.jpg)

jesusmtzarvizu:
Could you please paste the code from the first and second case to see it??

maxlibin:
The screenshot is taken from official site using chrome on macOS

bardiharborow:
Confirmed on Mac Chrome 55.

Johann-S:
Confirmed on Windows 10 Chrome 54

ArunHub:
confirmed on windows firefox developer edition 51.0a2

bardiharborow:
I used `setInterval(function() { console.log($('.bs-tether-element')[0].outerHTML) }, 1000);` to pull the HTML for the tooltips.

First activation:

```html
<div class="tooltip fade bs-tether-element bs-tether-enabled bs-tether-abutted bs-tether-abutted-top bs-tether-element-attached-bottom bs-tether-element-attached-center bs-tether-target-attached-top bs-tether-target-attached-center in" role="tooltip" id="tooltip258569" style="left: 0px; position: fixed; transform: translateX(208px) translateY(-140px) translateZ(0px); bottom: 0px;"><div class="tooltip-inner">Tooltip on top</div></div>
```

Second activation:

```html
<div class="tooltip bs-tether-element bs-tether-abutted bs-tether-abutted-top bs-tether-element-attached-bottom bs-tether-element-attached-center bs-tether-target-attached-top bs-tether-target-attached-center fade bs-tether-enabled in" role="tooltip" id="tooltip373736" style="left: 0px; position: absolute; transform: translateX(208px) translateY(2368px) translateZ(0px); top: 0px;"><div class="tooltip-inner">Tooltip on top</div></div>
```

Third activation:

```html
[...]

PR #22276, opened 2017-03-26
Fix different tooltips offset when hovering

Johann-S:
Fix different tooltips offset when hovering by resetting tip classes 

/CC : @mdo and @bardiharborow 

Close : #21146

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -34,6 +34,7 @@ const Tooltip = (($) => {
   const JQUERY_NO_CONFLICT  = $.fn[NAME]
   const TRANSITION_DURATION = 150
   const CLASS_PREFIX        = 'bs-tether'
+  const TETHER_PREFIX_REGEX = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g')
 
   const Default = {
     animation   : true,
@@ -441,18 +442,9 @@ const Tooltip = (($) => {
 
     _cleanTipClass() {
       const $tip = $(this.getTipElement())
-      let tabClassStr = $tip.attr('class')
-      let tabClass = tabClassStr.split(' ')
-      let i = 0
-      while (tabClassStr.indexOf(CLASS_PREFIX) !== -1) {
-        if (tabClass[i].indexOf(CLASS_PREFIX) !== -1) {
-          $tip.removeClass(tabClass[i])
-          tabClassStr = $tip.attr('class')
-          tabClass = tabClassStr.split(' ')
-          i = 0
-        } else {
-          i++
-        }
+      const tabClass = $tip.attr('class').match(TETHER_PREFIX_REGEX)
+      if (tabClass !== null && tabClass.length > 0) {
+        $tip.removeClass(tabClass.join(''))
       }
     }
 
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -340,6 +340,7 @@ const Tooltip = (($) => {
           tip.parentNode.removeChild(tip)
         }
 
+        this._cleanTipClass()
         this.element.removeAttribute('aria-describedby')
         $(this.element).trigger(this.constructor.Event.HIDDEN)
         this._isTransitioning = false
@@ -436,6 +437,15 @@ const Tooltip = (($) => {
 
     _getAttachment(placement) {
       return AttachmentMap[placement.toUpperCase()]
+    }
+
+    _cleanTipClass() {
+      const $tip = $(this.getTipElement())
+      if (this.config.toggle === 'tooltip') {
+        $tip.attr('class', 'tooltip fade')
+      } else {
+        $tip.attr('class', 'popover fade')
+      }
     }
 
     _setListeners() {
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -441,10 +441,18 @@ const Tooltip = (($) => {
 
     _cleanTipClass() {
       const $tip = $(this.getTipElement())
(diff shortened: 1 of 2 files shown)


Johann-S:
The test failed on trying to stop the tunnel with SauceLabs
https://travis-ci.org/twbs/bootstrap/jobs/215178467#L431-L451

Johann-S:
Squashed my commits and no more failed on Travis

js/src/tooltip.js
```diff
@@ -438,6 +439,15 @@ const Tooltip = (($) => {
       return AttachmentMap[placement.toUpperCase()]
     }
 
+    _cleanTipClass() {
+      const $tip = $(this.getTipElement())
+      if (this.config.toggle === 'tooltip') {
+        $tip.attr('class', 'tooltip fade')
```
Johann-S: good feedbacks @vanduynslagerp thank you 👍

js/src/tooltip.js
```diff
@@ -438,6 +439,15 @@ const Tooltip = (($) => {
       return AttachmentMap[placement.toUpperCase()]
     }
 
+    _cleanTipClass() {
```
Johann-S: Yes my bad I'll do that 😄

Johann-S: Finally I found a generic solution but thank you @vanduynslagerp 👍

vanduynslagerp: Using the `CLASS_PREFIX` is clever !
You can make this code shorter and more efficient thought:
```js
// On top of the file
const TETHER_PREFIX_REGEX = new RegExp(`(\S*)${CLASS_PREFIX}(\S*)`, 'g');

const $tip = $(this.getTipElement())
$tip.removeClass($tip.attr('class').match(TETHER_PREFIX_REGEX);
```
That makes the code more readable and  would avoid the multiple calls to Jquery to remove the classes one by one. It also avoid the loop and the assigning and reassigning of temporary variables multiple time.

Johann-S: You're right nice regex thanks you @vanduynslagerp

Johann-S:
I changed a bit your regex @vanduynslagerp but thank you it works very well 👍

Johann-S:
This PR (https://github.com/twbs/bootstrap/pull/17094) tried to do the same thing but ``addTargetClasses`` only remove Tether class from the element which trigger the tooltip not the tooltip.

vanduynslagerp:
> This PR (#17094) tried to do the same thing but addTargetClasses only remove Tether class from the element which trigger the tooltip not the tooltip.

I didn't check the Tether code, but is it possible that's it's actually a Tether bug ? It would be reasonable to think that when you activate the same element twice Tether should position it at the same position.
Or is the problem due to something Bootstrap does that prevent Tether to handle its own generated classes ?

Johann-S:
Currently they are no way of cleaning the tooltip element in Tether and I didn't found any PR which target this issue on Tether repository. 
IMO it's due to Tether classes so until they don't fix this issue it's better to have this PR

vanduynslagerp:
> Currently they are no way of cleaning the tooltip element in Tether and I didn't found any PR which target this issue on Tether repository.
IMO it's due to Tether classes so until they don't fix this issue it's better to have this PR

So in that case it might be better to open an issue with Tether and apply this PR if it's not fix fast enough? And we could remove this workaround once it's fixed ?

Johann-S:
Finally after a quick search is possibly an issue in BS classes (but not sure about that) see : 
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_tooltip.scss#L17-L59

FYI this issue exist since the merged of #17094 because before this PR BS4 cleaned Tether classes

vanduynslagerp:
> Finally after a quick search is possibly an issue in BS classes (but not sure about that) see :
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_tooltip.scss#L17-L59
FYI this issue exist since the merged of #17094 because before this PR BS4 cleaned Tether classes

Right ! Actually the code removed in #17094 was mentioning the Tether issue https://github.com/HubSpot/tether/issues/36, that is still opened....since March 2014...

Johann-S:
Until they are a fix published by Tether I'll merge this PR.

Thank you for your help and feedbacks @vanduynslagerp 👍

IdanCo:
@Johann-S kuddos for the fix, but even after the merge there is a strange behavior - when scrolling away and then back, the margin changes for the **top** and **left** tooltips. try scrolling down and up here -
http://jsbin.com/gopevakala/edit?html,output

I'm assuming it has something to do with the negative margins **top** and **left** tooltips has -
```sass
// scss/_tooltip.scss
  &.tooltip-top,
  &.bs-tether-element-attached-bottom {
    margin-top: -$tooltip-margin;
```
```sass
// scss/_tooltip.scss
  &.tooltip-left,
  &.bs-tether-element-attached-right {
    margin-left: -$tooltip-margin;
```

While right and bottom tooltips has positive margin -
```sass
// scss/_tooltip.scss
  &.tooltip-right,
  &.bs-tether-element-attached-left {
    margin-left: $tooltip-margin;
```
```sass
// scss/_tooltip.scss
  &.tooltip-bottom,
  &.bs-tether-element-attached-top {
    margin-top: $tooltip-margin;
```
Also, you can notice that the initial state lacks margins -
tooltips-margin

I tried looking for a quick win here by replacing the negative with positive, but the inconsistency continues. If you want i'll be happy to dig deeper.

Johann-S:
Please @IdanCo open a new issue to expose this problem thank you 👍

PR #22279, opened 2017-03-26
Changed Display Utilities classnames

simPod:
I personally think names should be as much descriptive as possible and that applies for classes as well. I don't fancy naming display class names as `.d*` so I changed it to `.display*`.

`.d*` reminded me my old ways of programming like naming variable containing User object as  `$u` instead of `$user` or even worse UserModel as `$um` instead of `$userModel` The problem with this usually becomes when there's another object with same first letters. eg. UserManager and UserModel. Someone resolves it with `$um` and `$um2` which is not the best way to solve it :)

So IMO class name should say what it does. `.display-sm-none` says it very well I think. For `.d-sm-none` I would have to search the docs if I'm user not familiar with Boostrap.

### DIFF
--- a/docs/_includes/nav-docs.html
+++ b/docs/_includes/nav-docs.html
@@ -1,4 +1,4 @@
-<form class="bd-search d-none d-sm-block">
+<form class="bd-search display-none display-sm-block">
   <input type="search" class="form-control" id="search-input" placeholder="Search..." aria-label="Search for..." autocomplete="off">
   <div class="dropdown-menu bd-search-results" id="search-results"></div>
 </form>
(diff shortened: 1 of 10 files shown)


mdo:
> So IMO class name should say what it does. .display-sm-none says it very well I think. For .d-sm-none I would have to search the docs if I'm user not familiar with Bootstrap.

This holds true no matter the class name. Thanks, but we'll leave it as-is <3.

simPod:
Is there any rational reason?

mdo:
Yes, though not one you're likely going to agree with. These utilities are likely to see a ton of use, so keeping the syntax short and memorable is top of mind. Just as `d` is shorter than `display`, `m` and `p` are shorter than `margin` and `padding`. The display utilities are part of a larger suite of classes. Where necessary, we break that pattern though.

Also worth noting that `display` is the arguably the most used property starting with `d`. Only likely runner-up is `direction` and we don't make extensive use of that.

PR #22286, opened 2017-03-28
Improve card image top and card image bottom can use with another car…

nickjazz:
Now when we use card-header and card-img-top or card-footer card-img-bottom at the same time, the border-radius of image should be remove. 

Improve card image top and card image bottom can use with another card components

- Only card-img-top is first-child add border-radius

- Only card-img-bottom is last-child add border-radius


mdo:
These classes are meant to be first and last; I don't see this as necessary. Thanks though! Also next time please don't push compiled `dist` files, we just need the source Sass files :).

nickjazz:
@mdo  Thanks, i will keep in mind next time.
About why need add this class, please reference this codepen demo
http://codepen.io/kimix/pen/evLxpV

if i use .card-header and card-img-top, or card-footer and card-img-bottom. the border-radius on img is not necessary.

so just need border-radius if they are the first or last one.

PR #22288, opened 2017-03-28
Card image fixes

### LINKED ISSUE #21650: Card images distorted

### LINKED ISSUE #21885: Wrong Img-Height when using 'img-fluid' inside a card
When you use an image with class 'img-fluid' within a 'card', the width is correct but the height is the original height of the image.

https://jsfiddle.net/6Lef7tzo/

This happens because the card has now the style 'display: flex'. If i style the card with 'display: block' everthing works fine!

Bootstrap 4 Alpha 6

Chrome + Firefox OK!
IE 11 = bug

mdo:
Closes #21857, fixes #21650, and closes #21885 (dupe of #21650).

The card image stretching affected `.card-img`, `.card-img-top`, and `.card-img-bottom` as we utilize flexbox and `flex-direction: column` on `.card`. Intrinsic image sizing is basically borked with flex due to the default `stretch` value for aligning the content. There's no real way around this it seems _other_ than `width: 100%;` as no other flex property will address the aspect ratio.

### DIFF
--- a/docs/components/card.md
+++ b/docs/components/card.md
@@ -644,7 +644,7 @@ Cards can be organized into [Masonry](http://masonry.desandro.com)-like columns 
 {% example html %}
 <div class="card-columns">
   <div class="card">
-    <img class="card-img-top img-fluid" data-src="holder.js/100px160/" alt="Card image cap">
+    <img class="card-img-top" data-src="holder.js/100px160/" alt="Card image cap">
     <div class="card-block">
       <h4 class="card-title">Card title that wraps to a new line</h4>
       <p class="card-text">This is a longer card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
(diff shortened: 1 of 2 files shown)


mdo:
Example of new styles at https://jsbin.com/qelenuh/.

Ponant:
Thanks @mdo , that works in my case. I hope the beta will be released now. I am very happy with the alpha 6 version :)

Ponant:
@mdo  , it does not work for IE11 unfortunately.

Ponant:
To solve the issue for IE 11 , I had to add not only width:100% but also
```
.card {
    display: block;
}
```

floscher:
This pull request does not fix #21885. The issue is still present in `v4-beta` and `v4-beta.2`.
The workaround by @Ponant does work, but only when the card is not part of a card group.

jeroendesloovere:
The fix from @Ponant works.
I used the class `d-block` instead.
Example `<div class="card d-block">`

PR #22294, opened 2017-03-28
Don't test IE9 on Win7 anymore

mdo:
/cc @Johann-S


Johann-S:
So easy 😲 thank you 👍

Issue #22296, opened 2017-03-28
collapse.js does not work anymore


git-marijus:
Since commit 24924c23b24c1b196c42166ebe9a17d31f0ee720 `<a data-toggle="collapse" href="#collapseExample">Some text</a>` reloads the page. `event.preventDefault()` is not being called anymore.

mdo:
Do you have a demo that reproduces this?

/cc @Johann-S

Johann-S:
Yes that's true I'll fix that now thank you @git-marijus 

My bad sorry 😢

git-marijus:
The code below works well in alpha6 and does not work in current git.
The logic in https://github.com/twbs/bootstrap/commit/24924c23b24c1b196c42166ebe9a17d31f0ee720#diff-6ef28a975ef633bb3b3318b9917df9e5R366 is obviously wrong. It only calls `event.preventDefault()` on inputs and textareas instead of the other way around.
```
<a data-toggle="collapse" href="#collapseExample" aria-expanded="false" aria-controls="collapseExample">
  Link with href
</a>
<div class="collapse" id="collapseExample">
  <div class="card card-block">
    Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid. Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
  </div>
</div>
```

Johann-S:
That's true, that's what happened when you are in a hurry

git-marijus:
No big deal ;)

vanduynslagerp:
As the `if` logic was wrong, shouldn't the unit have failed ?

PR #22297, opened 2017-03-28
Collapse - Fix check to not prevent event for input and textarea

Johann-S:
Thank you @git-marijus

Close : #22296

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -357,7 +357,7 @@ const Collapse = (($) => {
    */
 
   $(document).on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-    if (/input|textarea/i.test(event.target.tagName)) {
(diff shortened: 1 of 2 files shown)


mdo:
Does the test need updating per https://github.com/twbs/bootstrap/issues/22296#issuecomment-289909869?

Johann-S:
I improved my unit test

vanduynslagerp:
In you new unit test you just added `assert.ok(e.isDefaultPrevented() === false)` that test if the `shown.bs.collapse` event is prevented. That will never be prevented.

Your modification in the code prevent the event from `Event.CLICK_DATA_API`.
They are not the same event. Your test still doesn't do what you intend.

You should test if the `<div id="collapsediv1"/>` has the class `show` instead, to make sure that the click on the checkbox did expand the collapsible area

Johann-S:
Thank you @vanduynslagerp 👍 Time for me to sleep I make to much mistakes tonight

PR #22298, opened 2017-03-28
Add border width to input height variables

jacobmllr95:
This commit adds the border width (top + bottom) to the input height variables by default. This makes IMHO much more sense because every element is set to `box-sizing: border-box`, so the border height should count to the total element height.

This commit introduces some handy helper functions to convert `rem` units to `px` and vice versa. This is required to calculate the total input height because the border width is defined in `px` and the other height related variables (`padding`, `font-size` etc.) are defined in `rem`.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -135,8 +135,7 @@
 .custom-select {
   display: inline-block;
   max-width: 100%;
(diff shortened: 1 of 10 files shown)


mdo:
This will be addressed by #22395.

Issue #22300, opened 2017-03-29
Inconsistency Between Tooltip's Margins


IdanCo:
Following issue #21146 and PR #22276, tooltips still exhibit inconsistencies with the margins.

The issue manifests itself in two ways:

1. Initial state lacks margins:
![tooltips-margin](https://cloud.githubusercontent.com/assets/10364499/24447346/4b21aedc-1479-11e7-9b13-1a4cc2586658.png)

2. When the browser re-renders the tooltips, the margins changes. open [this bin](http://jsbin.com/gopevakala/edit?html,output) and try scrolling down and then back up. notice how the margins of the **top** and **bottom** tooltips changes.

Johann-S:
Confirmed for me.

If you want to work on this issue @IdanCo do not hesitate 👍

mdo:
No longer an issue in v4-dev.

PR #22301, opened 2017-03-29
fix tooltips margins

IdanCo:
## Summary:
As described in #21146 and #22300, and [shown here](http://jsbin.com/gopevakala/edit?html,output), left and top tooltips exhibit inconsistencies in regards to margins.

## before
Top and left tooltips sometimes ignore margins

## After
All tooltips now use padding instead of margin

## Technical 
[As shown here](https://www.diffchecker.com/kyilr4EM), **top** and **left** tooltips switches between _absolute_ and _fixed_ positions. No doubt this is what causes the inconsistencies, but I admit the logic behind this switch, and the reason it differs from **bottom** or **right** tooltips remains an enigma to me. [It happens here](https://github.com/HubSpot/tether/blob/3d7119e590661f8c9e9e566c8a7640c189687215/src/js/tether.js#L734), but since it's deep inside tether's logic, I tried looking for a workaround in bootstrap's code. Since the margins seems to be what causes the issue, i've successfully replaced them with paddings, while preserving the spacing by adding it to the arrow position. 

Example of fix available here -
http://jsbin.com/noyitof/edit?html,css,js,output

### DIFF
--- a/scss/_tooltip.scss
+++ b/scss/_tooltip.scss
@@ -15,11 +15,10 @@
 
   &.tooltip-top,
   &.bs-tether-element-attached-bottom {
-    padding: $tooltip-arrow-width 0;
-    margin-top: -$tooltip-margin;
+    padding: $tooltip-arrow-width + $tooltip-margin 0;
 
     .tooltip-inner::before {
-      bottom: 0;
+      bottom: $tooltip-margin;
       left: 50%;
       margin-left: -$tooltip-arrow-width;
       content: "";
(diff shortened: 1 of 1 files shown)


IdanCo:
updated jsbin - http://jsbin.com/noyitof/edit?html,css,js,output

Johann-S:
I close this PR because #22444 fixed that and will be merge on sunday.

Thank you @IdanCo

PR #22315, opened 2017-03-30
[DOCS] Corrected jquery activation example

friedow:


Johann-S:
Hi @friedow, currently ``option`` is not an option of Tooltip plugin.
``option`` allow to pass options like ``title`` or ``template``

for example : 
```javascript
$('#tooltip').tooltip({
  triggered: 'hover',
  title: 'Johann-S'
})

friedow:
@Johann-S : Thanks for your quick and helpful reply! You should probably add this to the documentation. The current documentation is a bit misleading.

PR #22316, opened 2017-03-30
Popover + Tooltip - fix error when content or title is a number

### LINKED ISSUE #20193: Popover can't show number as content

Johann-S:
As @cvrebert suggest (https://github.com/twbs/bootstrap/pull/20210#issuecomment-229783242) convert content and title to string

Close #20193

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -605,6 +605,10 @@ const Tooltip = (($) => {
         }
       }
 
+      if (config.title && typeof config.title === 'number') {
+        config.title = config.title.toString()
+      }
+
       if (config.content && typeof config.content === 'number') {
         config.content = config.content.toString()
       }
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -605,6 +605,10 @@ const Tooltip = (($) => {
         }
       }
 
+      if (config.content && typeof config.content === 'number') {
+        config.content = config.content.toString()
+      }
+
       Util.typeCheckConfig(
         NAME,
         config,
--- a/js/tests/unit/popover.js
+++ b/js/tests/unit/popover.js
@@ -365,7 +365,7 @@ $(function () {
       .modal('show')
   })
 
-  QUnit.test('should convert number content to string without error', function (assert) {
+  QUnit.test('should convert number to string without error for content and title', function (assert) {
     assert.expect(2)
     var done = assert.async()
     var $popover = $('<a href="#">@mdo</a>')
--- a/js/tests/unit/popover.js
+++ b/js/tests/unit/popover.js
@@ -364,4 +364,24 @@ $(function () {
       })
       .modal('show')
   })
+
+  QUnit.test('should convert number content to string without error', function (assert) {
+    assert.expect(2)
+    var done = assert.async()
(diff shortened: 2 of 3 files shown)


vanduynslagerp:
I don't think it's the right fix. What you are doing is preemptively converting the `data-title` and `data-content` to string, effectively you are bypassing the checks done in `Util.typeCheckConfig`.
I think that checks were put in place for a reason (I guess controlling the content to avoid XSS).
Moreover the `data-title` and `data-content` can be an `element` or a `function` and you are converting it to a string.

I think what @cvrebert meant in this [comment](https://github.com/twbs/bootstrap/pull/20210#issuecomment-229783242) was to consider the `title` and `content` as a string when you add them to the popover/tooltip not preemptively before the check. But that's not possible as the check is done before that.

The proper fix is to merge PR #20210 for the `data-content` of the popover and apply the same fix for the `data-title` in the tooltip (and add unit test that are not present in #20210)

Johann-S:
> Moreover the data-title and data-content can be an element or a function and you are converting it to a string.

That's not true, I only convert when it's a number that's all.
Moreover same is already done for ``delay``

vanduynslagerp:
The delay is not converted:
```js
if (config.delay && typeof config.delay === 'number') {
    config.delay = {
        show : config.delay,
        hide : config.delay
     }
}
```
The delay can be an object with `show`/`hide` or a single `number`. This code handle the second case (a single number meaning both `show` and `hide` are the same).  No conversion here and the original parameter still goes through `Util.typeCheckConfig`.

Johann-S:
For me it's a conversion but that's not the subject here.
Thank you for your feedbacks 👍

vanduynslagerp:
if `config.delay` is a string then `config.delay.show` will also be a string and will fail in `Util.typeCheckConfig` as intended. What I meant by 'not a conversion' is that is doesn't change the type. You code change the type by doing `toString()`.

Johann-S:
At the beginning ``config.delay`` is a Number and at the end ``config.delay`` is an Object that's why I said it's a conversion.

vanduynslagerp:
Actually I think the code regarding the delay, should be done after the call to `Util.typeCheckConfig` otherwise it defeat the purpose of the check and let the user pass any type as the delay.
Basically we shouldn't change the `config` before checking it.

Johann-S:
It's a point of view but, we only accept `string`, `element `and `function` for `title` and `content` so if it's not a string we convert the argument.

About `delay` to be compliant with documentation we have to perform the conversion.
> Delay showing and hiding the tooltip (ms) - does not apply to manual trigger type
If a number is supplied, delay is applied to both hide/show
Object structure is: delay: { "show": 500, "hide": 100 }

For me they are not a better point of view than an other, your point of view is equal to mine but a path have been chosen that's all.

Issue #22302, opened 2017-03-29
button group accessibility - Ensures every form element has a label


Lausselloic:
https://v4-alpha.getbootstrap.com/components/button-group/#button-toolbar

Ensures every form element has a label, the two sample for mail adress doesn't have a valid label (aria-described is not enough). I suggest to add an aria-label, which will reuse the palceholder content

e.g:
```
<span class="input-group-addon" id="btnGroupAddon">@</span>
<input type="text" class="form-control" placeholder="Input group example" aria-describedby="btnGroupAddon" aria-label="Input group example">
```

Lausselloic:
Maybe need also an update for all form elements also on input-group page. And need to update the accessibility best practice : aria-describedby and placeholder are NOT a valid technique to provide a label input information
http://v4-alpha.getbootstrap.com/components/input-group/#accessibility

patrickhlauke:
on the last point about `aria-describedby` and `placeholder`: while not ideal, it is my understanding that on inputs at least the latter DOES act as a "last resort" way of providing a label via the accessibility API if none of the better ways are present (see http://rawgit.com/w3c/html-api-map/master/index.html#accessible-name-and-description-calculation). so based on that, the current example code is not exactly elegant, but "valid" (/cc @stevefaulkner)

having said that, adding an explicit `aria-label` seems the cleaner option for those inputs. and removing `aria-describedby` (as it's clearly wrong) and `placeholder` (as it's contentious) from the advice section makes sense.

patrickhlauke:
xref https://github.com/w3c/html-aam/issues/84 which resulted in related bugs being filed against iOS/VoiceOver and Android/Chrome/Talkback

PR #22321, opened 2017-03-30
Add explicit aria-label to placeholder-only inputs

patrickhlauke:
While `placeholder` is nominally valid per spec
http://rawgit.com/w3c/html-api-map/master/index.html#accessible-name-and-description-calculation
it is inelegant, a fallback, and not supported in ios/safari/voiceover
and android/chrome/talkback, to name a few combinations

`aria-describedby` is also not really the right tool for providing a
name/label equivalent, so remove that from the list of possible
alternative methods / clarify how it can be used

closes https://github.com/twbs/bootstrap/issues/22302


patrickhlauke:
xref https://github.com/w3c/html-aam/issues/84 - in theory, IF iOS/VO and Android/Chrome/TalkBack fix their bugs, this PR's main change (adding extra `aria-label`) may be superfluous...but right here, right now, it's needed for these platforms (and even once fixed it's not actually wrong to have `aria-label` explicitly, just perhaps a bit "belt and braces" redundancy)

patrickhlauke:
adding to this: since I raised UA inconsistencies, it's now been officially decided to remove `placeholder` from the official "accessible name calculation" algorithm in HTML. user agents may still do this, but it's outside of spec, and won't be guaranteed to work cross-browser. in short: this PR was the right call https://github.com/w3c/html-aam/issues/87

PR #22324, opened 2017-03-31
Added mixin for better z-index management

naicko:
Combined all z-index variables into one list, added function for extracting the correct index, plus a better way for requesting a backdrop index. Only the main controls are listed in the z-index list, lower indexes can be achieved by passing an extra variable to the mixin.


mdo:
This doesn't really provide any improvement over the current variable z-index list. Sass maps are really only helpful for iterating over, but there's never an instance where we'd want to build a series of classes from this map (unlike with our grid, for example).

Thanks though!

PR #22336, opened 2017-04-02
Docs and CSS updates for #21807

mdo:
#21807 merged without a thorough review of the new docs content, and without some changes to the CSS reverted or changed. This PR aims to take care of those things.

This removes the CSS that was added for the justified and fill variants; that was tackled by requiring `.nav-item` on `.nav-link` in caebfcd246d7a09f311091adf2dbe835dee244b2. This way, we're not setting CSS just to override it in the following ruleset.

The docs updates are relatively straightforward. Some grammar fixes and minor tweaks here and there mostly. The biggest change was redoing the scrollspy's docs intro; I hate seeing huge callouts. Instead of that, let's just lead with "here's what you need."

Might keep pushing some tweaks here, unsure yet if this is ready to go. Still reading all the docs changes from that PR.

/cc @vanduynslagerp @Johann-S


vanduynslagerp:
Docs changes looks good to me. English is not my mother tongue, so there is surely improvement possible in the doc I wrote :-)

Regarding the CSS, as mentioned in this [comment](https://github.com/twbs/bootstrap/pull/21807/files#r107007012) the change [here](https://github.com/twbs/bootstrap/pull/21807/files#r107007012) create an issue with the bottom border of active tabs with this markup:
```css
<nav class="nav nav-tabs">
    <a class="nav-link" href="#">Item</a>
</nav>
```
See screenshot

PR #22337, opened 2017-04-02
Fix #22332: Adjust tests for hidden scrollbar

### LINKED ISSUE #22332: Units tests failed on OSX and Linux
Due to this commit https://github.com/twbs/bootstrap/commit/a14451deb3b8368de76aad6e728b27032bf8097a

job : https://travis-ci.org/twbs/bootstrap/jobs/217756286

Maybe @Deilv, do you have any idea why ? 

Edit maybe it would be better to move unit test relative to this commit in a visuel unit test

deilv:
On macOS and mobile devices the scrollbar is not part of the document, but rather floats above it. This means that it does not affect body/fixed element padding and scrollbarWidth is equal to 0.

The problem in #22332 was caused by the fact that the test expected padding-right to be increased while opening a modal, but threw an error when it was not (because padding + 0 = padding).

To address the issue the following changes have been made:

* A copy of the original getScrollbarWidth() function has been added (because the modal function is private)
* Tests now check if the padding or margin is changed **as expected** (even if zero) rather than just checking if it changed

### DIFF
--- a/js/tests/unit/modal.js
+++ b/js/tests/unit/modal.js
@@ -12,6 +12,15 @@ $(function () {
     before: function () {
       // Enable the scrollbar measurer
       $('<style type="text/css"> .modal-scrollbar-measure { position: absolute; top: -9999px; width: 50px; height: 50px; overflow: scroll; } </style>').appendTo('head')
+      // Function to calculate the scrollbar width which is then compared to the padding or margin changes
+      $.fn.getScrollbarWidth = function () {
+        var scrollDiv = document.createElement('div')
+        scrollDiv.className = 'modal-scrollbar-measure'
+        document.body.appendChild(scrollDiv)
+        var scrollbarWidth = scrollDiv.offsetWidth - scrollDiv.clientWidth
+        document.body.removeChild(scrollDiv)
+        return scrollbarWidth
+      }
     },
     beforeEach: function () {
       // Run all tests in noConflict mode -- it's the only way to ensure that the plugin works in noConflict mode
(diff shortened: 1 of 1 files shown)


js/tests/unit/modal.js
```diff
@@ -12,6 +12,14 @@ $(function () {
     before: function () {
       // Enable the scrollbar measurer
       $('<style type="text/css"> .modal-scrollbar-measure { position: absolute; top: -9999px; width: 50px; height: 50px; overflow: scroll; } </style>').appendTo('head')
+      $.fn.getScrollbarWidth = function () {
```
Johann-S: Maybe you can add here, why did you add this function

js/tests/unit/modal.js
```diff
@@ -341,24 +349,27 @@ $(function () {
...
-        done()
-      })
-      .on('shown.bs.modal', function () {
-        var currentPadding = $body.css('padding-right')
-        assert.notStrictEqual(currentPadding, originalPadding, 'body padding should be adjusted while opening')
-        $(this).bootstrapModal('hide')
-      })
-      .bootstrapModal('show')
+    // This test only works when the vertical scrollbar is visible and part of the document (which is not the case for macOS and mobile devices)
+    if ($(this).getScrollbarWidth()) {
```
Johann-S: It will fail because no test will be run here you should add an ``else`` condition

Johann-S (changes requested):

Deilv:
There's no point running the test when no changes are made to the DOM. I will move the conditional outside and if that fails too I will remove the opening test altogether.

wuxue12:
thank you for your comments.


983299179@qq.com
 
From: Ilias
Date: 2017-04-03 17:20
To: twbs/bootstrap
CC: Subscribed
Subject: Re: [twbs/bootstrap] Fix #22332: Check if the scrollbar affects the document before running the padding tests (#22337)
There's no point running the test when no changes are made to the DOM. I will move the conditional outside and if that fails too I will remove the opening test altogether.

Deilv:
@Johann-S I modified the tests to check if the padding/margin is changed as expected, even if that is zero. Should no longer give errors, but please give it a try.

js/tests/unit/modal.js
Deilv: Conditional removed.

Johann-S (approved):

Johann-S:
Thank you @Deilv every tests passed (https://travis-ci.org/twbs/bootstrap/jobs/218028318)

PR #22342, opened 2017-04-03
Updates docs to match Tooltip type checking

gareth:
Spotted when looking up [documentation for Popups](http://getbootstrap.com/javascript/#popovers-options), and wondering why a raw element couldn't be passed in as a `container` according to the documentation - I looked in the source and [saw that it *could*](https://github.com/twbs/bootstrap/blob/c78cbb275b0c559f5c9436ee3df816e1f6195201/js/src/tooltip.js#L284) – whatever's in `container` (if truthy) gets passed into [jQuery's appendTo](http://api.jquery.com/appendto/)!

I also spotted that the Tooltip type checking [explicitly allows for an element to be passed](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L65), it's literally just that the documentation is out of sync.

With this change, the documentation for Tooltip and Popover now matches the types defined in the DefaultType constant in js/src/tooltip.js

### DIFF
--- a/docs/4.0/components/popovers.md
+++ b/docs/4.0/components/popovers.md
@@ -179,7 +179,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
     </tr>
     <tr>
       <td>container</td>
-      <td>string | false</td>
+      <td>string | element | boolean</td>
       <td>false</td>
       <td>
         <p>Appends the popover to a specific element. Example: <code>container: 'body'</code>. This option is particularly useful in that it allows you to position the popover in the flow of the document near the triggering element - which will prevent the popover from floating away from the triggering element during a window resize.</p>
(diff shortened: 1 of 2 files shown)


docs/components/popovers.md
```diff
@@ -181,7 +181,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
     </tr>
     <tr>
       <td>container</td>
-      <td>string | false</td>
+      <td>string | element | boolean</td>
```
samjewell: If you felt like it, you could also change `false` to `boolean` on the tooltip container too:

From:
```
<td>string | element | false</td>
```
to:
```
<td>string | element | boolean</td>
```

Johann-S: Agreed with @samjewell

gareth: Interestingly,  if you actually set `container: true` then the tooltip/popover won't work properly, because of this line:

```
var container = this.config.container === false ? document.body : $(this.config.container);
```

I'm actually going to change these updates back to `false` to make that clear. The important thing for this PR is that `element` is added.

Johann-S:
@gareth you have conflicts in your issue can you update your branch please ?

Johann-S (approved):

PR #22354, opened 2017-04-05
Fix #22225: use $btn-focus-box-shadow in the button-variant mixin

### LINKED ISSUE #22225: $btn-focus-box-shadow variable does not trickle down to button variants
In v4, we now have:
`$btn-focus-box-shadow:           0 0 0 2px rgba($brand-primary, .25) !default;`

If I change it to:
`$btn-focus-box-shadow:           0 10px 25px rgba($black, .15) !default;`

Doesn't work. :/ This is overridden by individually by each button variant. Take this for example:
screen shot 2017-03-20 at 12 50 37 pm

There's 2 solutions that would make sense to me:
1. individual button styles shouldn't override this
2. there should be separate variables for focus-box-shadow on each button variant (for example: $btn-info-focus-box-shadow)

Thank you!

bwalendz:
Make sure variable is used here.

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -24,7 +24,7 @@
     @if $enable-shadows {
       box-shadow: $btn-box-shadow, 0 0 0 2px rgba($border, .5);
     } @else {
-      box-shadow: 0 0 0 2px rgba($border, .5);
+      box-shadow: $btn-focus-box-shadow;
     }
   }
 


scss/mixins/_buttons.scss
```diff
@@ -24,7 +24,7 @@
     @if $enable-shadows {
       box-shadow: $btn-box-shadow, 0 0 0 2px rgba($border, .5);
```
bwalendz: Not too familiar with enable-shadows and the thought process behind this line but possibly $btn-box-shadow should be $btn-focus-box-shadow here as well.

mdo:
This is done so that if you enable shadows, you get the inner input shadows plus the focus with a custom color. When you don't have shadows enabled, you still get the custom focus styles.

PR #22355, opened 2017-04-05
css: .navbar elements aligning

zalog:
Vertical centering `.navbar-brand` and `.navbar-toggler`.

Please check this pen: http://codepen.io/zalog/pen/zZXbwz

Since this great fix #22230, the comment here https://github.com/twbs/bootstrap/pull/22355/files#diff-df1f74faf207a474f1d78c86393221c8L105 is not relevant anymore.
cc: https://github.com/twbs/bootstrap/pull/21626#issuecomment-292129473

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -19,6 +19,7 @@
   position: relative;
   display: flex;
   flex-wrap: wrap; // allow us to do the line break for collapsing content
+  align-items: center;
   justify-content: space-between; // space out brand from logo
   padding: $navbar-padding-y $navbar-padding-x;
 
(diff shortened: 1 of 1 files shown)


zalog:
Or... I think is better this way: http://codepen.io/zalog/pen/bqygdZ
To keep aligning logic on the `.navbar` / parents only, and delete it from any child.
We already have `align-items: center`, but it's only for `.navbar-expand-*`

mdo:
Aligning the navbar contents to a smart default as you suggest in https://github.com/twbs/bootstrap/pull/22355#issuecomment-292133457 would be rad.

zalog:
Yep, I'll do that soon.

zalog:
Done.
You can check this demo: http://jsbin.com/fibini

Issue #22061, opened 2017-02-24
carousel aria-role listbox require owned element option - Accessibility


Lausselloic:
Actual carousel use role listbox, but doen't provide role option on each slide
https://v4-alpha.getbootstrap.com/components/carousel/

But aria role listbox require at least one option element inside listbox
https://www.w3.org/TR/wai-aria/roles#listbox

patrickhlauke:
I would actually say that the `role="listbox"` on the carousel is wrong. listboxes are conceptually similar to `<select>` elements, meaning a user can go in and choose/select one or more options. this is different from the generic carousel (as used in bootstrap's docs) which is a widget to simply show information/navigate through discrete chunks of content. it's actually, in a way, more akin to a proper tabbed interface (if we envisages each slide as a separate tabpanel), but its presentation and interactions with it obviously don't match that model.

the difficulty with carousels is that they are among the most "alien" widgets in terms of accessibility since there's no directly comparable standard interface widget that they  conceptually borrow from or mimic (unlike, say, tabbed interfaces, which are used - at least on the Windows side - frequently in native apps and have therefore been given explicit ARIA roles and behaviors). it's often quite use-case-dependent, which makes it even trickier for a generic framework like bootstrap...

having said all that, i think as a first step, i'll *remove* the `role="listbox"` which, to my mind, is clearly wrong. nonetheless, we'll need some more thinking around what the best, most generic approach could be for carousels (in terms of markup, keyboard interaction, focus handling, any roles/attributes needing to be exposed to assistive tech)...

patrickhlauke:
this may be useful (for context and later more thorough fixes to carousel) https://www.w3.org/WAI/tutorials/carousels/

Mikeks81:
I understand and mostly agree with what @patrickhlauke has stated but since removal the carousel component is essentially unusable to screen readers until there is a more appropriates decision made to approach this component. Is keeping it accessible ( even arguably wrong in implementation ) better than removing the listbox and making the carousel inaccessible??? At the very least screen readers were able to read the carousels with listbox and options but now it's completely hidden.

PR #22365, opened 2017-04-06
accessibility - carousel aria fix

Lausselloic:
Propose a PR to fix aria role listbox required children


patrickhlauke (changes requested):
see https://github.com/twbs/bootstrap/issues/22061#issuecomment-292926679 - i would actually suggest removing the `role="listbox"` at this stage, rather than adding `role="option"`. happy to do this myself in a separate PR, if you'd prefer closing this one (as it seems a bit messy already with various extraneous commits)

patrickhlauke:
xref https://github.com/twbs/bootstrap/issues/22061

Lausselloic:
Yes I will close it

patrickhlauke:
i hope i didn't scare you off from contributing though...would have been happy for you to rework the PR (but it would mean a rebase etc). let me take opportunity though to say thank you for raising these things which i knew were there but never got around to methodically look at :+1:

Lausselloic:
No don't wory I won't be affraid :-) 
I agree with the proposition to remove the listbox. 
I need to cleanup my dedicated PR repository, as I know my PR are not very clean and doesn't only contain proper PR code.

I will continue to push accessibility improvement.

PR #22366, opened 2017-04-06
Add border for group input in the navbar-inverse

### LINKED ISSUE #21098: Invisible border of input-group in inverse navbar
When in a **inverse** navbar I add a input-group containing a input and an add-on, the border of the input is badly rendered while the border of the add-on is ok. If the navbar is not inverse, all is fine.
Here is a jsfiddle showing the problem : [https://jsfiddle.net/mw7ts10m/](https://jsfiddle.net/mw7ts10m/)

temp
The problem is with version alpha 5 (tested on firefox and IE11).

With the alpha 2 version, the borders was well rendered, as showed in this jsfiddle : [https://jsfiddle.net/mw7ts10m/2/](https://jsfiddle.net/mw7ts10m/2/)
It seems broken since alpha 3.

bastienmoulia:
Add `$btn-secondary-border` border color for form control in input group in the navbar when there is a dark background.

Because the normal border color is `rgba(0,0,0,.15)` and depend of the background color the input in input group seems to have no border.

Fix #21098
Close #21318

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -279,4 +279,9 @@
   .navbar-text {
     color: $navbar-inverse-color;
   }
+  
+  .input-group .form-control {
+    border-color: $btn-secondary-border;
+  }
+  
 }


mdo:
This isn't the right solution to this.

Issue #22328, opened 2017-04-01
Do not let system-ui come before "Segoe UI" in font-family—or remove it from the font stack altogether


infinnie:
Reverting #21356.

See https://infinnie.github.io/blog/2017/systemui.html for the reason.

mdo:
We ran into some issues with this in GitHub, too. Will push an update.

bardiharborow:
Is there an upstream bug with the Chrome team, or is this just entirely unfixable?

infinnie:
@bardiharborow I believe it an unwanted feature.

patrickhlauke:
It does seems that the feature itself - that with `system-ui` the browser uses the default defined by the system - is actually working exactly as it should, so not a bug in Chrome itself. the problem lies with the OS' choice of system font. e.g. looking at the screenshot on https://infinnie.github.io/blog/2017/systemui.html i would say that Chrome here behaves exactly as instructed - "use the system font".

in short, i don't believe this needs an upstream bug for Chrome, since it's behaving correctly. the fact that this then leads to sites that look horrid/broken when the system font is horrid/broken is purely a Bootstrap issue.

bardiharborow:
@patrickhlauke I was about to ask if Windows had a correct fallback algorithm when dealing with English text in UI, but then I realised that the whole point of switching the OS language is that the OS is fully localised to Chinese, so this may not be something that exists. As an alternative, is there a bug lodged with Microsoft to fix the font (though good luck getting any progress with that). I notice that even Chrome is using the broken font for it's UI (omnibar, etc.), so I don't know if they would consider implementing a platform-specific fallback for this situation (though Chrome seems like they might be stubborn on the issue to try to force Microsoft into acting instead). Failing all that, we may just have to wait 6 months and see.

patrickkettner:
thanks for the headsup @patrickhlauke - will pass over to the font folks

alanhogan:
I haven't seen any screenshots of `system-ui` causing unwanted effects on OSes newer than Windows Server 2008, although I read infinnie’s blog post mentioning font-weight issues on Windows 8.1 and/or 10.

1. Does anyone have relevant screen shots to share?

2. @mdo, does GitHub localize their interface? that is, would GitHub ever render in e.g. Chinese instead of English?

patrickkettner:
This is the same text rendered in SegoeUI and YaHei, with default weights.

[...]

PR #22377, opened 2017-04-07
Remove system-ui from font stack due to i18n issues

mdo:
On Windows, `system-ui` causes all sorts of issues. This PR removes it for now, reverting #21356 and closing #22328.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -244,7 +244,7 @@ $transition-collapse:    height .35s ease !default;
 //
 // Font, line-height, and color for body text, headings, and more.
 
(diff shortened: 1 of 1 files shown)


2dareis2do:
system-ui seems to have made it back into the latest release

https://github.com/twbs/bootstrap/blob/72d3b6efc4b92e83a4abca6f5bc0cd7e6fd25931/scss/_variables.scss#L606

Artoria2e5:
That's a good thing: the i18n concern around Chinese is unfounded. `system-ui` is either same as, or much better than (Microsoft Yahei UI) what you'd get with a bare `sans-serif` (SimSun). The "same as" case occurs only on Windows 7 and older -- check their 2024 market share.

mirabilos:
It’s still a cross-platform concern. On Linux with no desktop environment, there is not even a way to *set* that.

Please do **not** use `system-ui`.

Issue #20925, opened 2016-10-17
[Chrome] Cards inside .card-columns stack differently


glebez:
In case when there's a few cards inside card-columns container they get distributed differently in Chrome, comparing to FF and Safari.

For example, with 2 cards, in Chrome they stay in the same column and are placed one under another. In FF and Safari they are distributed among 2 columns and displayed side by side.

Adding 3rd card leads to this: In chrome there are 2 cards in the 1st column and 1 in the 2nd. In other browsers there is 1 card in each of 3 columns.

Here is the [reduced test case for 2 cards](http://codepen.io/glebez/pen/QKVokJ).
Here is the [reduced test case for 3 cards](http://codepen.io/glebez/pen/qarzwa).

I'm using BS4 alpha 3, running Mac Os 10.11.6, Chrome 53.0.2785.116 (64-bit), Firefox 49.0.1 and Safari 9.1.2 (11601.7.7).

mdo:
Can you try with the latest from `v4-dev`? I imagine it's something to do with `inline-block` on those elements, but I don't recall seeing this issue before in Chrome.

glebez:
Same behaviour with the v4-dev css: http://codepen.io/glebez/pen/pExPdP

scottgrayson:
This customization worked for me
```
.card-columns .card {
    display: block;
}
```

glebez:
@scottgrayson, be aware that using this workaround causes the card to break in multiple columns. This was addressed in this ticket: https://github.com/twbs/bootstrap/issues/20654

jeronimok:
Hi @glebez  
Did you get any solution to this problem?

trackleft:
https://github.com/twbs/bootstrap/pull/22378

DLongoni:
Hello, I see that @trackleft solution has been closed. I am still experiencing the same problem, though. Have you found a solution?

ilromolo:
Hello, same for me. Using bootstrap-4.0.0-alpha-6, if .card are less than three, cards stack differently in Chrome and Opera.

I modified the last part of _card.scss:

````
//
// Columns
//

@include media-breakpoint-up(sm) {
    .card-columns {
        column-count: $card-columns-count;
        column-gap: $card-columns-gap;

        .card {
            -webkit-column-break-inside: avoid;
            //display: inline-block; // Don't let them vertically span multiple columns
            width: 100%; // Don't let their width change
            margin-bottom: $card-columns-margin;
            
        }
    }
[...]

PR #22378, opened 2017-04-07
#20925 [Chrome] Cards inside .card-columns stack differently

trackleft:
https://github.com/twbs/bootstrap/pull/22378

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -268,9 +268,17 @@
     column-gap: $card-columns-gap;
 
     .card {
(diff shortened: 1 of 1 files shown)


mdo:
Have any explanation for the changes? Or a JS Bin demo of the before/after?

trackleft:
http://codepen.io/trackleft/pen/ZedaXb

mdo:
Never saw an updated example, and given the lack of context here, I'm going to close this out for now. Happy to revisit in the future with an example that explains the changes!

PR #22379, opened 2017-04-07
css: just some ordering

zalog:
Order x & y axis.
Now that's really nothing fancy :)


mdo:
I think I had these first because in the two-value shorthand, the y-value comes first :grin:

zalog:
Yep, if you think about `padding`. I'll correct this PR soon.
I think the file was written like that because most of us associate this with `background-position` 😁

zalog:
Done

PR #22388, opened 2017-04-08
eslint: fix semicolon and some duplicate key

zalog:
- some ESLint semicolon validations
- some duplicate object key

Test plan: `grunt test`

### DIFF
--- a/js/.eslintrc.json
+++ b/js/.eslintrc.json
@@ -72,7 +72,6 @@
     "no-new": "error",
     "no-new-func": "off",
     "no-new-wrappers": "error",
-    "no-new": "error",
     "no-octal-escape": "error",
     "no-param-reassign": "off",
     "no-proto": "error",
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -7,16 +7,16 @@
  */
 
 module.exports = function (grunt) {
-  'use strict'
+  'use strict';
 
   // Force use of Unix newlines
-  grunt.util.linefeed = '\n'
+  grunt.util.linefeed = '\n';
 
   RegExp.quote = function (string) {
-    return string.replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&')
-  }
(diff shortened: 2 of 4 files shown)


Johann-S (commented):
I don't see any interest to add semicolon to our ``Gruntfile.js`` because currently we do not use semicolon on our client side JS and in the others JS files

zalog:
There are IDE like Visual Studio Code that comes with ESLint by default and every time we open `bootstrap` folder/project we get this validation errors that mess up with other possible real problems :)
If we decide to ignore this problems, they should be added in `.eslintignore` file.

js/.eslintrc.json
```diff
@@ -72,7 +72,6 @@
     "no-new": "error",
     "no-new-func": "off",
     "no-new-wrappers": "error",
-    "no-new": "error",
```
Johann-S: Good catch

Johann-S:
Update ``.eslintignore`` should be a good way to do this.

@bardiharborow any concern about that ?

zalog:
Sorry, later edit to be more clear.

On our generated client side JS files, like `js/dist/`, there are semicolons. I'm not sure, but probably babel do that.
But semicolons are not in `js/src`, ESLint does not require that because of some specific settings in `js/.eslintrc.json`.

zalog:
Guys, do you want me to make another PR with the `no-new` duplicate key?
And rebase this PR without that and leave it here for further discussions?

bardiharborow:
PR for this coming soon.

bardiharborow:
Closing for #22477.

zalog:
Yes, but still https://github.com/twbs/bootstrap/pull/22388#issuecomment-292885868

bardiharborow:
Yep, I think maybe just moving the `.eslintrc.json` to the root of the project will fix that issue (from memory we were just trying avoid adding yet another file there).

zalog:
Yes, no other files, they are hard to maintain.
I'll try to make a new PR with this, this days.

PR #22395, opened 2017-04-09
Form updates: Bug fixes, docs update, and validation overhaul

### LINKED ISSUE #21581: Revisit stacked and inline form layout options
Folks often want to mix the standard form group and inline form options in the same form. I'll build some example layouts out and see what we can do to help.

### LINKED ISSUE #21922: Custom select border radius has default browser value when $enable-rounded is set to false
I'm using Bootstrap v4.0.0-alpha.6 on OS X 10.2.2.

Setting `$enable-rounded` to `false` instead of zeroing the value of border radius it allows to use browser default. However it works as expected when `$enable-rounded` is set to `true` and `$custom-select-border-radius` is set to `0`.

### LINKED ISSUE #22105: (v4) customize input height not working for IE 9 to 11
Just notice that the height for inputs  is commented out:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_forms.scss#L11
```
  // // Make inputs at least the height of their button counterpart (base line-height + padding + border)
  // height: $input-height;
```
I want to customize my input height to ```48 px```:
I do the following change: 
```
$input-line-height : 30px;
$input-padding-y: 8px:
$input-btn-border-width: 1px; 
base line-height + padding + border = 48 px
```
It works perfectly for Chrome + Firefox+ IE edge, but it does not work for IE 9-11. 

Plunker:
https://plnkr.co/edit/kJdhvFyAZmDVkeTATkiW?p=preview

Just open above link in Chrome and IE 11. You will find the difference.

Is there any workaround for this?

mdo:
- Changes horizontal input and button padding from `1rem` to `.75rem`.
- Closes #21970 by using the `$btn-focus-box-shadow` variable on form controls.
- Fixes #21922 by nuking `border-radius` if global radius is disabled.
- 7efe4dd Adds same focus shadow effect to the form validation mixin.
- Revamp input and select height calculations. Previously, `$input-height` and it's sm/lg variants didn't account for top and bottom `border-width`. This has been redone to utilize three total variables and one `calc()` function because of mixed units. Closes #22298, closes #22484 which brought the issue to light, and fixes #22105.
- Renames `$custom-select-sm-*` variables to put the `-sm` at the end, matching our other variables.

---

Additional commits addressed the following:

- Fixes #21581 with additional grid options
- Closes #22538 by cherry-picking the relevant bits into the new forms docs


mdo:
I've merged my form validation branch into this PR and as such have a huge update here.

- I've dropped all the previous form validation styles and docs for new ones based on HTML5 native form validation APIs and styles. This means there's support for `:invalid` and `:valid` psuedo-classes on inputs, selects, and textareas.

- Invalid and valid states are styled for textual inputs, checkboxes and radios, selects, textareas—and all their custom versions. I've done my best to _limit_ these styles for the time being so that they can more easily be built upon in the future, and customized by folks.

- There's example JS included in the docs example for how to disable browser validation and add your own custom styles.

I'm still working on this, but initial reviews from folks would be awesome. I've been tweeting around ideas for the last few days or so already, so I feel pretty confident in this direction at the very least.

Still to do:

- [ ] Remove unused Sass variables
- [ ] Add more Sass variables for the validation styles
- [ ] Wrap the validation styles in a global option variable so folks can enable/disable as they please
- [ ] Explore docs for customizing feedback message via JS with `setCustomValidity`
- [ ] Rearrange the form docs even more (trying to find the right flow—ideas would are welcome!)

But alas, peep some screenshots!

screen shot 2017-06-06 at 9 32 36 pm

screen shot 2017-06-06 at 9 32 48 pm

mdo:
Will follow up with restoring optional form validation icons in the inputs with #22762.

pkozlowski-opensource:
@mdo just to be clear that I'm reading this PR correctly: doest it meant that from now on styling will be based on internal state of native HTML controls? I'm wondering how this will work with JS frameworks that add their own notations of validation constraints and validity state (ex.: some frameworks will have a "pending" state where async validation is in progress).

If we can't style inputs by simply adding appropriate classes and are forced to interact with native HTML controls validity it might be very difficult to use Bootstrap with frameworks that don't reflect validity state in native HTML controls.

I guess the simple question is: will it be possible to style for valid / invalid state by simply adding appropriate CSS classes?  I'm seeing that this PR removes `.form-control-success` and friends, so getting worried here....

Francismori7:
Also, think about server side validation fails.. this pr wasn't well thought out IMO... Needs revisit

mdo:
@pkozlowski-opensource @Francismori7 I've pushed some changes:

- Added a new `form-validation-state()` mixin for generating the class names and styles.
- Added class fallbacks for `:valid` and `:invalid`, `.is-valid` and `.is-invalid`.
- Documented the new styles and fixed some docs examples.

Lemme know how that looks, please!

screen shot 2017-06-10 at 4 37 08 pm

Francismori7:
This does look good!

Thumbs up from me

Issue #22362, opened 2017-04-06
Modal "esc" key closes fullscreen browser


callumstrubi:
Pressing the "esc" key to close a modal will close the browser in macOS from fullscreen using the default browser action

Tested on:
macOS Sierra, Safari Version 10.1

Simple JSBin:
https://jsbin.com/buciferovu/edit?html,css,output

Proposed fix:
preventDefault(); added to modal close action when triggered from keypress

patrickhlauke:
sounds fair enough to me. targetting this for v4, as v3 is only in maintenance mode

bkdotcom:
Modal Feature Request:  Keyboard Arrow Key's changing modal button focus should be core functionality.    In addition, when a modal opens, it should  `.find("[autofocus]").focus()`

patrickhlauke:
> Keyboard Arrow Key's changing modal button focus should be core functionality. 

modals are used for more than just OK/CANCEL type dialogs, so this would likely not work in our more generalised model, i think

callumstrubi:
To specifically support @patrickhlauke here, we use the modal for our search implementation, any bootstrap focus intervention would kill that functionality.
A suggestion to get your focus to work though:
$("#modal").on('shown.bs.modal', function(){
      $(this).find('[autofocus]').focus();
});
This is how we focus on the search input when the search menu is initialised.
<img width="1440" alt="screen shot 2017-04-06 at 15 53 02" src="https://cloud.githubusercontent.com/assets/5045779/24760788/8e519096-1ae1-11e7-974e-e93700a9fd72.png">

bkdotcom:
>any bootstrap focus intervention would kill that functionality.

sorry to derail...
how so?  If you don't want something autofocused don't add the autofocus attribute.  If you do want something autofocused, add the autofocus attribute.   Autofocus on an input element inside a modal currently doesn't have any effect (unless implementor is listening for on show/shown and setting the focus).  I don't see any potential breakage or undesired behavior.

patrickhlauke:
i think @callumstrubi was referring to focus intervention in the context of the left/right cursor key to set focus to OK/CANCEL type buttons. agree with you that the autofocus thing is likely fine since it's up to the author to either add any autofocus attributes or not

bkdotcom:
final comment regarding keyboard arrows and buttons:
perhaps enabled via a data-attr on the modal?
`<div class="modal" data-keyboard-btn="true"></div>`

patrickhlauke:
[...]

PR #22398, opened 2017-04-09
Prevent default for ESC in modal.js

patrickhlauke:
ESC can be used to close modals, but on OS X/macOS this also jumps out
of full-screen mode. `preventDefault` suppresses this.

Closes https://github.com/twbs/bootstrap/issues/22362


mdo:
> OS X/macOS this also jumps out of full-screen mode.

Holy shit thank you, I hate when this happens.

Issue #22040, opened 2017-02-21
Radio buttons group: incorrect aria-pressed behavior


thbt:
When clicking on a [radio button in the v4 demo]( https://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons) and using chrome devtools, I noticed a weird behavior : 

- clicking on one of the radio button `Radio 2` of the group will set it's `aria-pressed` attribute to `true`.
- clicking on `Radio 3` will set it's `aria-pressed` attribute to `true`, and `Radio 2` will still have `aria-pressed=true` as well, even though it's not toggled anymore.
- clicking on `Radio 3` again will set it's `aria-pressed` attribute to `false`, even though it's still toggled.

It looks like the `aria-pressed` logic is the same as the one from the checkbox. Testing it with NVDA gives the wrong output (the checked button is not the intended one). I have yet to test it with JAWS and VoiceOver.

Is this the correct behavior ?

patrickhlauke:
There seems to indeed be something broken here. Fundamentally, `aria-pressed` should not be set at all on the `<label>` elements, as it makes no sense there - it's the visually hidden radio buttons that carry the checked/unchecked state.

I worked on some form of simplication of the whole checkbox/radio button logic ages ago...will try to dig this out and address the problem.

thbt:
> Fundamentally, aria-pressed should not be set at all on the <label> elements, as it makes no sense there

That's what I thought as well

joe-watkins:
@patrickhlauke I took a quick look at this and wondering if ARIA is needed at all on the `checkbox` and `radio` examples. AT will communicate checked state naturally because they are native elements and well supported.

If the roles of these elements were being changed I'd think this may be required but don't here.. what ya think?

patrickhlauke:
handling `aria-pressed` in the checkbox/radio case, if implemented the way we show in the documentation, is indeed superfluous. the JS here is only really there to set the correct CSS class to the `<label>` (in absence of a parent selector in CSS). i don't believe it's the intention for the CSS to handle arbitrary elements being used as radio/checkbox elements (as it would otherwise need lots more logic here and ideally only work with the appropriate `role="..."` elements etc).

i think fundamentally this whole "button groups for checkboxes/radio buttons" thing could be handled differently and split out/away from JS completely...but for now i've made an initial patch here https://github.com/twbs/bootstrap/pull/22400

PR #22400, opened 2017-04-09
Only change aria-pressed if it's not an input-based radio or checkbox group

patrickhlauke:
`aria-pressed="true"`/`aria-pressed="false"` is really only useful for making on/off toggles out of, say, `<button>` elements. the attribute is useless (and potentially confusing/conflicting) on, say, `<label>` elements for an existing `<input type="radio">` or similar.

closes https://github.com/twbs/bootstrap/issues/22040

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -66,6 +66,7 @@ const Button = (($) => {
 
     toggle() {
       let triggerChangeEvent = true
+      let addAriaPressed = true
       const rootElement      = $(this._element).closest(
         Selector.DATA_TOGGLE
       )[0]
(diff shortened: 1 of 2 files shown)


Johann-S (dismissed):
Maybe is it possible to add a unit test here ?

patrickhlauke:
@Johann-S ah yes, good call on unit tests. will get on that this morning

patrickhlauke:
@Johann-S mind having a look over the unit test i just added? admittedly, i'm not so hot on qunit nor clean jquery, so likely i'm going about this in an unnecessarily convoluted way

Johann-S:
It seems you let some trailing spaces see : https://travis-ci.org/twbs/bootstrap/jobs/220509346#L220

But your unit test is fine 👍

patrickhlauke:
@Johann-S cool. trailing spaces killed. will go ahead and merge. thank you :)

Issue #22416, opened 2017-04-12
document that dropdown-backdrop is shown for touch devices


SimonHeimberg:
Please update the documentation to report that the dropdown-backdrop is used on devices with **touch** support (and not only on **mobile** devices).
* on http://getbootstrap.com/javascript/#dropdowns-usage
* in the souce in [dropdown.js](https://github.com/twbs/bootstrap/blob/f2f805128508e82f0adc6e57b421dfb46d65a434/js/src/dropdown.js#L115)

And maybe there should be some discussion when the backdrop is necessary. For all touch devices or only some? Or only when the pointing device with touch support is used? Some toughts on [mozilla hack](https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/).

Background: Firefox reports touch support on my computer since 52.0 (indicated in note [2] on [touchstart](https://developer.mozilla.org/en-US/docs/Web/Events/touchstart)). (There is a touchpad and a trackpoint.) Because of a bug in my dropdown code, the backdrop was in front of the dropdown. Finding out why it happens on my computer and not on similar ones took much time. The documentation mentioning that the backdrop is enabled on touch support would have saved time.
Similar situation have happended in the past for others ([SO](http://stackoverflow.com/questions/30925699/hide-bootstraps-dropdown-backdrop-css-only), [SO](https://stackoverflow.com/questions/35612895/links-and-ng-clicks-in-bootstrap-dropdown-not-working-on-chrome-48-for-windows-8)).
Of course I will fix the z-order of the dropdown now. (I do not want to hide the backdrop, to support closing menu on real mobile devices.)

browser: firefox 52.0.2
os: windows 10, 64bit

patrickhlauke:
v3 is really just in maintenance mode, but i've made this PR against v4 https://github.com/twbs/bootstrap/pull/22422

patrickhlauke:
related, i may have found a way to completely avoid this backdrop altogether in v4 https://github.com/twbs/bootstrap/pull/22426

PR #22422, opened 2017-04-12
Reword "mobile" to "touch-enabled"

patrickhlauke:
...as touch is not exclusive to "mobile" anymore nowadays. also explicitly clarifies this is a fix for iOS, and that it impacts touch laptops etc as well

closes https://github.com/twbs/bootstrap/issues/22416

### DIFF
--- a/docs/components/dropdowns.md
+++ b/docs/components/dropdowns.md
@@ -480,7 +480,7 @@ Add `.disabled` to items in the dropdown to **style them as disabled**.
 
 Via data attributes or JavaScript, the dropdown plugin toggles hidden content (dropdown menus) by toggling the `.show` class on the parent list item.
 
-On mobile devices, opening a dropdown adds a `.dropdown-backdrop` as a tap area for closing dropdown menus when tapping outside the menu, a requirement for proper iOS support. **This means that switching from an open dropdown menu to a different dropdown menu requires an extra tap on mobile.**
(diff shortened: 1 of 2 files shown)


patrickhlauke:
@mdo more fundamentally, as this is really a fixup only necessary for iOS (and it causes issues on touch-enabled laptops like my surface 3, where - as expected - even if i use a mouse it means that once a dropdown is open, any click anywhere else on page is swallowed by the generated div), if we can just drop this whole backdrop (meaning iOS users would only be able to close the dropdown by explicitly tapping the dropdown opener, or opening another dropdown) or if we can somehow limit this to only kick in on iOS (ugly UA sniffing?). or, only if the menu was opened as the result of a touch (would require major recoding i think). thoughts?

patrickhlauke:
@mdo heck, maybe we could even programmatically add an explicit passive `touchstart` listener to the body in that chunk of code, instead of generating the div, and remove it once the dropdown is closed. would possibly be less intrusive. i'll have a play...but for now, this PR at least clarifies current behavior

patrickhlauke:
went ahead and merged this, as it seems fairly non-controversial to me. however, i believe i have a far slicker, more fundamental solution (indeed adding `touchstart` to body programmatically) that has far fewer side effects. that'll be in a separate PR which i WILL want reviewed to make sure i'm not losing my mind ;)

Issue #22196, opened 2017-03-15
Bootstrap 4 with npm and webpack


IdanCo:
I've been spending WAY too much time on adding bootstrap 4 with npm to a webpack project. In hope to save others the time, I thought i'll share the final recipe -

start by installing bootstrap in your project -
```
npm install bootstrap@4.0.0-alpha.6 --save
```
(notice bootstrap has two dependencies - jquery and tether. If you would rather have explicit versions of those two, you should install them as well)

import bootstrap's javascript through index.js -
```javascript
import 'bootstrap';
```
(i'm assuming you're using es6, in case of es5, use `require('bootstrap')`)

The previous line will only import the js part of bootstrap. for the styling part you have two options -

### 1. Precompiled SASS
Inside one of your sass files (index.scss for example) add this line -
```SASS
@import "~bootstrap/scss/bootstrap.scss";
```
(notice the ~ (tilde) which points to the node modules folder)
This mehtod is beneficial if you plan on using your own version of the wonderful __variables_ file bootstrap comes with. just make sure you import __variables_ before _bootstrap_. Also, now you can use all the cool mixins bootstrap has.

### 2. Compiled CSS only
If you're not planning on using the variables or the mixins, and prefer the precooked css file, simply add this line to index.js or any other js file -
```javascript
import 'bootstrap/dist/css/bootstrap.css';
```
(btw - you can also import this way the sass file, but it's nicer to import it via another sass file as shown in pervious mehtod)

now comes the webpack part. for _jquery_ and _tether_ to be available as a global variable throughout the project, you'll have to add this to your webpack plugins sections -
```javascript
new webpack.ProvidePlugin({ // inject ES5 modules as global vars
  $: 'jquery',
  jQuery: 'jquery',
  'window.jQuery': 'jquery',
  Tether: 'tether'
})
```
The different jquery definition are meant to answer requirements of different libraries (for example angular looks for 'window.jQuery'). I'm assuming your webpack.config already has rules for scss and/or css.

And that's it! now you have bootstrap in your webpack project.
Let me know if any further explanation is needed, and if anyone knows of a better way, please share.

Olivia101:
> `@import "~bootstrap/scss/bootstrap.scss";`
> (notice the ~ (tilde) which points to the node modules folder)

@IdanCo Do you need to install a dependency to be able to use `~` instead of the `node_modules` folder?
I tried it but I didn't work.

IdanCo:
[...]

PR #22423, opened 2017-04-12
add webpack to documentation

IdanCo:
## Summary
Add a guide to the official docs on how to integrate Bootstrap with Webpack.

## Motivation
- Webpack's popularity is growing, but there's a lack of reliable up-to-date resources on how to integrate it with Bootstrap
- Following @mdo question in #22196 

## Background
During the past few months i've been integrating bootstrap into several new and existing projects using webpack version 2. This led me to conduct a thorough research and tests on the best practices to achieve this integration. The results of these efforts are in this guide.

## Technical
Check out a sample project which implements the steps detailed in this guide -
https://github.com/IdanCo/webpack-modular/tree/bootstrap4


mdo (approved):
Looking at this, the only question I have is around the placement of this on a separate page—need to think about that a bit more. Also, is there anyone else out there using Webpack that would like to chime in with a review here?

docs/getting-started/webpack.md
```diff
@@ -0,0 +1,110 @@
...
+To enjoy the full potential of Bootstrap, use the source files as a part of your project's bundling process. Add this to your main sass file:
+
+{% highlight scss %}
+@import "~bootstrap/scss/bootstrap";
+{% endhighlight %}
+
+{% callout info %}
+#### Customization
+
+To implement the powerful [options](/getting-started/options/) Bootstrap offers, create `_custom.scss` and import it **before** Bootstrap.
```
ajacksified: Since you include examples elsewhere, may be nice to do the same here for clarity:

```
To implement the powerful [options](/getting-started/options/) Bootstrap offers, create a file for variables, such as `_custom.scss`, and import it **before** Bootstrap:

{% highlight scss %}
// theme.scss
$brand-primary: $green;
{% endhighlight %}

{% highlight scss %}
@import "theme.scss";
@import "~bootstrap/scss/bootstrap";
{% endhighlight %}
```

docs/getting-started/webpack.md
```diff
@@ -0,0 +1,110 @@
+---
+layout: docs
+title: Webpack
+description: Learn how to install Bootstrap using webpack 2
+group: getting-started
+---
+
+Use [webpack](https://webpack.js.org/) to bundle Bootstrap into your project.
```
ajacksified: May be worth mentioning that this is for webpack v2.

ajacksified (approved):
I have a few minor optional comments, but otherwise this very similar to my own webpack config.

docs/getting-started/webpack.md
IdanCo: @ajacksified thanks for the feedback! i'll push a new PR later

IdanCo:
@ajacksified , i've explicitly mentioned webpack v2 and put more emphasize on custom variables (after all, why even use webpack if you're not planning on customizing?)

ajacksified:
Sorry! Must have missed it.

ajacksified:
Disregard that, misread the notification email. Looks good to me!

samsch:
I would suggest using [ExtractTextWebpackPlugin](https://webpack.js.org/plugins/extract-text-webpack-plugin/) in the example, rather than style-loader. With style-loader, the styles are injected into `<style>` tags in the html head. With ExtractTextWebpackPlugin, the styles are extracted into a separate css file to be `<link>`ed in.

The downside is the slightly added complexity, but the upside is less "magic" around importing styles from JS.

IdanCo:
@samsch , I agree that using ExtractText in many cases would be the best practice. It's actually used it in the demo project I attached to this PR - https://github.com/IdanCo/webpack-modular/tree/bootstrap4

But I'm not convinced it should be a part of the documentation. There are other best-practices (such as using source maps), but adding them to the documentation might imply that they are **essential** for compiling Bootstrap, even though they are in no way such. Also, i can think of many use cases in which using ExtractText plugin would be counter productive, such as serving dev server or running tests.

I hope other resources (blogs, stackoverflow, etc.) will expand and elaborate this basic setup. I know I plan to...

samsch:
Just a quick counter argument. I'm actually less concerned with ExtractText
being a best practice, and more interested in having a separate css file to
be more similar to environments which users may be used to.

I do see your point about it seeming to be "required" when it is not.
Perhaps that could be satisfied with an additional sentence or two in the
guide noting that it's optional.

On Apr 24, 2017 4:44 PM, "Idan Cohen" <notifications@github.com> wrote:

@samsch <https://github.com/samsch> , I agree that using ExtractText in
many cases would be the best practice. It's actually used it in the demo
project I attached to this PR - https://github.com/IdanCo/
webpack-modular/tree/bootstrap4

But I'm not convinced it should be a part of the documentation. There are
other best-practices (such as using source maps), but adding them to the
documentation might imply that they are *essential* for compiling
Bootstrap, even though they are in no way such. Also, i can think of many
use cases in which using ExtractText plugin would be counter productive,
such as serving dev server or running tests.

I hope other resources (blogs, stackoverflow, etc.) will expand and
elaborate this basic setup. I know I plan to...

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/twbs/bootstrap/pull/22423#issuecomment-296815953>, or mute
the thread

mdo:
Any other feedback before we merge this? (Also, hi @ajacksified! Long time no see! :D)

Johann-S:
My only concern about this PR is, is it the role of BS to show how to integrate BS in any plugins, module bundler etc... ? 
I know Webpack have a huge community and I love this tool but if we do that we'll have to do for others plugins too so we should create a dedicated page for that on our documentation

mdo:
> My only concern about this PR is, is it the role of BS to show how to integrate BS in any plugins, module bundler etc... ?

I think so, yes. I'd love more in-depth and helpful guides for using, extending, and contributing to Bootstrap—regardless of the package manager or community. Bootstrap is meant to be accessible to everyone, beginner or seasoned veteran, so I'm a fan :).

IdanCo:
@mdo totally agree.

@Johann-S i see your point - Bootstrap can't be held responsible for what other bundlers are doing. But on the other hand, the bundlers also can't commit to how Bootstrap is structured (especially while in alfa). So the outcome is that integrations fall "between the chairs". So taking responsibility for documenting may not be the obvious choice, but it will help bring BS to more people, and i think that's worth it.

Regarding which bundlers should be included, IMO as long as there's a demand for integration documentation, it's worth considering adding.

Johann-S:
That's ok for me 👍 it's just a choice to do

cr101:
We are integrating Bootstrap into our web app with [Create React App](https://github.com/facebookincubator/create-react-app)

Johann-S:
I think we can add this page on : Write a **Learn & Extend** or **Approach** section.

Related to #17021 (docs section)

Johann-S:
@IdanCo can you update your branch ? because now we use Popper instead of Tether

IdanCo:
Done! and congrats on the huge Popper PR @Johann-S , can't wait to give it a try!

Johann-S:
Thank you very much @IdanCo ❤️

docs/getting-started/webpack.md
```diff
@@ -0,0 +1,106 @@
...
+Bootstrap is dependent on jQuery and Tether, so npm will install them for you if needed. But they must be explicitly provided by webpack. Add the following code to the `plugins` section in your webpack config file:
+
+{% highlight js %}
+  plugins: [
+    ...
+      new webpack.ProvidePlugin({
+        $: 'jquery',
+        jQuery: 'jquery',
+        'window.jQuery': 'jquery',
+        Tether: 'tether',
```
Johann-S: Can you change that by `Popper` ?

docs/getting-started/webpack.md
```diff
@@ -0,0 +1,106 @@
...
+
+Alternatively, you may **import plugins individually** as needed:
+
+{% highlight js %}
+import 'bootstrap/js/dist/util';
+import 'bootstrap/js/dist/dropdown';
+...
+{% endhighlight %}
+
+Bootstrap is dependent on jQuery and Tether, so npm will install them for you if needed. But they must be explicitly provided by webpack. Add the following code to the `plugins` section in your webpack config file:
```
Johann-S: Can you change that by `Popper` ?

Johann-S (changes requested):

docs/getting-started/webpack.md
IdanCo: On it

Johann-S (approved):

Johann-S:
LGTM 👍 

The merge is up to @mdo because he said : 
> Looking at this, the only question I have is around the placement of this on a separate page—need to think about that a bit more.

IdanCo:
Changed the webpack configuration to accommodate popper instead of tether.

@Johann-S , any reason why [bootstrap.js](https://github.com/twbs/bootstrap/blob/v4-dev/dist/js/bootstrap.js) in v4-dev is not yet updated and still dependent on thether? I built it manually to do my tests and find the correct recipe, but I'm curious about the repo build process...

BTW, at first glance the new tooltips/popovers/dropdown looks great!

IdanCo:
> Looking at this, the only question I have is around the placement of this on a separate page—need to think about that a bit more.

no problem, let me know if there's anything i can do to help

Johann-S:
Yes we build manualy `dist` files, currently they are outdated but we will update them soon 👍

docs/getting-started/webpack.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Installing Bootstrap
+
+[Install bootstrap](/getting-started/download/#npm) as a node module using npm.
+
+## Importing JavaScript
+
+Import [Bootstrap's JavaScript](/getting-started/javascript/) by adding this line to your app's entry point (usally `index.js` or `app.js`):
+
+{% highlight js %}
```
micky2be: Why don't you use the triple back tick?
```js
import 'bootstrap';
```

mdo: Because we use `highlight` everywhere else due to our Jekyll setup.

mdo:
Thanks y'all!

petrpacas:
@mdo I'd like to point out an issue of compiling the JS with Bootstrap's individual `/dist/` files (as hinted [here](https://github.com/shakacode/bootstrap-loader/issues/172#issuecomment-247164418)).

I tested @IdanCo's repo and it looks fine when using the `bootstrap.js` scripts bundle.

However, if you include the `/dist/` files individually (and necessarily reference them with `ProvidePlugin`), the code is duplicated in the resulting JS file (only referencing them in `ProvidePlugin` and not importing them doesn't work).

**The solution** for individual imports is, as suggested in the first link, to include the individual `/src/` files instead and use a transpiler (Babel etc.).

IdanCo:
Thanks @petrpacas , I'll have a look at it this weekend

mselerin:
Hi @petrpacas. Personally, I don't use the Provide Plugin but the Expose Loader instead (see my comment [here](https://github.com/twbs/bootstrap/issues/22196#issuecomment-288321383))

Maybe you could adapt your import with this loader and see what happen ?

petrpacas:
For those interested, here's my repo with working example of using the individual source files w/ Babel instead:

https://github.com/petrpacas/webpack-bootstrap-4-setup

ccalvarez:
Hi.  

I followed the [guide](https://v4.bootcss.com/docs/4.0/getting-started/webpack/) and for some reason it kept telling me that Tether was needed.  I had to replace Popper with Tether.

I've installed Bootstrap with `npm install --save bootstrap@4.0.0-alpha.6` and in my package.json the dependency is `"bootstrap": "^4.0.0-alpha.6",`.

How can I get the dependency to Popper as explained in the documentation?  
Thanks.

IdanCo:
@ccalvarez, These instructions refer to the develop branch which replaced _Tether_ with _Popper_.

For version Alpha 6 (which you probably use), You can browse [the previous commit](https://raw.githubusercontent.com/twbs/bootstrap/4a907b32573fc9b4b37cacab9538a380d80b5bbc/docs/getting-started/webpack.md), or have a look [in the issue discussion](https://github.com/twbs/bootstrap/issues/22196).

ccalvarez:
@IdanCo I understand, thank you very much!

chrisalexander55:
Although the documentation is very good, I still found it incomplete regarding the actual NPM packages that need to be installed and interop with other loader(s) who generate source-maps. Here is a Gist that brings it together:

[https://gist.github.com/chrisalexander55/81f3d5db057fbcbd74d62d488adbca8a](https://gist.github.com/chrisalexander55/81f3d5db057fbcbd74d62d488adbca8a)

Please feel free to use this to update the docs page.

PR #22426, opened 2017-04-12
Replace dropdown backdrop hack with cleaner JS-only hack

patrickhlauke:
As discussed in #22422 the current approach of injecting a backdrop (to work around iOS' broken event
delegation for the `click` event) has annoying consequences on touch-enabled laptop/desktop devices.

Instead of a backdrop `<div>`, here we simply add extra empty/noop mouse listeners to the immediate children of `<body>` (and remove them when the dropdown is closed) in order to force iOS to properly bubble a `click` resulting from a tap (essentially, method 2 from https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html)

This is sufficient (except in rare cases where the user does manage to tap on the body itself, rather than any child elements of body - which is not very likely in an iOS phone/tablet scenario for most layouts) to get iOS to get a grip and do the correct event bubbling/delegation, meaning the regular `click` event will bubble back to the `<body>` when tapping outside of the dropdown, and the dropdown will close properly (just like it already does, even without this fix, in non-iOS touchscreen devices/browsers, like Chrome/Android and Windows on a touch laptop).

This approach, though a bit hacky, has no impact on the DOM structure, and has no unforeseen side effects on touch-enabled laptops/desktops (nor non-iOS mobile/tablet devices). And crucially, it works just fine in iOS (tested on iPhone and iPad)

Additionally, this PR removes the styles associated with the now unnecessary dropdown backdrop, and updates the documentation (possibly in far too much "under the hood" detail?)

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -42,13 +42,11 @@ const Dropdown = (($) => {
   }
 
   const ClassName = {
-    BACKDROP : 'dropdown-backdrop',
     DISABLED : 'disabled',
     SHOW     : 'show'
   }
 
   const Selector = {
-    BACKDROP      : '.dropdown-backdrop',
     DATA_TOGGLE   : '[data-toggle="dropdown"]',
     FORM_CHILD    : '.dropdown form',
     ROLE_MENU     : '[role="menu"]',
@@ -108,16 +106,13 @@ const Dropdown = (($) => {
         return false
       }
 
-      // set the backdrop only if the dropdown menu will be opened
+      // if this is a touch-enabled device we add extra
+      // empty mouseover listeners to the body's immediate children;
+      // only needed because of broken event delegation on iOS
+      // https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
       if ('ontouchstart' in document.documentElement &&
          !$(parent).closest(Selector.NAVBAR_NAV).length) {
-
-        // if touch-enabled device we use a backdrop because click events
-        // don't delegate on iOS - see https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
-        const backdrop     = document.createElement('div')
-        backdrop.className = ClassName.BACKDROP
-        $(backdrop).insertBefore(this)
-        $(backdrop).on('click', Dropdown._clearMenus)
+        $('body').children().on('mouseover', '*', $.noop)
       }
 
       this.focus()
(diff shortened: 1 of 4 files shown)


js/src/dropdown.js
```diff
@@ -192,10 +187,10 @@ const Dropdown = (($) => {
...
         }
 
-        // remove backdrop only if the dropdown menu will be hidden
-        const backdrop = $(parent).find(Selector.BACKDROP)[0]
-        if (backdrop) {
-          backdrop.parentNode.removeChild(backdrop)
+        // if this is a touch-enabled device we remove the extra
+        // empty mouseover listeners we added for iOS support
+        if ('ontouchstart' in document.documentElement) {
+          $('body').children().off('mouseover', $.noop)
```
Johann-S: Here `$.noop` will continue to be called because the second parameter of ``off()`` is a selector see https://api.jquery.com/off/#off-events-selector

To check what I said you can test that : 
```javascript
var lala = function () { console.log('lala') }
var toto = function () { console.log('toto') }
$('body').children().on('mouseover', lala)
$('body').children().on('mouseover', toto)
```
And after try to do this : `$('body').children().off('mouseover', lala)`
You'll see what I explained before.

To do what you tried you should use this : https://api.jquery.com/off/#off-events-selector-handler
And you code will became 
```javascript
$('body').children().on('mouseover', '*', $.noop)
$('body').children().off('mouseover', '*', $.noop)

Johann-S (changes requested):

js/src/dropdown.js
patrickhlauke: crikey, i feel like a doofus. thanks for catching that. made the change and force-pushed the update. one of these days i may even go and learn some jquery...

Johann-S: You're welcome, happy to help you 😄

Johann-S (approved):

js/src/dropdown.js
patrickhlauke: although this is now merged and closed ... i actually got a follow-up here: it seems that even without the selector, it works as expected. when i ran your test in chrome just now, once i did the `$('body').children().off('mouseover', lala)` it actually removed the lala function correctly (so only toto was triggered).

further, am i right in thinking that `$('body').children()` and `$('body').children().find("*")` (which would be what is selected using your proposed change) are not the same thing, as i'm going one level deeper?

(asking because i'm about to add this exact same tweak to tooltip.js, and it made me look deeper into jquery, which is not usually my thing)

patrickhlauke: IF we want to keep the explicit selector, and target the children (rather than going one level down), would it actually be better to do something like `$('body').children().on('mouseover', null, $.noop)` / `$('body').children().off('mouseover', null, $.noop)` ?

Johann-S: Yes you're right I checked now and it works but when I checked `$('body').children().off('mouseover', lala)` didn't worked...
I made a mistake sorry about that :/

patrickhlauke: UNACCEPTABLE. no, seriously, no worries. i'll do a quick patch and still use the second parameter as selector, but as `null` (to make absolutely clear i'm wanting to bind directly to the child elements, and not do event delegation or some such)

Issue #22432, opened 2017-04-13
Wrong Color in Popover Arrow


IdanCo:
When popover direction is set to **bottom**, the arrow has the wrong color:

![image](https://cloud.githubusercontent.com/assets/10364499/24996862/27296536-203d-11e7-9949-76e59c6da489.png)

Couldn't find any related PRs, i'll push one soon...

Johann-S:
Closed by : #22433

IdanCo:
@Johann-S , the plot thickens -

After further research, I'm assuming the reason for the mixup with the color variable was intended to handle cases where the popover has a title:
![image](https://cloud.githubusercontent.com/assets/10364499/25005856/3b7a709a-2063-11e7-8ae6-372122c8efc1.png)

The problem was that it affected all bottom-popovers, even those without a title.

This conflict can be solved by using the already existing [title::before](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_popover.scss#L93) element to "cover" the existing arrow with the right color. I'll do a PR with this solution and let you decide what's best.

Johann-S:
Oh nice catch I didn't know that. Yeah sure create your PR but this time I'll CC @mdo because I don't know all the use case

PR #22436, opened 2017-04-13
cover popover arrow when title is present

IdanCo:
## Summary
Following issue #22432 and PR #22433 , change the arrow color of popover when title is present -
image

See it in action here - http://jsbin.com/tubuwu/edit?html,output

## Before
Arrow color was the same, regardless of the presence or absence of a title in the popover

## After
Arrow color is now dependent on the presence or absence of a title in the popover

## Technical
I've used an existing pseudo element of the title, originally intended to fix a gap between the popover and the arrow, and redesigned it to cover the arrow entirely with the title's background color.

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -89,16 +89,19 @@
       border-bottom-color: $popover-arrow-color;
     }
 
-    // This will remove the popover-title's border just below the arrow
+    // This will cover the popover's arrow with the title's background color
     .popover-title::before {
       position: absolute;
-      top: 0;
+      top: -($popover-arrow-width - 1);
       left: 50%;
+      z-index: $zindex-popover + 1;
       display: block;
-      width: 20px;
-      margin-left: -10px;
+      width: 0;
+      margin-left: -$popover-arrow-width;
       content: "";
-      border-bottom: 1px solid $popover-title-bg;
+      border: $popover-arrow-width solid transparent;
+      border-top-width: 0;
+      border-bottom-color: $popover-title-bg;
     }
   }
 


Johann-S:
X-Ref : https://github.com/twbs/bootstrap/pull/22433

Johann-S (changes requested):
Please can you fix that : https://travis-ci.org/twbs/bootstrap/jobs/221759321#L284

IdanCo:
Fixed 👍

Johann-S (approved):

IdanCo:
To be honest, this solution (or the previous one for that matter) is not the best i can think of. No doubt it works, but it lacks the simplicity and cleanness we've come to expect from Bootstrap. I'd like to suggest two alternatives -

## 1. Use `has-titlte` class
Since javascript is already injecting the popover, we can add ([here maybe](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/popover.js#L113)) a single line to toggle between the states -
```js
$tip.toggleClass('has-title', this.getTitle())
```
And the with a single line of css (instead of the [current dozen](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_popover.scss#L93)) set the arrow color:
```sass
&.has-title::after {
 border-bottom-color: $popover-title-bg;
}
```

## 2. Replace pseudo element with a real element
If the arrow will be an element instead of a psudo-element, it will be possible to use the [sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_selectors) to target the arrow like this -
```sass
.popover-title + .popover-arrow::after {
  border-width: $popover-arrow-width;
}
```
This [cannot be done](http://stackoverflow.com/questions/7735267/can-i-target-a-before-or-after-pseudo-element-with-a-sibling-combinator) with a pseudo element. I realize pseudo element for arrows were [intoduced recently](https://github.com/twbs/bootstrap/pull/17238), but since the markup is injected by js anyway, I don't see what is gained by using pseudo element, but this is one example of what is lost.

Thoughts?

mdo:
Thanks, but we'll leave it as-is without the extra element or color change. The white arrow is intentional.

IdanCo:
Sounds good to me, i'll just recommend in that case to remove [this rule](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_popover.scss#L93) completely.

PR #22437, opened 2017-04-13
Update modal.md

tmdh:
The code of "Live demo modal" doesn't work. I have updated it and now it works.


patrickhlauke:
the `id` issue is alsready fixed (as can be seen by actually looking at your diff https://github.com/twbs/bootstrap/pull/22437/files)

Issue #22406, opened 2017-04-11
docs - Which accessibility standard is Bootstrap 4 following? WCAG 2.0 A, AA or AAA?


changLiuUNSW:
We are using Bootstrap 4 to build public facing application and the key requirement is that the site must meet WCAG 2.0 AA compliant.
Just want to know which accessibility standard is Bootstrap 4 following. WCAG 2.0 A, AA or AAA?
It is better to clarify it in the documentation.
Thanks

bkdotcom:
https://www.w3.org/TR/WCAG20/#guidelines

> 1 Perceivable
> 
> 1. Provide text alternatives for any non-text content so that it can be changed into other forms people need, such as large print, braille, speech, symbols or simpler language.
> 2. Provide alternatives for time-based media.
> 3. Create content that can be presented in different ways (for example simpler layout) without losing information or structure.
> 4. Make it easier for users to see and hear content including separating foreground from background.
> 
> 2 Operable
> 
> 1. [Make all functionality available from a keyboard.](/twbs/bootstrap/issues/22402) 😃
> 2. Provide users enough time to read and use content.
> 3. Do not design content in a way that is known to cause seizures.
> 4. Provide ways to help users navigate, find content, and determine where they are.
> 
> 3 Understandable
> 
> 1. Make text content readable and understandable.
> 2. Make Web pages appear and operate in predictable ways.
> 3. Help users avoid and correct mistakes.
> 
> 4 Robust
> 
> 1. Maximize compatibility with current and future user agents, including assistive technologies.

Bootstrap is a frame work.  It makes it possible to create compliant websites, but compliance is still up to the developer.  It also allows you to create a website that is very much not in compliance.

patrickhlauke:
we strive towards making interactive controls/widgets expose all the correct roles/values/state properties (to satisfy, say, 4.1.2), to make things works with keyboard (2.1.1). Some of the default colours / their combinations do not meet color contrast requirements (so would fail 1.4.3 / AA).

but beyond that, as @bkdotcom noted, this is simply a framework. a set of tools. still up to the developer to make sure what they're building WITH it meets accessibility requirements. but for the most part (barring bugs of course), it should be at least feasible to use vanilla bootstrap and (provided you know what you're doing) build something A or even AA (depending on use of colours) compliant.

changLiuUNSW:
@patrickhlauke 
I totally agree that Bootstrap is a framework and developers can customize it to meet A or AA.
[...]

PR #22440, opened 2017-04-13
Rewrite getting started/accessibility docs

patrickhlauke:
A long overdue rewrite of the accessibility section - instead of the few
snippets of strangely superficial and out-of-context advice (skip links,
use correct heading levels), this tries to answer some of the
fundamental questions about "is Bootstrap accessible", with emphasis on
the fact that the final result will depend in large part on what BS is
applied to/on (since BS relies on the markup etc authored by
developers). This also sets out our ambition to have things work for
keyboard and assistive tech users, and that we strive to make all our
examples etc accessible and semantic.

closes https://github.com/twbs/bootstrap/issues/22406

### DIFF
--- a/docs/getting-started/accessibility.md
+++ b/docs/getting-started/accessibility.md
@@ -1,11 +1,11 @@
 ---
 layout: docs
 title: Accessibility
-description: A brief overview of Bootstrap's features and limitations for the creation of accessible content
+description: A brief overview of Bootstrap's features and limitations for the creation of accessible content.
 group: getting-started
 ---
 
-Bootstrap's aim is to provide an easy-to-use framework of ready-made styles, layout tools and interactive components which  allows developers to create web sites and applications that are not only visually appealing and functionally rich, but also accessible out of the box.
+Bootstrap provides an easy-to-use framework of ready-made styles, layout tools, and interactive components, allowing developers to create web sites and applications that are visually appealing, functionally rich, and accessible out of the box.
 
 ## Contents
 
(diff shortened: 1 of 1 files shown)


stephenbe:
May I suggest you add https://www.w3.org/TR/wai-aria-practices/ in  Additional resources.

I am so happy about this, accessibility is so important, especially when so many people use bootstrap. Many disabled users will thank you for all the efforts that is put into accessibility :+1:

patrickhlauke:
yeah, open to adding some more resources (was actually thinking https://www.w3.org/TR/aria-in-html/). i may push a few more updates to that effect to this PR

docs/getting-started/accessibility.md
```diff
@@ -1,57 +1,59 @@
...
 
-### Button groups
+### Structural markup
 
-In order for assistive technologies–such as screen readers–to convey that a series of buttons is grouped, an appropriate `role` attribute needs to be provided. For button groups, this would be `role="group"`, while toolbars should have a `role="toolbar"`.
+Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples which not only demonstrate the use of Bootstrap itself, but also illustrate appropriate semantic markup, including ways in which potential accessibility concerns can be addressed.
```
mdo: > Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples to demonstrate the use of Bootstrap itself and illustrate appropriate semantic markup, including ways in which potential accessibility concerns can be addressed.

docs/getting-started/accessibility.md
```diff
@@ -1,57 +1,59 @@
...
 
-In addition, groups and toolbars should be given an explicit label, as most assistive technologies will otherwise not announce them, despite the presence of the correct `role` attribute. In the examples provided here, we use `aria-label`, but alternatives such as `aria-labelledby` can also be used.
+### Interactive components
 
-## Skip navigation
+Bootstrap's interactive components – such as modal dialogs, dropdown menus and custom tooltips – are designed to work for touch, mouse and keyboard users. Through the use of relevant [<abbr title="Web Accessibility Initiative">WAI</abbr> <abbr title="Accessible Rich Internet Applications">ARIA</abbr>](https://www.w3.org/WAI/intro/aria) roles and attributes, these components should also be understandable and operable using assistive technologies (such as screen readers).
```
mdo: > Bootstrap's interactive components—such as modal dialogs, dropdown menus, and custom tooltips—are designed to work for touch, mouse, and keyboard users. Through the use of relevant [<abbr title="Web Accessibility Initiative">WAI</abbr> <abbr title="Accessible Rich Internet Applications">ARIA</abbr>](https://www.w3.org/WAI/intro/aria) roles and attributes, these components are also understandable and operable using assistive technologies (such as screen readers).

docs/getting-started/accessibility.md
```diff
@@ -1,57 +1,59 @@
...
 
-If your navigation contains many links and comes before the main content in the DOM, add a `Skip to main content` link before the navigation (for a simple explanation, see this [A11Y Project article on skip navigation links](http://a11yproject.com/posts/skip-nav-links/)). Using the `.sr-only` class will visually hide the skip link, and the <code>.sr-only-focusable</code> class will ensure that the link becomes visible once focused (for sighted keyboard users).
+Because Bootstrap's components are purposely designed to be fairly generic, authors may – depending on their specific use case – need to include further <abbr title="Accessible Rich Internet Applications">ARIA</abbr> roles and attributes, as well as JavaScript behavior, to more accurately convey the precise nature and functionality of their component. This is usually noted in the documentation.
```
mdo: Nixing the `depending on...` part here since `authors may need` should cover the one-off nature of these changes.

> Because Bootstrap's components are purposely designed to be fairly generic, authors may need to include further <abbr title="Accessible Rich Internet Applications">ARIA</abbr> roles and attributes, as well as JavaScript behavior, to more accurately convey the precise nature and functionality of their component. This is usually noted in the documentation.

docs/getting-started/accessibility.md
```diff
@@ -1,57 +1,59 @@
...
 
-{% callout danger %}
-Due to long-standing shortcomings/bugs in Internet Explorer (see this article on [in-page links and focus order](http://accessibleculture.org/articles/2010/05/in-page-links/)), you will need to make sure that the target of your skip link is at least programmatically focusable by adding `tabindex="-1"`.
+### Color contrast
 
-In addition, you may want to explicitly suppress a visible focus indication on the target (particularly as Chrome currently also sets focus on elements with `tabindex="-1"` when they are clicked with the mouse) with `#content:focus { outline: none; }`.
+Most colors that currently make up Bootstrap's default palette – used throughout the framework for things such as button variations, alert variations, form validation indicators – lead to *insufficient* color contrast (below the recommended [WCAG 2.0 color contrast ratio of 4.5:1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)) when used against a light background. Authors will need to manually modify/extend these default colors to ensure adequate color contrast ratios.
```
mdo: Need to fix the dashes here as noted in the main review comment. Should be `—` instead of ` – `.

mdo (changes requested):
Few copy changes. Rewrote some things here in the blockquotes and made some other suggestions where it was easier to do so.

Few writing style notes:

- We use Oxford commas for clarity.
- Try to limit using `not only`—it typically produces longer sentences.
- Breaks in sentences should use em dashes without spaces on either side, not en dashes. So `—` instead of ` – `.

Curious if, given the amount of docs we have, we should lint for these kind of things.

docs/getting-started/accessibility.md
patrickhlauke: Personally, I find the "no spaces around the dash" part a bit antiquated (and I'm fairly sure I've seen examples of spaces around dashes used in other parts of the docs), but fair enough...

patrickhlauke:
@mdo pushed up changes which I think cover your suggestions :+1:

mdo (approved):

PR #22442, opened 2017-04-14
Fix carousel "hover" behavior on touch-enabled devices

### LINKED ISSUE #20847: [Firefox] Touch event support has been re-enabled on Windows desktop
https://www.fxsitecompat.com/en-CA/docs/2016/touch-event-support-has-been-re-enabled-on-windows-desktop/

patrickhlauke:
Currently, `mouseenter`/`mouseleave` pause the carousel, but these listeners are not added if it's a touch-enabled device. This causes problems on touch-enabled laptop/desktop devices, or when a mouse is paired with, say, an Android phone/tablet, as then the mouse listeners are not added and the carousel does not pause as it should when the user interacts with the mouse.

This PR modifies the behavior: mouse listeners are always added. Additionally, on touch devices, an extra `touchend` listener is included. Mouse users on touch-enabled devices now can pause the carousel as expected. If the user interacts with the carousel with the touchscreen, the carousel is now paused for the duration of `this._config.interval` (plus an extra 500 milliseconds, just to make sure we always wait until the mouse compat events have definitely been fired - which is what made the original code necessary in the first place since touch events also fire a `mouseenter` compatibility event), and then the cycling is explicitly restarted.

This now allows touchscreen users some time to read/interact with the carousel slide, before cycling is resumed. And it fixes the broken behavior for mouse users on touch-enabled devices too.

Closes #20847 (but the problem is not limited to Firefox, same happens on, say, Chrome on a touch-enabled device)


patrickhlauke:
This PR does not contain a unit test, as I can't figure out how this new behavior could even be tested...any suggestions welcome...

mdo (commented):
Nice, sounds awesome! Sorry I can't review the JS closer—perhaps @Johann-S can weigh in?

Johann-S (approved):
LGTM 👍

patrickhlauke:
thanks @Johann-S @mdo

PR #22444, opened 2017-04-14
v4 - Remove Tether for Popper

Johann-S:
Remove Tether for Popper

### Reasons
- @FezVrasta and his community are very active
- Popper fix a lot of bug that Tether didn't handle (see listed issues below)
- Tether recommand the use of Popper.js

### TODO
- [x] Stop using Tether for Tooltip, Popover
- [x] Use Popper for Dropdown
- [x] Check dropup for Dropdown
- [x] Fix unit test (#22464)
- [x] Remove Tether as a dependency
- [x] Add `update` method to update position for Tooltip/Popover/Dropdown
- [x] Add `flip` attribute for Dropdown to disable the flip behavior of Popper.js
- [x] Add `fallbackPlacement` for Tooltip and Popovers
- [x] Update documentation
  - [x] Remove Tether
  - [x] Add that Dropdown now depend on Popper
  - [x] Add `offset` option in documentation for Dropdown

~Work in progress...~

Close : #19685, #19279, #22261, #22249, #22022, #21962, #19888, #17167, #17669


FezVrasta:
Note that Popper by default applies some padding (virtually, not CSS) to the container element (in practice, the tooltips will never touch the edges of the container/window).

I think it's the same behavior that the old Bootstrap v3 tooltips had, but I'm not 100% sure, it is worth to review if this is something you want to keep or if you prefer to disable this feature setting the `padding` option to `0` on both `preventOverflow` and `flip` modifiers.

js/tests/index.html
```diff
@@ -7,7 +7,7 @@
 
     <!-- jQuery -->
     <script src="../../docs/assets/js/vendor/jquery-slim.min.js"></script>
-    <script src="../../docs/assets/js/vendor/tether.min.js"></script>
+    <script src="../../docs/assets/js/vendor/popper.min.js"></script>
```
Johann-S: Hi @FezVrasta,

For an unknown reason my unit tests failed because it seems when I run ``grunt qunit`` on my bash Popper is undefined but as you can see I added Popper. 

Do you have any idea ? 

In `tooltip.js` we have this check : 
```javascript
  /**
   * Check for Popper dependency
   * Popper - https://popper.js.org
   */
  if (typeof Popper === 'undefined') {
    throw new Error('Bootstrap tooltips require Popper (https://popper.js.org)')
  }
```
And it's this exception which failed tests

Johann-S: I found the reason why...
I didn't take the ES5 version of Popper... Sorry @FezVrasta 😄

Johann-S:
It seems so tests are failing on iOS see : https://github.com/FezVrasta/popper.js/issues/210

FezVrasta:
Is there any way to see the full log of the iOS9 tests?

Johann-S:
I tried our tests in SauceLabs with an iOS simulator on iOS 9.3 and everything works fine... 😩 
https://saucelabs.com/jobs/284bc995f8274d51b57b6f6fdeeada94/0004screenshot.png

FezVrasta:
Now it failed on IE10? Seems like you have some flaky test

Johann-S:
All tests passed but for an unknown reason Travis failed to close connection with Saucelabs...
https://travis-ci.org/twbs/bootstrap/jobs/222521829#L420

FezVrasta:
😂😂😂 it's still a progress

FezVrasta:
I can't see iOS in the logs, isn't that one that's failing?

js/src/dropdown.js
```diff
@@ -1,3 +1,5 @@
+/* global Popper */
```
FezVrasta: Just wondering, why do you need to set it as global? Wouldn't it cleaner to import it with ESM and mark it as external dependency in your modules resolver?

Johann-S:
Yes I don't see any iOS test in log, strange behavior...

js/src/dropdown.js
Johann-S: I'll tackle that later, they are a lots of work to do in this PR 👍

Johann-S:
Again on iOS : https://travis-ci.org/twbs/bootstrap/jobs/222776929#L411 
But when I tested manually on SauceLabs everything works...

FezVrasta:
@Johann-S can you enable the screen recording of SauceLabs and see what's happening there?

Johann-S:
Thank you @FezVrasta I'll take about that to the team 👍

FezVrasta:
Looks like screen recording is already on, you just have to visit:
https://saucelabs.com/beta/users/bootstrap-prs

And find the iOS test that failed.

I can't do it by myself because the user page is available only to people with the login credentials. (I think you have them since you are running manual builds?)

Johann-S:
The only failed test found is this one : https://saucelabs.com/beta/tests/604d33a88ea947e89df0e2fd11d65b4e/watch

But no iOS 😩

FezVrasta:
Does Bootstrap v4 support Safari 5? 😵

Johann-S:
No only those browsers : https://github.com/twbs/bootstrap/blob/v4-dev/grunt/sauce_browsers.yml

But maybe they are something wrong in this file

wolfy1339:
Bootstrap v4 doesn't support Safari on Windows, per http://v4-alpha.getbootstrap.com/getting-started/browsers-devices/#desktop-browsers

FezVrasta:
So why is there that test in the SauceLabs screen records?

Also, is there any way to output the logs from the failing browser? They look like black boxes to me 🤕

Johann-S:
After a lot of tests my conclusion is, currently we thought our test were runs on iOS but it wasn't the case it was run on Win7 and Safari 5...

I'll do some research about how to test on iOS with SauceLabs and if it's possible because I'm not sure about that

FezVrasta:
Take a look at the AngularJS repository, they do test on iOS.

FezVrasta:
It's still running on Safari 5 on Win 7:
https://saucelabs.com/beta/tests/6d948b0f09f44691b26e4b536d79f539

Johann-S:
Build passed ! Woot woot ! 💪

js/src/tooltip.js
```diff
@@ -13,11 +13,11 @@ import Util from './util'
...
   /**
-   * Check for Tether dependency
-   * Tether - http://tether.io/
+   * Check for Popper dependency
+   * Popper - https://popper.js.org
    */
-  if (typeof Tether === 'undefined') {
-    throw new Error('Bootstrap tooltips require Tether (http://tether.io/)')
+  if (typeof Popper === 'undefined') {
+    throw new Error('Bootstrap tooltips require Popper (https://popper.js.org)')
```
FezVrasta: Not to be picky, but the full name is "Popper.js" 😇

Johann-S: Sure I'll change that 😉 
I have to update those lines : https://github.com/twbs/bootstrap/blob/v4-dev/grunt/npm-shrinkwrap.json#L3795-L3798

Do you know the correct url for `resolved` ?

FezVrasta: I think it's `https://registry.npmjs.org/popper.js/-/popper.js-1.8.1.tgz`

docs/components/popovers.md
```diff
@@ -255,16 +255,10 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
...
-      <td>An array of constraints - passed through to Tether. For more information refer to Tether's <a href="http://tether.io/#constraints">constraint docs</a>.</td>
-    </tr>
-    <tr>
       <td>offset</td>
-      <td>string</td>
-      <td>'0 0'</td>
-      <td>Offset of the popover relative to its target. For more information refer to Tether's <a href="http://tether.io/#offset">offset docs</a>.</td>
+      <td>number</td>
+      <td>0</td>
+      <td>Offset of the popover relative to its target. For more information refer to Popper's <a href="https://popper.js.org/popper-documentation.html#new_Popper_new">offset docs</a>.</td>
```
FezVrasta: It accepts both a number or a string as explained here:
https://popper.js.org/popper-documentation.html#Modifiers.offset

js/src/dropdown.js
Johann-S: Currently they are no global solution for that issue (see : https://github.com/twbs/bootstrap/issues/17201#issuecomment-274243781).
So for now I'll pass on that.

Maybe @bardiharborow have some feedbacks ?

docs/components/dropdowns.md
```diff
@@ -527,7 +533,35 @@ Regardless of whether you call your dropdown via JavaScript or instead use the d
...
+      <td>
+        <p>How to position the popover - top | bottom.</p>
+        <p>When a function is used to determine the placement, it is called with the popover DOM node as its first argument and the triggering element DOM node as its second. The <code>this</code> context is set to the dropdown instance.</p>
+      </td>
+    </tr>
+    <tr>
+      <td>offset</td>
+      <td>number | string</td>
+      <td>0</td>
+      <td>Offset of the dropdown relative to its target. For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#Modifiers.offset">offset docs</a>.</td>
```
FezVrasta: I updated the docs, now they should be much clearer. This is the right URL to refer to these options:
https://popper.js.org/popper-documentation.html#modifiers..offset.offset

Sorry for the broken URL :-(

Johann-S: Haha that's ok, I'll update our doc 👍

docs/components/dropdowns.md
```diff
@@ -527,7 +533,35 @@ Regardless of whether you call your dropdown via JavaScript or instead use the d
...
+      <th style="width: 100px;">Type</th>
+      <th style="width: 50px;">Default</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>placement</td>
+      <td>string | function</td>
+      <td>'right'</td>
```
wojtask9: shouldn't be default 'bottom' (according to source)

Johann-S: yes nice catch @wojtask9 ! 👍

js/src/dropdown.js
FezVrasta: I need some context because I'm not confident with the Bootstrap v4 code base, what do you use to bundle the package and convert `import`s?

[Webpack](https://webpack.github.io/docs/library-and-externals.html), [Rollup](https://github.com/rollup/rollup/wiki/JavaScript-API#external) and [Browserify](https://benclinkinbeard.com/posts/external-bundles-for-faster-browserify-builds/) allow to define a dependency as external, thus avoiding to bundle it together with your own source code.

Johann-S: We just build our JS files with Babel that's all

FezVrasta: Babel is a transpiler, it doesn't handle the bundling of the resources AFAIK 🤔

wojtask9: bootstrap uses grunt-contrib-concat 
see
https://github.com/twbs/bootstrap/blob/v4-dev/Gruntfile.js#L96

FezVrasta: I see, but I don't understand how does it handle the `import` and `export` logic...
It first transpiles each module to Require.js and then it concats them together right? But where?

Johann-S: Each module is compiled by Babel and grunt-contrib-concat concatenate each files in boostrap.js and at the end we use uglify for minify our JS.

Import and export are native in ES6

Johann-S:
I think we make the good decision to remover Tether for Popper.js see : https://github.com/HubSpot/tether/issues/252#issuecomment-294961650

js/src/dropdown.js
wojtask9: @FezVrasta 
look at https://github.com/twbs/bootstrap/blob/v4-dev/.babelrc
in short it doesn't transform and removes import/export declarations.

It simply works :)

scss/_popover.scss
```diff
@@ -153,8 +149,8 @@
 //
 // .popover-arrow is outer, .popover-arrow::after is inner
 
-.popover::before,
-.popover::after {
+.arrow::before,
```
wojtask9: this should be moved  to `.popover` because this will overlaps with  `.arrow` from `.tooltip`

scss/_popover.scss
```diff
@@ -21,69 +21,66 @@
...
 
-  &.popover-top,
-  &.bs-tether-element-attached-bottom {
-    margin-top: -$popover-arrow-width;
+  &.bs-popover-top {
+    margin-bottom: $popover-arrow-width;
 
-    &::before,
-    &::after {
+    ::before,
```
wojtask9: this is inconsistent. In this file sometimes we use `.arrow` and sometimes pseudo-elements to style "arrows"

js/src/dropdown.js
```diff
@@ -55,6 +64,21 @@ const Dropdown = (($) => {
     VISIBLE_ITEMS : '.dropdown-menu .dropdown-item:not(.disabled)'
   }
 
+  const AttachmentMap = {
+    TOP    : 'top',
```
wojtask9: shouldn't be better if default will be 'top-start' and 'bottom-start'.
For my eyes centered dropdowns looks wired :)

wojtask9 (commented):
great work @Johann-S and @FezVrasta (for popper.js)

I've tested a little bit (with project based on BS4) and here my thoughts:
a) BS should be consistent with .arrow styling for tooltip and popover.
In tooltip we use .arrow and in popover we use pseudo-elements.
Popper.js adds [x-placement] attribute so this attribute can be used too.

Not sure what is the best choice :)

b) Default placements for dropdowns are 'top' and 'bottom'.
For these placements dropdown position is centered to element.
It looks a bit wired for me.
Shouldn't be changed to 'top-start' and 'bottom-start' and behave like current (before popper.js) BS dropdown?

c) popovers and tooltips 
When tooltip or popover flips (because there is not enough space) position classes are not updated.
This leads to wrong "arrows". There are two cases:
-on initial 'show'
-on scroll
This could be resolved by using callback onCreate and onUpdate (adding/removing classes) or use [x-placement] attribute.
Maybe there is another solution using modifier.flip option?

d) .scss changes are breaking for me :) so maybe this PR should target 'beta' not 'beta2'?  
Or remove bs-* prefix and style "arrows" using pseudo-elements then this can go 'beta2' :)

Johann-S:
Thank you very much for your great review @wojtask9 ! I'll update my PR 👍

FezVrasta:
About point C of wojtask9's comment:

You may also define a custom modifier that runs on each update, it would look like:

```
const updateClasses = {
  order: 1000,
  enabled: true,
  function: function(data) {
    // use `data.placement` to update your classes
    return data;
  }
}
```

Johann-S:
Why not with this way @FezVrasta  ? 
https://github.com/twbs/bootstrap/blob/8360ed1525b25377b1a3fc943328a49b978fe5a8/js/src/tooltip.js#L284-L297

FezVrasta:
@Johann-S it works as well, it's a matter of preferences.

alecpl:
I'm all for this change. Thank you. I think this should be included in the first beta.

js/src/tooltip.js
```diff
@@ -636,6 +620,20 @@ const Tooltip = (($) => {
...
+    _cleanTipClass() {
+      const $tip = $(this.getTipElement())
+      const tabClass = $tip.attr('class').match(BSCLS_PREFIX_REGEX)
+      if (tabClass !== null && tabClass.length > 0) {
+        $tip.removeClass(tabClass.join(''))
+      }
+    }
+
+    _handlePopperPlacementChange(data) {
+      if (data.originalPlacement !== data.placement) {
```
wojtask9: this `if` should be removed because it omits update if popper.js places tooltip again to default position.
Then data.orginalPlacement === data.placement but we changed class before.

@FezVrasta 
What is purpose of `data.flipped` flag?
According to my tests it's equivalent to `data.originalPlacement !== data.placement` but it will be more valuable if `data.flipped` will store information when "flip" happens during current update.

@Johann-S 
the same change should go for popover (onCreate, onUpdate)

FezVrasta: `flipped` is used by the `flip` modifier to avoid "flip loops"

Johann-S: @wojtask9 This change work on Popover too because Popover extend from Tooltip.

And the `if` shouldn't be removed because on `hide` for Popover and Tooltip we destroyed Popover instance

js/src/tooltip.js
```diff
@@ -288,19 +281,24 @@ const Tooltip = (($) => {
...
-          addTargetClasses: false
+        this._popper = new Popper(this.element, tip, {
+          placement : attachment,
+          modifiers : {
+            offset : {
+              offset : this.config.offset
+            }
+          },
+          onCreate : (data) => {
+            if (data.originalPlacement !== data.placement) {
```
FezVrasta: Wouldn't be cleaner to manage the classes only here instead of attaching them outside and then having to check that they are still correct?

js/src/tooltip.js
wojtask9: @Johann-S 
you are right. 
There is something wrong with my rollup configuration :/

js/src/tooltip.js
Johann-S: It cause some issues because the Tooltip or Popover is already shown and our CSS apply some styles because of that Popper do not calculate the good position for our Tooltip/Popover

FezVrasta: What if you keep the tooltip invisible until Popper finished doing its stuff? In any case, showing the tooltip while it's still being positioned isn't something desirable.

Johann-S: This case only happened if the Tooltip/Popover flipped.
And at the end we should also managed the add of `.hide` and the remove of this class during Popper calculations.
As this case isn't usual I won't change that for now

Johann-S:
@FezVrasta I updated Popper.js to 1.8.2 and I chose the UMD release, is it a good choice for BS ? Because before this release we only have to choose the ES5 release

FezVrasta:
@Johann-S the UMD release is compatible with a wider set of environments and is the same as the old "ES5" release, the ESM release is compatible only with the EcmaScript Modules, which are supported by Safari, Webpack and Rollup AFAIK.

FezVrasta:
Is there any additional work needed before this PR can be merged?

Johann-S:
Yes I want before the merged of this PR a review from @mdo and @bardiharborow + I want to clarify documentation for Dropdown (#22479)

Johann-S:
@FezVrasta @wojtask9 I add an option to disable `flip` behavior for Dropdown and update documentation if you have any time for a review do not hesitate 😄

FezVrasta:
it looks good!

FezVrasta:
Is in this PR scope to support the additional placement variations provided by Popper.js?

For instance, each placement can have `-start` and `-end` variations (`bottom-start`, `top-end` etc)

719media:
@Johann-S I've closed #19692 in preference of this issue/pull, as this is much more all-encompassing. And popper is nicer than tether for sure. I haven't built the pull request branch to check, but one thing that would be nice is if something like <form><.dropdown /></form> will still work (i.e. when the <.dropdown-menu> gets absolutely positioned and attached to the <body> when it opens, the <.dropdown-menu /> gets "put back" to the original DOM position on dropdown close, such that any <inputs> will be included in a <form> submit).

FezVrasta:
Popper.js doesn't move the posiotioned element to body so I don't think it will be a problem

Johann-S:
@FezVrasta I stop to add feature to this PR except bug fix or documentation improvement. But yes after the merged of this we are open for PRs that handle all the placements provided by Popper.js 😉 

@719media I encourage you to check what you reported on my PR and let me know if they are a bug or a bad behavior 👍

wojtask9:
@Johann-S 
One more thing. It would be great if we can disable popper.js positioning in dropdown (maybe `placement: static` ?).
Sometimes people don't want use dynamic positioning. For example dropdown in navbar, notification drawlers etc.
Maybe this can be achieved with combination of popper modifiers. If yes then note in docs would be great

Johann-S:
@wojtask9 already done : Add flip attribute for Dropdown to disable the flip behavior of Popper.js

I update my first comment with all the things I added

If I misunderstood what you said don't hesitate to @ me 👍

wojtask9:
It's hard to explain :) because I don't know where the problem is (it can be bug in popper.js or desired behaviour)

https://jsfiddle.net/gp2b6zc7/

If dropdown button is near bottom edge of browser window try open dropdown.
Usually (but not always and probably depends on browser scroll) it appears in correct position.
Next without closing dropdown-menu scroll mouse wheel. You will see inconsistent behavior (position is not correct).

Screenshots:
correct (because `flip:false`) 
image

wrong (`flip: false`) (after scroll)
image

FezVrasta:
If you see incorrect positioning is always Popper.js' fault. But I can't reproduce the problem you describe in your fiddle unfortunately. May you record a small clip (using Licecap or [kap](https:/geetkap.co)) to show how to reproduce it?

wojtask9:
@FezVrasta
As you can see when first time clicked on dropdown position is correct.
Second click is incorrect (in my opinion :) ) 
It happens on FF/Chrome (Linux and Windows)

dropdown2

FezVrasta:
Umh, I still can't reproduce it... I'm on macOS tho

wojtask9:
After scrollbar appears you must scroll mouse wheel.
This is reduced testcase. With navbar and very long dropdown I can achieve this situation without scrolling.

FezVrasta:
I see now, I'll investigate.

wojtask9:
@FezVrasta 
Try this. 
I've added more dropdow-items and script that should trigger behavior on my computer
`$('#dropdownMenu1').click();
 window.scrollTo(0,1000);`

https://jsfiddle.net/gp2b6zc7/1/

wojtask9:
@Johann-S 
another issue with tooltip. I can't reproduce this with 
https://popper.js.org/tooltip-examples.html 
so this must be bootstrap specific.

If you move fast mouse sometimes tooltip appears in wrong position
https://jsfiddle.net/gp2b6zc7/3/

tooltip

This happens only if you move fast mouse.

Johann-S:
Ok I found why it's because the `fade` animation you have a duration to show the tooltip...

If you do this you wont see this effect : 
```javascript
$('[data-toggle="tooltip"]').tooltip({
  animation: false
})
```
I'll try to find a way to avoid that effect

Johann-S:
@wojtask9 I fixed what you reported thank you 👍

wojtask9:
@Johann-S 
Great it works now :) Thank you.

One another issue :/
Tooltip adds margins to all classes .bs-tooltip-{left,right,top,bottom}.
This almost works beside .bs-tooltip-right (see https://github.com/twbs/bootstrap/issues/22300)
Right tooltip have unnecessary margin. 

I think a proper fix is to add ` margin: $tooltip-margin;` to `.tooltip` class and remove all `margin-{top,right,bottom,left}` from `bs-tooltip-{left,right,top,bottom}`

see https://jsfiddle.net/gp2b6zc7/11/

Johann-S:
Nice fix @wojtask9 ! I'll add this 👍 

And I'll add back `auto` placement because it was remove on Alpha 5 (see : #19796)

Johann-S:
It didn't work what you suggest @wojtask9, in your example it works because you added
`margin: 0 !important;`

wojtask9:
Yeah I've added !impartant because I'm lazy :)
If you remove margin-left,margin-right,margin-top,margin-bottom from all `bs-tooltip-{left,right,top,bottom}` then you can remove !important.

Value `0px` is only for test.
Boostrap v3 adds default space between button and arrow too (http://getbootstrap.com/javascript/#tooltips)

my updated example with removed margins and without !important

https://jsfiddle.net/gp2b6zc7/12/

Johann-S:
I'm sorry @wojtask9 but in your JSFiddle Tooltips aren't stick to there button

Johann-S:
@FezVrasta sorry to bother you but I tried to add `auto` placement to our Tooltip plugins but Tooltips isn't placed well maybe they are something bad I made in how to use Popper.js

FezVrasta:
There's some small annoyance with the `auto` placement at the moment, maybe you hit that problem.
There's a PR that should improve the `auto` detection but it's blocked because of missing CI test.
https://github.com/FezVrasta/popper.js/pull/235

If you have time to add one it would be great. Unfortunately these days I'm quite busy.

wojtask9:
@Johann-S 
Yes :) because of `margin:3px`
in `scss` there will be `margin: $tooltip-margin;`
default value from _variables.scss is `3px !default` so someone who want to change that must set to `0px`

For me $tooltip-margin is orthogonal to $tooltip-arrow-width and indicates space between tooltip+arrow from element.

Not sure what the default value should be. If we want tooltips arrows sticks to elements then default should be `0px`.
In Bootstrap v3 tooltip have space between element na arrow so it's up to you what default should be.

My fix proposal is about margin consistency. With current code right tooltip doesn't stick to element but left, top, bottom stick correct.

FezVrasta:
Note that you can define the margin in Popper.js...

```
new Popper(el1, el2, {
 modifiers: {
  offset: { offset: "0, 5px" },
 },
});
```

Johann-S:
@FezVrasta yes it's already set to `0` by default 👍 

@wojtask9 It's not a good fix for me because what we want is, each tooltips must stick to their button

FezVrasta:
@Johann-S you can define the margin there (`"0, 5px"`) and get rid of any margin from the CSS.

wojtask9:
@FezVrasta 
my proposal is about exactly about the same

here is diff 

https://gist.github.com/wojtask9/ec29db8435893dff63afcb22046e6193

FezVrasta:
@wojtask9 my proposal is a step further, get rid of the margin from the CSS at all, and delegate it to Popper.js. If you prefer to manage it from CSS, yours is a good solution.

Johann-S:
Thank you @wojtask9 I implemented what you did and it works well 👍 💪

wojtask9:
@FezVrasta 
I created issue (https://github.com/FezVrasta/popper.js/issues/265) for comment https://github.com/twbs/bootstrap/pull/22444#issuecomment-300834292

@Johann-S 
It looks now really great.
One thing
https://jsfiddle.net/gp2b6zc7/12/

if you open tooltip and scroll down and up you will see that tooltip 'jumps'.
This happens because `fallbackPlacement`. If I remove this then everything is OK (no wired jumps).
Another reason to remove `fallbackPlacement` is wired placement of tooltips :)
Wired means that currently tooltips don't flip opposite direction.
For example tooltip-top goes tooltip-left. For me natural way will be top goes down, down goes top, left goes right, etc :)

Johann-S:
@wojtask9 I didn't removed `fallbackPlacement` because it allows folks to customise the flip behavior but I set `fallbackPlacement` to `flip` the default value used by Popper.js

Thank you for the feedback you helped me a lot 👍

Johann-S:
Special thanks for @FezVrasta and @wojtask9 who helped me a lot 👍 

We made awesome works together 💪

valleyspirit:
"@FezVrasta and his community are very active"
"Popper fix a lot of bug that Tether didn't handle"

Aside from the bad English, I did not find this helpful. I found popper in v4-dev by looking at the github pulse, not by issues history. This was a bad first start. Then I noticed that the documentation on the v4 site was horribly misaligned with v4-dev, and the live documentation included hipster-ipsum (tight pants!) and inconsistent styles, and poor English. Then I found grammar errors in the merge referenced in the pulse. Next, I located a comment by mdo from December, where he said he would pass on switching to popper for what were apparently no reasons. I then went to https://popper.js.org/ on my mobile phone and was greeted by a dancing yellow popper tooltip pinned to the top of the site (which begged me to star the github project while blocking text), and noticed that when I scroll down on my shiny new android, the popup twitches and dances all over my screen, as if it's having a fit of content-blocking vibrational joy. So, there is my introduction, a very very bad one.

With no specifics it certainly ~appears~ that the developer of popper.js has simply had a pull request denied and then successfully rammed his pet project into bootstrap. I hope the behavior in bootstrap will be considerably better than the behavior on the developer's own site.

It took some work on my own to realize that there may be good reasons for the change, no help from this repo. My research continued and I discovered that popper.js is actually being recommended in the github project readme for tether.js, since HubSpot is not able to maintain tether. So, that is a great reason to migrate. But I had to find that on my own.

I have yet to find any list of problems which popper has "fixed", or any reasons for why the change was made, here on the bootstrap site. Those who are administrating v4-dev branch changes and tickets may want to bother including the rest of the world in their thinking. This experience has left me with mixed feelings about a significant change, while I am actively busy developing with bs4.

That being said, I have a slightly better understanding of why the change was made, to popper. I hope it's easily excluded since, like the developer, I also work with view engines that use a virtual DOM, and the twitching behavior on mobile android certainly won't make it past my boss or QA. 

It would have helped me as a bootstrap developer to have found any documented reasons for a change to popper while browsing issues regularly in the last month, instead of finding a questionable merge in v4-dev accompanied by new lows in the live documentation pages.

FezVrasta:
Hi @valleyspirit, you can find all the related issues linked to this PR right above your message. I hope it helps

Johann-S:
Thank you @valleyspirit for your opinion and feedbacks.
You're welcome if you want to submit a PR to fix my bad English in the documentation 😉 
And I'm a bit sad for you because at the end of my first comment you have a list of fixed issue by this PR + all the information you found by yourself are available in the comments of this PR

cr101:
@Johann-S  I'd like to thank you for your time and the awesome work you are doing on this awesome open-source project. Much appreciated.

Johann-S:
Thank you @cr101 much appreciate ❤️

docs/components/tooltips.md
```diff
@@ -136,7 +140,7 @@ You should only add tooltips to HTML elements that are traditionally keyboard-fo
 
 <!-- Generated markup by the plugin -->
 <div class="tooltip tooltip-top" role="tooltip">
```
Lausselloic: need to rename ```tooltip-top``` to ```bs-tooltip-top``` if I've understand other changes

Johann-S: nice catch @Lausselloic feel free to open a PR 👍

mdo:
@valleyspirit I appreciate the feedback as well, but there are a few things we should clarify about this PR.

- You're calling out someone's English language skills who's first language isn't English. There likely wasn't any malice to that, but it didn't seem like there was any _understanding_ either. Help @Johann-S and myself out by opening an issue or a PR to suggest changes, please.

- We don't develop Bootstrap like you think we do. The development branch, in this case `v4-dev`, always lags behind the docs. We only update our hosted docs when there is a release to publish. Doing otherwise would be chaotic for us and our users.

- Yes, I've previously turned down the move to Popper, but if I recall correctly that issue was from last year when I didn't have enough time or experience of my own to push it through. Changing dependencies is no small matter, and the project lacked the time from team members to make it happen. With @johann-s coming aboard, we've been able to tackle _much_ more. Naturally we've revisited some of these older decisions to help us release the most stable v4 we can.

- Speaking of which, v4 is still in alpha. Yes, it sucks that it is and it's absolutely crazy it's taken this long. You're developing on it at your own risk—anything and everything is subject to change until we publish our first beta.

When we publish our first beta, we'll have a blog post, updated documentation, and more to share with folks. Until then, if you're using the bleeding edge changes from `v4-dev`, expect some crazy changes and brace yourself.

PR #22445, opened 2017-04-14
Make "Copy" buttons actual buttons

patrickhlauke:
This makes them keyboard-accessible.

For mouse users, the only change here is that the focus outline will remain on the button once clicked (in future, this can be solved with :focus-ring, but for now this would require a polyfill).

the tooltip is explicitly hidden on `mouseleave`, so even though the `<button>` retains focus after clicking, the tooltip won't stay visible once mouse user moves away.

the mouse hover styles have explicitly NOT been make to also apply to :focus, so as to minimise the visual impact for mouse users (though the browser's default outline will remain visible).

### DIFF
--- a/docs/assets/js/src/application.js
+++ b/docs/assets/js/src/application.js
@@ -55,9 +55,16 @@
 
     // Insert copy to clipboard button before .highlight
     $('.highlight').each(function () {
-      var btnHtml = '<div class="bd-clipboard"><span class="btn-clipboard" title="Copy to clipboard">Copy</span></div>'
+      var btnHtml = '<div class="bd-clipboard"><button class="btn-clipboard" title="Copy to clipboard">Copy</button></div>'
       $(this).before(btnHtml)
-      $('.btn-clipboard').tooltip()
+      $('.btn-clipboard')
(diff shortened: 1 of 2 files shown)


docs/assets/js/src/application.js
```diff
@@ -55,9 +55,16 @@
 
     // Insert copy to clipboard button before .highlight
     $('.highlight').each(function () {
-      var btnHtml = '<div class="bd-clipboard"><span class="btn-clipboard" title="Copy to clipboard">Copy</span></div>'
+      var btnHtml = '<div class="bd-clipboard"><button class="btn-clipboard" title="Copy to clipboard">Copy</button></div>'
       $(this).before(btnHtml)
-      $('.btn-clipboard').tooltip()
+      $('.btn-clipboard')
+        .tooltip()
+        .on('mouseleave', function (e) {
```
wolfy1339: This argument is not needed, as it is not used anywhere

patrickhlauke: ah, you're right. good spot. removed and force-pushed an update

mdo (approved):

PR #22453, opened 2017-04-16
Fix Card Deck and Card Group when .card-title contains a long word

pitpit:
Here's a caption and demo of behaviour before :

capture d ecran 2017-04-16 a 14 29 10

> You need to adjust width of the HTML preview frame to the size of a *small* device and see the bug

https://jsfiddle.net/pitpit/ed3fcfx3/

Here's a caption and a demo after the fix:

capture d ecran 2017-04-16 a 14 53 07

https://jsfiddle.net/pitpit/tcsy9fte/

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -20,6 +20,7 @@
 
 .card-title {
   margin-bottom: $card-spacer-y;
+  hyphens: auto;
 }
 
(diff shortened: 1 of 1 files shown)


mdo:
Alternate fix without hyphenation coming in #22454.

PR #22454, opened 2017-04-16
Prevent long words in `.card-title`s from breaking auto layout

mdo:
Alternate fix to card titles breaking the automatic card sizing in decks and groups without use of `hyphen`s property. Closes #22453.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -20,6 +20,7 @@
 
 .card-title {
   margin-bottom: $card-spacer-y;
+  word-break: break-all;
 }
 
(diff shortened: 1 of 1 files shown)


tillkruss:
This produces weird behaviour for me. Card title: `Fooooobar-Baaaaaaaaaaaz`.

Before:
```
Fooooobar-
Baaaaaaaaaaaz
```

After:
```
Fooooobar-Baaaa
aaaaaaaz
```

Is there a way for `word-break` to use regular hyphens and only break words if necessary?

Issue #21867, opened 2017-01-28
Missing button group role attribute


thbt:
In the [v3 documentation](https://getbootstrap.com/components/#btn-groups) you were using `role="group"` or `role="toolbar"` for buttons group, but you dropped these in the [v4 documentation](https://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons). This should be included in the v4 doc as it pushes good accessibility practices.

patrickhlauke:
the v4 equivalent of the v3 page you mention is https://v4-alpha.getbootstrap.com/components/button-group/ which does still include those `role` attributes. the example code from the button plugins lacks the `role` in both the v3 and v4 documentation currently.

what v4 does lack is the accessibility callout in the button group documentation, something that should probably be re-included.

it's debatable whether or not the `role="group"` makes any tangible difference in terms of accessibility in the case of those visually grouped radio buttons...it serves roughly the same purpose as a `<fieldset>`, and the `aria-label` acts a bit like a `<legend>` - though the behavior is not exactly the same in AT. i guess it could be included, but more for consistency than any hard benefit for AT users.

one concern i'd have is that `role="group"` and `role="toolbar"` have some quite specific connotations - see https://www.w3.org/TR/wai-aria/roles#group and https://www.w3.org/TR/wai-aria/roles#toolbar - and depending what authors end up doing (e.g. "perverting" the use of button group/button toolbars to do things other than grouped controls/toolbars) it may result in pages that are more broken than those without any roles defined...but i guess that's a problem that can never be fully addressed.

patrickhlauke:
the roles have now, from what i can see, been added to https://v4-alpha.getbootstrap.com/components/button-group/

leaving https://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons aside for now as i plan to redo this part separately, and current example works/makes sense even without `role="group"` (as it's more of a nice-to-have)

thbt:
Great ! Please notify me when you do the rework of the checkbox and radio buttons as I'm more interested in those (regarding [ng-bootstrap](https://ng-bootstrap.github.io/#/components/buttons))

PR #22455, opened 2017-04-16
Port button group accessibility callout from v3 to v4

patrickhlauke:
Added straight after the TOC because it applies to groups/toolbars overall, and it would arguably feel strange having it as a child section in one or the other.

Closes https://github.com/twbs/bootstrap/issues/21867


patrickhlauke:
Taking executive decision not to bother @mdo with reviewing this minute change. Again, think it's quite non-controversial (only thing I was concerned about was the positioning of the callout straight after the TOC, but I personally think that's the most logical place)

PR #22461, opened 2017-04-17
Add dropdown menu accessibility callout and fix <a> example

patrickhlauke:
As the question often comes up about why Bootstrap does not do "true"
ARIA menus (with their `role="menu"` etc), add an admittedly lengthy
note (tl;dr because BS is generic, and ARIA menus are specific and quite
limiting).

Additionally, fixes up the `<a>` example for dropdown trigger, with the
missing `role="button"` and neutering the `href` (which would be useless
anyway since BS overrides the link-like nature of the `<a>` so it could
never be triggered/followed anyway)


patrickhlauke:
originally had @mdo set to review, but i think this is fairly non-controversial (though yes, a tad long). going to merge...

PR #22464, opened 2017-04-18
Fix running test on iOS with Saucelabs

Johann-S:
Currently we thought our test were runs on iOS 9.3 but due to this bad configuration or maybe a change in Saucelabs it wasn't the case. Our tests were run on Windows 7 Safari 5 😨 

It was because of that my branch `v4-experimental-tooltip` cannot build correctly (#22444)

Thank you @FezVrasta for the idea to check on Angular project 👍 

/CC @mdo

### DIFF
--- a/grunt/sauce_browsers.yml
+++ b/grunt/sauce_browsers.yml
@@ -48,10 +48,10 @@
   # Win Opera 15+ not currently supported by Sauce Labs
 
   {
-    browserName: "Safari",
+    browserName: "iphone",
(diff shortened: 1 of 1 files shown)


Johann-S:
Ok now I have one build on Travis which is working but not the over with this error : 
`Your test errored. No active tunnel found for identifier 261735881`

Johann-S:
For me everything is fine here. So if they are someone who think it wasn't the good way please open a PR and /CC me 😉

PR #22473, opened 2017-04-19
Add myself to core team

Johann-S:
First of all I would like to thank every members of Bootstrap core team because
I learned so much thanks to you, thank you so much 👍 

I'm now a member of that team, it's a huge responsability but I'm very proud of that 😄 
And maybe one day I'll speak correctly in English 😆 

/CC @mdo


patrickhlauke:
:+1:

mdo:
Woohoo! Welcome aboard :D.

Issue #16028, opened 2015-03-09
Tooltips cannot be dismissed on iOS


upngo:
Tooltips can be shown on mobile by clicking, but you cant get rid of them.

They swallow up click events too so you can't add click handlers to the body to hide them.

Same as the problem noted in #3417 #6232 - both full of exasperated users..

If this isn't going to be fixed, at least a note could be put in the docs saying don't use tooltips if you want it to work on mobile without custom code disabling them for mobile devices...

cvrebert:
You're referring only to iOS and not Android, correct?

We could give [`body { cursor: pointer; }`](https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile) a try.

upngo:
Yes. That would make it more intuitive for iOS users but the problem is you
can't get them to disappear after the tap on iOS

cvrebert:
Uh, no, I mean that should literally theoretically fix the dismissibility problem. See linked article.

upngo:
Woops! True, that's convenient. Good find

fat:
@cvrebert is this a css issue?

cvrebert:
@fat It's an iOS Safari quirk/bug: https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile

fat:
@cvrebert i just meant more, that if we wanted to pursue a fix, woudl we just add that css snippet to the tooltip component… or did you expect the js to fix this somehow? (label wise)

cvrebert:
@fat I guess it would need to be in the JS. Something like:
```js
if ('ontouchstart' in document.documentElement) { // or whatever "is this a touch device?" test we want to use
  $('body').css('cursor', 'pointer');
}
```

patrickhlauke:
> `or whatever "is this a touch device?" test we want to use`

there is no bullet-proof way for this part though...and it would likely cause quirky effects on hybrid devices like touch-enabled laptops, surface, etc

cvrebert:
I guess we'd ought to toss in some iOS Safari detection/UA-sniffing for good measure.

kkirsche:
@cvrebert Why not just use the `Touch` object safari defines on mobile?

```javascript
// Safari Mobile
if(window.Touch) { /* JavaScript for your touch interface */ }
```

cvrebert:
@kkirsche I'm pretty sure that's not specific enough. There are browsers besides Safari that support Touch Events.

[...]

PR #22481, opened 2017-04-20
Add iOS hack to tooltips

patrickhlauke:
Same hack as in https://github.com/twbs/bootstrap/pull/22426 (modulo the selector, which is wrong in that PR and will be updated in a separate PR) to get tooltips to work correctly on iOS. Dynamically adds/removes empty (`noop`) `mouseover` event handlers to all children of `<body>` in order to coax iOS into proper event delegation/bubbling

Closes https://github.com/twbs/bootstrap/issues/16028

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -304,6 +304,15 @@ const Tooltip = (($) => {
 
         $(tip).addClass(ClassName.SHOW)
 
+        // if this is a touch-enabled device we add extra
+        // empty mouseover listeners to the body's immediate children;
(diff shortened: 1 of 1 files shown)


patrickhlauke:
urgh, and because my build system is broken, i completely missed that this was from an earlier version and needs tweaking. apologies for the second tweak today...once i can build and test properly again,  it won't happen :(

AdrianHL:
Could this fix/hack be migrated to v3 branch? Thanks in advance.

PR #22482, opened 2017-04-20
Tweak iOS hack for dropdown

patrickhlauke:
Tweak to https://github.com/twbs/bootstrap/pull/22426, where the wrong selector slipped through the net (selecting all of `<body>`s grand-children rather than children)

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -111,7 +111,7 @@ const Dropdown = (($) => {
       // https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
       if ('ontouchstart' in document.documentElement &&
          !$(parent).closest(Selector.NAVBAR_NAV).length) {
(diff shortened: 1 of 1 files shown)


js/src/dropdown.js
```diff
@@ -111,7 +111,7 @@ const Dropdown = (($) => {
       // https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
       if ('ontouchstart' in document.documentElement &&
          !$(parent).closest(Selector.NAVBAR_NAV).length) {
-        $('body').children().on('mouseover', '*', $.noop)
+        $('body').children().on('mouseover', null, $.noop)
```
alecpl: You can just omit the 2nd argument in this case. Right?

patrickhlauke: yes, but i thought i'd make it explicit (since that's what tripped up an earlier review)

PR #22484, opened 2017-04-20
`$input-height` and `$input-height-*` should include borders, no need to calculate select.form-control heights, and calculate form validation icon sizes

pat270:
Updated so $input-height, $input-height-lg, and $input-height-sm reflect the height of inputs with borders. Select elements should be the same size as other inputs when the value of $input-btn-border-width is greater than 1px.


mdo:
This will be tackled by #22395.

PR #22488, opened 2017-04-20
Arguably more professional pictures of Rick Ross

zachleat:
Not trying to body shame Rick Ross here but my assumption is that not all business clients using Bootstrap will celebrate half naked Rick Ross smoking blunts in the test files.

This PR keeps using images of Rick Ross but with additional clothing.

Also moves to host with imgur instead of tumblr, for no good reason.

Images maintain the same 1280x853 dimensions. If you’d like, I can make local copies of them in the repo.

### DIFF
--- a/js/tests/visual/carousel.html
+++ b/js/tests/visual/carousel.html
@@ -20,13 +20,13 @@
         </ol>
         <div class="carousel-inner">
           <div class="carousel-item active">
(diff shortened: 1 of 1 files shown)


zachleat:
Alright, I did the squashing.

patrickhlauke:
leaving this up to @mdo to decide ;)

mdo:
Lol, thanks for this <3.

PR #22495, opened 2017-04-22
First child of input group segmented buttons are off by $input-btn-border-width

pat270:
http://v4-alpha.getbootstrap.com/components/input-group/#segmented-buttons

Input group with input-group-btn as first child that have multiple buttons inside are off by $input-btn-border-width.

Edit: Made the selector more specific


oktalk:
Is this in reference to a known issue? Your code change doesn't do anything since the left margin is already `0`. Also, I'm not sure what you're trying to solve.

PR's have a better chance of getting through if you can create an issue requesting the PR first.

pat270:
Hey @oktalk,

I don't think this is a reference to a known issue. It's just something I stumbled across while using v4. It's hard to see on the docs page because the segmented button component isn't working. If you add `flex-direction: row;` to `.input-group` you can see it a bit better. Here is a gif of what I'm seeing.

bs4-segmented-buttons

It's not that noticeable because the borders are only 1px, but it becomes more apparent if you increase `$input-btn-boder-width` to something like 5px. This is what it looks like with this fix.

bs4-segmented-buttons-fixed

oktalk:
Thanks for the GIF. I see it now. That would be hard to explain without visuals.

mdo (approved):

mdo:
Looks good, and added another fix for a related issue.

PR #22496, opened 2017-04-22
New watch task

mdo:
Adds an `watch` task to compile CSS while the local server is running. Likely need to expand to JS, or at least rename to `watch-css` if it only affects the CSS. To get the right output, I removed the `docs/examples/` directory for the prefix task; it was rewriting all the example CSS and adding source maps for them (not needed).


mdo:
Hmm, seeing Sauce fail on the grunt task:

```
> bootstrap@4.0.0-alpha.6 js-test-cloud /home/travis/build/twbs/bootstrap
> ruby -run -ehttpd . -p3000 > /dev/null & grunt saucelabs-qunit
sh: 1: grunt: not found
```

Johann-S:
I added several commands

`watch-css` : which watch only css changes
`watch-js`: which watch only js changes
`watch`: which watch css and js changes

bardiharborow (approved):

PR #22477, opened 2017-04-20
More minor build tweaks.

bardiharborow:
Closes #22388. Closes #22476. Closes #22490.

PR #22500, opened 2017-04-22
Documentation cleanup (inc. use of <input ... readonly> for static form controls example)

patrickhlauke:
following up from all the html linting errors i was getting with the new build system (see https://github.com/twbs/bootstrap/pull/22477#issuecomment-296172499)

- fixes broken code indentation (and that was a joy to do...urgh)
- remove incorrect `<label>`, change the static text approach and explainer to encourage use of `readonly` inputs instead of meaningless paragraphs
- removes mention of `cursor:not-allowed` which was removed in https://github.com/twbs/bootstrap/pull/22403
- removes unnecessary vendor prefixes for `box-sizing` code sample (which also fixes the indentation linting error)

### DIFF
--- a/docs/.htmllintrc
+++ b/docs/.htmllintrc
@@ -9,6 +9,7 @@
   "html-valid-content-model": false,
   "id-class-ignore-regex": "(onclick|content|[a-z]+([A-Z][a-z])+)",
   "id-class-style": "dash",
+  "img-req-src": false,
   "img-req-alt": false,
   "indent-style": "spaces",
   "indent-width": 2,
--- a/docs/components/carousel.md
+++ b/docs/components/carousel.md
@@ -186,45 +186,45 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
 
 <table class="table table-bordered table-striped table-responsive">
   <thead>
-   <tr>
-     <th style="width: 100px;">Name</th>
-     <th style="width: 50px;">Type</th>
-     <th style="width: 50px;">Default</th>
-     <th>Description</th>
-   </tr>
+    <tr>
+      <th style="width: 100px;">Name</th>
+      <th style="width: 50px;">Type</th>
+      <th style="width: 50px;">Default</th>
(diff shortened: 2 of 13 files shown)


patrickhlauke:
@mdo i'd be particularly interested in your thoughts around the spacer.png use, and the changes to forms.md regarding static text

patrickhlauke:
potentially that "Static controls" section itself could be moved after the "Readonly inputs" section, as it follows logically from there.

Starsam80:
Adding `src` to the image elements completely breaks holder-js, adds multiple `src` attributes in the example, and shows usage of the assets folder (which I don't think is a good idea). Instead, why don't you turn off that linting error in `docs/.htmllintrc` by adding `"img-req-src": false`?

patrickhlauke:
@Starsam80 urgh, was so focused on getting the new build system to actually allow me to build, that i then didn't explicitly check the output. well, seems that holder is a bit...limited in that sense then.
tried an alternative approach of using `src="holder.js/..."` instead, but that just results in 404s in the console on those pages (plus the need to use separate live example code and then a `{% highlight html %}` block to make it look clean in the visible code output.
forcing the linter to ignore the issue is certainly a way around it, but opens up possibility of then missing actual valid problems where images have indeed been added without a proper `src` though...but i guess barring a change to holder's code itself to allow for and ignore any existing `src` attribute, or the above-mentioned 404s + extra duplication of live and highlighted code, that's the best/only option around the build problem. @mdo @bardiharborow thoughts?

patrickhlauke:
@mdo any feedback on the `readonly` change here (which is substantive)?

mdo (approved):

mdo:
Sorry to have been a blocker here! Formatting changes look good and I'm in favor of the move from non-input markup to readonly inputs for the "static controls." That said, maybe we open an issue to follow up there and specifically target `readonly` inputs, either with the attribute selector or a new class name?

PR #22510, opened 2017-04-24
Add responsive headings

houfio:
- Add `display-x-x` and `heading-x-x` classes for responsive headings
- Remove hx classes
- Update documentation accordingly


oktalk:
I like the concept of this, and the code LGTM.

mdo:
For responsive headings, we won't be going this route. Instead, we'll scale up each heading size without adding a new class for each breakpoint. Thanks though!

PR #22513, opened 2017-04-24
Nav Pills with open dropdown don't reflect the nav-pills active state…

### LINKED ISSUE #23482: Collapsed navbar navpill links are all in active state
Bootstrap 4.0.0-beta1

Navpill links in a navbar are all displayed in an active state when collapsed. Tested with Edge/Firefox/Chrome. Example:

https://www.bootply.com/VAyFaYVO64

(please click on the mobile icon to show the collapsed view)

pat270:
The active state should also be reflected when using collapse plugin inside nav-pills.

Latest v4-dev branch doesn't work like the current docs site http://v4-alpha.getbootstrap.com/components/navs/#pills-with-dropdowns. I think it was introduced at https://github.com/twbs/bootstrap/commit/91b62941afb7115807fc2925398ebccfc68f5377.

Fixes #23482.

### DIFF
--- a/scss/_nav.scss
+++ b/scss/_nav.scss
@@ -74,12 +74,12 @@
 .nav-pills {
   .nav-link {
     @include border-radius($nav-pills-border-radius);
+  }
 
-    &.active,
-    .show > & {
-      color: $nav-pills-link-active-color;
-      background-color: $nav-pills-link-active-bg;
-    }
+  .nav-link.active,
+  .show > .nav-link {
+    color: $nav-pills-link-active-color;
+    background-color: $nav-pills-link-active-bg;
   }
 }
 


scss/_nav.scss
```diff
@@ -74,12 +74,13 @@
...
   .nav-link {
     @include border-radius($nav-pills-border-radius);
+  }
 
-    &.active,
-    .show & {
-      color: $nav-pills-active-link-color;
-      background-color: $nav-pills-active-link-bg;
-    }
+  .nav-link[aria-expanded="true"],
```
mdo: We don't put aria roles in our selectors. This should likely be removed.

scss/_nav.scss
```diff
@@ -74,12 +74,13 @@
...
+  .nav-link.active,
+  .nav-item.show .nav-link {
```
mdo: Consider removing the `.nav-item` from `.nav-item.show .nav-link` so it's a slightly less specific selector?

pat270:
I removed the aria role selector and made `.nav-item.show .nav-link` less specific. I used the selector `.nav-item.show .nav-link` so the highlighted styles wouldn't bleed to nested `.nav-link`'s e.g. using the collapse plugin inside `.nav-pills`.

Johann-S:
This PR will close : #23482

mdo (approved):

PR #22538, opened 2017-04-27
Rename .form-control-static to .form-control-readonly-plain

patrickhlauke:
Following on from https://github.com/twbs/bootstrap/pull/22500 and its more semantically appropriate use of `<input ... readonly>` elements for "static controls", this renames the actual class. Admittedly the new name is a bit long, but very descriptive.

Didn't simply target `[readonly]`, because authors may want to retain the input-like look of `readonly` inputs, and only apply this new class when necessary/desired.


patrickhlauke:
ping @mdo

scss/_forms.scss
```diff
@@ -123,12 +123,12 @@ select.form-control {
...
-// Static form control text
+// Readonly controls as plain text
 //
-// Apply class to an element to make any string of text align with labels in a
-// horizontal form layout.
+// Apply class to a readonly input to make it appear like regular plain
+// text (without any border, background color, focus indicator)
 
-.form-control-static {
+.form-control-readonly-plain {
```
mdo: Would it be enough to call this `.form-control-readonly` instead? Do we envision having a readonly and readonly-plain in the future? Or would `.form-control-plaintext` be better?

PR #22541, opened 2017-04-28
Pagination that is wider than it's container overflows horizontally

pat270:
Pagination component with a lot of items causes horizontal overflow. Not sure if this is intended (like use a flex utility), I would expect it to break to a new line by default.


mdo:
Pagination shouldn't wrap by default, given it's current styling. Thanks though!

PR #22556, opened 2017-05-02
Input-group inside d-flex with a tall item causes input and input-group-addon to be too tall

pat270:
See https://jsbin.com/tanijet/edit?html,css,output


mdo:
Note for myself when revisiting this in Beta 2 release—will need to check out all the other input groups before trying to merge.

PR #22580, opened 2017-05-08
Busy sliding event

QwertyZW:
The branch adds a 'busy sliding' event that goes off every time a slide request is submitted while the carousel is already sliding.

The justification for something like this is to allow developers to display an indicator such as a 'spinner' to indicate that the carousel has indeed recieved the request from the UI controls/That the UI controls are being 'spammed'. An indicator such as this can be embedded as a part of the UI as well, but an event like this merely allows the developer to implement their own indicators, by utlizing both the current 'slid' event and the newly added event.

The image URLs in the visual test seemed to be broken, so I replaced them with placeholdits

Note: I wasn't sure about the mapping of next/prev to left/right respectively in the current state of the carousel module. I assumed a mapping of next/prev to right/left, which goes against what the code is already doing but which also made more sense for the time being.

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -124,9 +124,11 @@ const Carousel = (($) => {
     // public
 
     next() {
-      if (!this._isSliding) {
-        this._slide(Direction.NEXT)
+      if (this._isSliding) {
+        return
       }
+
+      this._slide(Direction.NEXT)
     }
 
     nextWhenVisible() {
@@ -137,9 +139,11 @@ const Carousel = (($) => {
     }
 
     prev() {
-      if (!this._isSliding) {
-        this._slide(Direction.PREV)
+      if (this._isSliding) {
+        return
       }
+
+      this._slide(Direction.PREV)
     }
 
     pause(event) {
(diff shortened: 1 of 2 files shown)


Johann-S:
Hi @QwertyZW thank you for you PR 👍 

I looked what you did but I don't see interest to add a busy slid event because we already have `slide.bs.carousel` which is fired immediately when the slide instance method is invoked. 
So you can bind this event and add your indicator.

The `slid.bs.carousel` is fired when the carousel has completed its slide transition.

QwertyZW:
Yes, you can use slid.bs.carousel and slide.bs.carousel to implement something such as a 'static' 'loading' indicator. However creating a dynamic indicator will come with an overhead of adding your own listeners to all the UI controls that can switch the carousel. Why would we do that if we already have listeners setup by the carousel?

We can also go down the route of firing a different kind of event that goes off every time someone uses a control that would cause the carousel to slide. But in order to detect 'spam' we would still have to listen to this new event in conjunction with slide.bs.carousel and slid.bs.carousel to keep track of whether or not the carousel is currently sliding in order to finally be able to conclude that the carousel is being 'spammed' or not.

XhmikosR:
@Johann-S: is this still relevant?

mdo:
Closing as stale given the conflicts and lack of progress here. Happy to revisit in the future though, so please open an issue and we can discuss. <3

PR #22581, opened 2017-05-08
No check on element is null causes error in tooltip.show.complete

NorbertKramer:
When, while destroying a tooltip, a mouseover (on the element with that tooltip) can raise an error.

This change checks for the existence of that.$element before doing anything,


Johann-S:

If you find same issue on V4 please open an issue

Issue #22600, opened 2017-05-11
Using collapse on display:flex elements ie: row


iatek:
I found this same problem in closed #20442, and I was wondering if it will be revisited for beta. Since row is `display:flex`, and `collapse.show` sets `display:block`, the displayed row is off.

```
<div class="container">
    <div class="row">
        <h4 class="col">
        <a href="#details" data-toggle="collapse">
            <span>Expand</span>
        </a>
      </h4>
    </div>
    <div id="details" class="row collapse" aria-expanded="false">
        <div class="col-6">
            <h3>Text</h3>
        </div>
        <div class="col-6">
            <h3>Text</h3>
        </div>
    </div>
    <div class="row" aria-expanded="true">
        <div class="col-6">
            <h3>Text</h3>
        </div>
        <div class="col-6">
            <h3>Text</h3>
        </div>
    </div>
</div>
```

http://codeply.com/go/UuAQvoRMiC

emma-burrows:
In case I'm not the only one who can't access that codeply snippet, I also included one in the question that probably prompted this issue: http://stackoverflow.com/questions/43909742/bootstrap-4-alpha-6-collapsed-two-divs-side-by-side-are-stacked-when-expanded

mdo:
The solution here isn't to swap the `display` value on `.collapse` classes. @Johann-S, I'm curious if the collapse plugin could read the initial display somehow? Otherwise there's no way around collapse being somewhat limited.

Johann-S:
I don't see any reliable method to check that, but we can check if the element to collapse have `.row` class and apply a different class with `display: flex;`.
Does it sounds a good solution for you @mdo ?

mdo:
That fixes it for row, but what if someone has overridden a one-off row to say `inline-flex` that also collapses? This fixes one edge case, but doesn't provide much help otherwise.

Johann-S:
I added a way to detect the use of flexbox in our Collapse plugins 🎆

DDDgfx:
I think I am having this problem. I add the collapse class to a .row, and the result is child columns stack as if block elements. Is there a fix?

PR #22605, opened 2017-05-12
Use flex instead of block for Collapse .show

Johann-S:
Use `flex` for Collapse `.show` class

Not sure about if it's the good solution or not so I hope feedbacks from @mdo 😄 

Close : #22600


mdo:
Unfortunately we can't—this would introduce more problems than it'd solve I'm afraid. It'd trigger different behaviors than `block`.

PR #22612, opened 2017-05-13
Fix border on input group elements and custom form inputs

### LINKED ISSUE #21098: Invisible border of input-group in inverse navbar

nwoltman:
c3e4cbd changed the border of input elements to be translucent. This makes input group elements and custom form inputs follow the same pattern.

Fixes #21098

bootstrap

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -140,6 +140,7 @@
   color: $custom-select-color;
   vertical-align: middle;
   background: $custom-select-bg $custom-select-indicator no-repeat right $custom-select-padding-x center;
+  background-clip: padding-box;
   background-size: $custom-select-bg-size;
   border: $custom-select-border-width solid $custom-select-border-color;
   @include border-radius($custom-select-border-radius);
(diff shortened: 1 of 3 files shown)


mdo:
This might be perfect! I'm trying to prioritize only the essentials for our first beta though, so going to wait just a bit on merging this <3.

mdo (approved):

arnauddrain:
Shouldn't we change the color of the border when the input is focused as well?
screen shot 2017-08-16 at 17 00 35

PR #22615, opened 2017-05-14
Change "Max CDN" to "MaxCDN"

AnandChowdhary:
MaxCDN is one word.


Johann-S:
Nice catch thank you 👍

PR #22640, opened 2017-05-17
Use popper to align dropdown menu instead of using css with important

Johann-S:
Remove the use of `important` in css because it's a bad practice and now we have Popper 🎆 
@FezVrasta any feedbacks on this PR ? And thank you for your help 👍


Johann-S:
@FezVrasta any feedbacks on this PR ? And thank you for your help 👍

FezVrasta:
Umh, does it actually work? I thought Popper.js would freak out if you change the `right: 0` property to  `auto`

Johann-S:
Yes it works very well

FezVrasta:
Umh, I'm not so confident, but if you tested it, great.

Johann-S:
As you can see : https://codepen.io/Johann-S/pen/mmGrJe

FezVrasta:
👍

PR #22641, opened 2017-05-17
fix tooltip documentation page missing classname prefix "bs-" to tooltip top

Lausselloic:
fix doc error


Johann-S:
Thank you @Lausselloic and nice catch 😉

Issue #22628, opened 2017-05-16
Dropdown and navbar


danijelGombac:
I try use the new dropdown with popper, but i can't use it in navbar. Popper always add position absolute to dropdown. How to be used dropdown and navbar in mobile?

Johann-S:
Bug reports must include a live demo of the problem. Per our contributing guidelines, please create a reduced test case via [JS Bin](https://jsbin.com/) or [CodePen](http://codepen.io/) and report back with your link, Bootstrap version, and specific browser and OS details.

danijelGombac:
I was tried to create a JS Bin test, but alerted me: Sorry this bin is too large for as to save.

danijelGombac:
The navbar css:

```
 .navbar-expand-md .navbar-nav .dropdown-menu {
    position: static;
    float: none;
}
```

But what can i see the position was overridden by popper and cannot be applied.

Johann-S:
I made a CodePen and I don't see any issues here : https://codepen.io/Johann-S/pen/RVBoGM
Maybe I misunderstood something 😣

danijelGombac:
Alpha 6 navbar for compared. Is different in mobile.

https://jsbin.com/cafeceyuwu/edit?html,css,js,output

Johann-S:
Yes I see the display is different 👍 
But why is it a problem for you ?

danijelGombac:
I don't use navbar like in documentation. In mobile i move it to the left and is opened like a drawer:

http://git.blivesta.com/drawer/

This dropdown menu is not very good for my design. The old version can expand and is better looking.

Johann-S:
Hmm ok I see thank you for the explanation 👍

PR #22644, opened 2017-05-17
fix(css): force .navbar .dropdown-menu positioning

zalog:
**Demo**
- Before: http://codepen.io/zalog/pen/eWLvRP
- After: http://codepen.io/zalog/pen/WjgGzM

**Problems:**
Please follow my example in _before_, the problems are in the same order.
- for navbars that never collapse `.navbar.navbar-expand`, `.dropdown-menu` has `position: static` and push parent `.navbar` container down
- if `.dropdown-menu` it's on the edge of browser, popper will move it and seems unnatural for `.navbar`
- if we use only `.navbar` (for navbar that always collapse), `.dropdown-menu` act in _popper style_

--

I tried to solve that by forcing `.dropdown-menu` to act like before popper, and in a mobile first abordation.

cc @Johann-S, #22628, #22630

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -75,6 +75,12 @@
     padding-right: 0;
     padding-left: 0;
   }
+
+  .dropdown-menu {
+    position: static !important;
+    float: none;
+    transform: unset !important;
+  }
 }
 
 
@@ -136,14 +142,6 @@
 
     &#{$infix} {
       @include media-breakpoint-down($breakpoint) {
-        .navbar-nav {
-          .dropdown-menu {
-            position: static !important;
-            float: none;
-            transform: unset !important;
-          }
-        }
-
         > .container,
         > .container-fluid {
           padding-right: 0;
@@ -160,7 +158,8 @@
           flex-direction: row;
 
           .dropdown-menu {
-            position: absolute;
+            position: absolute !important;
+            top: 100% !important;
           }
 
           .nav-link {


scss/_navbar.scss
```diff
@@ -160,7 +158,8 @@
           flex-direction: row;
 
           .dropdown-menu {
-            position: absolute;
+            position: absolute !important;
```
Johann-S: Not sure about that because Popper always use `absolute` position

Johann-S:
About that : 
> if .dropdown-menu it's on the edge of browser, popper will move it and seems unnatural for .navbar

If you user wants to disable that feature they can add `data-flip="false"` to disable that behavior

scss/_navbar.scss
zalog: Yep, I know.
But now, because of mobile first abordation, we need to overwrite [L80](https://github.com/zalog/bootstrap/blob/bfbb443b9e1bd305afdc17b8c557606e64639f03/scss/_navbar.scss#L80).

Why mobile first abordation:
- for navbars that never collapse, `.navbar.navbar-expand`. My first navbar in the _before_ example
- for navbars that always collapse, just `.navbar` class. My third navbar example in that _before_ like

zalog:
We can use `data-flip` but not for `.navbar`, I think. Here, `.dropdown-menu` must stay in place by default.
What do you think?

Preview:
image

FezVrasta:
I'm not sure what are you trying to fix, the first codepen looks better than the second one...?

zalog:
Before: http://codepen.io/zalog/pen/eWLvRP
- first navbar example is for `.navbar.navbar-expand`, just click ok "Dropdown link" button please.
- second example, just click on "Dropdown link"
- third navbar example is for `.navbar` without any complementar `.navbar-expand` or `.navbar-expand-*` classes. Click on the sandwitch button to collapse menu, and after that, click on the "Dropdown link".

FezVrasta:
I see the dropdowns broken only in the version when you add `static !important` and `unset !important` 😢

zalog:
Yes, it's because of this [commit](https://github.com/twbs/bootstrap/commit/ba2352c14ee1524a44e26f3b2393d4ed3195d99f).
The _before_ example is with the current v4-dev branch.

Johann-S:
With your PR this issue will stay fixed : https://github.com/twbs/bootstrap/issues/22628 ?
And can you update your branch please

zalog:
Yep, #22628 stays fixed.
Yes, I just rebased the branch.
Did you guys saw this bugs?

Johann-S (approved):

Johann-S:
I approve but it's a real problem to override Popper.js css with `!important`

zalog:
Yep I know... But this is the only way to override Popper.js and have `.dropdown-menu` controlled in the `.navbar` way.

Issue #22313, opened 2017-03-30
Button do not have cursor:pointer ? Why?


sahilweb6:
I am making an app and after checking the buttons I saw there is no cusror:pointer on the buttons . Why have you removed or you forget to add this? Is there any specific reason for this? Because if there will be no cursor the user will be confused about its functions. 

Note : **I can add custom css but I want to know the reason**

patrickhlauke:
the decision was made mainly in light of this https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

PR #22660, opened 2017-05-19
Add pointer cursor on buttons for v4

v-stein:
Added the pointer cursor on .btn class in v4 so that all buttons (a.btn, input.btn, button.btn) have the pointer.


hebbet:
see [#22313](https://github.com/twbs/bootstrap/issues/22313#issuecomment-290381725)

PR #22666, opened 2017-05-20
Make anchors in .thead-inverse elements white

### LINKED ISSUE #22665: .thead-inverse a should be white, not blue
I'm migrating to bootstrap 4, and have the following html structure:

```
<table class="table">
  <thead class="thead-inverse">
    <tr>
      <th>#</th>
      <th><a>Created Time</a></th>
    </tr>
  </thead>
</table>
```

However, the anchor is styled as blue, not white:
screenshot

hughbe:
Fixes #22665

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -109,6 +109,10 @@
     color: $table-inverse-color;
     background-color: $table-inverse-bg;
   }
+
+  th a {
+    color: $table-inverse-color;
+  }
 }
 
 .thead-default {


mdo:
I wanted to suggest we rewrite this to be a bit more friendly to other components and styles:

```
th > a {
  color: inherit;
}
```

So that it'd only affect immediate children anchors, and not screw up say dropdowns or custom markup. But, that's just as problematic because there _is_ custom markup. My gut says to let folks use a `.text-white` utility class instead.

Thoughts?

mdo:
Closing per above comment—let's have folks handle this with the `.text-white` utility.

hughbe:
Oh dear I missed your comment - sorry.

I understand your reasoning, but this now means we have to use text-white on each `th` element, right? Or maybe just on the `thead` element.

But I disagree, as this makes us inconsistent between things like navbar-inverse, which automatically makes links white. This inconsistency is unexpected - what do you think about that?

mdo:
The difference here is that the colors in the navbar are scoped to specific classes and aren't applied to the `.navbar` or a generic `.navbar a` selector. Changing the colors there is done by modifying specific elements with the right class.

We could add `.table-inverse th a { color: white }`, but what if your table headers had dropdowns in them—all of a sudden, those links would be white, too, unless you do what I mentioned in https://github.com/twbs/bootstrap/pull/22666#issuecomment-304969998.

hughbe:
Yeah you're right - this selector is too generic.

I think we should leave this as is and encourage users to style it themselves. I also realised that selector you suggested would break the coloring of buttons in headers so there's
 no real good solution.

Issue #22669, opened 2017-05-21
Automatic popover arrow placement


alecpl:
Using the latest build based on Popper.js I don't see how to automatically place the popover arrow. When the button is on an edge of the screen the arrow is still in the middle of the popover.
```
NOW:-----------------------------------------------------------------------|
                                                                  ---------|
                                                                  |  link ||
                                                                  ---------|
                                           ---------------^----------------|
                                           |           popover            ||

EXPECTED:------------------------------------------------------------------|
                                                                  ---------|
                                                                  |  link ||
                                                                  ---------|
                                           ---------------------------^----|
                                           |           popover            ||
```
I see in the Popper demos it can position the arrow properly, so this is Bootstrap specific issue. Because the arrow uses :after selector it might be tricky to fix it programmatically. I suppose arrow positioning should be delegated to Popper.

Johann-S:
Thank you @alecpl for your feedback 👍 
Is it possible to post a CodePen with steps to reproduce that ?

alecpl:
It's as simple as creating a popover for a button that is on the right side of the screen. Something like:
```
$(button).popover({content: 'some content longish...', placement: 'bottom', trigger: 'click'})
```
and make sure the popover does not wrap the content text so it's much wider than the button.

Johann-S:
So I made a Codepen to demonstrate what you reported @alecpl see : https://codepen.io/Johann-S/pen/dWwQyX

The first button is our current dist of Bootstrap
And the second button use `x-arrow` attribute provided by Popper (see https://popper.js.org/popper-documentation.html#modifiers..arrow) but I'm not sure it's what you want in term of result

alecpl:
@Johann-S Thanks for working on this. Popover2's popover position is out of screen, that's what I see. And arrow is in the middle, which is also wrong. The first button displays popover in a correct position only the arrow position is wrong. So, x-arrow here somehow broke the popover positioning nothing more.

Johann-S:
Maybe @FezVrasta can help us on that issue ?

alecpl:
@Johann-S aside of the popover position issue (popper.js bug?) I see it set `left:0px` on the arrow element, but Bootstrap overwrites this with `left:50%` on `.popover.bs-popover-bottom .arrow::after, .popover.bs-popover-bottom .arrow::before`. `left:0` will of course not look good with Bootstrap arrow, some more work will be needed.

FezVrasta:
[...]

Issue #22572, opened 2017-05-05
Tooltips arrow breakaway


huang-x-h:
Hi

Browser：**Chrome 56.0.2924.87 / IE 10.0.9200.16384 / Firefox 53.0**

View http://getbootstrap.com/javascript/#tooltips, the tooltips arrow breakaway

![f4 0 t k7 6g7waby 2 3mf](https://cloud.githubusercontent.com/assets/381860/25737013/03169f34-31a8-11e7-8d44-9acc7030a48a.png)

But Chrome 57 / IE 11 it's ok.

Johann-S:
Tested on Win10 Firefox 53 not able to reproduced your issue.
FYI we don't support browser zooming see : http://getbootstrap.com/getting-started/#support-browser-zooming

huang-x-h:
We do not zoom the browser, it's default 100%, and we test on Win7 and Win8.1

oktalk:
I can't duplicate this either. Tested in all mentioned browsers and OS's.

jipexu:
the given example is BS 3 ??

..."View http://getbootstrap.com/javascript/#tooltips, the tooltips arrow breakaway" ...

it's labelled V4 ?

Johann-S:
Bootstrap 3 is essentially in maintenance mode as we focus on working towards a stable v4. As such, we're only accepting changes to v3's code on a case-by-case basis, usually only for critical bug fixes or docs improvements.

huang-x-h:
Sorry for reply late

We test v4 version also have the same problem

https://v4-alpha.getbootstrap.com/components/tooltips/

```
OS: Windows8  6.2.9200 N/A Build 9200
Browser: Chrome 56.0.2924.87
```

jipexu:
seems to happens only with your config ?
OS: Windows8  6.2.9200 N/A Build 9200
Browser: Chrome 56.0.2924.87

i have not this on 
Chrome Version 58.0.3029.96 (64-bit)
Safari
FF
macOS sierra 10.12.4

huang-x-h:
@jipexu chrome 58 we test it's ok, but not on chrome 56, and also on `IE10 / FF54`

dillix:
I also have this issue with tooltips on Chrome 56 (mac) & ElectronJS (build on Chromium 56)

jipexu:
so seems chrome 56 "bug" on both mac/win ?

dillix:
[...]

PR #22681, opened 2017-05-23
Use Popper.js to manage arrow position

Johann-S:
Use Popper.js to manage arrow position instead of using CSS

Close : #22669, #22572

### DIFF
--- a/docs/components/popovers.md
+++ b/docs/components/popovers.md
@@ -56,32 +56,32 @@ $(function () {
 Four options are available: top, right, bottom, and left aligned.
 
 <div class="bd-example bd-example-popover-static">
-  <div class="popover bs-popover-top">
-    <div class="arrow"></div>
+  <div class="popover bs-popover-top bs-popover-top-docs">
+    <div class="arrow" x-arrow></div>
     <h3 class="popover-title">Popover top</h3>
     <div class="popover-content">
       <p>Sed posuere consectetur est at lobortis. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>
     </div>
   </div>
 
-  <div class="popover bs-popover-right">
-    <div class="arrow"></div>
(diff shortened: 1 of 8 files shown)


docs/components/popovers.md
```diff
@@ -56,32 +56,32 @@ $(function () {
...
     <h3 class="popover-title">Popover top</h3>
     <div class="popover-content">
       <p>Sed posuere consectetur est at lobortis. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>
     </div>
   </div>
 
-  <div class="popover bs-popover-right">
-    <div class="arrow"></div>
+  <div class="popover bs-popover-right bs-popover-right-docs">
+    <div class="arrow" x-arrow></div>
```
FezVrasta: Note that you may override the `x-arrow` selector and simply use `.arrow` as selector if you prefer.

Johann-S: Yes, but do you know if it will cause Bootstrap some issues to use `x-arrow` attribute ? If not I'll `x-arrow`.
But thank you for letting me know that 👍

pkozlowski-opensource: IMO this would be preferable since it would decouple Bootstrap from a library to manage positioning. This is especially important for libraries that build on top of Bootstrap's CSS and re-do JavaScript logic (ex.: https://ng-bootstrap.github.io/#/components/popover)

FezVrasta: There aren't problems using `x-arrow`, but as @pkozlowski-opensource said, you may want to reduce the points of junction between Bootstrap and Popper.js to avoid to coupling them too much.

Johann-S: Feel free to open a pull request @pkozlowski-opensource if you want to handle that 👍

pkozlowski-opensource: Will do. Thnx @Johann-S @FezVrasta

PR #22684, opened 2017-05-24
Share button hover and focus styles

### LINKED ISSUE #22657: Button text color problem on focus state
If I use btn classes on an anchor link element, the color of the element will not become white in focus state as it inherits the variable value of the $link-color. 
https://v4-alpha.getbootstrap.com/components/buttons/#button-tags

I suggest that we should complete the focus part with color property in @button-variant mixin.

schpet:
this fixes a bug where button variants did not share the styles between focus and hover. 

fixes #22657


mdo:
This has since been solved. See #21625.

PR #22686, opened 2017-05-24
form-control states apply to all sizes

silverbackdan:
Modify a mixin for form-control states and apply to all form-control sizes. Noticed the issue when I changed from form-control to form-control-lg and the placeholder colour didn't update. Then I noticed that the disabled state wasn't applied ... the focus style may not be applied...?

Comments still need tidying up if this looks like to correct fix.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -38,26 +38,8 @@
   }
 
   // Customize the `:focus` state to imitate native WebKit styles.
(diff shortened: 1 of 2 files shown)


mdo:
`.form-control-lg` is a modifier class, not a replacement for `.form-control`.

silverbackdan:
Of course, very sorry!

PR #22689, opened 2017-05-24
Sync code shown with code actually used

heikojansen:
Remove class and attributes from buttons which are not expected to trigger dropdowns.
Code used to render the examples is correct but code shown was not.


Johann-S:
If you remove `data-toggle="dropdown` Dropdown won't work because of that selector : https://github.com/twbs/bootstrap/blob/v4-dev/js/src/dropdown.js#L63

`.dropdown-toggle` is usefull because it add the arrow next to the button to show it's a Dropdown

and `aria` attribute is for accessibility so it's usefull too

heikojansen:
Sorry for not providing a more detailed description. Yes, the class and attributes are required, but not on two button elements within the _same_ split button instance (".btn-group"). Please review once more.

Johann-S:
Please provide a CodePen to demonstrate what you explained because if I remove what you suggested Dropdown don't works

acshef:
@heikojansen is saying that for a split button, the NON-DROPDOWN side doesn't need the extra attributes.

image

Johann-S:
Understood thank you @acshef for the explanation 👍

heikojansen:
Thanks @acshef for the screenshot. Would have done that next, too, but was on holiday yesterday.
I'll try to provide more context, next time around...

PR #22695, opened 2017-05-25
Update Bootstrap masthead - Vertically center "B"

sandrosc:
This change would vertically center the B and make it look a bit better IMO.
I have included screenshots, judge for yourself.

Before:
bootstrap_before

After:
bootstrap_after


mdo:
How's this look everywhere else? `8rem` on macOS looks just as wrong. I think instead we'll want to swap this for an SVG.

mdo:
Superseded by #22708.

sandrosc:
Very cool 👍

PR #22697, opened 2017-05-26
Update equal-width example numbering

Scoutski:
PR simply updates the numbering in the equal-width example in the grid docs.


mdo:
<3

Issue #22702, opened 2017-05-26
progress-bar: When adding labels to non-default height bars, text displays at the top of the bar


davethegr8:
Apologies if this is a duplicate or intentional, I searched but wasn't able to find any matches.

In the latest bootstrap 4 (alpha6), when adding a non-default height to a `.progress-bar`, labels render at the top of the bar. 

For example, taking the basic example from https://v4-alpha.getbootstrap.com/components/progress/#labels and adding a height, the label shows at the top of the bar (I'd guess it should be middle?)

```
<div class="progress">
  <div class="progress-bar" role="progressbar" 
       style="height: 2rem; width: 25%; align-items: center;" 
       aria-valuenow="25" aria-valuemin="0" aria-valuemax="100">25%</div>
</div>
```

![screen shot 2017-05-26 at 10 20 39](https://cloud.githubusercontent.com/assets/55392/26505276/fa66e8ca-41fc-11e7-8740-d41e43eacf57.png)

This could be addressed by adding `display: flex;` and `align-items: center;` to `.progress-bar` but that feels a bit weird to me.

While considering the impacts to other things in these changes, I also noticed that when using multiple bars, non-default heights behave more weirdly:

![screen shot 2017-05-26 at 10 21 26](https://cloud.githubusercontent.com/assets/55392/26505340/3d4c1afc-41fd-11e7-8c0d-5c00a63ae397.png)

Using the flex/center changes makes it a little better, but not by much.

This all seems like a lot of work & headaches, so I'd like to make a small suggestion: change up progress bars so that the height gets set on `.progress`, and `.progress-bar` just has `height: 100%`. With the flexbox changes above, this leads to looking like so:

![screen shot 2017-05-26 at 10 33 09](https://cloud.githubusercontent.com/assets/55392/26505720/b965b656-41fe-11e7-8e88-ae48559275f5.png)

I think this ends up being a bit more flexible and maintainable, but I didn't want to drop a PR over the wall or come in and say "you should do things this way" because I don't know any context about why things are the way they are - I just know what I see and my uninformed assumptions.

Hopefully this helps, and I'll make a PR to detail all the differences just in case.

PR #22703, opened 2017-05-26
Progress bars: center labels vertically and set height on parent .progress

davethegr8:
These changes address the suggestions outlined in #22702

The changes include:

- Setting labels to center vertically: There is no visual difference in the base case, but in cases where the height of the bar is not the default, behaves better.

Before:

screen shot 2017-05-26 at 10 58 37

After: 

screen shot 2017-05-26 at 10 59 22

- Moving where custom heights are set from `.progress-bar` to `.progress`. Again, there is no change in the base case, but on non-default heights it behaves better (more flexible/ maintainable). This is especially useful when using multiple bars.

Before (you have to set them all to be the same height):

```
<div class="progress">
  <div class="progress-bar" style="width: 15%; height: 3rem;">1</div>
  <div class="progress-bar bg-success" style="width: 30%; height: 3rem;">2</div>
  <div class="progress-bar bg-info" style="width: 20%; height: 3rem;">3</div>
</div>
```

After (only one change):

```
<div class="progress" style="height: 3rem;">
  <div class="progress-bar bg-warning" style="width: 15%;">1</div>
  <div class="progress-bar bg-success" style="width: 30%;">2</div>
  <div class="progress-bar bg-info" style="width: 20%;">3</div>
</div>
```

You can still change the heights to be different if you want, but it's less work in the base case.

```
<div class="progress" style="height: 3rem;">
  <div class="progress-bar bg-success" style="width: 30%;">2</div>
  <div class="progress-bar bg-info" style="width: 20%; height: 1rem;">3</div>
</div>
```

screen shot 2017-05-26 at 11 04 49


scss/_progress.scss
```diff
@@ -1,32 +1,37 @@
+// Progress animations
```
mdo: You can drop these newly added comments—they don't add any new context that the code itself doesn't already tell you :).

scss/_progress.scss
```diff
@@ -1,32 +1,37 @@
...
-  @include box-shadow($progress-box-shadow);
 }
 
 .progress-bar {
-  height: $progress-height;
-  line-height: $progress-height;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
```
mdo: This isn't necessary as far as I can tell. See https://jsbin.com/wahusavehu/edit?html,css,output.

scss/_progress.scss
```diff
@@ -1,32 +1,37 @@
...
 }
 
+// Basic progress bar
 .progress {
   display: flex;
+  align-items: center;
+  height: $progress-height;
   overflow: hidden; // force rounded corners by cropping it
   font-size: $progress-font-size;
   line-height: $progress-height;
```
mdo: Probably safe to remove this? Or we might need to change the value at least.

mdo (commented):
Thanks for tackling this! Just a couple questions here.

scss/_progress.scss
davethegr8: ah, it looks like align-items is indeed safe to remove, but the height is needed (empty bars will collapse to zero) with no explicit height.

scss/_progress.scss
davethegr8: Great, I'll remove height. :)

scss/_progress.scss
davethegr8: Oh whoops, sorry. I was editing the npm package for testing in my project, and those snuck into the dist scss files.

scss/_progress.scss
mdo: Height is good to stay here, but I meant we can probably remove the `line-height`.

davethegr8: oh, gotcha, removing. (align-items seems to _also_ be safe with having it on the `.progress-bar`)

scss/_progress.scss
```diff
@@ -5,21 +5,19 @@
...
 
 .progress-bar {
-  height: $progress-height;
-  line-height: $progress-height;
+  display: flex;
+  align-items: center;
+  justify-content: center;
   color: $progress-bar-color;
   background-color: $progress-bar-bg;
-  @include transition($progress-bar-transition);
```
tmorehouse: Any particular reason why the transition was removed from progress in this PR?

davethegr8: I don't believe so, probably an oversight on my part.

PR #22705, opened 2017-05-26
Minor: Documentation Copy and Paste error

danrye:
I know this isn't against master, but the error doesn't exist in master.


mdo:
This was just fixed in #22697.

Issue #22699, opened 2017-05-26
[v4] Right aligned dropdown in nav causes horizontal overflow


kedarv:
This issue occurs on all desktop platforms. When you have a right aligned dropdown in the navigation bar, it overflows off the page creating a horizontal scroll and misplaced dropdown.

Demo v4 alpha 6: http://jsbin.com/sozuva/1
Demo v3 (desired behavior): http://jsbin.com/jorobic/1

I'm not sure if this issue is related to #22636

Johann-S:
Currently with the last dist of our v4 code : https://codepen.io/Johann-S/full/Pmgado/

/CC @zalog

Issue #22636, opened 2017-05-17
New dropdown misplaced


danijelGombac:
When i updated to the new dropdown v-4-dev , dropdown-right is misplaced.

https://codepen.io/danijelGombac/pen/eWLmOK

FezVrasta:
`.dropdown-menu-right` defines `left: auto !important; right: 0`, this breaks the inline style applied by Popper.js.

cc @Johann-S

Johann-S:
I don't see any `.dropdown-menu-right` here @FezVrasta but yes I'll have to find a better way to do this instead of add `!important` any ideas ? 

but yes @danijelGombac the dropdown is misplaced

FezVrasta:
I can't always get that class applied, I'm not sure how it works. By the way I think that if you upgrade to 1.9.4 one of the bugs will go. But the problem with the `.dropdown-menu-right` must be solved Bootstrap side.

Johann-S:
`.dropdown-menu-right` is applied manually so if it's not present in the HTML this class isn't applied.

Yes I agreed about `.dropdown-menu-right` must be solved on Bootstrap side but I just asked if you have any ideas how to fix it 😄

danijelGombac:
Hy,

i try to update the pen with the new code and i don't see any difference. When i add min-width to dropdown is worse.

https://codepen.io/danijelGombac/pen/eWLmOK

 I'm doing something wrong?

Johann-S:
In the previous version of you CodePen you didn't use `.dropdown-menu-right` and `.dropdown-menu-large` if you remove those class they are no issue. But you find a new one, so can you open a new issue please

im-bilaljafri:
Hi, I've an issue with drop-down menu. It goes under the navigation bar and hide half.
Do you guys have any idea or solution?

julien-deramond:
@im-bilaljafri This issue is closed, please create [a new discussion](https://github.com/orgs/twbs/discussions/new/choose) if you need help from the Bootstrap community describing precisely your problem.

PR #22707, opened 2017-05-26
fix(css): .dropdown-menu-right positioning

zalog:
This is a fix for placing `.dropdown-menu-right` on the right position for `.navbar` and override Popper.js.

Problem: http://output.jsbin.com/sozuva/1
This PR: https://codepen.io/zalog/pen/BREOMv

cc #22636, #22640, #22699

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -162,6 +162,11 @@
             top: 100% !important;
           }
 
+          .dropdown-menu-right {
+            right: 0;
+            left: auto !important;
+          }
+
           .nav-link {
             padding-right: .5rem;
             padding-left: .5rem;


mdo:
I don't understand what this fixes exactly. The problem link here has no `.dropdown-menu-right` on the `.dropdown-menu`, and you shouldn't need any additional styles to override that since our navbar uses static positioning for dropdowns in narrower viewports.

Johann-S:
The jsbin use Tether and is outdated.
With our last dist file this is what we get : https://codepen.io/Johann-S/full/Pmgado/

zalog:
Click on "Dropdown link", the bug is still there.
Please take a look on my pens:
before: https://codepen.io/zalog/pen/pPmzaE
after: https://codepen.io/zalog/pen/BREOMv

css is generated local: `npm run css-compile` & `npm run css-prefix`
js files loaded from cdnjs & rawgit

Johann-S:
Yes it fixed the issue but on the last `navbar` for me static position should be applied only for narrower viewports.

zalog:
That's a navbar that always collapse, a mobile style. I think the default behaviour for `.dropdown-menu` should open static on all the viewports sizes.
I belive it's much [cleaner this way](https://github.com/twbs/bootstrap/pull/22644#discussion_r117102469).

Or perhaps we find a way to get popper.js manage this, but I think we don't want all this mess on js.

I don't like override popper.js or use `!important` eighter... 

@mdo, have some thoughts?

mdo:
Ugh, yeah, we should work with the Popper team to get some better support for controlling the styles here. Having all these overrides kinda sucks. Ideally, instead of `position: absolute; transform: translate3d(0px, 38px, 0px); top: 0px; left: 0px; will-change: transform;` on every default dropdown, we apply as little as possible. There's no need for any of these given our CSS.

I'm unsure how best to rectify this in short order though.

Johann-S:
Maybe @FezVrasta have some ideas how we can handle that in a short term

FezVrasta:
Hey, Popper.js "team" here (😂).

If you want to make Popper.js apply only some specific styles you can inject a modifier before the applyStyle one and strip any unwanted properties from the styles object.

But in my opinion you should override the applyStyles modifier all together and create one that reads the class applied to the popper element and applies the needed styles accordingly

Johann-S:
I will try to be a bit more precise : 
- For me here they are no problem of positionning by Popper.js
- The problem is present on the first two navbars because of those lines : [_navbar.scss#L79-L82](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L79-L82) and [_navbar.scss#L160-L162](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L160-L162). Without those lines they are no problem of positionning
- What I currently don't understand is on those lines  [_navbar.scss#L79-L82](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L79-L82) we add a `static` position without taking into account : it's for all the different viewports. 
When I track the history of this file (`_navbar.scss`) we can see (https://github.com/twbs/bootstrap/commit/ba2352c14ee1524a44e26f3b2393d4ed3195d99f#diff-df1f74faf207a474f1d78c86393221c8) they were only on css rule where position were `static` and it was in a certain breakpoint and not for all the viewports/breakpoints 🤔 

Maybe someone can help me to understand why now it's needed to have a `static` position for all the viewports/breakpoint and before it wasn't the case and they wasn't any issues about that.

Johann-S:
I made a commit (see : https://github.com/twbs/bootstrap/commit/ca60fe4f23581d62b69b2d040bbd6aa5afd7ade6) which reset Popper.js styles for Dropdowns in Navbar.
I don't know if it's the best solution but it fix our issue.

I post that here because I hope some feedbacks or ideas

zalog:
We have now 2 ways of fixing this:
- this PR #22707 - the css way with `!important`
- Johann's [branch](https://github.com/twbs/bootstrap/compare/v4-dropdown-navbar) with some optimisations #22776 

In my opinion we shoud step the Johann way :)

Johann-S:
Superseded by #22782

PR #22709, opened 2017-05-27
Add `docs-clean` task to watch scripts

mdo:
Without this, the CSS and JS files don't get copied over to the `docs/dist/` folder during development. I imagine we can simplify this and not run it twice, and instead run it once? Unsure.

/cc @johann-s


Johann-S (approved):
LGTM 👍 

I don't see any way to do that in an easy way.
In an other way we can create a `watch.js` node script which use nodemon and can order tasks

PR #22716, opened 2017-05-29
Docs redesign

mdo:
Been derping around on this one for awhile now. Finally feeling good about it to open the PR.

- [x] Figure out a way to do Docsearch locally?
- [x] Move the docs to a subfolder in preparation of the beta?
- [ ] Redesign homepage
- [x] Copy masthead button styles to navbar download button
- [x] `target="_blank"` on social links in navbar
- [ ] Update examples page
- [ ] Redesign footer

screen shot 2017-05-27 at 9 40 35 pm

PR #22717, opened 2017-05-29
Rearrange all the docs to allow for a docs/major.minor/ setup

mdo:
This PR rearranges almost all of the `docs/` folder to better allow us to group actual documentation at `getbootstrap.com/docs/4.0/`. It builds on #22716 and would merge into that branch if accepted. The best way to take in the changes is to [browse the branch](https://github.com/twbs/bootstrap/tree/docs-rearrange).

Given the size of the diff, here's what's changed.

- The Jekyll `source` is no longer `docs`—it's the entire project.

- All the existing docs pages, minus the home page and about section are in `docs/4.0/`.

- The `docs/4.0/` folder will be renamed (via `git mv`) with each minor/major release. Deploying these changes to our production site is a manual copy-pasta dance, so we'll never lose old versions of the core docs.

- We no longer have two `dist` directories. Previously, we had one in the root and one more duplicated via npm script in `docs/` because Jekyll couldn't read from outside the `source` directory. This is the bulk of those 15,000 lines deleted.

- Moves some root dot files around. The `.github` is new and contains the contributing guidelines and issue template. The `build` directory contains the `.htmllintrc` and `.htmlhintrc` files.

- There are a ton more files and folders in our root directory. Not great, but not the end of the world to shave some code from the project.

I have no idea if this is the right approach, but it works for now. If this ends up shipping, we'll need to tackle a few more todos:

- [x] Update all URLs within docs pages
- ~~Update links to contributing docs and issue template.~~ This will be done at launch.
- [x] Look into consolidating more root directory files elsewhere (travis, hound, bable)
- [x] Verify we need all these package manager files (sache.json, package.js, composer.json, etc)
- [x] Consider nuking root changelog.md since we don't use it.
- [x] Figure out if htmllint and htmlhint are both needed (real talk this messed with me a ton)

Gimme your thoughts—is there a better way?


tomscholz:
Over Dogfalo/materialize we are (still) using jade and also have the docs in the root folder. In my opinion this is really messy. I always loved bootstrap for it's clean and simple folder structure. Put docs inside `docs/` is just my favorite way to handle docs.

> There are a ton more files and folders in our root directory. Not great, but not the end of the world to shave some code from the project.

Putting these files in the root will IMO confuse the users. Not everyone even knows what jekyll is and what it does.

I currently don't have a better solution but I still wanted to give some feedback ;) 

Best wishes from Hamburg

mdo:
@tomscholz Yeah, I've had the same opinion for awhile now. Here's what's run through my head while working on this the last few days (and years honestly):

- Separate the docs from the CSS/JS. Not feasible; it'd be a pain to develop any of the source code.

- Separate the docs _site_ from the main repo. Also not the best given we'd have to copy-pasta (or abstract) some elements of the site into the docs, and back and forth.

- Keep everything in `docs`, and deal with a weird nested docs directory and duplicate `dist` directory within that. Adds 15,000 lines of code (whoa now) and an extra script (eh, not a big deal), but woof that nested directory would be confusing as hell.

- Keep it altogether and deal with a funky directory structure. It's not as clean, but it's _straightforward_, and that's what I want as much as possible throughout Bootstrap.

I'm sure there are other options too, but I think this one is the best right now.

tomscholz:
Then the _funky directory structure_ it should be! ;)

mdo:
This has no merged into #22716! Consolidating work over there to finish that PR out next.

PR #22722, opened 2017-05-31
V4 Rename `$card-link-hover-color` to `$card-inverse-link-hover-color`

pat270:
I think `$card-inverse-link-hover-color` is a better description of what this variable does.


mdo:
Thanks!

PR #22731, opened 2017-06-01
Fix search redirect

maxiloc:
Small fix to fix the search redirect, without this even in production the search will redirect to localhost.


mdo:
<3

mdo:
@maxiloc You're a champ! Thank you so much for contributing, iterating, and giving feedback.

PR #22733, opened 2017-06-01
Docs - avoid Tooltips/Popovers overflow on documentations's navbar

Johann-S:
This PR avoid this effect on our new awesome documentation 

popovers bootstrap - mozilla firefox

By the way, thank you @mdo for your awesome work on our new documentation ❤️ 

/CC @mdo

### DIFF
--- a/assets/scss/_nav.scss
+++ b/assets/scss/_nav.scss
@@ -33,7 +33,7 @@
     position: -webkit-sticky;
     position: sticky;
     top: 0;
-    z-index: 1030; // over everything in bootstrap
+    z-index: 1071; // over everything in bootstrap
   }
 
   .navbar-nav {


mdo:
Is this a problem for all popovers and tooltips, for all navbars? Do we need to revisit the z-index of a sticky navbar, or is my docs navbar (which uses a custom sticky style) missing the z-index?

Johann-S:
This problem is just fort Tooltip/Popover examples in our documentation.
I should change my fix to target only examples Tooltip/Popovers.
It's due to a low z-index of the sticky navbar in our documentation

mdo:
> I should change my fix to target only examples Tooltip/Popovers.
> It's due to a low z-index of the sticky navbar in our documentation

Yup, sounds right to me! ;)

Johann-S:
finally the best solution was to change navbar's `z-index` because dynamic Tooltips/Popovers example are added to the end of the `<body>` 
I chose `1071` because `1070` it's the Tooltips `z-index`

PR #22739, opened 2017-06-01
Text Truncate should truncate text on non block level elements

pat270:
Demo: https://jsfiddle.net/ybzL5vu9/1/

`.text-truncate` doesn't work on non block level elements, used inline-block and width: 100% just incase someone wanted to truncate two sections of text on the same line.

Edit: I just saw the comment at the top of the mixin file. I think this change would be easier for end users to use. Changed to `display: block` because of vertical spacing issues with `inline-block`. I think block would be the more common use case.

### DIFF
--- a/scss/mixins/_text-truncate.scss
+++ b/scss/mixins/_text-truncate.scss
@@ -2,6 +2,7 @@
 // Requires inline-block or block for proper styling
 
 @mixin text-truncate() {
+  display: block;
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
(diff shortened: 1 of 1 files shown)


mdo:
I pushed a change to this to assume inline-block and allow for a parameter, but I'm still not sure on this one. I feel like maybe we should encourage folks to actually set a `.d-*` class instead? With your original proposal, if you're using the `.text-truncate` class on a `<span>` and wanted it `inline-block`, you'd have to use a display utility. If you did what I changed here and assumed `inline-block`, you'd have to do `d-block` on `<div>`s and the like.

All that to say, I'm going to close as a won't fix and modify the docs to make mention of supported `display` values with the text truncation.

mdo:
Welp looks like we already have that in our docs. No need for a follow up.

PR #22752, opened 2017-06-05
Fix for Issue #22225: Replaced hardcoded box-shadow values with $btn-focus-box-shadow scss …

geocode64:
…variable. Will work with $enable-shadows set or not, as per behaviour of .btn base style in scss/_buttons.scss file.


mdo:
Closing per https://github.com/twbs/bootstrap/issues/22225#issuecomment-306106453

PR #22758, opened 2017-06-06
Collapse with flex elements

Johann-S:
Detect if the element use flexbox, if it's the case add `.flexshow`

- [x] Add Visual test
- [x] Detect all posible flex values (inline-flex, flex, etc...)

We detect flex by creating hidden div(s) and we add each class to this hidden div to detect if this div is displayed in flex mode or not. 

We do that only when we call the constructor of Collapse plugins that's why I added an `update` method if the collapsible element change, to detect once again the appropriate show class

Close : #22600

/CC : @mdo

### DIFF
--- a/docs/4.0/components/collapse.md
+++ b/docs/4.0/components/collapse.md
@@ -226,6 +226,10 @@ Shows a collapsible element. **Returns to the caller before the collapsible elem
 
 Hides a collapsible element. **Returns to the caller before the collapsible element has actually been hidden** (i.e. before the `hidden.bs.collapse` event occurs).
 
+#### `.collapse('update')`
+
+Update a collapsible element if it changed during its lifetime.
+
 ### `.collapse('dispose')`
 
 Destroys an element's collapse.
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -46,6 +46,7 @@ const Collapse = (() => {
 
   const ClassName = {
     SHOW       : 'show',
+    FLEXSHOW   : 'flexshow',
     COLLAPSE   : 'collapse',
     COLLAPSING : 'collapsing',
     COLLAPSED  : 'collapsed'
@@ -113,7 +114,7 @@ const Collapse = (() => {
     // public
 
     toggle() {
-      if ($(this._element).hasClass(ClassName.SHOW)) {
+      if ($(this._element).hasClass(ClassName.SHOW) || $(this._element).hasClass(ClassName.FLEXSHOW)) {
         this.hide()
       } else {
         this.show()
@@ -121,8 +122,9 @@ const Collapse = (() => {
     }
 
     show() {
+      const showClass = this._getShowClass()
       if (this._isTransitioning ||
-        $(this._element).hasClass(ClassName.SHOW)) {
+        $(this._element).hasClass(showClass)) {
         return
       }
 
@@ -176,7 +178,7 @@ const Collapse = (() => {
         $(this._element)
           .removeClass(ClassName.COLLAPSING)
           .addClass(ClassName.COLLAPSE)
-          .addClass(ClassName.SHOW)
+          .addClass(showClass)
 
         this._element.style[dimension] = ''
 
@@ -201,8 +203,9 @@ const Collapse = (() => {
     }
 
     hide() {
+      const showClass = this._getShowClass()
       if (this._isTransitioning ||
-        !$(this._element).hasClass(ClassName.SHOW)) {
+        !$(this._element).hasClass(showClass)) {
         return
       }
 
@@ -221,7 +224,7 @@ const Collapse = (() => {
       $(this._element)
         .addClass(ClassName.COLLAPSING)
         .removeClass(ClassName.COLLAPSE)
-        .removeClass(ClassName.SHOW)
+        .removeClass(showClass)
 
       if (this._triggerArray.length) {
         for (let i = 0; i < this._triggerArray.length; i++) {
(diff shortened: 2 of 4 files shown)


js/src/collapse.js
```diff
@@ -298,6 +309,31 @@ const Collapse = (($) => {
       }
     }
 
+    _getShowClass() {
+      const tabClass = this._element.classList
+
+      // Detect flex for inline style
+      let useFlex = $(this._element).css('display').indexOf('flex') !== -1
+
+      // Create a wrapper to hide our flex detection
```
glebm: Can you not use `getComputedStyle(this._element).display` instead?

Johann-S: I can use that for the line 316 yes, but not for the detection below because getComputedStyle return the current applied style and that's not what I want here. 
I want to know if they are a class which contain a flex display

mdo:
Is anything here considered to be a breaking change @Johann-S? Might be beneficial to wait for Beta 2 to address this.

Johann-S:
That's right this can wait for beta 2

js/src/collapse.js
```diff
@@ -314,6 +325,31 @@ const Collapse = (($) => {
       }
     }
 
+    _getShowClass() {
+      const tabClass = this._element.classList
+
+      // Detect flex for inline style
+      let useFlex = $(this._element).css('display').indexOf('flex') !== -1
```
XhmikosR: How about just using https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes?

js/tests/visual/collapse.html
```diff
@@ -55,9 +55,25 @@ <h5 class="mb-0">
...
+                Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid. Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
     </div>
 
-    <script src="../../../docs/assets/js/vendor/jquery-slim.min.js"></script>
+    <script src="../../../assets/js/vendor/jquery-slim.min.js"></script>
```
XhmikosR: Hmm, how did it work without this?

docs/4.0/components/collapse.md
```diff
@@ -226,6 +226,10 @@ Shows a collapsible element. **Returns to the caller before the collapsible elem
 
 Hides a collapsible element. **Returns to the caller before the collapsible element has actually been hidden** (i.e. before the `hidden.bs.collapse` event occurs).
 
+#### `.collapse('update')`
+
+Update a collapsible element if this element changed during his lifetime
```
XhmikosR: `its lifetime`

XhmikosR: And maybe add a `.` at the end.

js/tests/visual/collapse.html
Johann-S: It didn't work before that change and yes it's an unrelated change

js/src/collapse.js
Johann-S: I switched to `includes` but they are no support on IE 10 so back to `indexOf`

XhmikosR: But it's transpiled isn't it?

docs/4.0/components/collapse.md
```diff
@@ -226,6 +226,10 @@ Shows a collapsible element. **Returns to the caller before the collapsible elem
 
 Hides a collapsible element. **Returns to the caller before the collapsible element has actually been hidden** (i.e. before the `hidden.bs.collapse` event occurs).
 
+#### `.collapse('update')`
+
+Update a collapsible element if this element changed during its lifetime.
```
XhmikosR: How about: "Update a collapsible element if it changed during its lifetime."

js/src/collapse.js
Johann-S: I didn't checked but unit tests failed

docs/4.0/components/collapse.md
Johann-S: Yep that's better thanks 👍 I'll update that soon

js/src/collapse.js
XhmikosR: But since we use babel it should be taken care of. Please try one more time.

XhmikosR: I looked into it a bit more and it seems Babel doesn't do this. Just verify one more time and if it doesn't just use `indexOf`. It's not worth using a polyfill for this.

XhmikosR: Just keep in mind we can do this if we just load the babel polyfill and use ES6 methods. Not sure what the exact size diff is, just thinking out loud for another PR.

Johann-S: Do you know which polyfill to use ? I can make a test 👍

XhmikosR: https://babeljs.io/docs/usage/polyfill/

Johann-S:
For me everything is fine @mdo so this one can be merged 👍

js/src/collapse.js
Johann-S: I checked `babel-polyfill` and currently when we import babel-polyfill we will import all the available polyfill not the `includes` one so for now we should merge this PR and open an issue on how to integrate babel-polyfill with only specific es6 polyfill

XhmikosR: No worries, it's not so important currently as long as we don't use features that aren't available for the browsers we support.

vanillajonathan:
Keep in mind that the other `.flex-*` classes are separated with a hyphen.
Example `.flow-row`, `.flex-column`, `.flex-wrap`, etc.'.

Perhaps `.flex-show` should be considered.

scss/_transitions.scss
```diff
@@ -14,16 +14,21 @@
...
   }
+  &.flexshow {
+    display: flex;
+  }
 }
 
 tr {
-  &.collapse.show {
+  &.collapse.show,
+  &.collapse.flexshow {
```
vanillajonathan: Should `.flexshow` be named `.flex-show` to be consistent with the hyphen separation used by the other `.flex-*` classes?

Johann-S:
Not sure to keep that PR, because I have to create an hidden div to detect if folks use `flex` attributes or not, a better solution would be CSS one

PR #22773, opened 2017-06-07
Updated flex-basis property to fix issue #18569 in .scss files

obinnangini:
Tied to my comment on issue [18569](https://github.com/twbs/bootstrap/issues/18569)

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -197,7 +197,7 @@
 
     .card {
       display: flex;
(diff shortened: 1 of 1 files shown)


wolfy1339:
The commit message and title should be fixed. It should be `issue #18569 `and not `issue# 18569`

mdo:
Incorporating your commit into #22862. Thanks!

PR #22776, opened 2017-06-07
Some .navbar .dropdown-menu optimisation

zalog:
We do not need [those css rules](https://github.com/twbs/bootstrap/blob/ca60fe4f23581d62b69b2d040bbd6aa5afd7ade6/scss/_navbar.scss#L144) because are inherited from [here](https://github.com/twbs/bootstrap/blob/ca60fe4f23581d62b69b2d040bbd6aa5afd7ade6/scss/_navbar.scss#L79).

Also, I think we do not need `.dropdown-menu-left`, because this is inherited from [`.dropdown-menu`](https://github.com/twbs/bootstrap/blob/ca60fe4f23581d62b69b2d040bbd6aa5afd7ade6/scss/_dropdown.scss#L38).

I'm not sure about the existence of member `_navbarPositioning()`, those rules are already aplyed from css.

### Test plan
- `npm run js-compile` & `js-minify`
- `npm run css-compile` & `css-prefix` & `css-minify`
- https://codepen.io/zalog/pen/RgWKKo

cc: #22707

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -251,21 +251,6 @@ const Dropdown = (($) => {
       return $(this._element).closest('.navbar').length > 0
     }
 
-    _navbarPositioning() {
-      const $parentNavbar = $(this._element).closest('.navbar')
-      if ($(this._menu).hasClass(ClassName.MENURIGHT)) {
-        if (!$parentNavbar.hasClass('navbar-expand')) {
-          return {
-            position: 'static',
-            transform: '',
-            float: 'none'
-          }
-        }
-      }
-
-      return {}
-    }
-
     _getPopperConfig() {
       const popperConfig = {
         placement : this._getPlacement(),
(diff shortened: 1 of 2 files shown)


Johann-S:
Thank you for the clean of my code 😆 I'll check that soon

Johann-S:
I check what you did but I have this behavior and I don't think it's the appropriate behavior : 
image

zalog:
That is a simple default .dropdown-menu, it has no .dropdown-menu-right :)

Johann-S:
Yes I understand that, but IMO this menu shouldn't add an horizontal scroll

zalog:
For horizontal scroll not to show, we need to add `overflow: hidden` to `.navbar`. We can not do that because that css rule will hide `.dropdown-menu` too.
By default, `.dropdown-menu` it's positioned on the left side of x axis. If the user want to align this menu, he must use `.dropdown-menu-right` class.
Or... maybe if we let popper.js to manage only x axis align?

I made some tests with v4.0.0-alpha.6 and the horizontal scroll is there :)

Johann-S:
Thank you @zalog to improve my branch 👍

PR #22782, opened 2017-06-08
Use css to display Dropdown in Navbar

Johann-S:
In this PR, we remove CSS aplied by Popper.js (only for Dropdown in navbar) to let our CSS do the job.

We detect a weird behavior, If we have a navbar in 100% width container an horizontal scroll appear.
See this CodePen : https://codepen.io/zalog/pen/Rgrbod (first Navbar)

I'm not sure if it's really a bad behavior, maybe @mdo can help us in this use case ? 

Thank you to @zalog to review/improve my first commit 👍 

Close : #22699

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -186,6 +186,7 @@ const Dropdown = (($) => {
     }
 
     update() {
+      this._inNavbar = this._detectNavbar()
       if (this._popper !== null) {
         this._popper.scheduleUpdate()
       }
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -251,21 +251,6 @@ const Dropdown = (($) => {
       return $(this._element).closest('.navbar').length > 0
     }
 
-    _navbarPositioning() {
-      const $parentNavbar = $(this._element).closest('.navbar')
-      if ($(this._menu).hasClass(ClassName.MENURIGHT)) {
-        if (!$parentNavbar.hasClass('navbar-expand')) {
-          return {
-            position: 'static',
-            transform: '',
-            float: 'none'
-          }
-        }
-      }
-
-      return {}
-    }
-
     _getPopperConfig() {
       const popperConfig = {
         placement : this._getPlacement(),
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -96,10 +96,11 @@ const Dropdown = (($) => {
   class Dropdown {
 
     constructor(element, config) {
-      this._element = element
-      this._popper  = null
-      this._config = this._getConfig(config)
-      this._menu = this._getMenuElement()
+      this._element  = element
+      this._popper   = null
+      this._config   = this._getConfig(config)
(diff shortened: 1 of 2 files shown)


scss/_navbar.scss
```diff
@@ -77,9 +77,8 @@
   }
 
   .dropdown-menu {
-    position: static !important;
+    position: static;
```
mdo: What if we scoped this to a `max-width` media query instead? Would that mean we don't need to reset the styles later for `.dropdown-menu` and `.dropdown-menu-right`?

Johann-S: `.dropdown-menu-right` only exist here, because we handle that on our JS for the other Dropdown which aren't in a navbar.
But yes I think it's possible to add a media query just for the Dropdown position

Johann-S:
Currently we have an issue on navbar (https://github.com/twbs/bootstrap/issues/22831) to fix before the merge of this PR

scss/_navbar.scss
zalog: I'm on your branch `.v4-dropdown-navbar` and made two pens:
- one with the last commit: [e836f5](https://codepen.io/zalog/pen/OgbVzE)
- the other with one commit before the last commit: [7d3368](https://codepen.io/zalog/pen/bRBdLm)

The problem is that with our last commit:
```
.navbar-expand {
    $next: breakpoint-next($breakpoint, $grid-breakpoints);
    $infix: breakpoint-infix($next, $grid-breakpoints);

    &#{$infix} {
        .dropdown-menu {
          position: static;
          float: none;
        }
...
```
the rules `position: static; float: none;` won't apply in case of:
- navbars that always collapse `.navbar`
- navbars that never collapse, `.navbar-expand`

This two use cases (navbar that always/never collapse) are already documented [here](https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/components/navbar.md#responsive-behaviors).

I mentioned this, first time, [here](https://github.com/twbs/bootstrap/pull/22644#discussion_r117097343). I hope now It's much more clear :)

js/src/dropdown.js
```diff
@@ -250,13 +242,40 @@ const Dropdown = (($) => {
...
+          },
+          flip : {
+            enabled : this._config.flip
+          }
         }
       }
-      return placement
+
+      if (this._inNavbar) {
+        popperConfig.modifiers.AfterApplyStyle = {
```
FezVrasta: Any reason to make Popper apply styles and then remove them?

You could set the `enabled` property of the `applyStyle` modifier to `!this._inNavbar`

Johann-S: Oh thank you @FezVrasta ! Nice review !

Johann-S: Finally I tried what you suggested but it add inline style see https://github.com/FezVrasta/popper.js/issues/309

FezVrasta: Gonna fix this tomorrow morning!

Issue #22745, opened 2017-06-03
Inconsistent navbar classes naming


marko-avlijas:
Opposite of light is dark, not inverse. Dark is also shorter.

Please use class `navbar-dark` instead of `navbar-inverse`.

Varunram:
@Johann-S May I put in a PR for this? Thanks!

Johann-S:
Yes you can @Varunram but I'm not sure if this change will be accepted

Varunram:
Thanks!

mdo:
We use inverse for cards, navbars, tables, and background utilities. For now, we'll leave them as-is.

Varunram:
Sure @mdo Thanks!

PR #22793, opened 2017-06-10
Proposal: Rename navbar-inverse to navbar-dark

Varunram:
Intended as a fix for #22745. I'm new here, so please do comment your views :)

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -239,25 +239,25 @@
 }
 
 // White links against a dark background
-.navbar-inverse {
+.navbar-dark {
   .navbar-brand {
-    color: $navbar-inverse-active-color;
+    color: $navbar-dark-active-color;
 
     @include hover-focus {
-      color: $navbar-inverse-active-color;
+      color: $navbar-dark-active-color;
     }
   }
 
   .navbar-nav {
     .nav-link {
-      color: $navbar-inverse-color;
+      color: $navbar-dark-color;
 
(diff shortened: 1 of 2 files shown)


Johann-S:
It seems you have some errors @Varunram see : https://travis-ci.org/twbs/bootstrap/jobs/241576992#L227

Varunram:
Ah, expected it. I thought I'd start off with the class definition first and based on feedback, go about editing the others. Will edit this shortly. Thanks for your time!

Varunram:
Done! If the changes are accepted, then we can modify the docs :)

akhyarrh:
What if user set darker color as base color ? Using `navbar-dark` but output bright color makes no sense

Varunram:
Well, the same logic applies to `navbar-light` too

akhyarrh:
As the linked issue stated, the problem is with classes name. `navbar-light` & `navbar-dark` or `navbar-inverse` is okay. But what if user use different base & second color

mdo:
Closing per https://github.com/twbs/bootstrap/issues/22745#issuecomment-308639412.

dinatih:
For people (like me) who thinks that it is not integrated, because of the closed status, it is integrated since 26 June. https://github.com/twbs/bootstrap/commit/36e482ed2728eec37188b5ec486283334bb3034f

also related issue : https://github.com/twbs/bootstrap/issues/22745

PR #22800, opened 2017-06-12
Use the box-shadow mixin for .btn on focus

valexiev:
Apply the box-shadow mixin to .btn:focus as well so it can be customizable by $enable-shadows.


mdo:
This one doesn't use a mixin to ensure there's always a focus state. Thanks though!

PR #22821, opened 2017-06-14
v4: Remove the confusing _custom.scss file

mdo:
Original issue: #17322. Original PR: #18398.

Honestly, this one is more pain than it's worth. Looking back, the file was originally added because I had left a comment in the variables file that you could copy-pasta some lines into that file for easier customization. I don't think that's a helpful way forward for a few reason:

- It's been brought up in so [many issues](https://github.com/twbs/bootstrap/issues?q=is%3Aissue+_custom.scss+is%3Aclosed).
- We flipped how it's used, by changing the import order (and then back again) after it screwed up some other stuff.
- In the current implementation, you cannot use other variables in the custom file—only copy-paste and change the values of existing ones. So, for example, you cannot add `$my-new-var: $blue`.

Nuking it in favor of relying on imports from package manager directories (where you're most likely to be using source Sass files) and manual forks of the project.


andersmoldin:
@mdo Could you perhaps describe, or point me to a good source on how best to customize boostrap now?

duggum:
@andersmoldin 

My solution was to create my own `_custom.scss` file and import `bootstrap.scss` after my style customizations:

```sass
@import "my-functions";
@import "my-sass-file";
@import "my-other-sass-file";

// bootstrap overrides
$body-bg: red;

// this has to use my custom function since bootstrap's functions get imported after
$link-color:  my-theme-color("info");

// import main bootstrap
@import "/path/to/node_modules/bootstrap/sass/bootstrap";
```

The only issue is I had to duplicate/make my own functions for choosing colors when overriding a variable that uses a function from bootstrap's `_functions.scss` file, such as:

```sass
// from bootstrap's _variables.scss
$link-color:  theme-color("primary") !default
```

gthomas-appfolio:
Am I correct in reading this as Bootstrap 4 does not support an 'official' way to create a customized theme any longer?

rcroeder:
MDO Please has someone remove http://v4-alpha.getbootstrap.com/getting-started/options/ on Customize Bootstrap 4, this is now misleading and deprecated .

mdo:
You’re looking at alpha docs with clear banner at the top that says there’s a newer release. The alpha docs stay; move to the newer beta docs.

Issue #22786, opened 2017-06-09
Navbar collapsed dropup caret should point down


Hopper262:
When in collapsed mode, a navbar's dropdowns and dropups behave identically (opening down). However, the triangle indicator still reflects the non-collapsed behavior, which for dropups, points away from the content.

Here's a test case on alpha 6, based on one of the documentation examples: https://jsfiddle.net/2htxfvz1/

Hopper262:
The following CSS fixes the issue in my own use cases:

    @media (max-width: 767px) {
    .navbar-toggleable-sm .dropup .dropdown-toggle::after {
      border-top: .3em solid;
      border-bottom: 0;
    }
    }
    @media (max-width: 992px) {
    .navbar-toggleable-md .dropup .dropdown-toggle::after {
      border-top: .3em solid;
      border-bottom: 0;
    }
    }

Johann-S:
If you want that behaviour you should add `.navbar-expand`
This class prevent navbar from expanding.
In a mobile first design this is a normal behaviour

Hopper262:
No, the behavior itself is working perfectly fine. This is about a visual glitch:

![caret-bug](https://user-images.githubusercontent.com/2176248/26985597-2da51d36-4d12-11e7-9be0-baff251a2798.png)

I would expect that arrow to point downward, since the full-width menu appears below the toggle.

Varunram:
Correct me if I'm wrong, but in the example you've used the class `dropup` which specifies for the caret to be bottom up. On removing the same, it defaults to the behaviour expected. Any reasons why? Thanks!

Hopper262:
This is a fixed-bottom menubar. At desktop window sizes, the menus appear horizontally and I do want them to open upward toward the middle of the screen. At mobile resolutions, there's no behavioral difference between `dropdown` and `dropup`, but there's a valid reason to select the latter for desktops.

Basically, the problem is that the responsive behavior switches the direction of dropup menus at certain screen sizes, but it fails to switch the direction of the caret.

Varunram:
Thanks for your clarification! Made it more clear for me to approach the same.

PR #22823, opened 2017-06-14
Modify caret mark behaviour wrt screen size

Varunram:
Intended as a fix for #22786. I'm new here, so feel free to comment your issues. One issue from my side is that I hardcoded the value `768px` into the file. Is this fine or do I use the functions from `_breakpoints.scss`?

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -27,9 +27,17 @@
 // Just add .dropup after the standard .dropdown class and you're set.
 .dropup {
   .dropdown-toggle {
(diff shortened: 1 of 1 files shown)


mdo:
This is the wrong place to do this to fix #22786. This PR would affect all `.dropup`s, when we only really care about `.dropup`s in `.navbar`s.

> Is this fine or do I use the functions from `_breakpoints.scss`?

Yes, you'll want to use the media queries. There are examples in many of the Sass files.

PR #22826, opened 2017-06-14
Update Popper.js 1.10.1

Johann-S:
Update Popper.js to 1.10.1 and use relative paths in our examples


FezVrasta:
Is this process automatized? If not, it seems like a PITA to perform on each new version 🤔

PR #22848, opened 2017-06-17
Docs: accessibility and IE fixes for SVG-based top navigation

patrickhlauke:
- fixes issue of some browser/AT combinations not reading an `<svg>`'s `<title>` for the link text (by adding an explicit `aria-label` to the link)
- fixes issue of IE explicitly focusing on `<svg>` elements (resulting in two tab stops for each `<svg>`-based link - one for the link, one for the nested `<svg>`)


patrickhlauke:
hmmm, travis is taking an inordinately long time to run here it seems...

PR #22858, opened 2017-06-18
Add text-align to `<body>` so we can inherit it in `<th>`s

### LINKED ISSUE #22519: .text-center inside th
By using `<table class="table text-center">` the text is only centered in `<td>` but not in `<th>`.

Happens at the latest version:
`<link rel="stylesheet" href="http://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css">`

mdo:
This sets an explicit `text-align` on the `<body>` element so we can use `inherit` to override the default `<th>` `text-align` value. Most browsers center this text, so we previously applied a specific `text-align: left` to the `<th>`s. That had a side effect of not responding in the same way as `<td>`s do when setting an alignment value on the `<table>`.

Fixes #22519.


mkamensky:
Would it be possible to change this into `text-align: start`? Otherwise it breaks things on RTL pages (generally speaking, it would help to change all the left/right into start/end, respectively)

PR #22860, opened 2017-06-18
Redo .card-deck and .card-group so they're always flex

### LINKED ISSUE #22206: Card order not working on XS devices; .card-deck is not a flex
Rearranging cards does not work on mobile devices as expected. 
Setting an "order" has no effect because .card-deck on XS devices is a "display: block" instead of a "display: flex".

I would suggest to use flexbox even on XS devices so ordering of the cards in CSS is possible:

```
.card-deck {
    display: -webkit-box;
    display: -webkit-flex;
    display: -ms-flexbox;
    display: flex;
    -webkit-flex-flow: row wrap;
    -ms-flex-flow: row wrap;
    flex-flow: row wrap;
}
```

mdo:
This rearranges the Sass source code for card decks and groups so that they are always flex containers. It also addresses the vertical spacing for them on the xs breakpoint by adding some default margin there (that's later overridden).

Fixes #22206


gijsbotje:
@mdo i'm getting this behavior on the card-groups on the currect v4-dev branch, and i think it's caused by this pr
schermafbeelding 2017-08-16 om 20 45 43

is the margin on the bottom supposed to be overridden?
i'll happily make a pr to fix it

mdo:
Hmm yeah margin should be overridden. Those should be equal height.

gijsbotje:
i'll see if i can fix it

gijsbotje:
@mdo PR is up #23505

PR #22863, opened 2017-06-18
Buttons cursor is not pointer / no-allowed

artyomtrityak:
Buttons hover state for `btn` class does not have `cursor: pointer` or `cursor: not-allowed` as it used to be in v3.

Even for CSS spec `cursor: pointer` is primarily designed for links (`<a/>` tag) it is still very useful for modern web to have indication that this button is clickable via cursor change.

Optional: `cursor: not-allowed` is much less viable but to be consistent and keep backward compatibility it was added either.

It is possible to check difference here:
v3 http://getbootstrap.com/components/#btn-dropdowns
v4 https://v4-alpha.getbootstrap.com/components/buttons/

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -12,6 +12,7 @@
   vertical-align: middle;
   user-select: none;
   border: $input-btn-border-width solid transparent;
+  cursor: pointer;
   @include button-size($input-btn-padding-y, $input-btn-padding-x, $font-size-base, $input-btn-line-height, $btn-border-radius);
   @include transition($btn-transition);
 
@@ -28,6 +29,7 @@
   // Disabled comes first so active can properly restyle
   &.disabled,
   &:disabled {
+    cursor: not-allowed;
     opacity: .65;
     @include box-shadow(none);
   }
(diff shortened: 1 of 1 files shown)


skmedix:
See https://github.com/twbs/bootstrap/issues/22313#issuecomment-290381725

Johann-S:
the decision was made mainly in light of this https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

artyomtrityak:
@Johann-S @skmedix It is opinion right? For example Chrome itself still uses button-pointer in many places like extensions tab etc.
This is one of the examples. 

Extensons tab trash btn:
```css
.trash {
    -webkit-appearance: none;
    background: none;
    border: none;
    cursor: pointer;
    display: inline-block;
    outline: none;
    padding: 0;
    position: relative;
    width: 30px;
}
```

Main menu styles:

```css
button {
    background-color: white;
    border: 0;
    color: #999;
    cursor: pointer;
    font: inherit;
    line-height: 1.417em;
    margin: 6px 0;
    padding: 0;
}
```

It __maybe__ make sense to remove hover but it's almost a standard buttons-hover over internet.

PR #22864, opened 2017-06-18
Fix modal overflow with our navbar documentation and fix typo

Johann-S:
Some quick fix : 
- Typo: usally -> usually 
- Modal overflow with our documentation navbar see : 

modal bootstrap - google chrome

/CC @mdo

### DIFF
--- a/assets/scss/_component-examples.scss
+++ b/assets/scss/_component-examples.scss
@@ -285,6 +285,10 @@
   }
 }
 
+.modal.show {
+  z-index: 1072;
+}
+
 // Example tabbable tabs
 .bd-example-tabs .nav-tabs {
   margin-bottom: 1rem;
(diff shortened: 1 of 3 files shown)


assets/scss/_component-examples.scss
```diff
@@ -285,6 +285,10 @@
   }
 }
 
+.modal.show {
+  z-index: 1072;
```
mdo: Does this happen to all navbars or just the one for our docs?

Johann-S: IMO it's just for the one in our documentation because we want this navbar on top of everything

mdo: We'll also need to move the backdrop, right? The navbar doesn't need to be above the modal and it's backdrop here.

PR #22869, opened 2017-06-19
Fix removed ".modal-open" in "body"

krzysztofgwiazda:
When I close modal and another is open. Class ".modal-open" in "body" tag is removed

ybkzjvjgqxs8kbvbxc8-sw


Johann-S:
Sorry but Bootstrap do not support multiple modal open for us it's a bad practice

PR #22874, opened 2017-06-20
`$list-group-color` isn't used in `.list-group`

pat270:


mdo:
This isn't needed given it's value is the `<body>` default color—it should just inherit. See 4e067f722.

PR #22879, opened 2017-06-21
Update Carbon Ads script

sayzlim:
Replaced Carbon ad code with the latest version that is whitelisted
from https://adblockplus.org/. The preview for both pages can be seen
on:

Homepage: https://s.codepen.io/sayzlim/debug/dRWQPG
Docs: https://s.codepen.io/sayzlim/debug/PjmxEQ

### DIFF
--- a/docs/_includes/ads.html
+++ b/docs/_includes/ads.html
@@ -1,1 +1,1 @@
-<div id="carbonads-container"><div class="carbonad"><div id="azcarbon"></div><script>var z = document.createElement("script"); z.async = true; z.src = "http://engine.carbonads.com/z/32341/azcarbon_2_1_0_HORIZ"; var s = document.getElementsByTagName("script")[0]; s.parentNode.insertBefore(z, s);</script></div></div>
+<script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?zoneid=1673&serve=C6AILKT&placement=getbootstrapcom" id="_carbonads_js"></script>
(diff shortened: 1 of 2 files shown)


Johann-S:

But we use Carbon in `v4` too, so do not hesitate to create a PR which target v4

sayzlim:
Got it, the v4 is already using the latest Carbon ad code. We still see large amount of impressions from v3, so updating the ad code to latest version is going to receive the benefit from additional impressions.

bardiharborow:
Pinging @mdo as this may affect the bottom line on your end.

mdo:
This is on my list to audit and update with our next release :).

docs/_includes/ads.html
```diff
@@ -1 +1 @@
-<div id="carbonads-container"><div class="carbonad"><div id="azcarbon"></div><script>var z = document.createElement("script"); z.async = true; z.src = "http://engine.carbonads.com/z/32341/azcarbon_2_1_0_HORIZ"; var s = document.getElementsByTagName("script")[0]; s.parentNode.insertBefore(z, s);</script></div></div>
+<script async type="text/javascript" src="//cdn.carbonads.com/carbon.js?zoneid=1673&serve=C6AILKT&placement=getbootstrapcom" id="_carbonads_js"></script>
```
XhmikosR: I'd remove the type since it's the default and also use `https` explicitly.

mdo:
Closing as out of date. Will tackle in `gh-pages` branch.

PR #22887, opened 2017-06-22
Remove dots in installation option lines

maximium:
Makes copy-paste easier


Johann-S:
Not present in our `v4-dev` branch and Bootstrap 3 is essentially in maintenance mode as we focus on working towards a stable v4. As such, we're only accepting changes to v3's code on a case-by-case basis, usually only for critical bug fixes or docs improvements.

Issue #22783, opened 2017-06-08
[Question] Should we provide a bundled release of Bootstrap with Popper.js


Johann-S:
Hi folks,

Currently if you want to use Tooltips/Popovers and Dropdown, you should include Popper.js before Bootstrap and I think it's a huge drawback that you cannot just include Bootstrap and all the provided plugins doesn't work.

So a simple solution should be to provide a bundled release of Bootstrap with Popper.js included.
The size of our minified JS will increase of 16kb see my test : 

![js](https://user-images.githubusercontent.com/1689750/27172372-4afd176a-51b5-11e7-8037-7c6c70f057a5.png)

Is it a good solution for you ? Or should we continue to provide a release without Popper.js ? 

/CC @mdo

acshef:
Why not both?

Johann-S:
Yes that's a possibility 👍

mdo:
Yikes, that's a 26kb increase, not 16kb. 😬  A 50% increase is pretty bad.

Johann-S:
Yes a bit strange because Popper's weight is 19kb 🤔

deandob:
Are tooltips that difficult that a 3rd party dependency is needed for bootstrap? Another thing to go wrong, additional load time (as needs to be loaded before bootstrap) is a disadvantage for v4. If it has to be a separate plugin, agree with Johann-S to provide a pre-bundled package.

Johann-S:
It's a decision we made at the beginning of `v4` release, because they are a lot's of use case to handle for positionning Tooltips (scroll, overflow, ...)

pat270:
I think it should be bundled with Bootstrap.js. Popper documentation is kind of confusing. I couldn't find a link to their cdn on https://popper.js.org/. I had to go to their github repo to find a link to the cdn which provides a confusing list https://cdnjs.com/libraries/popper.js, then pick and choose to find the right one.

Johann-S:
Don't worry @pat270 we will provide the appropriate Popper.js link in our documentation 👍

Sungodragon:
Maybe, it is better if its bundled release.
Perhaps someone can investigate to make the bundle size smaller. 
@vanduynslagerp

Johann-S:
I have some issue to build our `dist` files that's, I fixed that and finally this is the correct screenshot : 

![image](https://user-images.githubusercontent.com/1689750/27172372-4afd176a-51b5-11e7-8037-7c6c70f057a5.png)

[...]

PR #22888, opened 2017-06-22
Better modularization of our plugins

Johann-S:
This PR will allow folks to import only modules they want to use for example : 
```javascript
import { Util, Modal } from 'bootstrap'
```

This PR introduce the use of `rollup` and add two new dist files : 
- `bootstrap.bundle.js`
- `bootstrap.bundle.min.js`

This two files include Popper.js so when you include those files in your HTML pages you don't have to include Popper.js before Bootstrap.

/CC @mdo @petetnt @bardiharborow @hnrch02 
Close: #19017, #22783

Edit : 
/CC @IdanCo if they are something to change in our webpack get started example
I know the branch of this PR is `webpack` but finally `rollup` allow us to have smalls dist files


Johann-S:
/CC @IdanCo if they are something to change in our webpack get started example

IdanCo:
What a blessed PR! Finally bootstrap is aligning with modern module systems! Can't wait to dig in, probably this weekend

js/tests/visual/alert.html
```diff
@@ -45,6 +45,7 @@
     </div>
 
     <script src="../../../assets/js/vendor/jquery-slim.min.js"></script>
+    <script src="../../../assets/js/vendor/popper.min.js"></script>
```
Johann-S: Something to remove 😆

petetnt:
Cool stuff! I'll be pretty offline for the finnish midsummer time, but I'd love to review this early next week!

FezVrasta:
I'd suggest to go with Rollup instead of Webpack since it's best suited for libraries bundling (the webpack maintainers think the same).

This would also reduce the bundle size

Johann-S:
I'll give another try to Rollup in another branch and if I have something something better I'll incorporate that here thanks @FezVrasta 👍

Johann-S:
First try with Rollup for our `dist/js/bootstrap.js` : 
- With webpack : `132 kb`,
- With Rollup : `106 kb`

I'll switch to Rollup 😄

Johann-S:
Switch to `rollup` done 👍

package.json
```diff
@@ -57,7 +60,7 @@
   },
   "style": "dist/css/bootstrap.css",
   "sass": "scss/bootstrap.scss",
-  "main": "dist/js/bootstrap",
+  "main": "src/js/index",
```
IdanCo: not sure what you were aiming for, but no need to change the main file

Johann-S: Because with that entry point we can load only required plugins it allow folks to make small bundle

FezVrasta: You shouldn't point to the source but to a es2015 transpiled version

FezVrasta: You shouldn't point to the source but to a es5 transpiled version

IdanCo: I agree with @FezVrasta (and in any case i think you meant `js/src` and not `src/js`)

Nevertheless, there is still some issue with importing specific modules (it happens with either `main` file) . I havn't figured out the root cause, but the symptom is that both `import "bootstrap"` and `import { module } from "bootstrap"` ends up loading **ALL** modules.

To test it i've created a [new branch of an empty scaffold project](https://github.com/IdanCo/webpack-modular/tree/test-bootstrap-webpack), and added @Johann-S [latest commit](https://github.com/twbs/bootstrap/pull/22888/commits/ffca50924eb519fe62bccf30c9d899c12eaf887c) as a dependency. Feel free to clone and do `npm install`. Then `npm run serve` will run the app with the new dependency. **notice** that you should manually fix "package.json" in the bootstrap directory to point to the right file (`dist/js/bootstrap`) before serving.

And on a side note - great job @Johann-S ! With this new build, importing bootstrap with webpack (or any other bundler for that matter) becomes a simple one liner. Sadly it will render the entire installation guide i wrote obsolete, but i'll cope ;-)

FezVrasta: Sorry I forgot to mention that the Es5 + es modules version should be defined in the "module" property, while the es5 w/o es modules (umd usually) should point to "main"

petetnt: Re: `ends up loading ALL modules`  issue is documented in my original issue https://github.com/twbs/bootstrap/issues/19017. Re-exported modules will call all the IIFEs when the index file is evaluated, leading into namespace pollution.  See https://codesandbox.io/s/vLXO14lg for an example.

In my [branch](https://github.com/petetnt/bootstrap/commits/v4-js-modules) I opted in for having module users to initialize their own components with `ModuleNameHere.init()`, while the browser bundles (including single modules, like tooltip.js) auto-initialize themselves with a secret variable ([[1]](https://github.com/petetnt/bootstrap/blob/v4-js-modules/js/src/index.js) [[2]](https://github.com/petetnt/bootstrap/blob/v4-js-modules/js/webpack.config.js#L49-L54)) which solves the problem but creates some cognitive load for the end user.

petetnt: Another way would be lazily evaluate the modules with `import()` call, but not sure if that would either.

FezVrasta: Lazy load is overkill, better use named experts for each plugin/module so that people can import just what they need

petetnt: That's only possible by importing the modules with `import Tooltip from 'bootstrap/js/some-module-name'`. Importing the modules anywhere (like in `index.js`) evaluates the files as evident in https://codesandbox.io/s/Mjgm6qzAm, rendering `import { SomeModule } from 'bootstrap'` pretty moot as it's always equivalent of `import Bootstrap from 'bootstrap'`.

Userland could adapt some bundler that's smart enough to tree-shake the other modules but even that might be a scretch. Do correct me if I am wrong though.

petetnt:
I was thinking about the modularization part last night and my conclusions were something like this (if my idea of https://github.com/twbs/bootstrap/pull/22888#discussion_r123908978 is correct):

- Without any escape hatch for preventing auto-init importing all the modules essentially boils down to `import 'bootstrap'` as the modules don't really have that many static methods available that would be useful outside of the modules themselves.
- Importing `import { Tooltip, Modal } from 'bootstrap'` most likely is not going to happen without any side-effects...
- however, releasing the modules as namespaced versions (for example `@bootstrap/tooltip`, `@bootstrap/modal` on `npm` would provide modularity for those who want to include only parts of the modules themselves (which is actually pretty nice, as proven for example by https://lodash.com/ and its namespaced modules (https://www.npmjs.com/browse/keyword/lodash-modularized).
- one could also import the modules from the folders with `import 'bootstrap/js/tooltip'` or `import 'bootstrap/js/modal'`.

Johann-S:
I have some tests to do with all your comments thanks @petetnt, @FezVrasta and @IdanCo 

I'm not sure if it's the best way but something like this works very well to import only needed plugins : 
`import Modal from 'bootstrap/js/dist/modal'`

FezVrasta:
This is the same as `import { Modal} from 'bootstrap'`, but the latter works better and follows the conventions

Johann-S:
@petetnt that's right what did lodash here : https://github.com/lodash/lodash/tree/npm-packages allow to export plugins separately, but it seems pretty hard to maintain a separate branch only for our js plugins

FezVrasta:
This means there's something wrong with your code. I'd say you may want to merge this PR just to provide the needed tooling. Then we may work together on future PRs to improve the code splitting.

IdanCo:
I think you're in the right direction. IMO, loading the entire library should remain as simple as possible. importing individual modules is usually the doing of more advanced users, in which case bootstrap can allow itself a bit more complex implementation.

In any case i'll do some testings later today to see if it's possible to hold the stick on both ends by fixing the "all or nothing" issue.

petetnt:
@FezVrasta There's nothing wrong with the code, as I stated above, `import { Modal } from 'bootstrap'` is not equivalent of `import Modal from 'bootstrap/js/dist/modal'. It's just that the `index.js` file is evaluated when `import * from 'bootstrap'` is called like I explained before.

I think `lodash` is now built with [lodash-cli/](https://github.com/lodash-archive/lodash-cli/) nowadays but building 9 individual files with RollUp and creating a simple release script would not be too time consuming, it can be done later too though.

Johann-S:
I wont merge this PR untily Beta 1 is shipped, but you can submit a PR which target this branch @FezVrasta 👍 

Thanks @IdanCo, do not hesitate to let us know what result you had

@petetnt with this PR rollup build seperately each of our files with `build/build-plugins.js` script but yes it's a good idea to create a release script for that 👍 

With all your feedbacks a good beginning should be what we have currently, allow user to import only needed plugins with `import { <plugin name> } from 'bootstrap/js/dist/<plugin name>` and for folks who needs the entire libs allow us to do `import { <plugin name> } from 'bootstrap'` or simply `import 'bootstrap'`

FezVrasta:
@petetnt the whole idea about named exports is that you can import the library and have three shake remove unused code super easily.

Doing `import Modal from 'bootstrap/js/dist/modal'` is just a way to import a single file, but it doesn't take advantage of the ES Modules architecture and relies on internal structure of the library, which may change in the future and break others code.

petetnt:
@FezVrasta I already mentioned userland treeshaking in https://github.com/twbs/bootstrap/pull/22888#discussion_r123910574.  The issue is that will tree shakers eliminate the IIFEs that pollute global namespace, especially when it's done against the already-once built files (when things like Webpack 2 have had problems with treeshaking and re-exports in general: https://github.com/webpack/webpack/issues/2867).

If it works, then fine, but the side-effect is still unwanted and unexpected in general.

FezVrasta:
@petetnt I think we should look forward and comply with the modern standards (in this case, named exports + three shaking) and don't stagnate on old habits.

In this case, IIFE and UMD aren't bundle formats that ever considered three shaking as part of them, and nobody is trying to make them work with it.
We should provide ability to three shake just to who actually uses a tooling capable to handle it properly.

js/src/index.js
```diff
@@ -0,0 +1,32 @@
...
+import Util from './util'
+
+/**
+ * --------------------------------------------------------------------------
+ * Bootstrap (v4.0.0-alpha.6): index.js
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ * --------------------------------------------------------------------------
+ */
+
+export default {
```
Johann-S: Currently I export our plugins with `export default` but I'm not sure if this is the good way to have allow folks to : 
`import { Util, Modal } from 'bootstrap'` ?

FezVrasta: Nope, if you `export default { stuff }` you are exporting an object with inside the modules.

If you want to allow users to import single modules, you have to do `export { stuff1, stuff2, etc }`

Johann-S: Thanks @FezVrasta 👍

FezVrasta:
Related to this:
https://github.com/twbs/bootstrap/pull/22888#issuecomment-311031927

Nothing prevent us from providing both the solutions, if people use modern toolings, they will want to use the ES5 + ESM bundle to take advantage of three shaking.

If they use old tooling, they'll want to use the UMD/IIFE bundle, and, in this case, we may provide sub-bundles with the specific modules inside.

Johann-S:
I agree with @FezVrasta, if I try to allow folks to only import specific plugins it's to allow them to have small bundle when they import bootstrap in their projects so tree-shaking is an important thing to concider here

IdanCo:
Hope i'm not barging into an open door, but looks to me that the main problem with implementing a truly modular architecture is that each module is also initializing itself with an IIF such as this -
```js
$(document).on(
  Event.CLICK_DATA_API,
  Selector.DISMISS,
  Alert._handleDismiss(new Alert()) // <-- BOOM!!
)

```
from https://github.com/twbs/bootstrap/blob/webpack/js/src/alert.js#L168

Johann-S:
I'm sorry @IdanCo, I made some test on Bootstrap and I don't have any errors with Alert plugin, maybe you can add some context when this error happened

FezVrasta:
I think it's not about errors but more about the fact that the modules will auto initialize when you import them?

IdanCo:
@FezVrasta - Bingo for the clarification, sorry for not making it clearer.

And also your solution sounds promising!

Johann-S:
I understand your issue, but if we don't find any alert using our data-api, they aren't any self initialisation.

@petetnt made a solution with an `init` method to initialise our plugins, but to use that we have to detect in which environment Bootstrap is used

IdanCo:
> @petetnt made a solution with an init method to initialise our plugins, but to use that we have to detect in which environment Bootstrap is used and call this method only in a browser environment >

I see now -
https://github.com/petetnt/bootstrap/commits/v4-js-modules
looks interesting!

Apologies for the ignorance/laziness, but can someone explain to me
A. Why the env detection is necessary? 
B. Assuming that it is necessary, what's wrong with using it?

petetnt:
@FezVrasta @Johann-S yep! My branch used a hidden variable to see the environment the script was used in. I think it was posted somewhere already, but here's the relevant lines for Webpack:

https://github.com/petetnt/bootstrap/blob/v4-js-modules/js/webpack.config.js#L49-L54
https://github.com/petetnt/bootstrap/blob/v4-js-modules/js/src/index.js#L13-L24

and example usage in https://github.com/petetnt/bootstrap/blob/v4-js-modules/js/src/dropdown.js#L264-L304

FezVrasta:
That's why I suggested to auto init everything on `bootstrap.js` import, and don't load in case of module import.

The module import is something new, so people before use it will have to learn about it, and the page where they learn it will explain how to init them as well.

Johann-S:
So @FezVrasta what you suggest is what we have with this PR ?

FezVrasta:
I don't think?

I mean:


```
// modal.js
export default class Modal {
  init() {}
}
```

```
// tooltip.js
export default class Tooltip {
  init() {}
}
```

```
// index.js
import Modal from './modal';
import Tooltip from './tooltip';
Modal.init();
Tooltip.init();

export { Modal, Tooltip };
```

Johann-S:
Thank you for the clarification 👍 
What you described init plugins in the `index.js` what's the difference between initialise our plugins in the plugin file and in the index.js ?

I looked for the best way to detect if we are in a Node environment or if we are in a Browser environment and we can check if `global` and `window` are undefined to call `init` in case of a Browser and not call `init` in a Node environment

IdanCo:
@FezVrasta - nice direction but using this

```js
// index.js
import Modal from './modal';
import Tooltip from './tooltip';
Modal.init(); // <--IIF *********
Tooltip.init(); // <--IIF *********

export { Modal, Tooltip };
```

Still leaves you with IIFs, only now they're in index.js. So even if i only import Modal, i'll still get Tooltip

FezVrasta:
@IdanCo you are right 🤔

Johann-S:
Maybe we should let user call `init` method for plugins they want to use... I'm not sure it's the best solution but with that it's avoid global polution

IdanCo:
It is possible, but i would still want to enable users to drop the bootstrap.js file in their HTML (like we did in the good ol' days) and run with it. So i think we should keep looking for some kind of a combined solution.

Can this be useful? 
https://github.com/umdjs/umd/blob/master/templates/jqueryPlugin.js

Johann-S:
It's how to bundle a libs with an UMD support @IdanCo 

Maybe we can listen for `document` ready event and init our component when this event is triggered by the `window`

IdanCo:
>It's how to bundle a libs with an UMD support @IdanCo
>
But the rollup/webpack implementation of UMD is indifferent to whether or not we use jQuery. I though perhaps in this implementation there's a clue to how we can use it...

>Maybe we can listen for document ready event and init our component when this event is triggered by the window
>
I like it! and have every module register a callback for that event... worth a shot

Johann-S:
@IdanCo I'll look at that thank you 👍 I made a shot to init our plugins only on a browser environment and when DOM is ready

Johann-S:
Maybe @FezVrasta can help me on that but, if I compile our plugins in ESM will it be possible to do : `import { Modal } from 'bootstrap'` ?

FezVrasta:
It depends if you are using named exports or not

Johann-S:
I want to compile this file : https://github.com/twbs/bootstrap/blob/webpack/js/src/index.js in ESM (and compile other plugins too because import will fail if the other plugins aren't in ESM) and we are using named export here.
So in theory ESM will allow us to do that ? 
I just have to update the `module` in `package.json` with index (in ESM) path, that's right ?

FezVrasta:
Yup theorically it should work, but as we discussed previously, three shaking may not work if you auto initialize the plugins.

Johann-S:
I changed that, I added the init of our plugins only when DOM is ready so I'm not sure this is auto initialisation

Johann-S:
Maybe a good solution should be to have an `indexUMD.js` which auto initalize our plugins and an `indexESM.js` for modules

FezVrasta:
Ideally the different bundles should work all the same, just with different module spec.

It should be easy to test if the three shaking works tho, just bundle a js file that imports and uses just one plugin with Rollup and see what happens.

Johann-S:
That's what I do, but even if I remove the DOM ready listener everything is included 😟 
maybe my rollup config isn't good : 
```javascript
import resolve from 'rollup-plugin-node-resolve';
import babel from 'rollup-plugin-babel';

export default {
  entry: 'index.js',
  format: 'umd',
  moduleName: 'test',
  dest: 'index.bundle.js',
  plugins: [resolve(), babel({
	  exclude: 'node_modules/**' // only transpile our source code
  })],
  external: ['jquery'],
  globals: {
	jquery: '$'
  }
};
```

and my index.js 
```javascript
import $ from 'jquery'
import { Util } from 'bootstrap'
console.log(Util)
```

FezVrasta:
is the rollup config used to bundle bootstrap or the test script? because in the first case it should be `es`, not `umd`

Johann-S:
No that's my test script, I'll push compiled ESM files

Johann-S:
If they are folks who want to test that in Local this my test folder : 
[testBsModules.zip](https://github.com/twbs/bootstrap/files/1133103/testBsModules.zip)

To use that you have to pull this branch in the root of your bootstrap folder run : `npm link`
After that you have to go to the root of my test folder and run : `npm link bootstrap`

everything is good 👍 you just have to install deps (`npm i`) and run `npm run build`

XhmikosR:
@addyosmani: do you have any idea how to solve this? Thanks in advance!

> To sum up if someone trys to build a project with Bootstrap and just want to include our `Util.js` with `import { Util } from 'bootstrap'` currently the other source files are added too... They isn't any tree shaking done.

mdo:
We, uh, have some conflicts here lol.

Johann-S:
Yep I'll take care of that 😆 
About that PR I can't go further here, everything is done, if someone have other feedbacks don't hesitate 👍

addyosmani:
@XhmikosR  Just saw this. Sorry for the delayed response. 

@Johann-S Do you know to what extent your latest commits using RollUp are failing to tree-shake as expected? Do you still see the problem of other modules (e.g beyond the 'Util' example) getting pulled in?

Johann-S:
@addyosmani yep I still have this issue, when we import a module every others modules are bundled too... 
Maybe that's because Bootstrap depends a lot of jQuery.
I read a few days the Troubleshooting section of Rollupjs (see : https://github.com/rollup/rollup/wiki/Troubleshooting#tree-shaking-doesnt-seem-to-be-working) and it indicate that if a library is too complex Rollup won't tree shake because they don't want to break everything.

If you have any idea about what I did wrong ? Or any other ideas it would be great, thank you 👍

build/rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+
+module.exports = {
+  entry: path.resolve(__dirname, '../js/src/index.js'),
+  format: 'umd',
+  moduleName: 'bootstrap',
+  dest: path.resolve(__dirname, `../dist/js/${fileDest}`),
+  external: external,
+  globals: {
+    jquery: '$',
+    'popper.js': 'Popper'
```
FezVrasta: These dependencies should be defined as external, but not global. Bootstrap should actively try to import them (mostly to fix #23381)

Johann-S: if I remove globals I have those warnings :
image

FezVrasta: I think it's because you don't have any `import $ from 'jquery'` and `import Popper from 'popper.js'` in the code?

Johann-S: That's wrong in this PR we import Popper.js and jQuery (for example see : https://github.com/twbs/bootstrap/blob/webpack/js/src/tooltip.js)

FezVrasta: Sorry, probably you still need those for the UMD and IIFE bundles, but you must make sure no never reference to them in the source, but always require them with `import`

Johann-S: we always import our dependencies with `import` thanks to this PR 😃

FezVrasta: Probably it would make sense to add some automated integration tests to check that the various ways to import Bootstrap work fine?

I can think about UMD and ESM, not sure fyou want to support IIFE

Johann-S: Yes automated integration tests would be rad ! 👍 

I checked for those 3 environments and it works pretty well but not sure how to automise that 😟

FezVrasta: So does it work without requiring the `window.Popper = Popper`?

Johann-S: For ESM and the Bundled release (Bootstrap + Popper.js) yes

I have a doubt for UMD 😟 

Do you know how to add integration tests ?

FezVrasta: Umh nope, you should create HTML pages with the needed code and have them report back something when Bootstrap is properly loaded. But I'm not sure how to achieve this.

Johann-S: If someone want to use Bootstrap in a UMD module a rollupjs config like this would work perfectly without : `window.Popper = Popper` : 

image

FezVrasta: With this we can do `require(['jquery', 'popper.js', 'bootstrap'], ($, Popper, Bootstrap) => {})`?

Johann-S: Currently I have this issue : 
⚠️   'jquery' is imported by ..\bootstrap\js\dist\esm\scrollspy.js, but could not be resolved – treating it as an external dependency

So I'm not sure in an UMD build everything works too...

Johann-S: My ESM files import jQuery and that's why I have this issue 😩

Johann-S: jQuery is difficult to import... For now I cannot make it work without something like this : 
`<script src="https://code.jquery.com/jquery-3.2.1.slim.min.js"></script>`
But for Popper.js it works perfectly

Do you have any ideas @FezVrasta  ?

FezVrasta: Nope sorry, It's been a while since last time I used jQuery :/

Johann-S: I've got to be sure but it seems that's because jQuery do not have an ES6 release 😟

prograhammer:
Has anyone seen this: 

> To get most out of tree shaking with external packages, you have to use babel-plugin-transform-imports to rewrite imports so that they work with webpack's tree shaking logic. See [webpack issue #2867](https://github.com/webpack/webpack/issues/2867) for more information.

From Webpack's docs [here](https://survivejs.com/webpack/optimizing/tree-shaking/). It seems tree shaking doesn't work on external packages yet. You would have to explicitly do `import Bar from 'foo/bar'`.

FezVrasta:
Here they are using Rollup, not webpack

Johann-S:
Closing for now, I'll come back with a better PR, thanks to all you helped me a lot

PR #22890, opened 2017-06-22
fix a couple dead links to the navbar docs

scottpetrovic:
not sure if we want to update these as they now link to page not found. There are more of them, but wanted to get some feedback before I might update any more.


scottpetrovic:
I am playing around with these links and it is an interesting dilemma. The HTML examples just show the source code if you view it on github, so the links themselves aren't really that useful.

I personally cloned the project and was viewing the HTML pages in Chrome. Of course the MD files don't render the markdown in Chrome, so the links show as plain text.

mdo:
These links are correct for building the compiled docs.

mdo:
Expanding on that, our docs are compiled with Jekyll. There are source Markdown files, yes, and they can also be browsed on GitHub, but they're not the final version.

PR #22895, opened 2017-06-22
fix location of favicons so it loads when viewing the examples. 

scottpetrovic:
This is the only favicon I found in the project, so I am assuming it is the right one to point to


scottpetrovic:
this might not be correct too. I might have to get a Jekyll instance going to see how this code would be in its final form. I didn't realize that the code in this area is used for another process.

PR #22896, opened 2017-06-22
Error in current navbar documentation

lemuelbarango:
Currently, the docs indicate that the navbar toggler could be on the left but the example in the documentation shows otherwise.

localhost-9001-docs-4 0-components-navbar- 1

I have updated the example and here is what it shows at the moment:
localhost-9001-docs-4 0-components-navbar-


mdo:
Closing as we picked this up elsewhere it seems. Thanks!

PR #22905, opened 2017-06-24
Update _card.scss

liuguilai:
i want to try one request function


jipexu:
hi 
why to try here ? ..  you are wasting time on the developer ...

mdo:
Do not test on public repos. You just notified thousands of people in the process.

PR #22912, opened 2017-06-26
New docs page: Extending Bootstrap

mdo:
Right now this is a placeholder PR to stir some conversation around how to _extend_ Bootstrap. This might not be the right list to start with, but I wanted to open the PR to gather more feedback before jumping into it full-steam. Some of this might make more sense in our Getting Started pages, but let it rip if you have ideas no matter the place just yet.

*Will aim to tackle this after Beta 1.*

### DIFF
--- a/docs/4.0/extend/index.md
+++ b/docs/4.0/extend/index.md
@@ -3,4 +3,13 @@ layout: docs
 title: Extend
 ---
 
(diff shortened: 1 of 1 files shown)


mdo:
This is covered in other pages.

Issue #22916, opened 2017-06-27
Progress margin-bottom deleted


marcoscv-work:
In 4.0.0-alpha.6

Progress works with margin-bottom: X (default 20px) in previos bootstrap versions.

This change make more difficult migrations between BS3 and BS4.
To make it easier I think we should persist this margin-bottom.

mdo:
Leaving it without the margin given our utilities. Thanks though!

PR #22918, opened 2017-06-27
Add margin-bottom to .progress

marcoscv-work:
For https://github.com/twbs/bootstrap/issues/22916

### DIFF
--- a/scss/_progress.scss
+++ b/scss/_progress.scss
@@ -6,6 +6,7 @@
 .progress {
   display: flex;
   overflow: hidden; // force rounded corners by cropping it
+  margin-bottom: $progress-margin-bottom;
   font-size: $progress-font-size;
   line-height: $progress-height;
   text-align: center;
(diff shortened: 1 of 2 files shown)


afuno:
It seems to me it is superfluous. For these tasks there are corresponding helpers: https://v4-alpha.getbootstrap.com/utilities/spacing/

marcoscv-work:
Spacing utilities are great! 💃

But I think we should care more the HTML markup changes between versions, if not the jump from BS3 to BS4 will be more complex.

Tanks!

afuno:
@marcoscv-work There is nothing complicated in this migration. In addition, no one prevents you from rewriting this class in person. No need to spoil the logic of this version.

Do you know about migration Panel to Card? Again, this is purely my opinion. And this is a crazy idea. One should stick to one course. And the use of helpers - it's cool.

marcoscv-work:
Ok guys! thanks a lot, PR closed! 👍

Issue #22318, opened 2017-03-30
move .dropdown-toggle (caret) to mixin


wojtask9:
Currently .dropdown-toggle is hard to customize.
If it will be mixin then people can override this mixin and create own caret (.dropdown-toggle::after) (for example using font-awesome).

This is only my idea to improve bootstrap :)

Johann-S:
Bug reports must include a live demo of the problem. Per our contributing guidelines, please create a reduced test case via [JS Bin](https://jsbin.com/) or [CodePen](http://codepen.io/) and report back with your link, Bootstrap version, and specific browser and OS details.

wojtask9:
Maybe I wasn't clear. 
This isn't a issue, problem or bug with bootstrap. It's just about moving dropdown-toggler::after to mixin.

My thinking:

move 
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_dropdown.scss#L9
and
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_dropdown.scss#L24

to some mixins

@mixin dropdown-caret() {
  &::after {
    display: inline-block;
    width: 0;
    height: 0;
    margin-left: $caret-width;
    vertical-align: middle;
    content: "";
    border-top: $caret-width solid;
    border-right: $caret-width solid transparent;
    border-left: $caret-width solid transparent;
  }
}

@mixin dropup-caret() {
  &::after {
    border-top: 0;
    border-bottom: $caret-width solid;
  }
}

and then replace referenced lines with

.dropdown-toggle {
  // Generate the caret automatically
  @include dropdown-caret();
}

and 

.dropup {
  .dropdown-toggle {
    @include dropup-caret()
  }
}

Then in my project I'm able to override mixin (create another one and include after bootstrap mixin) and generate my custom style for dropdown-toggler caret (for example using font-awesome glyph instead of current bootstrap based on border triangle)

mdo:
I'll pass on this one for now—not an immediate gain for the project itself and I'm unsure just how much this would change customization of the class.

PR #22926, opened 2017-06-28
Add option to disable carets on dropdowns through `$enable-caret`

pat270:
Disables carets on .dropdown-toggle if `$enable-caret: false;`, is useful for those that want to add their own custom carets on dropdowns.


mdo:
Hmm, I'm a maybe on this one. I'll leave it open for folks to comment on while we get into beta.

afuno:
@mdo In fact, this is a good idea.

pat270:
I was on the fence about this one too since it seemed a bit overkill for one caret. I figured it would be better not to have the rule-sets output if you didn't want the carets. Also the mixin could be used to add carets to random components if you wanted.

wojtask9:
I had the same idea https://github.com/twbs/bootstrap/issues/22318

I wanted caret to be mixin because currently it's hard to customize.

mdo:
Could you add something to the `options.md` docs page, too? We have our global Sass variables listed out there.

pat270:
Added description for `$enable-caret`. Congrats on the Beta release!

mdo (approved):

Issue #21416, opened 2016-12-23
Dropping IE9 support or other older browsers


pattonwebz:
Recent days have shown that maintainers and some developers are in favour of dropping IE9 (and maybe even IE10). Can we please reopen that discussion? I don't think that this decision can be explored fully in the space of 24-48 hours, especially when it's the week before Xmas and many people are not around to weigh in with their opinion.

This issue and PR are both under 48 hours old and both are locked without allowing everyone who is following V4 development the time and opportunity to say what they think.

https://github.com/twbs/bootstrap/issues/21387
https://github.com/twbs/bootstrap/pull/21389

If anyone has an opinion to share, or any data to support their decision, please share so that the decision to drop support can be explored fully to reach the best outcome for developers and end-users.

Thanks in advance for any input you have :)

browner12:
i don't think there's any real interest in dropping support for IE10. The only reason IE9 was targeted was because of flexbox support.  And I think the decision has been made already. A lot of the excess IE9 code has been removed, and we're going all in on flexbox. Basically the answer is if you need support for <=IE9, use BS3.

pattonwebz:
@browner12 yeah it appears that the decision has been made and already a PR is ready that drops most of the support (with the rest forthcoming in the next few days). Personally I also want IE9 to stop being used but people do continue to use it. Stats vary greatly among different user groups and geographical location and the current window of discussion between maintainers and developers hasn't been open long enough to get a proper overview of that situation.

myorangeca:
@browner12  I second your opinion. We closed it too quick and it was not discussed in detail. It looks as if we are skipping the work just to push the release. :\

mdo:
I was unaware my PR to make this change was locked, apologies for that and thanks for opening a new issue to continue discussion.

While the decision was made _publicly_ in a day or two, I've spent the last several weeks (months?) exploring and debating the move to drop IE9 alongside the IE8 drop. I was already convinced of the change before asking on Twitter or opening that PR, but I figured asking couldn't hurt. I think I saw maybe a handful of negative reactions.

At this point, however, I'm sticking with IE10+. I'm incredibly excited to be building a leaner and more forward-thinking v4 with this change (if the number of PRs didn't already show that :P). Bootstrap 3 is now the go-to version for folks who require IE9 (or IE8) support.

Please continue bringing up potential issues, though. I'll happily take inventory of bugs, grievances, and the like so that after v4 is finalized, I can swing back to v3 and make some quality updates there, too.

[...]

PR #22943, opened 2017-06-30
Update browser support

mdo:
I'd like to be more aggressive in our browser support—this is a first pass at it. Looking at our compiled CSS, we've added a ton. :sob: This list has largely been out of date aside from the dropping of IE9/10. I'm thinking about this because of the number of flex properties that require vendor prefixing—it literally triples line of flex code you write. This knocks down some of it.


mdo:
Fwiw, worst case scenario after merging something like this is that we have to support older browsers, which isn't a breaking change. On the flip side, wanting to drop older browsers is a breaking change.

yasinkocak:
Drop older browsers

Johann-S:
We are still in Alpha so breaking changes it's something which can happen, I'm in favor of dropping old browsers, maybe you'll find interesting information here : http://caniuse.com/#feat=flexbox

lorvent:
eventhough i love dropping support for old browsers for obvious reasons,
we develop products and customers with old browsers will cry now.... (i really hope everyone ditches IE and moves to firefox/chrome)

Superpencil:
To summarise caniuse as @Johann-S stated: 12.29% of current active internet users do not have working flexbox support, including the entire Indian subcontinent

Mustafa-Online:
Supporting old browsers is a good thing in some aspects, especially Flex property, because it is the core of the framework now 😃

mdo:
v3 will handle all the older browsers out there if you need to support them. We'll get a v3 quality of life update out after v4 ships. That's always been the plan :).

andrerom:
@mdo Based on both support data and usage data in http://caniuse.com/#feat=flexbox, and considering the two years that has passed since alpha 1. Shouldn't this rather bump to `Explorer >= 11` and `Edge >= 14` for the beta?

Herst:
@andrerom #21416 was when the decision for dropping IE9 but not IE10 was made. @mdo said in that thread that supporting IE10 hasn't been an issue at all.

PR #22947, opened 2017-06-30
Rearrange make-container mixin

mdo:
This moves the `max-width: 100%` from the media query and puts it at the root of the mixin so we're not unnecessarily repeating `max-width` across all breakpoints at compilation


mdo:
No more need for the original purpose of this PR, but there is a linting problem, so fixing that.

PR #22952, opened 2017-06-30
Update carousel.md

wil93:
fix typo


mdo:
There's a `slide` and a `slid` event, not a typo.

PR #22958, opened 2017-07-01
Feature: Add carousel fade option

### LINKED ISSUE #20974: Crossfade transition option for Carousel
Hi,

I apologize if this has been asked before, but Im having trouble finding a simple solution to have an image sequence crossfading transition (like carousel with an image fading effect rather than slide, a slow opacity fade into new image (1-3 second animation time) with 5-8 seconds interval between images.  Tried tons of examples, bootply, custom ones ppl made, none work for me.  I want a smooth, slow crossfade thats enjoyable to watch.  Its odd that bootstrap 3 ( i use 3.3.5) has such an easy copy/paste carousel slide, with no CSS needed, yet something comparable like an image fade carousel doesnt exist without convoluted css and js and  jquery, media queries and tons of configs which do not work, because of absolute positioning tricks etc.  Wish this feature could be added.  

Examples of what Im looking for are here (tried them all- doesnt work with my standard carousel:

Demo 3:  http://css3.bradshawenterprises.com/cfimg/#cfimg3
http://www.1squarepear.com/_docs/bootstrap_carousel_fade.html
https://codepen.io/crashy/pen/JoKMgG

mdo:
Fixes #20974.

Adds `.carousel-fade` for transitioning via `opacity` instead of sliding the classes horizontally. Note that due to how our JS is built, the `.slide` class still must be present. I've also cleaned up some code comments here to help folks (like me haha) when customizing this component.

This is slated for a v4.1 or v4.2 release.


mattonit (commented):
I just tested this. Works great and doesn't break anything

odahcam:
Why you keep working in a carousel that doesn't supports touch? 🤔

mdo:
@odahcam Because I'm working on what I _can_ work on with the carousel. I don't write JavaScript, so I can't solve that problem myself right now. You and anyone else are welcome to take a pass at it though! I'd love to see that issue resolved.

odahcam:
@mdo It's all right, but why not use some already stable and big solution like Swiper, OwlCarousel or some other nice carousel?

mdo:
> It's all right, but why not use some already stable and big solution like Swiper, OwlCarousel or some other nice carousel?

We don't really embed other folks' components into Bootstrap. Only exception has been Popper.js for it's JS positioning—all the HTML and CSS is our own. Not ruling it out, but it's not my favorite.

odahcam:
Sure embedding external libraries would a mess, but I'm not talking about put more code inside BS, but about let other tools do the job for us.

Carousels are very complex components and I think is too hard for a library like Bootstrap, which has no focus in carousels, beat one of the big carousel libraries like Swiper.

I think that there's no need of have all the possible components inside Bootstrap, we already have awesome components that integrate like a charm:
- [Datepicker](https://bootstrap-datepicker.readthedocs.io/en/latest/)
- [Timepicker](http://weareoutman.github.io/clockpicker/)
- [Complex select boxes](https://select2.github.io/select2-bootstrap-theme/)

Like the examples above, I think that libraries like Fancybox and OwlCarousel could have something like an Bootstrap theme. In that way Bootsrap could focus more in the base styling than advanced components (more CSS than JS).

If I'm talking bullshit sorry, but I really wanna know your opinion this kind of things. 😀

smalpierre:
Yo mdo - I have swipe working without any plugins at all. No jquery mobile even.

`    <script type="text/javascript">

        // ENABLE TOUCH SWIPE FOR CAROUSEL - NO PLUGINS NEEDED!
        $(".carousel").on("touchstart", function (event) {
            var xClick = event.originalEvent.touches[0].pageX;
            $(this).one("touchmove", function (event) {
                var xMove = event.originalEvent.touches[0].pageX;
                if (Math.floor(xClick - xMove) > 5) {
                    $(".carousel").carousel('next');
                }
                else if (Math.floor(xClick - xMove) < -5) {
                    $(".carousel").carousel('prev');
                }
            });
            $(".carousel").on("touchend", function () {
                $(this).off("touchmove");
            });
        });
    </script>`

XhmikosR:
@Johann-S: see the above comment for touch support.

PR #22961, opened 2017-07-02
Small reorg to rearrange utilities SCSS and docs

### LINKED ISSUE #19314: Add utility class which adds ellipsis if text overflows
As explained in the below StackOverflow question, one might often want to add an ellipsis to some text if it overflows.

http://stackoverflow.com/questions/802175/truncating-long-strings-with-css-feasible-yet/22811590#22811590

I created a pull request for this feature https://github.com/twbs/bootstrap/pull/19313

It's my first time contributing to Bootstrap so forgive me if I'm doing anything incorrectly.

mdo:
This breaks apart the responsive helpers docs into separate pages for floats and embeds. It also moves the embed SCSS file into the utilities subfolder, where it should be given it's docs organization.

Fixed the docs for `.text-truncate` as well. Fixes #19314.


abrambailey:
How do we apply this within breadcrumbs? The class seems to do nothing.

PR #22963, opened 2017-07-02
Add .font-weight-light class, add !important to the classes

### LINKED ISSUE #20528: Add more font-weight-* utility classes
In #18433 it was proposed to add a number of `font-weight-*`-classes. Only `font-weight-bold` and `font-weight-normal` were added because these "apply rather universally to default web fonts".

However, since then #19098 has been implemented, using the "Native system font stack". Most (if not all) of these fonts do provide a wide array of font weights. I do believe including an utility class for all of them is perhaps a step to far, but I think that at the very least the inclusion of `font-weight-light` has to be reconsidered. As (at least) Segoe UI, San Fransisco and Roboto all support this font weight, making up the vast majority of devices.

mdo:
Adds the class and `!important` to match other utils.

Fixes #20528.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -240,6 +240,7 @@ $font-size-lg:   1.25rem !default;
 $font-size-sm:   .875rem !default;
 $font-size-xs:   .75rem !default;
 
+$font-weight-light: light !default;
 $font-weight-normal: normal !default;
 $font-weight-bold: bold !default;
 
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -240,6 +240,7 @@ $font-size-lg:   1.25rem !default;
 $font-size-sm:   .875rem !default;
 $font-size-xs:   .75rem !default;
 
+$font-weight-light: light !default;
```
mkasperski: There's no such thing as `font-weight: light` - there's `lighter`, but it's relative and I'm not sure it's what you wanted to achieve here. Alternatively it could be `300`, but it would need to be tested with the font stack if it reliably supports this weight (or at least does a fallback to other weight that 'looks light')

mkasperski (approved):

PR #22980, opened 2017-07-03
v4.2: Add new toasts component

### LINKED ISSUE #21769: [New Component] Toast / Notification / Notify / Confirm dialog (like Alert but small and fixed to screen corner)
Couldn't find any similar component currently in BS4.

Notifications ("toasters") are very common part of modern UI which can be used to send small notifications to a user (confirm action, new message received, action with confirmation, etc)

I have own Notify component based on BS4 Alert styles.

What is your opinion on adding a new component or extending Alert with classes like `alert-fixed-top-right`?

mdo:
screen shot 2017-07-03 at 4 08 39 pm

Messed around with this toasts component on account of the extended holiday weekend here. There's no JS for this, and I doubt there will be unless someone like @Johann-S or @bardiharborow want to take a crack at it :). However, what we do have is a super simple and highly customizable component for showing push notification style messages to folks.

Fixes #21769.


patrickhlauke:
Saying right now that toast messages will need to have an `aria-live` attribute to ensure that they're announced to assistive technologies when they're shown / updated https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

Johann-S:
Toast, Spinner... You're on fire @mdo 👍 
I totally like it !

Yes why not add some JS for example, to set Toast placement or to allow folks to set when the toast should be dismissed

mdo:
> Yes why not add some JS for example, to set Toast placement or to allow folks to set when the toast should be dismissed

If you have some vanilla JS to include in the docs, I'll be happy to see it added! :D

odahcam:
Guys, I have something here: https://codepen.io/odahcam/pen/oeoYxm
Also:  https://github.com/odahcam/bootoast

It was designed for BS3, but it works on BS4 and I'm working on a version for 100% over BS4. 😀

jipexu:
this is so cool ..

mdo:
@patrickhlauke I've resolved conflicts and added a quick section to stub something out about accessibility for live regions. I'm not 100% on how they work or what's required though—can you take a pass at it before we merge this?

ysds:
Do you have any thoughts about z-index of this toast component?

ysds:
@mdo Thanks for your response. Are you planning on adding z-index? If not, I think it should be documented.

mdo:
I'll jump on it after we confirm a11y todos.

@odahcam See the variables file for our current z-index stack:

https://github.com/twbs/bootstrap/blob/96a9e98e49cdd32148981ef71617b1cfc6054470/scss/_variables.scss#L560-L571

Tooltips/popovers are currently the highest priority, above modals and navbars, because any element within a modal or navbar can tigger a tooltip/popover. Toast could arguably have tooltips/popovers within them, but that doesn't seem as likely. I'm thinking they might go above? Will need to experiment a bit more, but they cannot go below modals and dropdowns.

odahcam:
sorry @mdo, I did not had time to take a deep look at the way this kinda things work. 

So if everything is fixed, when I have an open popover and I open some modal, wont the popover persists over the modal? If it doesn't and the same logic works for the toasts, I think toasts could go between the modal and the popover, so users would be free to use tooltips and popovers as they want from the toast.

I just think that in no way the toast can have a higher `z-index` than the tooltip, because as long as toasts usually shows some short timing information, toasts could be well used to show extra timing info. Also, popovers could be used to confirm some toast's action.

patrickhlauke:
would be nice to see this in action with a super-basic dynamic example. looking at this "cold", the live region advice seems appropriate (noted a few tweaks). sadly, live regions are quite fickle when it comes to the precise timing...generally, you'll want to keep a container with `aria-live="..."` "primed" in the document at first, and then inject things (like toasts) into it...generating both at the same time generally confuses browsers/assistive tech, as they don't have enough time to work out/watch the new live region for changes.

bastienmoulia:
@mdo: Can you add a close button in an example with the CSS to position it correctly?
```html
<div class="toast-header">
    <img class="rounded mr-2" data-src="holder.js/20x20?size=1&text=.&bg=#007aff" alt="">
    <strong class="mr-auto">Bootstrap</strong>
    <small class="text-muted">11 mins ago</small>
    <button type="button" class="close" data-dismiss="toast" aria-label="Close" style="">
      <span aria-hidden="true">&times;</span>
    </button>
  </div>
```
```scss
.toast-header .close {
  padding: 0 0.25rem;
  margin: -0.25rem -0.75rem 0 0;
}
```
capture d ecran 2018-03-02 a 22 40 58

bastienmoulia:
I made some changes with a fork of this to add theme color, JS example and z-index. Should I do a new PR ?
https://github.com/bastienmoulia/bootstrap/commit/a94aeac5a1e0007c8afd1012097ecc8f26a9c7b7

GuillaumeWrobel:
@mdo can we help somehow? Thats a great component!

Johann-S:
Hi @mdo, I updated your branch because I work on a small plugins for the Toastr

Johann-S:
I have to do the unit test and update the documentation 👍

scss/_toasts.scss
```diff
@@ -0,0 +1,34 @@
...
+  display: none;
+  max-width: $toast-max-width;
+  overflow: hidden; // cheap rounded corners on nested items
+  font-size: $toast-font-size; // knock it down to 14px
+  background-color: $toast-background-color;
+  background-clip: padding-box;
+  border: $toast-border-width solid $toast-border-color;
+  border-radius: $toast-border-radius;
+  box-shadow: $toast-box-shadow;
+  backdrop-filter: blur(10px);
```
XhmikosR: I'm not sure I'd go with this. I mean it's not even in WD phase :/

XhmikosR:
@Johann-S: that's a lot of code to review :) Thanks for adding tests BTW!

That being said, only @mdo can tell us if he thinks he's done with it or not.

Johann-S:
A simple demo of our Toast plugin : https://codepen.io/Johann-S/full/QVjPrB/ 😉

patrickhlauke:
@Johann-S ... needs `aria-live` and `aria-atomic` to be actually accessible/announced by AT https://codepen.io/anon/pen/xawowW

Johann-S:
That's something we should add in our docs @patrickhlauke because we don't manage the container of our toasts. Can you handle that ?

patrickhlauke:
there's some basic info here in the proposed docs addition already about `aria-live`, but i'll review and expand it (i've since worked a bit more on MDN stuff relating to live regions, so may even be good to point to that...) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

Lausselloic:
some improvement proposed on the toast branch : https://github.com/twbs/bootstrap/pull/27155

Feel free to comment, improve. Could update it if needed

Johann-S:
@mdo if think everything is done here, so we'll wait for your final decision 👍

andresgalante:
@mdo should we float this feature back?

mdo:
@andresgalante Not sure what you mean by float it back?

---

I just pushed some updates to clear up some docs, add dismiss buttons to every example, and move from `display: inherit` to `display: block`. The latter change has to be here; `inherit` is super messy if someone has a parent with any `display` value other than `block` or `inline-block`.

Johann-S:
Not sure if it's a good things to add close button in every Toast because that's not something required, Toast can work without that close button

XhmikosR:
Also, when I tried this branch yesterday, a lot of text was not readable most because of the dark background in docs.

Johann-S:
@XhmikosR I fixed that yesterday 😉

XhmikosR:
This LGTM. Only thing I'd maybe change is remove holder.js and use something like `<span class="rounded mr-2 d-block bg-primary" style="width: 20px; height: 20px;" aria-hidden="true"></span>`.

Not sure if we need `aria-hidden` or not there...

XhmikosR:
@patrickhlauke can you comment on my comment [above](https://github.com/twbs/bootstrap/pull/22980#issuecomment-429545472)?

I think it's the only thing left for this to land. We could land it with holder.js too but I find a pure CSS solution cleaner personally.

patrickhlauke (commented):
> This LGTM. Only thing I'd maybe change is remove holder.js and use something like:
> 
> ```
> <span class="rounded mr-2 d-block bg-primary" style="width: 20px; height: 20px;" aria-hidden="true"></span>
> ```
> 
> Not sure if we need `aria-hidden` or not there, though.

if there's nothing there, you don't need to hide it no

XhmikosR:
@patrickhlauke: thanks for the reply, I'm gonna switch to a pure CSS solution for the images here.

Can you make the text changes you commented above? I will clean the branch before the final merge.

XhmikosR:
@patrickhlauke: I added your suggested text changes. If you could add the remaining aria-related suggestions we could merge this soon after a cleanup.

mdo:
Some some docs changes, but I went and updated the aria roles. May have done too many of those? @patrickhlauke's comments mentioned polite, but the rest were `assertive`—which is it?

XhmikosR (approved):

PR #22984, opened 2017-07-04
fixed doc vertical-align.md "inlie" to "inline"

Wcc723:
fixed a word.


Johann-S:
Good catch thank you 👍

PR #22986, opened 2017-07-04
overflow-y auto on toc menu

### LINKED ISSUE #22977: Overflowing TOC hidden  due to large amount of content and position:sticky
When looking at the latest docs for v4 Layout -> Grid 
The TOC on the right overflows on smaller screens. The bottom links are hidden. 
That element is `position:sticky` so scrolling down the page doesn't help. 
I'm sure this is happening elsewhere. 

Shall we add `overflow-y:scroll;` to .bd-toc ? 

Double scrollbars suck but I'm not sure on a better solution ? 

- Browser and version: Chrome, Firefox
image

gijsbotje:
fixes #22977

### DIFF
--- a/assets/scss/_sidebar.scss
+++ b/assets/scss/_sidebar.scss
@@ -9,12 +9,11 @@
     position: sticky;
     top: 4rem;
     max-height: calc(100vh - 4rem);
+    overflow-y: auto;
   }
   order: 2;
   padding-top: 1.5rem;
   padding-bottom: 1.5rem;
-  overflow-x: visible;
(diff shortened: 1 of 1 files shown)


mdo:
Thanks! I've moved the `overflow-y` to the `@supports` feature query, that way it only applies if the column is truly "stuck" and needing the scrollbar. Also removed the unnecessary `overflow-x` I added some time ago.

gijsbotje:
aaah now i get that '@supports', smart!
I was wondering if that overflow-x was needed, looks good to me

PR #22991, opened 2017-07-04
fixes #22976

### LINKED ISSUE #22976: Remove position: static from dropdowns in navbar
Is there a reason to use `position: static;` for dropdowns in navbar? I use code from `v4-dev` and it works fine, but it causes a problem in your new website

The problem isn't visible in `v4-dev` because Popper set the position of the dropdown to absolute (as it should be). Previously my PR (#22010) was rejected because you said that was intentional to use `position: static`, but when currently it's fixed by popper.js

Here's what I'm talking about:
image

And without `position: static;`
image

Isn't that better?

Tested on latest Chrome, Win 7/Win 10/Ubuntu 17.04/OS X Sierra 10.12.5

mattonit:


Johann-S:
That's not a good fix see my comment : https://github.com/twbs/bootstrap/issues/22976#issuecomment-312596241

mdo:
Superseded by #23014.

Issue #22090, opened 2017-03-01
Add Slider and toggle components


natcohen:
I was wondering since bootstrap has been built with "mobile first" in mind, why there is no slider and toggle (which are key components in the mobile world)!

Shouldn't be new features to add to the new v4?

HarelM:
Should be one of the top priorities, it's so basic.

saad749:
+1 its an essential component for a lot of websites

shyamal890:
Angular Material too has these components. Its a basic need in mobile.

HarelM:
I migrated to material for several reasons: slider, RTL and pagination support out of the box, I don't understand how this is not implemented already in bootstrap, especially the RTL - this is a requirement as old as bootstrap...

shyamal890:
@HarelM The only problem I have with material is that it is too style heavy. I guess we would need more designers to make sure that custom elements gel well with material elements.

natcohen:
@HarelM Don't forget that nobody at bootstrap is paid for this project (AFAIK) while material is backed by Google... so yeah, it's a must have but be patient!

HarelM:
True, I can't argue about it, although I don't think it's a matter of payment but priorities - bootstrap 4 has a lot of code changes but still doesn't have RTL and slider - which means it's not highly prioritized, it's not like this project doesn't have contributors and maintainers...

cinqlair:
Totally agree, I clearly miss sliders and toggle components. 
If I had the skills, I would do it !

natcohen:
@mdo Any chance this can be implemented in the next release?

gijsbotje:
slider has allready been added, the switch is still in progress i guess. although i couldn't find it on any project board

guledali:
Exactly slider has already being added. The toggle-switch, seems that they are reviewing it. 
Man I really hope it becomes part of this release.

mdo:
Custom switch was added for v4.2 (launching soon), and the custom range was added in v4.1. Closing out!

septatrix:
Is the "Feature Idea" project still used? If so this issue should be removed from there to keep it clean and up to date

PR #23004, opened 2017-07-06
v4.2: switch custom form control

gijsbotje:
This adds a neat styling feature for the custom form controls.
Markup stays the same as the checkbox but displays it as a toggle switch.

schermafbeelding 2017-07-06 om 11 25 55

https://codepen.io/gijsbotje/pen/rwKMmy

https://deploy-preview-23004--twbs-bootstrap4.netlify.com/docs/4.1/components/forms/#switches


docs/4.0/components/forms.md
```diff
@@ -1004,6 +1004,23 @@ Custom checkboxes and radios are inline to start. Add a parent with class `.cust
 </div>
 {% endexample %}
 
+### switches
```
skmedix: First letter should be capitalized

skmedix (changes requested):

docs/4.0/components/forms.md
gijsbotje: fixed

gijsbotje:
fixes #22090

mdo:
Dope! Thank you for jumping in on this. I'll take a look at it for our v4.1 ship :).

gijsbotje:
@mdo @XhmikosR @andresgalante I messed up here :( and cant figure out how to revert it, any suggestions?
Sorry for this... don't understand what wen't wrong

XhmikosR:
@gijsbotje: you can just do `git rebase -i upstream/v4-dev` and clean up your local branch.

gijsbotje:
maybe i did something wrong, but it didn't work...

XhmikosR:
@gijsbotje: you had messed up your branch. Do not merge from upstream. Just rebase it, if needed.

Now, I cleaned it up and force pushed it. So, make sure you do `git fetch origin && git reset --hard origin/custom-switch-form-control` before anything else otherwise you will overwrite the cleanup.

gijsbotje:
@XhmikosR thanks dude, will do
i tried to make a pr for a different issue and my forked v4-dev branch is messed up i think
i'll see if i can fix it all

docs/4.0/components/forms.md
```diff
@@ -1013,6 +1013,32 @@ Custom checkboxes and radios can also be disabled. Add the `disabled` boolean at
...
+{% endexample %}
+
+#### Animated switch
+
+Optionally you can add the class `.custom-switch-animated` to add a transition to the indicator.
+
+{% example html %}
+<div class="custom-control custom-switch custom-switch-animated">
+  <input type="checkbox" class="custom-control-input" id="customSwitchAnimated">
+  <label class="custom-control-label" for="customSwitchAnimated">toggle this switch element</label>
```
patrickhlauke: Probably want to uppercase the "Toggle" text

gijsbotje: fixed :)

m5o:
@gijsbotje / @mdo the solution to modifying the custom-checkbox with `.custom-checkbox-switch` is super handy. 
http://wingman.mediumra.re/components-bootstrap.html#forms

docs/4.0/components/forms.md
```diff
@@ -1118,6 +1118,32 @@ Custom checkboxes and radios can also be disabled. Add the `disabled` boolean at
...
+  <label class="custom-control-label" for="customSwitch2">Disabled switch element</label>
+</div>
+{% endexample %}
+
+#### Animated switch
+
+Optionally you can add the class `.custom-switch-animated` to add a transition to the indicator.
+
+{% example html %}
+<div class="custom-control custom-switch custom-switch-animated">
```
m5o: `.custom-switch-animated` is not defined.

gijsbotje: have a look at https://github.com/twbs/bootstrap/pull/23004/files#diff-e5540857a54676b43146ad92add2dbdeR182
it's actually is defined as a nested sass selector

wolfy1339: They don't like that kind of stuff. Don't use nesting to create new classes

ayoubkhan558-zz (approved):
I think it should be added to bootstrap, so that we don't need any external library for this purpose.

ayoubkhan558-zz (changes requested):
Its a good option to have

claudioalmeiida:
"mdo removed this from Inbox in v4.2 13 days ago" :(

guledali:
@mdo  @XhmikosR Any chance we could see the toggle-switch in this release?

scss/_custom-forms.scss
```diff
@@ -161,6 +161,39 @@
...
+      background-size: $custom-control-indicator-size;
+    }
+  }
+
+  .custom-control-input:checked ~ .custom-control-label::after {
+    background-position: right center;
+  }
+
+  &-animated .custom-control-label::after {
+    transition: background-position .1s;
```
MartijnCuppens: I think we don't need a animated and non animated version, you may use the transition in the default version. Also, use the `transition` mixin for this.

XhmikosR:
Personally I'm not sure it's ready yet.

scss/_variables.scss
```diff
@@ -542,6 +542,10 @@ $custom-checkbox-indicator-indeterminate-border-color: $custom-checkbox-indicato
 $custom-radio-indicator-border-radius:          50% !default;
 $custom-radio-indicator-icon-checked:           str-replace(url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='-4 -4 8 8'%3e%3ccircle r='3' fill='#{$custom-control-indicator-checked-color}'/%3e%3c/svg%3e"), "#", "%23") !default;
 
+$custom-switch-indicator-border-radius:         $custom-control-indicator-size / 2 !default;
+$custom-switch-indicator-icon:                  str-replace(url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-4 -4 8 8'%3E%3Ccircle r='3' fill='#{$custom-control-indicator-checked-color}'/%3E%3C/svg%3E"), "#", "%23") !default;
```
MartijnCuppens: It's possible to do this without the base64 icon, just use `border-radius` for the element

scss/_custom-forms.scss
```diff
@@ -161,6 +161,39 @@
...
+      left: -$custom-control-gutter * 1.5;
+      width: $custom-control-indicator-size * 1.75;
+      background-image: $custom-switch-indicator-icon;
+      background-position: left center;
+      background-size: $custom-control-indicator-size;
+    }
+  }
+
+  .custom-control-input:checked ~ .custom-control-label::after {
+    background-position: right center;
```
MartijnCuppens: Don't use background-position for this, use the transform property, this is more performant.

MartijnCuppens (dismissed):

XhmikosR:
@gijsbotje: will you be able to address @MartijnCuppens's comments above?

gijsbotje:
sorry got pre-occupied with work. i'll try and fix it before the weekend

gijsbotje:
i'm struggling a little bit...
as the colour scheme changed, the unchecked state of the switch now doesn't have a gray background. which results into the indicator being not visible until the switch is checked.

to illustrate my issue:
image

@MartijnCuppens @XhmikosR any suggestions on what colors to use for what?

XhmikosR:
Yeah, that was the issue I noticed too when I said I didn't think this was ready along with transitioning costly properties.

I'll try to have a look tomorrow, but @MartijnCuppens will probably beat me to it.

gijsbotje:
would this be ok? i used the color of the border for the indicator
image

rafalp:
Can you include screenshot of other special control (eg. checkbox) bein on/off/disabled?

Also, is there "disabled + on" state?

gijsbotje:
@rafalp i've paired the switch with its checkbox equivalent
image

thanks for reminding me about the `:checked:disabled` state, forgot about that

gijsbotje:
@MartijnCuppens fixed your comments as best as I could think of.
main question to resolve is if the color for the circle when it's unchecked is ok

rafalp:
Those are looking very good, thanks 👍

scss/_custom-forms.scss
```diff
@@ -161,6 +161,48 @@
...
+    }
+
+    &::after {
+      top: calc(#{(($font-size-base * $line-height-base - $custom-control-indicator-size) / 2)} + #{$custom-control-indicator-border-width * 2});
+      left: calc(#{-$custom-control-gutter * 1.5} + #{$custom-control-indicator-border-width * 2});
+      width: $custom-switch-indicator-size;
+      height: $custom-switch-indicator-size;
+      background-color: $custom-control-indicator-border-color;
+      background-size: $custom-control-indicator-size;
+      border-radius: $custom-switch-indicator-border-radius;
```
MartijnCuppens: This is always `50%` I think?

scss/_custom-forms.scss
```diff
@@ -161,6 +161,48 @@
...
+    }
+
+    &::after {
+      top: calc(#{(($font-size-base * $line-height-base - $custom-control-indicator-size) / 2)} + #{$custom-control-indicator-border-width * 2});
+      left: calc(#{-$custom-control-gutter * 1.5} + #{$custom-control-indicator-border-width * 2});
+      width: $custom-switch-indicator-size;
+      height: $custom-switch-indicator-size;
+      background-color: $custom-control-indicator-border-color;
+      background-size: $custom-control-indicator-size;
```
MartijnCuppens: `background-size` is only relevant for background images which we don't have.

MartijnCuppens (dismissed):

MartijnCuppens:
@gijsbotje, the color of the circle looks ok.
IMO the switch looks a bit small, maybe we should try to increase the size a bit? 
It would also be nice to have total control over the track width and height and the spacing between the switch and the label text. With 'the track', I mean this element:
image

Maybe it's possible to tweak this a bit to so that it's possible to set the following variables:
- The track width
- The track height
- Margin between the switch and the label text
- The size of the circle

If we have control over this in `_variables.scss`, it would be easier to control the look of the switch and it would also be possible to make them look like in this PR:
https://github.com/twbs/bootstrap/pull/26691

scss/_custom-forms.scss
MartijnCuppens: Forget this, you actually also took care of the ability to use switches that aren't rounded:
image

Sweet!

rafalp:
I would be careful with making switch too big - you'll potentially end with component being too big to use "inline", requiring special positioning or, worse, custom form layout altogether.

MartijnCuppens:
Good point, @rafalp. Forget about size increment than.

rafalp:
Yeah, I am thinking now how this component should be sized. Right now its sized same way checkbox is, meaning its interchangeable with it. Perhaps it should also support `.form-control-lg` and `.form-control-sm` for extra variety?

Just dropping ideas - it may be too late for this, or material for new PR bumping extra sizing to custom controls?

gijsbotje:
> 💡 Support for contextual classes could be interesting, too.

I would do that as a seperate PR as it also affects the other controls

 

> Yeah, I am thinking now how this component should be sized. Right now its sized same way checkbox is, meaning its interchangeable with it. Perhaps it should also support `.form-control-lg` and `.form-control-sm` for extra variety?
> 
> Just dropping ideas - it may be too late for this, or material for new PR bumping extra sizing to custom controls?

Indeed as you said yourself `material for new PR` 😄 
 

> Maybe it's possible to tweak this a bit to so that it's possible to set the following variables:
> 
> - The track width
> - The track height
> - Margin between the switch and the label text
> - The size of the circle
> If we have control over this in _variables.scss, it would be easier to control the look of the switch and it would also be possible to make them look like in this PR:
> #26691 

I was fiddling around with the variable `custom-control-gutter`, but this isn't really the gutter, right? It doesn't define the space between the indicator and the label, but rather defines the space that the label is pushed over. Then the indicator is pulled back the same amount.
So by default the gutter is set to 1.5rem. This gives the checkbox and radio a breathing room of .5rem. Therefore shouldn't gutter be .5rem? And the left properties set to `$custom-control-indicator-size + $custom-control-gutter` which would result in 1.5rem?

If we implement this change, controlling `the track width` and `space between switch and the label` would be much easier. to get it to be accurate in the current setup i would need to do something like:
`$custom-switch-width + ($custom-control-gutter - $custom-control-indicator-size)`
So if you guys agree with the change to `$custom-control-gutter` I will make a separate PR with those changes. I can imagine this would have some impact on people that make use of the sass, so that would be your call.

What i'm really struggling with and get wrap my head around is making the height and circle size customizable. I'm currently using the calc css function to get the height and with like this:
`calc(#{$custom-control-indicator-size} - #{$custom-control-indicator-border-width * 4})`
First the 1rem indicator-size then subtracting the border width * 4 to create space between the circle and the track. This can not be done in sass math as `$custom-control-indicator-border-width` is a pixel value and `$custom-control-indicator-size` is a rem value, which are incompatible in sass.

i'm also struggling with the positioning of the circle when you increase or decrease the size of the circle. currently this calc function is taking care of the vertical positioning:
`calc(#{(($font-size-base * $line-height-base - $custom-control-indicator-size) / 2)} + #{$custom-control-indicator-border-width * 2})`

I'm far from a math guy so if so if someone could assist me in these issues that would be much appreciated.

MartijnCuppens (approved):
LGTM, more control over sizing is something we can look into later

PR #23010, opened 2017-07-06
Fixing docs example in input-group section

vleput:
One of the examples in the Input Group -> Button Addons section was broken because of a missing quote.
Before:
image

After:
image


patrickhlauke:
Good catch, thanks

PR #23011, opened 2017-07-07
dadayi.txt

laijie1994:
this is my firsy commit

PR #23012, opened 2017-07-07
Colors theme

Phantomphyre:
[penetration-testing-with-kali.pdf](https://github.com/twbs/bootstrap/files/1130195/penetration-testing-with-kali.pdf)


scss/utilities/_text.scss
```diff
@@ -34,25 +34,21 @@
```
Phantomphyre: #23011

PR #23019, opened 2017-07-08
Modal - prevent bubbling of [esc] when closing

itgoldman:
This is also fixed in version 4.


itgoldman (commented):
Please accept this fix? It was also fixed in version 4.

Johann-S:

PR #23026, opened 2017-07-09
fix navbar class

ghost:
navbar-default like thumbnail at https://v4-alpha.getbootstrap.com/examples/screenshots/sticky-footer-navbar.jpg


mdo:
Navbars are now `.navbar-light` and `.navbar-dark`.

PR #22804, opened 2017-06-12
Generate series of `.table-responsive-*` classes to accomodate overflowing tables in a variety of screen sizes

pat270:
The old `.table-responsive` was nice because it helped manage unwieldy tables. This adds responsive tables at all screen sizes and at each breakpoint, `.table-responsive`, `.table-responsive-sm`, `.table-responsive-md`, `.table-responsive-lg`, and `.table-responsive-xl` respectively.

PR #23028, opened 2017-07-09
Fix .table-responsive breakpoint

PerfectPixel:
This fixes the threshold for .table-responsive to perform its magic. `media-breakpoint-down(md)` is equal to <991px, while `media-breakpoint-down(sm)` is the correct breakpoint according to the comment.

While this is superseeded by #22804 , you might want to merge this to have properly functioning responsive tables in the meantime.


mdo:
Merged #22804, thanks!

PR #23041, opened 2017-07-10
Horizontal card support

### LINKED ISSUE #20794: Add support for horizontal card
I think it's very common use case that you want to have card content vertically in small screen and in bigger screens content is aligned horizontally. For example: http://www.apple.com/newsroom/

example

I implemented this for my own project roughly so that:

```
.card {    display: flex; flex-direction: column;}
.card-img {    order: -1;}
.card-block {    order: 0;}
.card-img-last {    order: 1;}
```

.card-img-last is basically same than .card-img-bottom, but it doesn't care about order in source and also naming works better with horizontal cards where last refers to the right side.

Then I have horizontal classes for cards which needs to be added in addition to basic card class:

```
.card-horizontal {    flex-direction: row;}
.card-horizontal-sm-up { ... }
.card-horizontal-md-up { ... } etc.

.card-horizontal .card-img {    flex: 0 0 $card-img-width;}
.card-horizontal .card-block {    flex: 0 0 calc(100% - $card-img-width)}
```

Something like that, not a perfect example but gives the basic idea how it is working. I'm using card-block now as a flex item but it won't work if I want to have card header or footer. There should be some wrapper div which contains all content expect image.

Only problem with this is to make image behave well in horizontal view. In my implementation .card-img is just a div container, then img is inside that and I use object-fit (IE/Edge fix with Javascript) to cover the whole div. I've set minimum and maximum height for .card-block to have always reasonably sized image there.

I would like to hear if this is wanted feature and then opinions what would be the correct implementation. Obviously we need to have also non-flex-box solution for that.

gijsbotje:
This PR is a first take at supporting horizontal cards
This only supports image caps on the left and right side of a `.card-body`

might need more work to support other card components

fixes #20794


kevin-jj:
It would be great to have this PR merged. @mdo

gijsbotje:
sorry got some stuff in this pr that shouldn't be here i'll fix that

scss/_card.scss
```diff
@@ -131,6 +131,21 @@
...
+
+.card-img-left {
+  height: 100%; // Required because we use flexbox and this inherently applies align-self: stretch
+  @include border-left-radius($card-inner-border-radius);
+}
+
+
+// Horizontal cards
+.card-horizontal {
+  flex-direction: row;
```
andresgalante: Hi @gijsbotje This could have been a clever and simple solution, but you have to remember that body and image are not the only elements a card can have. For example, if you add a `card-footer` it will go to the right and it'll look really bad.
Although it had potential I don't think it's general and flexible enough.

osc2nuke: @andresgalante it is exact my simple mind came across.

osc2nuke: Not elegant, and NOT how it is supposed to.
Given from the Template example, using inside the card columns made the `card-footer`
at least working:
        `<div class="col-md-6">
			<div class="card mb-4 box-shadow ">
				<div class="row">
					<div class="col-md-8">
						<div class="card-body flex-column align-items-start">
							<strong class="d-inline-block mb-2 text-primary">World</strong>
							<h3 class="mb-0">
							<a class="text-dark" href="#">Featured post</a>
							</h3>
							<div class="mb-1 text-muted">Nov 12</div>
							<p class="card-text mb-auto">This is a wider card with supporting text below as a natural lead-in to additional content.</p>
							<a href="#">Continue reading</a>
						</div>
					</div>
					<div class="col-md-4">            
						Card image cap
					</div> 			
				</div>
				<div class="card-footer">test</div>			
			</div>
        </div>`

gijsbotje:
closing this as you can actually do this with utility classes, although still far from perfect

an example of how to do it with utility classes 
https://codepen.io/gijsbotje/pen/ePQRoQ?editors=1000

MartijnCuppens:
In the example you provide, there is a small issue with the border radius:
image

But you can add the `.overflow-hidden` to the card and remove `.rounded-left` from the image to fix this.

I think closing this was a good idea, these horizontal cards have responsive issues and that can be solved with the responsive utility classes we provide.

tarkhov:
Worked solution for horizontal cards https://tarkhov.github.io/postboot/#card-image

Issue #21771, opened 2017-01-18
First-letter prints incorrectly in Chrome


zam6ak:
When using [display headings](https://v4-alpha.getbootstrap.com/content/typography/#display-headings) and trying to print (Ctrl+P) the first letter of the heading breaks layout.
Tested on Chrome v55

 
Here is sample HTML code
```
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <title>Print Test</title>

    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css">
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/js/bootstrap.min.js">
</head>

<body>

    <div class="container-fluid">
        
        <div class="row">
            <div class="col">
                <h1 class="display-4">This heading fails to print</h1>
                <p class="lead">This heading fails to print correctly.</p>
            </div>
        </div>

        <div class="row">
            <div class="col">
                <h1 class="display-1">Display 1</h1>
                <h1 class="display-2">Display 2</h1>
                <h1 class="display-3">Display 3</h1>
                <h1 class="display-4">Display 4</h1>
            </div>
        </div>

        <div class="row">
            <div class="col">
                <h1 id="h1-heading">h1 heading</h1>
                <p>Curabitur blandit tempus porttitor. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>

                <h2 id="h2-heading">h2 heading</h2>
                <p>Curabitur blandit tempus porttitor. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>

                <h3 id="h3-heading">h3 heading</h3>
                <p>Curabitur blandit tempus porttitor. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>

                <h4 id="h4-heading">h4 heading</h4>
                <p>Curabitur blandit tempus porttitor. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>

                <h5 id="h5-heading">h5 heading</h5>
                <p>Curabitur blandit tempus porttitor. Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis vestibulum.</p>

                <h6 id="h6-heading">h6 heading</h6>
[...]

PR #23044, opened 2017-07-11
Remove ::first-letter and ::first-line selectors from print styles

patrickhlauke:
Bootstrap itself doesn't use any `::first-letter` or `::first-line`
styles. These selectors also cause(d) problems in IE11 (see
https://github.com/h5bp/html5-boilerplate/pull/1799) and currently cause
a nasty rendering bug in Chrome where the first letter is vertically
shifted (see https://github.com/twbs/bootstrap/issues/21771)

Closes https://github.com/twbs/bootstrap/issues/21771

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -12,15 +12,20 @@
   @media print {
     *,
     *::before,
-    *::after,
-    p::first-letter,
-    div::first-letter,
-    blockquote::first-letter,
-    li::first-letter,
-    p::first-line,
-    div::first-line,
-    blockquote::first-line,
-    li::first-line {
+    *::after {
+      // Bootstrap specific; comment the following selectors out
+      // as we don't set any shadow on first line/letter in Bootstrap
+      // and these selectors trigger a nasty rendering bug in current Chrome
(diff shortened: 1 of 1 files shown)


patrickhlauke:
I know this doesn't address the fundamental problem (which is due to a Chrome bug), but rather circumvents it...nonetheless, as it's not really a problem we actually have (since we don't set any `::first-letter` / `::first-line` text or box shadows, I'd say it's ok to leave that battle to be fought by somebody else / let Chrome fix their bug

patrickhlauke:
h5bp have now removed those styles as well (after i filed an issue), so i might be bold enough to merge this  without bothering @mdo for a review (as it was mainly to know if it was cool to comment those things out/diverge from h5bp)

mdo:
This issue (and many others) was slated for Beta 2 to limit the remaining changes ahead of Beta 1. It's fine that it merged, but I'm trying to get us to do a better job of organizing bugs and new features into more coherent and regularly shippable releases :).

I've updated the issue this fixed to reflect it's status in the Beta 1 project.

patrickhlauke:
@mdo oops sorry, missed that we're now using the "projects" thing

PR #23051, opened 2017-07-12
Fixed a word (connectinh / connecting)

Wcc723:


Johann-S:
Good catch thank you 👍

Issue #22065, opened 2017-02-24
v4 custom-select does not support "multiple"


jsmad:
The current bootstrap v4 beta 6 comes with a new custom-select class, but it appears broken when the "multiple" attribute is also defined.

For example:
```<select multiple class="form-control" id="exampleSelect2">```

changes into
```<select multiple class="custom-select" id="exampleSelect2">```

Maybe the documentation could mention that multiple is not supported, or if it is not too much work, to implement custom-select a bit better for displaying on a multiple select.

Thank you.

C-Lodder:
PR: https://github.com/twbs/bootstrap/pull/23055

PR #23055, opened 2017-07-12
Fix custom-select with "multiple" attribute

C-Lodder:
This PR fixes the issue mentioned here: https://github.com/twbs/bootstrap/issues/22065

`height: auto;` is used to override `height: calc(2.25rem + 2px);`

and

`background-image: none;` is used to override the SVG used on the `.custom-select` element


scss/_custom-forms.scss
```diff
@@ -164,6 +164,11 @@
     }
   }
 
+  &[multiple] {
+    height: auto;
+    background: none;
```
pat270: Changing this to `background-image` would be helpful so we don't lose the `background-color` set above.

C-Lodder:
@pat270 - updated

mdo (approved):

PR #23060, opened 2017-07-12
Add badge counter example and accessibility advice

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/18544


patrickhlauke:
capture

Issue #22091, opened 2017-03-01
Add utility class for overflow property


dkrutsko:
Apologies if this was already brought up but I didn't find any mention of it. I've come across use cases where you might want to make a div `overflow: hidden` for improved design. I would like to propose a new class (or set of classes) relating to controlling the overflow CSS property. Perhaps something like `.overflow-{visible|hidden|scroll|auto}`. These classes do not have to be responsive.

prateekgoel:
I would like to take this up.

mdo:
Closing per comment in #23085.

PR #23085, opened 2017-07-14
Overflow Utitliy class

chiraggmodi:
Have added Overflow utility class #22091


mdo:
I don't see us adding utilities for all of these values, and then getting the inevitable request of making them responsive. For now, passing on this.

darkguy2008:
1 year has passed, I think this functionality would still be useful. Come on, how long are we going to wait?

mdo:
@darkguy2008 Code existing in a PR is not reason enough to merge it into our codebase. We have plenty of code lying around in PRs that will never see the light of day.

As for this, please search around—there are additional issues and PRs that are tackling it. See #26479 for example.

PR #23091, opened 2017-07-15
Start testing on Node v8.

bardiharborow:
Bootstrap informally supports **the last two LTS releases and the current release** of Node. v7 is in maintenance-only mode and v8 (the current release) will begin LTS in October 2017, so this PR moves testing from v7 to v8. v4 is in maintenance-only mode, but is currently being retained for legacy support (as one of the last two LTS releases). Bootstrap's v4 support will likely be deprecated in October 2017. See https://github.com/nodejs/LTS for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ git:
   depth: 3
 node_js:
   - "4"
-  - "7"
+  - "8"
 before_install:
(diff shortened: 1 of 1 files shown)


Johann-S:
That's a good things thank you @bardiharborow 👍

PR #22744, opened 2017-06-03
Update Bundle and NPM Dependencies

CDAGaming:
Listen 3.1.2+ Incompatible at this time, not updated

Signed-off-by: CDAGaming <cstack2011@yahoo.com>

PR #23094, opened 2017-07-15
Update dependencies and switch to using NPM lockfiles.

bardiharborow:
* Jekyll has been held back at v3.4.5 due to jekyll/jekyll-redirect-from#159.
* The ESLint `indent` rule has set to v3 compatibility mode, pending other fixes.
* Change from using shrinkwrap to [NPM lockfiles](https://docs.npmjs.com/files/package-locks) (similar to how `Gemfile.lock` functions).

I'm happy to expand on my rationale for NPM lockfiles if needed, but basically they don't ship when publishing to NPM (and hence don't need to be moved to a subdirectory like shrinkwraps), and integrate very nicely with `npm update`.

Closes #22744.


Johann-S:
Overlapping with : https://github.com/twbs/bootstrap/pull/22744

XhmikosR:
@bardiharborow: all is good, just one question. Why did you remove anchor etc? Aren't those used?

XhmikosR:
Also, I wanted to point out we should specify clean-css's compatibility for IE to be safe.

bardiharborow:
@XhmikosR I didn't (?). I've just updated anchor from`v3.2.2` to `v4.0.0`, etc.

Is there something with upgrading `clean-css` that I've missed?

XhmikosR:
You are right, it's just that someone had switched to the non-minified versions and now it's again the minified versions. It beats the point though because of the number of lines changed but anyway, maybe it won't happen again in the future.

Re clean-css: it's not exactly related to this PR; I just noticed that with the switch to npm scripts someone missed to specify the minimum IE compatibility option. This should be fixed asap to ensure we don't hit any weird issues.

PR #23095, opened 2017-07-15
Add position utilities class

### LINKED ISSUE #18476: Add position utility classes

CVarisco:
Utility class to set `position` property. Old PR #19292 with discussion. Fixes #18476.


scss/utilities/_position.scss
```diff
@@ -23,3 +23,8 @@
     z-index: $zindex-sticky;
   }
 }
+
+.pos-s { position: static; }
```
mdo: You're missing `position: sticky`, which, with this naming scheme, would be problematic. We'll need that utility and a different name.

mdo (changes requested):

scss/utilities/_position.scss
CVarisco: You are right!
Could be a good naming scheme ?
`pos-sti`
`pos-sta`
`pos-rel`
`pos-abs`
`pos-fix`

mdo: At that point, we might as well use the full value names. `.p-sticky`, `.p-static`, etc.

CVarisco: Cool!
Thank you, I will a commit with this edit 🙂

jipexu: why not keep pos- or po- as prefix as p is use for padding ?

mdo: Because there's no risk of overlap with padding—these values are unique to positioning.

mdo:
I think your rebase went awry a bit :).

CVarisco:
ops, I'm sorry for that 🙁 
would you like that I squash commits ?

mdo:
I can squash as I merge when it comes to that :)

mdo:
Closing as this history got messed up.

PR #23099, opened 2017-07-16
Errata from #23094.

bardiharborow:
`package-lock.json` didn't check in properly for #23094 due to ba3c61b32edac65d35823b1130a513cd121f0891 (now reverted).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,3 @@ docs/.jekyll-metadata
 # Folders to ignore
 bower_components
 node_modules
-
-# node
-package-lock.json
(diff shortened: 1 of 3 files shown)


mdo:
Nooooooo the `package-lock.json`. 😭 

Mostly jk-ing. I'm just amazed that we have a 6,000 line file for our dependencies.

bardiharborow:
@mdo, remember that we've always had `build/npm-shrinkwrap.json`, this just switches to `package-lock.json` which is better for several reasons (happy to discuss if need be).

mdo:
Hah, just playing around. Front-end development is all sorts of messed up these days 😆

PR #23101, opened 2017-07-16
docs: New favicons and add a service worker

XhmikosR:
Include everything for most browsers to work including Android.

Kept favicon.ico to the root dir.

/CC @mdo @bardiharborow @Johann-S


Johann-S:
Overlapping with #22895 but I prefer this way to update our favicon 👍

XhmikosR:
Doesn't overlap because I kept favicon.ico in root for this reason :)

XhmikosR:
Added a service worker.

Johann-S:
What's the goal of this Service Worker ?

XhmikosR:
https://developers.google.com/web/progressive-web-apps/

First thing I want to make https://developers.google.com/web/updates/2017/02/improved-add-to-home-screen work and then we can look into the offline stuff.

mdo:
#22895 was slated for Beta 2, not Beta 1. Similarly, this PR isn't in scope for our first beta release either. Before we attempt to merge, let's better understand what this is and why we need it.

As an aside, 3% of our docs traffic is from Android. I have a hard time justifying new code and optimizations there given the audience size. Not saying that PWA are bad in general, just that we don't really gain much of anything by doing this.

XhmikosR:
@mdo: this isn't only for Android, since PWAs are the future. It's just that currently that's where it works.

This is a nice addition and along with offline capabilities which I plan to add, would help reading our docs offline etc.

sw.js
```diff
@@ -0,0 +1 @@
+// empty for now
```
Johann-S: Is it ok to merge an empty file ?

XhmikosR: Well, otherwise it's a 404. This is until we actually add any code for service workers.

Johann-S (approved):

PR #23104, opened 2017-07-16
Docs fixes for badges in buttons

mdo:
This corrects some changes from #23060:

- Buttons must have a base and modifier class. I've added `.btn-primary` to fix that.
- Given the primary buttons, I've changed the badges to `.badge-light` for contrast.
- `.bd-example` and `{% highlight %}` don't need to be separate, so I've combined them with our `{% example %}` blocks designed for these situations.
- I've tweaked some docs content: added a serial comma, broke apart a rather large paragraph, and consolidated some text.

/cc @patrickhlauke


gijsbotje (approved):
found no typo's and looks good

PR #23106, opened 2017-07-17
Fix #23100: Add xs: 100% in $container-max-widths

### LINKED ISSUE #23100: Remove duplicate max-width and make it as default
when put container in flex box. work well but not in xs size.

before : https://codepen.io/kimix/pen/VWNaXj

and remove duplicate max-width for each media query, just put width 100% as default

after: https://codepen.io/kimix/pen/NgmNYY

chiraggmodi:
Fix issue mention in #23100, i have added xs: width 100% in _variable.scss file $container-max-widths.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -189,7 +189,6 @@ $grid-breakpoints: (
 // Define the maximum width of `.container` for different screen sizes.
 
 $container-max-widths: (
-  xs: 100%,
   sm: 540px,
   md: 720px,
   lg: 960px,
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -189,6 +189,7 @@ $grid-breakpoints: (
 // Define the maximum width of `.container` for different screen sizes.
 
 $container-max-widths: (
+  xs: 100%,
   sm: 540px,
   md: 720px,
   lg: 960px,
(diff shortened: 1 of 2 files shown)


hobroker (approved):

gijsbotje:
I think the fix that was provided in the issue you mentioned is a better fix
He proposed to make `width: 100%` a default for `.container` and only use max-width for the breakpoints

chiraggmodi:
so this PR is good, right? no need to change as per his suggestion?

gijsbotje:
you added a xs value to the `container-max-widths` map
what #23100  suggest is chaging the function that creates the sass for every breakpoint to only contain the `max-width` property, and to set `width: 100%` on `.container`

chiraggmodi:
@gijsbotje have made change as per suggestion in #23100

mdo:
Digging this change for Beta 1—testing and will merge with some additional changes shortly. Here's a quick recap on why it's helpful:

- We've had hella duplicate issues of #21590, which dealt solely with navbar containers collapsing. This issue obviously extends beyond that one component.
- #21722 aimed to resolve the issue with auto margins. #22383 built on that margin fix with an explicit `width: 100%` for a subset of our breakpoints.

With this PR, we can remove both those previous fixes for the original navbar issue, thereby fixing all containers in and outside navbars.

chiraggmodi:
Thanks for merging
and did you have any idea how could i see all contributor list of this project.
on this URL only top 100 contributors are displaying.
https://github.com/twbs/bootstrap/graphs/contributors

PR #23108, opened 2017-07-17
Fix #23013: Add boder-top:0 in .list-group-flush .list-group-item

### LINKED ISSUE #23013: .list-group-flush not working properly with Chrome 59
The following code used to work just fine in all browsers:
```
<div class="card mx-3 my-3" style="width: 300px">
    <h5 class="card-header list-group-header text-muted">Card header</h5>
    <div class="selection list-group-flush">
        <span class="list-group-item list-group-item-action">List item A</span>
        <span class="list-group-item list-group-item-action">List item B</span>
        <span class="list-group-item list-group-item-action">List item C</span>
        <span class="list-group-item list-group-item-action">List item D</span>
    </div>
</div>
```
Since the Chrome 59 update however, it does not work properly. The borders between the list items are missing.

Result in Edge (expected result):
image

Result in Chrome 59:
image

jsfiddle: https://jsfiddle.net/6oggze6h/

chiraggmodi:
Fixes #23013. I have added `boder-top:0` in **file _list-group.scss** in class `.list-group-flush .list-group-item` this isssue only shows in chrome browser.

### DIFF
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -89,6 +89,7 @@
     border-bottom: 0;
     border-left: 0;
     border-radius: 0;
+
     &:first-child {
       border-top: 0;
     }
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -85,9 +85,9 @@
 
 .list-group-flush {
   .list-group-item {
+    border-top: 0;
     border-right: 0;
     border-left: 0;
-    border-top: 0;
     border-radius: 0;
   }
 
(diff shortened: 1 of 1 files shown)


gijsbotje:
i think you misread the issue, we want the border to be visible not to get rid of the border
Your change gets rid of the border

chiraggmodi:
@gijsbotje mistakenly i have removed border from top instead of bottom, currently there are border top as well as bottom which creating this issue, so i have remove border bottom to fix issue, i am also attaching screenshot of chrome, firefox and edge browser. Also i have remove double border from card header. Due to `border-top` in `.list-group-item` and `border-bottom` in `.card-header`

hirenbhut93:
@mdo This PR is break the following card layout 

screenshot-getbootstrap com-2017-10-04-23-16-24

PR #23109, opened 2017-07-17
fix 23034: added Variable for Navbar padding

### LINKED ISSUE #23034: Update responsive navbar padding to use variables
Currently .nabvar-expand-* classes adds padding to .nav-link wrapped by `media-breakpoint-up`
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L168-L171

This should be moved few lines above to `media-breakpoint-down`  https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L143 

It would be great if those paddings were variables.

EDIT:
should be moved to `media-breakpoint-down`

chiraggmodi:
I have added `$nav-link-padding` in `variables.scss`. Fixes #23034.

### DIFF
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -85,9 +85,9 @@
 
 .list-group-flush {
   .list-group-item {
+    border-top: 0;
     border-right: 0;
     border-left: 0;
-    border-top: 0;
     border-radius: 0;
   }
 
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -85,7 +85,6 @@
 
 .list-group-flush {
   .list-group-item {
-    border-top: 0;
     border-right: 0;
     border-left: 0;
     border-radius: 0;
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -87,6 +87,7 @@
   .list-group-item {
     border-right: 0;
     border-left: 0;
+    border-top: 0;
     border-radius: 0;
   }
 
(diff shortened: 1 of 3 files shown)


scss/_list-group.scss
```diff
@@ -85,6 +85,7 @@
 
 .list-group-flush {
   .list-group-item {
+    border-top: 0;
```
wojtask9: this change is unrelated.

scss/_variables.scss
```diff
@@ -528,6 +528,7 @@ $zindex-tooltip:            1070 !default;
 
 // Navs
 
+$nav-link-padding:              .5rem !default;
```
wojtask9: I would call this $navbar-link-padding and put with others navbar-* variables.
Also consider if split between x/y axis wouldn't be better

scss/_list-group.scss
chiraggmodi: yes sorry for that, actually this is for another PR

scss/_variables.scss
chiraggmodi: yes that would be better will update this changes are PR again.

chiraggmodi: I have made your requested changes, but for navbar link variable i have splited  in left/right instead of x/y as this are for left and right side not left/right and top/bottom.

chiraggmodi:
because if some folks wants to give different value then they can overwrite those variable.

gijsbotje:
I get that but then all the values with `-x` and `-y` would have to be changed to
I would suggest to leave this to the programmer to overwrite this in custom scss and just stick with `-x`

chiraggmodi:
have change `$navbar-link-padding` variable to `$navbar-link-padding-x`

mdo:
This was superseded by #25117

PR #23025, opened 2017-07-09
Fixed Chrome box sizing inheritance bug (#22872)

prateekgoel:
I have changed:
```
*,
*::before,
*::after {
  box-sizing: inherit; }
```
to
```
*,
*::before,
*::after {
  box-sizing: border-box; }
```

Fixes #22872.

PR #23118, opened 2017-07-18
Fix to Chrome inheritance bug (#22872)

### LINKED ISSUE #22872: Chrome details element doesn't correctly propagate box-sizing
There is a [bug in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=589475&q=details%20box-sizing&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified) where inheritance of box-sizing does not happen correctly with the `details` element. This results in all `input` (and every other element that is `width: 100%`)  overflowing the `details`.

This is demonstrated by [this codepen](https://codepen.io/andrewbelcher/pen/mwWZEM).

The solution is to add:
```
details > * {
  box-sizing: border-box;
}
```

Would post a PR, but can't get grunt working locally and don't have time to fix it now.

prateekgoel:
I have changed:

```css
*,
*::before,
*::after {
  box-sizing: inherit; }
```
to

```css
*,
*::before,
*::after {
  box-sizing: border-box; }
```

Also, to address duplication of `box-sizing` from the `html` selector above, changed `box-sizing:border-box;` to `box-sizing:inherit;` from `html`.

Fixes #22872.
Resolves the issue with PR https://github.com/twbs/bootstrap/pull/23025.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -18,20 +18,19 @@
 //    we force a non-overlapping, non-auto-hiding scrollbar to counteract.
 // 6. Change the default tap highlight to be completely transparent in iOS.
 
+*,
+*::before,
+*::after {
+  box-sizing: border-box; // 1
+}
+
 html {
-  box-sizing: inherit; // 1
   font-family: sans-serif; // 2
   line-height: 1.15; // 3
   -webkit-text-size-adjust: 100%; // 4
   -ms-text-size-adjust: 100%; // 4
   -ms-overflow-style: scrollbar; // 5
   -webkit-tap-highlight-color: rgba(0,0,0,0); // 6
-}
-
-*,
-*::before,
-*::after {
-  box-sizing: border-box; // 1
 }
 
 // IE10+ doesn't honor `<meta name="viewport">` in some cases.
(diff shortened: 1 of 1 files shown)


scss/_reboot.scss
```diff
@@ -19,7 +19,7 @@
 // 6. Change the default tap highlight to be completely transparent in iOS.
 
 html {
-  box-sizing: border-box; // 1
+  box-sizing: inherit; // 1
```
mdo: This should just be removed outright, and the comments above updated.

prateekgoel: Ok I am on it. I don't think we need to update comments. As the comments are still correct.

prateekgoel:
@mdo This is done. Let me know if any other changes are required.

scss/_reboot.scss
prateekgoel: This is done. Let me know if any other changes are required.

mdo (approved):
Looks good, thanks! We'll get this in Beta 2.

prateekgoel:
Ok thanks.

PR #23121, opened 2017-07-19
fix .col-form-label paddings

### LINKED ISSUE #23447: Form elements position is off
On the ["Forms" page](https://getbootstrap.com/docs/4.0/components/forms/) In the docs some (not all) form elements position are off:

1. Bad:
    twbs_form_example_bad_1

2. Good:
    twbs_form_example_good_2

3. Bad:
    twbs_form_example_good_3

Versions I'm looking with:

- macOS 10.12.3
- Chrome 60.0.3112.90

wojtask9:
Fixes #23447.

Current calculations are wrong for me.
First we shouldn't multiply border by two. Second we should add (not substract) border to padding.
Also we want the same line-height like `.form-control` have. 
Without line-height `.from-control-plaintext` isn't properly aligned with *-sm/lg modifiers.

We can avoid `calc()` too. Just add to `.col-form-label` border:
` border: $input-btn-border-width solid transparent;`


wojtask9:
Forgot to mention that `.form-control-*` also have line-height so `.col-form-label-*` need line-height too.

mdo (approved):
Verified this locally and it looks good. Thanks!

Hima1717:
How do I stop all this going to my mail

Johann-S:
https://help.github.com/articles/about-notifications/ @Hima1717 :wink:

PR #23122, opened 2017-07-19
Remove the Internet Explorer 10 for Windows Phone 8 workaround

coliff:
I suggest that there's no longer a need to include the section on workaround for the IE10 bug on Windows Phone 8. Not only was this bug patched in a later release of Windows Phone 8 (which shipped in 2013), but Windows Phone 8's support ended a long time ago - even its successor Windows Phone 8.1 is no longer supported by Microsoft. The potential audience affected by this bug is likely microscopic so I suggest it can be safely removed. 

I suggest the workaround JS (http://getbootstrap.com/assets/js/ie10-viewport-bug-workaround.js) can also be safely removed from the docs and examples but I haven't done that in this PR.


mdo (approved):
Nice, thanks! We'll get this in for Beta 2.

XhmikosR:
Is it possible to do it in this PR @coliff?

Also, CC @patrickhlauke for confirmation.

XhmikosR:
@coliff: ping ^^

patrickhlauke (approved):

XhmikosR:
@mdo: we need to remove the JS from docs too.

coliff:
@XhmikosR - thanks for taking care of the additional PR to remove the JS from the docs. I'd been away travelling for past 3 weeks and couldn't do it myself.

PR #23138, opened 2017-07-22
Add margin variable for breadcrumb

prateekgoel:
I have added a variable for breadcrumb margin to `_variables.scss` and replaced static margin in `_reboot.css` with that variable.

Fixes https://github.com/twbs/bootstrap/issues/23016.

PR #23140, opened 2017-07-22
Add margin variable for paragraphs

### LINKED ISSUE #23031: [v4] Introduce variable for customizing margin-bottom for "p"
_reboot.scss has:

```
p {
  margin-top: 0;
  margin-bottom: 1rem;
}
```

However the value is hardcoded. It would be nice to have it exported to _variables.scss so that people can customize (_custom.scss was nice but looks like it's removed now).

prateekgoel:
I added a section for paragraphs in _variables.scss, used that variable to change hard coded margin value for p element.

Fixes #23031

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -172,7 +172,7 @@ $link-hover-decoration: underline !default;
 //
 // Style p element.
 
-$para-margin: 0 0 1rem;
+$paragraph-margin-bottom: 1rem;
 
 
 // Grid breakpoints
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -13,6 +13,7 @@
 // Spacing
 // Body
 // Links
+// Paragraphs
 // Grid breakpoints
 // Grid containers
 // Grid columns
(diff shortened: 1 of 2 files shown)


mdo (changes requested):
Paragraphs don't have horizontal margin on them (by default or from Bootstrap), so there's no need to reset them if there is nothing there to begin with.

Please restore the `margin-top: 0;` and use a variable for the `margin-bottom`.

Also, no need to shorten the word to four letters—use `$paragraph-margin-bottom` or something like that maybe?

prateekgoel:
Ok. I'll do these changes. Thanks.

prateekgoel:
@mdo For now, I have restored `margin-top: 0;`
Let me know if anything else need to be done.

prateekgoel:
@mdo Any update on this?

scss/_variables.scss
```diff
@@ -787,6 +794,8 @@ $breadcrumb-padding-y:          .75rem !default;
 $breadcrumb-padding-x:          1rem !default;
 $breadcrumb-item-padding:       .5rem !default;
 
+$breadcrumb-margin:             0 0 1rem !default;
```
mdo: This should be here

mdo (changes requested):
This'll go in for Beta 2 at the earliest.

scss/_variables.scss
prateekgoel: @mdo Should I add breadcumb-margin too? I mean this PR is for paragraph-margin. For breadcrumb there's a different PR #23138 .

mdo: Crap, I typo-ed—I meant shouldn't be here. It's in the wrong PR, so you can remove it.

prateekgoel: @mdo Did the changes.

PR #23143, opened 2017-07-23
v4 sticky footer example with flexbox

sepiariver:
In July 2017 flexbox support is at [97.48%](http://caniuse.com/#search=flexbox) (prefixed). In the past, the sticky footer element has relied on absolute positioning and a fixed height. The utility of a fixed height footer is limited, not to mention the unsightly mess created when the footer content expands beyond the predetermined fixed height. 

With the use of flexbox, the sticky footer example CSS becomes lighter and the sticky footer element's height can be `auto`.

### DIFF
--- a/docs/4.0/examples/sticky-footer-navbar/index.html
+++ b/docs/4.0/examples/sticky-footer-navbar/index.html
@@ -54,7 +54,9 @@
 
     <footer class="footer">
       <div class="container">
-        <span class="text-muted">Place sticky footer content here.</span>
+        <p class="text-muted">Place sticky footer content here.</p>
+        <img class="img-fluid" src="data:image/gif;base64,R0lGODlhAQABAIAAAHd3dwAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==">
+        <p class="text-muted">With Flexbox the footer height doesn't need to be fixed.</p>
       </div>
     </footer>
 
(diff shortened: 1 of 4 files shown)


alanstratos:
`justify-content: space-between` will mess things up with more elements in the page.
I have achieved better results with this code:

```
body {
    min-height: 100vh
    display: flex
    flex-direction: column
}

.flex-grow {
   flex: 1 1 auto
}
```
Now just add `.flex-grow` to your main `.container` and it will push the footer to the bottom.

sepiariver:
Thanks @alanstratos that looks pretty good. Just to be clear, that will push down everything that comes after the `.flex-grow` element in the source. Not to say there's a problem with that—it could actually come in handy—but just to understand the ramifications, is that correct?

Another question: do you want to submit a PR with it, in which case I could close this one, or would you like me to incorporate it into this one?

Lastly it's worth mentioning IE11 has a [bug](https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview) that makes it incompatible with this whole approach, unless you can explicitly set a height on the `<body>`. To support IE11 a polyfill would be required. If it's deemed worth pursuing, I could add a polyfill specific to this bug, and this implementation of flex, to the PR. Thoughts?

XhmikosR:
Why don't you use the helper classes instead?

andrerom:
I tried to do more or less exactly the same thing when testing out boostrap 4.0-beta now, both using plain css with flexbox on body and footer, and using the flex box utility classes.

However this messes with the height of the navbar when page is bigger than screen size, seen this?

ysds:
Solved by #26674 so you can close it.

Issue #23053, opened 2017-07-12
Responsive font sizes


MartijnCuppens:
Currently there is no font-resizing implemented in Bootstrap. Font-resizing is needed to make large texts readable on all devices and to prevent longer words and [compounds](https://en.wikipedia.org/wiki/Compound_(linguistics)) to fall off screen. 

It's currently possible to alter the font-size of the html element with media queries (see [Responsive typography](https://getbootstrap.com/docs/4.1/content/typography/#responsive-typography)), but this method can cause the font-size of text to be too small on mobile devices and because the rem-value is overridden, users can't change their font-size in their browser anymore.

[RFS](https://github.com/MartijnCuppens/rfs) can solve this by because this mixin automatically calculates the appropriate font-size based on the width of the device. 

RFS is also available in Sass, Less, PostCSS and Stylus.

**Example problem with bigger font-sizes:** 
http://getbootstrap.com/docs/4.1/content/typography/#display-headings

**RFS solution:**
https://project-rfs.github.io/bootstrap/docs/4.1/content/typography/#display-headings

**Bootstrap implementation demo:** 
https://project-rfs.github.io/bootstrap/

**Github page:** 
https://github.com/project-rfs/rfs

**Pull request**
https://github.com/twbs/bootstrap/pull/23816

gijsbotje:
would love this feature, how do you suggest implementing this though?
what would be my first thought is a class that can be added to the `html` element, like `.responsive-font-size`

and would you rather have responsive or [fluid](https://madebymike.com.au/writing/fluid-type-calc-examples/) font size?

just some questions that pop in to my head

MartijnCuppens:
@gijsbotje, for the demo I changed the font-sizes in the `_variables.scss`-file into unitless numbers like this:

```
$font-size-h1: 40 !default;
$font-size-h2: 32 !default;
$font-size-h3: 28 !default;
$font-size-h4: 24 !default;
$font-size-h5: 20 !default;
$font-size-h6: 16 !default;
```

In the `.scss`-files I changed the `font-size` -declarations to the RFS-mixin:
```
h1, .h1 { @include rfs($font-size-h1); }
```

It's also still possible to use fixed values like `15px` or `1.25rem` by passing them to the mixin.

<br>
So at this moment there is not really a way to enable/disable it per page. However it is possible to alter the mixin by adding a class inside it.

[...]

PR #23157, opened 2017-07-26
Responsive font-sizes implementation

MartijnCuppens:
Implementation of responsive font-size system [RFS](https://github.com/MartijnCuppens/rfs/).

Link to feature-request:
https://github.com/twbs/bootstrap/issues/23053

CC: @XhmikosR, @kimamil, @mdo


mabar:
[Poly fluid sizing](https://www.smashingmagazine.com/2017/05/fluid-responsive-typography-css-poly-fluid-sizing/#the-holy-grail-of-css-sizing) would be better. More fluent.

MartijnCuppens:
@mabar, this implementation doesn't require to define mappings for every font-size so it can easily be used by users who do not have much frontend experience.

kimamil (changes requested):
Can you add an example in the documentation on how to use this with rem? As I understood, rfs must be 'false' and unit set to 'rem'. Then I could use media-breakpoints on the html tag to set the font-size in 'rem'? But basically you suggest using fixed px sizes to set the base font-size in all breakpoints, correct?

Else, nice catch!

docs/4.0/content/responsive-font-sizes.md
```diff
@@ -7,48 +7,111 @@ toc: true
 ---
 
 ## Responsive font sizes
-Bootstrap uses <a href="https://github.com/MartijnCuppens/rfs" target="_blank" rel="noopener">RFS</a> to control its font-size. RFS is an SCSS-mixin which automatically calculates the correct font size based on the browsers screen width. You just have got to define your font-size for big screens and the font size will automatically decrease for smaller screens.
+Bootstrap uses <a href="https://github.com/MartijnCuppens/rfs" target="_blank" rel="noopener">RFS</a> to control its font size. RFS is an SCSS-mixin which automatically calculates the correct font size based on the browsers screen width. You just have got to define your font size for big screens and the font size will automatically decrease for smaller screens.
```
XhmikosR: Please change this to a normal markdown link. Same for other ones too if any.

MartijnCuppens:
@kimamil, the previous documentation might indeed be a little bit confusing since it didn't explain how the configuration works but explained how you disable RFS and use rem-units for font-resizing just like you can do now.

I've updated the documentation and tried to explain how every setting of the configuration changes the way the responsive font size is calculated. Is this a bit clear or does anyone has any suggestions on how to clarify this even more?

docs/4.0/content/responsive-font-sizes.md
MartijnCuppens: Done: https://github.com/twbs/bootstrap/pull/23157/commits/aa676531a18d3a099de676e9cfcbdad6fb205ed1

PR #23161, opened 2017-07-27
fixed top-right/left bottom-right/left style

huang-x-h:
when tooltip set class `top-right` or `top-left` , the tooltip height doesn't equal when class is `top`

see example http://output.jsbin.com/dayucif


Johann-S:

huang-x-h:
OK, thanks

PR #23162, opened 2017-07-27
Update Popper.js to 1.11.0

Johann-S:
Update Popper.js to have those awesome bugfixes : https://github.com/FezVrasta/popper.js/releases/tag/v1.11.0

IMO this PR should be merged before our first beta @mdo
and I removed the line which add an error because we don't have Popper's `.js.map` file @XhmikosR 😉


package-lock.json
```diff
@@ -3267,9 +3267,9 @@
       "dev": true
     },
     "popper.js": {
-      "version": "1.10.8",
```
XhmikosR: Are you sure there are no other deps due to the popper.js update?

Johann-S: Popper.js has no deps so it shouldn't have other deps

bardiharborow: Popper.js has no dependencies.

XhmikosR: Right. Looks good to me then. 👍

XhmikosR (approved):

bardiharborow (approved):
Approved on the code. I can't offer an opinion as to whether this should ship in Beta 1 or 2.

PR #23186, opened 2017-07-31
Change base 64 encoded GIFs to SVGs

### LINKED ISSUE #23185: Replace base64 gif placeholder images with SVG ones
Currently we use gif base64 placeholder images in a few pages. We should switch to SVGs.

tom-sherman:
Fixes #23185 

For reference, an example of the decoded SVG is as follows:

```svg
<svg width="1" height="1" xmlns="http://www.w3.org/2000/svg"><path fill="#00dc83" d="M0 0h1v1H0z"/></svg>
```

Minified using SVGOMG: https://github.com/jakearchibald/svgomg

### DIFF
--- a/site/docs/4.1/examples/carousel/index.html
+++ b/site/docs/4.1/examples/carousel/index.html
@@ -103,19 +103,19 @@
         <!-- Three columns of text below the carousel -->
         <div class="row">
           <div class="col-lg-4">
-            <svg width="140" height="140"><circle cx="70" cy="70" r="70" fill="#777"/></svg>
+            <span class="d-block mx-auto rounded-circle" style="background-color: #777; width: 140px; height: 140px;"></span>
             <h2>Heading</h2>
             <p>Donec sed odio dui. Etiam porta sem malesuada magna mollis euismod. Nullam id dolor id nibh ultricies vehicula ut id elit. Morbi leo risus, porta ac consectetur ac, vestibulum at eros. Praesent commodo cursus magna.</p>
             <p><a class="btn btn-secondary" href="#" role="button">View details &raquo;</a></p>
           </div><!-- /.col-lg-4 -->
           <div class="col-lg-4">
-            <svg width="140" height="140"><circle cx="70" cy="70" r="70" fill="#777"/></svg>
+            <span class="d-block mx-auto rounded-circle" style="background-color: #777; width: 140px; height: 140px;"></span>
             <h2>Heading</h2>
             <p>Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit. Cras mattis consectetur purus sit amet fermentum. Fusce dapibus, tellus ac cursus commodo, tortor mauris condimentum nibh.</p>
             <p><a class="btn btn-secondary" href="#" role="button">View details &raquo;</a></p>
           </div><!-- /.col-lg-4 -->
           <div class="col-lg-4">
-            <svg width="140" height="140"><circle cx="70" cy="70" r="70" fill="#777"/></svg>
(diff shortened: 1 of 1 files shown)


XhmikosR:
Thanks for the patch!

But, you shouldn't base64 encode the SVGs. That being said, we need to make sure some characters are escaped otherwise FF can have issues.

bardiharborow:
Also `data:image/svg+xml;base64,DQo8c3ZnIHdpZHRoPSIxIiBoZWlnaHQ9IjEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHBhdGggZmlsbD0iIzc3NyIgZD0iTTAgMGgxdjFIMHoiLz48L3N2Zz4=` has a linebreak at the start of the file.

tom-sherman:
@XhmikosR You can blame 36 hours of no sleep for that one! I'll change it to use SVG elements isntead

XhmikosR:
Also, we might as well have 2 instances for proper aspect ratio. It's worth looking into it.

mdo:
What's the difference in size between the two formats?

tom-sherman:
@mdo encoded they have about twice the char length, amounts to about an additional 16 bytes per svg. I would guess that this number would decrease if they weren't encoded

XhmikosR:
Yup, that is why I said you shouldn't base64 encode the SVGs :)

Without that, it should be even shorter and more flexible with custom width/height etc.

XhmikosR:
@tom-sherman: any updates?

tom-sherman:
@XhmikosR changed to inline SVGs

docs/4.0/examples/carousel/index.html
```diff
@@ -99,19 +99,25 @@
...
+          <svg alt="Generic placeholder image" height="140" width="140">
+            <circle cx="70" cy="70" r="70" fill="#777" />
+          </svg>
           <h2>Heading</h2>
           <p>Duis mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit. Cras mattis consectetur purus sit amet fermentum. Fusce dapibus, tellus ac cursus commodo, tortor mauris condimentum nibh.</p>
           <p><a class="btn btn-secondary" href="#" role="button">View details &raquo;</a></p>
         </div><!-- /.col-lg-4 -->
         <div class="col-lg-4">
-          <img class="rounded-circle" src="data:image/gif;base64,R0lGODlhAQABAIAAAHd3dwAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==" alt="Generic placeholder image" width="140" height="140">
+          <svg alt="Generic placeholder image" height="140" width="140">
```
XhmikosR: Could you not pretty-print these like you do with the above SVGs?

XhmikosR: Or rather, maybe just pretty-print the 3 SVGs since they are less :)

docs/4.0/examples/carousel/index.html
```diff
@@ -49,7 +49,7 @@
       </ol>
       <div class="carousel-inner">
         <div class="carousel-item active">
-          <img class="first-slide" src="data:image/gif;base64,R0lGODlhAQABAIAAAHd3dwAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw==" alt="First slide">
+          <svg width="100%" height="100%" class="first-slide" alt="First slide" xmlns="http://www.w3.org/2000/svg"><path fill="#777" d="M0 0h1v1H0z"/></svg>
```
XhmikosR: Please use the pixel values for width and height in both examples.

XhmikosR:
@tom-sherman: do you plan to take care of the feedback above?

XhmikosR:
Closing this since we are going to go with another solution.

PR #23191, opened 2017-08-01
Reuse of the color-yiq function

### LINKED ISSUE #23114: Reuse of the color-yiq function
With a simple change it is possible to use the color-yiq function with other properties, not just "color"

```sass
// Color contrast
@mixin color-yiq($color, $prop: 'color') {
  $r: red($color);
  $g: green($color);
  $b: blue($color);

  $yiq: (($r * 299) + ($g * 587) + ($b * 114)) / 1000;

  @if ($yiq >= 150) {
    #{$prop}: #111;
  } @else {
    #{$prop}: #fff;
  }
}
```

```sass
.element-inside-primary {
  @include color-yiq(theme-color("primary"), 'background-color');
}
```

Just add the parameter "**$prop: 'color'**" that we can use other properties
This can be interesting when you want to assign a background that contrasts with a color.

envolute:
Fixes #23114.

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -49,7 +49,7 @@
 }
 
 // Color contrast
-@mixin color-yiq($color, $prop: 'color') {
+@mixin color-yiq($color, $prop: "color") {
   $r: red($color);
   $g: green($color);
   $b: blue($color);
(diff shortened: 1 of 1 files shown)


mdo:
Punting on this one in favor of either #23748 or #23943. Thanks though!

PR #23196, opened 2017-08-01
Update .babelrc

tonywudiji:
just for test


patrickhlauke:
don't use public repositories to just "test" how to use git/github

PR #23206, opened 2017-08-02
Adds 'system-ui' to the font stack following W3C new font spec

andresgalante:
There is a new `system-ui` value defined for `font-family` at [W3C CSS Fonts Level 4](https://www.w3.org/TR/css-fonts-4/#system-ui-def) working draft.

> `system-ui` is intended to let text render with the default user interface font on the platform on which the UA is running.

Since Bootstrap emulates a system-ui font stack, I though that by adding this value we are future proofing the stack.

What do you think?

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -231,7 +231,7 @@ $transition-collapse:    height .35s ease !default;
 //
 // Font, line-height, and color for body text, headings, and more.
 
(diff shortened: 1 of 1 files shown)


andresgalante:
I am going to close it due to #22328

bardiharborow:
Yep. We're unfortunately not going to see much movement on this until the font issue is fixed with Microsoft upstream, and most machines are updated or become obsolete.

PR #23208, opened 2017-08-02
Reduce number of spacing rules

mstone6769:
The goal was to reduce the number of generated spacing rules. We do this by combining x & y rules with top, bottom, right, and left rules

### DIFF
--- a/scss/utilities/_spacing.scss
+++ b/scss/utilities/_spacing.scss
@@ -10,11 +10,11 @@
         .#{$abbrev}#{$infix}-#{$size}  { #{$prop}:        $length !important; }
         .#{$abbrev}t#{$infix}-#{$size},
         .#{$abbrev}y#{$infix}-#{$size} {
-          #{$prop}-top:    $length !important;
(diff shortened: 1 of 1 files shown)


scss/utilities/_spacing.scss
```diff
@@ -8,34 +8,42 @@
...
-    .ml#{$infix}-auto { margin-left:   auto !important; }
+    .mt#{$infix}-auto,
+    .my#{$infix}-auto {
+      margin-top:    auto !important;
+    }
+    .mr#{$infix}-auto,
     .mx#{$infix}-auto {
-      margin-right: auto !important;
-      margin-left:  auto !important;
+      margin-right:  auto !important;
```
XhmikosR: Can you remove the spaces after `:` that don't make sense to have like here and the other places you edited?

mstone6769:
During my testing I saw a reduction 6,228 bytes in the minified file by combining the x & y rules with the t, r, b, l rules

mdo (approved):

PR #23244, opened 2017-08-08
Move jQuery and Popper to peerDependencies

### LINKED ISSUE #23204: Remove `jquery` from required packages
Relates to #22194

`jQuery` should be removed from the mandatory dependencies for Bootstrap library.

Since Bootstrap is posioned as CSS framework (not JS one) and it is declared it can function without JQuery (even while it is limiting its features), there should be a way to install pure Bootstrap without JQuery from the repositories like Bower/Npm.

In particular, these lines should be removed:
https://github.com/twbs/bootstrap/blob/v4-dev/bower.json#L34-L35

In the modern web application development more and more projects relies on suhc JS libraries like AngularJS, ReactJs, EmberJS and so on, in pursuit for 'on-page-site' feature implementation.

If particular developer wishes to use AngularJS and Bootstrap, why should he install a 'jquery' package, which will lie as a garbage among the project source files?
Keeping 'jquery' as mandatory dependency looks unprofessional.

Ideally JS related code should be moved into a separated repository, which will depend on 'twbs/bootstrap' and 'jquery', while 'twbs/bootstrap' itself will be requirement free.

If this be a Composer installation, the quick solution will be move 'jquery' into [suggest](https://getcomposer.org/doc/04-schema.md#suggest) section, but for Bower/Npm - I can not say.

mdo:
Fixes #23204.

I've made the change to `package.json` as @bardiharborow suggested. I've also taken a quick pass at some docs updates to clarify usage around source files, compiled files, and that our JS plugins are optional and depend on jQuery and Popper.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,8 @@
   },
   "license": "MIT",
   "dependencies": {
+  },
+  "peerDependencies": {
     "jquery": ">=1.9.1",
     "popper.js": "^1.11.0"
   },
--- a/docs/4.0/getting-started/download.md
+++ b/docs/4.0/getting-started/download.md
@@ -6,26 +6,42 @@ group: getting-started
 toc: true
 ---
 
-## Bootstrap CSS and JS
+## Compiled CSS and JS
 
-**Download Bootstrap's ready-to-use code to easily drop into your project.** Includes compiled and minified versions of all our CSS bundles (default, grid only, or Reboot only) and JavaScript plugins. Doesn't include documentation or source files.
+Download ready-to-use compiled code for **Bootstrap v{{ site.current_version}}** to easily drop into your project, which includes:
 
-**Current version:** v{{ site.current_version}}
+- Compiled and minified CSS bundles (default, grid-only, and Reboot-only)
+- Compiled and minified JavaScript plugins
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -66,6 +66,8 @@
   },
   "license": "MIT",
   "dependencies": {
```
RinkAttendant6: Is it necessary to keep an empty `dependencies` object in package.json?

mdo: Probably not? I did to show we have no explicit dependencies...unsure what the best option is there.

RinkAttendant6: Oh ok, I'm unsure as well but just wanted to point that out, I don't think it is harmful to keep it there or to remove it.

XhmikosR (approved):
LGTM

The empty dependencies object is fine, doesn't hurt and in fact `npm init` adds it too when there are no dependencies IIRC.

Johann-S (approved):
LGTM 👍 
It won't cause any issues with my PR to improve our modularisation 😄

package.json
dkrutsko: It's safe to remove `dependencies` if needed. Some of my libraries do that and it works just fine.

XhmikosR: Of course it's safe. It's left on purpose if you read the above comments, to indicate that there are no dependencies.

bardiharborow (approved):

PR #23248, opened 2017-08-08
normalize.css 7.0 updates

rouralberto:
Updating the relevant parts of normalize.css from v5 to v7. In this case the removal of `font-family: sans-serif`. Discussion on https://github.com/necolas/normalize.css/issues/517


mdo:
Per https://github.com/necolas/normalize.css/issues/517#issuecomment-187983661, this isn't an update we want, and is part of the reason we move away from Normalize itself. Without this, the default font for `<html>` elements will be different by browsers.

roura356a:
Ohh ok I see, I thought setting `sans-serif` later made this part irrelevant.

PR #23253, opened 2017-08-08
Make sticky header offset CSS not selectable

### LINKED ISSUE #23252: `::before` from headings overlap previous element

/CC @Johann-S

mdo:
Should fix #23252. Really the `user-select` is probably all we need, but figured I'd also tweak the hack to use `padding-top` instead of a fixed `height`.

### DIFF
--- a/assets/scss/_content.scss
+++ b/assets/scss/_content.scss
@@ -15,10 +15,11 @@
       position: relative;
       z-index: -1;
       display: block;
-      height: 6rem;
+      padding-top: 6rem;
       margin-top: -6rem;
       visibility: hidden;
       content: "";
+      user-select: none;
     }
   }
 


XhmikosR (commented):
Doesn't seem to fix the issue. Tested latest Chrome and FF.

XhmikosR:
Using `4rem`, which is the min-height for the navbar, improves things a bit but still not perfect.

XhmikosR:
That along with increasing margin-top for the header itself almost fixes the issue. Something like

```css
.bd-content {
  > h2[id],
  > h3[id],
  > h4[id] {
    &::before {
      position: relative;
      z-index: -1;
      display: block;
      padding-top: 4rem;
      margin-top: -4rem;
      visibility: hidden;
      content: "";
      user-select: none;
    }
  }
.bd-content > h3 {
    margin-top: 3rem;
}
```

XhmikosR:
@mdo this can be closed in favor of #23257

mdo:
This isn't closed in favor of that PR, but it is closed as it's an incomplete solution—the `::before` content will still overlay the content that comes before it. Going to try a few more options still.

PR #23255, opened 2017-08-08
Fix sourcemaps

### LINKED ISSUE #23254: CSS sourcemaps are broken
I'm on Windows and tested Firefox but shouldn't really matter.

/CC @Johann-S @mdo @bardiharborow

XhmikosR:
Without this, the sourcemaps aren't found for me.

Fixes #23254


XhmikosR:
Actually, can someone try this locally who is not on Windows and see if it fixes the issue?

I have opened an issue with clean-css because I see the paths use Windows separator on Windows.

XhmikosR:
OK, I confirmed it fixes the issue for me.

Johann-S (approved):
LGTM 👍 on Win10

XhmikosR:
Thanks @Johann-S for confirming!

@mdo: should we merge this now? Currently the sourcemaps are broken for v4.

mdo:
I've no idea why this fails elsewhere, but it doesn't break source maps for me, so :shipit:!

PR #23256, opened 2017-08-08
Update sticky header offset CSS & JS

MartijnCuppens:
Should fix #23252. This sets `user-select: none;` on the heading-elements and `user-select: auto;` on the direct child. All content in the heading-elements are wrapped in spans by jQuery.

PR #23257, opened 2017-08-08
Fix headers conflicting with `pre`

### LINKED ISSUE #23252: `::before` from headings overlap previous element

MartijnCuppens:
Fixes #23252. This sets `pointer-events: none;` on the heading-elements and `pointer-events: auto;` on the direct child. All content in the heading-elements is wrapped in spans by jQuery.


mdo:
Does #23244 not handle this well enough?

XhmikosR:
@mdo you probably meant #23256, which does not fix the issue as I mentioned there.

As for this, @MartijnCuppens this has nothing to do with **M**icro**s**oft. Unless you mean something else with "ms".

Also, please skip the generated files.

Other than that this is the proper fix so 👍

XhmikosR (changes requested):
After removing the generated files and rebasing skipping @mdo's patch completely, we can merge this

mdo:
Cleaned up and superseded with #23259.

PR #23259, opened 2017-08-09
Fix sticky headings

### LINKED ISSUE #23252: `::before` from headings overlap previous element

mdo:
Isolates work from @MartijnCuppens in #23257 to fix #23252. Main difference is placement of the JS and more specific CSS so that we're not using unnecessary universal selectors. I wasn't stoked on injecting markup, but any other CSS-only solution seems to have significant drawbacks.

### DIFF
--- a/assets/js/src/application.js
+++ b/assets/js/src/application.js
@@ -96,15 +96,10 @@
         .tooltip('_fixTitle')
     })
 
+    anchors.options = {
+      icon: '#'
+    }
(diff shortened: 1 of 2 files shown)


XhmikosR:
LGTM if you like the new behavior; now the anchor shows up only when you hover over the `#` and not the whole header like it did before.

XhmikosR:
@mdo How about just making that `span` `block`? It would have the same behavior as before.

Issue #23112, opened 2017-07-17
How to override theme colors now?


martinbean:
I’ve noticed in [_variables.scss](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss) that the brand colours have been moved to a nested object.

Up to now, I‘ve been specifying variable overrides in a *_custom.scss* in my projects like this:

```scss
@import "custom";
@import "node_modules/bootstrap/scss/bootstrap";
```

I’d then override the variables I wanted in my *_custom.scss* file like this:

```scss
$brand-primary: #fc0;

$font-family-sans-serif: "Open Sans", sans-serif;
```

How do I override values on a per-variable basis now that they’re inside an object? For example, I have a project where I want the brand’s primary color to be green and not blue.

Sorry if this a n00b question, but I’m under the impression that object members like `$theme-colors[primary]` _aren’t_ overridable in Sass without overriding the whole object, in which case these _seriously_ reduces the configurability of Bootstrap in version 4.

If I’m doing something wrong, or I _can_ override specific members of objects in Sass, then please let me know.

bkdotcom:
http://sass-lang.com/documentation/Sass/Script/Functions.html#map_merge-instance_method

mdo:
While it does appear you cannot override part of the Sass maps, much like our variables, they do have the `!default` flag. [See this example](https://codepen.io/anon/pen/KqLzzX) for how you can change the variable or Sass map. IMO this is a fine approach as it maps to variables—copy, paste, and change the value.

> How do I override values on a per-variable basis now that they’re inside an object?

We also still have all the regular color variables, so you can customize the Sass map or you can change the base variables. See the variables file:

```scss
$blue:    #007bff !default;
$indigo:  #6610f2 !default;
$purple:  #6f42c1 !default;
$pink:    #e83e8c !default;
$red:     #dc3545 !default;
$orange:  #fd7e14 !default;
$yellow:  #ffc107 !default;
$green:   #28a745 !default;
$teal:    #20c997 !default;
$cyan:    #17a2b8 !default;

$colors: (
  blue: $blue,
  indigo: $indigo,
  purple: $purple,
  pink: $pink,
  red: $red,
  orange: $orange,
  yellow: $yellow,
  green: $green,
  teal: $teal,
  cyan: $cyan,
  white: $white,
  gray: $gray-600,
  gray-dark: $gray-800
) !default;

$theme-colors: (
  primary: $blue,
  secondary: $gray-600,
[...]

Issue #22891, opened 2017-06-22
_custom.scss alternatives


silverwind:
Now that `_custom.scss` has been remove in https://github.com/twbs/bootstrap/pull/22821, I wonder how builds with custom colors and the likes could be made. My build so far was

1) Clone `twbs/bootstrap` to a temporary location
2) Copy my `_custom.scss` into the `scss` directory
3) Run npm build scripts
4) Copy the compiled .css and .js files to my project

The PR mentions it was nuked "in favor of relying on imports from package manager directories" but I have a hard time following? Anyone care to explain?

TheBuzzer67:
I always use my own css file loaded at the end to make customization, so it could be a _custom.css file suggested!?

silverwind:
> I always use my own css file loaded at the end to make customization, so it could be a _custom.css file suggested!?

That's what I do too, but it results in duplicate/unneccesary rules from Bootstrap's style being left over in the resulting CSS. It'd be more efficient to override rules before the scss is compiled.

Meanwhile, I've found a workaround to get the old behaviour of `_custom.scss` by prepending it to `_variables.scss` (appending doesn't work):

````sh
$ cat _custom.scss _variables.scss > _variables_new.scss
$ mv -f _variables_new.scss _variables.scss
````

rmoorman:
@silverwind I think that we are supposed to use bootstrap through npm (or the like) and then include/compile the sass files using the facilities the bundler in question gives us.
See the [example over here](https://github.com/mrholek/CoreUI-Free-Bootstrap-Admin-Template/tree/master/React_Full_Project/scss). Even though bootstrap is inlined into the repository (which is not exactly the same as going through npm and node_modules) in this example, you can see [over here](https://github.com/mrholek/CoreUI-Free-Bootstrap-Admin-Template/blob/master/React_Full_Project/scss/style.scss#L15) how it is imported and how variables are adjusted to fit the theme (didn't build the code in there myself yet though).
Would that be a solution for you (sans inlining bootstrap into your repo that is)?

silverwind:
I haven't thought about importing it because most of my projects are pure CSS, so I can only use the compiled CSS for those. But yes, that would work for SCSS projects, which I coincidentally also have a few.

So the gist is that CSS and preprocessors other than Sass are not supported for customization? Isn't that a bit opinionated?

undying-fish:
[...]

PR #23260, opened 2017-08-09
Sass map trickery for easier customization

mdo:
**Peep the [CodePen demo](https://codepen.io/anon/pen/XaJeVd).**

This PR creates a `!default` empty Sass maps for our `$grays`, `$colors`, and `$theme-colors` lists and `map-merges` a default setting for all colors into them. This way, anyone can continue to `map-merge` in additional or replacement `key: value` pairs as needed without having to replace the _entire_ original Sass map as @martinbean pointed out in #23112.

This technique was pointed out by @hokiecoder in https://github.com/twbs/bootstrap/issues/22891#issuecomment-316463806.

---

Thoughts? If folks are interested, we'll need to update the `options.md` docs page to explain this, and add some code comments in our Sass files.


bardiharborow (approved):

martinbean:
@mdo It’s an improvement in that we don’t have to re-define unnecessary variables to override a sub-set now 😀 

Curious as to why the move to holding theme colours in a map now though, rather than simple variables? As my feeling is it was a lot easier to reference `$brand-primary` in custom Sass style sheets than `map-get($theme-colors, "primary")`. I think for designers and front-end developers not familiar with things like maps in Sass, this change may make Bootstrap less accessible to them and steepen the learning curve for them.

kartikluke:
@martinbean They iterate over the map when generating buttons, etc. so just by adding to the array you can add new theme colors for buttons, etc. which is useful.

If it's possible to dynamically generate Sass variables you could iterate through the map and make flat variables which people could use.

liquidvisual:
I agree with @martinbean regarding ease of use. I'm a designer, and I find the ```map-get($theme-colors, "primary")``` so cumbersome to write. I use $brand-primary everywhere, and it's a no-brainer to type.

martinbean:
I’m getting an error when trying to override the primary colour.

My *app.scss*:

```scss
@import "custom";
@import "node_modules/bootstrap/scss/bootstrap";
```

My *_custom.scss* file:

```scss
$theme-colors: (
  primary: #c00
);
```

The error I get when trying to compile:

> Error: node_modules/bootstrap/scss/_forms.scss
> Error: argument `$color` of `rgba($color, $alpha)` must be a color
> 
> Backtrace:
> &nbsp;&nbsp;node_modules/bootstrap/scss/_forms.scss:280, in function `rgba`
> &nbsp;&nbsp;node_modules/bootstrap/scss/_forms.scss:280
> &nbsp;&nbsp;&nbsp;&nbsp;on line 280 of node_modules/bootstrap/scss/_forms.scss
>  &gt;&gt;   background-color: rgba($form-feedback-invalid-color,.8);

mdo:
> I'm a designer, and I find the `map-get($theme-colors, "primary")` so cumbersome to write. I use $brand-primary everywhere, and it's a no-brainer to type.

@liquidvisual You don't have to pull from the map at all times—we have all the variables there still. We use the Sass map to reduce overhead and iterate over all our variables to generate every component's themed modifier classes

> How come the implementation demonstrated in the aforementioned Pen not implemented for the beta? As that’s what I followed.

I didn't have time to get this into the beta and this isn't a breaking change. It'll make it into beta 2.

dfinucane:
@martinbean I think you are getting that error because you don't import functions before variables

SirCmpwn:
Having to use map-get everywhere is *awful*. Strong :-1:

martinbean:
> now have no idea how to make basic customizations

This.

Bootstrap used to be so simple to customise. Maps may be a great data structure to hold things like colours of a theme, but they’re not accessible or great for beginners or non-full-time front-end developers to customise.

It’d be great if the top-level `$brand-` variables were brought back and a map used “under the hood” to hold the colours belonging to the theme, but I think the current solution is just going to frustrate and turn away a _lot_ of users.

SirCmpwn:
It definitely turned me away. I reverted back to v4.0.0-alpha.6 and will not be upgrading so long as this is a problem.

Soullivaneuh:
> I didn't have time to get this into the beta and this isn't a breaking change. It'll make it into beta 2.

@mdo You you have any eta for the release of the beta 2?

braginteractive:
So does anyone know how to change the theme colors so it will edit all the elements that use primary, secondary, etc? 

For example, I am looking to change the $primary color, so when I use things like btn-primary it will be my custom color. The following doesn't work:

```
$custom: #123;

$theme-colors: (
  primary: $custom
);
```

Is there another step to override and use the map defaults?

dfinucane:
I have a file I named _custom_bootstrap.scss that contains something like the following
`
@import './bootstrap/scss/functions';

$gray-100: #f8f9fa;
$gray-600: #868e96;
$gray-800: #343a40;
$black:  #000;
$blue:    #0082cc;
$red:     #dc3545;
$yellow:  #ffc107;
$cyan:    #17a2b8;

$theme-colors:(
    primary: $blue,
    secondary: $gray-600,
    success: $blue,
    info: $cyan,
    warning: $yellow,
    danger: $red,
    light: $gray-100,
    dark: $gray-800
);
`
then in my applications styles.scss I have the following
`@import 'custom_bootstrap';

@import 'bootstrap/scss/bootstrap';
`
The thing that helped me understand how this had to work was I realized that what SASS calls variables are really constants in that once they are defined they can't be changed.  So knowing that all you need to do is import your _custom_bootstrap.scss before bootstrap defines those same 'variables'.

braginteractive:
@dfinucane worked perfectly. Thanks

martinbean:
> And it needed the whole map

Yup. This is what I’ve been trying to get over. It’s overkill when all people will want to do is change the brand primary colour.

kartikluke:
@braginteractive This PR is meant to fix the issue you just had. The name is a bit confusing but this is the main part.

> This way, anyone can continue to map-merge in additional or replacement key: value pairs as needed without having to replace the entire original Sass map as @martinbean pointed out in #23112.

philipchristopher:
@liquidvisual You can reduce some of the complexity of the "map-get" function by using bootstrap's "theme-color" function which simply returns the theme-color. Makes it a little less cumbersome.

```css
@import "~bootstrap/scss/functions";

h2 {
  color: theme-color("primary");
  /* color: map-get( $theme-colors, primary); */
}
```

d4rky-pl:
It's still more cumbersome because now my IDE does not autocomplete this which ruins my flow, making me pause to stop and think and triple-check if I didn't make a mistake or typo. 

It was so great to just type `$brand-` and get a full list of available options... :(

martinbean:
@mdo Thanks for making customisation easier again in beta 2! 😘

9peppe:
I believe the base color constants (`$white` to `$cyan`) should probably be separated in a different scss file. So that they can be imported *before* the maps are created to easily define the logical function colors (`$primary` to `$dark`) in customizations. 

In beta2 this is really confusing, because you can easily override variables like `$body-bg` and `$body-color` after `@import "variables";` but you **cannot** effectively do the same with `$primary` and the like.

PR #23286, opened 2017-08-11
Fix SRI hashes.

### LINKED ISSUE #23284: CDN integrity in new beta docs doesn't match
I'm using the cdn from the new documentation but i get an error about integrity not matching.

XhmikosR:
Fixes #23284


mdo (approved):
<3

XhmikosR:
Note that none of the non owners can merge when CI fails.

PR #23301, opened 2017-08-11
Fix HTML errors

XhmikosR:
@Johann-S @mdo @bardiharborow: please approve because non-owners can't merge when CI isn't passing.


Johann-S (approved):
LGTM 👍

XhmikosR:
We really need to switch to html-validator-cli. I'll work next on that because what we have now is pretty bad.

PR #23310, opened 2017-08-11
Replace dependencies badge by peer dependencies badge

### LINKED ISSUE #23282: Add a peerDependencies badge in README.md

Johann-S:
Fixes #23282

/CC @XhmikosR and @mdo


XhmikosR:
Perhaps just adding the new badge and not replacing the dependencies one so that it shows `none` would be better?

Johann-S:
If you prefer like that, why not, personnaly I don't have any advice on what should be better here 🤔

XhmikosR:
No preference either, just wanted to point this out. I guess @mdo will decide

XhmikosR (approved):

mdo (approved):
Given there are no dependencies now, I'm inclined to replace it.

Issue #22762, opened 2017-06-07
Restore icons on valid/invalid form states as optional styles


mdo:
#22395 is removing the icons for now (to limit scope), so we'll need to restore those in a later release. We'll need to tackle some other issues, so I'm consolidating and referencing all of those here. Hoping to get to it in Beta 2, but it might wait to a patch or minor bump.

- [ ] #22670: Icon should be positioned from the top for textareas
- [ ] #22696: Sometimes icons don't appear in inputs in modals in Edge
- [ ] #19852: Custom selects mess up icon size due to custom background-image for the caret

jimjaeger:
@mdo Hi Mark, can you look on the comment in #22762. Can we reopen #22762 or may can you look into as part of this item here?

mdo:
We should also consider replacing the &times; icon with something else, as noted in #23274.

PR #23312, opened 2017-08-11
added icon in form validation in #22762

chiraggmodi:
added form validation icon mention in #22762

### DIFF
--- a/scss/mixins/_forms.scss
+++ b/scss/mixins/_forms.scss
@@ -53,10 +53,12 @@
   }
 
   // Validation icon top for textarea
-  textarea.form-control,
-  textarea.custom-select {
-    &.is-#{$state} {
-      background-position: top calc(#{$input-height-inner} / 4) right calc(#{$input-height-inner} / 4);
+  textarea {
+    &.form-control, 
+    &.custom-select {
+      &.is-#{$state} {
+        background-position: top calc(#{$input-height-inner} / 4) right calc(#{$input-height-inner} / 4);
+      }
     }
   }
 
--- a/scss/mixins/_forms.scss
+++ b/scss/mixins/_forms.scss
@@ -42,7 +42,7 @@
       }
 
       &:focus {
-        box-shadow: 0 0 0 0.2rem rgba($color, 0.25);
+        box-shadow: 0 0 0 .2rem rgba($color,.25);
       }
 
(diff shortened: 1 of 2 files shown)


chiraggmodi:
can any one let me know what is the right way for this **Avoid qualifying class selectors with an element.** validation issue?

scss/mixins/_forms.scss
```diff
@@ -40,6 +52,13 @@
     }
   }
 
+  // Validation icon top for textarea
+  textarea.form-control,
```
chiraggmodi: can any one please guide me on this issue?

XhmikosR:
Doesn't this work?

```scss
element {
  .class {
  }
}
```

scss/mixins/_forms.scss
XhmikosR: Doesn't this work?

```scss
element {
  &.class {
  }
}
```

EDIT: nah, it would still complain.

XhmikosR: See https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#qualifyingelement

chiraggmodi: @XhmikosR  Thank you!!

scss/mixins/_forms.scss
```diff
@@ -10,24 +10,36 @@
...
+      padding-right: $input-btn-padding-x * 3;
+      background-repeat: no-repeat;
+      background-position: center right calc(#{$input-height-inner} / 4);
+      background-size: calc(#{$input-height-inner} / 2) calc(#{$input-height-inner} / 2);
+
+      @if $state == "valid" {
+        background-image: $form-icon-success;
+      }
+
+      @if $state == "invalid" {
```
vsn4ik: Maybe `@else if`...

XhmikosR: Yup definitely good suggestion @vsn4ik

chiraggmodi: @vsn4ik and @XhmikosR made changes as per suggestion.

XhmikosR:
@mdo: what do we do with this? I think we better drop this from a quick look...

mdo:
New PR at #26824.

PR #23314, opened 2017-08-11
Remove IE 10 viewport bug workaround JS.

### LINKED ISSUE #23281: Remove JS from #23122
Follow-up to #23122, gotta remove the JS, too.

XhmikosR:
Followup to #23122.

Fixes #23281


mdo (approved):
Thank you <3

PR #23315, opened 2017-08-11
Fix broken links in documentation.

jankdc:
Repairs 404 links by adding missing `{{ site.baseurl }}/docs/{{ site.docs_version }}/` prefix.


XhmikosR:
@jankdc: are those all of them?

We should really automate this testing maybe with node simple crawler or something. Any PRs welcome.

XhmikosR (approved):

jankdc:
I _think_ that's all of them. That crawler sounds interesting, will look into that.

bardiharborow:
@XhmikosR I used to run a crawler on the docs periodically, but I've been lazy recently. There were thoughts at one point about automating it, but it's probably too much extra CI weight for little gain.

PR #23317, opened 2017-08-11
Make placeholder links optional

jalkoby:
Long story short - after update to Bootstrap beta 4 I've noticed that all `a.btn` has black color text. As I investigated there were discussion in #19402 which come up to neutralize links without href to the plain text style. Also some people mentioned that it breaks the style of their react app so their removed "reboot.scss" from their build. I think removing "reboot" is not a good option as there are many useful things, but make the reset of placeholder links optional(enabled by default) is a perfect compromise for rich web app users

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -204,18 +204,19 @@ a {
 // It would be more straightforward to just use a[href] in previous block, but that
 // causes specificity issues in many other styles that are too complex to fix.
 // See https://github.com/twbs/bootstrap/issues/19402
-
-a:not([href]):not([tabindex]) {
-  color: inherit;
-  text-decoration: none;
-
-  @include hover-focus {
+@if $enable-placeholder-links {
+  a:not([href]):not([tabindex]) {
     color: inherit;
     text-decoration: none;
-  }
 
-  &:focus {
-    outline: 0;
+    @include hover-focus {
+      color: inherit;
+      text-decoration: none;
+    }
+
+    &:focus {
+      outline: 0;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


mdo:
Are there that many `<a>`s out there without `href`s that need the clickable link styling?

jalkoby:
As it was mentioned in the prev discussion any app with react and react router. The truth is the current bootstrap state follows the w3c, but the fact is when you work on a react app with react, angular, etc you need a button like control which partly changes current state without submitting form. So many developers as me tend to use <a> tag

rafalp:
Speaking from experience using href'less a="" to fire the state change in JS app is bad pattern as it breaks middle mouse click - the shortcut for new tab, especially in something routable like react-router.

Having styles letting you mimic links appearance with buttons seems better IMHO. This is why its so good to be able to put buttons in dropdowns and pills in v4. ;)

mdo:
Thoughts from @patrickhlauke maybe?

patrickhlauke:
personally, i'd be hardline and say that if you're using `<a>` without an `href`, you ARE defining an anchor rather than a link and the styling should follow this - and that react should stop abusing this pattern this way. making this a switch that can be enabled if needed sounds like a pragmatic compromise. it feels dirty, but at least it's optional.

andresgalante:
I agree with @patrickhlauke. Plus `a` tags without `href` are not focusable, off the tabindex and not keyboard accessible.

Although @jalkoby is right about React application doing this, I don't like it to feel dirty 😃. I would encourage Bootstrap user to overwrite it on their code base if they want to.

@mdo up to you.

mdo:
Closing out then—we can revisit in the future, though. I haven't seen a huge number of complaints about this yet, so if we don't have to add another option, I'd like to avoid it.

PR #23327, opened 2017-08-11
Switch to html-validator.

### LINKED ISSUE #23171: Switch back to the official vnu.jar for HTML validation
I might have a look myself when I have some time.

Note that alternatively we could use https://github.com/zrrrzzt/html-validator-cli which shouldn't require Java installed.

XhmikosR:
~~This is a first try.~~

~~We could even use [html-validator](https://github.com/zrrrzzt/html-validator) API. I had just started this.~~

Feel free to push to this branch if you have any ideas or want to use the API package which now that I think about it makes more sense.

TODO:

- [x] ~See why ignore doesn't work~ It seems if `verbose` is used then everything is shown, even ignores
- [x] Switch to [html-validator](https://github.com/zrrrzzt/html-validator)
- [ ] See if there's a way to run multiple checks at the same time
- [ ] Handle error codes and return appropriately

Fixes #23171


XhmikosR:
So, anyone willing to help with this? Or switch to html-validator instead of the CLI?

bardiharborow:
Can we write a PR to html-validator-cli to add wildcard support?

XhmikosR:
My main problem is how to handle this upstream or not. Do we make one request with multiple input files? I doubt this is supported. /CC @sideshowbarker

sideshowbarker:
> My main problem is how to handle this upstream or not. Do we make one request with multiple input files? I doubt this is supported. /CC @sideshowbarker

It’s not supported.

To be clear, I mean that the HTML checker network API (or HTTP API or REST API or whatever you want to call it) doesn’t allow checking multiple files during a single request. The API and (servlet) backend are entirely built around checking one document per request.

XhmikosR:
@sideshowbarker: thanks for confirming!

Is there a limit on multiple requests server-side? We could do like 10 requests in parallel to speed things up.

sideshowbarker:
> Is there a limit on multiple requests server-side?

Yes there is, if you make the requests to https://validator.w3.org/nu/. We have rate limiting in place there that will cause your IP address and range to get blocked for some period of time if you send more requests per minute than we allow. And we intentionally don’t document what the threshold is. 

> We could do like 10 requests in parallel to speed things up.

Yeah that’s likely to cause you to hit rate limiting unless you throttle the rate at which you make the requests, which of course defeats the entire purpose.

If you don’t want to hit rate limiting, you can use https://checker.html5.org/ instead, which exposes the same network API—and which receives only a minuscule fraction of the traffic that https://validator.w3.org/nu/ does, so the lack of rate limiting there has never led to any observable problems in the availability of the service.

XhmikosR:
@bardiharborow @Johann-S: any chnace you could help with this PR?

XhmikosR:
I'm gonna close this after discussing the potential solutions with @bardiharborow. We'll go with another solution which will need Java but be optional.

PR #23333, opened 2017-08-11
Fix markdown link formatting on migration page

clarkwinkelmann:
I've not tried to generate the docs locally but the link is currently not rendered on https://getbootstrap.com/docs/4.0/migration/ and I guess it's because of that double parenthesis breaking the markdown format.


clarkwinkelmann:
Didn't search before sending that PR, my bad. It's a duplicate of #23330

PR #23335, opened 2017-08-11
docs: fix Popper.js name and removed comma

FezVrasta:
I'm sorry for being nitpicking 😅

commans before "and" aren't usually needed, and the Popper.js name was misspelled.


mdo:
> commans before "and" aren't usually needed

We subscribe to the serial comma, so it should be left in there :)

FezVrasta:
it really hurts my eyes, btw restored 😇

Issue #23343, opened 2017-08-11
Broken Links on getbootstrap.com


trdwll:
In the footer of the site Examples links to http://getbootstrap.com/examples/ which is a 404 and About links to http://getbootstrap.com/about/history/ which is another 404.

ocelotluke:
This is because the examples & history content is stored on the following links: 

https://v4-alpha.getbootstrap.com/examples/
http://v4-alpha.getbootstrap.com/about/history/

Someone must have changed the links target, but not the file path.

XhmikosR:
Pretty sure this is a duplicate.

trdwll:
@XhmikosR I did a quick search before posting as requested and I didn't find anything. It seems that @StevenBlack has fixed it in the PR above. Which I see that he didn't fix the about link in that PR.

XhmikosR:
About was fixed in https://github.com/twbs/bootstrap/commit/bfd05c3423b8d92721444f002311e7d261ef30a7

And I don't know about his example "fix". Works fine locally.

PR #23348, opened 2017-08-12
Fix broken Examples link in the docs footer include – `footer.html`.

StevenBlack:
Re:  #23343 – **Broken Links on getbootstrap.com**

Clicking the `Examples` link in the home page footer yields 404.

If merged, this makes the footer `Examples` link like the one in docs-navbar.html which is the same source code folder.

Therefore `_includes/footer.html` and `_includes/docs-navbar.html` now use the same link to `Examples`.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -3,7 +3,7 @@
     <ul class="bd-footer-links">
       <li><a href="{{ site.repo }}">GitHub</a></li>
       <li><a href="https://twitter.com/getbootstrap">Twitter</a></li>
-      <li><a href="{{ site.baseurl }}/examples/">Examples</a></li>
+      <li><a href="{{ site.baseurl }}/docs/{{ site.docs_version }}/examples/">Examples</a></li>
       <li><a href="{{ site.baseurl }}/about/">About</a></li>
     </ul>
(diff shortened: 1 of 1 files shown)


XhmikosR:
I don't have an issue with examples locally.

XhmikosR:
BTW, I meant that there is a redirect in place.

StevenBlack:
@XhmikosR  the problem is on the live website at http://getbootstrap.com, on every page that includes the footer.

XhmikosR:
I know what you mean ofc.

I'm just saying, I don't get why this happens since the redirect **is** there.

@mdo: I guess it should hurt to merge. Can you update the gh_pages so that we stop getting such reports?

XhmikosR (approved):

XhmikosR:
I'd still like to know why it happens though since the redirect is there.

StevenBlack:
@XhmikosR my guess: a different `.htaccess` that's, understandably, not in the repo.

XhmikosR:
There is no .htaccess. Everything is hosted on GitHub pages, thus the jekyll-redirect-from plugin.

StevenBlack:
@XhmikosR this is still broken on the live website.

I understand that "docs go with the release" as @mdo said in a related PR, but I wonder if that's wise because, man, this error is all over the current getbootstrap site.

PR #23368, opened 2017-08-12
fix docs modal example fade out

gijsbotje:
the z-index was only corrected for `.modal.show` which made the modal drop behind the header and backdrop when fading out
simply dropping the `.show` from this selector will correct this issue

@XhmikosR

### DIFF
--- a/assets/scss/_component-examples.scss
+++ b/assets/scss/_component-examples.scss
@@ -285,7 +285,7 @@
   }
 }
 
-.modal.show {
+.modal {
   z-index: 1072;
 
   .tooltip, .popover {


Johann-S (approved):

XhmikosR:
LGTM too but since I'm on mobile feel free to merge

coolcat-creations:
I proposed another fix in #23423 because solving it with z-index can cause another issues in custom templates. Just putting it beside the header is not the right thing IMHO.

gijsbotje:
@coolcat-creations this fix is only for the documentation, not for the distributed version
I don't really see what your PR fixes as the backdrop gets injected into the dom when a modal is openend, and deleted from the dom when the modal has closed
so it could never block anything when a modal is closed

coolcat-creations:
@gijsbotje for me it´s not deleted from the DOM when closing the modal?

gijsbotje:
that's weird, it should be @coolcat-creations

coolcat-creations:
http://recordit.co/g2cVlGrxX5

gijsbotje:
@coolcat-creations did you define the `modal-backdrop fade` div yourself? because is think that's your problem
could you make a pen or fiddle?

coolcat-creations:
No, I just used the basic code from the BS4 documentation Button + Modal Window, maybe some complications with other libraries then? I check and get back here.

PR #23375, opened 2017-08-12
Basic release bash script

mdo:
Creates a bash script for me to run to go through our shipping checklist.


XhmikosR:
I guess it's OK as a first step, but personally

1. I'd add it in an npm script too
2. Make it cross-platform

That being said, since you make the releases and works for you, the aforementioned points are low priority :)

build/ship.sh
```diff
@@ -0,0 +1,44 @@
...
+printf "\n${magenta}🚧  Pulling latest changes...\n\n${end}"
+git pull origin v4-dev
+
+# Update version number
+printf "\n${magenta}🚧  Updating version number...\n${end}"
+npm run release-version $current_version $1
+
+# Compile latest CSS and JS
+printf "\n${magenta}🚧  Compile latest CSS and JS...\n${end}"
+npm run dist
```
XhmikosR: Here you need to add a step for docs css and JS until we change our scripts; see #24121

build/ship.sh
```diff
@@ -0,0 +1,44 @@
...
+# Get current version from package.json
+current_version=$(node -p "require('./package.json').version")
+
+if [[ $# -lt 1 ]]; then
+  printf "\n${red}⚠️  Shipping aborted. You must specify a version.\n${end}"
+  exit 1
+fi
+
+# Pulling latest changes, just to be sure
+printf "\n${magenta}🚧  Pulling latest changes...\n\n${end}"
```
XhmikosR: I'd also add a `git reset --hard origin/v4-dev` step so that your local branch is clean.

PR #19468, opened 2016-03-11
Add custom range styles

wolfy1339:
Fixes #17916
I'm open for comments and able to do a rebase.
This was inspired by [m8tro's range style](https://github.com/idleberg/m8tro-bootstrap/blob/master/src/themes/m8tro/input/slider.less) but fully changed to use Bootstrap default colors.

PR #23380, opened 2017-08-13
V4.1: Custom range control

gijsbotje:
This pr adds styling to the range input type conform the BS custom control styling
i would like for people on windows to check this codepen: https://codepen.io/gijsbotje/pen/zddJXL
i've tried my best to get it uniform across browsers but i can't really test windows that good

hope you like it :)

schermafbeelding 2017-08-13 om 03 39 23


wolfy1339:
Please see my PR #19468, I'm sure you can take the styles for Firefox and Chrome for the equivalent of ::-ms-fill-lower

gijsbotje:
@wolfy1339 i tried to make the element look the same accros browsers, that's why in didn't add styling for the progress like you did
only firefox and ie have a way of showing progress

scss/_custom-forms.scss
```diff
@@ -253,3 +253,105 @@
...
+
+  &:focus {
+    outline: none;
+
+    &::-webkit-slider-thumb, &::-moz-range-thumb, &::-ms-thumb {
+      box-shadow: $custom-control-indicator-focus-box-shadow;
+    }
+  }
+
+  &::-moz-focus-inner, &:-moz-focusring {
```
XhmikosR: I think we have one rule per line so try updating your patch.

scss/_custom-forms.scss
```diff
@@ -253,3 +253,110 @@
...
+  }
+
+  // 	IE styling
+  &::-ms-thumb {
+    width: $custom-control-indicator-size;
+    height: $custom-control-indicator-size;
+    background-color: $custom-control-indicator-checked-bg;
+    border: 0;
+    border-radius: $custom-control-indicator-size;
+    -webkit-appearance: none;
```
XhmikosR: I believe this should be unprefixed and let autoprefixer do its thing. The same goes for other prefixes.

gijsbotje:
@XhmikosR fixed it all :)

mdo:
Conflicts resolved and a few more issues being tackled in #25600. Thanks for this, picking it up there.

PR #23384, opened 2017-08-13
Style redirect page

JasonEtco:
Style the redirect page thats visible for a brief second (or more on slower connections). [jekyll-redirect-from](https://github.com/jekyll/jekyll-redirect-from) can use a `redirect.html` layout 🙌 I've used inline styles to save on load time. Here's a screenshot:

image

It ain't a work of art but its more Bootstrappy then the default. Let me know what you think 👍

### DIFF
--- a/_layouts/redirect.html
+++ b/_layouts/redirect.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8">
+    <title>Bootstrap - Content moved</title>
+    <link rel="canonical" href="{{ page.redirect.to }}">
+    <meta http-equiv="refresh" content="0; url={{ page.redirect.to }}">
+    <style>
+      .title { font-size: 2.8rem; margin-top: 1rem; margin-bottom: .5rem; font-weight: 300; }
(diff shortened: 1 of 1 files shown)


XhmikosR (changes requested):
Good idea, but I'm not sure I dig the implementation. Better use our classes to style the page and definitely not inline styles.

JasonEtco:
@XhmikosR Are you sure? The redirect page should be as tiny as possible to limit the amount of extra time needed to actually redirect, so I don't think it should try to load in any CSS (or other unnecessary files). Otherwise users will have to wait for that request while they're still being redirected. I'm not typically an advocate for inline styles, but this seems like a good time for them.

XhmikosR:
@JasonEtco; the styles should be already cached for most users.

XhmikosR:
That being said, I personally don't feel so strong about either solution so if someone else approves this, that's OK.

I mean it just shows up for an instant for the time being, it's not so important. :)

JasonEtco:
I hear you; you're right that in most cases it will be cached but it feels wrong to me to request the whole lib for 2 lines of text (for those few times it won't be cached already). I'll leave this PR to simmer for a bit, if anyone else has thoughts please chime in 👍 I'm happy to link the CSS if thats what we decide to do 🙌

Issue #23390, opened 2017-08-13
Different buttons looks almost identical: "btn-outline-secondary" and "btn-outline-dark disabled"


vaags:
See plunker: https://plnkr.co/edit/hBi8RhSqgtnLtki0K0KS?p=preview

mdo:
Yeah, not much to do about this tbh. I'm inclined to wait it out and see how much of an issue it truly becomes for folks. Yes, we could decrease opacity, but that's a color contrast issue. We could further differentiate the grays, but that doesn't buy us much either.

vaags:
I understand your reasoning. I'm thinking perhaps that btn-outline-secondary looks too much like a disabled button to begin with (and the regular btn-secondary too). For now I've simply stopped using these variants.

PR #23405, opened 2017-08-14
reduce opacity for disable button #23390

chiraggmodi:
Reduce opacity for disable button to differentiate between disable and normal button as per #23390


mdo:
This would break color contrast requirements, thanks though.

Issue #23417, opened 2017-08-14
Broken link for old ver 3.3.7


chiraggmodi:
there is still old version link 3.3.7 is broken
https://getbootstrap.com/3.3.7/

mdo:
Fixed in 96e4efb49.

PR #23421, opened 2017-08-14
Change URL for doc

chiraggmodi:
in doc dropdown v3.3.7 URL is broken. 
live url i [3.3](https://getbootstrap.com/docs/3.3/) while in doc 3.3.7

#23417


mdo:
You have an unrelated change in here with the opacity.

PR #23423, opened 2017-08-14
Update _modal.scss

coolcat-creations:
Adding display:none; to .modal-backdrop and display:block; to .modal-backdrop.show because otherwise the modal backdrop is still there after closing the modal and overlays the site and disables so other navigation elements.

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -68,11 +68,11 @@
   left: 0;
   z-index: $zindex-modal-backdrop;
   background-color: $modal-backdrop-bg;
-  display:none;  
+  display: none; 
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
@coolcat-creations: please fix the errors.

Johann-S:
Sorry but we remove the backdrop in our JS see : https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js#L387
So we have no need to change the `display` of our modal

Johann-S:
Unless you provide a Codepen with a reproductible issue I won't reopen this PR

PR #23424, opened 2017-08-14
Adds a top margin to the jumbotron

patkub:
This PR adds a top margin to the jumbotron component. This makes the top and bottom margins equal.


mdo:
We try to only specify `margin-bottom`. Thanks, though!

PR #23439, opened 2017-08-14
Changes badge-pill-border-radius variable from 10px to 50 percent to ensure ro…

andresgalante:
Hi! Variables.scss reads:

```scss
// Use a higher than normal value to ensure completely rounded edges when
// customizing padding or font-size on labels.
$badge-pill-border-radius:    10rem !default;
```
If we want to ensure completely rounded edges, that value should be 50%. This PR changes it to:

```scss
$badge-pill-border-radius:    50% !default;
```

What do you think?


mdo:
Doesn't work like that: https://codepen.io/anon/pen/vJezWQ

PR #23440, opened 2017-08-14
Add contrast colors Sass map

### LINKED ISSUE #23418: Cannot customize Button foreground color
In the new beta the `color` style on buttons are calculated based on the background color to be either `#111` or `#fff`, I have NO idea why are this limited now? It's a huge limitation when compared to the earlier versions.

In all other versions we have been able to set the exact `color` by changing the LESS (SASS in v4 alpha) variable which was an easy way to get it exactly how we needs it, now we will have to write our own .btn-* overrides to achieve the same with risk of loosing backwardscompatiblity later.

Please bring back the ability to set `color` using variables as before. The new system are really frustrating to work with.

jacobmllr95:
This commit adds a `$contrast-colors` Sass map to the variables. This makes it possible to define a custom contrast color for a given color. Closes #23418.

Let's say someone wants the color for `btn-secondary` to be black:
```sass
$contrast-colors: (
  $gray-600: $black
);
```
bootstrap-v4-beta-btn-secondary-black-color

Or if you want to restore the Bootstrap v4 Alpha colors:
```sass
$blue:   #0275d8;
$green:  #5cb85c;
$teal:   #5bc0de;
$yellow: #ffd500;
$red:    #d9534f;

$theme-colors: (
  secondary: $white
);

$contrast-colors: (
  $blue: $white,
  $white: $gray-600,
  $green: $white,
  $cyan: $white,
  $yellow: $white,
  $red: $white
);
```
bootstrap-v4-alpha-button-syles


mdo:
Closing for a simpler approach of adding vars for these two colors: #24426. It's not the per-theme-color approach as here, but it provides some flexibility.

Issue #23349, opened 2017-08-12
Proposal: CSS Variables read-only reflection of some global Sass variables


passcod:
I tried to search, but turns out searching for "css variables" or even "custom properties" isn't particularly helpful. In any case I did go through a bunch of issues and didn't see this mentioned. Thus:

Background: situations where adding Sass to the build pipeline is not feasible or not wanted, but one still wants to use Bootstrap values (e.g. colours, sizes, timings) in custom CSS.

Proposal: to add a `:root {}` section in the _default_ generated CSS that writes out something like:

```css
:root {
  --blue: $blue;
  --indigo: $indigo;
  --purple: $purple;
  --pink: $pink;
  /* etc */
}
```

Usage would then look like:

```css
.custom-component {
  box-shadow: var(--dark-gray);
  color: var(--indigo);
}
```

That would either be [directly supported by browsers](https://caniuse.com/#feat=css-variables) (except IE, Edge, some mobile browsers) or more likely right now compiled-in by the build tool (e.g. PostCSS plugin [`postcss-custom-properties`](https://github.com/postcss/postcss-custom-properties)). Because this is supported right now in modern / dev-used browsers, it also brings a newish component to live editing via the Inspector.

Compatibility: in the event that someone is already using custom properties named the same as the bootstrap variables, then this would only affect them if the bootstrap ones come _after_ their own (otherwise the cascade would take care of it). Apart from that, I don't see any other compatibility problems, but I might have overlooked some.

mdo:
Interesting idea! Could be cool to do.

mdo:
Doing something like this:

```css
:root {
  @each $color, $value in $theme-colors {
    --#{$color}: $value;
  }
}
```

Could output this:

```css
:root {
  --primary: #007bff;
  --secondary: #868e96;
  --success: #28a745;
  --info: #17a2b8;
  --warning: #ffc107;
  --danger: #dc3545;
  --light: #f8f9fa;
  --dark: #343a40;
}
```

mdo:
Tried a PR over at #23446.

passcod:
I was thinking about this and I think a good minimal base could be:

- The `$theme-colors` as in the PR
- The `$colors` to have the full bootstrap-picked palette
- The `$grid-breakpoints` to use in media queries
- The `$font-family-*` stacks

The many other variables could also be exposed as an option, but I don't think most of them would be as useful as those ones, especially since a lot of them are derived from these base ones.

[...]

PR #23446, opened 2017-08-15
Generate CSS variables for -colors

mdo:
Interesting idea proposed by #23349. We could build with Sass, but compile some utilities to help those not using our Sass-based compilation flow. I'd say we could even wrap these in conditionals to not include them unless asked for, but that'd defeat the purpose a bit.

I dunno how far we'd take this though...


passcod:
Yay! Okay, this works great. I've tried it with postcss-custom-properties for non-Sass build usage, and also directly in the browser. Here's a GIF of the latter:

bootstrapvars

PR #23448, opened 2017-08-15
Beautify _config.yml.

XhmikosR:
See the non-whitespace diff https://github.com/twbs/bootstrap/pull/23448/files?w=1.


mdo (approved):
Nice, thanks!

PR #23459, opened 2017-08-15
Add css-mqpacker for the minified files only.

XhmikosR:
Without css-mqpacker:

```
bootstrap-grid.min.css      23.937
bootstrap-reboot.min.css    3.730
bootstrap.min.css           125.440
docs.min.css                23.082
```

With css-mqpacker:

```
bootstrap-grid.min.css      23.727
bootstrap-reboot.min.css    3.730
bootstrap.min.css           124.362
docs.min.css                22.406
```


XhmikosR:
OK, the way we have our scripts, i'm not sure how to make this apply only to minified files; it works for docs.min.css since that is the filename we output from postcss, but for the rest we use clean-css.

So we need to run postcss with a different config before running clean-css.

XhmikosR:
@Johann-S: so any suggestions how to handle this? We run postcss against all CSS files including minified ones. I think I will try to split this so that it runs only for non-minified files and just change the clean-css script to run postcss first or something.

Johann-S:
Maybe a node script would be better instead of splitting the calls to postcss for each of our minified files 🤔

mdo:
That's not a lot of file size savings. I don't think we really need this given it potentially changes our output in a significant way from what others could expect.

XhmikosR:
It is definitely worth it. And like I said I will make it run only for the
minified files.

mdo:
> It is definitely worth it.

How? It saves less than 1kb in each of those bundles, at the expense of adding another dependency. I don't see an upside to adding more code on our end to save so little.

XhmikosR:
You think 1 kb less from the minified file isn't good enough?
As far as the dependency come on man this is just fine here. :) Eventually
I plan to add more stuff to improve the minified build size.

bangzek:
Are you sure it is safe to use with bootstrap css?

The [First Win](https://github.com/hail2u/node-css-mqpacker#the-first-win-algorithm) algorithm could break havoc on ruleset order if we're not carefull.

XhmikosR:
I'm using it for months in personal projects without a hiccup.

bangzek:
Because how massive the changes to final css will be, I suggest this is done on the normal, non minified css.

I don't want to have a big difference in rulesets order between normal and minified css.

Another thing maybe add in the beginning of sass file a series of empty media query in the order we prefer, so we will not be catch unaware with the First Order algorithm.

mdo:
Yeah, potential issues with CSS being moved around is always of concern to me. While we try to limit issues from the order of CSS, there's always something. Would definitely be good to see a diff of the file after running it on `bootstrap.css`.

bangzek:
@mdo they are quite large, see it in [my fork](/bangzek/bootstrap/commit/ac19d93d66ccf883c55e32450a975d801f2bfd72)

XhmikosR:
I update the description with the final findings using css-mqpacker.

XhmikosR:
...and the gzip size is larger for bootstrap.min.css: 18.83KB vs 19.02KB.

So I'd say this is not worth doing.

PR #23468, opened 2017-08-15
Convert px values to rem on tooltip variables

andresgalante:
To help move Bootstrap units to rems this PR changes tooltip variables for maxwidth and paddings from px to rem.

what was `8px` is now `.5rem` (= 8px with a 16px base)
what was `3px` is now `.2rem` (= 3.2px with a 16px base)
what was `200px` is now `20rem` (= 192px with a 16px base)

What do you think?


mdo:
Are there any rendering issues with the partial pixel value of `3.2px` there? Tooltips and popovers have historically had potential to be blurry or placed on subpixels is why I ask.

andresgalante:
@mdo let me check in different browsers and send you a couple of images. We can dial it up to 2.5rem which should compute to 4px, what do you think?

andresgalante:
Sarafi has a slightly difference, but it doesn't matter if it is `.2rem`, `.25rem` or even `3px` it always shows that. Is as if the font gets lighter. This is an image of it in my non retina display, I get the same difference always:

screen shot 2017-08-15 at 2 57 28 pm

Chrome is ok and Firefox is perfect, I couldn't test on Edge or IE.

andresgalante:
I've updated the value to `2.5rem`  (= 4px with a 16px base) and rebased the branch.

mdo:
Should `$tooltip-arrow-width` and $tooltip-arrow-height` be updated as well?

```scss
$tooltip-arrow-width: .4rem !default;
$tooltip-arrow-height: $tooltip-arrow-width !default;
```

NielsHolt:
Hi
I think it is a good idea to convert from px to rem but I think that changing $tooltip-arrow-width from px to rem will introduce an error in css for the arrow In scss/tooltip.scss :        
`margin-left: -($tooltip-arrow-width - 2);`

Both in scss/tooltip.scss and scss/popover.scss there are several calculations a la 
```
    .arrow::after {
      bottom: -($popover-arrow-outer-width - 1);
      margin-left: -($popover-arrow-outer-width - 5);
      border-top-color: $popover-arrow-color;
    }

Where my quess is that `1` is `default border-width` and `5` is default `$popover-arrow-height` or half default `$popover-arrow-width`

I will look into it and get back to you




```

andresgalante:
@NielsHolt is right, having `$tooltip-arrow-width` in rems puts the arrows all over the place.

I've changed it back to `5px` and I'll review the calcs.

Thanks a lot @NielsHolt for reviewing it

andresgalante:
Hi @mdo and @NielsHolt I've made the adjustments so we can now use `$tooltip-arrow-width` in ems

They look like this:
  
screen shot 2017-08-29 at 9 26 06 pm

What do you think?

NielsHolt:
Hi @andresgalante 
It looks very nice :-)
I have tried to convert popover from `px `to `rem `(as you) but also changing `$border-width` from `px` to `rem` and include `$popover-border-width` in the css for popover
I will try to send a PR soon

andresgalante:
Hi @NielsHolt I wouldn't do that, borders should be measured in pixels.

NielsHolt:
Hi @andresgalante 
I have spend some (too much) time looking into the popover-array.
It seems that there is a bug in Bootstrap. It doesn't account for the `$border-width` and the placement of the arrow is hard-coded to  `border-width: 1`
If you try setting `$border-width: 10px` and build the css the placement of the popover-box and the arrow is wrong.
It looks like the `.arrow` element isn't placed 100% correct. If I increase the dim of the element with 2*the 
border-width of the popover the placement get right
I have made a solution (with border-width in px :-)) that seems to fix it and allows for arrow-width in rem
Perhaps it can also solve the issues with tooltips
My fork of Bootstrap has a branch "popover-in-rem" with the my solution. I will make a PR to Bootstrap later this weekend

andresgalante:
@NielsHolt I am lost, in here #23776 the position of the arrow is "arrow-width-1px" it just works

NielsHolt:
Hi @andresgalante 
I think I figured it out. All the `$arrow-width - 1` and `$arrow-width - 5` is hard coded based on fixed border-width and arrow-width. If you change any of the dimensions in popover or tooltip the arrow are placed wrong. The same happen when the dimensions is changed from px to rem
But I think I have figured it out and I will send a PR to Bootstrap asap
Thank you for your help and input :-)

andresgalante:
@NielsHolt Man, you have me completely lost. My solution #23776  works perfectly fine even if I change the size of the box. I don't see where the problem is.

NielsHolt:
Hi @andresgalante 
The issue is that the width of the element's and the popover's border must be taken into calculation
All the `- 1` in the scss are hard coded assuming that `border-width: 1px`

If you in Bootstrap 4.0.0-beta set 
```
$border-width: 20px;
$popover-border-width:10px;
```
You will get
bootstrap-popover-4-beta

If you use #23468 and set `$popover-arrow-width: 1rem;` you get
bootstrap-popover-galante

If you use #23820 you get
bootstrap-popover-nielsholt

andresgalante:
@NielsHolt ok, I see what you mean. thanks for explaining it to me 😄

mdo:
Slating for a review in Beta 3.

Issue #23476, opened 2017-08-15
Proposal: the way to apply fix for SVG images w/ .img-fluid in IE 9-10 automatically


juliyvchirkov:
as declared at https://getbootstrap.com/docs/4.0/content/images/ IE 9-10 sizes SVG images w/ .img-fluid disproportionately, but no fix is applied automatically 'cause the offered workaround fixes responsivity for SVG images on one hand but breaks it for other image formats on another

![2017-08-16 00 10 56](https://user-images.githubusercontent.com/12705661/29336986-6c7069e0-8218-11e7-983f-a784de0c8519.png)

the proposal is to add

```
.img-fluid[src$=.svg] {
    width: 100% \9;
}
```

since the fix «width: 100% \9» is targeted to IE & ignored as malformed by other browsers, & the selector «src$=.svg» forces the rule to be applied only if image url in src attribute has «svg» extension, this should achieve the goal

chiraggmodi:
we have dropped support for IE 9, so I don't think it is necessary to add

juliyvchirkov:
yep, as I see IE9 support has been dropped, but beta 4.0 still declares the support for IE10 declared https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/
 
![2017-08-16 14 09 31](https://user-images.githubusercontent.com/12705661/29360675-be0848bc-828c-11e7-8c82-b79a132364d7.png)

so since the IE10 inherits the bug on SVG proportions w/ .img-fluid from v9 while percent of SVG images is widely growing nowadays, maybe there's a reason to apply the workaround for v10 till the support for IE will be dropped completely

Herst:
IE9 was dropped in December (see #21416) but IE10 was not; though people were told to _"keep the feedback coming in"_.

mdo:
No plans to add this automatically at this point. Have heard very few issues about this one since originally opening and don't want to cause more confusion by making it the default now. Thanks though!

PR #23483, opened 2017-08-16
include img-fluid width 100% for svg files #23476

chiraggmodi:
have added `width:100%` for SVG image src as per reference in #23476


mdo:
Closing per #23476.

Issue #23460, opened 2017-08-15
Disabled cursor on form labels not displaying correctly


lowegreg:
My apologies if this should be a feature request and not an issue.  I noticed when adding a disabled form control the label still uses the pointer cursor instead of the not-allowed cursor.  I tested this on both version 3 and 4.

Cheers,
Greg

mdo:
This isn't something we can change with CSS—there's no way for the label to know if the input is disabled.

PR #23485, opened 2017-08-16
Added not allowed cursor in disable form

chiraggmodi:
as per suggestion in #23460 i have added not allowed cursor.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -28,6 +28,7 @@
   // Disabled comes first so active can properly restyle
   &.disabled,
   &:disabled {
+    cursor: not-allowed;
     opacity: .65;
     @include box-shadow(none);
   }
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -58,6 +58,9 @@
     // iOS fix for unreadable disabled content; see https://github.com/twbs/bootstrap/issues/11655.
     opacity: 1;
   }
+  &:disabled {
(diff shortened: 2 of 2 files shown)


patrickhlauke (commented):
I'm personally not a fan of `cursor: not-allowed` for aesthetic reasons (on windows, it results in an ugly red "not allowed" sign which looks rather jarring), and because default native forms don't have this sort of indication on disabled form controls. leaving it up to @mdo to make the call (but since we removed that particular cursor in a previous PR, this would mean reintroducing it / making sure it's consistently re-added)

lowegreg:
I agree with @patrickhlauke and also not a fan of the cursor: not-allowed.  I only added it as an issue because the control had it.  In my opinion regardless of which way you go, the control and label should have the same cursor...

Greg

mdo:
We'll pass on it for now then given @patrickhlauke's comment. Thanks though!

danjford:
Just adding my experience on this. I did some UX testing with some partially sighted users who used a mixture of screen readers, braille keyboards and screen magnifiers.

I believe it was the users who were using screen magnifiers who gave the feedback that they found the not-allowed image on windows particularly useful as the opacity did not help them to distinguish that something was disabled.

chiraggmodi:
i agree, `cursor: not-allowed` is useful as UX point of view. Believe me i am also not fan of `cursor: not-allowed` property.

PR #23496, opened 2017-08-16
allow default font family control

michaelreynolds:
Allow users to override and change default `font family`. 

Without this, to use a custom font, users must manually override default `font-family: san-serif` in each and every place were `font-family` is not specified within a selector declaration.

TDLR; This changes nothing except allows users to control default font families without tons of work.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -25,7 +25,7 @@
 }
 
 html {
-  font-family: sans-serif; // 2
+  font-family: $font-family-base; // 2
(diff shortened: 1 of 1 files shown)


mdo:
> Without this, to use a custom font, users must manually override default font-family: san-serif in each and every place were font-family is not specified within a selector declaration.

How's that? `font-family` is inherited in nearly every other place, and it's usually inherited from the `<body>` or an element's nearest parent.

mdo:
Closing as a won't fix.

PR #23503, opened 2017-08-16
(Fixes #23449) Normalize Custom Checkbox and Radio with Form Check

### LINKED ISSUE #23449: Normalize Custom Checkbox and Radio with Form Check
What do you think about normalizing the markup in custom checkboxes and radios to match form-check (https://github.com/twbs/bootstrap/pull/23444) ? Something like:

```
<div class="custom-radio">
    <label class="custom-control">
        <input class="custom-control-input" type="radio">
        <span class="custom-control-indicator"></span>
        <span class="custom-control-description"></span>
    </label>
</div>
```

have it stacked by default and have helper classes to inline them? I think it makes sense to keep the pattern consistent for both. It would also prevent `label.custom-control` from expanding the full width of the container with `.custom-controls-stacked` (https://getbootstrap.com/docs/4.0/components/forms/#stacked).

pat270:
Fixes #23449.


mdo:
Merged this into #23444.

Issue #23506, opened 2017-08-17
Carousel Documentation contains broken Link (B4)


herbalite:
Navigate to "Methods" on the carousel page => https://getbootstrap.com/docs/4.0/components/carousel/#methods

Broken Link has the following text=> "See our Javascript documentation for more informations.

Link points to => https://getbootstrap.com/getting-started/javascript/

Link probably should point to => https://getbootstrap.com/docs/4.0/getting-started/javascript/

XhmikosR:
This is already fixed.

shamotufei:
shaziJ8

PR #23508, opened 2017-08-17
fix broken link in footer

chiraggmodi:
there was a broken link in footer for about menu. have fixed that. also check #23506 this issue but it is not in latest ver.


XhmikosR:
This is not needed.

shamotufei:
shazi  J8

PR #23510, opened 2017-08-17
Added `@content` directive to `media-breakpoint-only` mixin for min and max cases

AlexisVK:
In current implementation `@content` directive is missed in mixin `media-breakpoint-only` for `$max == null` and `$min == null` cases


mdo (approved):

mdo:
Fixed #23516.

PR #23519, opened 2017-08-17
esc should not close modal when child dropdown is open

### LINKED ISSUE #23507: Esc key closes modal and dropdown if dropdown is open inside it
When I have a modal and in the model I have a DropDown.  If I click the DropDown and it opens clicking ESC doesn't close de DropDown but the entire modal.

If it's by design I don't understand.  ESC should check if a DropDown is opened and if so closing it then if there's nothing else opened the second ESC should close de dialog

This happens right now in 3.3.7 but I want to know if this changed in 4 or if this will remain like that ?

For instance : I pick a DropDown but it was not the right one.  Now I have to click somewhere else with the mouse but many people uses TAB to go from one control to the next

bkdotcom:
fix #23507


Johann-S:
Seems fine to me, any advice @patrickhlauke ?

patrickhlauke:
not checked this live, but a perhaps naive question: should it not be fixed at the dropdown script? i.e. that the dropdown should swallow the ESC and `preventDefault()` / cancel the event / stop the bubbling up to the modal? ditto in case of alerts that may be open when the modal is open etc.

Johann-S:
I totally agree with @patrickhlauke this use case should be handled by our Dropdown instead of our Modal plugins, thanks @patrickhlauke 👍 

maybe @bkdotcom can you change that in this PR or create a new one ?

bkdotcom:
@Johann-S : der, that makes sense.    I'll try to get to it today

bkdotcom:
Turns out removing this logic from the modal plugin (as it should be) is a bit more complex than I've got time to tackle right now...

**Currently:**
open modal → listen for esc on modal element → open dropdown → listen for esc on dropdown element

press esc:   modal gets escape event first

**crazy out of the box idea:**
A single bootstrap-wide always-on document keydown-escape listener (Utils?)
open modal → push closure to escape-handler-stack → open dropdown → push closure to escape handler-stack

press escape:  
pop closure off of stack and call it.
preventDefault depending on return value of the closure(s)

if a modal containing an open dropdown is closed...   will that trigger a hide.bs.dropdown?
trash-collection would need to be preformed on the stack?
or perhaps say the dropdown escape closure is called but the dropdown is already closed/:hidden (no-trash-collection)... closure could return `false`... telling the escape handler to pop-and-run the next closure on the stack?

PR #23520, opened 2017-08-17
Fix dropup in navbar and the caret position accordingly 

### LINKED ISSUE #22786: Navbar collapsed dropup caret should point down

### LINKED ISSUE #23429: Dropdown (dropup) in bottom fixed navbar opens down
Either a bug or something I'm missing... When having a dropdown set to dropup in a bottom navbar, it opens the dropdown down instead of up (probably because the container limits the opening to the top). FYI it was working in alpha 6, the bug appeared in beta 1

See the example here: https://plnkr.co/edit/7iaSiawpvF6yqiL3DxGC?p=preview

Johann-S:
We choose to disable Popper.js style for Dropdown in navbar but we didn't handle the case when we have a dropup in a bottom navbar and the change of the dropdown caret

- [x] Add documentation example

Fixes #23429
Fixes #22786


mdo:
Can you take a stab at adding a docs demo for this with a fixed bottom navbar?

Johann-S:
Sure I'll take time to give some love to this PR 😄

mdo:
Also, what about #22786?

Johann-S:
I added an example :wink: I'll take a look at this issue

Johann-S:
@mdo this PR fix now the issue you mentionned 👍 I think I've done my work on this PR if you have time to review my CSS before merging this one that's would be great 😄

Johann-S:
If you found something wrong here in my CSS @mdo and @andresgalante feel free to correct it :smile: I'll merge this PR for now

XhmikosR:
@Johann-S: the dropup in the example still opens to the bottom.

Johann-S:
I tested that a few minutes ago and the dropup in the example opens to the top :thinking:

Johann-S:
Hmm no sorry I just checked an it opens to the top : 

image


Make sur you checked on that URL : http://localhost:9001/docs/4.0/examples/navbar-bottom/

XhmikosR:
OK, it's because of the dist files. It seems to work fine.

XhmikosR (approved):

scss/_navbar.scss
```diff
@@ -74,6 +74,13 @@
     position: static;
     float: none;
   }
+
+  .dropdown-toggle {
```
wojtask9: For me this change is unnecessary. Why we need redefine `.dropdown-toggle` in .navbar?

Johann-S: because without that we cannot have the up caret

wojtask9: I just tested and works without these lines.
Still I don't understand why this is needed :)

scss/_navbar.scss
```diff
@@ -144,6 +151,18 @@
           padding-right: 0;
           padding-left: 0;
         }
+
+        .dropup {
```
wojtask9: Sorry @Johann-S  but here is another issue.
This should be moved to `media-breakpoint-up` because currently it only works with `.navbar-expand` but not with `.navbar-expand-{sm,md}`

Johann-S: you're right ! Can you open an issue which explain that ? Or a PR if you want to fix that ? :+1:

PR #23522, opened 2017-08-17
Fix error in Alerts component doc

drskullster:
In the examples introductory text, the sentence _use one of the four required contextual classes_ is wrong now that there are more contextual classes for alerts.

Could we use twig to do the count dynamically ? something like `{{ site.data.theme-colors | length }}`. Maybe I'm getting ahead of myself :)

### DIFF
--- a/docs/4.0/components/alerts.md
+++ b/docs/4.0/components/alerts.md
@@ -8,7 +8,7 @@ toc: true
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
You had a typo, I fixed it in your branch.

XhmikosR (approved):

Issue #23523, opened 2017-08-17
Carousel: stop cycling if there are other carousels in page and are not visible


XhmikosR:
Say, one has 3 carousels in the same page with autoplay on. An example would be 3 tabs with a carousel in each tab.

We shouldn't cycle any carousel that is not currently visible.

/CC @Johann-S @mdo @bardiharborow for thoughts

XhmikosR:
Definitely need to mention it in the release notes.

BTW I haven't been adding the PRs I merge in the release notes. I will from now on, I promise :)

But if we could add everything that's missing that would help.

PR #23524, opened 2017-08-17
Do not call next when the carousel or the parent isn't visible

Johann-S:
- [x] Unit test to be sure the carousel don't cycle when we cannot see it
- [x] Unit test to be sure the carousel don't cycle when is parent isn't visible

~Work in progress...~

Close : #23523

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -131,7 +131,9 @@ const Carousel = (($) => {
 
     nextWhenVisible() {
       // Don't call next when the page isn't visible
-      if (!document.hidden) {
+      // or the carousel or it's parent isn't visible
+      if (!document.hidden &&
+        ($(this._element).is(':visible') && $(this._element).css('visibility') !== 'hidden')) {
         this.next()
       }
     }
--- a/js/tests/unit/carousel.js
+++ b/js/tests/unit/carousel.js
@@ -879,4 +879,43 @@ $(function () {
       }, 80)
     }, 80)
   })
+
+  QUnit.test('Should not go to the next item when the parent of the carousel is not visible', function (assert) {
+    assert.expect(2)
+    var done = assert.async()
+    var html = '<div id="parent" style="display: none;">'
+             + '  <div id="myCarousel" class="carousel slide" data-interval="50" style="display: none;">'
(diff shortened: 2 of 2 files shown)


js/src/carousel.js
```diff
@@ -131,7 +131,9 @@ const Carousel = (($) => {
 
     nextWhenVisible() {
       // Don't call next when the page isn't visible
-      if (!document.hidden) {
+      // or the carousel or it's parent isn't visible
+      if (!document.hidden &&
+        ($(this._element).is(':visible') && $(this._element).css('visibility') !== 'hidden')) {
```
XhmikosR: Are you sure those 2 checks aren't identical?

Johann-S: Yes the first one just check if the parent or the carousel don't have `display: none;`

Johann-S: For more informations and precise explanations (with a better English 😆) : https://api.jquery.com/hidden-selector/

XhmikosR (approved):

XhmikosR:
So we just need a test with both cases and we are all set. Thanks for the fast patch!

Johann-S:
I added the two unit tests so @XhmikosR if you're available for a quick code review 😄

js/tests/unit/carousel.js
```diff
@@ -879,4 +879,43 @@ $(function () {
...
+    $html.appendTo('#qunit-fixture')
+    var $parent = $html.find('#parent')
+    var $carousel = $html.find('#myCarousel')
+    $carousel.bootstrapCarousel()
+    var $firstItem = $('#firstItem')
+
+    setTimeout(function () {
+      assert.ok($firstItem.hasClass('active'))
+      $carousel.bootstrapCarousel('dispose')
+      $parent.attr('style', 'visibility: hidden;')
```
XhmikosR: This add `visibility` but doesn't remove `display:none`, right?

Johann-S: It remove the `display: none;` because we set the attribute value no matter what we had before that

XhmikosR: OK, looks good to me then. If you can't think of anything else, please merge :)

PR #23533, opened 2017-08-17
Pre-cache key assets with Workboxjs.

MikeyBeLike:
First time attempting to contribute, so bare with me please if i've missed anything or done anything wrong.

But I noticed this was merged ( #23101 ) so in addition, thought it would make sense to pre-cache those main assets.

Also not too sure how it works but I think Jekyll is serving/parsing the markdown doc files dynamically. It would be a cool idea if we could maybe get them to static html files? That way we can have offline support with this too!

Screenshot to show speed difference:
https://snag.gy/KzD0Rv.jpg - throttled slow 3G
https://snag.gy/YHsyzB.jpg - throttled slow 3G w/ workboxjs

Also ran `grunt test` and everything passed fine.

### DIFF
--- a/assets/js/src/pwa.js
+++ b/assets/js/src/pwa.js
@@ -7,6 +7,21 @@
     window.addEventListener('load', function () {
       navigator.serviceWorker.register('/sw.js').then(function (registration) {
         console.log('ServiceWorker registration successful with scope: ', registration.scope)
+        registration.onupdatefound = function () {
+          var installingWorker = registration.installing
+          installingWorker.onstatechange = function () {
+            switch (installingWorker.state) {
+              case 'installed':
+                if (navigator.serviceWorker.controller) {
+                  console.log('new update available')
+                  location.reload(true)
+                }
+                break
+
+              default:
+            }
+          }
+        }
       }).catch(function (err) {
         console.log('ServiceWorker registration failed: ', err)
       })
--- a/build/workbox.config.json
+++ b/build/workbox.config.json
@@ -1,0 +1,8 @@
+{
+  "globDirectory": "./",
+  "globPatterns": [
+    "_gh_pages/**/*.{html,css,js,json,png,jpg}"
+  ],
+  "swSrc": "./sw.js",
+  "swDest": "./_gh_pages/sw.js"
+}
--- a/build/workbox.js
+++ b/build/workbox.js
@@ -1,0 +1,37 @@
+const fs = require('fs')
+const path = require('path')
+const swBuild = require('workbox-build')
+const config = require('./workbox.config.json')
+const buildPrefix = '_gh_pages/'
+
+const workboxSWSrcPath = require.resolve('workbox-sw')
+const wbFileName = path.basename(workboxSWSrcPath)
+const workboxSWDestPath = buildPrefix + 'assets/js/vendor/' + wbFileName
+const workboxSWSrcMapPath = `${workboxSWSrcPath}.map`
+const workboxSWDestMapPath = `${workboxSWDestPath}.map`
+
+fs.createReadStream(workboxSWSrcPath).pipe(fs.createWriteStream(workboxSWDestPath))
+fs.createReadStream(workboxSWSrcMapPath).pipe(fs.createWriteStream(workboxSWDestMapPath))
+
+const updateUrl = (manifestEntries) => manifestEntries.map((entry) => {
+  if (entry.url.startsWith(buildPrefix)) {
+    const regex = new RegExp(buildPrefix, 'g')
+    entry.url = entry.url.replace(regex, '')
+  }
+  return entry
+})
+
+config.manifestTransforms = [updateUrl]
+
+swBuild.injectManifest(config).then(() => {
+  const wbSwRegex = /{fileName}/g
+  fs.readFile(config.swDest, 'utf8', (err, data) => {
+    if (err) {
+      throw err
+    }
+    const swFileContents = data.replace(wbSwRegex, wbFileName)
+    fs.writeFile(config.swDest, swFileContents, () => {
+      console.log('Pre-cache Manifest generated.')
+    })
+  })
+})
--- a/package.json
+++ b/package.json
@@ -44,10 +44,11 @@
     "js-test-cloud": "ruby -r webrick -e \"s = WEBrick::HTTPServer.new(:Port => 3000, :DocumentRoot => Dir.pwd, :Logger => WEBrick::Log.new('/dev/null'), :AccessLog => []); trap('INT') { s.shutdown }; s.start\" & grunt saucelabs-qunit",
     "docs": "npm-run-all docs-compile docs-lint",
     "docs-lint": "htmllint --rc build/.htmllintrc _gh_pages/*.html _gh_pages/**/*.html js/tests/visual/*.html",
-    "docs-compile": "bundle exec jekyll build",
+    "docs-compile": "bundle exec jekyll build && npm run docs-workbox-precache",
     "docs-serve": "bundle exec jekyll serve",
     "docs-github": "shx echo \"github: true\" > twbsconfig.yml && npm run docs-compile -- --config _config.yml,twbsconfig.yml && shx rm ./twbsconfig.yml",
     "docs-upload-preview": "build/upload-preview.sh",
+    "docs-workbox-precache": "node build/workbox.js",
     "maintenance-dependencies": "ncu -a -x jquery && npm update && bundle update && shx echo 'Manually update assets/js/vendor/*, js/tests/vendor/* and .travis.yml'",
     "release-version": "node build/change-version.js",
     "release-zip": "cd dist/ && zip -r9 bootstrap-$npm_package_version-dist.zip * && shx mv bootstrap-$npm_package_version-dist.zip ..",
@@ -99,7 +100,8 @@
     "rollup-plugin-node-resolve": "^3.0.0",
     "shelljs": "^0.7.8",
     "shx": "^0.2.2",
-    "uglify-js": "^3.0.24"
+    "uglify-js": "^3.0.24",
+    "workbox-build": "^2.0.1"
   },
   "engines": {
     "node": ">=4"
--- a/sw.js
+++ b/sw.js
@@ -1,1 +1,4 @@
-// empty for now
+self.importScripts('/assets/js/vendor/{fileName}')
+
+const workboxSW = new self.WorkboxSW()
+workboxSW.precache([])
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "bootstrap",
   "version": "4.0.0-beta",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "abbrev": {
       "version": "1.1.0",
@@ -37,6 +38,10 @@
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-2.1.1.tgz",
       "integrity": "sha1-1t4Q1a9hMtW9aSQn1G/FOFOQlMc=",
       "dev": true,
+      "requires": {
+        "extend": "3.0.1",
+        "semver": "5.0.3"
+      },
       "dependencies": {
         "semver": {
           "version": "5.0.3",
@@ -74,7 +79,10 @@
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/ansi-align/-/ansi-align-2.0.0.tgz",
       "integrity": "sha1-w2rsy6VjuJzrVW82kPCx2eNUf38=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "string-width": "2.1.1"
+      }
     },
     "ansi-escapes": {
       "version": "2.0.0",
@@ -101,7 +109,11 @@
       "version": "1.3.2",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-1.3.2.tgz",
       "integrity": "sha512-0XNayC8lTHQ2OI8aljNCN3sSx6hsr/1+rlcDAotXJR7C1oZZHCNsfpbKwMjRA3Uqb5tF1Rae2oloTr4xpq+WjA==",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "micromatch": "2.3.11",
+        "normalize-path": "2.1.1"
+      }
     },
     "aproba": {
       "version": "1.1.2",
@@ -113,19 +125,29 @@
       "version": "1.1.4",
       "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.4.tgz",
       "integrity": "sha1-u13KOCu5TwXhUZQ3PRb9O6HKEQ0=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "delegates": "1.0.0",
+        "readable-stream": "2.3.3"
+      }
     },
     "argparse": {
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.9.tgz",
       "integrity": "sha1-c9g7wmP4bpf4zE9rrhsOkKfSLIY=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "sprintf-js": "1.0.3"
+      }
     },
     "arr-diff": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-2.0.0.tgz",
       "integrity": "sha1-jzuCf5Vai9ZpaX5KQlasPOrjVs8=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "arr-flatten": "1.1.0"
+      }
     },
     "arr-flatten": {
       "version": "1.1.0",
@@ -173,7 +195,10 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/array-union/-/array-union-1.0.2.tgz",
       "integrity": "sha1-mjRBDk9OPaI96jdb5b5w8kd47Dk=",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "array-uniq": "1.0.3"
+      }
     },
     "array-uniq": {
       "version": "1.0.3",
@@ -239,7 +264,15 @@
       "version": "7.1.3",
       "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-7.1.3.tgz",
       "integrity": "sha512-zGy6aRrZjWhR+25n89jYkpLXutg6xdDGKKnVNzkfamnnhK3mYchkKj5GJR14aJiU0x4UqZ1+N728tkFNUL5N2g==",
-      "dev": true
+      "dev": true,
+      "requires": {
+        "browserslist": "2.4.0",
+        "caniuse-lite": "1.0.30000726",
+        "normalize-range": "0.1.2",
+        "num2fraction": "1.2.2",
+        "postcss": "6.0.11",
+        "postcss-value-parser": "3.3.0"
+      }
     },
     "aws-sign2": {
       "version": "0.6.0",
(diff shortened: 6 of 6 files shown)


XhmikosR:
Jekyll builds the md files to html.

Please make sure we are not caching useless stuff.

sw-src.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
XhmikosR: Why do we need this file? Can't you do what you want in sw.js?

sw.js
```diff
@@ -1 +1,156 @@
-// empty for now
```
XhmikosR: If this file is generated, then it should be removed now.

XhmikosR: And output the generated sw.js in _gh_pages folder.

sw-src.js
Lyricalz: We can, but this serves as a template to generate the main sw.js file

XhmikosR:
Then see my comment about the dist folder.

Lyricalz:
Should be all good now?

build/workbox.config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "globDirectory": "./",
+    "globPatterns": [
+        "assets/**/*.{js,html,svg,png,css,jpg,ico}",
+        "dist/**/*.{js,html,svg,png,css,jpg,ico}"
+    ],
+    "swSrc": "sw-src.js",
```
XhmikosR: How about just naming this `sw.js`? We output the generated file in the _gh_pages folder so we don't have any conflicts.

sw-src.js
```diff
@@ -0,0 +1,6 @@
+importScripts('https://unpkg.com/workbox-sw@1.3.0');
```
XhmikosR: Why do you use unpkg and not a local devDependency?

package.json
```diff
@@ -52,7 +52,8 @@
     "test": "npm-run-all dist js-test docs",
     "watch": "npm-run-all --parallel watch-css watch-js",
     "watch-css": "nodemon --ignore js/ --ignore dist/ -e scss -x \"npm run css && npm run css-docs\"",
-    "watch-js": "nodemon --ignore scss/ --ignore js/dist/ --ignore dist/ -e js -x \"npm run js-compile-plugins\""
+    "watch-js": "nodemon --ignore scss/ --ignore js/dist/ --ignore dist/ -e js -x \"npm run js-compile-plugins\"",
+    "workbox-precache": "workbox-cli inject:manifest --configFile=\"build/workbox.config.json\""
```
XhmikosR: We need to hook this script in the other scripts, say `docs`. And also rename it to `docs-workbox-precache`.

Lyricalz: workbox-precache fails on node 4 via travis as of now so I removed it from docs build, have put in a pull

It fails on [travis](https://travis-ci.org/twbs/bootstrap/jobs/265776270) due to incompatibility with node 4, so I had to remove it from the docs script to pass, but i've put in a [pull request](https://github.com/GoogleChrome/workbox/pull/753/files) to fix it though, so waiting on that for now

sw-src.js
Lyricalz: that makes more sense actually, working on it.

build/workbox.config.json
Lyricalz: makes more sense, will update

XhmikosR:
@Lyricalz: apart from the previous stuff I commented, you need to update package-lock.json too after you fetch and rebase from v4-dev.

sw-src.js
```diff
@@ -0,0 +1,6 @@
...
+
+const VERSION = 1;
```
XhmikosR: Wouldn't it make sense to use the version from package.json + another variable?

Lyricalz: yeah that's a good idea, since this is linked to the docs, are there any vars that are updated whenever the docs are rebuilt?

Lyricalz:
Made some updates
-removed workbox cli, now using workbox build directly
-included generated html to be pre-cached too so the docs are now fully available offline
-removed version const, turns out that serviceworkers are prompted to update when there's a byte difference, the revision hashes take care of this automatically
-automatically copies local workbox-sw library in case of any version changes
-renamed npm script also added it to docs-compile script
-set page to reload automatically when there's a service worker update

Note: build still hasn't passed on node 4 but once the workbox-team have [merged](https://github.com/GoogleChrome/workbox/pull/753#issuecomment-323432229) & released i'll update to that version and it should pass then

build/workbox.js
```diff
@@ -0,0 +1,28 @@
...
+var workboxSWSrcPath = require.resolve('workbox-sw')
+var workboxSWDestPath = 'assets/js/vendor/workbox-sw.js'
+fse.exists(workboxSWDestPath, function (e) {
+  if (!e) {
+    fse.copy(workboxSWSrcPath, workboxSWDestPath)
+    console.log('copied workbox service worker.')
+  }
+})
+
+config.manifestTransforms = [mdToHtml]
```
XhmikosR: Why do we need workbox to do MD -> HTML?

The _gh_pages folder already has the html files.

build/workbox.config.json
```diff
@@ -0,0 +1,15 @@
+{
+    "globDirectory": "./",
+    "globPatterns": [
+        "assets/**/*.{js,svg,png,css}",
+        "dist/**/*.{js,css}",
+        "docs/**/*.{md,html,css,jpg,js}"
```
XhmikosR: We shouldn't cache md files.

XhmikosR: Also, this seems to generic. You should change the whole config to work against the _gh_pages folder which only has the needed assets and not the src stuff.

build/workbox.config.json
```diff
@@ -0,0 +1,15 @@
...
+    ],
+    "swSrc": "./sw.js",
+    "swDest": "./_gh_pages/sw.js",
+    "templatedUrls": {
+      "/": ["_gh_pages/docs/index.html"],
+      "docs": ["_gh_pages/docs/index.html"],
+      "docs/4.0/": ["_gh_pages/docs/4.0/index.html"]
```
XhmikosR: Why do we need this? If we cache the whole _gh_pages folder, doesn't workbox take care of this for us?

build/workbox.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+var fse = require('fs-extra')
```
XhmikosR: You don't list fs-extra in devDependencies? BTW can't we just use `fs`?

XhmikosR:
@Lyricalz : thanks for your patience, we'll get there :)

build/workbox.js
XhmikosR: Oh, also, you could/should ES6-ify this file.

build/workbox.config.json
Lyricalz: So 'templatedUrls' is just a way to redirect urls to their relative sources. Since the docs uses /docs/index instead.html of _gh_pages/docs/index.html and _gh_pages/* doesn't actually exist via 'http://getbootstrap.com/gh_pages/*', we're basically saying get the source contents of _gh_pages/docs/index.html and the cache the contents of it as if you're visiting the url '/'.

build/workbox.config.json
Lyricalz: Okay so I added a [transformer](https://github.com/twbs/bootstrap/pull/23533/files#diff-7382067a8603719f3dfffe1afa96738dR7) for this as if I specified the _gh_pages folder, workbox will attempt to pre-cache an url like.. http://getbootstrap.com/gh_pages/docs/4.0/**, instead of http://getbootstrap.com/docs/4.0/**. The transformer basically renames the .md files to their correct .html files but keeps the docs/ structure intact instead of _gh_pages/docs/

build/workbox.js
Lyricalz: see comment below

Lyricalz:
@XhmikosR actually I wasn't thinking right about the md-html thing, should just go straight to the _gh_pages as you suggested, just updated and moved back to just normal fs instead. I think we've got it now

build/workbox.js
```diff
@@ -0,0 +1,21 @@
...
+const config = require('./workbox.config.json')
+
+const updateUrl = (manifestEntries) => manifestEntries.map((entry) => {
+  if (entry.url.startsWith('_gh_pages/')) {
+    entry.url = entry.url.replace(/_gh_pages\//g, '')
+  }
+  return entry
+})
+
+const workboxSWSrcPath = require.resolve('workbox-sw')
```
XhmikosR: Are these 2 used?

XhmikosR:
One last thing before I test it locally. Please revert the change in package-lock.json and do `npm i workbox-build -D`

build/workbox.js
Lyricalz: yep workbox-build ships with workbox-sw as a dependency, so I thought it made better sense to just copy it directly from node_modules, that way everytime we bump the version of workbox-build the relevant version of workbox-sw gets copied too.

Also I just realised i'm copying to assets/js/vendor/
should this instead be copied to _gh_pages/assets/js/vendor/? i'm not sure as assets/js will work locally but gh_pages/assets/js would work remotely?

XhmikosR: Yup, everything should be in `_gh_pages` folder. That is what we commit to the gh-pages branch and thus the live site.

Lyricalz:
Done!

assets/js/src/pwa.js
```diff
@@ -7,6 +7,19 @@
     window.addEventListener('load', function () {
       navigator.serviceWorker.register('/sw.js').then(function (registration) {
         console.log('ServiceWorker registration successful with scope: ', registration.scope)
+        registration.onupdatefound = () => {
+          var installingWorker = registration.installing;
+          installingWorker.onstatechange = () => {
```
XhmikosR: Can you use ES5 here? We use uglify-js which doesn't support ES6 (there's a branch, I know, but we don't use ES6 here).

Lyricalz: whoops, updated now

Lyricalz:
@XhmikosR workbox just released [2.0.0](https://github.com/GoogleChrome/workbox/commit/5cfdcb37e9c59bdae96e5821d34fd62eb7244485) now. Fine to bump the version to that?

XhmikosR:
@Lyricalz: yes, please do so and fetch & rebase against latest v4-dev.

Lyricalz:
@XhmikosR all done now and Node 4 build has passed with the fix made to workbox

XhmikosR:
@Lyricalz: awesome, thanks! I'll have a look tomorrow. Did you verify everything works offline? Do we also need an offline page?

build/workbox.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs')
+const swBuild = require('workbox-build')
+const config = require('./workbox.config.json')
+
+const updateUrl = (manifestEntries) => manifestEntries.map((entry) => {
+  if (entry.url.startsWith('_gh_pages/')) {
```
XhmikosR: One minor change here, maybe move `_gh_pages` to a `const` instead of hardcoding it in this file.

Lyricalz: Just updated this

Lyricalz:
@XhmikosR yep everything works offline.. except for the 2 of the png favicons

See screenshots below:

https://snag.gy/gLjFRv.jpg
Also noticed this weird response from the favicons too: https://snag.gy/MHhauL.jpg

XhmikosR:
That is weird. But please try without any extensions enabled.

XhmikosR:
@Lyricalz: I rebased your branch and force pushed it so make sure you don't overwrite it :)

I also had to make a change in `sw.js` to use a root relative url in `importScripts`.

Lyricalz:
@XhmikosR disabled extensions, still same issue:

https://snag.gy/N4oiu1.jpg

XhmikosR:
Can you try with these 2 files?

favicon-32x32
favicon-16x16

Lyricalz:
hmm it could just be me.. but loading these directly from their URL displays this same issue in the preview:

https://snag.gy/1a6DvW.jpg

XhmikosR:
OK, so it seems it's not our issue :)

Can you please fetch your branch and try locally if everything is fine? It seems to work for me fine, I just want to be 100% sure.

Lyricalz:
@XhmikosR did so, all works fine both offline & online :)

XhmikosR:
@Lyricalz: can you fix the remaining one ESLint error? I fixed the others and force pushed.

XhmikosR:
Another thing I notice in the generated sw.js is the paths for URL

```
"url": "examples\\index.html",
```

Shouldn't those use the unix separator regardless of me being on Windows?

Lyricalz:
@XhmikosR i've updated the build file, I don't have a windows anymore so can't test, but I think that should do the trick

build/workbox.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs')
+const path = require('path')
+const swBuild = require('workbox-build')
+const config = require('./workbox.config.json')
+const buildPrefix = '_gh_pages/'
+
+const updateUrl = (manifestEntries) => manifestEntries.map((entry) => {
+  if (entry.url.startsWith(buildPrefix)) {
+    const regex = new RegExp(buildPrefix, 'g')
+    entry.url = path.normalize(entry.url.replace(regex, ''))
```
XhmikosR: I don't think this will fix the issue. `path.posix` should be used, but I'll try it later and see :)

sw.js
```diff
@@ -1 +1,4 @@
...
+self.importScripts('/assets/js/vendor/workbox-sw.js')
```
gauntface: you should include the version version here or add a revision hash otherwise you may not be able to update the library

build/workbox.js
XhmikosR: @Lyricalz Please remove the workaround since the issue shouldn't happen. I have opened an issue upstream.

Lyricalz: @XhmikosR done :)

XhmikosR:
Thanks! I'll have a final look when the issue is fixed upstream and then merge :)

gauntface:
Just an FYI - We have a 2 weekly push for releases so the next release is scheduled for 6th of August. Should be more than enough time to land your fix and get it pushed out.

Team calendar: https://workboxjs.org/community.html

Lyricalz:
@XhmikosR I've just rebased and upgraded workbox-build to 2.0.1 which includes your backslash fix, we ready to merge? 😄

XhmikosR:
@Lyricalz: I rebased the branch, please make sure you don't overwrite my changes.

Now, 2 final issues, I promise :)

1. We get some warnings from workbox:
    ```
    C:\Users\xmr\Desktop\bootstrap>node -v && npm -v
    v8.5.0
    5.3.0

    C:\Users\xmr\Desktop\bootstrap>npm run docs-compile

    > bootstrap@4.0.0-beta docs-compile C:\Users\xmr\Desktop\bootstrap
    > bundle exec jekyll build && npm run docs-workbox-precache

    Configuration file: C:/Users/xmr/Desktop/bootstrap/_config.yml
                Source: .
           Destination: ./_gh_pages
     Incremental build: disabled. Enable with --incremental
          Generating...
                        done in 4.297 seconds.
     Auto-regeneration: disabled. Use --watch to enable.

    > bootstrap@4.0.0-beta docs-workbox-precache C:\Users\xmr\Desktop\bootstrap
    > node build/workbox.js

    (node:3884) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
    (node:3884) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'replace' of undefined
    (node:3884) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
    ```
2. Not sure how to make our scripts generate the workbox stuff so that we can test things without us having to run 2 separate cmds simultaneously. Currently I have to do `bundle exec jekyll s` and `npm run docs-workbox-precache` in two separate cmd windows
3. I'm not sure how to debug this being that the service worker isn't being registered locally without https.

Lyricalz:
@XhmikosR i've updated 1) to get rid of this error
not sure about 2) but doesn't the script in `docs-compile` take care of this?
3) as far as I know localhost/127.0.0.1 is considered secure origin, so your service worker should still register, unless i'm missing something here

XhmikosR:
@Lyricalz:

The error I was must be because the service worker wasn't properly being registered without your last patch :)

Now I can navigate the docs offline.

1. `[2017-09-14 10:14:31] ERROR `/assets/js/vendor/workbox-sw.prod.v2.0.1.js.map' not found.` Maybe we should disable map for workbox or output it
2. I added *.json files too in precache otherwise manifest.json wasn't cached
3. And the favicon not being cached for some weird reason

I think we are almost done :)

XhmikosR:
@gauntface: do you have any idea why we get this for the favicons?

```
:9001/assets/img/favicons/favicon-32x32.png:1 GET http://localhost:9001/assets/img/favicons/favicon-32x32.png net::ERR_INTERNET_DISCONNECTED
10:19:54.421 :9001/assets/img/favicons/favicon-16x16.png:1 GET http://localhost:9001/assets/img/favicons/favicon-16x16.png net::ERR_INTERNET_DISCONNECTED
```

I see sw.js has them so something else must be going on.

Lyricalz:
@XhmikosR 
1) was thinking about this last night, i'll add the map file later this evening it should just be a matter of appending .map to the the workbox path and copying it too
2) yes, that's the right thing to do!
3) I think the favicon issue is independent of this as even viewing the file directly & outside of the docs produces this strange error: https://snag.gy/1a6DvW.jpg

Lyricalz:
Also re: `docs-serve` i'm assuming this command re-builds & runs a local server with that one command? If so, I guess we won't be able to generate the manifest for pre-caching before the local server starts. Is that something that we need to address? i've been just running `docs-workbox-precache` after `docs-serve`, which isn't a big deal.. But yeah, would be nice if we could get the serve to build docs, build pre-cache manifest & then run server

build/workbox.config.json
```diff
@@ -1,7 +1,7 @@
 {
   "globDirectory": "./",
   "globPatterns": [
-    "_gh_pages/**/*.{html,css,js,png,jpg}"
+    "_gh_pages/**/*.{html,css,js,json,png,jpg}"
```
addyosmani: With respect to the favicon issue you're seeing, is it possible to view your current build in Chrome Canary and take a screenshot of what is in Application Panel -> Cache Storage? Favicon issues like this are often due to the asset either not existing in the final build, the globPattern not being sufficient (don't believe this is an issue here) or something else.

XhmikosR: @addyosmani: I see the favicons being in the workbox cache:

cache

And thanks for the reply :)

XhmikosR (approved):

XhmikosR:
Thanks to everyone for the help and ofc @Lyricalz for your patience!

I'll keep an eye for any regressions or fixes for the favicon issue.

Lyricalz:
Awesome!

gauntface:
Just a cheeky FYI, if you have any further issues, feature requests or feedback for Workbox please raise an issue (thanks for you help fixing the windows bug ;) ).

Lyricalz:
@gauntface would you mind reviewing this [file](https://github.com/Lyricalz/bootstrap/blob/fa038339beab089ac1a2d88727fbff2ae8bf1563/assets/js/src/pwa.js) please? Would you say that is the recommended way to prompt an update with service workers?

XhmikosR:
@Lyricalz: another thing I noticed is that if you visit the homepage, go offline, then the docs page isn't cached. I think is something with the page being redirected.

Lyricalz:
@XhmikosR I noticed this too, but that's because the live link is hardcoded in the redirect, rather than the local url. I was thinking about making a PR, to make these links relative, rather than using absolute URLs.

When you click documentation it actually goes here: https://getbootstrap.com/docs/4.0/ looking at the source - `view-source:https://getbootstrap.com/docs/4.0/` you can see the redirects are absolute URLs

XhmikosR:
@Lyricalz: we use jekyll-redirect-from for the redirects so I'm not sure if it's even possible since I haven't looked at it. If you make a PR please CC me.

gauntface:
@Lyricalz that looks fine to me.

Workbox-sw does provide an alternative way to listen for a broadcast message when assets are updated: https://github.com/GoogleChromeLabs/workbox-microsite/blob/master/src/themes/scripts/sw-controller.js#L18

PR #23534, opened 2017-08-17
Update devDependencies.

XhmikosR:


Johann-S:
It's strange it's an update of our dependencies but they are no change in our `package.json` 🤔

XhmikosR:
Not strange at all. That's how it works when using `^x.y.z`.

tomscholz:
You are also correcting a typo. I think you should mention that somewhere. Otherwise looks good

Johann-S:
Oh yes I forgot that thank you 👍

Johann-S (approved):

XhmikosR:
@tomscholz: nah, not worth. Ideally the branch protection wouldn't be here and I could push directly like in the past. In that case I would do things differently but now I'm spending too much time on useless stuff like rebasing and waiting for tests to pass.

Issue #23489, opened 2017-08-16
Bootstrap 4 print problem in Firefox


jevanlingen:
Printing multiple pages in Firefox does not work. Only one page is printed, the rest of the pages is missing. Seems it is a known issue for years within the Mozilla community: https://bugzilla.mozilla.org/show_bug.cgi?id=104040.

**Case**
The Bootstrap v4 [documentation](https://getbootstrap.com/docs/4.0/) page. 

**Compare**
Version 3 of Bootstrap does not have this issue, compare the [old docs](https://getbootstrap.com/docs/3.3/css/) for example.

patrickhlauke:
![capture](https://user-images.githubusercontent.com/895831/29359784-e72d03c8-8277-11e7-8ecc-e63d7a5b5f9f.PNG)

prateekgoel:
It's the `display:flex` in `.row` that's causing the issue. Setting `display` to block for `print` resolves this.
Can I work on this?

mdo:
This doesn't appear to be an issue in Firefox 59 on Windows or macOS. Just printed to PDF to see and all the pages show.

kcantwel:
I am seeing this exact same issue with Firefox 60 and Bootstrap 4.1.1.

garthtee:
I see this issue also, but only when scrolling using arrow keys on my laptop when I use my mouse it's fine.

dozer75:
The referenced issue in the original post is marked as duplicate of duplicate in their bug tracking system, so just adding the real tracked [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=939897) for Firefox which still after 5 years (!!!) isn't solved (they do have ways to reproduce it since it's not possible to print their own bug tracking system).

The solution to use `display: block;` wasn't a good solution for me since it mess quite a bit with element margin usage, I will rather advice our users to avoid using print capabilities in Firefox.

Lowell130:
> It's the `display:flex` in `.row` that's causing the issue. Setting `display` to `block` for print resolves this.
> Can I work on this?

thanks work great!

PR #23547, opened 2017-08-18
add firfox bug in doc #23489

chiraggmodi:
add browser bug as mention in #23489. i have also tried as suggestion in comment to add `display: block` but it is not printing properly. have attached 
screenshot

### DIFF
--- a/_data/browser-bugs.yml
+++ b/_data/browser-bugs.yml
@@ -150,6 +150,16 @@
 
 -
   browser: >
+    Firefox
+  summary: >
+    Only one page printed
+  upstream_bug: >
+    Mozilla#104040
+  origin: >
+    Bootstrap#23489
+
+-
+  browser: >
     Firefox (Windows)
   summary: >
     Right border of `<select>` menu is sometimes missing when screen is set to uncommon resolution


prateekgoel:
@chiraggmodi I think you should not submit a PR for the issue where someone is asking or claiming to work on it.
/CC @mdo

mdo:
We're going to link to a 16 year old bug in Firefox? Surely there's something more up to date...

prateekgoel:
@mdo Seems like setting `display` to `block` instead of `flex` is the only solution right now.
I did a little research on this and found that firefox has issues with printing flex containers.
Also, noted here:
https://stackoverflow.com/questions/4609007/firefox-printing-only-1st-page
https://stackoverflow.com/questions/43492724/firefox-prints-first-page-only-but-other-browsers-doing-fine

I tried changing `display:flex;` to `display:block;` in Firefox web inspector and this solved the printing issue for me.

chiraggmodi:
yes, i do change to `display:block` but we need to change many classes to print appropriate as chrome

mdo:
Closing per #23489.

PR #23552, opened 2017-08-18
Helpers to add a shadow to text

afuno:


afuno:
I see my mistakes. Now I'll fix it.

afuno:
Done.

mdo:
Thanks, but I don't think we'll be implementing this as coded. I'm fine with a global flag for `text-shadow` properties, but I don't see a need for generating a series of utilities like this.

PR #23560, opened 2017-08-19
Rename `.table-inverse`, `.thead-inverse`, `.thead-default` to `.*-dark`, `.*-light`

pat270:
Sending another breaking change, this seemed like an oversight. I renamed `.table-inverse`, `.thead-inverse`, `.thead-default` to `.table-dark`, `.thead-dark`, `.thead-light`. I also changed the variables `$table-inverse-*` to `$table-dark-*`.

The last commit fixes a bug where `.thead-inverse` and `.thead-default` doesn't apply the proper variant border color, also made the selector more specific so that it will win over the parent `.table` class e.g. using `.table-inverse` with `.thead-default`.

### DIFF
--- a/docs/4.0/content/tables.md
+++ b/docs/4.0/content/tables.md
@@ -45,10 +45,10 @@ Using the most basic table markup, here's how `.table`-based tables look in Boot
 </table>
 {% endexample %}
 
-You can also invert the colors—with light text on dark backgrounds—with `.table-inverse`.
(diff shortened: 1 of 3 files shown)


mdo (approved):

hebeldark:
Hi

i'm download css and js files about v4 beta 2.

i rename thead-inverse to thead-dark in my files

but sometimes, i have the header of my table black, sometimes no


i don't understand why i have this difference ?

PR #23565, opened 2017-08-19
Remove jQuery from Util.js

Johann-S:
That's the first try to remove jQuery from our source, with this PR no more jQuery in Util.js and other plugins will follow the move 😄 

I added a new folder `dom` which will allow us to make some usefull functions to replace jQuery

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -9,23 +9,9 @@ import Event from './dom/event'
 
 const Util = (() => {
 
-
-  /**
-   * ------------------------------------------------------------------------
-   * Private TransitionEnd Helpers
-   * ------------------------------------------------------------------------
-   */
-
-  let transition = false
+  const transition = Event.getBrowserTransitionEnd()
 
   const MAX_UID = 1000000
-
-  const TransitionEndEvent = {
-    WebkitTransition : 'webkitTransitionEnd',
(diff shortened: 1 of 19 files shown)


js/src/util.js
```diff
@@ -100,7 +67,7 @@ const Util = (($) => {
 
   const Util = {
 
-    TRANSITION_END: 'bsTransitionEnd',
+    TRANSITION_END: transition.end,
```
Johann-S: Breaking change here, before that modification folks was able to listen to `bsTransitionEnd` event
Now they have to listen to their browser transition end event for example : `webkitTransitionEnd`

XhmikosR: See why I insisted on doing this after v4? :)

Johann-S: I'll find a solution 😉
and that's better to do it now instead of when we will have a stable release because we're still in beta so breaking change can happen and must happen now

XhmikosR: I don't agree... No breaking change should happen from now until v4 stable IMO.

Johann-S:
Close because I'll open a huge PR to track the removal of jQuery

XhmikosR:
@Johann-S: just make PRs against a new branch not v4-dev.

PR #23569, opened 2017-08-19
Add npm-check-updates as a devDependencies

tkrotoff:
npm-check-updates (`ncu`) can be a `devDependencies` inside package.json (instead of installing it manually). `npm run maintenance-dependencies` works with local npm-check-updates.

Command line used to generate package.json and package-lock.json: `npm install --save-dev npm-check-updates`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,11 +69,31 @@
       "integrity": "sha1-MU3QpLM2j609/NxU7eYXG4htrzw=",
       "dev": true
     },
+    "align-text": {
+      "version": "0.1.4",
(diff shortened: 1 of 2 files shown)


Johann-S (approved):
LGTM 👍

XhmikosR:
I'm not in favor because of the bloat and the fact that this is just an internal thing.

My guess is this was the reason @bardiharborow didn't add it either.

bardiharborow:
No one outside of the core team should really need to do this, and in the long term someone should probably send a PR to npm and create a flag for `npm update` that replicates `npm-check-updates`'s behaviour.

PR #23572, opened 2017-08-20
Switch to stylelint.

### LINKED ISSUE #23383: Switch to stylelint.

XhmikosR:
Fixes #23383 

Note to all: feel free to push to this branch, and merge from v4-dev, but please do not rebase until we are ready to merge to v4-dev.

TODO:

- [x] Make sure tests are green
- [x] Check for properties order
- [x] Tighten rules
- [x] Make sure Hound works


wojtask9:
Probably you also want to use `stylelint-order` plugin (https://github.com/hudochenkov/stylelint-order).

XhmikosR:
Yup, it's on my todo, but I thought I'd let the other team members chime in before I make a lot more changes myself :)

andresgalante:
It'd be great if we make style lint format the code or use another formater. 
Fornater + lint = happiness 

Love this change 👍🏻

XhmikosR:
I believe this is in a mergeable state; we can tighten the rules later more.

My only issue is freaking Hound. I don't know if it will stop complaining after this has been merged or not.

bardiharborow:
@XhmikosR bit of a shot in the dark, but you may need to pass `scss: enabled: false` in `.hound.yml` to disable the default linter (which is now malfunctioning because you've removed the config file). My recollection is that Hound configurations are somewhat unintuitive in that respect.

wolfy1339:
Hound uses the config of the parent branch, not the one in the current branch. The only way to test hound is to PR against this branch, in order to use the config in this branch

XhmikosR:
OK, then we can ignore Hound in this PR.

@mdo: please have a look when you have some time and merge or let me know if you want me to make any changes. We'll need to tweak the rules more after the merge anyway; I have some of the rules set to `null` so that things pass currently.

XhmikosR:
(I can't override Hound's blocking review, only repo owners can)

package.json
```diff
@@ -117,9 +122,7 @@
     "build",
     "js/.eslintrc.json",
     "js/**/*.js",
-    ".scss-lint.yml",
-    "scss/**/*.scss",
-    "LICENSE"
```
mdo: Shouldn't the license stay?

XhmikosR:
It's default included by npm.

package.json
XhmikosR: It's default included in npm.

XhmikosR:
@mdo: is there anything else holding us for this? Otherwise pull the trigger 🔫

PR #23577, opened 2017-08-20
Beta 2 migration docs

mdo:
Opening this as a WIP pull request to track breaking changes coming in Beta 2. We don't want to have any breaking changes, so these should be very few in number and clearly articulated here.

---

### Todo

- [x] #23560: Table modifiers
- [x] #23568: Drop Bower support
- [x] #23579: Grayscale
- [x] #22804: Responsive tables

### DIFF
--- a/docs/4.0/migration.md
+++ b/docs/4.0/migration.md
@@ -6,9 +6,11 @@ group: migration
 toc: true
 ---
 
-{% callout info %}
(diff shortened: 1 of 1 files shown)


Johann-S:
We should mention that now Bootstrap 4, require jQuery 3

mdo:
Will update as we go from now on.

PR #23582, opened 2017-08-21
Use postcss-calc to simplify css calc(), closes #23581

### LINKED ISSUE #23581: Feature Request: Use postcss-calc to simplify calc expresssion
They will simplify a few css _calc()_ on `.col-form-label`

bangzek:
It will simplify css calc() expression on `.col-form-label`, `.col-form-label-lg` and `.col-form-label-sm`.

### DIFF
--- a/build/postcss.config.js
+++ b/build/postcss.config.js
@@ -5,6 +5,7 @@ module.exports = (ctx) => ({
     sourcesContent: true
   },
   plugins: {
+    'postcss-calc': {},
     autoprefixer: {
       browsers: [
         //
--- a/package-lock.json
+++ b/package-lock.json
@@ -1309,6 +1309,12 @@
         "boom": "2.10.1"
       }
     },
+    "css-unit-converter": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/css-unit-converter/-/css-unit-converter-1.1.1.tgz",
+      "integrity": "sha1-2bkoGtz9jO2TW9urqDeGiX9k6ZY=",
+      "dev": true
+    },
     "currently-unhandled": {
       "version": "0.4.1",
       "resolved": "https://registry.npmjs.org/currently-unhandled/-/currently-unhandled-0.4.1.tgz",
@@ -2034,6 +2040,12 @@
         "graceful-fs": "4.1.11",
         "write": "0.2.1"
       }
+    },
+    "flatten": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/flatten/-/flatten-1.0.2.tgz",
+      "integrity": "sha1-2uRqnXj74lKSJYzB54CkHZXAN4I=",
+      "dev": true
     },
     "for-in": {
       "version": "1.0.2",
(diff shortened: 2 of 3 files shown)


XhmikosR:
@bangzek: you need to update package-lock.json too.

@mdo: this is what I meant we should use more postcss plugins. This is an easy optimization, same as the mqpacker one. :)

mdo:
Why do we need it?

XhmikosR:
It's one of the many available optimization steps we can take advantage of
to reduce the final bundle size to the minimum. It might seem like a small
improvement but the whole idea is to apply many more; at least that is my
plan and eventually we can save an important amount of CSS with these easy
changes.

mdo:
Okay, I think maybe I need to rephrase myself with these kind of questions. Here's what goes through my mind with PRs like this:

- What does `postcss-calc` do?
- How is it better or simpler than CSS's regular `calc`?
- Is it more approachable than the regular `calc`? 
- Does it add more burden to us, or users of Bootstrap?
- Why so many PostCSS plugins lately? We use Sass—how do we communicate this to folks?

None of these PRs lately have answers those questions @XhmikosR. You're telling me it's an optimization and it's better—you gotta tell me how exactly. I can go hunt some of this information down myself, but without it, I can't approve of the changes.

XhmikosR:
postcss-calc just reduces the calc references. In our case it's only 3 but it's a pretty safe and easy optimization we sould/should have in our toolbox.

The resulting CSS has still the same effect. See the diff:

```diff
 dist/css/bootstrap.css | 12 ++++++------
 1 file changed, 6 insertions(+), 6 deletions(-)

diff --git a/dist/css/bootstrap.css b/dist/css/bootstrap.css
index 9164d6876..8e667f466 100644
--- a/dist/css/bootstrap.css
+++ b/dist/css/bootstrap.css
@@ -1796,20 +1796,20 @@ select.form-control:focus::-ms-value {
 }
 
 .col-form-label {
-  padding-top: calc(0.5rem - 1px * 2);
-  padding-bottom: calc(0.5rem - 1px * 2);
+  padding-top: calc(0.5rem - 2px);
+  padding-bottom: calc(0.5rem - 2px);
   margin-bottom: 0;
 }
 
 .col-form-label-lg {
-  padding-top: calc(0.5rem - 1px * 2);
-  padding-bottom: calc(0.5rem - 1px * 2);
+  padding-top: calc(0.5rem - 2px);
+  padding-bottom: calc(0.5rem - 2px);
   font-size: 1.25rem;
 }
 
 .col-form-label-sm {
-  padding-top: calc(0.25rem - 1px * 2);
-  padding-bottom: calc(0.25rem - 1px * 2);
+  padding-top: calc(0.25rem - 2px);
+  padding-bottom: calc(0.25rem - 2px);
   font-size: 0.875rem;
 }
 ```

We do use Sass, but that doesn't mean we can't hook up Postcss plugins. We should definitely do this because Postcss is a post-processor not a pre-processor and can help a lot if we set up things carefully.

Note that one thing people blame us is the bloat/result size. We can't do miracles and keep the same functionality but we can squash a few KB here and there.

mdo:
Ahhh okay so it rewrites things for a simpler output. That's chill. I didn't know if it was just affecting compiled code or trying to change how we use `calc` in our source Sass files.

XhmikosR:
@mdo: I think we can drop this PR since the gain isn't big.

bangzek:
This no longer needed cause that `calc()` on **.col-form-label*** not exist anymore

PR #23584, opened 2017-08-21
Make jumbotron padding equal, fixes #23583

### LINKED ISSUE #23583: Fix jumbotron padding
Please do visual comparison between:
http://getbootstrap.com/docs/4.0/examples/narrow-jumbotron/
and:
https://getbootstrap.com/docs/3.3/examples/jumbotron-narrow/

You would find v4 jumbotron padding is so much longer on the vertical.

On the old v3 the space is 68px on the vertical and 60px on horizontal,
this is because `h1` has 20px _margin-top_.

In the current v4 **the vertical is 64px** (on normal 16px default font size) and **32px on horizontal**.
Probably v4 just copy the older padding proportion but failed to adjust with the removal of all margin-top.

I propose to just make the padding equal, so the end result will be 64px on both.

BTW, the example thumbnails needs updating.

bangzek:
See the issues for details.

### DIFF
--- a/docs/4.0/examples/narrow-jumbotron/narrow-jumbotron.css
+++ b/docs/4.0/examples/narrow-jumbotron/narrow-jumbotron.css
@@ -32,9 +32,9 @@ body {
 }
 
 /* Customize container */
-@media (min-width: 48em) {
+@media (min-width: 52em) {
   .container {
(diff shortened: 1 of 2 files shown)


mdo:
Closing per issue.

PR #23586, opened 2017-08-21
v5 without jQuery

Johann-S:
### Plugins:

- [x] Util (#23565)
- [x] Alert (#23596)
- [x] Button (#23600)
- [x] Carousel (#23658)
- [x] Collapse (#23958)
- [x] Dropdown (#24099)
- [x] Modal (#23955)
- [x] Popover (#24014)
- [x] ScrollSpy (#24879)
- [x] Tab (#24094)
- [x] Tooltip (#24014)

### Tests:

- [x] Remove the use of `Data` in our unit tests (#26685)
- [x] Remove the use of `EventHandler` in our unit tests (#26685)
- [x] Fix coverage by testing our new internal objects

### Build:

- [x] Remove jQuery in our Rollup config
- [ ] See why there is no `const` or `let` in the dist files. ([bootstrap.js](https://deploy-preview-23586--twbs-bootstrap.netlify.com/docs/4.3/dist/js/bootstrap.js))

### Docs:

- [ ] Update plugin pages
  - [x] Alert (#27062)
  - [x] Button (#27173)
  - [x] Carousel (#27173) - still has a few references to `$()`
  - [x] Collapse (#27173) - still has a few references to `$()`
  - [ ] Dropdown
  - [ ] Modal
  - [ ] Popover
  - [ ] ScrollSpy
  - [ ] Tab
  - [ ] Toasts
  - [ ] Tooltip

- [ ] Other doc pages
  - [ ] List group
  - [ ] Forms
  - [ ] Nav
  - [ ] getting-started/browsers-devices.md
  - [ ] getting-started/javascript.md

- [x] Stop using jQuery
- [x] Update `/getting-started/browsers-devices/` and remove any IE references
- [ ] Add a section explaining that it's possible to use this version with and without jQuery
- [ ] Search docs for `jquery` (not case sensitive) and `$(`
- [ ] Replace https://codepen.io/Johann-S/pen/djJYPb or add a codepen with vanilla JS
- [ ] Maybe explain when an element is considered visible in the scrollspy callout

### Browser compatibility:

- [x] Remove IE

### Various

- [x] Make sure we don't reference old versions like 4.1.3 in any newly added files
- [x] Same as above for copyright year
- [ ] Drop `Util.TRANSITION_END`?
- [ ] See if we really need `DOMContentLoaded` anymore in tests and docs.

### Demo:

<https://deploy-preview-23586--twbs-bootstrap.netlify.com/>

If you want to help us do not hesitate to create a PR which targets this branch `v4-without-jquery`.


XhmikosR:
@Johann-S: did you find any lib for namespaced events or is that still broken?

afuno:
@Johann-S It's really cool idea!

Johann-S:
Currently nothing is broken but I cannot finish to rewrite entirely the Alert plugin without a namespaced events...

dushkostanoeski:
Will there be both a jQuery and a non-jQuery version? If you're using libraries that require jQuery, you can use the jQuery version, and if not, the non-jQuery one.

Johann-S:
We will have only one release without jQuery but if we detect jQuery you'll be able to use Bootstrap's plugin as before 👍

danijelGombac:
How large will be bootstrap.min.js without jQuery?

Johann-S:
that's too early to have an estimation @danijelGombac sorry, but you can follow the development 👍

tmorehouse:
If you are interested in Vanilla JS versions of tooltip, popover, and scrollspy, We have versions that we use in [Bootstrap-Vue V1.0.0.beta.7](https://github.com/bootstrap-vue/bootstrap-vue)

- tooltip.js Class https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/classes/tooltip.js
- popover.js Class https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/classes/tooltip.js
- scrollspy (awaiting a re-write into ES6 class format) https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/directives/scrollspy.js

We have a few extra features added to Tooltips and popovers (such as a new `blur` leave event... i.e. `trigger='click blur'`, and auto appending tooltips/popover into modal... which is pending in PR https://github.com/bootstrap-vue/bootstrap-vue/pull/1024)

Johann-S:
thanks @tmorehouse we will take a look 👍

tmorehouse:
@Johann-S no problem.  There are only a few Vue.js centric items in the tooltip/popover classes (such as event emitting)

tmorehouse:
Boostrap-Vue is quite interested in the jQuery free version of some of the plugins, as we might be able to direclty import some of these modues (or the classes), to replace some of our code.

We have just finished porting ScrollSpy over to vanilla js (with some helper dom functions we have created).

If you like, I could see if I can port our vanilla version to this branch as a PR (the main things to change would be the event signatures)

Johann-S:
Sure @tmorehouse this branch is open to everyone who wants to help us 👍

tmorehouse:
OK, I ill try porting over in the next few days.

Johann-S:
@alekitto it seems we have issue on IE with your last commit see : https://saucelabs.com/beta/tests/af6abf5c5275433080920e7f34a234cc/commands#12 and https://travis-ci.org/twbs/bootstrap/jobs/276457805#L718

Johann-S:
@alekitto it's still failed on Edge see :  
- https://saucelabs.com/jobs/03afada422c7448d84ea067dee18906d
- https://travis-ci.org/twbs/bootstrap/jobs/276838570#L718-L721

alekitto:
@Johann-S It seems that MSEdge resets the `defaultPrevented` flag upon `dispatchEvent`, so when the event is re-dispatched after being passed into the jQuery event system the flag is set to false. Now I'm currently working on it..

Johann-S:
A great tool Bundlesize thanks @XhmikosR :+1: 

image

XhmikosR:
We'll need to visit the thresholds, but yeah, it works nice :)

js/src/dom/manipulator.js
```diff
@@ -0,0 +1,56 @@
...
+
+    element.removeAttribute(`data-${key.replace(/[A-Z]/g, (chr) => `-${chr.toLowerCase()}`)}`)
+  },
+
+  toggleClass(element, className) {
+    if (typeof element === 'undefined' || element === null) {
+      return
+    }
+
+    if (element.classList.contains(className)) {
```
dwightwatson: Would it not be possible to just call `element.classList.toggle(className)`?

alekitto: Is it supported on IE 11?

tmorehouse: IE (any version) doesn't support the `toggle` method of `Element.classList`: https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility

Herst: According to <https://caniuse.com/#search=classList> it just "does not support the second parameter for the `toggle` method".

tmorehouse: IE11 appears to have this feature, but IE10 doesn't.

thecodingdude:
@Johann-S, @XhmikosR @mdo, is there any progress updates in regards to this? It seems [tab](https://github.com/twbs/bootstrap/pull/23586/commits/2a48a82da0b38d0fe1b940313567beae8da4f5e1) and [scrollspy](https://github.com/twbs/bootstrap/pull/24879) have been rewritten without jQuery - which were the final two on the list.

@mdo stated "there will be no breaking changes from Beta 3 to stable". I am not sure if this will count as a breaking change, but I'd love to see this merged into v4 or 4.1 :)

Johann-S:
Not sure it'll arrive in v4 but maybe a v5 😉

dgt41:
It's a joke that v4 still has a jQuery dependency... Pity

thecodingdude:
I guess the jQuery dependency will be dropped by 2020, can't come soon enough :D

thecodingdude:
@Johann-S is there any reason why this _can't_ ship in v4? Would it be possible to create a bootstrap-nojquery.js and let people play around with it noting it's experimental and make not work 100%? This might accelerate the progress on getting this merged into 4.x. I don't consider this a breaking change since you'd literally be shipping the nojquery file alongside the others, only changing your script tag would be required.

I guess the biggest changes would be in documentation and how the methods are actually called in JS. But adding them to the docs shouldn't be too much work, and as far as I can see, this initiative feels good to go, not sure what the hold up is.

Can you spare a few details of what needs doing here? I'm keen to rid websites and projects of jQuery ASAP :P

Johann-S:
They are a lot of things to and to test and it's a huge work, as you can see it's in our v4.3 so maybe it'll happen 😉

LeeEvans1234:
Yes it is huge work but anything that's worth building is a lot of work no one has came forward to build a hub that is niether black or white up or down left or right it's the creation story it just there that s  why the work has to be created and in a brineray system that is open for knowloge for all to share please help and thank you

landsman:
Hi guys, can I help with something?

js/src/dom/manipulator.js
Herst: @tmorehouse The link you posted has been updated to reflect that IE10 does indeed have basic support for it.

Johann-S:
It's fail on : 

image

But I'm unable to reproduce this bug on Win10 Firefox...

More details : 
https://saucelabs.com/beta/tests/b5e40f5d444542ecb510df7f51fb8848/commands#13

And 

image

/CC @alekitto

alekitto:
I'm able to reproduce the error. Now investigating..

connorshea:
Just wanted to pop in and say thank you for your work on removing the dependency on jQuery! 🙇 🙇 🙇

ItaloBC:
Is pull this still on development?

.eslintrc.json
```diff
@@ -71,7 +71,6 @@
     "no-param-reassign": "off",
     "no-proto": "error",
     "no-restricted-properties": "error",
-    "no-return-assign": "error",
```
XhmikosR: I wouldn't remove this. Why can't we assign the value before the return call?

Johann-S: I don't remember when I changed that, but feel free to change that 😉

js/src/dom/polyfill.js
```diff
@@ -0,0 +1,94 @@
+import EventHandler from './eventHandler'
+
+const Polyfill = (() => {
+  // defaultPrevented is broken in IE.
+  // https://connect.microsoft.com/IE/feedback/details/790389/event-defaultprevented-returns-false-after-preventdefault-was-called
```
Herst: Link broken, possibly working link: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/103709/

Johann-S:
@XhmikosR I think SauceLabs have issues on their browsers... Because when I check manually on Win10 and the latest Chrome (67) everything works fine, Can you check too ?

aTable:
Looks like the majority of the work is done so I'll clean up the docs. Unless there is something specific?

Johann-S:
I think @mdo or @andresgalante will push some CSS features too here 😄 but yep you can work on changing our documentation

Johann-S:
@XhmikosR BrowserStack fix everything 😄

XhmikosR:
@Johann-S: remember that IE has issues with `forEach`. I'm seeing the error in this branch's docs:

```
SCRIPT438: Object doesn't support property or method 'forEach'
application.js (21,5)
```

Not sure if there are more cases either in core or docs. We could just polyfill this if we want to use `forEach` and not plain `for` loops.

Johann-S:
That's very weird there is an issue on SelectorEngine because tests passed on BrowserStack 🤔
We don't need to add a polyfill we need to convert the result of document.querySelectorAll to an array

XhmikosR:
Sure, I only suggested it so that we don't need to do this across the codebase. Because it seems tests didn't catch this.

Johann-S:
@XhmikosR I pushed two fixes, everything should be fine now 😉

nextgenthemes:
Why was this changed to v5? This looks like its ready and just the docs need to be updated? So you will bump to 5.0 when the docs are ready?

thehitechpanky:
Thanks for the reference. It would be awesome to see bootstrap independent of jQuery. Any estimated release date? And why not remove jQuery in v4 itself?

Johann-S:
@thehitechpanky because we broke everything and it's a major change in our API.
We plan to release that in 2019

thehitechpanky:
"broke everything"? Does that mean syntax is gonna change?

And thanks for the quick response.

Johann-S:
It means, in v5 you'll be able to use jQuery **if you want** and the syntax will be the same **but** if you don't want to use jQuery the syntax is different

thehitechpanky:
So what has been broken if both options are available?

Johann-S:
Compatibility and we plan to add other features too

Sebazzz:
> IE won't be supported

This is very sad, because even IE11 has kind-of OK support and also very much used in business. If this is only regarding the usage of ES6+, I suggest to downcompile to ES5.

iamandrewluca:
I am not sure, but I think he referred about web api's.  
And you'll have the possibility to polyfil them.

Sebazzz:
Right, per #25100 the opinion appears to be to drop IE10 only. That's fine, if you haven't upgraded to IE11 somehow by now you're a security risk to yourself and others.

corbinu:
@Sebazzz My understanding is IE 11 will be supported in v5 with jQuery. This PR will add the option to not use jQuery in v5 but that option will not support IE 11. That way users can choose what works for them.

Johann-S:
For those who wants to test that branch you can try it here: https://deploy-preview-23586--twbs-bootstrap4.netlify.com/

site/docs/4.1/assets/js/src/application.js
```diff
@@ -10,60 +10,96 @@
...
+(function () {
   'use strict'
 
-  $(function () {
+  document.addEventListener('DOMContentLoaded', function () {
     // Tooltip and popover demos
-    $('.tooltip-demo').tooltip({
-      selector: '[data-toggle="tooltip"]',
-      container: 'body'
+    var tooltipDemoList = [].slice.call(document.querySelectorAll('.tooltip-demo'))
```
XhmikosR: If we don't care about IE, we could just use `forEach` everywhere I think.

alekitto: Edge < 16 does not support forEach on `NodeList` objects, as well Safari < 10.
Is it planned to drop the support for these versions?

XhmikosR: Nope. I guess we could either polyfill that or add a helper method.

Johann-S:
```js
var tooltipDemoList = bootstrap.Util.makeArray(document.querySelectorAll('.tooltip-demo'))

**Comment by Johann-S** [comment]
Thanks @tmorehouse it's fixed now 👍

**Comment by thecodingdude** [comment]
Hi - could this work using browser modules/imports? The JS is es6 anyway.

**Comment by Johann-S** [comment]
@thecodingdude I think it's possible using directly our source code, but I have to try it myself 😁

**Comment by thehitechpanky** [comment]
Any plans / timeline for alpha / beta?

**Comment by Johann-S** [comment]
@thehitechpanky Not currently but one thing is sure ! It'll happen during this year 😉

**Comment by Ruffio** [comment]
Looking forward to this, so a lot of websites can be shrunked by at least 85 KB :-)

**Comment by webfolderio** [comment]
it's 2019, nobody needs bootstrap. Everybody should use css grid layout insteas of uber css library.

**Comment by ergo** [comment]
Any chance this can be implemented as custom elements, this way everything would be nicely compatible with all component frameworks. I think out-of-the-box IE/Edge compatibility could be a problem, but for example `lit-element` claims IE11+ compatibility and there are examples of wrapping systems like `material-web-components` with it (its 5kb gzipped dep), maybe it could be archived this way and we can finally get rid of `framework-bootstrap` packages and have canonical source shared across all solutions.

File: .browserslistrc [hunk 5]
```diff
@@ -6,7 +6,6 @@ not dead
 Chrome >= 45
 Firefox >= 38
 Edge >= 12
-Explorer >= 10
```
bcullman: IE11 will be supported by MS in some way under Oct 14th 2025. Consider changing to 
`Explorer > 10`

LifeIsStrange:
This should be correctly benchmarked, if it ends up not being faster than jquery then it would not be rational to merge this.

AlekseyKorzun:
Fiiiiireeee 🙌🙌🙏

ergo:
It is rational, it would enable easy wrapping into custom-elements without dependencies like jquery.

thorsten:
Any reason why you're not using TypeScript? :-)

.browserslistrc
dgrammatiko: Not really: https://techcommunity.microsoft.com/t5/Windows-IT-Pro-Blog/The-perils-of-using-Internet-Explorer-as-your-default-browser/ba-p/331732

iamandrewluca:
@thorsten first you need to have a setup for typescript, maybe it will come with time, but not now. Second, adding typescript with a feature branch is not a good idea. Typescript should be added in it's own feature branch. imo

Johann-S:
@LifeIsStrange you can test it here: https://deploy-preview-23586--twbs-bootstrap.netlify.com/ and it's pretty significant 😉 

@thorsten I don't think we'll rewrite everything (again 😆) in TypeScript in v5, but why not adding some types definitions for Typescript users 😉

iamandrewluca:
Bootstrap with jQuery typings
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bootstrap/index.d.ts

thecodingdude:
Hey @Johann-S is there any indication of how this will be compiled? Will we still have a .JS file to include? I wanted to spend some time seeing if it'd be possible to ship the src folder and let people import the features they need.

I also noticed certain files like [toast.js](https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/toast.js#L224) include jQuery references - is it worth creating different versions such as esm.js, jquery.js so users can pick the version they want?

Johann-S:
@thecodingdude Yep you'll have ton include one JS file

- this one **without** Popper.js inside: https://deploy-preview-23586--twbs-bootstrap.netlify.com/docs/4.3/dist/js/bootstrap.min.js
- Or this one **with** Popper.js inside: https://deploy-preview-23586--twbs-bootstrap.netlify.com/docs/4.3/dist/js/bootstrap.bundle.min.js

You can already import only plugins you need in v5, you'll find them in the `js/dist` folder and it'll be the same in v5, but the folder structure will change.

In v5 our plugins be able to work with or without jQuery, because if we detect jQuery we declare our Bootstrap plugin as a jQuery plugin.

We plan to ship an ESM version of Bootstrap too 👍

js/src/dom/manipulator.js
```diff
@@ -0,0 +1,109 @@
...
+  return val
+}
+
+function normalizeDataKey(key) {
+  return key.replace(regexDataKey, (chr) => chr.toLowerCase())
+}
+
+const Manipulator = {
+  setDataAttribute(element, key, value) {
+    element.setAttribute(`data-${normalizeDataKey(key)}`, value)
```
garygreen: Can't `element.dataset` be used here? It's supported in all browsers: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

js/src/dom/manipulator.js
```diff
@@ -0,0 +1,109 @@
...
+    element.removeAttribute(`data-${normalizeDataKey(key)}`)
+  },
+
+  getDataAttributes(element) {
+    if (typeof element === 'undefined' || element === null) {
+      return {}
+    }
+
+    let attributes
+    if (Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'dataset')) {
```
garygreen: Under what conditions would an element NOT have `dataset`  available? This whole block of code seems unnecessarily complex.

js/src/dom/selectorEngine.js
```diff
@@ -0,0 +1,100 @@
...
+  find(selector, element = document.documentElement) {
+    if (typeof selector !== 'string') {
+      return null
+    }
+
+    return find.call(element, selector)
+  },
+
+  findOne(selector, element = document.documentElement) {
+    if (typeof selector !== 'string') {
```
garygreen: Are these string checks really necessary? Seems like babysitting a bit?

garygreen (commented):
Only thing I would suggest is that polyfills should really be optional and not forced into every bundle.

I often see libraries make the mistake of forcefully including their own polyfills for *web standards* that just isn't necessary for evergreen browsers.

Imagine if every library out there included their own polyfills for `closest` or `promise` or `dataset` and other stuff. The web would be littered with duplicate functionality in bundles. I would prefer to have a separate polyfill file you can load.

js/src/dom/manipulator.js
Johann-S: Seems not, for Safari it's unknown 😟

js/src/dom/manipulator.js
Johann-S: some browsers don't have datasets like I mentionned above, we have to do some tests about Safari

Johann-S:
@garygreen I agree we should work on that to make those optional 👍

nikosgram:
> How large will be bootstrap.min.js without jQuery?

Probably really smaller, because jQuery is already a really heavy library, so it's an addon on downloading time. And most of the features that jQuery provide, es6 already implements.

I don't know though if bootstrap gonna support es6.

XhmikosR:
+parsing + execution time.

js/src/dom/manipulator.js
garygreen: Hmm that's odd, I just tested with Browserstack all the way to Mac Safari 5.1.9 (6534.59.8) (used on OSX Snow Leopard, released 9 years ago) and `dataset` appears to work fine:

image

garygreen: Also tested with iphones and works fine

z3ntu: https://caniuse.com/#feat=dataset

garygreen: So works in virtually all browsers including Safari except for SVG support

Seems unnecessary to roll own implementation of dataset when it's a web standard imo

amarkoulavkar:
Great move..when bootstrap 5 is releasing? Is there any new classes in BS 5??

js/src/dom/manipulator.js
Johann-S: I made some tests and it works on our supported browsers thanks @garygreen ! 👍 

Fixed here: https://github.com/twbs/bootstrap/pull/23586/commits/552f2905a577ac78b55700a95200e72005839916

js/src/dom/eventHandler.js
```diff
@@ -0,0 +1,330 @@
...
+const stripNameRegex = /\..*/
+const keyEventRegex  = /^key/
+const stripUidRegex  = /::\d+$/
+const eventRegistry  = {}   // Events storage
+let uidEvent         = 1
+const customEvents   = {
+  mouseenter: 'mouseover',
+  mouseleave: 'mouseout'
+}
+const nativeEvents   = [
```
garygreen: Is this list of every event the browser supports really needed for Bootstrap?

If it's just a case of checking if an event is natively supported, maybe something a simple as this would suffice:

```js
function isNative(eventName) {
  return 'on' + eventName in document.documentElement;
}
```

garygreen: Is this list of every event the browser supports really needed for Bootstrap?

If it's just a case of checking if an event is natively supported, maybe something a simple as this would suffice:

```js
function isNative(eventName) {
  return 'on' + eventName in document.documentElement;
}
```

Instead of:

```js
const nativeEvents = [
'click', 'dblclick', 'mouseup', 'mousedown', 'contextmenu',
  'mousewheel', 'DOMMouseScroll',
  'mouseover', 'mouseout', 'mousemove', 'selectstart', 'selectend',
  'keydown', 'keypress', 'keyup',
  'orientationchange',
  'touchstart', 'touchmove', 'touchend', 'touchcancel',
  'pointerdown', 'pointermove', 'pointerup', 'pointerleave', 'pointercancel',
  'gesturestart', 'gesturechange', 'gestureend',
  'focus', 'blur', 'change', 'reset', 'select', 'submit', 'focusin', 'focusout',
  'load', 'unload', 'beforeunload', 'resize', 'move', 'DOMContentLoaded', 'readystatechange',
  'error', 'abort', 'scroll'
];
```

Johann-S: It can works for some of them but not for all, for example it doesn't work for: `touchstart` or `mouseup` etc.

Akira0705:
Happy breaking up！

site/docs/4.3/getting-started/download.md
```diff
@@ -56,7 +55,7 @@ Install Bootstrap in your Node.js powered apps with [the npm package](https://ww
 npm install bootstrap
 {% endhighlight %}
 
-`require('bootstrap')` will load all of Bootstrap's jQuery plugins onto the jQuery object. The `bootstrap` module itself does not export anything. You can manually load Bootstrap's jQuery plugins individually by loading the `/js/*.js` files under the package's top-level directory.
+`require('bootstrap')` will load all of Bootstrap's plugins onto the `window` object. The `bootstrap` module itself does not export anything. You can manually load Bootstrap's plugins individually by loading the `/js/*.js` files under the package's top-level directory.
```
XhmikosR: This needs rewording @Johann-S

site/docs/4.3/getting-started/javascript.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: docs
 title: JavaScript
-description: Bring Bootstrap to life with our optional JavaScript plugins built on jQuery. Learn about each plugin, our data and programmatic API options, and more.
+description: Bring Bootstrap to life with our optional JavaScript plugins built on VanillaJS. Learn about each plugin, our data and programmatic API options, and more.
```
mdo:
```suggestion
description: Bring Bootstrap to life with our optional JavaScript plugins. Learn about each plugin, our data and programmatic API options, and more.
```

site/docs/4.3/getting-started/javascript.md
```diff
@@ -209,3 +209,45 @@ $('#yourTooltip').tooltip({
   }
 })
 {% endhighlight %}
+
+## Compatibility with IE 11
+
+Our v5 isn't designed to work with Internet Explorer 11, but you can add the following polyfills to make it work:
```
mdo:
```suggestion
Bootstrap v5 isn't designed to work with Internet Explorer 11, but you can add the following polyfills to make it work:
```

mdo: Reading this paragraph, I'd expect it to list out the polyfills instead of showing a code snippet. Can we add what the polyfill.io script does and make a bulleted list before the code snippet?

site/docs/4.3/getting-started/javascript.md
```diff
@@ -209,3 +209,45 @@ $('#yourTooltip').tooltip({
...
+
+{% capture callout %}
+Note that while we continue adding new features, it's possible that Internet Explorer 11 compatibility breaks in the future.
```
mdo: We can probably skip this—we don't support IE11, so we shouldn't have to warn against it possibly breaking in the future.

site/docs/4.3/getting-started/javascript.md
XhmikosR: Initially the plan was to indeed list the polyfills. But we opted for the auto-detection.

site/docs/4.3/getting-started/introduction.md
```diff
@@ -26,19 +26,16 @@ Copy-paste the stylesheet `<link>` into your `<head>` before all other styleshee
...
+Many of our components require the use of JavaScript to function. Specifically, they require our own JavaScript plugins and [Popper.js](https://popper.js.org/). Place the following `<script>`s near the end of your pages, right before the closing `</body>` tag, to enable them. Popper.js must come first, and then our JavaScript plugins.
 
 {% highlight html %}
-<script src="{{ site.cdn.jquery }}" integrity="{{ site.cdn.jquery_hash }}" crossorigin="anonymous"></script>
 <script src="{{ site.cdn.popper }}" integrity="{{ site.cdn.popper_hash }}" crossorigin="anonymous"></script>
 <script src="{{ site.cdn.js }}" integrity="{{ site.cdn.js_hash }}" crossorigin="anonymous"></script>
 {% endhighlight %}
 
-Curious which components explicitly require jQuery, our JS, and Popper.js? Click the show components link below. If you're at all unsure about the general page structure, keep reading for an example page template.
+Curious which components explicitly require our JS, and Popper.js? Click the show components link below. If you're at all unsure about the general page structure, keep reading for an example page template.
```
mdo:
```suggestion
Curious which components explicitly require our JavaScript and Popper.js? Click the show components link below. If you're at all unsure about the general page structure, keep reading for an example page template.
```

site/docs/4.3/getting-started/browsers-devices.md
```diff
@@ -121,7 +121,7 @@ For a list of some of the browser bugs that Bootstrap has to grapple with, see o
 
 ## Internet Explorer
 
-Internet Explorer 10+ is supported; IE9 and down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported in IE10, or require prefixed properties for full functionality. Visit [Can I use...](https://caniuse.com/) for details on browser support of CSS3 and HTML5 features. **If you require IE8-9 support, use Bootstrap 3.**
+Internet Explorer 11 is not supported officially. However, if you build Bootstrap yourself, you can add the needed CSS prefixes and follow the instructions on the [JavaScript page]({{ site.baseurl }}/docs/{{ site.docs_version }}/getting-started/javascript/#compatibility-with-ie-11). Please note that this might break at any time. **If you require Internet Explorer 10/11 support, you are better off using Bootstrap 4.**
```
mdo:
```suggestion
Internet Explorer 11 is not supported officially. However, if you build Bootstrap yourself, you can add the needed CSS prefixes and follow the instructions on the [JavaScript page]({{ site.baseurl }}/docs/{{ site.docs_version }}/getting-started/javascript/#compatibility-with-ie-11). Please note that this might break at any time. **If you require Internet Explorer 10/11 support, use Bootstrap 4.**
```

site/docs/4.3/assets/js/src/application.js
```diff
@@ -10,19 +10,23 @@
...
 
-/* global ClipboardJS: false, anchors: false, bsCustomFileInput: false */
+/* global ClipboardJS: false, anchors: false, bootstrap: false, bsCustomFileInput: false */
 
-(function ($) {
+(function () {
   'use strict'
 
-  $(function () {
+  document.addEventListener('DOMContentLoaded', function () {
```
XhmikosR: Might be worth double checking if we really need to wait until `DOMContentLoaded`.

XhmikosR (approved):

alekitto:
Boom! Awesome work!  🎉
Thanks to everyone who contributed!

XhmikosR:
Thanks to everyone!

I hope we didn't break a lot of things, we tried our best. That being said, please try the master branch and feel free to provide any doc tweaks or bugfixes.

We'll make a new issue with the remaining TODO tomorrow.

Johann-S:
Special thanks to @alekitto which helped me a lot to rewrite our plugins 👍

mdo:
Woohoo, well done y'all! <3

Kocal:
Nice job!
Finally we will be able to completely remove jQuery from our projects :clap:

ItaloBC:
Is this going to v5?

thehitechpanky:
Finally...waiting for it since very long. Thanks a lot everyone.

Razatastic:
Can't wait. Thank you Bootstrap team! 😄

3kynox:
Is it safe to use this 4.3.1 without jQuery in production? Or better to wait.

Sorry if it's an idiot question, I just came on this PR discussion, and accordingly to:
https://deploy-preview-23586--twbs-bootstrap.netlify.com/

We can use it already! No more jquery requirement

DarkGhostHunter:
From what I have read, Bootstrap without jQuery has been pushed to v5, so you will need to wait until it launches. Seems that v4 will still use jQuery.

v5 is not scheduled, but seeing by the PRs, it may come this year at best as long there is constant work on it.

XhmikosR:
You shouldn't use something which is in development on production, however dangerous it might be.

That being said, you can always experiment with the master branch. Just keep in mind, it's not done yet.

PR #23591, opened 2017-08-21
Tweak ESLint rules

XhmikosR:
* removed a few enabled rules which we are already using the default values
* enabled a few for strictness

@Johann-S: I know we disagreed last time on `undefined` but I still believe we should enable this. That being said, the other changes are good too, so I can revert that part if you insist.

We should also discuss which other rules we might enable. I know for sure I'd like semicolons, but it's a personal matter of taste plus this would conflict with your other PRs so it's something we should do in the future, if ever.

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -42,7 +42,7 @@ const Util = (($) => {
         if ($(event.target).is(this)) {
           return event.handleObj.handler.apply(this, arguments) // eslint-disable-line prefer-rest-params
         }
-        return undefined
+        return undefined // eslint-disable-line no-undefined
       }
     }
   }
@@ -55,7 +55,7 @@ const Util = (($) => {
     const el = document.createElement('bootstrap')
 
     for (const name in TransitionEndEvent) {
-      if (el.style[name] !== undefined) {
+      if (typeof el.style[name] !== 'undefined') {
         return {
           end: TransitionEndEvent[name]
         }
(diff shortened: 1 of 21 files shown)


bardiharborow (approved):
Looks good. The only thing is what's the rationale for disabling `no-else-return`, `eqeqeq`, `dot-notation` and `curly`, all of which don't seem to be included in `eslint:recommended`.

XhmikosR:
That is a mistake. I'm gonna restore them. Thanks for catching this.

js/src/util.js
```diff
@@ -42,7 +42,7 @@ const Util = (($) => {
         if ($(event.target).is(this)) {
           return event.handleObj.handler.apply(this, arguments) // eslint-disable-line prefer-rest-params
         }
-        return undefined
+        return undefined // eslint-disable-line no-undefined
```
XhmikosR: @Johann-S: maybe we should return `null` here?

js/src/util.js
```diff
@@ -42,7 +42,7 @@ const Util = (($) => {
         if ($(event.target).is(this)) {
           return event.handleObj.handler.apply(this, arguments) // eslint-disable-line prefer-rest-params
```
XhmikosR: Oh, also what's the reasoning for not using rest params here?

Johann-S:
I don't have any huge concern to switch from `element !== undefined` to `typeof element !== 'undefined'` 👍 
When I made that choice a long time ago that was because @cvrebert prefer that but for me that's ok

js/src/util.js
Johann-S: I don't know but that's related on how to bind special events to a native event with jQuery but you can make a try if you can still listen to `bsTransitionEnd` everything is fine

js/src/util.js
Johann-S: Sure 👍

js/src/util.js
bardiharborow: IIRC, I was just too scared to touch that at the time. No objection to changing as long as it doesn't break anything.

XhmikosR:
Let's just merge this as is then and we can look at the inline suppressions in another branch.

PR #23596, opened 2017-08-21
Removing jQuery from Alert

Johann-S:
I'm near the end for this plugin

The last thing I have to fix is : 
```javascript
  $(document).on(
    Event.CLICK_DATA_API,
    Selector.DISMISS,
    Alert._handleDismiss(new Alert())
  )
```
And how folks will user our plugins without jQuery, maybe we should enable to register our plugins to jQuery even if we do not use it anymore 🤔 

Work in progress...
@XhmikosR if you want to help me do not hesitate 😄

### DIFF
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -163,12 +163,7 @@ const Alert = (($) => {
    * Data Api implementation
    * ------------------------------------------------------------------------
    */
-
-  $(document).on(
-    Event.CLICK_DATA_API,
-    Selector.DISMISS,
-    Alert._handleDismiss(new Alert())
-  )
+  EventHandler.on(document, Event.CLICK_DATA_API, Selector.DISMISS, Alert._handleDismiss(new Alert()))
 
 
   /**
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -73,7 +73,7 @@ const Alert = (($) => {
       const rootElement = this._getRootElement(element)
       const customEvent = this._triggerCloseEvent(rootElement)
 
-      if (customEvent.defaultPrevented) {
+      if (customEvent.isDefaultPrevented()) {
         return
       }
 
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -73,7 +73,7 @@ const Alert = (($) => {
       const rootElement = this._getRootElement(element)
       const customEvent = this._triggerCloseEvent(rootElement)
 
-      if (customEvent.isDefaultPrevented()) {
+      if (customEvent.defaultPrevented) {
         return
       }
 
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -1,3 +1,6 @@
+import Data from './dom/data'
+import EventHandler from './dom/eventHandler'
+import SelectorEngine from './dom/selectorEngine'
 import Util from './util'
 
 
@@ -70,7 +73,7 @@ const Alert = (($) => {
       const rootElement = this._getRootElement(element)
       const customEvent = this._triggerCloseEvent(rootElement)
 
-      if (customEvent.isDefaultPrevented()) {
+      if (customEvent.defaultPrevented) {
         return
       }
 
@@ -78,7 +81,7 @@ const Alert = (($) => {
     }
 
     dispose() {
-      $.removeData(this._element, DATA_KEY)
+      Data.removeData(this._element, DATA_KEY)
       this._element = null
     }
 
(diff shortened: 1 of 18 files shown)


Johann-S:
It seems on Windows 8 and 8.1 for IE10 and IE11 `CustomEvent` produce errors 😢I'll add a polyfill

js/src/dom/selectorEngine.js
```diff
@@ -0,0 +1,66 @@
...
+    return element.closest(selector)
+  }
+}
+
+const SelectorEngine = {
+  matches(element, selector) {
+    return fnMatches.call(element, selector)
+  },
+
+  find(selector) {
```
XhmikosR: Perhaps we should have a `find` method for `querySelector`? It should be faster than` querySelectorAll` especially if we know we want one element.

XhmikosR: BTW, what do you think about adding helper methods around qsa, qs, etc? Instead of us calling the whole thing across the codebase this should result in shorter minified code.

Johann-S: Maybe a `findOne` method should be a good thing with `querySelector` and why not adding helpers methods but I prefer longer method because it's simple to know what they do

Johann-S: Do you have any ideas why it failed on IE ? I added a CustomEvent polyfill

XhmikosR: I still think we should have our helpers around qsa, qs, etc. The minified code will be a lot shorter that way. We can still name them `Util.getElementById` etc.

Johann-S: You should make a PR when this one will be merged 👍

XhmikosR: Did you try locally with IE?

Johann-S: Found on my local IE thanks 👍 btw I hate to debug on IE 😢

Johann-S:
I win against IE 💪 😆

XhmikosR:
IE keeps bugging us :/ Glad you solved it I'll try to have a look tomorrow.
BTW perhaps we should have more tests now for these cases.

js/src/dom/selectorEngine.js
```diff
@@ -0,0 +1,66 @@
...
+    if (typeof selector !== 'string') {
+      return null
+    }
+
+    let selectorType = 'querySelectorAll'
```
XhmikosR: After looking into this file I still believe we should have a wrapper generic function and instead do what I suggested and just provide qsa, qs, etc. It's one less thing for the engine to do so it should be as fast possible.

js/tests/unit/alert.js
```diff
@@ -64,16 +64,19 @@ $(function () {
...
-        done()
-      })
-      .on('closed.bs.alert', function () {
-        assert.ok(false, 'closed event fired')
-      })
-      .bootstrapAlert('close')
+    var $alert = $('<div class="alert"/>')
+    $alert.appendTo('#qunit-fixture')
+
+    EventHandler.on($alert[0], 'close.bs.alert', function (e) {
```
XhmikosR: So, things still work fine with jQuery?

package.json
```diff
@@ -33,7 +33,7 @@
     "js-lint": "eslint js/ && eslint --config js/tests/.eslintrc.json --env node build/ Gruntfile.js",
     "js-lint-docs": "eslint --config js/tests/.eslintrc.json assets/js/",
     "js-compile": "npm-run-all --parallel js-compile-*",
-    "js-compile-bundle": "shx cat js/src/dom/event.js js/src/util.js js/src/alert.js js/src/button.js js/src/carousel.js js/src/collapse.js js/src/dropdown.js js/src/modal.js js/src/scrollspy.js js/src/tab.js js/src/tooltip.js js/src/popover.js | shx sed \"s/^(import|export).*//\" | babel --filename js/src/bootstrap.js | node build/stamp.js > dist/js/bootstrap.js",
+    "js-compile-bundle": "shx cat js/src/dom/eventHandler.js js/src/dom/selectorEngine.js js/src/dom/data.js js/src/util.js js/src/alert.js js/src/button.js js/src/carousel.js js/src/collapse.js js/src/dropdown.js js/src/modal.js js/src/scrollspy.js js/src/tab.js js/src/tooltip.js js/src/popover.js | shx sed \"s/^(import|export).*//\" | babel --filename js/src/bootstrap.js | node build/stamp.js > dist/js/bootstrap.js",
```
XhmikosR: Justthinking out loud: this is getting too big. Would it be possible to move this outside of package.json maybe to an external json file file or something?

Johann-S: Yes that would be better if we have other file to add

js/tests/unit/alert.js
Johann-S: Yep everything is fine

js/src/dom/selectorEngine.js
Johann-S: I'm not against it but that's not something to prioritize here we should focus on event namespacing

package.json
XhmikosR: Yes, but this is just planning ahead. We'll definitely need it and this just seems like bad planning to me.

Johann-S: With my PR which improve our modularisation I introduced a bundler (Rollup) which handle that pretty well

js/src/alert.js
```diff
@@ -90,55 +93,50 @@ const Alert = (($) => {
       let parent     = false
 
       if (selector) {
-        parent = $(selector)[0]
+        const tmpSelected = SelectorEngine.find(selector)
```
XhmikosR: Do we really need this temp var?

Johann-S: No we can remove this var 👍

js/src/dom/selectorEngine.js
XhmikosR: @Johann-S: maybe we could make `find` accept an element parameter and if missing use `document`?

Regarding `findOne` since we already have a function I think it might be better if we just filter the selector and if it's one then call `querySelector`. For all this we should do some benchmarks because I'm not sure if it's worth in this case. If we called it directly it should be faster for sure.

js/src/alert.js
```diff
@@ -90,55 +93,50 @@ const Alert = (($) => {
...
-        .remove()
+      EventHandler.trigger(element, Event.CLOSED)
+      element.parentNode.removeChild(element)
     }
 
 
     // static
 
     static _jQueryInterface(config) {
       return this.each(function () {
```
XhmikosR: I wonder how this still works. `each` is jQuery specific, isn't it?

Johann-S: Yes but this method is called only when jQuery is present

XhmikosR:
Yup I just saw after I made the comment:)  I'm just getting a TypeError in
button but we'll see when time comes.

js/src/dom/selectorEngine.js
Johann-S: Yes I agree for your two proposals 👍

PR #23600, opened 2017-08-21
button: Convert to vanilla JS

XhmikosR:
Tasks:

- [x] Convert buttons.js
- [ ] Fix tests

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -75,7 +75,7 @@ const Button = (() => {
       )
 
       if (rootElement) {
-        const input = SelectorEngine.findOne(this._element, Selector.INPUT)
+        const input = SelectorEngine.findOne(Selector.INPUT, this._element)
 
         if (input) {
           if (input.type === 'radio') {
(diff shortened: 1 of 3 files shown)


Johann-S:
it seems your tests failed because phantomsJS don't know `closest` function and the polyfill is in my PR, before finishing this PR we should finish the Alert plugin and the Button plugin will be finish quickly after that 👍

Johann-S:
`v4-without-jquery` is up to date with `v4-dev` so you can rebase 👍

js/src/dom/selectorEngine.js
```diff
@@ -46,7 +46,7 @@ const SelectorEngine = {
     return fnMatches.call(element, selector)
   },
 
-  find(selector) {
+  find(element = document, selector) {
```
Johann-S: Nice change 👍 I did the same but without the default value and your change is much better 😄

Johann-S:
Can you update our button visual test ?

js/src/button.js
```diff
@@ -148,39 +151,56 @@ const Button = (($) => {
...
+      }
+      data.toggle()
     })
-    .on(Event.FOCUS_BLUR_DATA_API, Selector.DATA_TOGGLE_CARROT, (event) => {
-      const button = $(event.target).closest(Selector.BUTTON)[0]
-      $(button).toggleClass(ClassName.FOCUS, /^focus(in)?$/.test(event.type))
+
+  EventHandler.on(document,
+    Event.FOCUS_BLUR_DATA_API, Selector.DATA_TOGGLE_CARROT, (event) => {
+      const button = SelectorEngine.closest(event.target, Selector.BUTTON)[0]
```
XhmikosR: @Johann-S: this `[0]` indeed doesn't seem right.

Issue #23450, opened 2017-08-15
Cannot customize button border color


dozer75:
In the current _buttons.scss file, the same color is sent in as background and border color. This was not the case earlier where we could specify in our customization specific border and background colors.

We need to be able to customize border colors separate from the background. Otherwise why have borders at all?

I find this related to #23418

andresgalante:
Although the border is defined at the mixin:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_buttons.scss#L9

It's set as the same color of the background here:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_buttons.scss#L55

I think this is a bug. The solutions I can think of are:
- Make the border a darker shade of the background and keep this very nice definition of btns based on just one color set in the maps.
- Create a border variables (I don't like it)

What do you think?

wojtask9:
I described the same problem here https://github.com/twbs/bootstrap/issues/23126

dozer75:
@andresgalante  In my opinion this is one of the weakness with the new variable system. It is structured a lot better than earlier, but it isn't as flexible as it was, limiting us to either accept the default variations of basic color settings or to be forced to create our own definitions of the classes which I think will create more confusion, larger style sheets, more error prone and harder to read.

As for your suggestions, I think that the first one isn't an option. In our case, we want the background to be one color and the outer to be a completely other one ($white background and $blue border) and as far as I understand your suggestion it won't give me that.

The only way this can be solved for the usage I have is to be able to configure colors by themselves, and the only way is to have variables. However, instead of having lots of key/value pairs, why not use key/array structures instead? Like this: (I do not know scss that well, so just look at this example as pseudo code)

```
$buttons = (
    primary: (
        color: $white,
        background: theme-color("primary"),
        border: theme-color("primary"),        
        /* and other button related parameters */
   )
)
```

Then generate buttons based on this with:

```
@each $color, $values in $buttons {
  .btn-#{$button} {
    @include button-variant($values);
  }
}
```

and let `button-variant` read the configuration from `$values`?

As I said, it may not work since I don't know scss that well, but if it works it's worth a try.

I did get the issue described by @wojtask9 today and was planning to check it out tomorrow but it is clearly already registered. But I think that a similar construct like this (if it works) could be applied to lots of the similar issues.

andresgalante:
@dozer75 you are absolutely right it is limiting and my solution wouldn't help much.
[...]

PR #23611, opened 2017-08-22
New Theming docs page

### LINKED ISSUE #22891: _custom.scss alternatives

mdo:
This new docs page is intended to answer questions for where to put custom CSS (see #22891), how to customize and extend Sass maps and variables, and more. Ideally it'll replace the Options page. Much more still to come, but would love early feedback.

Closes #22891.


mdo:
Still needs something to solve #23418 and #23450.

tkuther:
I think the new doc isn't clear enough at this part:

> For example, to modify an existing color in our `$theme-colors` map, add the following to your custom Sass file:
> ￼ $theme-colors: (
> ￼   "primary": $red,
> ￼   "danger": $orange
> ￼ );

Where in my custom.scss would I put this? If I put this before the `@import bootstrap/scss/variables;` I also need to define $red and $orange first (because they are defined in _variables.scss which I import later on)

If I put it after the `@import bootstrap/scss/variables;`, $red and $orange are defined, but my $theme-colors override won't work because it's !default in _variables.scss which requires it to be redefined before the import.

mdo:
@tkuther Does the variable section cover that for you? Might just need another link to refer back to that section to connect Sass maps to that section since it's not immediately clear maps and variables are connected.

https://github.com/twbs/bootstrap/blob/0071bf52798515877bf1797e4332a11d92f1f7bb/docs/4.0/getting-started/theming.md#L62-L79

tkuther:
@mdo it does cover it, sure.

I'm just saying the example isn't clear because it reads "add following to your custom Sass file", but it actually requires two lines more to get it working that way. A link back to the variable defaults section would be enough to make it more clear, or even provide a full example like
```
$red: #f44336;
$orange: #ff9800;

$theme-colors: (
￼ "primary": $red,
￼ "danger": $orange
￼ );
```
I guess that could avoid some head scratching.

mdo:
Moved away from those variables—they were meant to pull from the included color variables in Bootstrap, but I can see the confusion. Moved to hex values.

Issue #23557, opened 2017-08-18
Split in multiple npm packages (was "Change peerDependencies to optionalDependencies") 


tkrotoff:
Edit: see comments below
> So, 3 packages, and make `bootstrap` depend on the other 2, `bootstrap-css` and `bootstrap-js`

<br>
<br>

-----

jQuery and Popper.js are defined as `peerDependencies`: https://github.com/twbs/bootstrap/blob/v4.0.0-beta/package.json#L69

Bootstrap can be used without the JS part (think of [reactstrap](https://github.com/reactstrap/reactstrap), [React-Bootstrap](https://github.com/react-bootstrap/react-bootstrap), [Angular UI Bootstrap](https://github.com/angular-ui/bootstrap), [NG Bootstrap](https://github.com/ng-bootstrap/ng-bootstrap), [BootstrapVue](https://github.com/bootstrap-vue/bootstrap-vue)...)

~~Suggestion: switch to [`optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies) so there is no warning when not installing jQuery and Popper.js.~~

```
$ yarn
...
warning "bootstrap@4.0.0-beta" has unmet peer dependency "jquery@>=3.0.0".
warning "bootstrap@4.0.0-beta" has unmet peer dependency "popper.js@^1.11.0".
```

XhmikosR:
@Johann-S: what do you think? We should probably add a check if it doesn't exist for popper.js. For jQuery we already check.

Johann-S:
Not sure because we switch from `dependencies` to `peerDependencies` because folks complaint about having unwanted libs installed... 
And with `optionalDependencies` (if I understand correctly) it will be the same because npm will try to install those libs

XhmikosR:
I think we should have used optional deps instead of peer deps.

tkrotoff:
@Johann-S 
> with `optionalDependencies` [...] npm will try to install those libs

Sure? I did some tests with npm and Yarn and now I don't even understand how `optionalDependencies` works and its purpose:

Bootstrap package.json:
```JSON
"optionalDependencies": {
  "jquery": ">=3.0.0",
  "popper.js": "^1.11.0"
}
```

Inside the package that uses Bootstrap (let's call it the "client"):
- With npm 5.3.0, jquery and popper.js are not present inside `node_modules` (*) + not present at root level of `package-lock.json` => OK
[...]

PR #23614, opened 2017-08-22
[WIP] Multiple packages using Lerna

peternewnham:
Here's an early proof of concept of how splitting into multiple packages using [Lerna](https://lernajs.io/) could work (#23557)

Created 3 packages located in the `packages` directory:
- bootstrap
  - devDependency on the css and jquery packages below
  - should be backwards compatible with current bootstrap package
- css (scss and css only)
- jquery (jquery plugins only)

Usage:
- Install  [Lerna](https://lernajs.io/) (`npm i -g lerna`)
- Run `lerna bootstrap` to initialise and link dependencies between packages
` Run `lerna run build` to run the build script for each package

Demos of them in action (published to a temporary scope):
- Full Bootstrap package (css + jquery) - https://jsfiddle.net/hpm0ku0c/1/
- CSS only - https://jsfiddle.net/n81gadbt/1/

To Do:
- Decide on package names 
  - scoped (@twbs/css) vs flat (bootstrap-css)
  - Consistent names between platforms Bootstrap is published on (npm, packagist, nuget... etc)
- Restructure rest of project and run scripts to work with the new format
- Consider the way the combined `bootstrap` package is generated (may be a better way than just copying files from it's dependencies?)
- Probably lots of other things I've not thought about

CC: @XhmikosR


Johann-S:
A `bootstrap-js` would be a good things too

wrakky:
`bootstrap-js` would be awesome if there are plans to have non jquery dependent plugins but my initial thought was to just port the current infrastructure to separate packages initially and then splitting them into extra packages can be done at a later date.

You could have:
- `bootstrap-js` - native js plugins
- `bootstrap-jquery`-  jquery wrappers for `bootstrap-js`
- Separate `grid` and `reboot` css packages
- `bootstrap-css` which uses `grid` and `reboot`

You could even split each individual JS plugin into it's own package if you wanted to be super granular and then projects won't even need the dependency on popper.js in your app if you don't need to use popovers or tooltips

Johann-S:
I agree that's a good long term vision 👍 
We will move from jQuery at v4.3 to let us the time to rewrite our plugins

FezVrasta:
Why jQuery and Popper.js set as `devDependencies`? They are needed at runtime

wrakky:
@FezVrasta they are peerDependencies (just like in the v4-dev branch)

FezVrasta:
I must have mixed up something in my head lol, ok great!

XhmikosR:
Well, that's why we say do not modify the dist files.

The end goal is to see how this works in our case and if we are going to go with it or not.

wrakky:
I've rebased now so the conflicts are gone (but things may no longer work as I haven't had time to test anything).

XhmikosR:
Unfortunately it doesn't seem to work for me.

The individual packages' package.json scripts need lots of work.

mdo:
Closing as stale. Happy to revisit in the future though, so please open an issue and we can discuss. <3

PR #23617, opened 2017-08-22
remove d-flex class from img tag in docs #23609

### LINKED ISSUE #23609: v4 Docs on Media objects are confusing and misleading.
I was a little bit confused when reading the Media Object documentation on the Bootstrap v4 docs that can be [found here](https://getbootstrap.com/docs/4.0/layout/media-object/#alignment).

In all of the examples, the author has added a `.d-flex` class on the `` elements. As I understand this class, it's the utility function that make the item, and any children of it be displayed as flex box. In all of the examples, this does not seem to be necessary. You can self-align the image no matter if it includes a `.d-flex` class or not. 

See example of using `.d-flex` and leaving it out:
[https://jsfiddle.net/ounyeh3p/](https://jsfiddle.net/ounyeh3p/)

My suggestion is to update the docs and remove the class since it's not needed and it has no effect on the content at all. Obviously this is **not** saying that you should remove any utility classes such as the margin or padding ones, they are used in a correct way.

chiraggmodi:
have remove all `.d-flex` class from `img` tag.
i have check in IE 10 and 11 and didn't found any bug related to `d-flex` class 

Fixes #23609

### DIFF
--- a/docs/4.0/layout/media-object.md
+++ b/docs/4.0/layout/media-object.md
@@ -14,7 +14,7 @@ Below is an example of a single media object. Only two classes are required—th
 
 {% example html %}
 <div class="media">
-  <img class="d-flex mr-3" data-src="holder.js/64x64" alt="Generic placeholder image">
+  <img class="mr-3" data-src="holder.js/64x64" alt="Generic placeholder image">
   <div class="media-body">
     <h5 class="mt-0">Media heading</h5>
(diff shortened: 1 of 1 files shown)


mdo:
For those curious, this isn't needed because `display: flex` (set via `.d-flex`) affects the children elements within. Given images have no contents, it was unnecessary.

mdo (approved):

PR #23625, opened 2017-08-22
Update which required jQuery release folks should use

Johann-S:
We already changed that in our `package.json` so we should do the same and warn folks which do not use jQuery 3


XhmikosR:
@Johann-S: I'd say we should do this for all other places.

* bootstrap.nuspec
* bootstrap.sass.nuspec
* bower.json

XhmikosR:
And maybe we should change the version range on package.json to reflect the fact that we need <4.

Johann-S:
That's true 👍

Johann-S:
Is it ok for you now @XhmikosR ?

XhmikosR (approved):

XhmikosR:
@mdo: not sure in which section to mention this change in the ship notes, so feel free to add it yourself.

PR #23634, opened 2017-08-23
Allow easy customization of docs variables

m5o:
* append `!default` to local docs variables

---
Hey bootstrappers 👋 

I'm generating some corp. docs and want to override the 
default docs variables, without touching the default vars.
Seems some `!default` are missing here.

### DIFF
--- a/assets/scss/_variables.scss
+++ b/assets/scss/_variables.scss
@@ -1,11 +1,11 @@
 // Local docs variables
-$bd-purple:          #563d7c;
-$bd-purple-bright:   lighten(saturate($bd-purple, 5%), 15%);
-$bd-purple-light:    #cdbfe3;
-$bd-purple-lightest: #f5f2f9;
-$bd-graphite:        #2a2730;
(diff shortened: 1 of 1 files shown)


mdo:
These aren't meant to be customized like this, thanks though. Adding `!default` doesn't make much sense—the variables would still be named "bootstrap docs purple" :).

m5o:
@mdo 
Ok, of course the naming of the variables could be improved as well. You should follow your own color system ;) Like: `$bd-primary: #563d7c;`

The 'Sass Colour Function Calculator' [1] is a nice helper to calculate other variations in colour.

```scss
$bd-primary:          #563d7c !default;
$bd-primary-bright:   lighten(saturate($bd-primary, 5%), 15%) !default;
$bd-primary-light:    lighten(saturate($bd-primary, 5.08%), 45.69%) !default;
$bd-primary-lightest: lighten(saturate(adjust-hue($bd-primary, 2%), 2.79%), 60%) !default;
```
[1] http://razorltd.github.io/sasscolourfunctioncalculator/

PR #23645, opened 2017-08-23
Add rounded prefixed css classes for each side with zero radius sizes

juanpabloprado:
Add top, right, bottom and left variants for achieving zero radius sizes.


mdo:
We'll punt on this for now, thanks though!

PR #23652, opened 2017-08-24
Fix navbar IE

### LINKED ISSUE #23531: Navbar content goes off page on IE / Windows
Hi,

on Windows 7, using Internet Explorer 10, and bootstrap 4 (beta) the navbar content goes off page. 

The easiest way to replicate the behavior is

**Option 1**

open Bootstrap's [fixed-navbar example](https://getbootstrap.com/docs/4.0/examples/navbar-top-fixed/) on the aforementioned combination of browser / OS. The rendered reply looks like in [this picture](https://www.browserstack.com/screenshots/c3197480c5c5ef0cfb7325d81ef945a08968f5c7/win7_ie_10.0.png) i.e. the search button goes off the page.

In case it is helpful - the same Bootstrap example works fine with other browser/OS combinations e.g. Edge14 / Windows 10, IE 11 / Windows 10.

**Option 2**

You can try to open the aforementioned [example](https://getbootstrap.com/docs/4.0/examples/navbar-top-fixed/) also on [Browserstack](https://developer.microsoft.com/en-us/microsoft-edge/tools/screenshots/)

win7_ie_10 0

Cheers,
     Thomas

MartijnCuppens:
Fixes #23531.


XhmikosR:
@MartijnCuppens: does this change the behavior with other browsers?

XhmikosR:
I can confirm it does fix the issue on IE 10.

/CC @mdo

XhmikosR (approved):

MartijnCuppens:
@XhmikosR, should be ok (tested in latest versions of Edge, Chrome, FF). `flex-grow: 1` also is a more logic choice over `flex-basis: 100%` because the element will never be 100% but just needs to fill the available space.

XhmikosR:
Indeed, makes more sense.

I'll let @mdo take the final decision.

PR #23658, opened 2017-08-24
Removing jQuery from Carousel

Johann-S:
Rewrite our Carousel plugin without jQuery

- ~Improve `off` method of our EventHandler to remove listeners in an entire namespace~

Work in progress...

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -89,6 +89,56 @@ const Util = (() => {
           }
         }
       }
+    },
+
+    extend(obj1, obj2) {
+      for (const secondProp in obj2) {
+        if (Object.prototype.hasOwnProperty.call(obj2, secondProp)) {
+          const secondVal = obj2[secondProp]
+          // Is this value an object?  If so, iterate over its properties, copying them over
+          if (secondVal && Object.prototype.toString.call(secondVal) === '[object Object]') {
+            obj1[secondProp] = obj1[secondProp] || {}
+            Util.extend(obj1[secondProp], secondVal)
+          } else {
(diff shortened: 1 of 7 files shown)


js/src/util.js
```diff
@@ -88,6 +88,22 @@ const Util = (() => {
           }
         }
       }
+    },
+
+    extend(obj1, obj2) {
```
wojtask9: Why not use Object.assign?
I thing another issue is that this implementation is `deepCopy` but $.extend by default is not recursive.

wojtask9: If Object.assign cannot be used I would use `arguments` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) and then you can change `Util.extend(Utils.extend...` to `Util.extend(obj1, obj2, obj3, ...)`

Johann-S: I didn't had time to think more of this implementation of `extend` but it did the job 😄
I'll take a look thanks @wojtask9 👍 
If you want you can submit a PR which target this branch (`v4-carousel-without-jquery`) with a better implementation

Johann-S:
last issue here, when I call `dispose` the first time all the event listeners were removed, but if click on the `next` arrow (a new carousel is created) and if I call `dispose` again the listeners are still alive 😢

Johann-S:
Now I have to be able to trigger events with parameters inside to be able to run our unit test for our Carousel plugins

Johann-S:
All test passed except one from our Button plugin but it's a known issue

PR #23663, opened 2017-08-25
Changes alerts to improve accessibility

andresgalante:
This PR is a simple order change of the close btn on alerts that improves the way a screen reader reads them. 

Instead of going:

>"Close, Holy guacamole! You should check in on some of those fields below, alert, Close, button"

now it goes:

>"Holy guacamole! You should check in on some of those fields below, Close, alert, Close, button"

Which IMO it makes more sense.


andresgalante:
Need to work on the CSS before it's ready

Issue #22866, opened 2017-06-19
Collapse click handler should only call preventDefault when a target was found


nelisbijl:
version 4.0.0-alpha.6

event.preventDefault is called regardless of whether a target is found.
In case no target exists, it prevents other code from being executed

https://codepen.io/anon/pen/eRvgYq

Johann-S:
Fixed in our current code base see : https://github.com/twbs/bootstrap/blob/v4-dev/js/src/collapse.js#L364
And in Alpha6 : https://github.com/twbs/bootstrap/blob/v4.0.0-alpha.6/js/src/collapse.js#L360

nelisbijl:
No it is not, have you tried it?

Johann-S:
Yes my bad 👍

PR #23666, opened 2017-08-25
Collapse - preventDefault only on <a> elements not inside the collapse element

Johann-S:
Should prevent default only on `<a>` elements not inside the collapse element

/CC @nelisbijl
Close: #22866

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -361,7 +361,8 @@ const Collapse = (($) => {
    */
 
   $(document).on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-    if (!/input|textarea/i.test(event.target.tagName)) {
+    // preventDefault only for <a> elements (which change the URL) not inside the collapsible element
+    if (event.target.tagName === 'A' && !$.contains(this, event.target)) {
       event.preventDefault()
     }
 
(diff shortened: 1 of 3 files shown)


js/tests/visual/collapse.html
```diff
@@ -57,7 +57,7 @@ <h5 class="mb-0">
       </div>
     </div>
 
-    <script src="../../../docs/assets/js/vendor/jquery-slim.min.js"></script>
```
XhmikosR: Can you go through all visual test files and see if there are more references like that and replace all of them?

js/src/collapse.js
```diff
@@ -361,7 +361,8 @@ const Collapse = (($) => {
    */
 
   $(document).on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-    if (!/input|textarea/i.test(event.target.tagName)) {
+    // preventeDefault only for <a> elements (which change the URL)
```
XhmikosR: `preventDefault`

XhmikosR (changes requested):

js/tests/visual/collapse.html
Johann-S: It was the only one I forget 😄

XhmikosR (approved):

PR #23669, opened 2017-08-25
Fix sticky margin when a modal is opened

### LINKED ISSUE #23661: Sticky navbar with container being shifted slightly to the left upon modal opened
Link to demo: https://jsfiddle.net/techdavid/g89ts0rs/1/

When opening a modal, where the page contains a sticky navbar with a container inside it, the container gets shifted slightly to the left due to the padding being applied to the body element.

The `.sticky-top` element gets a right-padding of 33px added, however there is also white space to the right of this due to the padding on the body. The `.sticky-top` element should also have a margin-right of -33px applied to compensate for this.

I think that when PR #18441 was submitted, it maybe wasn't tested with `.sticky-top` (only `.fixed-top`) and so this is why the problem is still sometimes occurring. I checked the source of the JS file I'm using and found that it does contain code from this PR.

Edit: checked to confirm the problem is only occurring with sticky-top (not fixed-top).

---

Before opening an issue:

- [Search for duplicate or closed issues](https://github.com/twbs/bootstrap/issues?utf8=%E2%9C%93&q=is%3Aissue) => supposedly fixed in #18441 but still broken in the latest beta
- [Validate](https://validator.w3.org/nu/) and [lint](https://github.com/twbs/bootlint#in-the-browser) any HTML to avoid common problems => done
- Prepare a [reduced test case](https://css-tricks.com/reduced-test-cases/) for any bugs => https://jsfiddle.net/techdavid/g89ts0rs/1/
- Read the [contributing guidelines](https://github.com/twbs/bootstrap/blob/master/CONTRIBUTING.md) => done

When asking general "how to" questions:

- Please do not open an issue here
- Instead, ask for help on [StackOverflow, IRC, or Slack](https://github.com/twbs/bootstrap/blob/master/README.md#community)

When reporting a bug, include:

- Operating system and version => Windows 10
- Browser and version => Chrome 60
- Reduced test cases and potential fixes using [JS Bin](https://jsbin.com) => https://jsfiddle.net/techdavid/g89ts0rs/1/

When suggesting a feature, include:

- As much detail as possible for what we should add and why it's important to Bootstrap
- Relevant links to prior art, screenshots, or live demos whenever possible

valenfys:
Fixes #23661

### DIFF
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -68,6 +68,7 @@ const Modal = (($) => {
     DATA_TOGGLE        : '[data-toggle="modal"]',
     DATA_DISMISS       : '[data-dismiss="modal"]',
     FIXED_CONTENT      : '.fixed-top, .fixed-bottom, .is-fixed, .sticky-top',
+    STICKY_CONTENT     : '.sticky-top',
     NAVBAR_TOGGLER     : '.navbar-toggler'
   }
 
@@ -441,6 +442,13 @@ const Modal = (($) => {
           $(element).data('padding-right', actualPadding).css('padding-right', `${parseFloat(calculatedPadding) + this._scrollbarWidth}px`)
         })
 
+        // Adjust sticky content margin
+        $(Selector.STICKY_CONTENT).each((index, element) => {
+          const actualMargin = $(element)[0].style.marginRight
+          const calculatedMargin = $(element).css('margin-right')
+          $(element).data('margin-right', actualMargin).css('margin-right', `${parseFloat(calculatedMargin) - this._scrollbarWidth}px`)
+        })
+
         // Adjust navbar-toggler margin
         $(Selector.NAVBAR_TOGGLER).each((index, element) => {
           const actualMargin = $(element)[0].style.marginRight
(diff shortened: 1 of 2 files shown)


js/src/modal.js
```diff
@@ -68,6 +68,7 @@ const Modal = (($) => {
     DATA_TOGGLE        : '[data-toggle="modal"]',
     DATA_DISMISS       : '[data-dismiss="modal"]',
     FIXED_CONTENT      : '.fixed-top, .fixed-bottom, .is-fixed, .sticky-top',
+    STICKY_CONTENT     : '.sticky-top',
```
Johann-S: `.sticky-top' is in `FIXED_CONTENT` so I don't see the interest to add a new constant here :thinking:

js/src/modal.js
```diff
@@ -464,8 +472,8 @@ const Modal = (($) => {
         }
       })
 
-      // Restore navbar-toggler margin
-      $(Selector.NAVBAR_TOGGLER).each((index, element) => {
+      // Restore sticky content and navbar-toggler margin
+      $(Selector.STICKY_CONTENT).add(Selector.NAVBAR_TOGGLER).each((index, element) => {
```
Johann-S: Maybe here you should do something like that : 
```javascript
$(`${Selector.STICKY_CONTENT} ${Selector.NAVBAR_TOGGLER}`)
```
instead of `.add`

js/src/modal.js
techdavid: Because sticky elements also need the padding which was and still is applied to all fixed elements, in addition to the new margin, to have the correct position. However, I could change the references to STICKY_CONTENT to the string `.sticky-top` if you want.

Johann-S: No that's fine I just wanted to understand your thinking 👍☺️

Johann-S (approved):
LGTM 👍

PR #23673, opened 2017-08-25
Changes alerts close btn to improve accessibility

### LINKED ISSUE #21889: Alert close button overflowing its container causing extra space to appear below
This is a continuation of #21733 with working examples, comment in SCSS for fix:

https://jsfiddle.net/kpyh492r/5/

The bug appears in all major browsers.

image

andresgalante:
Changes alerts close btn position on the markup to improve accessibility

This PR is a simple order change of the close btn on alerts that improves the way a screen reader reads them.

Instead of going:

>"Close, Holy guacamole! You should check in on some of those fields below, alert, Close, button"

now it goes:

>"Holy guacamole! You should check in on some of those fields below, Close, alert, Close, button"

Which IMO it makes more sense.

Plus:
- If the notification has an `a` a screen reader will tab order first the link then the close btn
- this solution is more flexible since the user can put the btn markup anywhere and it'll render always on the top right corner disregarding the amount of text that the alert has.

Fixes #21889.


patrickhlauke (approved):

patrickhlauke:
This also looks good to me, leaving it up to @mdo for final say though.

mdo:
It's unlikely to happen, but as it stands, someone could run into an issue where the dismiss icon overlays the text of the alert.

screen shot 2017-09-30 at 11 20 08 pm

mdo (commented):
This and the current version are both screwy, so merging away to at least solve one problem :D.

andresgalante:
@mdo I was aware of that problem, I'll try to think of a better solution. Thanks for reviewing it!

PR #23684, opened 2017-08-26
Adds accessibility notes to breadcrumbs

andresgalante:
This PR adds accessibility remarks on breadcrumbs following WAI-ARIA design pattern:

https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

What do you think?

### DIFF
--- a/docs/4.0/components/breadcrumb.md
+++ b/docs/4.0/components/breadcrumb.md
@@ -4,31 +4,36 @@ title: Breadcrumb
 description: Indicate the current page's location within a navigational hierarchy that automatically adds separators via CSS.
 group: components
 ---
+## Overview
 
 Separators are automatically added in CSS through [`::before`](https://developer.mozilla.org/en-US/docs/Web/CSS/::before) and [`content`](https://developer.mozilla.org/en-US/docs/Web/CSS/content).
 
 {% example html %}
-<ol class="breadcrumb">
-  <li class="breadcrumb-item active">Home</li>
-</ol>
-<ol class="breadcrumb">
-  <li class="breadcrumb-item"><a href="#">Home</a></li>
-  <li class="breadcrumb-item active">Library</li>
-</ol>
-<ol class="breadcrumb">
-  <li class="breadcrumb-item"><a href="#">Home</a></li>
-  <li class="breadcrumb-item"><a href="#">Library</a></li>
-  <li class="breadcrumb-item active">Data</li>
-</ol>
-{% endexample %}
 
-Similar to our navigation components, breadcrumbs work fine with or without the usage of list markup.
+<nav aria-label="breadcrumb" role="navigation">
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item active" aria-current="page">Home</li>
+  </ol>
+</nav>
 
-{% example html %}
-<nav class="breadcrumb">
-  <a class="breadcrumb-item" href="#">Home</a>
-  <a class="breadcrumb-item" href="#">Library</a>
-  <a class="breadcrumb-item" href="#">Data</a>
-  <span class="breadcrumb-item active">Bootstrap</span>
+<nav aria-label="breadcrumb" role="navigation">
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item"><a href="#">Home</a></li>
+    <li class="breadcrumb-item active" aria-current="page">Library</li>
+  </ol>
+</nav>
+
+<nav aria-label="breadcrumb" role="navigation">
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item"><a href="#">Home</a></li>
+    <li class="breadcrumb-item"><a href="#">Library</a></li>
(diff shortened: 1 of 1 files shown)


docs/4.0/components/breadcrumb.md
```diff
@@ -32,3 +32,22 @@ Similar to our navigation components, breadcrumbs work fine with or without the
   <span class="breadcrumb-item active">Bootstrap</span>
 </nav>
 {% endexample %}
+
+
+## Regarding accessibility
```
mdo: This would be the only heading on the page; can we change it to `## Accessibility` for brevity? And then also add an `## Overview` or `## Example` above line 8?

docs/4.0/components/breadcrumb.md
```diff
@@ -32,3 +32,22 @@ Similar to our navigation components, breadcrumbs work fine with or without the
...
+
+If you’re using breadcrumbs to provide a navigation, be sure to add a `aria-label="breadcrumb"` to describe the type of navigation provided in the `<nav>` element.
```
mdo: Should we be doing this in our examples?

docs/4.0/components/breadcrumb.md
```diff
@@ -32,3 +32,22 @@ Similar to our navigation components, breadcrumbs work fine with or without the
...
+
+Also apply an `aria-current="page"` to the last `<a>` of the set to indicate that it represents the current page.
+
+More information on [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#breadcrumb).
```
mdo: Feel like the above three lines should be a single paragraph vs three separate lines.

docs/4.0/components/breadcrumb.md
andresgalante: I think that if we can deliver accessible snippets from bootstrap we will be doing a huge dent to help people with disabilities.

So even though the other examples will render correctly, I think we should deliver just one example that follows an accessible pattern.

I'll change this PR and send a new option

andresgalante:
Hi @mdo 
I've reduce the accessibility text, removed the non accessible code snippets and replaced them for the recommended W3C design pattern.

As I wrote on the comment above, I think that the more accessibility we can deliver built into bootstrap components the better the web will be.

What do you think?

docs/4.0/components/breadcrumb.md
```diff
@@ -4,31 +4,36 @@ title: Breadcrumb
...
+    <li class="breadcrumb-item"><a href="#">Home</a></li>
+    <li class="breadcrumb-item"><a href="#">Library</a></li>
+    <li class="breadcrumb-item active" aria-current="page">Data</li>
+  </ol>
 </nav>
 {% endexample %}
+
+## Accessibility
+
+Since breadcrumbs provide a navigation, be sure to add a `aria-label="breadcrumb"` to describe the type of navigation provided in the `<nav>` element. Also apply an `aria-current="page"` to the last item of the set to indicate that it represents the current page.
```
patrickhlauke: I'd soften the language from "be sure to add" to something more like "it's a good idea to add". As it's up to the author to assign a label (i.e. "breadcrumb" isn't a mandated value, just a suggestion), I'd further change it to something like "it's a good idea to add a meaningful label such as `aria-label="Breadcrumb"`.

patrickhlauke (changes requested):
Barring the slight rewording I suggested, this looks good to me 👍

andresgalante:
Hey @patrickhlauke, that makes a lot of sense, thanks for the review.

I've also changed the second phrase. Can you please take another look at it?

docs/4.0/components/breadcrumb.md
```diff
@@ -4,31 +4,38 @@ title: Breadcrumb
...
 </nav>
 {% endexample %}
+
+## Accessibility
+
+Since breadcrumbs provide a navigation, it's a good idea to add a meaningful label such as `aria-label="breadcrumb"` to describe the type of navigation provided in the `<nav>` element, as well as applying an `aria-current="page"` to the last item of the set to indicate that it represents the current page.
+
+
+
+More information on [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#breadcrumb).
```
patrickhlauke: sorry, one final change request (missed it first time around): possibly reword to "For more information, see the [WAI-ARIA Authoring Practices for the breadcrumb pattern](...)." (and maybe remove two of the preceding line breaks, to keep it nice and tight)

andresgalante: @patrickhlauke Nothing to be sorry about, I'll update it now. I am not an english native speaker, so any help on wording is always welcome. I'll update the PR now

andresgalante:
Hi @patrickhlauke the change is done, let me know if I am missing anything else,
thanks a lot for taking the time to review this one.

patrickhlauke (approved):

patrickhlauke:
All good for me, thanks :)

PR #23688, opened 2017-08-26
Fix `npm run docs-github` script.

XhmikosR:
`$npm_config_tmp` is a non-existent variable.
Also, use double quotes since single ones don't work.
Finally, git ignore the temp file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,8 +44,9 @@ scss-lint-report.xml
 # grunt-contrib-sass cache
 .sass-cache
 
-# Jekyll metadata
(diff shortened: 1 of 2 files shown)


Johann-S (approved):

bardiharborow:
This workaround is fine, but you should lodge a bug report with NPM for your platform if `$npm_config_tmp` doesn't exist in a `npm run` context, as that's very much intended behaviour.

XhmikosR:
Nah. no time for that.

PR #23701, opened 2017-08-27
scss/_variables.scss

### LINKED ISSUE #23690: Sass issue using color function
When looking at [Bootstrap Sass Options](https://getbootstrap.com/docs/4.0/getting-started/options/), it gives the example 
``` scss 
// From the Sass map with our `color()` function
.beta { color: color("purple"); }
```
This doesn't work for me.
After debugging and putting lots of @debug statements in my own code and the BS4 _function.scss, I noticed that wrapping the color name in quotes stopped it working for me. It just returns null. 

``` scss 
// From the Sass map with our `color()` function
.beta { color: color(purple); }
```
This version works for me. 
Is any one else having this problem ? 

- Windows 10 
- Sass complied with Gulp-Sass

MartijnCuppens:
Fixes #23690


mdo:
What about the other maps?

MartijnCuppens:
@mdo, it would indeed make more sense if all maps use string-based keys. I just added this in the latest commit.

mdo (approved):

PR #23702, opened 2017-08-27
Fixes link color on navbar-text

### LINKED ISSUE #23693: Text link doesn't get styled when using color schemes for navbar
I am not sure if this actual a bug, or expected behaviour, but I think this might be worth changing.

**Expected behaviour:**
When I use certain color schemes (background) in combination with a navbar, while using text, while using plain links.
I expect the link to be colored in a way it's readable, thus enough contrast. Just like the text is readable.
screen shot 2017-08-26 at 18 29 18

**Actual behaviour:**
When I use certain color schemes (background) in combination with a navbar, while using text, while using plain links. 
The links do not get colored (with enough contrast). Rendering it unreadable while using it in combination with certain background color schemes.

```html
<div class="navbar navbar-expand navbar-dark bg-primary">
  <span class="navbar-text mx-auto">
    <small>We gebruiken cookies om de site te analyseren, te verbeteren en voor tracking doeleinden. Als je de website blijft gebruiken gaan we ervan uit dat je dat goed vindt. <a href="#">Lees Meer</a></small>
  </span>
</div>
```
screen shot 2017-08-26 at 18 15 54screen shot 2017-08-26 at 18 18 15

I understand I can solve it by adding a color scheme to the link (i.e.: `<a class="text-light" href="#">Lees Meer</a>`). But I think it should work without it.

andresgalante:
This PR fixes #23693 

It fixes colors for anchors inside `navbar-text`.

This is how they look like now:

screen shot 2017-08-27 at 7 11 32 am
screen shot 2017-08-27 at 7 11 40 am

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -235,6 +235,13 @@
 
   .navbar-text {
     color: $navbar-light-color;
+    a {
+      color: $navbar-light-active-color;
+
+      @include hover-focus {
+        color: $navbar-light-active-color;
+      }
+    }
   }
 }
 
@@ -280,5 +287,12 @@
 
   .navbar-text {
     color: $navbar-dark-color;
+    a {
+      color: $navbar-dark-active-color;
+
+      @include hover-focus {
+        color: $navbar-dark-active-color;
+      }
+    }
   }
 }


mdo:
I'm slightly concerned about this for folks who might nest other components within `.navbar-text`. Curious if this should be text utilities instead? I dunno; let me think about it for a bit :).

andresgalante:
@mdo Yeap, it makes sense to have it on text utilities since user will have this same problem anywhere that has light text on primary background with a link.

I know use cases can be weird, but what other components would anyone nest under to .navbar-text? a badge? I will check how specificity impacts in those cases.

Think about it and let me know if I should move it. In the meantime I'll run some tests in different scenarios to make sure it's useful.

mdo (approved):

PR #23703, opened 2017-08-27
RTL Support Idea

PascalPixel:
Lets help out the 500 million poor souls whose languages flow from right to left?

This doesn't change any of the classnames (left/right), but does flip around the attributes for them.
So you write `.float-left`, and it floats to the right. Madness. 

Would supplant our dependency on bootstrap-rtl versions and manual solutions.

**Edit**
- Need to rewrite any shorthand css notation with differing sizes for left and right (`margin: 0rem 2rem 0rem 5rem;`)
- Need to add support into the `js`, I haven't used any of these except modals, dropdowns, tooltips and popovers.

**Edit2**
There are multiple approaches for this:
1. Eliminate all left/right logic in Bootstrap with Flexbox (would be best)
2. Write and maintain overrides in a separate optional file (every other RTL solution out there, unmaintainable, file bloat)
3. Add a variable to export an RTL-only version (the solution in this PR)

Since 1. would require a bunch more internal tooling on BS it isn't really viable right now (I think?). 2 is too much work to maintain and has never found its way back into vanilla BS out of all the solutions online. Thus I am advocating solution 3 (this PR) for the time being.

At only 200 line changes, this solution solves the problem for a setup that can take care of compilation for you on the fly (sprockets, webpack) but won't serve basic users who use raw files or CDN.


Johann-S:
I remembered that @mdo said why not add RTL when our v4 will be in a stable release

mabar:
Remember when compass said us to use mixins for autoprefixing? Later autoprefixer shown us there is better way. I think this is the same situation. Why add complexity when automation tools (like gulp-rtlcss) can be used?

Superpencil:
@mabar 
- rtlcss is difficult to install due to limited support, which is why there are dozens of bootstrap-rtl packages
- it's a consideration of course, I don't think it's too much to ask writing `#{$right}` instead of `right`
- Bootstrap manually maintains `aria` labels that support a much smaller group of users, so the precedent is already in the code

andresgalante:
This is a smart solution and even though it adds an overhead to writing CSS I am all for it. 

@Superpencil is right, bootstrap does make an effort to make components accessible but ignores internationalisation.

In my opinion, a PR like this one must include a change in the linter to disallow `left` or `right` without the `#{$   }`, if not things will get messy **really** fast.

Superpencil:
@andresgalante our app’s rtl users have commented many times over the years what a feat it is that we have proper rtl support. And that’s really sad!

I’ll look into how to do the linter, unless you already know how?

andresgalante:
@Superpencil I don't know how to set up the linter :( sorry

I'd wait for @mdo to review this idea before moving forward though

myorangeca:
@andresgalante  I tried gulp-RTL, its not the solution we are all looking for.

Superpencil:
We've deployed this method on our site, available here: www.fishisfast.com/ar 

Our users have responded very positively, there have always been a list of issues that we'd constantly receive through support channels and would go into a file called 'rtf fixes'.

I'm going to list some examples, imagine we get about 10-20 of these reported monthly:
- A modal that went the wrong way
- A button that extends off-screen due to the content being English
- Alert closing tags not being clickable
- Tabs aligning the wrong way and overlapping themselves

The solution has always been to address these issues one-by-one as they were reported, resulting in a long list of manual fixes and hurting our ability to upgrade our vendor assets or implement other dependencies that will override our overrides.

With this PR, which is hosted in a fork every issue has disappeared, it is a breath of fresh air for both our users, our support team and for us devs and allows us to move forward.

Our forks are here:
https://github.com/Superpencil/bootstrap
https://github.com/Superpencil/bootstrap-rubygem

Since we're pretty intense users it's likely we'll update these forks every 6 months or so. I hope this helps others!

mtawil:
Thank you @Superpencil for hard work to make bootstrap support RTL.
I am wondering how your way support, for example, a padding with four values like this line https://github.com/twbs/bootstrap/blob/v4-dev/scss/_custom-forms.scss#L135?

Maybe you should take a look at my PR https://github.com/twbs/bootstrap/pull/10672

Superpencil:
Your right @mtawil it does look silly:
screen shot 2017-10-12 at 04 57 25
I think its best if we rewrite parts like this with `margin-left`, `margin-right` instead of shorthand?

mtawil:
Here is an article that might help you, @Superpencil :
https://github.com/MohammadYounes/rtlcss/wiki/Why-make-a-complete-RTL-version-%3F

johnnyshields:
@Superpencil I disagree with the approach of this PR, because users will need to have separate "bootstrap.css" and "bootstrap-rtl.css" files each which contain the entirely of bootstrap albeit compiled differently. This will require reloading the entire page (or loading a large CSS file) whenever the user changes locale.

It would be much better to use `html[dir="rtl"]` class as a switch which enables RTL-specific overrides like so:

```
.btn-group > .btn-group {
   float: left;
}

html[dir="rtl"] {
  .btn-group > .btn-group {
    float: right;
  }
}
```

Bonus points: it would be even better if we had some way to flip certain sections of the page back to ltr within an rtl block, e.g.:

```
<html> (ltr)
   <div class="direction-rtl">
       <div class="direction-ltr">
```

But this might be beyond the scope of the requirement; it can always be handled by the user manually if needed.

Superpencil:
@johnnyshields I feel like a small change like that would bloat up basic Bootstrap a lot more than loading a different file entirely for when RTL is needed. I don't have the compiled data to prove it however.

👉 I'd argue that if someone has an application that supports two different languages, there would already be a construct in there to deal with two different languages of content, and swapping out a CSS file would be trivial in that context?

johnnyshields:
> there would already be a construct in there to deal with two different languages of content

@Superpencil nope, my app supports 14 languages and I do not have such a construct. It would make things quite messy, increase page load, etc. etc. Moreover it means having to add TWO bootstrap files to your app rather than the current ONE.

All the existing Bootstrap RTL libraries rely on the override (rather than new file) approach.

johnnyshields:
Ultimately it's a question of do we make Bootstrap slightly more messy, or do we force a big amount of mess onto the user implementation (multiple BS files, on-the-fly loaders, etc.) The former is clearly preferrable.

Superpencil:
Quote @mdo https://github.com/twbs/bootstrap/pull/10672#issuecomment-30907288
> I overlooked this completely and hadn't realized it was open until now. I don't think we'll be going this route of having all these styles inline—too much bloat for those who don't want or need it. The optional file makes the most sense to me at this time.

> my app supports 14 languages (including Arabic) and I do not have such a construct.
What do you write your app with?

> All the existing Bootstrap RTL libraries rely on the override (rather than new file) approach.
Exactly, and those methods never found their way back into Bootstrap, thus I am presenting a ~200 line easily maintainable alternative that would solve our own particular use case.

> The former is clearly preferable.
You seem very convinced that we represent an edge case, yet we use Rails with SASS. The existence of an officially maintained `twbs/bootstrap-rubygem` has me believe we are not an edge case?

I agree that a 'drop-in' solution would be the best for end users, but I personally don't see a way to do that and keep a small file size.

johnnyshields:
Superpencil, I am CTO of a company which 3 apps which are fully localized including RTL, one uses Rails+Sass, one uses React+Sass, one uses Ember.js+Sass. All 3 use the approach that I am advocating.

We need to have component-specific overrides for RTL. Any other solution is gonna be a nightmare for frontend devs. Trust me :)

johnnyshields:
Re: file size, I do not think RTL in the way I am advocating will add significant bloat, probably 1kb after gzipping. That being said, RTL overrides could be contained in their own SASS file, OR we could wrap them all in an `@if $include-rtl { ... }` switch so you can omit them entirely from your bootstrap build.

Superpencil:
@johnnyshields can you submit a PR with that solution?

johnnyshields:
@Superpencil I'd be happy to do it, would like to get buy-in from @mdo that if I do it, it will actually be merged (given that there's a bazillion RTL patches floating around on the interwebs)

johnnyshields:
@mdo can you comment on this thread?

Superpencil:
Why not?

Superpencil:
@FezVrasta @mdo

Bootstrap is a drop-in solution for many beginners and a bit more for intermediates, myself included. `postcss-rtl` runs on `node`, and it would be very hard for people who work with simpler systems to set it up and use. In our case we ended up having to migrate from `sprockets` to `webpacker` to add `postcss` to our `rails` app, which was not a trivial task.

As stated before, `postcss-rtl` filters your `css`, but not your `js`, so even though we switched now, our popovers and tooltips are all in the wrong positions (off screen). We don't use other js-based parts, so I can't say what happened there.

Beginners have access to hundreds of `bootstrap-rtl` versions of course, but they are poorly maintained, and I had hoped that now thanks to bootstrap 4's new structure, a ~200+/- PR would be small enough to be considered as a first step.

_20% of the world population that uses rtl languages_
I re-iterate this because `atom` 'accidentally' dropped `rtl` support when they introduced a new `C++` based editor, and our own team also has to constantly be remembered of that as well.

FezVrasta:
Sorry but I think I didn't explain myself well, I was just proposing Bootstrap to use `postcss-rtl` in their build process.

They already use PostCSS for Autoprefixer.

Superpencil:
🤯 I completely misunderstood! I hadn't considered that possibility at all...

PR #23718, opened 2017-08-28
Fix unwanted body padding when a modal is opened

### LINKED ISSUE #23681: [v4b, modal, bug] Right padding added even if there is no scroll bar to hide
Even if there is no scroll bar on body element to be hidden, an extra right padding is added when the dialog is shown. Look at the gap that appears between the green bar and the red rectangle in the reduced test case.

Environment:
- Windows 2016 x64
- Chrome 60.0.3112.101
- Reduced test case: https://jsbin.com/hurotibesi/edit?html,output

valenfys:
Fixes #23681

This prevents padding being added to the body when it's not overflowing but has a margin (e.g. the default 8px) by measuring its outer width (i.e. including the margin).

@Johann-S:

> if it's isn't possible to test that with Phantomjs a visual test would be a solution

The problem is that this is breaking other existing PhantomJS tests... but as stated before, it works fine when running those same tests in Chrome.


techdavid:
Hmm, only one is failing on Travis. Guess I'll try pushing the jQuery outerWidth version and see what happens

techdavid:
So this is doing what I described - 8 tests are failing in PhantomJS but the same tests are all working in Chrome. And now I'll check using the first proposed solution (using `getBoundingClientRect`).

techdavid:
I figured out the problem! - in Phantom there are no vertical scrollbars, and this was breaking the other tests because they need vertical scrollbars to be present. However, these tests were previously working because of the buggy _checkScrollbar which thought that the 8px margin was a scrollbar - in other words, those tests were buggy in the first place but working because of #23681.

To fix the tests, I've given the html element padding-right to simulate a scrollbar, and disabled the scrollbar for the new test, as it tests for the body element when it _isn't_ overflowing.

Johann-S (approved):
Nice fix thanks @techdavid 👍

PR #23725, opened 2017-08-29
Guess text colour also on button hover background

IsraelOrtuno:
It should also guess the button hover colour based on the `$active-background` variable value as it may be a radically different color.


mdo:
Duplicate of #23657.

Issue #23730, opened 2017-08-29
Text Input and Date Input in form when using class .form-row


psingl:
There is a little alignment issue when using form->div.form-row When using text input and date input both together the date input height is more than text input as shown in image attatched.
![image](https://user-images.githubusercontent.com/13885706/29824857-38db5206-8cf0-11e7-83c3-ac25094ce8b7.png)
I am attaching the code also in below image for the form shown above
![image](https://user-images.githubusercontent.com/13885706/29825101-f0ba39f0-8cf0-11e7-983a-8d2f755aaf2b.png)

psingl:
Pull Request created #23731

diemedina:
Adding class :
`.form-row input[type=date] {
    height: 38px;
  }`

mdo:
Duplicate of #18842.

PR #23731, opened 2017-08-29
Issue #23730

itisparas:
Text input and date input together were not inline.


wolfy1339 (changes requested):
Please add only one selector per-line

mdo:
I don't think isn't an ideal solution, sorry. Definitely needs something, but overriding the padding like this isn't it.

Issue #18413, opened 2015-12-03
Export v4 Util class to make it usable in custom JS functions


wasinger:
For custom transition effects I currently use the transition helper functions `$.fn.emulateTransitionEnd` and `$.support.transition` from Bootstrap 3.

While porting my code to Bootstrap 4 I noted that while `$.fn.emulateTransitionEnd` is still available and works as expected, the `$.support.transition` function is no longer availabe because it got replaced by `Util.supportsTransitionEnd()`. Unfortunately, the `Util` class is available within a Closure only so I cannot access it from my code.

So I ask to export the Util class to the global or jQuery namespace (e.g. as `$.BootstrapUtil`) to make it available for use in custom JS code.

This is somehow related to #18136 which complains about the Util class being exported as `Util` in the global namespace, which is not true for the default build - it is not exported at all, but IMHO should be exported under some name other than plain "Util".

Johann-S:
Thanks to #23735 now you can do : `bootstrap.Util` because `bootstrap` is a global var

PR #23735, opened 2017-08-29
Create a bundled release of Bootstrap with Popper.js inside

Johann-S:
I know this PR looks like my other one "Better modularisation of our plugins" but that PR is too huge so I'll split that PR and that's begin with this one.

This PR add a two new dist files : `bootstrap.bundle.js` and `bootstrap.bundle.min.js`, both include Popper.js inside, so with this two files no need to include Popper.js

Also they are new dev dependencies : `rollup` which allow us to bundle our files

- [x] Update `stamp.js`
- [x] Make some tests with a bundler

I chose to build our `bootstrap` file in `iife` because currently that's what we have, we are not totally compatible with UMD

~Work in progress~

/CC @FezVrasta 

Close : #22783, #17201, #18413


package.json
```diff
@@ -65,7 +68,6 @@
     "url": "https://github.com/twbs/bootstrap/issues"
   },
   "license": "MIT",
-  "dependencies": {},
```
Johann-S: Ooups my bad I'll add back that

build/rollup.config.js
```diff
@@ -0,0 +1,36 @@
...
+  plugins.push(resolve())
+}
+
+module.exports = {
+  input: path.resolve(__dirname, '../js/src/index.js'),
+  output: {
+    file: path.resolve(__dirname, `../dist/js/${fileDest}`),
+    format: 'iife'
+  },
+  name: 'bootstrap',
```
Johann-S: FYI with that attribute a global `bootstrap` var will be added to the scope of folk's applications, Rollup do not accept to bundle without a name for `iife`

js/tests/.eslintrc.json
```diff
@@ -28,6 +28,7 @@
     "global-require": "off",
     "no-process-env": "off",
     "no-process-exit": "off",
+    "no-sync": "off",
```
XhmikosR: Isn't this off by default?

build/rollup.config.js
```diff
@@ -0,0 +1,31 @@
...
+var external  = ['jquery', 'popper.js']
+const plugins = []
+const globals = {
+  jquery: '$',
+  'popper.js': 'Popper'
+}
+
+if (BUNDLE) {
+  fileDest = 'bootstrap.bundle.js'
+  // remove last entry in external array ton bundle Popper
```
XhmikosR: A small typo here `ton`

package.json
```diff
@@ -33,9 +33,13 @@
...
     "js-lint-docs": "eslint --config js/tests/.eslintrc.json assets/js/ sw.js",
     "js-compile": "npm-run-all --parallel js-compile-*",
-    "js-compile-bundle": "shx cat js/src/util.js js/src/alert.js js/src/button.js js/src/carousel.js js/src/collapse.js js/src/dropdown.js js/src/modal.js js/src/scrollspy.js js/src/tab.js js/src/tooltip.js js/src/popover.js | shx sed \"s/^(import|export).*//\" | babel --filename js/src/bootstrap.js | node build/stamp.js > dist/js/bootstrap.js",
-    "js-compile-plugins": "babel js/src/ --out-dir js/dist/ --source-maps",
-    "js-minify": "uglifyjs --config-file build/uglifyjs.config.json --output dist/js/bootstrap.min.js dist/js/bootstrap.js",
+    "js-compile-standalone": "rollup --environment BUNDLE:false --config build/rollup.config.js",
+    "js-compile-bundle": "rollup --environment BUNDLE:true --config build/rollup.config.js",
+    "js-compile-plugins": "babel --no-babelrc js/src/ --out-dir js/dist/ --source-maps --presets=es2015 --plugins=transform-es2015-modules-strip",
+    "postjs-compile": "babel dist/js/bootstrap.js -o dist/js/bootstrap.js && babel dist/js/bootstrap.bundle.js -o dist/js/bootstrap.bundle.js && node build/stamp.js",
```
XhmikosR: Is this supposed to be included in another task?

XhmikosR: OK I think I get it now. This task is supposed to be the main JS build task. I think we should keep `js-compile` as the default one with stamp. I find the current tasks a little confusing :/

js/tests/.eslintrc.json
Johann-S: unfortunately no 😄

XhmikosR: Weird, doessn't seem enabled here https://eslint.org/docs/rules/#nodejs-and-commonjs

Johann-S: That's because of this line here : https://github.com/twbs/bootstrap/blob/v4-dev/js/.eslintrc.json#L122

package.json
Johann-S: that's cannot be possible because `js-compile` run all the `js-compile*` tasks so I should wait the end of those tasks to stamp our dist files

Johann-S:
BTW this PR will fix this too : https://github.com/twbs/bootstrap/issues/18413 because now folks will be able to do : `bootstrap.Util` to use our `Util` plugin 😃

XhmikosR (approved):
LGTM as far as I can tell. Note for the future, see if we can drop our custom stamp.js since rollup supports banner

PR #23757, opened 2017-08-30
Removed reference to .text-white

planetoftheweb:
The .text-white class doesn't exist, and is not listed as a color in this list, so this reference is wrong.

### DIFF
--- a/docs/4.0/utilities/colors.md
+++ b/docs/4.0/utilities/colors.md
@@ -11,7 +11,7 @@ toc: true
 <p class="text-{{ color.name }}">.text-{{ color.name }}</p>{% endfor %}
 {% endexample %}
 
-Contextual text classes also work well on anchors with the provided hover and focus states. **Note that the `.text-white` class has no link styling.**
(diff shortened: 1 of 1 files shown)


XhmikosR:
It exists just fine here.

/CC @Johann-S please verify and close this if it's indeed invalid.

andresgalante:
The doc is correct. For example for this PR #23702 we are adding colors to an anchor inside `text-white`

I'd close this PR

planetoftheweb:
If it does officially exist, then it should be listed in the color section of the document. I sort of like the concept of there being a .text-light color instead of a .text-white color. It seems more flexible.

andresgalante:
@planetoftheweb Sorry man! my bad, you are absolutely right, `text-white` doesn't exists!

But, the text should read
**Note that the `.text-light` class has no link styling.**

Don't remove it completely since links inside `text-light` have no styles

Sorry  for the wrong review 😞

mdo:
#23877 was merged, closing.

PR #23762, opened 2017-08-31
Fixes color of popover arrow on bottom popovers

andresgalante:
There is a bug on popovers, the arrow on the bottom pop overs is white:

screen shot 2017-08-30 at 9 37 42 pm

I've changed it to header grey:
screen shot 2017-08-30 at 9 37 36 pm

What do you think?

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -117,7 +117,7 @@
 
     .arrow::after {
       top: -($popover-arrow-outer-width - 1);
-      border-bottom-color: $popover-arrow-color;
+      border-bottom-color: $popover-header-bg;
     }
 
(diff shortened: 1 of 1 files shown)


wojtask9:
this is not right fix.
You assume that `.popover` always have `.popover-header`.

I had the same problem. My fix https://github.com/twbs/bootstrap/pull/23110

You also mentioned that you are working on `rems` in popover. Maybe my PR will help a little bit because right now in BS computations for `.arrow` assumes that border is always 1px width/height

andresgalante:
@wojtask9 thanks for spotting that usecase, You are absolutely right. I'll close this PR.

PR #23776, opened 2017-08-31
Changes popovers from px to rem

### LINKED ISSUE #23763: Remove inner white border effect on popovers
The popovers have an inner white border on the headers, it's a nice effect but I think it makes the styles too opinionated, plus it's a code overhead to generate them and it makes it harder to convert popovers from px to rems.

How about we remove that effect and simplify the popover code?

andresgalante:
Following the same idea of #23468 with tooltips, this PR is changing `px` units to `rem` to meassure popovers.

It also nukes the fancy white border as discussed and closes #23763

This is how it looks with shadows turn on:

screen shot 2017-08-31 at 10 37 30 am

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -5,7 +5,6 @@
   z-index: $zindex-popover;
   display: block;
   max-width: $popover-max-width;
-  padding: $popover-inner-padding;
   // Our parent element can be arbitrary since tooltips are by default inserted as a sibling of their target element.
   // So reset our font and text properties to avoid inheriting weird values.
   @include reset-text();
@@ -39,11 +38,11 @@
 
   .arrow::before {
     content: "";
-    border-width: $popover-arrow-outer-width;
+    border-width: $popover-arrow-width;
   }
   .arrow::after {
     content: "";
-    border-width: $popover-arrow-outer-width;
+    border-width: $popover-arrow-width;
   }
 
   // Popover directions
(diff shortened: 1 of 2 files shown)


scss/_popover.scss
```diff
@@ -61,14 +61,14 @@
...
-      margin-left: -($popover-arrow-outer-width - 5);
+      bottom: -$popover-arrow-width;
+      margin-left: -$popover-arrow-width;
       border-top-color: $popover-arrow-outer-color;
     }
 
     .arrow::after {
-      bottom: -($popover-arrow-outer-width - 1);
-      margin-left: -($popover-arrow-outer-width - 5);
+      bottom: calc((#{$popover-arrow-width} - 1px) * -1);
```
mdo: Change the `1px` here and elsewhere to `$popover-border-width`?

mdo (changes requested):

XhmikosR:
@andresgalante: I rebased and removed the commented out code. Only thing left is to take care of the other @mdo's comment.

scss/_variables.scss
```diff
@@ -626,27 +626,25 @@ $tooltip-arrow-color:               $tooltip-bg !default;
 
 // Popovers
 
-$popover-inner-padding:             1px !default;
 $popover-bg:                        $white !default;
-$popover-max-width:                 276px !default;
+$popover-max-width:                 17.5rem !default;
```
XhmikosR: Shouldn't this stay in `px`?

scss/_variables.scss
```diff
@@ -626,27 +626,25 @@ $tooltip-arrow-color:               $tooltip-bg !default;
...
 
 $popover-body-color:                $body-color !default;
-$popover-body-padding-y:            9px !default;
-$popover-body-padding-x:            14px !default;
+$popover-body-padding-y:            $popover-header-padding-y !default;
+$popover-body-padding-x:            $popover-header-padding-x !default;
 
-$popover-arrow-width:               10px !default;
-$popover-arrow-height:              5px !default;
+$popover-arrow-width:               .8rem !default;
```
XhmikosR: Same for these two?

andresgalante: @XhmikosR I would say for the max-width since we measure grids in `px` but not for the size of the arrow. But I am not sure, What do you think?

andresgalante:
@XhmikosR the change @mdo requested is done, thanks!

Yongjin0919:
Help me

scss/_variables.scss
andresgalante: @XhmikosR actually, if the root font size changes, we want to space of the  popover to be larger.
So, maybe the max-width should in rems too, right?

mdo (approved):

PR #23780, opened 2017-08-31
add iffy form validation state

browner12:
we have an `is-valid` and `is-invalid` validation state, but it would be nice to have something in the middle.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -277,6 +277,7 @@ select.form-control-lg {
 }
 
 @include form-validation-state("valid", $form-feedback-valid-color);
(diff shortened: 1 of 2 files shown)


mdo:
These two states are based on `:valid` and `:invalid` pseudo-classes; there's no `:iffy` :).

See https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation for more background on HTML5 form validation.

browner12:
that's fine and all that only valid and invalid exist in the HTML spec, but in the real world we have need for other validation states.  maybe the naming was off, but I think this should be reconsidered.

as precedent, in v3 we had 3 states:  `has-success`, `has-warning`, and `has-error`.

PR #23781, opened 2017-08-31
Improves badges docs

andresgalante:
Badges docs is using `highlight html` and HTML to demo badges on headers.

To keep docs DRY this PR is changing it to `example html`


mdo:
This was coded this way because of how our table of contents is generated; it'll render all those heading elements if we use their markup. For now we'll leave as separate until we rethink how the right sidebar contents list is generated.

PR #23782, opened 2017-08-31
Converts modal px units to rem

andresgalante:
Following changes here #23776 and #23468

This PR changes `px` units to `rem` on modals.

What do you think?


scss/_variables.scss
```diff
@@ -660,30 +660,30 @@ $badge-pill-border-radius:    10rem !default;
...
 $modal-footer-border-color:   $modal-header-border-color !default;
 $modal-header-border-width:   $modal-content-border-width !default;
 $modal-footer-border-width:   $modal-header-border-width !default;
-$modal-header-padding:        15px !default;
+$modal-header-padding:        1rem !default;
 
-$modal-lg:                    800px !default;
-$modal-md:                    500px !default;
-$modal-sm:                    300px !default;
+$modal-lg:                    50rem !default;
```
mdo: Widths should stay in pixels to match how we do grids.

mdo (changes requested):

scss/_variables.scss
andresgalante: Right! I was about to send them in pixels, but the part of me that wanted to transform breakpoints to rems won 😄  I've just updated the PR.
Thanks for the review @mdo !

mdo (approved):

PR #23783, opened 2017-08-31
Increases clickable area of the close btn on modals

andresgalante:
On alerts we increase the clickable area of the close btns to make them easier to click on devices where the pointer is coarse and improve accessibility.

This PR does 2 things:

1- Introduces the same concept for modals.
1- Changes the `align-items` to `flex-start` so the close btn always stays on the upper right corner when the `modal-title` text wraps.

I am not sure if this should be a global change on _close.scss

What do you think?


mdo:
Possible to see a quick demo of the before/after by chance?

andresgalante:
@mdo sure! here is the bug:
https://codepen.io/andresgalante/pen/NvopBJ?editors=1100

- close btn "X" centers instead of  been top right
- the x btn click are is smaller than a finger.

Here is the solution:
https://codepen.io/andresgalante/pen/MvLpZw?editors=1100

- The close btn in the top right corner
- the click area is larger

What do you you think?

patrickhlauke (approved):

patrickhlauke:
Seems fine to me. @mdo ?

XhmikosR (approved):

XhmikosR:
LGTM too, but I too wonder if we should change `close` instead of this specific case. Close seems to be using float currently.

mdo (approved):

PR #23784, opened 2017-08-31
Dismissable badges

andresgalante:
I have a use case for a dismissible badge. It’s a filter that generates badges.

toolbar-example

I don’t know if it’s common enough, but in case it is this PR introduces dismissible badges by accommodating the size of the `.close` btn when they are inside a `.badge`

What do you think? worth having it in case a user has a use case as twisted as mine? or should this just live in my codebase?

### DIFF
--- a/docs/4.0/components/badge.md
+++ b/docs/4.0/components/badge.md
@@ -76,3 +76,21 @@ Using the `.badge` classes with the `<a>` element quickly provide _actionable_ b
 {% for color in site.data.theme-colors %}
 <a href="#" class="badge badge-{{ color.name }}">{{ color.name | capitalize }}</a>{% endfor %}
 {% endexample %}
+
+
+## Dismisable badges
+
+{% example html %}
+<span class="badge badge-primary">
+  Dismissable badge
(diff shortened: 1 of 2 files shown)


dvmuccillo:
This would be a nice resource to have!

gijsbotje:
i made a demo of this once https://codepen.io/gijsbotje/pen/OjOmby
maybe a bit simpler but i guess with the padding you added the click area is bigger

andresgalante:
@gijsbotje nice demo! click areas a tiny, that's why I've increased the padding.

gijsbotje:
yeah good call on that one, though it might still be small

andresgalante:
@gijsbotje couldn't agree more, it's a tiny clickable are on a tiny component.

If we want to move forward with this idea, it might be a good case for [pointers media queries](https://css-tricks.com/touch-devices-not-judged-size/).

To me this PR is a good start and it solves the issue of

gijsbotje:
wow.. thanks for sharing that article, never heard of it before
would be a great addition to use them

mdo:
New PR at #26823.

PR #23801, opened 2017-09-01
fixes blockquote color whith text-light

andresgalante:
While doing #23800 I realized that `.blockquote-footer` has a hardcoded `grey-600` color. Meaning that if the text is `text-light` the footer will not be light:

screen shot 2017-09-01 at 4 09 29 pm

This PR is making that dynamic by making adding a 60% transparency to `.blockquote-footer` and it looks like this:

screen shot 2017-09-01 at 4 05 40 pm

This is a similar situation as #23702 where text color gets tricky.

What do you think?


scss/_type.scss
```diff
@@ -115,7 +115,7 @@ mark,
 .blockquote-footer {
   display: block;
   font-size: 80%; // back to default font-size
-  color: $blockquote-small-color;
+  filter: opacity($blockquote-footer-opacity);
```
wojtask9: Only technical comment :)
I would use `opacity` instead of `filter:opacity`

andresgalante: Hi @wojtask9 , you are absolutely right. I just like `filter` better but in this case it's a matter of taste. I'll change it to `opacity` and update the PR.

Thanks a lot for the review.

XhmikosR (approved):

XhmikosR:
Maybe 0.6 is too small?

mdo:
I think I'd rather see `rgba()` utilities for white and black colors. This way if someone adds an image in the footer, it's not at an opacity as well.

PR #23806, opened 2017-09-02
Fix input-group-btn height when next to textarea

### LINKED ISSUE #23745: Input Group w/ Textarea
The button does not align with the text-area anymore (in both Firefox and IE11).

This worked perfectly fine in Bootstrap 4 Beta:

```
<div class="input-group">
  <span class="input-group-addon">Id</span>
  <textarea class="form-control" maxlength="1000">Text</textarea>
  <span class="input-group-btn">
    <button class="btn btn-danger" type="button">Delete</button>
  </span>
</div>
```
Bootstrap 4 Beta:
input-group

Bootstrap 4 Alpha:
input-group-2

mattcheah:
As a fix for "Input Group w/ Textarea #23745", I added a `align-items:stretch;` to the .input-group-btn class. 
This stretches the height to be 100% of the textarea (or input) next to it. 

Fixes #23745

### DIFF
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -136,6 +136,7 @@
   // `font-size` in combination with `inline-block` on buttons.
   font-size: 0;
   white-space: nowrap;
+  align-items: stretch;
 
   // Negative margin for spacing, position for bringing hovered/focused/actived
   // element above the siblings.
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -132,11 +132,11 @@
 
 .input-group-btn {
   position: relative;
+  align-items: stretch;
   // Jankily prevent input button groups from wrapping with `white-space` and
   // `font-size` in combination with `inline-block` on buttons.
   font-size: 0;
   white-space: nowrap;
-  align-items: stretch;
 
   // Negative margin for spacing, position for bringing hovered/focused/actived
   // element above the siblings.


XhmikosR:
Please fix the build errors.

andresgalante:
@mattcheah The order of the properties is wrong, `  align-self: stretch;` should go after `position` that's why you are getting build errors.

I've tested this PR doesn't fix the issue. Stretching the `input-group-btn` does not do the trick, it needs more work.

If you want help I can open a new PR with a functional solution. Let me know 😄

mattcheah:
Thanks @andresgalante - I think I had fixed the location in my second commit. 
This is my first time working on bootstrap - would you mind pointing out where I went wrong? It looked to me as if stretching the input-group-btn worked ok in my fiddle (https://jsfiddle.net/fekdjctj/1/)

XhmikosR:
https://github.com/twbs/bootstrap/pull/23810

andresgalante:
@mattcheah you are right! I just tested it, your solution works stand alone but not in the docs, let me run one more test here and see what's up

andresgalante:
@mattcheah I found the problem, there are 2 difference between the branch v4-dev and the released beta:

- `.input-group` is set to top, it was introduced here: https://github.com/twbs/bootstrap/commit/02ae73fc0fc21049ea4199dfda18c87c9b709e52

- and most important `.input-group-btn` is not flex on v4-dev, that's why `align-items: stretch;` on itself makes no effect. I couldn't find the commit that did that change. On this PR #23810 I am bringing back flex and stretching the items.

mattcheah:
@andresgalante i see, that makes sense. Thanks for taking the time to explain! Much appreciated.

PR #23807, opened 2017-09-02
Fix migration button stateful link

vsn4ik:


Johann-S (approved):
Good catch thanks you 👍

PR #23811, opened 2017-09-03
Fix jQuery version

vsn4ik:
https://github.com/npm/node-semver#hyphen-ranges-xyz---abc: `3.0.0 - 4.0.0` := `>=3.0.0 <=4.0.0`
https://github.com/npm/node-semver#caret-ranges-123-025-004: `^3.0.0` := `>=3.0.0 <4.0.0`

For test - https://semver.npmjs.com/:
- pick a package - `jquery`
- enter a range - `2.0.0 - 3.0.0`
- enter a range - `^2.0.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   "license": "MIT",
   "dependencies": {},
   "peerDependencies": {
-    "jquery": "3.0.0 - 4.0.0",
+    "jquery": "^3.0.0",
     "popper.js": "^1.12.3"
   },
   "devDependencies": {
@@ -118,6 +118,7 @@
     "LICENSE"
   ],
   "jspm": {
+    "registry": "npm",
     "main": "js/bootstrap",
     "directories": {
       "lib": "dist"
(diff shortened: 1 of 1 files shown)


Johann-S:
Yes but the range we want is : `3.0.0 - 4.0.0`

XhmikosR:
I think `^3.0.0` is fine because if `4.0.0` it won't be installed anyway based on semver. The only thing is if we want us to be explicit here, nothing more.

About jspm I can't comment since I have never used it.

vsn4ik:
@Johann-S no, `if (version[0] < min || version[0] >= max) {` this is a test for incorrectness (inversion `if (version[0] >= min && version[0] < max`).

Johann-S:
Yes I agree but that's the limit `min` and `max` 😉 but that's ok for me and I never used jspm me too 😟

vsn4ik:
Fix conflict in `bower.json`

bower.json
```diff
@@ -1 +1 @@
-@$% Not a Bower package. Use a better package management tool instead. ^&*
\ No newline at end of file
+@$% Not a Bower package. Use a better package management tool instead. ^&*
```
XhmikosR: Can you skip touching this file altogether?

vsn4ik: Done

XhmikosR (approved):

XhmikosR:
@Johann-S: feel free to merge. From my understanding using `^` has the same effect.

PR #23814, opened 2017-09-03
Suggestion to use consistent information about running docs locally

alundiak:
Idea is about to leave Ruby/Bundler commands in tooling section/readme/space, and on main readme use npm scripts syntax.


Johann-S (approved):

Johann-S:
Thank you 👍

PR #23848, opened 2017-09-06
Add information about our bundle dist file which contain Popper.js inside

Johann-S:
Add information about our bundle dist file which contain Popper.js inside

/CC @XhmikosR and @mdo because I'm not sure of my english here 😄


XhmikosR:
I'm not a native English speaker either as you know :)

That being said, I think the word `inside` is redundant here.

FezVrasta:
> already includes

maybe

Johann-S:
I made some changes 👍

XhmikosR:
@Johann-S: I'd go with this:

```md
Be sure to include [popper.min.js]({{ site.cdn.popper }}) before Bootstrap's JavaScript or use `bootstrap.bundle.min.js` / `bootstrap.bundle.js` which contains Popper.js.
```

Johann-S:
yep that's better thanks @XhmikosR

XhmikosR (approved):

PR #23850, opened 2017-09-06
Update gems and package-lock.json.

XhmikosR:
/CC @Johann-S


Johann-S (approved):

Johann-S:
That's remind me we should upgrade the Popper.js version used by our docs 🤔

PR #23859, opened 2017-09-07
Typo in documentation

chrisfp:


hebbet:
i don't think it is a typo

patrickhlauke:
indeed, doesn't look like a typo (as the sentence talks about all breakpoints "from extra small to extra large")

chrisfp:
I see... I was just wondering how the last sentence of the paragraph ("The remaining breakpoints, however, do include a breakpoint abbreviation.") makes sense. Which are the "remaining breakpoints" when the spacing utilities apply already to all breakpoints? Another change that would make sense is changing the last sentence to "The remaining spacing utilities, however, do include a breakpoint abbreviation.")

PR #23866, opened 2017-09-07
Added type="button" to <button>s without any type

Herst:
https://github.com/twbs/bootlint/wiki/W007

Found as part of my work on bootlint for BS4: twbs/bootlint#410.

I created the following one-liner to do it:
```sh
find docs/ -type f -exec perl -pi -e \
  's/(?<=<button )(.*?)(?=>)/@{[(index($1,"type=")!=-1?"$1":"type=\"button\" $1")]}/g' {} +
```

BTW, I am occasionally rebasing my patch branches so you shouldn't need to merge the current HEAD of `v4-dev` into my patch.

### DIFF
--- a/docs/4.0/components/badge.md
+++ b/docs/4.0/components/badge.md
@@ -31,7 +31,7 @@ Badges scale to match the size of the immediate parent element by using relative
 Badges can be used as part of links or buttons to provide a counter.
 
 {% example html %}
-<button class="btn btn-primary">
+<button type="button" class="btn btn-primary">
   Notifications <span class="badge badge-light">4</span>
 </button>
 {% endexample %}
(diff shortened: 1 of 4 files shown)


patrickhlauke:
at `type`-less button is fine unless in a form. not completely sold on this PR being necessary, but not overly fussed.

XhmikosR:
I think this is a nice change for the reasons mentioned in the bootlint Wiki page.

XhmikosR (approved):

Herst:
I just noticed that my changes are incompatible with the attribute order as propagated by http://codeguide.co/#html-attribute-order but then there are numerous examples in the docs where this is the case. Should I change it?

BTW, I believe there are automated tools which can be used to enforce this.

mdo (approved):

PR #23872, opened 2017-09-07
Use border-radius mixin throughout scss

lglasgow:
As I was going through an update for my company's version of v4 bootstrap I noticed that the `border-radius` mixin wasn’t being used in some places where I thought it should be. I went through and updated code where it seemed that mixin should be used.


XhmikosR:
I'm not 100% sure about this. Sure, it's nice for consistency, but there's no other gain as far as I can tell.

Waiting for @mdo's final word.

scss/_custom-forms.scss
```diff
@@ -94,7 +94,7 @@
 
 .custom-radio {
   .custom-control-indicator {
-    border-radius: $custom-radio-indicator-border-radius;
+    @include border-radius($custom-radio-indicator-border-radius);
```
andresgalante: Please remove it from here, if $enable-rounded is set to false we will lose the styles for the radio

scss/_custom-forms.scss
```diff
@@ -140,9 +140,9 @@
   background-size: $custom-select-bg-size;
   border: $custom-select-border-width solid $custom-select-border-color;
   @if $enable-rounded {
-    border-radius: $custom-select-border-radius;
+    @include border-radius($custom-select-border-radius);
```
andresgalante: There is no need to use the mixin here, or replace the if statement with the mixin

scss/_forms.scss
```diff
@@ -20,10 +20,10 @@
   // Note: This has no effect on <select>s in some browsers, due to the limited stylability of `<select>`s in CSS.
   @if $enable-rounded {
     // Manually use the if/else instead of the mixin to account for iOS override
-    border-radius: $input-border-radius;
+    @include border-radius($input-border-radius);
```
andresgalante: same here, the if statement is present already

andresgalante (changes requested):
@XhmikosR I am also not sure if we need this. The main advantage of doing this is that if `$enable-rounded` is set to `false`, you'll get a compiled CSS that doesn't include these declarations. So there are some places on this PR where using the mixin is wrong or unnecessary.

mdo:
The advantage I think to using the mixin in more places is ideally Sass will remove all those selectors when the `$enable-rounded` is disabled and we're doing `@include border-radius(0)`.

PR #23873, opened 2017-09-07
Use touch-action: none instead of manipulation

jrchamp:
For whatever reason, in Chrome, setting touch-action: manipulation on a large interface causes significant performance issues (i.e. scrolling, resizing). If this is only here to support getting rid of the delay on IE/Edge, touch-action: none [achieves the same goal](https://patrickhlauke.github.io/touch/tests/results/#suppressing-300ms-delay) without wrecking the performance of other browsers.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -285,7 +285,7 @@ label,
 select,
 summary,
 textarea {
-  touch-action: manipulation;
+  touch-action: none;
 }
 
 


Herst:
I believe this would e.g. prevent pinch-zoom on `<textarea>`s, not the end of a world but worth noting.

BTW, is there Chromium bug tracker entry for this issue?

jrchamp:
I was wrong. The issue appears to be caused by having many elements with different touch-action values. This shows up in bootstrap because only the actionable items are having `touch-action` set to `manipulation` while the rest of the document remains `auto`. Using the glob `*` with manipulation or setting it on the `html` element itself avoided triggering the issue. I did not see an obvious tracker entry, so we'll probably do a little more testing before we reach out to Chromium.

patrickhlauke:
The use of `touch-action:manipulation` over `touch-action:none` is intentional. In theory there shouldn't be a perf impact here, unless there's other factors at play. Looping in @rbyers

patrickhlauke:
note that `touch-action:none` would also kill any other gestures (like the ability to scroll altogether) if the user happened to start one of those gestures on a control/button/link etc. so the impact would be quite huge.

Issue #19415, opened 2016-03-05
Simplify the customization of popover


fredgate:
This is a suggestion I have from a long time. I thought it was missing in version 3, and it might be well to allow it in version 4.
When we want to customize the popover (width, content padding...) we actually do not have other choice that to redefine the full template, which is not very readable, maintainable and evolutive.
A better option might be to allow to add a class to the popver, so we could add our custom styles to this class without impacting all the popovers.

Example :

```
.popover-custom .popover-content {
  padding: 0;
}

$('#foo').popover({
  title: ...,
  content: ...,
  class: 'popover-custom'
});
```

GEETAMAMUNI:
Challenge created successfully
Challenge Url: https://www.topcoder-dev.com/challenge-details/30049694/?type=develop&noncache=true

RyanZim:
I'm not sure what you're suggesting. Perhaps you could post a JSFiddle/JSBin of this in action?

staticshock:
I second this.

Take a look at the myriad answers to this question on stack overflow to get a sense for how often it comes up and what kinds of workarounds people are resorting to: http://stackoverflow.com/questions/12170357/dynamically-add-a-class-to-bootstraps-popover-container

If I want to style a specific popover, I have several possible approaches:

1. If I don't specify the popover's `container`, it's created adjacent to the target element. This gives me access to the popover from css via `.target + .popover`, but it also constrains the popover in some possibly breaking ways (e.g. `overflow: hidden` on an ancestor.) These things are annoying to work around in css.
2. If I specify `container: "body"`, it's appended to the `<body>`. This fixes the css problems inherent to the previous approach, but cripples my ability to style that specific popover from css.
3. If I append a custom container to the body, I could use it via `container: "#custom-container"`. This gives me access via `#custom-container .popover`, but has the downside of littering my base template with cruft like `<div id="custom-container"></div>`.
4. If I append it to the body and specify a custom template, I may use `.class-from-custom-template`, but the custom template is a hunk of HTML I'd rather avoid writing if I can. The only interesting part of that template would be the class, and that's unclear when confronted with the entire template.
5. I may resort to attaching the class from javascript, but that makes it hard to initialize popovers generically with a construct such as `$('.has-popover').popover()`.

In the current state of things, I think the best solution is combining the 5th approach with some code that copies something like `data-class` from the target to the popover.

awbergs:
[...]

PR #23874, opened 2017-09-07
Include option to add class to tooltips and popovers

mattcheah:
added feature as a response to https://github.com/twbs/bootstrap/issues/19415 in which people (myself included) wanted a way to add a class to a popover or group of popovers for easier styling. 

I added a default `class` key for `config` and added the class to the $(tip) element if the class value was present.  

So when you create a tooltip (or popover) the class you added in your options will be there, like so: 

```html
<a id="popover1" data-toggle="popover" title="popover 1" data-content="This is my content" data-class="fhqwhgads">Popover 1</a>
<a href="#" id="tooltip1" data-toggle="tooltip" title="Some tooltip text!">Tooltip 1</a>
```

```js
$("#popover1").popover()
$("#tooltip1").tooltip({class:"fhqwhgads"})
```

The class will be added to the element. 
add class to tooltips


js/src/tooltip.js
```diff
@@ -275,6 +276,10 @@ const Tooltip = (() => {
           $(tip).addClass(ClassName.FADE)
         }
 
+        if (this.config.class) {
```
Johann-S:
```javascript
if (this.config.class.length > 0) {
```
Would be better because here, we will always pass into this `if` because `''` is a valid string or you can change the default value to `null`

Johann-S (changes requested):

Johann-S:
if this feature is accepted you'll have to update our documentation 👍

mattcheah:
oop makes sense, thanks. Gimme all that documentation.

Herst:
I was just about to create a RFE for this as I want some of my popovers to have `.d-print-none`.

Johann-S (dismissed):

Johann-S:
I approved you changes but I want reviews from other team members

/CC @XhmikosR , @bardiharborow , @mdo

js/src/tooltip.js
XhmikosR: Isn't your request still valid @Johann-S? If so @mattcheah please take care of it.

Johann-S: @mattcheah made the change by changing the default value to `null` so now that's ok 👍

XhmikosR:
LGTM too, but I wonder, one might want to do this will the other JS plugins. So, I'm not sure if we should go this way...

XhmikosR:
@Johann-s: maybe we could add a helper function and call it in all of our plugins to add an optional class?

JasonRao (approved):
test

Johann-S:
Sorry @XhmikosR for the lack of answer 😟 but yes it could be better to allow that in our components instead of just one

XhmikosR:
@Johann-S: maybe we should just do it without a helper function after all? It just makes sense to have this behavior for all plugins instead of one. So, either in all of them, or none.

mdo:
Closing as stale.

truongnmt:
> @Johann-S: maybe we could add a helper function and call it in all of our plugins to add an optional class?

Is there any leads that I could find that helper function 🤔

PR #23875, opened 2017-09-07
Fix documentations component form example

vsn4ik:


andresgalante:
Hi @vsn4ik 
Why would you want to change that `select` for an `input`?
It makes sense the list the states on a `select`

vsn4ik:
@Hi @andresgalante!
Because by an example we show the flexibility of the grid.
Do not use a heavier element (`<select><option>...</option>...</select>`).

mdo:
We'll leave it as-is—it's helpful to show various elements respond to the grid.

vsn4ik:
@mdo, but the example `<select id="inputState" class="form-control">Choose</select>` is not a correct html element.

patrickhlauke:
@vsn4ik good catch. Fixing it https://github.com/twbs/bootstrap/pull/23912

PR #23876, opened 2017-09-07
Dont use short jQuery methods in documentations

vsn4ik:


docs/4.0/components/list-group.md
```diff
@@ -244,7 +244,7 @@ You can activate a list group navigation without writing any JavaScript by simpl
 Enable tabbable list item via JavaScript (each list item needs to be activated individually):
 
 {% highlight js %}
-$('#myList a').click(function (e) {
+$('#myList a').trigger('click', function (e) {
```
Johann-S: It doesn't seem a correct use of `trigger` see : http://api.jquery.com/trigger/
If you would like to change that line it would be : 
```javascript
$('#myList a').on('click', function (e) {
  e.preventDefault()
  $(this).tab('show')
})

vsn4ik: My bad. Fixed!

Johann-S (approved):

PR #23878, opened 2017-09-08
Use CSS3 pseudo elements in documentation

vsn4ik:


andresgalante:
👍  good point, `::after` and `::before` are psudo elements and not psudo classes

mdo (approved):

Issue #23880, opened 2017-09-08
Feature Request: Add text-underline utility class


PascaleBeier:
Hello there,

An utility class for text-underline would be awesome!

PascaleBeier:
Has PR https://github.com/twbs/bootstrap/pull/23881

mdo:
Closing per #23881.

PR #23881, opened 2017-09-08
Add text-underline Utility class, closing #23880

PascaleBeier:
Adding a Decoration block for utilites and adding .text-underline

Can be used to add the other text-decoration utilties.


andresgalante:
sorry @PascaleBeier but I wouldn't like to see this merged. We need to take really good care about not turning utility classes a 1:1 reflection of every CSS property.

PascaleBeier:
@andresgalante I guess one had to decide where to draw the line right? Interesting to see.

Johann-S:
Agreed with @andresgalante but it's @mdo who will have the final word

mdo:
I don't see us adding this one, but I would consider a `.text-decoration-none` or `.no-underline` utility added for that use case. Far more common than adding an underline via class.

PR #23890, opened 2017-09-08
v4 Collapse.js: Accordions with text-truncate causes the page to jump on close

pat270:
Demo of bug: https://output.jsbin.com/haxutuleju


Johann-S (commented):
Maybe that's possible to add unit test, which check if the scrollbar didn't move ?

Johann-S:
finally I checked and that's not the page which jump it's because we don't call `preventDefault` in this case and your fix do the job, nice catch 👍

Johann-S (approved):

Issue #23840, opened 2017-09-05
Use the 'main' element in documentation and examples


vanillajonathan:
Use the semantic `main` element rather than the ambiguous `div` element in the documentation and examples.

https://getbootstrap.com/docs/4.0/examples/

andresgalante:
I'd add to this issue: Go full accessible for examples, add aria tags, roles and `main` too 😄

vanillajonathan:
Also `nav` needs to be investigated.

asuh:
Is there a reason only `<main>` has an aria role and not `<header>` or `<footer>`? Further, while I support aria roles, are there still screen readers in 2017 that do not know the implicit landmark roles by default? My impression is that, at least for landmark roles like for the elements above, explicit aria roles don't add anymore value.

patrickhlauke:
IE11 doesn't understand/implicitly map `<main>` http://html5accessibility.com/

asuh:
Neat site, good reference. Bookmarked!

PR #23899, opened 2017-09-10
Structure examples with appropriate `<main>`, `<header>`, `<footer>`

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/23840


docs/4.0/examples/sticky-footer/index.html
```diff
@@ -19,13 +19,13 @@
   <body>
 
     <!-- Begin page content -->
-    <div class="container">
+    <main role="container" class="container">
```
XhmikosR: Typo here.

XhmikosR (changes requested):

docs/4.0/examples/sticky-footer/index.html
patrickhlauke: Good spot!

XhmikosR: vnu.jar in action, which is why we need to bring the HTML validator back. :)

patrickhlauke: Fixed

XhmikosR (approved):

docs/4.0/examples/sticky-footer/index.html
XhmikosR: Don't forget to add the PR in the ship notes.

vanillajonathan:
Consider also looking into the `nav` element.

patrickhlauke:
most examples seem to have it, but i wasn't looking for that specifically. good call, will be a follow-up pr

PR #23906, opened 2017-09-10
Adds nav tag around navs examples

### LINKED ISSUE #23831: Accecibility notes on navs
Navs have a section for [accessibility notes](https://getbootstrap.com/docs/4.0/components/navs/#regarding-accessibility).

Since most users will just copy and paste the code snippets and hardly anyone will actually read the docs I wonder if we should bake accessibility into every code snippet.

I know it will be way more verbose and it might not show the point of the example, but the dent Bootstrap can make just be changing this would be huge!

@patrickhlauke What do you think? If you like this idea I can work on a PR

### LINKED ISSUE #23894: Alert color variables missing
Beta-1 seems to be missing the alert color values from Alpha.6; was there any particular reason they were removed? How do we now control the background colors and borders for alerts?

andresgalante:
This PR closes #23831 it adds a `<nav>` tag around Navs snippets.

@patrickhlauke I am in the fence with this one, I thought it was a great idea but now that I think about it, users might end up adding `nav`s everywhere on their document even though their navs are not used as navigation. On the other hands chances are that navs are mostly use for navigation. If you think the accessibility notes are enough, please close this PR, and as always thanks a lot for the review, I learn a lot from them 😄

### DIFF
--- a/docs/4.0/components/navs.md
+++ b/docs/4.0/components/navs.md
@@ -17,20 +17,22 @@ The base `.nav` component does not include any `.active` state. The following ex
 {% endcallout %}
 
 {% example html %}
-<ul class="nav">
-  <li class="nav-item">
-    <a class="nav-link active" href="#">Active</a>
-  </li>
-  <li class="nav-item">
-    <a class="nav-link" href="#">Link</a>
-  </li>
-  <li class="nav-item">
-    <a class="nav-link" href="#">Link</a>
-  </li>
-  <li class="nav-item">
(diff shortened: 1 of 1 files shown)


patrickhlauke (commented):
I'd be ok with this, but I think @mdo is keen to keep example markup down to minimum, so leaving final decision up to him

patrickhlauke:
agree in principle with @mdo's rationale. we do have other cases though where we use `<nav>` in examples, but I guess we can't get around that unless we made every example its own document in an `<iframe>`

PR #23907, opened 2017-09-10
Accessibility fixes to dynamic tabs (`aria-selected`, remove dynamic tabs with dropdowns)

### LINKED ISSUE #23035: JavaScript enabled tab interfaces - is the ability to use dropdowns useful?
Currently the dynamic javascript enabled tab navs http://v4-alpha.getbootstrap.com/components/navs/#javascript-behavior allow for the use of a dropdown.

This particular use is fairly non-standard, and from an accessibility point of view this impedes adopting a "proper" tabbed interface pattern with correct `role="..."` etc https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel as that pattern does not foresee anything other than straight "tab" triggers.

As such, I'd like to get opinion from implementers about whether or not they find the current tab with dropdowns useful, or if dropdowns can be removed (which would then allow a proper ARIA tab implementation).

### LINKED ISSUE #23903: Documentation has misleading aria-expanded attribute
Hi all
currently the documentation says:

```
 <li class="nav-item">
    <a class="nav-link active" id="pills-home-tab" data-toggle="pill" href="#pills-home" role="tab" aria-controls="pills-home" aria-expanded="true">Home</a>
  </li>
  <li class="nav-item">
    <a class="nav-link" id="pills-profile-tab" data-toggle="pill" href="#pills-profile" role="tab" aria-controls="pills-profile" aria-expanded="true">Profile</a>
  </li>
```

It has twice`aria-expanded="true"` while one of them should be set to `false`. Also on the section **_And with vertical pills._** we have `aria-expanded="true"` for all of them. Yet in the documentation it says  _**If the collapsible element is closed by default, the attribute on the control element should have a value of aria-expanded="false".**_

Therefore either I am not getting it right, or there is a mistake in the docs.

patrickhlauke:
* Fixes our current incorrect use of `aria-expanded` rather than the proper `aria-selected` (see https://www.w3.org/TR/wai-aria-practices/#tabpanel)
* Removes dynamic tabbed interfaces with dropdowns, as they're not good for usability and more importantly not a pattern that can be expressed/conveyed to assistive tech

Closes #23035
Further closes #23903

### DIFF
--- a/docs/4.0/components/navs.md
+++ b/docs/4.0/components/navs.md
@@ -304,18 +304,18 @@ Dynamic tabbed interfaces, as described in the [<abbr title="Web Accessibility I
 <div class="bd-example bd-example-tabs">
   <ul class="nav nav-tabs" id="myTab" role="tablist">
     <li class="nav-item">
-      <a class="nav-link active" id="home-tab" data-toggle="tab" href="#home" role="tab" aria-controls="home" aria-expanded="true">Home</a>
+      <a class="nav-link active" id="home-tab" data-toggle="tab" href="#home" role="tab" aria-controls="home" aria-selected="true">Home</a>
     </li>
     <li class="nav-item">
-      <a class="nav-link" id="profile-tab" data-toggle="tab" href="#profile" role="tab" aria-controls="profile">Profile</a>
+      <a class="nav-link" id="profile-tab" data-toggle="tab" href="#profile" role="tab" aria-controls="profile" aria-selected="false">Profile</a>
     </li>
     <li class="nav-item dropdown">
       <a class="nav-link dropdown-toggle" data-toggle="dropdown" href="#" role="button" aria-haspopup="true" aria-expanded="false">
         Dropdown
       </a>
       <div class="dropdown-menu">
-        <a class="dropdown-item" id="dropdown1-tab" href="#dropdown1" role="tab" data-toggle="tab" aria-controls="dropdown1">@fat</a>
-        <a class="dropdown-item" id="dropdown2-tab" href="#dropdown2" role="tab" data-toggle="tab" aria-controls="dropdown2">@mdo</a>
+        <a class="dropdown-item" id="dropdown1-tab" href="#dropdown1" role="tab" data-toggle="tab" aria-controls="dropdown1" aria-selected="false">@fat</a>
+        <a class="dropdown-item" id="dropdown2-tab" href="#dropdown2" role="tab" data-toggle="tab" aria-controls="dropdown2" aria-selected="false">@mdo</a>
       </div>
     </li>
   </ul>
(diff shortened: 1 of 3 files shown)


patrickhlauke:
Further closes https://github.com/twbs/bootstrap/issues/23903

patrickhlauke:
This still leaves open a deviation from the spec with regards to keyboard navigation - per ARIA, the whole `tablist` should just be a single tab stop, and moving between tabs achieved through cursor keys (only in case of dynamic/js tabs, mind). May either note it or tackle it at a later point.

danijelGombac:
@patrickhlauke 

Hello, i have a question about aria values. i check this site for aria:

http://accessibility.athena-ict.com/aria/examples/tabpanel2.shtml

The role="tab" has aria-selected, but role="tabpanel" has aria-hidden.

The bootstrap tab.js add to both role="tab" and role="tabpanel" aria-selected. Is this ok?

patrickhlauke:
`aria-hidden` is not necessary if using `display:none`. and ah, hadn't looked at what the effect on the actual `tabpanel` is (was only focused on the effect on the `tab`s themselves). that shouldn't have any `aria-expanded` or `aria-selected` itself. will add that fix to this PR as well

patrickhlauke:
@danijelGombac pushed a new commit (and rebased) that addresses this. in short: `aria-selected` should only be set on the `role="tab"`, not the tab panel itself. thanks for the question which pointed me to the error 👍

patrickhlauke:
@Johann-S no idea why/how, but rebasing this seems to have pulled in some changes to collapse.js you made...admit at this point my github-fu is stretched, so this will likely need to be properly looked at by @mdo or whoever before being merge-able. sorry...

Johann-S:
your rebase result is a bit strange 😄 to rebase personaly I use this command : `git pull upstream v4-dev --rebase`

patrickhlauke:
ping @mdo ... keen to get this reviewed/merged (though merge may require some git-fu)

bardiharborow:
I'll send you a command to recover git-wise as soon I get home.

bardiharborow:
@patrickhlauke this should recover you (warning that this will wipe any changes on those branch `v4-dev` and `v4-dev-tabs-a11y-fix` other than those I can see on this PR)

you should ensure  are clean before you do this, as it :

```bash
git checkout v4-dev
git pull --ff-only # if this errors, then your v4-dev branch is not clean
git checkout v4-dev-tabs-a11y-fix
git reset v4-dev --hard
git cherry-pick c3acb29 7346eb3 812b069 9411480 62f4c21 3fa84b0
git push -f
```

patrickhlauke:
@bardiharborow thanks, i think that did it (though i now still need to "Update branch" here...is it safe to do, or will it get me in the same mess as before?)

patrickhlauke:
also, mind reviewing this so it can be merged?

bardiharborow:
@patrickhlauke it's safe now, but it says that because your branch is one commit out of sync and I'm surprised that `git pull` didn't pick up that last commit.

bardiharborow (approved):
I'm not familiar enough with the intricacies of ARIA to know if there's anything that can be done about dynamic tabbed interfaces, but I doubt anyone else on the team is either, so happy to defer to you on that one. `aria-expanded` makes more sense according to the spec, so that's fine. As long as you have tested the documentation on both Chrome, and some form of assistive tech (I have not), I don't see any issues with merging this.

patrickhlauke:
> As long as you have tested the documentation on both Chrome, and some form of assistive tech (I have not), I don't see any issues with merging this.

I did test it when I filed it (but can't double-check it now, due to the issue in https://github.com/twbs/bootstrap/issues/24034).

PR #23911, opened 2017-09-11
Changes active state color to make it different from hover

### LINKED ISSUE #23710: Buttons' :hover, :active, and checked background-colors should be different
I know that duplicated issue should not be reported. But, it seems my comment in old issue have been overlooked, so I created this new issue.

I can still confirm #18177 on https://getbootstrap.com/docs/4.0/components/buttons/#checkbox-and-radio-buttons.

### LINKED ISSUE #23897: Toggle button states are visually indistinguishable (from focus state)
Toggled button states are totally visually indistinguishable. It is no way to tell the end user is button toggled or not. Do you plan to change it?

andresgalante:
Since the hover and the active state are the same, then toggles are impossible to spot when the mouse is over the btn.

this PR makes `.active` state a bit darker than hover.

Image of the hover vs active toggle btn

Fixes #23897 
Fixes #23710


XhmikosR (approved):

patrickhlauke:
I'm assuming this only deals with hover, and does not touch on focus/focus+active styling?

andresgalante:
@patrickhlauke this PR makes active/focus a bit darker and leaves hover as it is now.

PR #23912, opened 2017-09-11
Fix malformed `<select>` in forms example

patrickhlauke:
x-ref https://github.com/twbs/bootstrap/pull/23875

### DIFF
--- a/docs/4.0/components/forms.md
+++ b/docs/4.0/components/forms.md
@@ -337,7 +337,10 @@ More complex layouts can also be created with the grid system.
     </div>
     <div class="form-group col-md-4">
       <label for="inputState" class="col-form-label">State</label>
-      <select id="inputState" class="form-control">Choose</select>
+      <select id="inputState" class="form-control">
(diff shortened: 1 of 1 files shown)


docs/4.0/components/forms.md
```diff
@@ -337,7 +337,10 @@ More complex layouts can also be created with the grid system.
     </div>
     <div class="form-group col-md-4">
       <label for="inputState" class="col-form-label">State</label>
-      <select id="inputState" class="form-control">Choose</select>
+      <select id="inputState" class="form-control">
+        <option selected>Choose...</option>
```
vsn4ik: The first option the attribute `selected` is not necessary.

patrickhlauke: sure, but also not harmful (and matches the existing use of `selected` in another example later on in this document)

mdo (approved):
Nice! Was thinking about this while reviewing that other PR and forgot to jump on it myself.

PR #23918, opened 2017-09-11
Moves theme variables outside the theme map

### LINKED ISSUE #23908: Provide variables for theme-colors
Pull request #22836 introduced the two color maps `colors` and  `theme-colors`:
```
$colors: (
  blue: $blue,
  indigo: $indigo,
  purple: $purple,
  pink: $pink,
  red: $red,
  orange: $orange,
  yellow: $yellow,
  green: $green,
  teal: $teal,
  cyan: $cyan,
  white: $white,
  gray: $gray-600,
  gray-dark: $gray-800
) !default;

$theme-colors: (
  primary: $blue,
  secondary: $gray-600,
  success: $green,
  info: $cyan,
  warning: $yellow,
  danger: $red,
  light: $gray-100,
  dark: $gray-800
) !default;
```

While updating a test project to v4-beta I noticed that it isn't possible anymore to override the `theme-colors` properly. To override `theme-colors.primary` one had to change the variable `blue`.

Would it be possible to introduce variables for the `theme-colors` which are then used in the color map? This way one could override single elements of the theme colors without doing strange things like `$blue: red;` or having to redefine the whole `theme-colors` map (while making sure that all other used variables are already defined, e.g. `$green`).

My suggestion is to do something like this:
```
$primary: $blue !default;
$secondary: $gray-600 !default,
$success: $green !default,
$info: $cyan !default,
$warning: $yellow !default,
$danger: $red !default,
$light: $gray-100 !default,
$dark: $gray-800 !default

$theme-colors: (
  primary: $primary,
  secondary: $secondary,
  success: $success,
  info: $info,
  warning: $warning,
  danger: $danger,
  light: $light,
  dark: $dark
) !default;
```

andresgalante:
This PR closes #23908 

@kriim makes a valid point on using and changing theme variables. You cannot overwrite the variables inside the map unless you actually change the source, which is something you wouldn't want to do.

This PR move theme variables outside the map as suggested by @kriim


XhmikosR:
@andresgalante: I rebased this. @mdo: lgty?

mdo (approved):

PR #23927, opened 2017-09-12
Update Gemfile.lock so that it works with platform ruby too. (#23275)

XhmikosR:


XhmikosR:
The failures seem unrelated to this change; something has changed on Travis regarding Java.

PR #23930, opened 2017-09-12
bug fix: fix navbar toggler in responsive mode #23926

### LINKED ISSUE #23926: Navbar responsive state is broken
The navbar is broken on `v4-dev` on a mobile size :

screen shot 2017-09-12 at 1 41 01 pm

it was introduced on this PR: https://github.com/twbs/bootstrap/pull/23652

`flex-grow: 1` makes element to grow, it will not wrap.

I am working on it, finding a solution that also works on IE, I'll send a PR soon

Fouzyyyy:
This PR fixes #23926; it seems to do the trick.

Tested in recent versions of Chrome, Firefox,...and IE10.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -105,6 +105,9 @@
 
 // Button for toggling the navbar when in its collapsed state
 .navbar-toggler {
+  position: absolute;
+  top: 0;
+  right: 0;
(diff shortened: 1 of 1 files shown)


andresgalante:
Sorry @Fouzyyyy but not only doesn't solve the issue but it's also not an elegant solution in my opinion. What I did on my PR works, I just need to adjust it to different breakpoints.

PR #23935, opened 2017-09-13
Forces the search box to stay inside the navbar

### LINKED ISSUE #23932: Dropmenu caret is wrapping in a strange way on the navbar
It only happens on a very specific width here:
http://getbootstrap.com/docs/4.0/examples/navbars/

Plus it pushes the search bar out. Play with it and you'll see it.

screen shot 2017-09-12 at 9 59 13 pm

If nobody wants to take this one, I'll probably work on it this week.

andresgalante:
This doesn't resolve #23932 since the caret still breaks, but it does solve the navbar overflow

It forces the search box to stay inside the navbar container.

@mdo I don't know if this is the most elegant of solutions since we should avoid contextual selectors, but I couldn't think of a better one unless we are willing to remove this line (I still couldn't figure out what it does):
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_forms.scss#L325

This is a codepen to show the fix:
https://codepen.io/andresgalante/pen/MvNNWe

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -32,6 +32,11 @@
     align-items: center;
     justify-content: space-between;
   }
+
+  // Forces the search box to stay inside the navbar container
+  .form-control {
+    width: 100%;
+  }
 }
 
 


mdo:
> I don't know if this is the most elegant of solutions since we should avoid contextual selectors

Agreed, not the right solution here I think. It also doesn't solve any potential issue with additional types of inputs like input groups or selects. Closing out for now given the issue was closed, too. We could revisit what's in that navbar instead maybe? Perhaps we look at `white-space: nowrap`?

Issue #23846, opened 2017-09-06
Arrow out of popover


alecpl:
Popover arrow can be placed incorrectly when button on the edge of the screen.
![arrow-issue](https://user-images.githubusercontent.com/546788/30098664-2ce12888-92e3-11e7-9d05-188237188b94.png)
Tested with most recent Bootstrap4 build.

Johann-S:
Can you make a CodePen of this issue please ? As it mentionned in our [contributing guidelines](https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md)

alecpl:
@Johann-S I thought it's obvious, but here you go: https://jsbin.com/horagubehe. See that horizontal as well as vertical position is buggy.

Johann-S:
it seem's has a CSS problem because Popper.js position correctly our arrow but our arrow width is too high 🤔 

yes @alecpl but that's easier for us to made a fix with a reproducible bug in live 👍

FezVrasta:
@Johann-S: Popper.js already takes care to include the arrow width in its computations (theorically), make sure the most outer arrow element has the right width set.

Johann-S:
Thank you so much @FezVrasta 👍 if you want to do that changes do not hesitate

alecpl:
Maybe #23820 should be considered first.

wojtask9:
Can you check this https://codepen.io/anon/pen/jLjraP ?

wojtask9:
I didn't saw #23820 
This PR should fix the issue

wojtask9:
I was wrong. #23820 isn't working
https://codepen.io/anon/pen/ZJdLpW (css compiled with popover changes from #23820)

Johann-S:
this CodePen : codepen.io/anon/pen/gxNxZw seems fine @wojtask9, isn't it ?

wojtask9:
@Johann-S 
Yes.
This pen contains my fix to Popper.js and css popover changes.
I'll make PR when I'll get back home

jipexu:
> this CodePen : http://codepen.io/anon/pen/gxNxZw seems fine @wojtask9, isn't it ?

the popover bottom left lost his left border ??

wojtask9:
@jipexu 
Something with popper position is wrong. I'm not sure what.

@alecpl 
my PR https://github.com/twbs/bootstrap/pull/23936 that should fix your issues

Johann-S:
[...]

PR #23820, opened 2017-09-03
Popup and tooltip: dimensions  in rem and bug fix of arrow placement

NielsHolt:
Hi
This PR will change the dimensions of popover and tooltip from px to rem AND fix a bug in the way the arrow of popover and tooltips are placed. The current placement of the arrow is only centered on the element when border-width and arrow-width are default values. This PR will allow both border-width and arrow-width to be changed
It need to be merged/integrated with #23763
Could solve #23793 and issues in #23468 
Setting `$border-width: 20px; $popover-border-width:10px;` in Bootstrap 4-beta would give
bootstrap-popover-4-beta
In this PR (with `$popover-arrow-width: 1rem;`) it would be
bootstrap-popover-nielsholt

PR #23936, opened 2017-09-13
fix popover arrow computations and .arrow position

wojtask9:
This fix is based on https://github.com/twbs/bootstrap/pull/23820 but it also fixes 'corner cases' from https://github.com/twbs/bootstrap/issues/23846

According to comment https://github.com/twbs/bootstrap/issues/23846#issuecomment-327465549 "The arrow element should have the same width of the actual space used by the arrow to properly get positioned" and "Note that the correct width is needed to position the arrow properly, and the margins are used to always ensure some spacing between the arrow and the popover's edge".

Also this PR fixes white arrow background with popover-header.
<b>Note</b> that 'old' solution is not valid anymore because `.arrow` position is managed by `popper.js` so it must be relative to `.arrow`

before:

after:

Notes:
1. This fix requires https://github.com/FezVrasta/popper.js/issues/417 if we want `.popover` border larger than 1px. 
2. converting to `rem` is very easy.
3. `calc()` can be removed if border and width use the same units
4. thanks to @NielsHolt and his PR (https://github.com/twbs/bootstrap/pull/23820) it should work also with browser zoom without any issues
5. var $popover-arrow-width currently means "whole" arrow width (before it was only half). The same with $popover-arrow-height.
6. $popover-arrow-margin is used to indicate margin between popover edge. We cannot set this to 0 because by default `.popover` use border-radius.

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -20,139 +20,135 @@
 
   // Arrows
   //
-  // .arrow is outer, .arrow::after is inner
+  // .arrow:before is outer, .arrow::after is inner
 
   .arrow {
     position: absolute;
     display: block;
-    width: $popover-arrow-width;
-    height: $popover-arrow-height;
   }
 
   .arrow::before,
   .arrow::after {
     position: absolute;
-    display: block;
+    content: "";
     border-color: transparent;
     border-style: solid;
   }
 
-  .arrow::before {
-    content: "";
-    border-width: $popover-arrow-outer-width;
-  }
-  .arrow::after {
-    content: "";
-    border-width: $popover-arrow-outer-width;
-  }
-
+  $popover-arrow-width-half: $popover-arrow-width / 2;
+  $popover-arrow-height-half: $popover-arrow-height / 2;
   // Popover directions
 
   &.bs-popover-top {
-    margin-bottom: $popover-arrow-width;
+    margin-bottom: $popover-arrow-height-half;
 
     .arrow {
-      bottom: 0;
+      top: 100%;
+      width: $popover-arrow-width;
+      height: $popover-arrow-outer-height;
+      margin-right: $popover-arrow-margin;
+      margin-left: $popover-arrow-margin;
     }
 
     .arrow::before,
     .arrow::after {
-      border-bottom-width: 0;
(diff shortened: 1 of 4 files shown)


scss/_popover.scss
```diff
@@ -20,139 +20,135 @@
 
   // Arrows
   //
-  // .arrow is outer, .arrow::after is inner
+  // .arrow:before is outer, .arrow::after is inner
```
vsn4ik: `CSS2` -> `CSS3`

NielsHolt:
Hi @wojtask9 and @vsn4ik  
I am not quite sure why you think that #23820 doesn't solve the corner problem
But there are some issues with #23936 when the border-width <> 1px

I have created tree demo to illustrate:
1: [Bootstrap 4-dev with $border-width: 10px](https://codepen.io/NielsHolt/pen/veLMeG): 
2: [This PR with $border-width: 1rem](https://codepen.io/NielsHolt/pen/JrGVvj)
3: [#23820 with $border-width: 1rem](https://codepen.io/NielsHolt/pen/yzBwEm)

Could you please have a look at them and let me hear what you think :-)

wojtask9:
@NielsHolt 
Have a look at this codepen https://codepen.io/anon/pen/wrGEWo (css compiled from https://github.com/twbs/bootstrap/pull/23820) with border 1px.

From `popper.js` docs about `.arrow`: "It will read the outer size of the arrowElement node to detect how many pixels of conjuction are needed"

In your PR you didn't take this into account and `.arrow::before` is bigger than `.arrow`.
It works with border 1rem but not necessary with other sizes.

Also `popper.js` takes arrow margins into account to compute distance between popover edge and arrow.
For example if we don't have border radius on popover then margin can be '0'.
But with rounded popover you cannot set margin to '0' because then you will see that .arrow is disconnected from popover.

In my PR (as you shown) there are issues but I believe these are issues in `popper.js`.

NielsHolt:
Hi @wojtask9 
My PR #23820 was closed by @mdo so I will leave it to you to fix it :-)

mdo:
Passing on this given recent changes to popovers using rems from another PR. I'd be happy to see another PR that addresses scaling the arrow borders with the popover's border width, but I don't see any advantage in the rest of these changes here. Thanks though!

PR #23940, opened 2017-09-13
Eight new popover placement

NielsHolt:
This PR adds eight new values for `placement` for popover and tooltips:
`topleft`, `topright`, `righttop`, `rightbottom`, `bottomleft`, `bottomright`, `lefttop`, `leftbottom`
### Example
eight new popover placement
### Demo
[Please see this demo](https://codepen.io/NielsHolt/pen/GMRwWm)
### Original
The PR is based on [bootstrap-popover-extensions](https://github.com/FCOO/bootstrap-popover-extensions) and adjusted with the appriciated help from @FezVrasta


FezVrasta:
Cool!

I wonder if it would make sense to use the same naming used by Popper.js tho.

js/src/tooltip.js
```diff
@@ -284,19 +311,37 @@ const Tooltip = (() => {
...
+          }
+
+        }
+
         this._popper = new Popper(this.element, tip, {
           placement: attachment,
           modifiers: {
             offset: {
-              offset: this.config.offset
+              offset: newOffset
```
Johann-S: by doing that you remove a custom configuration from our documentation see : https://getbootstrap.com/docs/4.0/components/tooltips/#options

js/src/tooltip.js
```diff
@@ -55,7 +55,31 @@ const Tooltip = (() => {
...
-    LEFT   : 'left'
+    LEFT   : 'left',
+    TOPLEFT    : 'top',
+    TOPRIGHT   : 'top',
+    RIGHTTOP   : 'right',
+    RIGHTBOTTOM: 'right',
+    BOTTOMLEFT : 'bottom',
+    BOTTOMRIGHT: 'bottom',
+    LEFTTOP    : 'left',
+    LEFTBOTTOM : 'left'
```
Johann-S: I don't understand why you didn't use Popper.js placement with `-start` or `-end` see : https://popper.js.org/popper-documentation.html#Popper.placements

NielsHolt: I had the same question from @FezVrasta 
Please see https://github.com/FCOO/bootstrap-popover-extensions/issues/1 for the differents

js/src/tooltip.js
NielsHolt: No - I only sets newOffset if it isn't already set (line 326)

js/src/tooltip.js
tmorehouse: Depending on what locate a person lives (i.e. with a LTR language or RTL language), strt and end could be opposite to what the user thinks.

For users whose written language is LTR (left to right), start is on the left, while those users whose language is RTL (right to left), start is on the right.  In my opinion start and end are ambiguous, depends on your written language direction.

Left and Right are more concrete values (left is always left, and right is always right, but start is not always on the left as it could be on the right, depending on perspective).

FezVrasta: But then you have to handle the additional logic of converting "left" to "top" when a tooltip changes its placement from top to let.

top-start -> left-start

Vs

top-left -> left-top

That's the main reason why Popper.js uses start and end

js/src/tooltip.js
```diff
@@ -284,19 +311,37 @@ const Tooltip = (() => {
...
+              break
+            case -1:
+              newOffset = `-50%p + ${arrowOffset}px`
+              break
+            default:
+              newOffset = 0
+          }
+
+        }
+
```
Johann-S: Can you move that part of this code in a private function please ? (offset calculation)

js/src/tooltip.js
```diff
@@ -70,6 +94,7 @@ const Tooltip = (() => {
     selector            : false,
     placement           : 'top',
     offset              : 0,
+    arrowPadding        : 6,
```
Johann-S: Here you added a new config option, can you update `DefaultType` to handle that new option

Johann-S (changes requested):
You should update our Tooltips/Popovers documentation to inform folks they will have those placements

js/src/tooltip.js
NielsHolt: Yes :-)
A small question - how do I prevent the build css- and js-files in `/dist` and `js/dist` from  being included in the PR?

Johann-S: you can revert those changes thanks to : `git checkout dist/ && git checkout js/dist/` :wink:

FezVrasta: Just don't commit them. You can decide what files you want git to stage

NielsHolt:
Hi @Johann-S 
I have made a new commit (actually two..) with the changes to tooltip.js you required
I will try to update the documentation asap

Johann-S:
yep saw it but can you revert changes into `js/dist/tooltip.js` please ? :smiley:

MiikaL:
Was this addition possibly still going ahead, or support for the native popper.js placements? Would be very interesting to me because of buttons at the edge of screen that need a popover below them which then goes outside the window.

NielsHolt:
Hi @MiikaL 
#23940 is still a open PR, but I don't know if or when it will be included in Bootstrap
But I have made the extentions in a packages https://github.com/FCOO/bootstrap-popover-extensions witch I hope you can use instead or until #23940 is accepted

BnitoBzh:
I just posted a note on a possible improvement of this feature before publishing it.
https://github.com/twbs/bootstrap/issues/26727

richaus:
Did this get merged? Do we have access to all the various placements now?

XhmikosR:
@Johann-S: what should we do with this?

Ruffio:
Is this PR still relevant?

mdo:
Closing as stale.

PR #23945, opened 2017-09-13
Zero Columns

supergibbs:
Add .col-<size>-0 classes with `overflow: hidden` and `padding-left/right: 0`

This was addressed a while back (https://github.com/twbs/bootstrap/issues/11391) but not included. As @tkaprol [commented](https://github.com/twbs/bootstrap/issues/11391#issuecomment-31082835), they would be useful in animation. I am working on a responsive collapsible sidebar and this is the missing piece. The sidebar uses col classes to responsively change it's expanded width and the new col-0 classes to collapse it.

### DIFF
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -42,9 +42,9 @@
 
       .col#{$infix}-0 {
         @include make-col(0, $grid-columns);
-        overflow: hidden;
-        padding-left: 0;
         padding-right: 0;
+        padding-left: 0;
+        overflow: hidden;
       }
 
       @for $i from 1 through $columns {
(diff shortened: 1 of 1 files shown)


andresgalante:
Hi @supergibbs I can't decide if this is a good idea or not. probably because I've never had to deal with animating the grid.

But I do think that if you want to add this to the scss, this PR should also update the grid docs to surface the new `col-0` class

supergibbs:
If approved by mdo, I add some docs too

Herst:
This also means that the [E037](https://github.com/twbs/bootlint/wiki/E037) rule for bootlint won't be needed in the version for v4 any more.

gijsbotje:
if this were to be merged, why not change the for loop to run through the column setup `from 0 through $columns` instead of adding a lone class?

supergibbs:
@gijsbotje because it also adds:
```
padding-right: 0;
padding-left: 0;
overflow: hidden;
```

mdo:
I don't see this being super widely used, so going to pass on it. Thanks for the suggestion though!

PR #23951, opened 2017-09-14
Update offset class in example

Herst:
Some BS3 remains… found using my bootlint adaption for Bootstrap 4: twbs/bootlint#410


XhmikosR:
Not sure if we should just use col-lg-6 offset-lg-3`  since that works again.

/CC @mdo

Herst:
Done.

mdo (approved):

PR #23955, opened 2017-09-15
Rewritten modal without jquery

alekitto:
To make it work correctly the following changes were required:

- Trigger events in jQuery if available
- Add util functions for manipulating data attributes and jQuery data object
- `Util.extend` now supports multiple objects

### DIFF
--- a/js/src/dom/eventHandler.js
+++ b/js/src/dom/eventHandler.js
@@ -134,6 +134,26 @@ function bootstrapDelegationHandler(element, selector, fn) {
   }
 }
 
+function removeHandler(element, events, typeEvent, handler) {
+  const uidEvent = handler.uidEvent
+  const fn = events[typeEvent][uidEvent]
+  element.removeEventListener(typeEvent, fn, fn.delegation)
+  delete events[typeEvent][uidEvent]
+}
+
+function removeNamespacedHandlers(element, events, typeEvent, namespace) {
+  const storeElementEvent = events[typeEvent] || {}
+  for (const handlerKey in storeElementEvent) {
+    if (!Object.prototype.hasOwnProperty.call(storeElementEvent, handlerKey)) {
+      continue
+    }
+
+    if (handlerKey.indexOf(namespace) > -1) {
+      removeHandler(element, events, typeEvent, storeElementEvent[handlerKey].originalHandler)
+    }
+  }
+}
+
 const EventHandler = {
   on(element, originalTypeEvent, handler, delegationFn) {
     if (typeof originalTypeEvent !== 'string'
@@ -187,40 +207,49 @@ const EventHandler = {
 
     const events      = getEvent(element)
     let typeEvent     = originalTypeEvent.replace(stripNameRegex, '')
+
     const inNamespace = typeEvent !== originalTypeEvent
     const custom      = customEvents[typeEvent]
     if (custom) {
       typeEvent = custom
     }
+
     const isNative = nativeEvents.indexOf(typeEvent) > -1
     if (!isNative) {
       typeEvent = originalTypeEvent
     }
 
-    if (typeof handler === 'undefined') {
-      const storeElementEvent = events[typeEvent]
-      for (const keyHandlers in storeElementEvent) {
-        if (!Object.prototype.hasOwnProperty.call(storeElementEvent, keyHandlers)) {
-          continue
-        }
-
-        const handlerKey = keyHandlers.replace(stripUidRegex, '')
-
-        // delete all the namespaced listeners
-        if (inNamespace && originalTypeEvent.indexOf(handlerKey) > -1) {
-          const handlerFn = storeElementEvent[keyHandlers]
-          EventHandler.off(element, typeEvent, handlerFn.originalHandler)
-        }
-      }
-    } else {
+    if (typeof handler !== 'undefined') {
+      // Simplest case: handler is passed, remove that listener ONLY.
       if (!events || !events[typeEvent]) {
         return
       }
 
-      const uidEvent = handler.uidEvent
-      const fn = events[typeEvent][uidEvent]
-      element.removeEventListener(typeEvent, fn, fn.delegation)
-      delete events[typeEvent][uidEvent]
+      removeHandler(element, events, typeEvent, handler)
+      return
+    }
+
+    const isNamespace = originalTypeEvent.charAt(0) === '.'
+    if (isNamespace) {
+      for (const elementEvent in events) {
+        if (!Object.prototype.hasOwnProperty.call(events, elementEvent)) {
+          continue
+        }
+
+        removeNamespacedHandlers(element, events, elementEvent, originalTypeEvent.substr(1))
+      }
+    }
+
+    const storeElementEvent = events[typeEvent] || {}
+    for (const keyHandlers in storeElementEvent) {
+      if (!Object.prototype.hasOwnProperty.call(storeElementEvent, keyHandlers)) {
+        continue
+      }
+
+      const handlerKey = keyHandlers.replace(stripUidRegex, '')
+      if (!inNamespace || originalTypeEvent.indexOf(handlerKey) > -1) {
+        removeHandler(element, events, typeEvent, storeElementEvent[keyHandlers].originalHandler)
+      }
     }
   },
 
@@ -231,6 +260,7 @@ const EventHandler = {
     }
 
     const typeEvent   = event.replace(stripNameRegex, '')
+    const inNamespace = event !== typeEvent
     const isNative    = nativeEvents.indexOf(typeEvent) > -1
 
     const $ = Util.jQuery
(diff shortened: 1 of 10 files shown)


js/src/dom/eventHandler.js
```diff
@@ -197,21 +198,18 @@ const EventHandler = {
...
-        for (const keyHandlers in storeElementEvent) {
-          if (!Object.prototype.hasOwnProperty.call(storeElementEvent, keyHandlers)) {
-            continue
-          }
-          // delete all the namespaced listeners
-          if (inNamespace && keyHandlers.indexOf(originalTypeEvent) > -1) {
-            const handlerFn = events[elementEvent][keyHandlers]
-            EventHandler.off(element, elementEvent, handlerFn.originalHandler)
-          }
+        const handlerKey = keyHandlers.replace(stripUidRegex, '')
```
Johann-S: Why did you add a new regex here ?  `off` method already work

js/src/dom/eventHandler.js
```diff
@@ -234,14 +232,31 @@ const EventHandler = {
...
     const isNative    = nativeEvents.indexOf(typeEvent) > -1
+
+    const $ = Util.jQuery
+    let jQueryEvent
+
+    let bubbles = true
+    let nativeDispatch = true
+    let defaultPrevented = false
+
+    if (!isNative && typeof jQuery !== 'undefined') {
```
Johann-S: here you check for `jQuery !== undefined` it should be better with `$` which come from your `Util.jQuery` getter

js/src/dom/eventHandler.js
```diff
@@ -250,7 +265,21 @@ const EventHandler = {
...
-    element.dispatchEvent(evt)
+
+    if (defaultPrevented) {
+      evt.preventDefault()
+    }
+
+    if (nativeDispatch) {
+      element.dispatchEvent(evt)
+    } else {
+      evt.stopImmediatePropagation()
```
Johann-S: why do you call `stopImmediatePropagation` here ? `evt` isn't dispatched

js/src/util.js
```diff
@@ -119,18 +136,54 @@ const Util = (() => {
...
+    setDataAttribute(element, key, value) {
+      const $ = this.jQuery
+      if (typeof $ !== 'undefined') {
+        $(element).data(key, value)
+      }
+
+      element.setAttribute(`data-${key.replace(/[A-Z]/g, (chr) => `-${chr.toLowerCase()}`)}`, value)
+    },
+
+    removeDataAttribute(element, key) {
```
Johann-S: `setDataAttribute` and `removeDataAttribute` should be in our DOM Manipulator here : https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/dom/manipulator.js

js/src/modal.js
```diff
@@ -537,53 +538,57 @@ const Modal = (() => {
...
-        if ($(this).is(':visible')) {
+      EventHandler.one(target, Event.HIDDEN, () => {
+        if (Util.isVisible(this)) {
           this.focus()
         }
       })
     })
 
+    const $ = window.$ || window.jQuery
     Modal._jQueryInterface.call($(target), config, this)
```
Johann-S: Without a change here, our Modal continue to depend on jQuery

Johann-S (changes requested):

Johann-S:
You made a very good work here @alekitto but they are few changes here before I can merge your PR 👍

js/src/util.js
```diff
@@ -64,7 +75,7 @@ const Util = (() => {
     },
 
     supportsTransitionEnd() {
-      return Boolean(transition)
+      return Boolean(EventHandler.getBrowserTransitionEnd())
```
Johann-S: Not a big fan here because `EventHandler.getBrowserTransitionEnd()` will be called each time `Util.supportsTransitionEnd()` is called

js/src/dom/eventHandler.js
Herst: `const` does not work in IE10

XhmikosR: We use babel.

Johann-S:
I updated the base branch with `v4-dev`, so that's why they are other commits

alekitto:
Ok, i've rebased my branch with upstream. Now working on your comments..

js/src/dom/eventHandler.js
alekitto: You're right. Fixed.

js/src/modal.js
alekitto: I've missed it. I've just changed it.

js/src/dom/eventHandler.js
alekitto: The `off` implementation was not complete and i think i've messed up some things. Now working on a cleaner and more jQuery-like solution.

js/src/util.js
```diff
@@ -64,7 +77,11 @@ const Util = (() => {
     },
 
     supportsTransitionEnd() {
-      return Boolean(transition)
+      if (transition === null) {
```
Johann-S: Here you add a check if transition is set or not, you should revert this change it was better before that, because they aren't any check done each time we call this method

Johann-S (changes requested):

js/tests/unit/modal.js
```diff
@@ -626,16 +635,19 @@ $(function () {
...
+        assert.notOk(event.defaultPrevented, 'navigating to href will happen')
 
         setTimeout(function () {
-          assert.ok(event.isDefaultPrevented(), 'model shown instead of navigating to href')
+          assert.ok(evt.defaultPrevented, 'model shown instead of navigating to href')
           done()
         }, 1)
       })
-      .trigger('click')
+    $('#test')[0].dispatchEvent(evt)
```
Johann-S: if you want you can use `EventHandler` here

js/src/util.js
alekitto: Cyclic import (util -> eventHandler -> util) was causing me troubles yesterday, but now seems working correctly. Probably it was another error.

Johann-S:
Oups last thing I forgot, can you update our visual modal test ? please

Johann-S (approved):

alekitto:
Ok, EventHandler off method has been rewritten to support namespaced events (`click.namespace`), native events without namespace (`click` removes `click.namespace`) and namespace only (`.namespace` removes `click.namespace` and `keydown.namespace`).
Visual modal test has been updated

Johann-S:
I don't understand why you made that change on `off` because `off` already support namespaced events 🤔 
What's the difference between what you've done and what we had ? and why you did that change ?

js/tests/unit/dom/eventHandler.js
```diff
@@ -0,0 +1,233 @@
+$(function () {
+  'use strict'
+
+  QUnit.module('event handler')
+
+  QUnit.test('should be defined', function (assert) {
+    assert.expect(1)
+    assert.ok(EventHandler, 'button method is defined')
```
Johann-S: that's not our `button` plugin 😄

Johann-S (changes requested):

alekitto:
The `off` method did not remove events if event type like `click.dismiss.bs.modal` was passed.  This would lead to events not removed in the `hide` method, especially from `document` and `window` nodes. Additionally, non namespaced events was removed only if an handler function was specified as third argument

Johann-S:
Ok thank you for the explanation 👍 and yep the second point wasn't implemented thank you

alekitto:
Yes, now `EventHandler.off(element, 'click.bs.alert')` is working as expected

js/tests/unit/dom/eventHandler.js
alekitto: Whoops

Johann-S (approved):

arcanedev-maroc:
About Modal events, can i suggest a feature request ?

How about adding a public method to listen a modal event:

```js
class Modal {
    //...

    on(event, callback) {
        this._element.addEventListener(`${event}${EVENT_KEY}`, callback);    
    }

    //...
}
```

So instead of:

```js
var myModalEl = document.getElementById('myModal')
myModalEl.addEventListener('hidden.bs.modal', function (e) {
    // do something...
})

// OR

let myModal = new bootstrap.Modal(document.getElementById('myModal'));
myModal._element.addEventListener('hidden.bs.modal', function (e) {
    // do something...
})
```

We can do something like this:

```js
var myModal = new bootstrap.Modal(document.getElementById('myModal'));
myModal.on('hidden', function (e) {
    // do something...
});
```

I don't know if it's going to have a conflict with jquery but we can change the method name `on` into something else like `listen('hidden', ...)`.

I can create a PR for this feature if it's OK :+1:

alekitto:
My 2 cents: This could be an interesting feature, but I think that should be implemented for all the components.
For example:

```js
const myCollapse = new bootstrap.Collapse(document.getElementById('test-collapse'));
myCollapse.on('shown', e => { ... });
```

could be equally good and useful.

@arcanedev-maroc probably you should open a new issue proposing a these changes to the event system of the components to better discuss pros and cons of this new design.

arcanedev-maroc:
@alekitto DONE :+1:

Issue #23914, opened 2017-09-11
Look into using Travis build stages


XhmikosR:
https://docs.travis-ci.com/user/build-stages/

And we should definitely try and fix #23387 too...

/CC @Johann-S

bardiharborow:
I literally came here to GitHub to open this exact issue, so thanks for reading my mind. :)

PR #23961, opened 2017-09-16
Move Travis to build stages and fix skip browser flag.

### LINKED ISSUE #23387: `skip browser` doesn't have any effect
/CC @bardiharborow

bardiharborow:
Fixes #23387. Begins to implement #23914, but I imagine there may be way more mileage we can get out of build stages. @XhmikosR maybe you have some advice?


XhmikosR:
Thanks for the patch! I'll have a look later and let you know.

XhmikosR (approved):
@bardiharborow: I think the current patch looks good so far. As long as `skip browser` works I say we merge it.

PR #22727, opened 2017-05-31
Use correct version of babel-preset-es2015

existentialism:
The 7.x-alpha preset won't work with 6.x core.

An alternative would be to bump babel-cli to 7.x-alpha (if you'd like to help us test :D) though it's probably best to wait until rc/release.

PR #23966, opened 2017-09-16
Babel 7

hzoo:
👋 Bootstrap team! Making a PR to change to v7 if desired! (instead of by 
 accident via https://github.com/twbs/bootstrap/pull/22727, hopefully we can fix that kind of dependency issue soon)

- [x] Update packages to Babel 7
- [x] Use https://github.com/babel/babel-preset-env instead of preset-es2015 (just deprecated it), has the same behavior if you don't specify a target
- feel free to edit the pr

## Diff

- New `loose` option for classes removes a bunch of helper checks, so they won't be in the bundle/files anymore, so a smaller output. Via https://github.com/babel/babel/pull/4850

- rm`'classCallCheck', 'inherits', 'possibleConstructorReturn'`
- add `inheritsLoose`

screen shot 2017-09-16 at 3 14 28 pm

- Aliases the class prototype via https://github.com/babel/babel/pull/5560

screen shot 2017-09-16 at 3 21 30 pm

- babel-generator defaults to `"` double quotes now via https://github.com/babel/babel/pull/5824
screen shot 2017-09-16 at 2 54 50 pm

- there are some newline changes because Babel won't read `tokens` from the AST anymore, so it may look different but it should compile faster (faster parse/print) via https://github.com/babel/babel/pull/5833
- Comments from dead code elim for a minifier if using exports. via https://github.com/babel/babel/pull/6209

screen shot 2017-09-16 at 2 55 21 pm

- default for template literals is using concat instead of `+`, can use `loose` to go back to `+` via https://github.com/babel/babel/pull/6098
screen shot 2017-09-16 at 2 55 14 pm


package.json
```diff
@@ -33,9 +33,9 @@
     "js-lint": "eslint js/ && eslint --config js/tests/.eslintrc.json --env node build/ Gruntfile.js",
     "js-lint-docs": "eslint --config js/tests/.eslintrc.json assets/js/ sw.js",
     "js-compile": "npm-run-all --parallel js-compile-*",
-    "js-compile-standalone": "rollup --environment BUNDLE:false --config build/rollup.config.js",
-    "js-compile-bundle": "rollup --environment BUNDLE:true --config build/rollup.config.js",
-    "js-compile-plugins": "babel --no-babelrc js/src/ --out-dir js/dist/ --source-maps --presets=es2015 --plugins=transform-es2015-modules-strip",
+    "js-compile-standalone": "ROLLUP=true rollup --environment BUNDLE:false --config build/rollup.config.js",
+    "js-compile-bundle": "ROLLUP=true rollup --environment BUNDLE:true --config build/rollup.config.js",
+    "js-compile-plugins": "PLUGINS=true babel js/src/ --out-dir js/dist/ --source-maps",
```
hzoo: maybe don't want to do it this way (or use https://www.npmjs.com/package/cross-env), don't think we support the `--environment flag` but we could think about it

XhmikosR:
Thanks for the PR @hzoo! One thing, you shouldn't include the dist files in your PR.

/CC @Johann-S for review

XhmikosR:
I wonder if we should tweak the options to use for example single quotes. And maybe switch to `loose` for concatenating strings.

hzoo:
> One thing, you shouldn't include the dist files in your PR.

Yeah I kept them in to explain the changes, can be removed later.

> And maybe switch to loose for concatenating strings.

I made the individual files use the same config (loose) so that is already done.

> should tweak the options to use for example single quotes. 

This we just defaulted to quotes and removed the options since it's just the output and not source code.

XhmikosR:
I'm aware it's the output. Too bad there is no option since I personally am
used to single quotes everywhere.

Anyway not a big deal I guess.

hzoo:
Yeah I hope not 😛, I just picked one - https://github.com/babel/babel/issues/5139

Johann-S:
Thanks for your PR @hzoo, I thought about doing that change but you were quicker than me 😄 

Just remove changes in our dist files and we will merge your PR 👍

Johann-S (approved):

hzoo:
Ok removed the dist, js/dist changes

Johann-S:
Sorry but they are warning in Travis see : https://travis-ci.org/twbs/bootstrap/jobs/276359245#L618
about : `createClass is not exported by babelHelpers` those warnings wasn't there before that

package.json
Johann-S: I don't understand that change, `--no-babelrc` is it deprecated ?

hzoo:
Hm good catch, didn't think that was needed so I removed in rollup.config.js

hzoo:
Ok I see a change we can make to loose mode for static getters https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAECCIFMBOAXaBvAUNaEVhQEthoBzBNANQFEAlAZQEkB5AOQAoBKDbHaJCgFckAO2g0GLVrwC-mGUA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015-loose&prettier=false&targets=&version=7.0.0-beta.0 which we can fix later.

I also excluded typeof transform so that helper can be dropped. I assume bootstrap doesn't use `Symbol` so it's not needed.

package.json
hzoo: no it's because there isn't a way to pass options to a preset via cli

Johann-S:
Thanks @hzoo 👍 
BTW I think you fix this issue : https://github.com/rollup/rollup-plugin-babel/issues/165 😄

PR #23972, opened 2017-09-17
Add eslint-plugin-compat

XhmikosR:
Note that we can move the browserlists config to a separate file; I went with package.json so that we don't have one more file.

### DIFF
--- a/build/postcss.config.js
+++ b/build/postcss.config.js
@@ -5,27 +5,6 @@ module.exports = (ctx) => ({
     sourcesContent: true
   },
   plugins: {
-    autoprefixer: {
-      browsers: [
-        //
-        // Official browser support policy:
-        // https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/#supported-browsers
-        //
-        'Chrome >= 45', // Exact version number here is kinda arbitrary
-        'Firefox ESR',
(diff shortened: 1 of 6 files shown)


build/postcss.config.js
```diff
@@ -5,27 +5,6 @@ module.exports = (ctx) => ({
...
-        'Explorer >= 10',
-        // Out of leniency, we prefix these 1 version further back than the official policy.
-        'iOS >= 9',
-        'Safari >= 9',
-        // The following remain NOT officially supported, but we're lenient and include their prefixes to avoid severely breaking in them.
-        'Android >= 4.4',
-        'Opera >= 30'
-      ]
-    }
+    autoprefixer: {}
```
Johann-S: why did you let that line ? their are nothing in `autoprefixer`

XhmikosR: How will autoprefixer be included otherwise?

XhmikosR:
I ignored the `serviceWorker` errors in pwa.js, but I wonder, should it even complain since we have that block wrapped in a `if ('serviceWorker' in navigator) {` check?

XhmikosR:
@Johann-S: what do we do with this?

Johann-S:
IMO we should merge this PR because currently folks PR aren't tested on Saucelabs so it will help folks to know if their code is compatible with our supported browsers

Johann-S (approved):

PR #23981, opened 2017-09-17
V4 without jquery: fix tests

alekitto:
Related to https://github.com/twbs/bootstrap/pull/23586#issuecomment-330031101
There were a couple of buggy tests for IE 11:
 - trying to fire a `click` on an element not added to the DOM does not produce any effect
 - `defaultPrevented` does not work for code-generated events

### DIFF
--- a/js/tests/unit/modal.js
+++ b/js/tests/unit/modal.js
@@ -627,7 +627,6 @@ $(function () {
 
   QUnit.test('should not follow link in area tag', function (assert) {
     assert.expect(2)
-    var done = assert.async()
 
     $('<map><area id="test" shape="default" data-toggle="modal" data-target="#modal-test" href="demo.html"/></map>')
       .appendTo('#qunit-fixture')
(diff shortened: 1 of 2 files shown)


js/tests/unit/modal.js
```diff
@@ -627,27 +627,26 @@ $(function () {
...
     $('<div id="modal-test"><div class="contents"><div id="close" data-dismiss="modal"/></div></div>')
       .appendTo('#qunit-fixture')
 
-    var evt = document.createEvent('HTMLEvents')
-    evt.initEvent('click', true, true)
+    // We need to use CustomEvent here to have a working preventDefault in IE tests.
+    var evt = new CustomEvent('click', {
+      bubbles: true,
+      cancelable: true
+    })
```
Johann-S: that's why previously I asked you to use `EventHandler` in our unit test :wink:

alekitto: I've tried to use `EventHandler.trigger` instead of creating a `CustomEvent` object, but `preventDefault` produces the same buggy result. Trying to replace the `preventDefault` in an `HTMLEvents` object works on IE, but breaks in phantom. This is the only working solution for all the browsers.

Johann-S: that's weird I didn't see that issue in our other plugins without jQuery

alekitto: I think this is the only test for the `defaultPrevented` property for a non-jQuery event by now. In other tests `trigger` is used to emit the event (which works correctly), but the Event object is discarded and never checked.

alekitto:
@Johann-S Now Edge is broken.. workin' on it.

Issue #23116, opened 2017-07-18
Move from grunt-saucelabs to Travis's built-in Sauce Connect


bardiharborow:
grunt-saucelabs is the last thing stopping us dropping grunt as a dependency altogether. I'd like to get this done before we ship the second beta. This will also allow us to start running Sauce tests against PRs.

https://docs.travis-ci.com/user/sauce-connect/

XhmikosR:
Agreed. I was looking at it too the other day. If you make any progress, feel free to CC me to have a quick look.

PR #23985, opened 2017-09-18
Revert "Move Travis to build stages and fix skip browser flag. (#23961)"

XhmikosR:
This reverts commit e296c1374104ef949ca8f6307f13b09dd4c4b09c.

@bardiharborow: @Johann-S noticed that the browser tests don't run in the PRs now. Since I don't know when you might be able to have a look, I make the PR to merge so that we have tests and we can look into it again in the future. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,24 +8,27 @@ node_js:
   - "8"
 before_install:
   - if [[ `npm -v` != 5* ]]; then npm install -g npm@5; fi
+  - "export TRAVIS_COMMIT_MSG=\"`git log --format=%B --no-merges -n 1`\""
+  - echo "$TRAVIS_COMMIT_MSG" | grep '\[skip browser\]'; export TWBS_DO_BROWSER=$?; true
 install:
   - bundle install --deployment --jobs=3 --retry=3
   - npm install
+script:
+  - npm test
+  - if [ "$TWBS_TEST" = browser -a "$SAUCE_ACCESS_KEY" ]; then npm run js-test-cloud; fi
 after_success:
   - if [ "$TRAVIS_REPO_SLUG" = twbs-savage/bootstrap ]; then npm run docs-upload-preview; fi
-stages:
-  - test
-  - name: browser
-    if: type = push
-jobs:
-  include:
-    - stage: browser
-      node_js: 8
-      script: if ! git log --format=%B --no-merges -n 1 | grep '\[skip browser\]'; then npm test && npm run js-test-cloud; fi
 cache:
   directories:
     - node_modules
     - vendor/bundle
+env:
+  - TWBS_TEST=core
+  - TWBS_TEST=browser
+matrix:
+  exclude:
+    - node_js: "4"
+      env: TWBS_TEST=browser
 notifications:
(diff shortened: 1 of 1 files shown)


bardiharborow (changes requested):
This is intended behaviour. PRs do not have access to the Sauce credentials, and never ran the browser tests anyway (note they did for team members, but team members are using branches and trigger a separate `push` build anyway). Savage used to run security checks on the PR and then ran the browser tests separately, but this has been broken for some time. e296c13 simply stops needlessly booting the VM, only to bail out because we don't have the correct environment variables. If we fix #23116 using the builtin Travis bridge (which I swear is still on my list), then we may be able to run browser checks securely for PRs and we can revisit this.

Johann-S:
My bad, folks PRs are not checked on Saucelabs but that's a huge issue for us because we try to build a Bootstrap release without jQuery.

I think we can close this PR, but maybe we should open an issue to keep in mind we should be able to run Saucelabs on folks PR

bardiharborow:
Closing out pending response from @XhmikosR. Ping me if anything is wrong here.

XhmikosR:
I'm still not bought on the advantages of stages at this point. Not having Saucelabs tests on PRs is a huge deal breaker for me.

And I bet the same is for @Johann-S.

At this point my suggestion would be to revert the stages change and just fix `skip-browser`.

XhmikosR:
So, I'm reopening this until we at least document this issue. Ideally we should have Saucelabs run on PRs too because things can break and we'll notice later, not before merging.

Johann-S:
Stages don't help us and I don't see any advantages to use that

bardiharborow:
> I'm still not bought on the advantages of stages at this point.

The advantages are minor:

- `.travis.yml` is slightly clearer (we're now defining Node.js versions in terms of inclusions, rather than exclusions, and there's better separation between different builds)
- Sauce isn't hammered for builds that fail local testing (Sauce reliability has always been an issue, and I'm not very familiar with our API limits, but I imagine less calls is better)
- The Travis npm package cache is reused for the `browser` VM rather that it creating it's own separate cache.
- Travis provides pretty graphical separation.

> Not having Saucelabs tests on PRs is a huge deal breaker for me.

@XhmikosR I think you've misunderstood. Travis PR builds **have never run Sauce testing**. Please read [the Travis documentation](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions), or my explanation above for why. e296c13 did _not_ change that. e296c13 simply used `if: type = push` to stop running an extraneous VM that didn't do anything (which was creating the illusion that we were running Sauce testing against PRs when we actually weren't). If you do need to revert something, just revert that line.

> So, I'm reopening this until we at least document this issue.

I agree that this is an issue, and it was tracked by #23116, but I've edited that issue to make it more explicit. Reverting this will not fix anything though. If you could give me some assistance fixing #23116, that would be great.

Feel free to hit me up on Slack if my reasoning is unclear.

XhmikosR:
OK, I think I completely misunderstood then. My bad, sorry!

So I guess we stick with the stages, right? Would be nice if stages were running in parallel but maybe it's because it's still a beta feature.

PR #23986, opened 2017-09-18
Add note about schema.org

mik-laj:


andresgalante:
I like this! schema is something you can add to almost any component, I wonder if it should be a gloacl comment instead of just related to breadcrumbs.

mik-laj:
What other components would you like to describe? I think we should use them as WAI-ARIA. ARIA improve accessibility for people, schema.org for search engines. ARIA speaks for each component separately.

andresgalante:
I agree, if you see accessibility notes missing please add them.

Most components on bootstrap docs have roles and aria tags: In the case of breadcrumbs it was merged here:
https://github.com/twbs/bootstrap/pull/23684

mik-laj:
I do not see the missing tags. I just used it as a comparison. WAI-ARIA attrubute  for humans are schema.org for search engines.

ad-m:
>  WAI-ARIA attrubute for humans are schema.org for search engines.
@mik-laj , do you mean "WAI-ARIA attrubute for humans as schema.org for search engines."?

mik-laj:
Yes😐. You're right.

patrickhlauke (commented):
Personally, I think Bootstrap documentation is not the place for suggesting how to make your site better for search engines (compared to the ARIA stuff, which is user-centric and without it AT users will simply not be able to understand/operate your site).

Further, why only add it to breadcrumbs? If we do want a concerted effort to give advice on how to make your site SE friendly, then it needs to go everywhere, not just here.

PR #23989, opened 2017-09-18
V4 without jquery: workaround for ms edge defaultPrevented reset

alekitto:
Related to https://github.com/twbs/bootstrap/pull/23586#issuecomment-330319137 and https://saucelabs.com/jobs/03afada422c7448d84ea067dee18906d failed job

Workaround for M$ Edge resetting defaultPrevented flag on dispatchEvent

@Johann-S I will really appreciate your opinion about this

### DIFF
--- a/js/src/dom/eventHandler.js
+++ b/js/src/dom/eventHandler.js
@@ -23,6 +23,8 @@ const workingDefaultPrevented = (() => {
   return e.defaultPrevented
 })()
 
+let defaultPreventedPreservedOnDispatch = true
+
 // CustomEvent polyfill for IE (see: https://mzl.la/2v76Zvn)
 if (typeof window.CustomEvent !== 'function') {
   window.CustomEvent = (event, params) => {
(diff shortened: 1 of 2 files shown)


js/src/dom/eventHandler.js
```diff
@@ -23,6 +23,19 @@ const workingDefaultPrevented = (() => {
   return e.defaultPrevented
 })()
 
+const defaultPreventedPreservedOnDispatch = typeof window.CustomEvent !== 'function' || (() => {
```
Johann-S: to avoid double check of `typeof window.CustomEvent` you should move this IIFE where we add a polyfill to add `CustomEvent`

Johann-S: and can you add a comment to explain the reason we added that please

Johann-S:
Thank you to handled that @alekitto :heart:

js/src/dom/eventHandler.js
alekitto: Ok, done

Issue #23329, opened 2017-08-11
V4.beta Dropdown menu focus/hover style constrast too low for accessibility


tmorehouse:
The current `:focus` / `:hover` styling is almost impossible to see (on some screens) as it is a very subtle change.

I would recommend that the  background be darkened slightly to make it better for keyboard-only users to know which `.dropdown-item` they have highlighted.

In this screen shot, I have `Another Item` focused, and at quick glance (without straining my eyes) I can't tell which menu item is focused:
![image](https://user-images.githubusercontent.com/2781561/29220721-6246c54c-7e92-11e7-93a1-19ab486119c9.png)

Doing the same thing, but adjusting the background color from `background-color:#f8f9f8` to `background-color:#e8e9e8` makes a big difference in being able to tell which menu item has focus:

![image](https://user-images.githubusercontent.com/2781561/29221029-8d27e1aa-7e93-11e7-8944-d80b92e3eaac.png)

tmorehouse:
Our current workaround is to darken the background:

```css
.dropdown-item:focus:not(.active),
.dropdown-item:hover:not(.active) {
    /* @See https://github.com/twbs/bootstrap/issues/23329 */
    box-shadow: inset 0px 0px 400px 110px rgba(0, 0, 0, .09);
}

.dropdown-item.active {
    box-shadow: initial;
}
```

browner12:
you can also use the following variables:

```sass
$dropdown-link-hover-color: #fff;
$dropdown-link-hover-bg: #f00;
```

pi0:
@browner12 While this suggestion is so nice, libraries like BootstrapVue just try to abstract bootstrap components (I mean not providing a pre-compiled BS css by default) and the hack above tries to override user them which is not a perfect solution. Maybe we can change variables for bootstrap defaults as an accessibility enhancement :)

browner12:
agreed, having a more contrasting default would be good.

patrickhlauke:
See https://github.com/twbs/bootstrap/pull/23990 for a partial fix for this.

tmorehouse:
Does this also address button focus styling?  It is possible to mix buttons and links together in a dropdown, and they should both look the same to the user

frederikhors:
Is there any news about it?

mdo:
This applies to v4 and v5. Would `$gray-200` be enough as opposed to the current `$gray-100`? And does this apply to our dark mode dropdown as well?

patrickhlauke:
[...]

PR #23990, opened 2017-09-18
Remove anchor focus override within dropdown menu

arizzitano:
The dropdown styles explicitly remove the browser's default focus outline for links within the menu, which [hurts accessibility](http://www.outlinenone.com/) as it ~~violates [WCAG 2.0 AA section 2.4.7](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#navigation-mechanisms-focus-visible)~~ [edit: technically, the focus highlight background color prevents it from failing]. Without the focus outline, keyboard-only users cannot orient themselves within a dropdown menu, which can be especially troublesome if the implementor has also applied a focus trap.

This PR removes the override, enabling the browser to apply its own native focus outline to links within the dropdown menu. The outline will only appear if a link has `:focus`, so **mouse users will see virtually no change**.

Suggestions welcome -- happy to amend as necessary!

### Screenshots
Before | After
|------|------|
image | image


patrickhlauke:
to clarify, no: the current situation does not fail wcag 2 2.4.7 as there is a visible focus indication (the background color applied to focused dropdown item). you can argue about whether or not the focus indication itself could be stronger, but as it stands it does nominally pass 2.4.7

arizzitano:
I updated the description accordingly, but I stand by the override removal. It's entirely reasonable for users to expect a focus outline on links, since that's a browser default. Is there a reason for keeping the override?

andresgalante:
I don't know about wcag 2 2.4.7, but I think that a focus ring can be stronger.

tmorehouse:
I too agree that the current dropdown focus background is rather light, and depending on screen contrast is almost invisible.  I have had keyboard only users complain about the inability to tell which dropdown item they have focused (screen readers users are OK with it).

We have added a box shadow with a large spread to darken the focus highlight.  Non keyboard users didn't like the focus ring we applied, so we went with a darker background.

arizzitano:
Just to be clear, this native focus ring will not show up for non-keyboard users. It only appears when focus is applied via the keyboard (or programmatically), and doesn't affect the hover state at all.

patrickhlauke (approved):

patrickhlauke:
LGTM. @mdo ?

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/23329

tmorehouse:
This addresses link dropdown items, but not button dropdown items.

PR #24000, opened 2017-09-19
Add back cursor: pointer for .btn-link

jeffbowen:
https://github.com/twbs/bootstrap/commit/232e86d0b4e1a4ff0dc3aa4e6d75c14a54fa8ac8 makes total sense but buttons with class `.btn-link` should still display a cursor.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -76,6 +76,7 @@ fieldset[disabled] a.btn {
   font-weight: $font-weight-normal;
   color: $link-color;
   background-color: transparent;
+  cursor: pointer;
   border-radius: 0;
 
   @include hover {


patrickhlauke (commented):
I'm split on this. I can see the rationale (it's styled to look like a link, so should have the standard pointer cursor for links), but the same argument could be made in reverse for `.btn` (it's styled like a button, so a link styled as button should have the default cursor for buttons). deferring to @mdo for decision.

patrickhlauke:
as an added point: i can see this whole thing confusing users either way. a button styled to look like a link, but without the `cursor:pointer`, will look strange to some users; a button styled as a link, and even more so when it DOES have `cursor:pointer`, may lead some users to think they can do things like middle-click to open in new tab, or get a link-specific context menu... (but that's a more fundamental issue with mimicking the look of a link)

mdo:
Nullified by #25082.

Issue #23928, opened 2017-09-12
bootstrap 3 demo errors


peter-mumford:
There are a bunch of errors on the Bootstrap 3 page. See screenshot.
![image](https://user-images.githubusercontent.com/31707793/30340892-0a981532-97c2-11e7-8a83-01e7ede87c71.png)

patrickhlauke:
images needs their initial slash removed (e.g. "/assets/img/expo-lyft.jpg" > "assets/img/expo-lyft.jpg")

XhmikosR:
If anyone makes a PR please CC me.

patrickhlauke:
any hints where these files (old docs) actually live in the repo?

XhmikosR:
It's the master branch I believe.

lucascono:
@mdo has closed the last PR as fixed. Is help still needed?

peter-mumford:
No more broken images.. there is still an https error and a 404 from a missing .swf file.

mdo:
No more errors for me.

PR #24013, opened 2017-09-20
[FIX] Docs site was not loading theme images because of URL path used.

bitsapien:
This PR relates to issue #23928

### DIFF
--- a/docs/_includes/footer.html
+++ b/docs/_includes/footer.html
@@ -6,8 +6,8 @@
     <ul class="bs-docs-footer-links">
       <li><a href="{{ site.repo }}">GitHub</a></li>
       <li><a href="https://twitter.com/getbootstrap">Twitter</a></li>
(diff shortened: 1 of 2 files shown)


XhmikosR:
Did you try the latest master? `baseurl` is defined there.

bitsapien:
@XhmikosR Yes I'm using `master`. 

`baseurl` is set correctly to `""`. But, our assets for docs do not live at `getbootstrap.com/`, instead they live at `getbootstrap.com/docs/3.3/`. Which is why having a `site.baseurl` does not make sense there.

XhmikosR:
@bitsapien: do we have any other cases like this? It's been a long time since I actually worked on v3 and its docs so I'm not sure.

bitsapien:
@XhmikosR The footer links for `Examples` and `About` also throw a `404` on the same page. Will push the changes in a moment.

mdo:
Works on my machine 😜. (jk i fixed it)

PR #24014, opened 2017-09-20
Rewritten tooltip and popover without jquery

alekitto:
Tooltip and popover rewritten without jquery

The event system has been modified to allow registering the same handler for more than one event, delegate events in `.one` method and unregistering one-off handlers through `.off`
The `delegateTarget` property has been added to the event object, replacing the jquery `currentTarget` (native events' currentTarget property is read-only)

### DIFF
--- a/js/src/dom/eventHandler.js
+++ b/js/src/dom/eventHandler.js
@@ -9,13 +9,11 @@ import Util from '../util'
 
 const EventHandler = (() => {
 
-
   /**
    * ------------------------------------------------------------------------
    * Polyfills
    * ------------------------------------------------------------------------
    */
-
 
   // defaultPrevented is broken in IE.
   // https://connect.microsoft.com/IE/feedback/details/790389/event-defaultprevented-returns-false-after-preventdefault-was-called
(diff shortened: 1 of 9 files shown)


Johann-S:
Did you update our event system because you had issue with it ? Because currently we are able to register for more than one event on the same element

js/src/dom/eventHandler.js
```diff
@@ -93,11 +93,13 @@ const eventRegistry = {}
 let uidEvent = 1
 
 function getUidEvent(element, uid) {
-  return element.uidEvent = uid && `${uid}::${uidEvent++}` || element.uidEvent || uidEvent++
```
Johann-S: change not relevant here, please keep the affectation here

js/src/util.js
```diff
@@ -77,6 +77,9 @@ const Util = (() => {
     },
 
     supportsTransitionEnd() {
+      if (transition === null) {
```
Johann-S: you changed that again :laughing:

js/src/tooltip.js
```diff
@@ -31,10 +33,13 @@ const Tooltip = (() => {
   const VERSION             = '4.0.0-beta'
   const DATA_KEY            = 'bs.tooltip'
   const EVENT_KEY           = `.${DATA_KEY}`
-  const JQUERY_NO_CONFLICT  = $.fn[NAME]
   const TRANSITION_DURATION = 150
   const CLASS_PREFIX        = 'bs-tooltip'
-  const BSCLS_PREFIX_REGEX = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g')
+  const BSCLS_PREFIX_REGEX  = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g')
+
+  function noop() {
```
Johann-S: `noop` should be added to our `Util` object

alekitto:
Before this, the same handler function cannot be registered and unregistered correctly for more than one element or event.
This would lead to `mouseover` event to be not unregistered on touch-enabled device (in tooltip.js:383)

That's why I removed the `uidEvent` setting in `getUidEvent` function, which is called with the original handler function as first argument in the `on` method

js/src/tooltip.js
```diff
@@ -468,12 +483,11 @@ const Tooltip = (() => {
 
       triggers.forEach((trigger) => {
         if (trigger === 'click') {
-          $(this.element).on(
+          EventHandler.on(this.element,
             this.constructor.Event.CLICK,
-            this.config.selector,
+            this.config.selector || '',
```
Johann-S: why ?

js/src/tooltip.js
```diff
@@ -482,29 +496,28 @@ const Tooltip = (() => {
...
-              (event) => this._enter(event)
-            )
-            .on(
-              eventOut,
-              this.config.selector,
-              (event) => this._leave(event)
-            )
+          EventHandler.on(this.element,
+            eventIn,
+            this.config.selector || '',
```
Johann-S: same here why ?

js/src/tooltip.js
```diff
@@ -190,14 +194,14 @@ const Tooltip = (() => {
...
+        let context = Data.getData(event.delegateTarget, dataKey)
 
         if (!context) {
           context = new this.constructor(
-            event.currentTarget,
+            event.delegateTarget,
             this._getDelegateConfig()
           )
-          $(event.currentTarget).data(dataKey, context)
+          Data.setData(event.delegateTarget, dataKey, context)
```
Johann-S: I don't think `setData` handle 3 arguments :thinking:

Johann-S (changes requested):

js/src/util.js
alekitto: Argh! This was an old branch, it seems that i've messed up this in a rebase..

js/src/tooltip.js
alekitto: 👍

js/src/tooltip.js
alekitto: Signature is `setData(instance, key, data)`

Johann-S: Oops, my bad :+1:

js/src/tooltip.js
alekitto: `.on` method explicitly checks for empty string, but selector can be `undefined` or `null`
Probably the check event handler should be modified

Johann-S:
Can you make some tests on IE, on your computer ? because currently we cannot run our Saucelabs tests on folks PRs :cry:

alekitto:
I've successfully executed the testsuite on chrome, chrome canary, firefox, safari, ie 11 and ms edge. Manual visual tests seem ok.

js/src/util.js
```diff
@@ -189,6 +189,10 @@ const Util = (() => {
       return false
     },
 
+    noop() {
+      return null
```
Johann-S: `noop` should return an empty function

js/tests/unit/dom/eventHandler.js
```diff
@@ -254,4 +254,18 @@ $(function () {
     EventHandler.trigger(element, 'click')
     document.body.removeChild(element)
   })
+
+  QUnit.test('off should remove a listener registered by .one', function (assert) {
+    assert.expect(0)
+
+    var element = document.createElement('div')
+    var handler = function () {
+      assert.ok(true, 'listener called')
```
Johann-S: you should use `notOk` here because it's a failed case

Johann-S (changes requested):

js/src/util.js
alekitto: `noop` should be an empty function, but ESLint does not allow empty functions

Johann-S: Disable that rule just for this line ;)

alekitto: Done

js/tests/unit/dom/eventHandler.js
alekitto: 👍

js/src/tooltip.js
```diff
@@ -614,10 +621,15 @@ const Tooltip = (() => {
     }
 
     _getConfig(config) {
-      config = $.extend(
+      if (typeof config !== 'undefined' &&
+        typeof config.container === 'object' && config.container.jquery) {
```
Johann-S: what is it `config.container.jquery` ?

alekitto: Checks if `config.container` is a jquery object. In that case i'll get the dom element via `[0]` .

Johann-S: that's right, I forgot about that property :smile:

PR #24015, opened 2017-09-20
Adds breakpoints to sizing utility classes

### LINKED ISSUE #24003: No responsive classes for the w-* class.
Will it be possible to add responsive classes to the w-* class?
For example, only w-0, w-25, w-50, w-75 and w-100 currently exist. Can classes such as w-sm-100 or s-xl-50 be added>

### LINKED ISSUE #24062: responsive sizing classes docs
Hi,

if there is no plan to add responsive sizing classes what do you think about adding something similar as code below to sizing docs for easy use?

What happens is that I don't think there is many cases where developer can use same width class for every screen.  For ex. `.w-25` or `.w-50` class mostly doesn't look good on small screen size.

```scss
$size-breakpoints: (
  //xs: 0, already exists
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;

@each $breakpoint in (map-keys($size-breakpoints)) {
  $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
  //@each $prop, $abbrev in (width: w, height: h) {
  @each $prop, $abbrev in (width: w) {
     @each $size, $length in $sizes {
        .#{$abbrev}#{$infix}-#{$size} { #{$prop}: $length !important; }
      }
    }
  }
}
```

andresgalante:
This PR closes #24003 and it adds breakpoints to sizing utility classes.

It also updates the docs to surface this change.

@Johann-S you know my feelings about utility classes 💩  but I think that since we are doing this for spacers, adding it for sizing is valid.

Closes #24062 too

### DIFF
--- a/scss/utilities/_sizing.scss
+++ b/scss/utilities/_sizing.scss
@@ -1,8 +1,15 @@
 // Width and height
 
-@each $prop, $abbrev in (width: w, height: h) {
-  @each $size, $length in $sizes {
-    .#{$abbrev}-#{$size} { #{$prop}: $length !important; }
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    @each $prop, $abbrev in (width: w, height: h) {
+      @each $size, $length in $sizes {
+        .#{$abbrev}-#{$size} { #{$prop}: $length !important; }
+        .#{$abbrev}#{$infix}-#{$size} { #{$prop}: $length !important; }
+      }
+    }
   }
 }
 
--- a/docs/4.0/utilities/sizing.md
+++ b/docs/4.0/utilities/sizing.md
@@ -6,7 +6,29 @@ group: utilities
 toc: true
 ---
 
-Width and height utilities are generated from the `$sizes` Sass map in `_variables.scss`. Includes support for `25%`, `50%`, `75%`, and `100%` by default. Modify those values as you need to generate different utilities here.
+## Overview
+
+Width and height utilities are generated from the `$sizes` Sass map in `_variables.scss`. Includes support for `25%`, `50%`, `75%`, and `100%` by default. Modify those values as you need to generate different sizing utilities.
+
+## Notation
+
+Sizing utilities are named using the format `{property}-{size}` for `xs` and `{property}-{breakpoint}-{size}` for `sm`, `md`, `lg`, and `xl`.
+
+Where *property* is one of:
+
+* `w` - for classes that set `width`
+* `h` - for classes that set `height`
+
+Where *size* is one of:
+
+* `25` - (by default) for classes that set the `width` or `height` to `25%`
+* `50` - (by default) for classes that set the `width` or `height` to `50%`
+* `75` - (by default) for classes that set the `width` or `height` to `75%`
+* `100` - (by default) for classes that set the `width` or `height` to `100%`
+
+(You can add more sizes by adding entries to the `$sizes` Sass map in `_variables.scss`)
+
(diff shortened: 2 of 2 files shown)


soullivaneuh (approved):

Herst:
While at it wouldn't printing-specific versions also be useful?

andresgalante:
@Herst yeap, it would make sense. Help me out a bit, what do you have in mind for printing? 

A question I keep asking myself when doing utility classes is how far do we want to take them, and when to stop.

Herst:
@andresgalante See for example: https://stackoverflow.com/a/32253969

Herst:
I separate PR makes much more sense, especially now that this one here has been reviewed. Sorry for hijacking it (I was sort-of thinking aloud).

andresgalante:
Closes #24062 too

XhmikosR:
@Johann-S: looks good to you?

XhmikosR (approved):

Johann-S:
Not against it but @mdo have concerns about our CSS file size, so I'll leave the end word to Mark here

andresgalante:
@XhmikosR thanks for reviewing it. I agree with @Johann-S, this is a "nice to have" to follow other utility classes, but I wouldn't mind if it doesn't get merged. it's not fixing any bug, just adding more options. So, it's more a matter of if we want this on Bootstrap or not.

On the other hand I have other open PRs that are fixing issues that we should merge.

mdo:
I'd rather hold off on this one—I want to verify these utilities get used much in the first place before we 5x the number of them by making them responsive :).

Herst:
@mdo Same concerning the idea with the printing classes I guess?

mdo:
Print-specific classes are probably fine? Depends on the situation. Open an issue to discuss?

scss/utilities/_sizing.scss
```diff
@@ -1,8 +1,15 @@
...
-@each $prop, $abbrev in (width: w, height: h) {
-  @each $size, $length in $sizes {
-    .#{$abbrev}-#{$size} { #{$prop}: $length !important; }
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    @each $prop, $abbrev in (width: w, height: h) {
+      @each $size, $length in $sizes {
+        .#{$abbrev}-#{$size} { #{$prop}: $length !important; }
```
iamandrewluca: @andresgalante it seems this line is not needed
when `$breakpoint` is `xs`, `$infix` is an empty string

iamandrewluca: Adding sizing for all breakpoints I use.
```scss
// Create sizing classes for all screen sizes
  $infix: breakpoint-infix($breakpoint, $grid-breakpoints);

  @if $infix != '' {
      @each $prop, $abbrev in (width: w, height: h) {
        @each $size, $length in $sizes {
          .#{$abbrev}#{$infix}-#{$size} { #{$prop}: $length !important; }
        }
      }
    }
  }
}
```

XhmikosR: @iamandrewluca: please make a PR; commenting in older PRs/issues will probably be ignored.

iamandrewluca: @XhmikosR I just wanted to mention this to @andresgalante . As I see this is not being merged. Should I create a PR in any case?

andresgalante: @iamandrewluca you are right, thanks a lot for reviewing it. 😄

XhmikosR:
I'd still like to have this merged.

It seriously improves usage.

XhmikosR:
@mdo: are you still opposed to this?

PR #24019, opened 2017-09-20
Carousel - Add swipe support.

dimasikdimasik:


Johann-S:
Thanks @dimasikdimasik to open PR with my work :smile: but I'm not sure if it's enough to add swipe support to our carousel, so I'll open a PR when I'll be sure of my work :+1:

PR #24026, opened 2017-09-21
remove .half-rule from docs CSS in favor of utils

new-object:


Johann-S:
Please create PR of your work :+1:

PR #24055, opened 2017-09-22
Improves colors on drop menus on navbar

### LINKED ISSUE #23915: Is this correct  rendering collapsed menu on mobile?
image
image

andresgalante:
This PR fixes #23915 and it looks like this:

Image of the navbar menu

Notice that I had to reverse the order of the declarations on the `_navbar.scss` putting first themes and then breakpoints to avoid increasing specificity on themes.

also, I've created variables for dropmenu styles on navbar so users have full control of them.

@mdo @xhmikosr @Johann-S  Since this is not a tiny change I did as much test I could and I think it's solid, but a fresh pair of eyes is really welcome 😄

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -130,6 +130,130 @@
   background-size: 100% 100%;
 }
 
+// Navbar themes
+//
+// Styles for switching between navbars with light or dark background.
+
+// Dark links against a light background
+.navbar-light {
+  .navbar-brand {
+    color: $navbar-light-active-color;
+
+    @include hover-focus {
+      color: $navbar-light-active-color;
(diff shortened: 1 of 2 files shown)


XhmikosR:
Judging for the end result (your gif) it looks fine for me. But we'll need @mdo to confirm.

mdo:
Going to hold off on this one for a bit longer...

mdo:
Going to punt on this, thanks though @andresgalante! We could revisit in a v5 at some point, where we change the expectations around navbars and dropdowns, but I think this is a bit much for right now. Happy to revisit though as always if sentiment changes :).

PR #24060, opened 2017-09-23
Set font-sizes relative to font-size-base

### LINKED ISSUE #24056: Set all font-sizes relative to $font-size-base
If all font-sizes are relative to $font-size-base, it would be simpler to change overall font-size:

```css
$font-size-base: 1rem !default;
$font-size-lg:   ($font-size-base * 1.25) !default;
$font-size-sm:   ($font-size-base * 0.875) !default;
```

tillmon:
This PR closes #24056


andresgalante:
😄 nice, I like this one! 

What about `$lead-font-size` ? should we base it on `$font-size-base` too?

scss/_variables.scss
```diff
@@ -216,12 +216,13 @@ $font-weight-bold: bold !default;
 $font-weight-base: $font-weight-normal !default;
 $line-height-base: 1.5 !default;
 
-$h1-font-size: 2.5rem !default;
-$h2-font-size: 2rem !default;
-$h3-font-size: 1.75rem !default;
-$h4-font-size: 1.5rem !default;
-$h5-font-size: 1.25rem !default;
-$h6-font-size: 1rem !default;
+$headings-font-size-base: $font-size-base !default;
```
andresgalante: I wouldn't introduce a new base for headers

andresgalante:
@mdo @XhmikosR I took another look at this PR. We have several font-size references across _variables.scss, and since they are all set in rem (or %) it might be easier to just to change the font-size base definition from `body` here:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L60

to under `html` here:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L28

That way by changing the `font-size-base` variable it will change the root and everything will follow. No changes needed any other variable and no extra layer of abstraction.

If you are ok with it, I can make the change and write a note on the docs to make this clear.

What do you think?

mdo (approved):

PR #24064, opened 2017-09-23
Update contributing.md with correct branch (`v4-dev` instead of `master`)

patrickhlauke:


bardiharborow (approved):

vsn4ik:
Merged 1 commit into `master`?

patrickhlauke:
yes, because contributing.md is only present in master (compared to any commits authors would want to make against actual bootstrap, where they'd need to do it against `v4-dev`)

PR #24069, opened 2017-09-24
Fix No Rounded Corners if there is one card in a card group.

drobson03:
This was actually killing me.


gijsbotje:
you included all the dist files in your pr, these should be left out as they will be generated on release

Johann-S:
yes please revert your changes in our `dist` folder

andresgalante:
Also avoid using !important

PR #24080, opened 2017-09-25
Fix no rounded corners if there is only one card in a card-group.

drobson03:


XhmikosR:
@andresgalante

andresgalante (approved):

andresgalante:
I've just pulled @SpicyDwarf branch and tested it. This PR fixes the issue. Let me describe it better:

When you have only one card inside a card-group the first and last child selector will not affect it since it's the only child. Then it looks like this:

screen shot 2017-09-25 at 5 51 01 pm

This PR is introducing the border radius if the card group container has only one card inside.

@SpicyDwarf Nice job!

XhmikosR (approved):

PR #24082, opened 2017-09-25
Use variable for heading margin bottom in reboot

psimyn:
This margin is overridden in [_type](https://github.com/twbs/bootstrap/blob/45722e94acdffc314d7f835ed8a73cd0d56d879a/scss/_type.scss#L7) anyway, but I figure the variable should be used in reset

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -93,7 +93,7 @@ hr {
 // margin for easier control within type scales as it avoids margin collapsing.
 h1, h2, h3, h4, h5, h6 {
   margin-top: 0;
-  margin-bottom: .5rem;
+  margin-bottom: $headings-margin-bottom;
 }
 
 // Reset margins on paragraphs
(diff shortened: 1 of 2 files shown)


andresgalante:
If we do this change, I'd remove it from here:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_type.scss#L7

So we don't repeat the same declaration in 2 places. It get confusing.

mdo:
Going to pass on this. The two rulesets here are actually a different set of selectors—note that the `_type.scss` one includes class versions of our heading elements. The Reboot one is only for the HTML elements. We could potentially remove the `margin-bottom` reset from Reboot and only have it in `_type.scss`, but that feels incomplete in a way.

psimyn:
@mdo it makes sense to keep both. This was originally only intended to replace hard-coded value with a variable. I will resubmit with just that change

Issue #24073, opened 2017-09-24
[BS 4.0.0-beta] Collapse parent option expects only a string


julianxhokaxhiu:
Hi all,

I was trying to use the new Boostrap 4.0.0 Beta and until know I must say I am very satisfied, looks great and everything is *much* better compared to 3.x :)

Although I think I found a bug, or maybe a regression regarding the Collapse component, compared to Bootstrap 3.x

---

## Bootstrap 4.x
**Test case:** http://jsbin.com/yomagujacu/edit?html,js,output
**Expected outcome:** The collapse just works.
**Current result:**
The collapse fires this error once the page is loaded:
```
Uncaught Error: COLLAPSE: Option "parent" provided type "element" but expected type "string".
```

**Report:**
- _Operating system:_ Windows 10 10.0.15063 x64
- _Browser:_ Google Chrome 61.0.3163.100 (Official build) (64 bit) (cohort: 61_Win_100)

## Bootstrap 3.x
**Test case:** http://jsbin.com/vufuhafuki/edit?html,js,output
**Expected outcome:** The collapse just works.
**Current result:** The collapse just works.

**Report:**
- _Operating system:_ Windows 10 10.0.15063 x64
- _Browser:_ Google Chrome 61.0.3163.100 (Official build) (64 bit) (cohort: 61_Win_100)

---

Thank you and best regards,
Julian

PR #24083, opened 2017-09-25
Collapse - Allow to pass jQuery object or DOM element to the parent option

Johann-S:
Allow to pass jQuery object or DOM element to the parent option

Close : #24073

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -33,7 +33,7 @@ const Collapse = (() => {
 
   const DefaultType = {
     toggle : 'boolean',
-    parent : 'string'
+    parent : '(string|element)'
   }
 
   const Event = {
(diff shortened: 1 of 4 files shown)


js/src/collapse.js
```diff
@@ -289,7 +289,18 @@ const Collapse = (() => {
     }
 
     _getParent() {
-      const parent   = $(this._config.parent)[0]
+      let parent = null
+      if (Util.isElement(this._config.parent)) {
+        parent = this._config.parent
+
+        // it's a jQuery object
+        if (typeof this._config.parent.jquery !== 'undefined') {
```
vsn4ik: Maybe
```
typeof this._config.parent.jquery !== 'undefined'
```
->
```
this._config.parent instanceof jQuery
```
?

Johann-S: yep works too :+1:

XhmikosR (approved):
LGTM since Browser tests pass.

Issue #23378, opened 2017-08-13
Popper.js messing up dropdown styles


myorangeca:
Loving the new popper.js but was wondering if there was a way to disable the custom styles it auto adds to elements?

Its adding the following:
```
    position: absolute;
    transform: translate3d(0px, 0px, 0px);
    top: 0px;
    left: 0px;
    will-change: transform;
```

This in turn overrides some custom animations I am trying to implement, it also messes up alignment if I wanted to put dropddowns in a container. The plugin is great but there must be a way to disable this custom "hardcode" style feature?

FezVrasta:
These styles are the ones needed to actually position the dropdown. Without them the dropdown would not be positioned.

Johann-S:
Yep but that's possible to disable Popper.js style we did it when Dropdown are in a navbar so why not add a feature to allow folks to do that

FezVrasta:
But then what's the point to use a dropdown?

Johann-S:
Yep I agree they won't take any advantage of Popper.js

mdo:
Perhaps @myorangeca can elaborate on the custom animations being applied so we can see what properties conflict? I definitely understand the desire to customize animations for positioned elements.

FezVrasta:
It's probably because of the `transform` property. Ideally dropdowns should allow an intermediate wrapper people can use to apply custom css and animations to it.

Doing so the outer most wrapper is used for the position and the inner one for the animations

myorangeca:
I have customized the dropdown to do the following animation below. I had to use !important style to override popper.js styles (clearly not the best solution). 

```
.show > .dropdown-menu {
    -moz-transform: scale(1) !important;
    -o-transform: scale(1) !important;
    -ms-transform: scale(1) !important;
    -webkit-transform: scale(1) !important;
    transform: scale(1) !important;
    moz-transform-origin: 29px -50px;
    -o-transform-origin: 29px -50px;
    -ms-transform-origin: 29px -50px;
    -webkit-transform-origin: 29px -50px;
    transform-origin: 29px -50px;
    opacity: 1;
    visibility: visible;
}

.dropdown-menu {
    -moz-transform: scale(0.8) !important;
    -o-transform: scale(0.8) !important;
    -ms-transform: scale(0.8) !important;
    -webkit-transform: scale(0.8) !important;
    transform: scale(0.8) !important;
    transition: all 270ms cubic-bezier(0.34, 1.25, 0.3, 1);
    opacity: 0;
[...]

PR #24092, opened 2017-09-25
Dropdown - Allow to disable Popper.js style

Johann-S:
Add `display` option to our Dropdown which will allow folks to use Popper.js if display is equal to `dynamic` (default value) or to not use Popper.js with display equal to `static`

Close: https://github.com/twbs/bootstrap/issues/23378

/CC @XhmikosR , @mdo

### DIFF
--- a/docs/4.0/components/dropdowns.md
+++ b/docs/4.0/components/dropdowns.md
@@ -778,6 +778,12 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>'toggle'</td>
       <td>Reference element of the dropdown menu. Accepts the values of <code>'toggle'</code>, <code>'parent'</code>, or an HTMLElement reference. For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#referenceObject">referenceObject docs</a>.</td>
     </tr>
+    <tr>
+      <td>display</td>
+      <td>string</td>
+      <td>dynamic | static</td>
+      <td>By default we use Popper.js but if you want to disable Popper.js use `static`</td>
+    </tr>
   </tbody>
 </table>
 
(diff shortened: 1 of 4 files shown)


XhmikosR:
I'm not very familiar with dropdowns since I don't use them currently. That being said, I get that you are trying to provide a way for people to override Popper.js. That seems good to me, but better wait for @mdo's confirmation.

mdo (approved):

mdo:
Looks good, but I'd suggest saving for v4.1 maybe?

Johann-S:
As you want @mdo :+1: they are no rush for this feature :smile:

Jake0oo0:
I'm not sure what the outlook for v4.1 is, but in the mean time a quick note in the docs about popperjs automatically being disabled in the Navbar might useful. 

I had to find these commits to realize that Bootstrap arbitrarily determines if the dropdown is within a navbar based on the existence of the `navbar` class. I'm using a custom navbar (not sure why I'm using BS at this point, but that's irrelevant), and had to find this to fix the broken dropdowns on mobile.

Johann-S:
Please @Jake0oo0 open a separate issue about that and we will fix that for sure 👍

Johann-S:
I updated that PR @mdo and @XhmikosR if it's planed to be in our v4.1

XhmikosR:
LGTM as long as tests pass.

XhmikosR:
But better wait for @mdo's feedback too.

XhmikosR (approved):

sagudev:
Why Bootstrap need Popper.js anyway

Herst:
@sagudev Popovers and tooltips.

Johann-S:
If we disable Popper.js in our dropdown we will have to do the positionning by CSS for dropleft and dropright

mdo (approved):

Johann-S:
@samdbeckham that's not secret it's written in our documentation see: https://getbootstrap.com/docs/4.1/components/dropdowns/#options

samdbeckham:
Just spotted it, I was looking at the 4.0 docs 🤦‍♂️

PR #24094, opened 2017-09-25
Rewritten tab without jquery

alekitto:
Related to #23586

### DIFF
--- a/js/src/dom/selectorEngine.js
+++ b/js/src/dom/selectorEngine.js
@@ -1,3 +1,5 @@
+import Util from '../util'
+
 /**
  * --------------------------------------------------------------------------
  * Bootstrap (v4.0.0-beta): dom/selectorEngine.js
@@ -29,17 +31,13 @@ const SelectorEngine = (() => {
   if (!Element.prototype.closest) {
     fnClosest = (element, selector) => {
       let ancestor = element
-      if (!document.documentElement.contains(element)) {
-        return null
-      }
-
       do {
         if (fnMatches.call(ancestor, selector)) {
           return ancestor
         }
 
         ancestor = ancestor.parentElement
-      } while (ancestor !== null)
+      } while (ancestor !== null && ancestor.nodeType === Node.ELEMENT_NODE)
 
       return null
     }
@@ -50,12 +48,67 @@ const SelectorEngine = (() => {
     }
   }
 
+  const scopeSelectorRegex = /:scope\b/
+  const supportScopeQuery = (() => {
+    const element = document.createElement('div')
+    try {
+      element.querySelectorAll(':scope *')
+    } catch (e) {
+      return false
+    }
+
+    return true
+  })()
+
+  let findFn = null
+  let findOneFn = null
+  if (supportScopeQuery) {
+    findFn = Element.prototype.querySelectorAll
+    findOneFn = Element.prototype.querySelector
+  } else {
+    findFn = function (selector) {
+      if (!scopeSelectorRegex.test(selector)) {
+        return this.querySelectorAll(selector)
+      }
+
+      const hasId = Boolean(this.id)
+      if (!hasId) {
+        this.id = Util.getUID('scope')
+      }
+
+      let nodeList = null
+      try {
+        selector = selector.replace(scopeSelectorRegex, `#${this.id}`)
+        nodeList = this.querySelectorAll(selector)
+      } finally {
+        if (!hasId) {
+          this.removeAttribute('id')
+        }
+      }
+
+      return nodeList
+    }
+
+    findOneFn = function (selector) {
+      if (!scopeSelectorRegex.test(selector)) {
+        return this.querySelector(selector)
+      }
+
+      const matches = findFn.call(this, selector)
+      if (typeof matches[0] !== 'undefined') {
+        return matches[0]
+      }
+
+      return null
+    }
+  }
+
   return {
     matches(element, selector) {
       return fnMatches.call(element, selector)
     },
 
-    find(selector, element = document) {
+    find(selector, element = document.documentElement) {
       if (typeof selector !== 'string') {
         return null
       }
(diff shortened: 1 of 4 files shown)


js/src/dom/selectorEngine.js
```diff
@@ -50,12 +48,71 @@ const SelectorEngine = (() => {
...
+  } else {
+    findFn = function (selector) {
+      if (!scopeSelectorRegex.test(selector)) {
+        return this.querySelectorAll(selector)
+      }
+
+      const hasId = Boolean(this.id)
+      if (!hasId) {
+        // eslint-disable-next-line no-magic-numbers
+        this.id = `q${Math.floor(Math.random() * 9000000) + 1000000}`
```
Johann-S: Use `Util.getUID` (https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/util.js#L51) instead of that

js/src/tab.js
```diff
@@ -46,10 +46,10 @@ const Tab = (() => {
     DROPDOWN              : '.dropdown',
     NAV_LIST_GROUP        : '.nav, .list-group',
     ACTIVE                : '.active',
-    ACTIVE_UL             : '> li > .active',
+    ACTIVE_UL             : ':scope > li > .active',
```
Johann-S: why did you changed that selector ?

alekitto: `> li > .active` is not a valid selector. Using with `querySelector/querySelectorAll` throws a `SyntaxError`.

Johann-S: yep so you should find `li` children and after that `.active` elements, no need here to implement `:scope`

alekitto: Sorry, can't understand. Are you suggesting not to use a selector? Or a double use of `.children`? In both cases it could be very slow

js/src/dom/selectorEngine.js
```diff
@@ -50,12 +48,70 @@ const SelectorEngine = (() => {
...
+          this.removeAttribute('id')
+        }
+      }
+
+      return nodeList
+    }
+  }
+
+  let findOneFn = null
+  if (supportScopeQuery) {
```
Johann-S: you should merge those two `if`

js/src/dom/selectorEngine.js
```diff
@@ -29,17 +31,13 @@ const SelectorEngine = (() => {
   if (!Element.prototype.closest) {
     fnClosest = (element, selector) => {
       let ancestor = element
-      if (!document.documentElement.contains(element)) {
```
Johann-S: this polyfill come from MDN, why did you change that ?

js/src/tab.js
Johann-S: Nope you're right that's better like that, sorry I didn't know `:scope` before your PR :smile:

js/src/dom/selectorEngine.js
```diff
@@ -64,21 +120,24 @@ const SelectorEngine = (() => {
...
+      return findFn.call(element, selector)
     },
 
-    findOne(selector, element = document) {
+    findOne(selector, element = document.documentElement) {
       if (typeof selector !== 'string') {
         return null
       }
 
-      let selectorType = 'querySelector'
```
Johann-S: Not a big fan of that, because if it's an ID you'll use `querySelectorAll` instead of `getElementById` which is faster

Johann-S (changes requested):

js/src/dom/selectorEngine.js
alekitto: Ok

js/src/dom/selectorEngine.js
alekitto: Tests are failing with that `if`. In any case, the DOM Standard specification (https://dom.spec.whatwg.org/#dom-element-closest) doesn't require the element to be a discendant of the `document` object

Johann-S: that's right but I don't see a use case where the element should be outside of the `document.documentElement` :thinking:

alekitto: Probably some tests don't add elements to the DOM before execution. If you prefer, I can revert this change and fix those tests.

Johann-S: No that's ok :+1:

js/src/dom/selectorEngine.js
alekitto: Multiple selector starting with `#` (ex: `#id > li, #id1 > .active`) will match and return a wrong result

alekitto: However, I think this is in the wrong PR, i've encountered an error on this while working on ScrollSpy 😅

alekitto: @Johann-S Should I revert this? Or fix testing the selector against a regex that excludes complex selectors?

js/src/tab.js
artemsky: :scope deprecated?) https://developer.mozilla.org/en-US/docs/Web/CSS/:scope

alekitto: Nothing indicates that `:scope` selector has been deprecated. It is part of the CSS Selectors Level 4 Working Draft. Some browsers have removed  the `scoped` style tag attribute, but the `:scope` pseudo-class is still supported and works correctly in CSS (works the same as `:root`) and in `querySelector/querySelectorAll`

js/src/dom/selectorEngine.js
Johann-S: Nope finally I think it's fine like this 👍

Johann-S:
@alekitto can you update your branch please ? 👍

alekitto:
@Johann-S done 👍

js/src/tab.js
```diff
@@ -171,54 +164,49 @@ const Tab = (($) => {
...
       }
 
-      $(element).addClass(ClassName.ACTIVE)
+      element.classList.add(ClassName.ACTIVE)
       if (element.getAttribute('role') === 'tab') {
         element.setAttribute('aria-selected', true)
       }
 
-      if (isTransitioning) {
+      if (isTransitioning || !active) {
```
Johann-S: It seems an unrelated change here 🤔 Maybe can you explain it ?

alekitto: This check was introduced to fix fading when no initial panel was active, but i've probably missed a test. I'll push it asap, but if you want to keep things separated, i can revert this check and submit a pr just for this case.

Johann-S: Is it an issue in our current `v4-dev` branch ?

alekitto: I’ve just tested it on the v4-dev, the issue is present in the visual test page

Johann-S: So I prefer if you can create a PR (which target `v4-dev`) to resolve that or if you can create an issue which describe that

alekitto: Correct. I've pushed a PR to explain and resolve the issue on `v4-dev` branch

alekitto: I've reverted this change for this PR, as it is not related

Johann-S (approved):

PR #24099, opened 2017-09-26
v4 Dropdown without jQuery

Johann-S:
Work in progress...

### DIFF
--- a/js/src/dom/eventHandler.js
+++ b/js/src/dom/eventHandler.js
@@ -339,7 +339,7 @@ const EventHandler = (() => {
       let defaultPrevented = false
 
       if (inNamespace && typeof $ !== 'undefined') {
-        jQueryEvent = new $.Event(event, args)
+        jQueryEvent = $.Event(event, args)
 
         $(element).trigger(jQueryEvent)
         bubbles = !jQueryEvent.isPropagationStopped()
@@ -347,8 +347,7 @@ const EventHandler = (() => {
         defaultPrevented = jQueryEvent.isDefaultPrevented()
       }
 
-      let evt           = null
-
+      let evt = null
       if (isNative) {
         evt = document.createEvent('HTMLEvents')
         evt.initEvent(typeEvent, bubbles, true)
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -1,5 +1,8 @@
-import $ from 'jquery'
+import Data from './dom/data'
+import EventHandler from './dom/eventHandler'
+import Manipulator from './dom/manipulator'
 import Popper from 'popper.js'
+import SelectorEngine from './dom/selectorEngine'
 import Util from './util'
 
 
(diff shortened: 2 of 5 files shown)


js/src/dropdown.js
```diff
@@ -416,35 +421,39 @@ const Dropdown = (() => {
...
-    .on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
-      event.preventDefault()
-      event.stopPropagation()
-      Dropdown._jQueryInterface.call($(this), 'toggle')
-    })
-    .on(Event.CLICK_DATA_API, Selector.FORM_CHILD, (e) => {
-      e.stopPropagation()
-    })
+  EventHandler.on(document, Event.KEYDOWN_DATA_API, Selector.DATA_TOGGLE, Dropdown._dataApiKeydownHandler)
+  EventHandler.on(document, Event.KEYDOWN_DATA_API, Selector.MENU, Dropdown._dataApiKeydownHandler)
```
Johann-S: @alekitto they are an issue here with our current EventHandler because if we have the same delegation handler for a different delegation selector no event listener is added due to that line : https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/dom/eventHandler.js#L225

Johann-S: Fixed by my last commit 👍

alekitto: Yes, you're right, I managed to reproduce the error in a test case. PR is coming to fix this.

js/src/dom/eventHandler.js
```diff
@@ -267,7 +257,19 @@ const EventHandler = (() => {
...
+        EventHandler.off(element, e.type, complete)
+        realHandler.apply(element, [e])
+      }
+      complete.isOne = true
+      if (isDelegation) {
+        addHandler(element, event, handler, complete)
+      } else {
+        addHandler(element, event, complete)
+      }
+      realHandler.uid = complete.uid.toString()
```
alekitto: @Johann-S Adding the `uid` property to the real handler function will cause troubles when using and removing the same handler for more than one event (like in https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/tooltip.js#L322).
It was changed in https://github.com/twbs/bootstrap/commit/fa90114c11212f1021ab22b5baf7201b4899f3a7 to take this into account, removing the setting of `uidEvent` property.
I thought that a test was written for this scenario, but i've probably missed it.

Johann-S: Yep that's because `Util.noop` always return the same function instead of a new one here but yes in case of a reusable handler it will be problematic

alekitto: Yes, with `Util.noop` shouldn't be a problem 😄 
However, https://github.com/twbs/bootstrap/pull/24487 should fix that, but i think you have to revert your last commit in order to merge it. If you can reopen it, I'll submit a couple of tests to cover this case.

Johann-S:
I'm stuck on this unit test : https://github.com/twbs/bootstrap/blob/v4-dropdown-without-jquery/js/tests/unit/dropdown.js#L598

It seems I don't understand how but when I trigger a click event on the input, the event is swallowed 🤔

PR #24100, opened 2017-09-26
Add image dimensions in stack image.

XhmikosR:


Johann-S (approved):

bardiharborow:
Can I confirm that we really want `alt` text for a purely decorative image?

patrickhlauke:
Personally, I'd say that image is mostly decorative, so `alt=""` is ok. The proposed new alt may actually make users who can't see the image wonder if this is some sort of diagram/visualization of the BS stack (with labels and arrows and such).

PR #24114, opened 2017-09-26
Fix .btn-group wrapping (#24111)

### LINKED ISSUE #24111: .btn-group should wrap buttons
I think Bootstrap should offer a behavior aligned with common responsive problems by default, so btn-groups should wrap buttons. And use "flex-nowrap" in "btn-group" when you need this specific infinite in-line behavior.

Besides, it would be more consistency with the previous version:

image

marcoscv-work:
Fixes #24111

Previous:
image

Wrap by default (this pull):
image

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -5,7 +5,7 @@
 .btn-group-vertical {
   position: relative;
   display: inline-flex;
-  flex-flow: row wrap;
+  flex-flow: wrap;
   vertical-align: middle; // match .btn alignment given font-size hack above
 
   > .btn {
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -5,6 +5,7 @@
 .btn-group-vertical {
   position: relative;
   display: inline-flex;
+  flex-flow: row wrap;
   vertical-align: middle; // match .btn alignment given font-size hack above
 
   > .btn {


scss/_button-group.scss
```diff
@@ -5,6 +5,7 @@
 .btn-group-vertical {
   position: relative;
   display: inline-flex;
+  flex-flow: row wrap;
```
andresgalante: The initial value of `flex-direction` is `row`. You just need to change the wrap here, not the directions.

I'd make this `flex-wrap: wrap;` instead.

andresgalante (changes requested):

scss/_button-group.scss
marcoscv-work: It make sense, PR re-send
https://github.com/twbs/bootstrap/pull/24120

marcoscv-work:
Resend PR here:
https://github.com/twbs/bootstrap/pull/24120

Johann-S:
you should update this one, no need to create a new one

andresgalante:
@marcoscv-work just make the change and push it again to your `origin btn-group-wrap` branch. It will update this PR

marcoscv-work:
Thanks, guys

I've added one commit more if you prefer a replace of the first commit only let me know :-)

Johann-S:
don't worry about that @marcoscv-work we have a `squash` option :wink:

andresgalante (approved):

mdo:
Closing per https://github.com/twbs/bootstrap/issues/24111#issuecomment-332282195.

PR #24115, opened 2017-09-26
Add composer component support

janseliger:
See [composer components](https://github.com/RobLoach/component-installer)

Supports easy use with require.js
I use the bundle, because of popper.js currently has no "component"

The styles are not in the possible styles section, f.e. `"component": { ..."styles": ["dist/js/bootstrap.css"] ...`, for the support of custom scss created styles


FezVrasta:
The problem with composer is that's just yet another Bower-like package manager.

It requires the dist files to be hosted in the git repository and it doesn't support momorepos.

If bootstrap is going to be a monorepo it will have to drop composer

Johann-S:
yep that's what we want to do in our next releases

P0rnflake:
@FezVrasta composer is capable of handling the distribution/compilation of the package itself by hooking into the distribution workflows (f.e. https://getcomposer.org/doc/articles/scripts.md). Even bower has the capabilty to trigger post-install scripts which can do the same. It's not impossible to have composer support, IMHO it would be a huge benefit for the community...

FezVrasta:
@P0rnflake, the problem is the monorepo support.

These package managers expect a single project for each git repository, while monorepos host several projects into a single repo.

P0rnflake:
@FezVrasta now I understood what your point is. Composer has this feature but only for SVN repositories, their opinion seems to be that git is not meant for handling multiple packages in one repository (@see https://github.com/composer/composer/issues/2588). Any thoughts on this?

Which package managers are you going to support in the long term?

FezVrasta:
I'm not part of the core maintainers so I don't decide anything.

cc @Johann-S

Johann-S:
In long term every package manager which allow us to ship different packages in one repository (npm, yarn, and maybe other)

FezVrasta:
Looks like Bootstrap shouldn't support Composer at all:

https://github.com/composer/composer/issues/2588#issuecomment-332563108

mdo:
Closing as a won't fix then. Holler if anyone on the team feels strongly about revisiting.

PR #24116, opened 2017-09-26
Minor Markdown consistency changes.

XhmikosR:


Johann-S:
Maybe can you update that file too : https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/getting-started/javascript.md#dependencies
to add that Dropdown depend on Popper.js too

Johann-S (approved):

PR #24127, opened 2017-09-26
example plugin: add `data-example-id` only if it's been supplied.

XhmikosR:
My Ruby knowledge is pretty basic, but this seems to do the job.


Johann-S:
What's the goal of this PR @XhmikosR ?

XhmikosR:
Without this, we insert `data-example-id` whenever the `example` filter is being used even without `id` being passed as a parameter.

bkeeton12 (approved):

bardiharborow (approved):
My Ruby is terrible too, but I think I can get away with reviewing a ternary operator. :)

PR #24132, opened 2017-09-27
prevent empty carousel from throwing, and staying in a 'sliding' state

jimmymain:
There are numerous questions on stack overflow relating to how the carousel becomes broken when it is emptied for any reason. This makes binding slides using angular or similar technologies very difficult, as they will routinely replace the set of slides with a new set.

whist I believe that calling `$.carousel()` should reset all the settings (particularly the `sliding` state), this simpler tweak at least ensures that if you use an angular / react binding to the carousel and it becomes empty, it does not get stuck in a sliding state.

please consider this pull - or something similar.

- https://stackoverflow.com/questions/29027830/getting-cannot-read-property-offsetwidth-of-undefined-with-bootstrap-carousel
- https://stackoverflow.com/questions/35918723/bootstrap-carousel-controls-causing-error
- https://stackoverflow.com/questions/29027830/getting-cannot-read-property-offsetwidth-of-undefined-with-bootstrap-carousel

### DIFF
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -144,7 +144,7 @@
     var slidEvent = $.Event('slid.bs.carousel', { relatedTarget: relatedTarget, direction: direction }) // yes, "slid"
     if ($.support.transition && this.$element.hasClass('slide')) {
       $next.addClass(type)
-      if (typeof $next == 'object' && $next.length) 
+      if (typeof $next == 'object' && $next.length)
       {
         $next[0].offsetWidth // force reflow
       }
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -144,7 +144,10 @@
     var slidEvent = $.Event('slid.bs.carousel', { relatedTarget: relatedTarget, direction: direction }) // yes, "slid"
     if ($.support.transition && this.$element.hasClass('slide')) {
       $next.addClass(type)
-      if (typeof $next == 'object' && $next.length) $next[0].offsetWidth // force reflow
+      if (typeof $next == 'object' && $next.length) 
+      {
+        $next[0].offsetWidth // force reflow
+      }
       $active.addClass(direction)
       $next.addClass(direction)
       $active
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -144,7 +144,7 @@
     var slidEvent = $.Event('slid.bs.carousel', { relatedTarget: relatedTarget, direction: direction }) // yes, "slid"
     if ($.support.transition && this.$element.hasClass('slide')) {
       $next.addClass(type)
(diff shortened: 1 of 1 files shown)


jimmymain:
I am not sure how to go about fixing the CI build.
Some documentation element is not correct - I am unsure of how this relates to the javascript change I made.

Please consider fixing this issue though - it's causing a fair few questions on github.

js/carousel.js
```diff
@@ -144,7 +144,10 @@
     var slidEvent = $.Event('slid.bs.carousel', { relatedTarget: relatedTarget, direction: direction }) // yes, "slid"
     if ($.support.transition && this.$element.hasClass('slide')) {
       $next.addClass(type)
-      $next[0].offsetWidth // force reflow
+      if (typeof $next == 'object' && $next.length)
```
Johann-S: please use `===` here

Johann-S (changes requested):

Johann-S:
Generally we don't accept any changes in v3 so for now I won't close this issue

/CC @mdo (for the final word)

jimmymain:
I understand that, but it's such a simple fix to a rather annoying problem - clearly not just experienced by myself.

js/carousel.js
jimmymain: i have made the requested correction.

XhmikosR:
My main concern is that this should target the master branch and not v3.4.0.

jimmymain:
I am happy to move it to another branch if there's a reasonable chance it will be pulled.

XhmikosR:
@mdo: should I merge this manually to master?

XhmikosR:
@Johann-S: does the change look good to you? Ideally I think we should also have a test case.

Johann-S:
Yep a unit test would be a good things because here I just see a way to force the reflow but not sure if it fix the issue

jimmymain:
Ideally Yes, but firstly it's a difficult test and secondly - the fix is
just a null check.

Johann-S:
Maybe with a visual test ? on a Codepen ?

Hima1717:
Nothing Found

jimmymain:
I will put a code pen together and post the link here.

jimmymain:
I have created a code pen that clearly shows the issue.
https://codepen.io/jimmymain/pen/boYeZV?editors=1111

If you click "Toggle Slide List" the list of slides is emptied, and immediate the javascript error is raised in the console. "Cannot Read Property 'offsetWidth'".

The slide list is then in a **stuck** state where it can no longer slide.
My fix resolves the issue.

Johann-S (approved):

XhmikosR:
Manually merged to master. Thanks!

Issue #23994, opened 2017-09-18
Custom input file after choice file nothing change.


madmax:
When we choice file nothing change (we don't see filename). 
https://getbootstrap.com/docs/4.0/components/forms/#file-browser

This should work this way? We need write JS to look change event and update filename inside custom-file-control? 

Tested on: 
Beta 4.0.0
Safari 11  and Chrome  60
MacOS 10.13

Johann-S:
That's true no update, should we add a simple plugin which handle that ? 
@mdo, @XhmikosR, @bardiharborow

Gaurav-Kumar-Rana:
Test

tmorehouse:
We have added file name update support in `<b-form-file>`, which is the Bootstrap-Vue implementation of Bootstrap's custom file input.  YOu might be able to make a plugin out of our file handler code:  https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/lib/components/form-file.vue

Basically it places the file names (or file count, in the case of multiple) in a `data-*` attribute which then picks up the file name(s) via CSS `content()` selector.

![image](https://user-images.githubusercontent.com/2781561/30622149-2036c194-9d86-11e7-8aad-5882d46dbcd5.png)

bardiharborow:
I thought we had code for this at some stage (not in a core plugin, but as an optional extra). Where did that go?

Johann-S:
I don't remember if we had code about that 😯

jplaverdure:
I'm preety sure the doc mentions someplace that filename update is not implemented 

But it's fairly easy to put in place, here's a solution:
`$('.custom-file-input').on('change', function() {
    let fileName = $(this).val().split('\\').pop();
    $(this).next('.custom-file-control').addClass("selected").html(fileName);
});`

istratos:
@jplaverdure this code won't work for multiple file input.

But I agree this should be a feature, it just feels weird to use this component and not have the file name auto update.

Superpencil:
Thanks @jplaverdure

PrinceOfAbyss:
This but still exists on v4.0.0 beta2

mdo:
Closing as duplicate of #20813.

shankyb123:
Thanks @jplaverdure 
$('.custom-file-input').on('change', function() { 
    let fileName = $(this).val().split('\\').pop(); 
    $(this).next('.custom-file-control').addClass("selected").html(fileName); 
});
This code is working on single file upload

[...]

PR #24138, opened 2017-09-27
Add js helpers + input change helper for custom file control

Johann-S:
Just an idea but that's something useful for folks, this file will help us to add some little helpers without creating a new plugin

Close: https://github.com/twbs/bootstrap/issues/23994

/CC @mdo, @bardiharborow, @XhmikosR

### DIFF
--- a/js/src/helper.js
+++ b/js/src/helper.js
@@ -1,0 +1,50 @@
+import $ from 'jquery'
+
+/**
+ * --------------------------------------------------------------------------
+ * Bootstrap (v4.0.0-beta): helper.js
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ * --------------------------------------------------------------------------
+ */
+
+const Helper = (() => {
+
+
+  /**
+   * ------------------------------------------------------------------------
+   * Constants
+   * ------------------------------------------------------------------------
+   */
+
+  const DATA_KEY  = 'bs.helper'
+  const EVENT_KEY = `.${DATA_KEY}`
+
+  const Selector = {
+    CUSTOMFILE : '.custom-file input[type="file"]'
+  }
+
+  const Event = {
+    INPUTCHANGE : `change${EVENT_KEY}`
+  }
+
+  const ClassName = {
+    CUSTOMFILECONTROL : 'custom-file-control'
+  }
+
+  function handleInputChange() {
+    const $fileControl = $(this).parent().find(`.${ClassName.CUSTOMFILECONTROL}`)
+    if ($fileControl.length === 1) {
+      $fileControl.text(this.value)
+    }
+  }
+
+  /**
+   * ------------------------------------------------------------------------
+   * Data Api implementation
+   * ------------------------------------------------------------------------
+   */
+
+  $(document).on(Event.INPUTCHANGE, Selector.CUSTOMFILE, handleInputChange)
+})(jQuery)
+
+export default Helper
--- a/js/src/helper.js
+++ b/js/src/helper.js
@@ -20,14 +20,17 @@ const Helper = (() => {
   const EVENT_KEY = `.${DATA_KEY}`
 
   const Selector = {
-    CUSTOMFILE : '.custom-file input[type="file"]'
+    CUSTOMFILE : '.custom-file input[type="file"]',
+    FORM: 'form'
   }
 
   const Event = {
-    INPUTCHANGE : `change${EVENT_KEY}`
+    INPUTCHANGE : `change${EVENT_KEY}`,
+    FORMRESET   : `reset${EVENT_KEY}`
   }
 
   const ClassName = {
+    CUSTOMFILE : 'custom-file',
     CUSTOMFILECONTROL : 'custom-file-control'
   }
 
@@ -38,6 +41,13 @@ const Helper = (() => {
     }
   }
 
+  function handleFormReset() {
+    const $filesControl = $(this).find(`.${ClassName.CUSTOMFILE}`).children(`.${ClassName.CUSTOMFILECONTROL}`)
+    if ($filesControl.length > 0) {
+      $filesControl.text('')
+    }
+  }
+
   /**
    * ------------------------------------------------------------------------
    * Data Api implementation
@@ -45,6 +55,7 @@ const Helper = (() => {
    */
 
   $(document).on(Event.INPUTCHANGE, Selector.CUSTOMFILE, handleInputChange)
-})(jQuery)
+  $(document).on(Event.FORMRESET, Selector.FORM, handleFormReset)
+})($)
 
 export default Helper
--- a/js/src/helper.js
+++ b/js/src/helper.js
@@ -25,8 +25,10 @@ const Helper = (() => {
   }
 
   const Event = {
-    INPUTCHANGE : `change${EVENT_KEY}`,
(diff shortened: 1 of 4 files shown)


js/src/helper.js
```diff
@@ -0,0 +1,50 @@
...
+const Helper = (() => {
+
+
+  /**
+   * ------------------------------------------------------------------------
+   * Constants
+   * ------------------------------------------------------------------------
+   */
+
+  const DATA_KEY            = 'bs.helper'
```
XhmikosR: Lots of spaces :)

XhmikosR:
How about adding it outside the input? What will happen if the path is too long? It will stretch the input with your current way.

Also, when I try the current patch, I'm getting C:\fakepath\filename.ext. Is that normal?

Johann-S:
yep that's the normal behavior for security reason :smiley: 

and yes they are possible issue with our CSS

lucascono:
I have another idea on this subject. I'm going to work on this today.

Johann-S:
before working on a another idea it would be better if you can explain your idea here to let you know if we're open to this idea or not @lucascono :+1:

lucascono:
I have tried this PR and I find some issues to solve:

1. What happens if the control is used inside a form and the form is reset? According to my tests with this PR, the value does not change.

2. What happens if the input is multiple? According to my tests with this PR, only the first file is shown. The normal behavior of the file controls is to show the amount of selected files (in the case of Firefox displays, for example: "2 files selected.").

3. To the previous point I add: assuming that, as does the native multiple control, we show the number of selected files, in what language would the suffix text be? Could the text of the suffix be translated as with the "placeholder" and the "button-label"?

4. What issues would the "helper.js" file handle? It seems to me that we have to consider the complexity of the forms and the way each browser treats them. If I'm not mistaken, it's the reason why custom-controls are implemented in Bootstrap: try to solve the inconsistencies.

These are just a few questions, but there are others. For these reasons, I consider some changes necessary, such as:

1. In my opinion, the forms deserve a separate plugin. Is there anything more inconsistent between different browsers? We think that, in the future, we also have to solve the color and rank controls, among others (there are already some issues open about it).

2. Consider multiple file controls. And as mentioned @XhmikosR, we must also consider the length of the control. And the best way, it seems to me, is how native control does: by number of files. And if so, it should be able to translate. Then you need to add some scss.

For these and other reasons, it seems to me that this matter must be dealt with in another way. And I would like to make a contribution that obviously I have in my mind but I must elaborate.

Otherwise, incorporate this into an upcoming release, I think we would then force us to reverse these changes.

Johann-S:
Not sure if we will go to a new plugin :thinking: to create a new plugin we should have differents use cases to handle, or currently we only have one and we won't add any validation.

I created this file just to add some helpers without creating new plugins and to open a discussion about that.

lucascono:
Ok. And about the other issues?

Johann-S:
I added a listener to handle form reset and I'll try to find a good way to display files when we have a multiple input file :+1:

lucascono:
I make a suggestion. What I had thought. Instead of adding the file as text inside the `custom-file-control`, make it as `[data-file]` and `[data-files]` attributes and use the `[attr]` selectors and `attr()` function of css in the `content` property of the `::after` pseudoelement. In this way, it could be translated using the variables. I explain?

lucascono:
And additionally would apply the mixin `text-truncate` to `::after` pseudoelement.

Johann-S:
Thanks it sems a good idea @lucascono :+1: 
before working on that I'll wait for @mdo feedbacks if he want to add that to Bootstrap or not

lucascono:
@Johann-S:

I share this code with you to try. I have tried it and I have loved it. It works just like the native file control.

In **`_variables.scss`**:

```scss
// stylelint-disable function-parentheses-newline-inside, function-comma-newline-after
$custom-file-text: () !default;
$custom-file-text: map-merge((
  placeholder: (
    en: "Choose file...",
    es: "Seleccione un archivo..."
  ),
  button-label: (
    en: "Browse",
    es: "Navegar"
  ),
  multiple-suffix: (
    en: "files selected",
    es: "archivos seleccionados"
  )
), $custom-file-text);
// stylelint-enable function-parentheses-newline-inside, function-comma-newline-after
```

In **`helpers.js`**:

```js
function handleInputChange() {
  const $fileControl = $(this).parent().find(`.${ClassName.CUSTOMFILECONTROL}`)
  if ($fileControl.length === 1) {
    $fileControl
      .attr('data-file', this.files.length === 1 ? this.files[0].name : '')
      .attr('data-files', this.files.length)
  }
}
```

And finally in **"_custom-forms.scss"** replace the rules with `:empty` pseudoclass for this:

```scss
&:not([data-files]),
&[data-files="0"] {
  &::after {
    content: map-get(map-get($custom-file-text, placeholder), en);
  }
}

&[data-files="1"] {
  &::after {
    content: attr(data-file);
  }
}

&[data-files]:not([data-files="0"]):not([data-files="1"]) {
  &::after {
    content: attr(data-files) " #{map-get(map-get($custom-file-text, multiple-suffix), en)}";
  }
}

@each $lang, $text in map-get($custom-file-text, placeholder) {
  &:lang(#{$lang}):not([data-files]),
  &:lang(#{$lang})[data-files="0"] {
    &::after {
      content: $text;
    }
  }
}

@each $lang, $text in map-get($custom-file-text, multiple-suffix) {
  &:lang(#{$lang})[data-files]:not([data-files="0"]):not([data-files="1"]) {
    &::after {
      content: attr(data-files) " #{$text}";
    }
  }
}
```

I understand that you wanted to wait for feedback from @mdo, but I just wanted to share it with you so that you try it sometime you have it free and tell me what you think. I hope you can try it.

It includes:

1. Code of PR #24259 that fixes a bug.
2. Variables with my language (es - spanish) so you can check how it works.

Greetings. :satisfied:

tmorehouse:
@Johann-S  Something you might want to consider adding in this helper is applying the class `.focus` to the `.custom-file-input`, when the hidden file input has focus (and remove the class when blurred).

**Background:**
Firefox does not work with the `input[type="file"]:focus ~ .custom-file-control` sibling selector (not sure if this is in the "wall of browser bugs" or not), as it considers the `browse` button inside the file input to have focus and not the actual `input[type="file"]. If the user clicks the filename area of the input (rather than the browse button) then the sibling selector works. See the answer for question https://stackoverflow.com/questions/20095105/input-type-file-focus-adjacent-selector-in-css (its an old question from 2015, but it is still applicable).

**Possible solution:**
The `<label>` wrapper does make it work for mouse users, but **does not work for keyboard users** (the inner browse button of the file input gets focus, not the actual file input when tabbing into the file input.

If the above focus/blur handler + `.focus` class tweak were used in the helper module, then all that would need  to be updated in `_custom-forms.scss` would be:

```css
.custom-file-input {
  min-width: $custom-file-width;
  max-width: 100%;
  height: $custom-file-height;
  margin: 0;
  opacity: 0;

  &.focus ~ .custom-file-control, /* add this line */
  &:focus ~ .custom-file-control {
    box-shadow: $custom-file-focus-box-shadow;
}
```

Johann-S:
Hello everybody, finally I made a plugin outside of Bootstrap for the custom input file you'll find it here: https://github.com/Johann-S/bs-custom-file-input

So I close this PR

PR #24143, opened 2017-09-27
Fix sidebar HTML entity encoding.

bardiharborow:
`Browsers & devices` needs the ampersand encoded as `&amp;` to pass validation.

### DIFF
--- a/_includes/docs-sidebar.html
+++ b/_includes/docs-sidebar.html
@@ -19,7 +19,7 @@
 
   <div class="bd-toc-item{% unless active == nil %} {{ active }}{% endunless %}">
       <a class="bd-toc-link" href="{{ site.baseurl }}/docs/{{ site.docs_version }}/{{ group_slug }}/{{ link_slug }}{% if link_slug %}/{% endif %}">
-        {{ group.title }}
+        {{ group.title | xml_escape }}
       </a>
 
       <ul class="nav bd-sidenav">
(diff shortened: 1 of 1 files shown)


XhmikosR:
I don't think that's true. We are using `utf-8` which makes the ampersand just fine.

Johann-S:
Agreed with @XhmikosR, which issue are you trying to fix ?

patrickhlauke:
Yup, surprisingly, a naked `&` is fine (and don't even think it's related to utf-8, just general HTML5 parsing) https://validator.w3.org/nu/?doc=https%3A%2F%2Fgetbootstrap.com%2Fdocs%2F4.0%2Fgetting-started%2Fintroduction%2F

XhmikosR:
My guess is @bardiharborow was trying to implement the vnu.jar integration and thought this was causing issues :)

Here are the current HTML errors (minus the missing `src` for `img`): https://gist.github.com/XhmikosR/9e417efb51b3b7b86c8502c8bb36b7a0

Most of those come from jekyll_toc and from the remaining ones, most of them must be by design.

bardiharborow:
It appears that my understanding came from [Section 5.3.2 in the HTML4 spec](https://www.w3.org/TR/html4/charset.html#h-5.3.2), where it says:

> Authors should use "&amp;" (ASCII decimal 38) instead of "&" to avoid confusion with the beginning of a character reference (entity reference open delimiter).

I cannot find an equivalent section in the HTML5 spec, and in fact I'm beginning to suspect that entity references have been generally depreciated. You learn something new every day I guess.

PR #24145, opened 2017-09-27
Add bundlesize.

### LINKED ISSUE #23009: Explore adding bundlesize
Could be rad to see something like https://github.com/siddharthkp/bundlesize in our CI status. It's easy for compiled code to get outta hand with small changes. Could help with that.

XhmikosR:
Fixes #23009.

So, the sizes bundlesize reports are the gzip ones. I went with pretty strict values and separate entries, but we can always change them if needed.

Now, in order for the GitHub integration to work, @mdo needs to take care of https://github.com/siddharthkp/bundlesize#2-build-status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: node_js
 git:
   depth: 3
 node_js:
-  - "4"
+  - "6"
   - "8"
 before_install:
   - if [[ `npm -v` != 5* ]]; then npm install -g npm@5; fi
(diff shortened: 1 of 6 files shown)


XhmikosR:
Oh, it seems bundlesize doesn't work on node.js 4 so we might want to bump the minimum version to 6 or run bundlesize only on node 8.

Johann-S:
Not against moving to node 6 plus node 4 maintenance will end in April 2018

XhmikosR:
Should I also add the CSS files?

Johann-S:
IMO @mdo would be interested to have our CSS file too

bardiharborow (approved):
Looks good. Happy to drop Node.js v4 given that it's in maintenance. v8 is about to hit LTS, so we'll be testing on both current LTS versions.

XhmikosR:
@mdo: will you add https://github.com/siddharthkp/bundlesize#2-build-status or should I merge now?

Johann-S (approved):

mdo:
Added env variable to Travis

PR #24146, opened 2017-09-27
Tweak npm scripts.

### LINKED ISSUE #24121: Refactor our npm scripts
OK, so currently, I find it quite annoying that

1. I need to open 2 separate cmds to do local development (one with `bundle exec jekyll watch` and another one with `npm run watch`)
2. Most importantly, I feel like our scripts don't include important parts.

When I do `npm run css` it makes sense to me to run the `docs-css` script too, and so on.

Opening the issue to track it and to find solutions to improve this.

/CC @Johann-S @bardiharborow

XhmikosR:
* make `css` and `js` "super" scripts which will run both main and docs scripts
* make `js-minify` run all js minify tasks
* use parallel in more places

It's a first pass but I think it covers more cases than before.

Suggestions to improve this are welcome!

Fixes #24121


Johann-S (approved):

XhmikosR:
So what should I do with this, guys?

@mdo @bardiharborow @Johann-S

bardiharborow (approved):
Seems sane to me, provided you've tested it.

PR #24147, opened 2017-09-27
Move uglify-js's config back to CLI.

### LINKED ISSUE #24139: assets/js/docs.min.js loses the comments
Not sure when this broke exactly, but after running `npm run js-docs`, assets/js/docs.min.js has the (important) comments removed.

I think we should try and fix this for proper attribution.

/CC @bardiharborow @Johann-S

XhmikosR:
We need to pass different options and this seems cleaner at this point.

In order for docs.min.js to keeps all comments, we don't pass `compress`.

This basically reverts #22990.

Fixes #24139


Johann-S:
I had an issue that's why I moved our uglifyjs config to a separate file :worried:

XhmikosR:
What issue exactly? I don't see anything wrong here...

Johann-S:
No but when I did the move to a config file it was because we had an issue but I tried now and it's seems fine :+1:

Johann-S (approved):

Johann-S:
can you keep `typeof: false` please ?

package.json
```diff
@@ -37,9 +37,9 @@
     "js-compile-bundle": "cross-env ROLLUP=true rollup --environment BUNDLE:true --config build/rollup.config.js",
     "js-compile-plugins": "cross-env PLUGINS=true babel js/src/ --out-dir js/dist/ --source-maps",
     "js-minify": "npm-run-all --parallel js-minify-standalone js-minify-bundle",
-    "js-minify-standalone": "uglifyjs --config-file build/uglifyjs.config.json --output dist/js/bootstrap.min.js dist/js/bootstrap.js",
-    "js-minify-bundle": "uglifyjs --config-file build/uglifyjs.config.json --output dist/js/bootstrap.bundle.min.js dist/js/bootstrap.bundle.js",
-    "js-minify-docs": "uglifyjs --config-file build/uglifyjs.config.json --output assets/js/docs.min.js assets/js/vendor/anchor.min.js assets/js/vendor/clipboard.min.js assets/js/vendor/holder.min.js assets/js/src/application.js assets/js/src/pwa.js",
+    "js-minify-standalone": "uglifyjs --compress --mangle --comments \"/^!/\" --output dist/js/bootstrap.min.js dist/js/bootstrap.js",
+    "js-minify-bundle": "uglifyjs --compress --mangle --comments \"/^!/\" --output dist/js/bootstrap.bundle.min.js dist/js/bootstrap.bundle.js",
```
Johann-S: should be `--comprend typeofs=false --mangle --comments \"/^!/\" `

Johann-S (changes requested):

package.json
XhmikosR: Will do in a few min.

Johann-S (approved):

PR #24149, opened 2017-09-27
Add back vnu-jar for HTML validation

### LINKED ISSUE #23171: Switch back to the official vnu.jar for HTML validation

XhmikosR:
After discussing this with @bardiharborow, I think this is the best middle solution; if Java is available HTML validation will run, otherwise it will be skipped.

Also, this is a lot faster than the html-validator solution and we don't introduce another point of external failure (like with Sauce Labs).

TODO:

* [x] Fix ignores
* [x] Fix multiple ignores
* [x] Combine ignores to an array instead of manually doing this?
* [x] Update jekyll-toc when they fix the HTML validation issues upstream
* [ ] Verify ignores

Fixes #23171


XhmikosR:
I believe this is ready. Perhaps we'd like to wait until the jekyll-toc errors are fixed upstream before merging this... or not.

@patrickhlauke: which of the other errors should be ignored?

Note that I'm ignoring warnings on purpose. We might want to revisit this later.

bardiharborow (approved):

bardiharborow:
@XhmikosR I think you should add `Element "ul" not allowed as child of element "ul" in this context.` as an ignore for now. Without further investigation I'm unsure about the other errors.

Johann-S (approved):

XhmikosR:
I think I'll wait a couple of days since the jekyll-toc issue seems to be WIP.

We still need @patrickhlauke's feedback about the remaining errors and which we should ignore.

XhmikosR:
Does anybody have any idea why ignoring specific errors doesn't work?

XhmikosR:
All right, after discussing this with @sideshowbarker on IRC, it appears the messages are inconsistent thus leading to this issue. `grunt-html` filters the errors differently and doesn't pass `filterpattern` thus the issue isn't present there.

I'll wait a bit for this to be addressed and we should be ready to merge in a few days, hopefully.

XhmikosR:
@bardiharborow @Johann-S @mdo: I changed this PR to not remove htmlhint and just add vnu-jar which will run if Java is available.

What do you think?

There's also the need to verify what we ignored in this branch. @patrickhlauke: can you have a quick look?

build/htmllint.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+'use strict'
+
+const childProcess = require('child_process')
+const vnu = require('vnu-jar')
+
+childProcess.exec('java -version', function (error) {
+  if (error) {
+    console.error('Skipping HTML lint test; Java is missing.')
```
Johann-S: If you return directly after that console.error you should throw an error instead

Johann-S (approved):
LGTM 👍 just a question, so we will use your script only if folks have Java ? Not on Travis ?

build/htmllint.js
XhmikosR: We don't want to fail here; if Java isn't installed we are just skipping vnu-jar.

bardiharborow:
@XhmikosR seems fine to me.

- `Attribute “autocomplete” is only allowed when the input type is “color”, “date”, “datetime-local”, “email”, “hidden”, “month”, “number”, “password”, “range”, “search”, “tel”, “text”, “time”, “url”, or “week”.` is likely intentionally non-standard.
- `Attribute “autocomplete” not allowed on element “button” at this point.` is likely intentionally non-standard.
- `Attribute “title” not allowed on element “circle” at this point.` I'm not sure about, likely an SVG issue (does the validator handle inline SVG correctly?).
-  `Bad value “tablist” for attribute “role” on element “nav”.` I'm not sure about, patrickhlauke will likely know.
- `Element “img” is missing required attribute “src”.` is a false positive, due to how we use Holder.js.
- `Element “legend” not allowed as child of element “div” in this context.*` I'm not sure about, I'd have to run tests locally to find the context.

I don't have an issue with merging now and then removing some ignores later.

bardiharborow (approved):

XhmikosR:
@Johann-S: Travis has Java so it will run there. See also this PR's Travis builds :)

XhmikosR:
I'm gonna merge this but I still think we need to go through our ignores.

For example, the error about title on circle seems legit to me and it's the issue I was telling @Johann-S the other day on Slack; this is non standard.

PR #24150, opened 2017-09-27
Darken outline hover color to match default button hover

### LINKED ISSUE #24144: Outline button hover should apply same darken as filled button.
If you have two buttons with the same theme color next to one another, one filled and one outline, the outline button's hover state is visually confusing. When hovered it fills and matches the already filled button exactly. When the outline button it hovered the color should fill and have the same darken applied as the standard button to prevent this.

https://codepen.io/afholderman/pen/rGyqbK

afholderman:
Alter button-outline-variant mixin to darken hover and active background in same fashion as filled button.

Fixes #24144


scss/mixins/_buttons.scss
```diff
@@ -42,16 +42,16 @@
   }
 }
 
-@mixin button-outline-variant($color, $color-hover: #fff) {
+@mixin button-outline-variant($color, $color-hover: #fff, $active-background: darken($color, 7.5%), $active-border: darken($color, 10%)) {
```
andresgalante: Since you are using the same parameters as in `button-variant` wouldn't it be better is we put outline variant under the same umbrella as normal variants? 
That way we keep them DRY and give less places to change to users.
What do you think?

andresgalante (approved):
Hi @afholderman I've pulled your branch, tested it, and it works as it should. Good job! 👍 

I am ok to merge this, but I'd keep in mind for the future to look for a better way to merge the `button-outline-variant` and `button-variant` mixins so we don't repeat the parameters. Set the colors once, once them in both places. specially the hover/active darken function.

afholderman:
@andresgalante I agree that these two mixins may need to be merged in to one. I just pushed another commit to fix another issue with the outline buttons, the lack of color-yiq usage. This was causing btn-light-outline to be illegible when hovered, and not match btn-light

Issue #24159, opened 2017-09-28
Proposal - New Bootstrap Pagination Solution


wilkmit:
I came across this pagination solution and thought it would work perfectly with bootstrap responsive design. When pagination falls to a second or third row I find it to be a waste of space and it also looks ugly. Anyway this is the most elegant solution I have seen so far.

![pagination](https://user-images.githubusercontent.com/6516828/30970011-103f8994-a432-11e7-94b8-e153fcb1ab58.PNG)

andresgalante:
Hi @wilkmit that's a good option. 
I have another idea, easier show than tell, so I'll send a PR 😄

andresgalante:
So, I don't know if this is a good idea, but it was fun to make. I stole it from an [article](https://medium.com/@srambach/why-do-i-have-to-learn-geometry-35189e5babab) @srambach  wrote.

By keeping the exact same flex structure and changing the flex values we can achieve this:

![example](http://g.recordit.co/ZQ1kV1qjeQ.gif)

What do you think? worth it or useless?

wilkmit:
I mean it looks cool and all but eventually the problem would still be a problem wouldn't it? The only reason I like the dropdown is because your pagination would be clean and consistent no matter the amount of pages. I have also seen other solution such as the one below but I am still stuck on the dropdrown when I first saw it I was like ahhh so simple why is everyone not doing pagination this way.

![pagination-using-ellipsis](https://user-images.githubusercontent.com/6516828/30983734-44198a04-a459-11e7-9788-c8f71643bda9.png)

gijsbotje:
was wondering if we could do this with all BS classes, just 1 custom css property
https://codepen.io/gijsbotje/pen/xXrKxd

andresgalante:
Are we looking at a new alternative for the pagination component or to make the existing component responsive?

Either way I think we should think about this issue, and find a solutions. Pagination as it stands is a poor option for small view ports of touch devices.

wilkmit:
I don't see any reason not to use as much of bootstrap as possible (ie. buttons, drop downs).

I think a component replacement would be better because it would allow the problem to be solved right but I am not sure what's possible I would need to research the current component.

wilkmit:
If I designed it I would do something like this and then use js to resize the select based on the option selected.

![capture](https://user-images.githubusercontent.com/6516828/31018178-15b07b46-a4f9-11e7-8357-16f4bd070c9a.PNG)

[...]

PR #24164, opened 2017-09-28
Adds a flex wrap to pagination

andresgalante:
Even though this PR doesn't solve #24159 it's related to it.

Pagination doesn't wrap, so when the user has lots of pages it generates a horizontal scroll bar:

screen shot 2017-09-28 at 2 30 13 pm

This PR introduces a wrap, it's not a perfect solution for responsive paginations but at least it doesn't break the layout:

screen shot 2017-09-28 at 2 30 57 pm

What do you think?

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -1,5 +1,6 @@
 .pagination {
   display: flex;
+  flex-wrap: wrap;
   // 1-2: Disable browser default list styles
   padding-left: 0; // 1
   list-style: none; // 2


mdo:
This was proposed somewhere else too and I passed on it. Ideally the developer should implement a better solution to the pagination than showing this many links :). An easier option might be adding `.d-none.d-md-block` or something to each of those inner links, leaving just `previous` and `next`.

PR #24168, opened 2017-09-29
iPhone X Fix

Dygear:
Adds `viewport-fit=cover` to the meta viewport tag allowing for page headers and footers to span the entire screen with the iPhone X.


Dygear:
Ok, so that wasn't the cleaest way to do that. I didn't read the `Add more commits by pushing to the patch-1 branch` Until after the fact. But that commit cleans up the rest of the places where the meta viewport is used within adding `viewport-fit=cover` for the iPhone X.

andresgalante:
@Dygear Don't worry if it gets merge it'll be squashed.

Let me add this article about "the notch" for more clarification on what this PR is doing:
https://css-tricks.com/the-notch-and-css/

patrickhlauke (commented):
Would be good to test all of this to ensure that there's no content now falling into the safe area that would require additional padding or similar before merging.

Dygear:
I’m actually using it. But I can provide example screen shots if you wish based on these code changes here to prove it doesn’t require additional padding. (On the iPhone Simulator for the iPhone X.)

mdo:
Before and after screenshots here. Going to need a bit more than the viewport change I think.

screen shot 2017-10-02 at 5 41 19 pm
screen shot 2017-10-02 at 5 41 07 pm

XhmikosR:
I'm personally not sure about this change. It's not like things work fine out of the box by just adding `viewport-fit=cover`.

Dygear:
I’m going to have to add some padding to make this work correctly on the iPhone X without the content div wrapper.

mdo:
Closing this out for now given it doesn't cleanly solve the problem. Looking forward to other ideas though! I can test on mine in 2-3 weeks. 😆

ThomasBerends:
@mdo The addition of a padding fix could solve the problem:

https://booking.design/how-we-adapted-the-booking-com-mobile-site-for-the-iphone-x-notch-e880fa00e989

PR #24033, opened 2017-09-21
Making the customization option docs clearer about variable overrides

Bobeta:
Current docs on the subject can easily be interpreted as the user should import their modified variables **after** importing bootstrap sass files, this PR aims to make it clearer #23963

PR #24186, opened 2017-09-30
Updates to customizing variables docs

mdo:
I believe #24033 ~~is factually inaccurate~~, while also containing some typos and grammar problems. This PR addresses a few things:

- ~~Variable overrides can come before or after you import Bootstrap—they'll still override the default value in our code.~~
- Variable overrides within the same Sass file can come before or after the default value, but across files they must come first.
- Removed the double `::` at the end of the paragraph.
- Fixed usage of `overwrite` when we mean `override`.

/cc @johann-s @xhmikosr @bobeta


mdo:
Ah, yeah, we're both right. I only tested variable overrides within the same Sass file, not across say a custom variables file and an import of Bootstrap. Testing it out in both cases again and will make edits.

XhmikosR (approved):

PR #24187, opened 2017-09-30
Update some grid docs

### LINKED ISSUE #23754: Columns sound no good
[The documentation's first section on the Grid system](https://getbootstrap.com/docs/4.0/layout/grid/#how-it-works) asks "Sounds good?". It is hard to answer, but it sounds vague.

The whole section brings more questions than it answers, but one particularly problematic part is the item 
> Content should be placed within columns, and only columns may be immediate children of rows.

At this point, the meaning of "column" has not been defined, and I do not see it defined later neither.

### LINKED ISSUE #24076: Inappropriate paragraph in grid system documentation's "How it works" section ("Sounds good?")
[The documentation's first section on the Grid system](https://getbootstrap.com/docs/4.0/layout/grid/#how-it-works) asks "Sounds good?".

As pointed out in issue #23754 , "sounds not so good", so assuming a positive answer with "Great, let’s move on to seeing all that in an example." is pretentious at this time. In any case, it would be preferable to avoid unnecessarily questioning the reader. Since the paragraph is not immediately followed by examples, I would suggest to just remove it, or to fix and move down.

mdo:
After a couple reports (fixes #23754 and closes #24076), I'm clarifying some things in the `How it works` section of the docs. I'm doing my best to be very specific and explain grid concepts and terms where we use them. However, not everything can be explained here—working with Bootstrap does require some experience with HTML, CSS, and building responsive web projects.

#24076 is only getting closed here because it's factually inaccurate. I don't disagree with the tone of it or think it's inappropriate. There's language like this all over the docs.


bardiharborow (approved):
Looks good, although as a side note I think a little flair in our documentation and blog posts has always been a part of the Bootstrap brand, and it's a key part of the community building attitude that we try to foster.

mdo:
Totally agreed, @bardiharborow! <3

Issue #23400, opened 2017-08-14
Add an order class to be before unspecified columns (like flex-first did previously)


chrisblakley:
Like I mentioned in another issue, I love how intuitive the new `order` classes are! However, when playing with them tonight I found that there is no way to move a `div` unless other `div`s in the row are also specified.

Since the default flex order is `0`, the Bootstrap order classes can only move `div`s to the right of an unspecified `div` (such as `.col` or `.col-md`).

```html
<div class="row">
	<div class="col-md">
		Unspecified column width. This will be first no matter what (default flex order of 0).
	</div>
	<div class="col-md order-1">
		Order declared, but remains right of unspecified column(s).
	</div>
</div>
```

### Previous solution
In the alphas, we had access to the class `flex-first` which would force the `div` to be the first column regardless (by setting the order to `-1`). I think the idea of this can be included while still maintaining the simplicity of the `order` classes.

### Proposed solutions
 - Having a class with the property of `order: -1` would be one idea. Maybe `order-first`, `order-before`, `order-0`, or even the former `flex-first`, or something completely different (just spitballing).
   - I'd hate to slip down the slope of requesting 12 more order classes for negative numbers, but maybe we could introduce one class to force a column first?
 - The default order could be changed as well, but this would cause the opposite problem.
 - Another solution could be to ignore it and just require devs to implement their own class if they want to reorder a column to be before an unspecified Bootstrap column `div`.

I want to make sure Bootstrap retains as much functionality as possible to allow these "width agnostic" columns because it is such a powerful feature that allows for so much elegant flexibility when developing.

lukefrake:
👍  this, would be great to have 
```
.order#{$infix}-first{
   order: -1
}
```.

On the same point, in the beta we hard

```
.order#{$infix}-unordered {
   order: 0;
}
```
This is really useful as well and would be great to have this back in.

I much prefer having the `.order-n` classes, they're powerful but think first and unordered need to be added back in.

If this is acceptable I'll put a PR together to put this back in and update the docs

Macintosher:
-- 



Bernhard Knab
Sebastian-Münster-Straße 3a
67549 Worms
Telefon: +49(0)6241-970020
Telefax: +49(0)6241-970749
bk@bknab.de

Macintosher:
HELP Where can I stop getting 1000s of emails

-- 



Bernhard Knab
Sebastian-Münster-Straße 3a
67549 Worms
Telefon: +49(0)6241-970020
[...]

PR #24202, opened 2017-10-02
Add order first to grid

lukefrake:
Fix : #23400

### DIFF
--- a/docs/4.0/layout/grid.md
+++ b/docs/4.0/layout/grid.md
@@ -530,9 +530,6 @@ Use `.order-` classes for controlling the **visual order** of your content. Thes
     <div class="col order-12">
       Third, but last
     </div>
-    <div class="col order-first">
-      Fourth, but first
-    </div>
   </div>
 </div>
 {% endexample %}
--- a/docs/4.0/layout/grid.md
+++ b/docs/4.0/layout/grid.md
@@ -524,11 +524,11 @@ Use `.order-` classes for controlling the **visual order** of your content. Thes
     <div class="col">
       First, but unordered
     </div>
-    <div class="col order-1">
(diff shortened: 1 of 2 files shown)


docs/4.0/layout/grid.md
```diff
@@ -515,7 +515,7 @@ In addition to column clearing at responsive breakpoints, you may need to reset
 
 ### Order classes
 
-Use `.order-` classes for controlling the **visual order** of your content. These classes are responsive, so you can set the `order` by breakpoint (e.g., `.order-1.order-md-2`). Includes support for `1` through `12` across all five grid tiers.
+Use `.order-` classes for controlling the **visual order** of your content. These classes are responsive, so you can set the `order` by breakpoint (e.g., `.order-1.order-md-2`). Includes support for `1` through `12` across all five grid tiers. You also set the class of an element to be `first` which will give it negative order and so put it before unordered elements.
```
XhmikosR: You **can** also set the class of an element to be first with `.order-first`, which will give it a negative order and so place it before the unordered elements.

mdo:
Updated the docs here to fix @XhmikosR's feedback by revamping them. We're introducing a new class name, which is best served by a new code example rather than mixing the two.

mdo (approved):

lukefrake:
@mdo that is much clearer, I considered doing that but didn't want to bloat the docs. LGTM though

Do you want me to revert the order-first in the code element above or are you happy to keep that?

XhmikosR (approved):

mdo:
@lukefrake Just took care of that, good catch.

lukefrake:
@mdo no problem, happy to help!

PR #24210, opened 2017-10-02
Normalize examples' CSS.

XhmikosR:
Ideally, I think we should use autoprefixer and also lint these to automate the process...


mdo:
The examples have been a weird one to be honest. I've tried to keep them as isolated as possible, ensuring they run if you copy-paste, but that's kinda difficult. I'd prefer not to add a bunch of tooling to them though, but perhaps there's another approach I haven't considered.

mdo (approved):

XhmikosR:
We already use autoprefixer so it shouldn't be too hard to hook something up.

I'll make an issue to track this so that I don't forget.

PR #24217, opened 2017-10-03
Switch to lowercase `doctype`.

XhmikosR:
Having it uppercase doesn't offer anything and the lowercase version slightly improves GZIP compression.

This was proposed a long time ago (can't find the issue/PR now) but we rejected it back then. Now I find no real gain in having an uppercase `doctype` thus this PR.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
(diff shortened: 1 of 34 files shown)


mdo (approved):

patrickhlauke (approved):
I just double-checked, and yes it's valid (the doctype is case insensitive - see https://www.w3.org/TR/html51/syntax.html#the-doctype)

PR #24221, opened 2017-10-03
Hound: Disable scss.

XhmikosR:
Refs #24225

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -3,6 +3,9 @@ fail_on_violations: true
 stylelint:
   config_file: build/.stylelintrc
 
+scss:
+  enabled: false
+
 jshint:
   enabled: false
 


Johann-S (approved):

mdo:
How trivial is it to setup stylelint to report back via CI instead of code review like Hound?

bardiharborow:
Are you able to clarify what you're trying to achieve here? You're spawning [three separate Node 8 containers](https://travis-ci.org/twbs/bootstrap/builds/282794243), two of which seem to do the same thing.

bardiharborow:
@mdo we're already doing that. Hound as an additional layer of linting is just being mad, presumably for the reason mentioned in https://github.com/twbs/bootstrap/pull/23572#issuecomment-333347685, and XhmikosR additionally says in #24225 that there may be some bugs in Hound's stylelint implementation.

XhmikosR:
From what I understand, Hound is using their own configs which is why this doesn't work properly.

@bardiharborow: this PR is just for testing crap.

PR #24208, opened 2017-10-02
Add support for calculate and dynamically change offset position for …

romanlex:
Sorry for bad English.
Add support for calculate and dynamically change offset position for dropdown menu.
Function call always when call Popper.js modifierFn.
Now I can change dynamically position of dropdown menu 
```javascript
            $('[data-toggle="dropdown"]').dropdown({
                offsetFn: function(data) {
                    data.left = -100;
                    return data
                }
            });
```

PR #24222, opened 2017-10-03
Offset option for dropdown can be function

romanlex:
Fix https://github.com/twbs/bootstrap/pull/24208

Offset option can be function (Popper.js)

> Add support for calculate and dynamically change offset position for dropdown menu.
Function call always when call Popper.js modifierFn.

With this code I can change any dropdown position as on picture
```javascript
const dropdownMegaToggle = $('.dropdown-toggle--mega');
dropdownMegaToggle.dropdown({
    offset: function(data) {
        let popper = data.popper;
        let reference = data.reference;
        let w = $(window).width();
        let buttonOffset = dropdownMegaToggle.offset();
        popper.top = 100;
        reference.width = w / 2; // f*cking Popper.js crutch
        popper.left = w / 2 - popper.width / 2 - buttonOffset.left;
        return data
    }
});
```

Fix : https://github.com/twbs/bootstrap/issues/24223


js/src/dropdown.js
```diff
@@ -39,6 +39,8 @@ const Dropdown = (() => {
   const ARROW_DOWN_KEYCODE       = 40 // KeyboardEvent.which value for down arrow key
   const RIGHT_MOUSE_BUTTON_WHICH = 3 // MouseEvent.which value for the right button (assuming a right-handed mouse)
   const REGEXP_KEYDOWN           = new RegExp(`${ARROW_UP_KEYCODE}|${ARROW_DOWN_KEYCODE}|${ESCAPE_KEYCODE}`)
+  const POPPER_OFFSET_KEY        = 'offset'
+  const POPPER_OFFSET_FN_KEY     = 'fn'
```
Johann-S: Not a big fan to add new const for that :thinking:

js/src/dropdown.js
```diff
@@ -241,16 +243,28 @@ const Dropdown = (() => {
...
+
     _detectNavbar() {
       return $(this._element).closest('.navbar').length > 0
     }
 
     _getPopperConfig() {
+      let key = POPPER_OFFSET_KEY
+      if (typeof this._config.offset === 'function') {
+        key = POPPER_OFFSET_FN_KEY
+      }
```
Johann-S: IMO something like that would be enough : 
```javascript
const offsetConf = {}
if (typeof this._config.offset === 'function') {
  offsetConf.offsetFn = (data) => this._getOffset(data)
} else {
  offsetConf.offset = this._config.offset
}

**Review by Johann-S — CHANGES_REQUESTED** [review]

File: js/src/dropdown.js [hunk 3]
```diff
@@ -241,16 +243,28 @@ const Dropdown = (() => {
       return placement
     }
 
+    _getOffset(data) {
+      const popperOffset = $.extend({}, data.offsets, this._config.offset(data.offsets) || {})
+      Object.keys(popperOffset).forEach((key) => {
```
Johann-S: Sorry to ask I don't know well `offset` function from Popper.js but why do you need that ?

romanlex: data object has many another keys. We need to change only data.offsets and not other. If we return only extended object with data.offsets - Popper is broken

Johann-S: That's right 👍 but instead of a Foreach you just have to do : 
`data.offsets = popperOffset`

js/src/dropdown.js
romanlex: I agree with your. It's better solution. Const is redundant
```javascript
_getPopperConfig() {
      const offsetConf = {}
      if (typeof this._config.offset === 'function') {
        offsetConf.fn = (data) => this._getOffset(data)
      } else {
        offsetConf.offset = this._config.offset
      }
      const popperConfig = {
        placement : this._getPlacement(),
        modifiers : {
          offset : offsetConf,
          flip : {
            enabled : this._config.flip
          }
        }
      }

      // Disable Popper.js for Dropdown in Navbar
      if (this._inNavbar) {
        popperConfig.modifiers.applyStyle = {
          enabled: !this._inNavbar
        }
      }
      return popperConfig
    }
```

js/src/dropdown.js
romanlex: Yup. Agree :) Sorry don't know how I can stick new changes in this PR https://github.com/romanlex/bootstrap/commit/9be393b3a567e327d07c7513e570b4fb1a3d2d41

js/src/dropdown.js
romanlex: Agree, const is redundant. Updated

Johann-S (approved):
Thank you very much @romanlex to tackled that :+1:

js/src/dropdown.js
```diff
@@ -241,17 +241,26 @@ const Dropdown = (() => {
...
+
     _detectNavbar() {
       return $(this._element).closest('.navbar').length > 0
     }
 
     _getPopperConfig() {
+      const offsetConf = {}
+      if (typeof this._config.offset === 'function') {
+        offsetConf.fn = (data) => this._getOffset(data)
+      } else {
```
Johann-S: I was thinking maybe we should remove `_getOffset` method because that's not something we will use outside of this and remplaced it by : 

```javascript
offsetConf.fn = (data) => {
     data.offsets = $.extend({}, data.offsets, this._config.offset(data.offsets) || {})
     return data
}

Can you give it a try ?

Johann-S (changes requested):

js/src/dropdown.js
romanlex: Remove this method but in my opinion this is less readable. On the other hand, the method is not needed, because obviously not planned to manage offseting inside boostrap

Johann-S (approved):

PR #24229, opened 2017-10-03
make-col-offset: Remove percentage if it's zero.

XhmikosR:
Another case is `flex: 1 0 0%;`. Is this on purpose? In theory using `0` should work just fine.

Or maybe `flex: 1 0`?


XhmikosR:
Another case is `flex: 1 0 0%;`. Is this on purpose? In theory using `0` should work just fine.

mdo:
Flexbugs #4: https://github.com/philipwalton/flexbugs#4-flex-shorthand-declarations-with-unitless-flex-basis-values-are-ignored

We should add a note to the source code for it.

XhmikosR:
I will add the note tomorrow, thanks.

XhmikosR:
@mdo: done

mdo:
Now that I think about it, wasn't this `%` stripping something CSS Comb or a linter was doing for us?

XhmikosR:
We just used CSSLint back then in v3. In stylelint I already have it enabled but apparently it doesn't warn for `0%`. `0px` works as expected. We'll need to report it.

I have a branch where I attempt to tighten things a little more but we'll probably need to work on it together.

XhmikosR:
I reported the issue here: https://github.com/stylelint/stylelint/issues/2940

XhmikosR:
@mdo: I don't think this will be supported so we can just merge this (just approve).

This didn't work with scss-lint either BTW. I'll see if I can come up with something else later.

mdo (approved):

PR #24234, opened 2017-10-04
Enable stylelint for Hound

### LINKED ISSUE #24225: Fix Hound Sass checking
- [x] We probably need to disable `scss` since we are using stylelint now
- [ ] Hound doesn't work with our stylelint config, see #24234

~I already contacted the Hound people; we might be doing something wrong or their stylelint implementation doesn't work for Scss files. If it's the later we'll need to move away from Hound (at least) for scss files.~ Already did this.

XhmikosR:
Fixes #24225

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -2,16 +2,17 @@ fail_on_violations: true
 
 stylelint:
   config_file: build/.stylelintrc
+  enabled: true
+
+eslint:
+  config_file: js/.eslintrc.json
+  ignore_file: .houndignore
 
 scss:
   enabled: false
 
 jshint:
   enabled: false
-
-eslint:
-  config_file: js/.eslintrc.json
-  ignore_file: .houndignore
 
 ruby:
   enabled: false


XhmikosR:
I'm not sure if this is even needed.

/CC @gylaz

XhmikosR:
@gylaz: see here the error that Hound is throwing? I was right to assume that Hound's stylelint config doesn't cover Scss checking.

gylaz:
@XhmikosR I opened https://github.com/houndci/linters/pull/157 to add `stylelint-config-recommended-scss` plugin. Is there any other plugin you anticipate will be needed?

XhmikosR:
@gylaz: thanks! I think we need more packages for scss linting to work, at least in our case here.

https://github.com/twbs/bootstrap/blob/347b3ac1c01ee9996daf542fb816a7dc34923d8a/package.json#L108-L112

XhmikosR:
Scratch that, spoke too soon. I think your PR adds the only missing package. :)

XhmikosR:
Thanks @gylaz!

gylaz:
@XhmikosR I just deployed the new plugin and re-ran checks for this PR. Looks like there's a config error?

XhmikosR:
I think it's the other issue I mentioned yesterday. Your stylelint packages
are old. Can you update all of them ?

gylaz:
Opened this PR: https://github.com/houndci/linters/pull/159. Will report back once it's merged and deployed.

XhmikosR:
@gylaz: isit normal we don't get any errors here even with all the plugins installed on Hound? Is it because the base branch v4-dev doesn't have Hound enabled?

Johann-S (approved):

XhmikosR:
It worked!

PR #24238, opened 2017-10-04
Suggest CodePen along with JSBin.

XhmikosR:
I'm using both but I personally find CodePen better. That being said, I think this is a fair addition.

PS. We could move the old jsbins to codepen if we want, but I'd rather have @mdo do that under his CodePen account.


Johann-S (approved):
I prefer CodePen too :+1:

mdo (approved):
I still prefer JS Bin over CodePen for it's simpler interface and sharing, however, JS Bin has had a rough time staying ahead of perf and spam lately it seems. Let's leave both in there for now, please.

PR #24240, opened 2017-10-04
Fix modal bug

### LINKED ISSUE #24117: Opening a modal window twice makes it impossible to close it again.
Opening a modal window twice makes it impossible to close it again.

So if you open the modal window #modal and then execute $("#modal").modal("show"); in the browser console, it is no longer possible to close the modal window.

I'm using Bootstrap 4.0.0-beta on Chrome 60.0.3112.113.

ghost:
Fixes #24117.

Note that the `hide()` function checks, before executing the routine, that the modal is not hidden. This is logical: why should I try to hide something that is already hidden?

```js
hide(event) {
  if (this._isTransitioning || !this._isShown) {
    return
  }
}
```

However, the `show()` function does not do this same check and runs the entire routine even though the modal is visible.

```js
show(event) {
  if (this._isTransitioning) {
    return
  }
}
```

That is why I check, before running the whole routine, that the modal is not visible. And I have verified that this indeed solves the bug.

PS: Sorry for the English. I'm using Google Translator.


Johann-S:
Please can you add a unit test ? 

And about your change, LGTM :+1:

lucascono:
@Johann-S I'm learning to use git and I still do not quite understand it. I published the unit test? Can you confirm that? Greetings from Argentina and thank you.

XhmikosR:
All is good @lucascono.

@Johann-S: please merge if it looks good to you too.

Johann-S (approved):
Perfect @lucascono thank you :+1:

XhmikosR:
@Johann-S: you should have squashed the useless commits.

Johann-S:
Oops... you're right... my bad :worried: 
Is it possible to fix it ?

XhmikosR:
Nope since the branch is protected we can't force push. Only the repo owners can.

XhmikosR:
But this should happen fast so that no other people get the change and make PRs otherwise it beats the point. So by the time @mdo will be up it'll be too late :P

Johann-S:
:worried: most of the time I think about squashing commits...

PR #24241, opened 2017-10-04
npm scripts: quote globs.

XhmikosR:
OK this was very annoying and was driving me crazy.

I had noticed for a long time that htmlhint on Travis said, 13 HTML files scanned without any errors. That was clearly wrong since we have a lot more HTML files.

Then I noticed that after the stylelint switch, Travis builds would sometimes be marked as passed when locally I was getting errors.

So, for anyone reading this, **always quote your globs in npm scripts**.

PS. this is the reason htmlhint was basically not doing anything.


Johann-S (approved):
Nice catch :+1:

PR #24249, opened 2017-10-04
Add CSS Utilities for 'auto' width and height Properties

taufek:
Currently bootstrap only provides 4 variations of `width` and `height` css utilities, which are:

1. 25%
2. 50%
3. 75%
4. 100%.

Here, I'm adding `auto` to the list. This is useful when we need to reset the width/height value
back to the browser default value at some point.


andresgalante:
Hi @taufek. I am having a hard time finding this useful. Can you please help me with a use case example when you would need to reset the `width` or `height` of an element to `auto`?

taufek:
Hi @andresgalante ,

Below is one of the example. There are 2 `select` element.

The first one has `form-control` class which gives `width: 100% !important` to the element.

The second one has `form-control` and `w-auto` which overrides the  `width` of `select` element to expand the width according to its content.

demo

https://jsbin.com/xecutulusi/edit?html,css,output

taufek:
Below is another example where I want to avoid the `select` being truncated when it is inside a narrow cell.

demo2

https://jsbin.com/sagigawofe/edit?html,css,output

andresgalante (approved):
Unlike #24015, adding `auto` to the size map has a small impact since it's only used on sizing utilities, and I think the usecases @taufek provide are valid.

@mdo up to you

taufek:
@mdo, 

What is your thought on this change?

mdo (approved):

PR #24254, opened 2017-10-04
Tighten stylelint rules

XhmikosR:
@mdo: feel free to push here. Remember the rules that we can enable are the ones set to `null` in .stylelinrc.


mdo (approved):
Looks good to me so far.

Issue #24251, opened 2017-10-04
Big dropdowns are unusable when used inside an `overflow:scroll` container, due to being constrained inside it


ronjouch:
## Problem

When using a [Bootstrap (4.0.0beta) Dropdown](https://getbootstrap.com/docs/4.0/components/dropdowns/) in an element that is a child of an `overflow:scroll` element, the dropdown is "stuck" inside the container, making it unusable if the dropdown contents are too big horizontally/vertically:

![overflow-scroll_bootstrap-dropdown-fs8](https://user-images.githubusercontent.com/522085/31194832-54d24bda-a916-11e7-9ad3-85541dd35f62.png)

Contrarily, a vanilla HTML `<select>` will merrily bleed out of the `overflow:scroll` element as necessary:
![overflow-scroll_regular-select-fs8](https://user-images.githubusercontent.com/522085/31194823-50451b9c-a916-11e7-85be-41e113fc0895.png)

See demo at [JSFiddle/ronj/t6z6Lnfb](https://jsfiddle.net/ronj/t6z6Lnfb/).

## Use case

A few words on the use case justifying this combination, as shown in the JSFiddle which mimics a webapp I'm working on:

- In the *left sidebar* live a set of filters. It may contain several filters that will consume undetermined vertical space, thus lives in an `overflow-y:scroll` div. Each filter has a dropdown which lets users modify filter options.
- In the *right pane* lives the content, filtered by the filters in the left sidebar.

→ As a user,

1. I *do need* that `overflow-y:scroll` sidebar scrolling behavior (because the list of filters is of undetermined and potentially large height, depending on the number of filters)
2. But filter dropdowns (which can be large, both in width due to length of options and height due to count of options) should be able "escape out of" / not be constrained by the sidebar.

And as shown in the demo/screenshots, vanilla HTML `<select>` don't suffer from this problem, but it's not always practical/possible to revert back to using them, e.g. due to using features specific to Bootstrap Dropdown, or for consistent styling.

## Details

Bootstrap version: 4.0.0-beta

OS/Browser: Ubuntu 16.04.3 / {Chrome 63, Firefox 58}

[...]

PR #24257, opened 2017-10-05
Add container option to our Dropdowns

Johann-S:
Add container option to our Dropdowns

For those who wants to try : https://codepen.io/Johann-S/pen/EwoPYo

Fixes : #24251

### DIFF
--- a/docs/4.0/components/dropdowns.md
+++ b/docs/4.0/components/dropdowns.md
@@ -728,7 +728,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>container</td>
       <td>boolean | string | element</td>
       <td>false</td>
-      <td>Appends the dropdown to a specific element. Example: <code>container: 'body'</code>.</td>
+      <td>Appends the dropdown to a specific element. Example: <code>container: 'body'</code>. Allowed only for dropdowns not in a navbar.</td>
     </tr>
   </tbody>
 </table>
--- a/docs/4.0/components/dropdowns.md
+++ b/docs/4.0/components/dropdowns.md
@@ -724,6 +724,12 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>true</td>
       <td>Allow Dropdown to flip in case of an overlapping on the reference element. For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#modifiers..flip.enabled">flip docs</a>.</td>
     </tr>
+    <tr>
+      <td>container</td>
+      <td>boolean | string | element</td>
+      <td>false</td>
+      <td>Appends the dropdown to a specific element. Example: <code>container: 'body'</code>.</td>
+    </tr>
   </tbody>
 </table>
 
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -54,11 +54,13 @@ const Dropdown = (($) => {
   }
 
   const Selector = {
-    DATA_TOGGLE   : '[data-toggle="dropdown"]',
-    FORM_CHILD    : '.dropdown form',
-    MENU          : '.dropdown-menu',
-    NAVBAR_NAV    : '.navbar-nav',
-    VISIBLE_ITEMS : '.dropdown-menu .dropdown-item:not(.disabled)'
+    DATA_TOGGLE    : '[data-toggle="dropdown"]',
+    FORM_CHILD     : '.dropdown form',
+    MENU           : '.dropdown-menu',
+    NAVBAR_NAV     : '.navbar-nav',
+    VISIBLE_ITEMS  : `.dropdown-menu .dropdown-item:not(.${ClassName.DISABLED})`,
+    ACTIVE_ITEMS   : `.dropdown-item:not(.${ClassName.DISABLED})`,
+    DROPDOWN_SHOWN : `.${ClassName.SHOW} > [data-toggle="dropdown"]`
   }
 
   const AttachmentMap = {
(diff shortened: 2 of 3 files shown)


danijelGombac:
I was try this changes and it works fine, only the position of the dropdown is wrong on the first click. When you click again the position is ok.

danijelGombac:
First click: translate3d(1625px, 406px, 0px)
Second click: translate3d(1642px, 406px, 0px)

Johann-S:
Seems related to Popper.js :worried: 
/CC @FezVrasta

Johann-S:
I'm able to reproduce what's described by @danijelGombac on this CodePen : https://codepen.io/Johann-S/pen/WZdrKj

tmorehouse:
One caveat... if someone specifies a container when the dropdown is inside a collapsible navbar.  Would need to have a guard in there to prevent the dropdown from being re-parented in that case.

Johann-S:
Good catch @tmorehouse ! :+1: 
I'll enable that feature only for dropdowns not in a navbar

danijelGombac:
If you use this option with dropdown inside modal, the dorpdown position is wrong when scrolling modal.

Johann-S:
You can call `update` of our Dropdown plugin it will update the position of your Dropdown 👍

danijelGombac:
This happen when dropdown is inside overflow. Try open modal and dropdown.

https://codepen.io/danijelGombac/pen/OxZxZM

Johann-S:
Not sure it's related to Bootstrap here @danijelGombac because it's Popper.js which handle the position

danijelGombac:
Yes the problem is in Popper.js. It can't handle two overflow container if the first container is not position static.

https://codepen.io/danijelGombac/pen/OxZxZM

Johann-S:
maybe @FezVrasta will have time to take look at this issue :smile:

danijelGombac:
I hope. There could be also problems with tooltips and popover inside modal and table-responsive.

Johann-S:
I made an issue to track that : https://github.com/FezVrasta/popper.js/issues/459

Johann-S:
I found a solution to fix that @danijelGombac :+1:

danijelGombac:
@Johann-S 

Why not apply this config option only when config.parent was used?

Johann-S:
yep you're right seems better with that way !

Johann-S:
done @danijelGombac :+1:

tmorehouse:
Just thinking ARIA and tab order.

Will re-parenting to the body change the tab order of the menu items, since the menu items are no longer siblings of the menu trigger button?

patrickhlauke:
indeed. unless focus is moved into the dropdown when it's opened and then kept there, this would be quite a broken experience for kbd users. (in the example in codepen, simply add some other focusable elements to the page to see the problem). so (just going by the example posted in the starter, not tried the actual code in this PR) i'd say no to this...

tmorehouse:
The only way would be to test the focus order (simulating tab presses) before opening/reparenting and trapping tab/shift-tab to move focus to the correct items, but that would be really tricky and probably really buggy.

tmorehouse:
WAI-ARIA recommends focusing the first item in dropdown menus when first opened, so that screen reader users are "inside" the dropdown, and that tabbing while inside the menu would close the menu and tab to the next document focusable item. Moving between menu items while opened is only handled by cursor up/down/left/right (although this doesnt work well for form controls (especially textarea)

This is what we have done in the Bootstrap-Vue implementation of Bootstrap V4 dropdowns (at the moment)

patrickhlauke:
> WAI-ARIA recommends...

but these are not menus in the WAI-ARIA sense, merely generic expand/collapse dropdowns (as they can contain more than simply menu items). see https://getbootstrap.com/docs/4.0/components/dropdowns/#accessibility

tmorehouse:
@patrickhlauke yeah, I've seen that section before, and it is why we haven't implemented forms inside dropdowns (yet), as the native Bootstrap V4 implementation is not as intuitive to some screen reader users.

We are thinking of adding n option to change from role `menu` to role `popup` when the dropdown contains form elements (other than standard acceptable menu items), and change from keyboard navigation (up/down) to pure tab only navigation.  Although this switch might be confusing to keyboard users (but not to screen-reader users).

tmorehouse:
But back to the re-parenting discussion,  I think moving the menu to somewhere else in the body will be confusing to keyboard users (due to the possibility of getting "lost" in the document when tabbing around).

patrickhlauke:
> We are thinking of adding n option to change from role `menu` to role `popup`

assuming you mean `dialog`, that was actually something i considered not long ago...making the trigger elements `aria-haspopup="dialog"` and treating the dropdown itself as a form of non-modal dialog (including managing the focus to remain inside it, possibly)

tmorehouse:
Yep, meant to say 'dialog'.

Johann-S:
I updated my Codepen : https://codepen.io/Johann-S/pen/EwoPYo
with the keyboard navigation fixed :+1:

danijelGombac:
@Johann-S The tab key not working how it should. When tab is pressed the menu close.

patrickhlauke:
keyboard navigation fixed how? see https://codepen.io/patrickhlauke/pen/ZaQzMR ... tab to open the dropdown, then tab again - focus moved to the "foo" links. and yes as noted, once user eventually gets to the dropdown itself, tab closes the menu.

i think the reparenting of the dropdown is not a good idea, as it just adds too many complications (in our current implementation)

Johann-S:
Yep and change the behavior on how our Dropdowns work on TAB adds a lot of complexicity :worried: 
So we are stuck for this issue https://github.com/twbs/bootstrap/issues/24251

FezVrasta:
I added some information useful to provide a workaround for #24251 and that could possibly replace the most basic usages of this very own PR

tmorehouse:
Would some of the Popper.js overflow modifiers fix this? https://popper.js.org/popper-documentation.html#modifiers..preventOverflow

tmorehouse:
I think PR #24976 provides a better solution without the need for re-parenting the `.dropdown-menu`, taking @FezVrasta 's idea (`position: static`) suggestion, and adding in Popper's `boundariesElement` modifier.

Johann-S:
closed because https://github.com/twbs/bootstrap/pull/24976 is better :+1:

Issue #23612, opened 2017-08-22
File browser not working for non `en` language pages


ztvMark:
Bootstrap 4
Chrome Versión 59 And Firefox Last Version
Ubuntu 16.04

File browser No Working
https://getbootstrap.com/docs/4.0/components/forms/#file-browser

Appears in white as in the following image
http://imgur.com/a/sPJgo

Do not see choose file or browse

patrickhlauke:
cannot reproduce. displaying fine for me. what browser/OS?

ztvMark:
Chrome Versión 59 And Firefox Last Version
Ubuntu 16.04

samhoward86:
I'm seeing the same in Chrome Version 60.0.3112.101 (Official Build) (64-bit)

The ::before and ::after CSS selectors are not being rendered.

patrickhlauke:
Works for me in Chrome 60.0.3112.101 (Official Build) (64-bit), Firefox, IE 11. So not sure what's going on there...

![capture](https://user-images.githubusercontent.com/895831/29585519-5348a8bc-877f-11e7-93b2-307d4c268c2e.PNG)

samhoward86:
This simple codepen doesn't render properly for me.

https://codepen.io/anon/pen/EvLqdq

samhoward86:
I added ` lang="en"` to the html tag and it is now rendering the "Browse" text. It is still not displaying the "Choose file..." text.

patrickhlauke:
aha, confirmed. seems the CSS is specifically targetting English only

samhoward86:
Thanks @patrickhlauke

ztvMark:
True if you change lang = "en" it works in Spanish no

mdo:
I imagine this still holds true for Beta 3. Thinking about it, we might want to include some default `content` values and then override for the languages. That way, anyone without an explicit `lang` attribute could pick up at least the English version.

Borkason:
Does not work for me even with `<html lang="en">`. Best example is https://getbootstrap.com/docs/4.0/components/forms/#file-browser

In Edge I have a working File Browser, but in Firefox 57.0.4 it just looks like this:

![grafik](https://user-images.githubusercontent.com/2971735/34774403-7d858b32-f60f-11e7-9b6f-85f71d49d43d.png)

patrickhlauke:
@Borkason that particular thing is likely due to https://github.com/twbs/bootstrap/issues/25112 which has been fixed in the repo already

PR #23632, opened 2017-08-23
Add default en label in custom file button

chiraggmodi:
currently there is no default content added in custom file button. i have add English test as default test.

Fixes #23612

PR #24259, opened 2017-10-05
Fix file browser for non `en` language pages 

ghost:
Fixes: #23612

I propose that the text variable for the custom file browser follow the same logic as the color variables (see the beginning of the variables file).

And, in addition, establish a text by default (in english), so that the custom file browser works even if it uses an unconfigured language and even if the "lang" attribute is not present.


scss/_variables.scss
```diff
@@ -447,14 +447,18 @@ $custom-file-border-radius: $input-border-radius !default;
 $custom-file-box-shadow:    $input-box-shadow !default;
 $custom-file-button-color:  $custom-file-color !default;
 $custom-file-button-bg:     $input-group-addon-bg !default;
-$custom-file-text: (
+
+// stylelint-disable
```
XhmikosR: Why do you disable stylelint here exactly?

XhmikosR: Why do you disable stylelint here exactly?

XhmikosR (dismissed):

scss/_variables.scss
lucascono: Stylelint suggests that maps begin on separate lines. I could have put the parentheses in separate lines but as in the color variables it is disabled and so it was approved, I thought it was the right way. It can be re-enabled but to pass the test have to put all parentheses in separate lines. And it should also be done in color variables.

XhmikosR: Ok, thanks for the explanation. There must be some way to tweak the rule for this, but I didn't have the time to look at it.

Either way, can you just disable the specific stylelint error instead of all?

```
// stylelint-disable rule
...
//stylelint-enable rule
```

lucascono: Yes, of course, I do now.

scss/_variables.scss
```diff
@@ -448,7 +448,8 @@ $custom-file-box-shadow:    $input-box-shadow !default;
 $custom-file-button-color:  $custom-file-color !default;
 $custom-file-button-bg:     $input-group-addon-bg !default;
 
-// stylelint-disable
+// stylelint-disable function-parentheses-newline-inside
```
XhmikosR: Can you just combine the rules with a comma between?

lucascono: I did not know it. Everyday learn something new. Thank you.

lucascono:
@XhmikosR Can you cancel the last change? It was a mistake. It's for another PR.

docs/4.0/getting-started/download.md
```diff
@@ -10,7 +10,7 @@ toc: true
 
 Download ready-to-use compiled code for **Bootstrap v{{ site.current_version}}** to easily drop into your project, which includes:
 
-- Compiled and minified CSS bundles (default, grid-only, and Reboot-only)
+- Compiled and minified CSS bundles (see [comparative of css files]({{ site.baseurl }}/docs/{{ site.docs_version }}/getting-started/contents/#comparative-of-css-files))
```
XhmikosR: Wrong branch, please revert the unrelated changes and push them to a new branch which will only have those.

lucascono: Yes, I apologize. All the contributions I'm making to Bootstrap today is to learn how to use git. It will not happen again.

XhmikosR: No worries :)

The patches will be squashed into one when merge. /CC @mdo

XhmikosR:
@mdo: we have #23632 for the same issue too. Choose which one to merge when you tackle this.

XhmikosR (changes requested):
Wrong merge. Would be better if you didn't do any manual changes unless you are 100% sure.

PR #24265, opened 2017-10-05
split input-btn variables into btn and input variables (issues #23550)

### LINKED ISSUE #23550: Padding for buttons and inputs
Buttons and input inherit the padding from the _$input-btn-padding-x_ and _$input-btn-padding-y_. For _$input-btn-padding-y_ is entirely true, for there is a problem with _$input-btn-padding-x_, y buttons, large side indent and have indented input is much less. Please add to _$input-btn-padding-x_ and _$input-btn-padding-y_ variables **_$input-padding-x_** and **_$input-padding-x_**. Let them be the default

isychev:
Fixes #23550.


scss/_variables.scss
```diff
@@ -299,17 +299,29 @@ $table-dark-color:           $body-bg !default;
 //
 // For each of Bootstrap's buttons, define text, background and border color.
 
-$input-btn-padding-y:       .375rem !default;
-$input-btn-padding-x:       .75rem !default;
-$input-btn-line-height:     $line-height-base !default;
+$input-padding-y:       .375rem !default;
```
andresgalante: Can you please move `$input-*` variables under "Forms" (after line 331).

scss/_variables.scss
```diff
@@ -299,17 +299,29 @@ $table-dark-color:           $body-bg !default;
...
+$input-line-height-sm:  $line-height-sm !default;
 
-$input-btn-padding-y-lg:    .5rem !default;
-$input-btn-padding-x-lg:    1rem !default;
-$input-btn-line-height-lg:  $line-height-lg !default;
+$input-padding-y-lg:    .5rem !default;
+$input-padding-x-lg:    1rem !default;
+$input-line-height-lg:  $line-height-lg !default;
+
+$btn-padding-y:       $input-padding-y !default;
```
andresgalante: Once we move input variables under Form we will be changing the order, so we will have to refence `btn-*` to `input-*` instead of the other way around.

@mdo we don't usually cross reference variables. Do you think we should add values absolute values in rem instead of abstracting them this way?

andresgalante (changes requested):
@isychev I made a few notes about the order and position of the variables, other than that nice work! I think this is a great move. 😃

andresgalante:
This PR closes #24265

scss/_variables.scss
isychev: @andresgalante good correct.

scss/_variables.scss
isychev: @mdo I did so as not to break the old structure. No differences in the absolute values or copied from the variables. I think it is better to make the absolute values, it is also easier to change

isychev:
@andresgalante  send new commit

andresgalante (approved):
I just tested the latest changes and it works as it should.

Thanks a lot @isychev ! 👍
 
@mdo up to you

scss/_variables.scss
```diff
@@ -301,17 +301,20 @@ $table-dark-color:            $body-bg !default;
...
-$input-btn-padding-y:         .375rem !default;
-$input-btn-padding-x:         .75rem !default;
-$input-btn-line-height:       $line-height-base !default;
 
-$input-btn-padding-y-sm:      .25rem !default;
-$input-btn-padding-x-sm:      .5rem !default;
-$input-btn-line-height-sm:    $line-height-sm !default;
+$btn-padding-y:                  .375rem !default;
+$btn-padding-x:                  .75rem !default;
+$btn-line-height:                $line-height-base !default;
```
XhmikosR: Please don't change indentation in this block; just follow the previous one.

XhmikosR (changes requested):

mdo:
We had this at one point, and we moved this direction to ensure inputs and buttons (which are used in very close proximity to one another) are sized correctly relative to one another. I think this is ultimately too  much of a departure from the current setup, especially for moving from beta 1 to beta 2.

I could potentially see us having an abstracted set of vars with the current `$input-btn-` prefix and then assigning those to `$btn-` and `$input-` variations. That's a bit of overhead, but it might add the clarity you're looking for while maintaining our current implementation and intent.

If you're interesting, please open a new PR for that. Thanks!

PR #24271, opened 2017-10-06
Describe compiled css

### LINKED ISSUE #24262: Docs: Describe compiled CSS files in download bundle
I downloaded the pre-compiled bootstrap-4.0.0-beta-dist.zip and it includes three flavors of CSS files:

- bootstrap-grid.css
- bootstrap-reboot.css
- bootstrap.css

It didn't seem evident from a few searches whether I needed all 3 files, whether they referenced each other with `@import`, or what. The only reference I can find in the documentation is this on the Download page (emphasis mine):

> Download ready-to-use compiled code for Bootstrap v4.0.0-beta to easily drop into your project, which includes:
> - Compiled and minified CSS bundles (**default, grid-only, and Reboot-only**)
> - Compiled and minified JavaScript plugins

These references cannot be found by search, since they don't match the file name.

It would be nice if the documentation went into more detail about these.

ghost:
Fixes #24262. Thank you @XhmikosR for the help.


XhmikosR (approved):
LGTM Waiting for @mdo's word too

docs/4.0/getting-started/contents.md
```diff
@@ -18,14 +18,70 @@ bootstrap/
...
 └── js/
     ├── bootstrap.js
     └── bootstrap.min.js
 {% endhighlight %}
 
 This is the most basic form of Bootstrap: precompiled files for quick drop-in usage in nearly any web project. We provide compiled CSS and JS (`bootstrap.*`), as well as compiled and minified CSS and JS (`bootstrap.min.*`). CSS [source maps](https://developers.google.com/web/tools/chrome-devtools/javascript/source-maps) (`bootstrap.*.map`) are available for use with certain browsers' developer tools.
 
+### Comparison of CSS files
+
+<div class="table-responsive">
```
andresgalante: Responsive tables have a different markup:
https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables

docs/4.0/getting-started/contents.md
```diff
@@ -18,14 +18,70 @@ bootstrap/
...
+  <table class="table">
```
andresgalante: I'd make this `table table-bordered` to make it easier to read

docs/4.0/getting-started/contents.md
```diff
@@ -18,14 +18,70 @@ bootstrap/
...
+    <thead>
+      <tr>
+        <th scope="col">CSS files</th>
+        <th scope="col" class="w-25 text-center">Layout</th>
```
andresgalante: Please remove all `w-*` classes

andresgalante (changes requested):
Thanks a lot @lucascono for the changes.

@XhmikosR I wonder if we need the table or just a list of 3 items that tells what each CSS file has. What do you think?

XhmikosR:
@andresgalante: I agree with your changes, except maybe for `w-25`; without it the columns don't take the whole space.

XhmikosR:
OK, I made the changes @andresgalante requested. @mdo: please have a look when you can.

An alternative solution would be just a list, but I feel like a table is better.

lucascono:
Hi, I had not responded before because of health problems. Class w-25 was added to make the width of the columns equal in a table with 100% width. Anyway it's just a question of criteria, there is no problem in changing it.

andresgalante (approved):

PR #24274, opened 2017-10-06
Update Popper.js to 1.12.5

Johann-S:


skmedix (commented):
I don't think the new version of popper has the same hash as the old one

Johann-S:
@skmedix https://www.srihash.org/ give me the same hash for 1.12.5 with this url : https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.5/umd/popper.min.js :worried:

skmedix:
@Johann-S Because there is no difference between these two versions. Compare them on https://diffnow.com

Johann-S:
That's strange according to the release note here : https://github.com/FezVrasta/popper.js/releases/tag/v1.12.5

XhmikosR:
What I understand from the release notes is that we basically don't need this update since it's the same as 1.12.3.

Johann-S:
So true @XhmikosR

anve:
I'm kind of new to this, so my apologies if this is already covered (maybe in upcoming beta 2?) or posted in the wrong place;

When using Popper v1.11.0, as suggested by the docs (in Boostrap v4 beta), I have a very hard time getting tooltips to position correctly when using 100% body height. This seems to be the issue described here:

https://github.com/FezVrasta/popper.js/issues/302
See demo: https://codepen.io/skunkbad/pen/prrOWR

Upgrading to Popper v1.12.5 solves the issue for me.

Johann-S:
That's right @anve but Popper.js 1.12.3 and 1.12.5 are the same 😉

PR #24278, opened 2017-10-06
fix broken footer links

leggiero:


XhmikosR:
We don't make changes like that to gh-pages. Everything goes into v4-dev and committed to gh-pages when a release is made.

Issue #23999, opened 2017-09-19
Logic of grid gutter units


Kopyov:
What is the reason of using px in $grid-gutter-width? All other components are using REM units for horizontal spacing (navbar, navs etc), so if i want to align padding in components: navbar + content container with nested grids + footer, i need to add left/right padding to my container and footer in REM value, but nested grids will have padding in PX. What's the reason behind this?

andresgalante:
Hi @Kopyov I am with you on this one, I'd rather do grid in rem. If you search you'll find a lot of issues related to this explaining the reasoning.

mdo:
The answer is in the docs—this is because of viewports being measured in pixels.

PR #24283, opened 2017-10-06
Change `$grid-gutter-width` to rem from px

pat270:
This changes `$grid-gutter-width` from 30px to 1.875rem. The px value makes it hard to calculate things based on `$grid-gutter-width` when everything else is based on rems.


andresgalante:
Hi @pat270 this has come up many times, for example here #23999
The grid is measured in pixels. I don't think we want to measure breakpoints in `px` and gutters in `rem`

PR #24296, opened 2017-10-08
Fix dropdowns

PascalPixel:
Dropdowns don't work on master, there is no `.show` applied to the `.dropdown-menu`, but there is an `.open` on the parent `.dropdown` now, fixed scss accordingly...

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -88,7 +88,7 @@
   }
 }
 
-.dropdown-menu.show {
+.dropdown.open .dropdown-menu {
   display: block;
 }
 


XhmikosR:
I doubt this is right. There is `.show` added via JS.

Superpencil:
@XhmikosR Perhaps this is just me, I have other issues too:
- Popovers remain invisible due to the `.fade` on them having `opacity: 0` and never being overwritten with a `.show`, they do get an `.in` but there is no class for it in `bootstrap/scss/_transitions.scss`
- Modals also remain invisible due to the same `.fade`
- modal-backdrops are full black and have no transparency (due to there being an .in but not a .show)

XhmikosR:
These are unrelated issues to this; you should make new issues after you search the issue tracker.

This PR is definitely wrong, I'll wait for someone else to confirm before closing it.

patrickhlauke:
building the latest `v4-dev` locally from scratch and running local documentation, dropdowns, popovers, modals all work as they should, so the problem definitely lies at whatever you're doing @superpencil and not with bootstrap.

Superpencil:
I made a mistake and was using the ES6 JavaScript files directly in the browser, only the dropdowns, popovers and modals didn't work, guessing they missed a dependency 😩

Sorry for wasting your time @XhmikosR @patrickhlauke

PR #24298, opened 2017-10-08
Update devDependencies.

XhmikosR:


tomscholz:
We could use greenkeeper.io to automate dependencies updates

XhmikosR:
I'm not very keen of it; I'd rather do it myself. That being said, if the other team members want such solution, we can try that.

Johann-S (approved):

Issue #24137, opened 2017-09-27
Navbar expand breaks at 767px


mhverbakel:
The layout breaks at exactly 767px wide screens. The easiest example is the documentation.
1. Go to https://getbootstrap.com/docs/3.3/css/#responsive-utilities-tests in Chrome or Firefox.
2. Use the emulator to limit your screen width to 766 pixels (height is not relevant).
3. Scroll to the green checks that show which sizes are shown and hidden (it's under Test cases).
4. Check that the page works as intended (e.g. there are at least some green checks visible).
5. Increase the screen width to 767 pixels.
6. Check that the page breaks (e.g. there are no more checks for visible or hidden).
7. Increase the screen width to 768 pixels.
8. Check that the page works as intended (e.g. there are green checks visible again).

My OS is Windows 10, I've been able to reproduce this on different machines using Chrome and Firefox (I was unable to reproduce this on MS Edge).

I think this is a browser issue, but it happens in multiple browsers, so this might be another issue. (I narrowed it down to 'media queries under 768 are exclusive, media queries above 767 are inclusive', and I have a minimal html/css test page if needed). Is there a way Bootstrap can work around this issue?

In case you cannot reproduce step 6, this is what it looks like on my PC:
![image](https://user-images.githubusercontent.com/11353785/30907633-70bb5756-a37b-11e7-9ab7-2cc290040d20.png)

Possibly related issues: #18845, #15787

mhverbakel:
In case you cannot reproduce step 6, this is what it looks like on my PC:
![image](https://user-images.githubusercontent.com/11353785/30907633-70bb5756-a37b-11e7-9ab7-2cc290040d20.png)

mhverbakel:
Possibly related issues: #18845, #15787

Johann-S:
Bootstrap 3 is essentially in maintenance mode as we focus on working towards a stable v4. As such, we're only accepting changes to v3's code on a case-by-case basis, usually only for critical bug fixes or docs improvements.

mhverbakel:
Okay, then let me provide an example of the same issue in v4.

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
[...]

Issue #21552, opened 2017-01-05
.hidden-* broken for subpixel display widths between breakpoints.


federico-peralta-schaffner:
Consider the following two spans:
```
<span class="hidden-lg-up" style="color: blue; ">Visible on xs, sm and md</span>
<span class="hidden-md-down" style="color: red; ">Visible from lg on, I can be longer</span>
```
The idea is that only one span is displayed at any given time. The blue one should be displayed on `xs`, `sm` and `md` viewports, while the red one is for `lg` and `xl` viewports.

However, in my laptop, when the width of the viewport is reported to be `991px`, **both spans are displayed**.

Here's the reduced test case: [JS Bin](http://jsbin.com/cimuzukada/edit?html,output)

To reproduce, manually change the width of the `Output` iframe until the reported width reaches `991px`.

I think that this happens because Chrome reports a width of **`991.4` pixels** (I've used Chrome DevTools to check this). And the relevant media queries for Bootstrap's hidden responsive utilities are as follows:
```
@media (max-width: 991px) {
  .hidden-md-down {
    display: none !important;
  }
}
@media (min-width: 992px) {
  .hidden-lg-up {
    display: none !important;
  }
}
```
This means that, when the width of the viewport is between `991px` and `992px` (and mine seems to be `991.4px`), none of the media queries above evaluates to true, so both spans end up being displayed.

I know that these pixels are not real pixels, but logical CSS pixels, abstracted by the browser. Despite this, I think that Bootstrap should not depend on how the browser converts real pixels to CSS abstracted pixels in order to behave as expected.

Maybe the media queries can be modified to include decimals?
I.e. `@media (max-width: 991.999999px)`
Here's the idea: [JS Bin](http://jsbin.com/xujakifuxu/edit?html,output)

OS: windows 10 home
browser: chrome v 55.0.2883.87 m
hardware: asus ux303u notebook pc
display configuration: 3200 real pixels wide, but screen is OS-zoomed to 250% recommended setting

mdo:
This has come up in a couple issues I think, but I don't see that there's anything we can do about it. Modifying media queries to use crazy decimals doesn't seem like the ideal answer either as it'll just bring up more questions and potential confusion.

For the time being, I'm thinking this is a won't fix. Ideally this is something that is resolved with browser vendors who have the power to control how their pixels are rendered more than us.

patrickhlauke:
FWIW filed https://github.com/w3c/csswg-drafts/issues/984

[...]

PR #24299, opened 2017-10-08
Add support for fractional viewport widths (zoom/high-dpi displays)

patrickhlauke:
Current breakpoint `min-width`/`max-width` calculations have a gap when viewport width is fractional (as can happen when the user zooms, but more importantly when the user has a high-dpi display with a fractional dpr/devicePixelRatio, such as the Surface 3 at `1.5`). See the related clarification I added to CSS MQ4 https://github.com/w3c/csswg-drafts/pull/1083

Instead of spacing the `min-` and `max-` values by `1px`, precision is now increased to `0.01px`, which should cover most fractional values in the foreseeable future.

xref https://github.com/twbs/bootstrap/issues/24137, https://github.com/twbs/bootstrap/issues/21552


patrickhlauke:
@mdo i know you said this would not be the ideal solution https://github.com/twbs/bootstrap/issues/21552#issuecomment-275019087 ... but I think the impact is minimal (in terms of confusion) and it elegantly solves a problem we'll hit more and more with high-dpi displays that use fractional dpr

XhmikosR:
@patrickhlauke: I suppose the two decimal points are enough, right?

Otherwise it seems like a fine solution until this is sorted by browsers, if ever.

XhmikosR (approved):

patrickhlauke:
> until this is sorted by browsers

once browsers support proper range context queries, we'll be able to do things like

```
@media (width < 576px) { ... }
@media (width >= 576px) and (width < 768px) { ... }
@media (width >= 768px) and (width < 992px) { ... }
...
```

but yes, for the time being, using decimal points is the best solution (and to be clear, current behavior is not a browser bug per se, just an inadequacy in the original `min-` and `max-` spec which didn't foresee the situation with fractional values)

patrickhlauke:
going to wait for @mdo to give his final blessing :)

andresgalante (approved):
Hi @patrickhlauke I wonder if it's worth adding a comment on the scss that explains why the `*.99px`

Other than that I am for merging this PR 😄

patrickhlauke:
@mdo unless you're philosophically opposed to this, I'd love to merge this for beta 2
@andresgalante i think it's fairly self-explanatory, no need to document?

also worth keeping an eye out for this bug https://bugs.webkit.org/show_bug.cgi?id=178261 if it starts having an impact

mdo:
I'd prefer to wait on this—we don't see that many problems with this in the grand scheme and the approachability of our setup goes down significantly with changes like these.

patrickhlauke:
I can't quite see what the problem is, @mdo...

Going from "-1px" logic to "-0.01px" doesn't seem to be a significant barrier to approachability to me. Would adding a note in the docs alleviate your concern?

As for problems, this also tackles some of the problems brought up by zooming (when zooming to values that result in fractional viewport widths), so would solve some (but not all) long-standing "we don't support this" type of issues, issues on devices like Surface 3 with non-integer dpr, and future-proofs the queries.

patrickhlauke:
@mdo FWIW i just pushed a new commit with some extra documentation (as an include) and comment for the SCSS. see if that would help sway you?

further, i'll be sending a proposed PR against the CSS4 MQ spec to include the example of high-precision values to deal with fractional viewports as a solution until range queries are supported in browsers. this way, we'll be using an approach that will be actually covered in the relevant spec itself.

patrickhlauke:
xref https://github.com/w3c/csswg-drafts/pull/1884

simevidas:
@patrickhlauke W.r.t. my WebKit bug that you’ve linked above, it might be worth performing a few more tests on iOS devices. According to [my test](https://twitter.com/simevidas/status/918843619326550016), `(max-width: 374.99px)` evaluates to `true` on an iPhone with a 375-pixel viewport (that’s the bug), but `(max-width: 374.98px)` doesn’t. Additional tests on other devices (iPhone Plus, iPad, etc.) could effectively confirm that the `.99px` fraction is problematical, and that `.98px` is safe, in which case Bootstrap could just use that fraction instead.

patrickhlauke:
but it's a clear bug in webkit, and as those breakpoints don't fall on any specific phone/tablet sizes, i'd rather keep it clean and understandable (otherwise the question arises "why -0.02px and not -0.01px" for the docs...

patrickhlauke:
"those breakpoints" meaning the ones BS uses, i.e. 575.99px, 767.99px, 991.99px, 1199.99px

patrickhlauke:
@mdo as the CSS 4 spec now explicitly has my example of using higher-precision values, and the callout should clarify any minor confusion seeing those ".99px" values may cause...are your concerns alleviated? I'd really like to put this one to bed and merge it, as I don't think in the grand scheme of things it will cause much, if any, confusion for authors

liquidvisual:
Just an FYI - this change had the adverse effect in my unzoomed MacOS Safari (hdpi and normal screens); causing elements that should be hidden to show during a 1px gap.

simevidas:
@liquidvisual For the record (and for what it’s worth), this issue does not manifest if the MQ breakpoint is changed from `767.99px` to `767.98px`, according to my test in Safari.

patrickhlauke:
so -0.02px instead of -0.01px? ok, i'll make a patch with a comment in the CSS later...sigh

patrickhlauke:
@simevidas et al - see https://github.com/twbs/bootstrap/pull/25177

PR #24303, opened 2017-10-09
enable sourcemaps for bundles, include sourcemap for plugins in npm package

delagen:
related #19063


XhmikosR:
@Delagen: why haven't you enabled sourcemaps for all JS files? Also, you should inline the source files to avoid issues like #19063.

@mdo @Johann-S: do you remember what was the reason for us not to include sourcemaps for JS files? It's like this from v3, so there must be some reasoning.

Delagen:
@XhmikosR As i see, dist/js/bootstrap.bundle.js.map contains all sources, and others also.

XhmikosR:
@Delagen: even so, I'll verify after you address the other issue I mentioned above.

Delagen:
@XhmikosR Which issue? Include recompiled files in commit?

XhmikosR:
@Delagen: you haven't enabled sourcemaps for all of our dist JS files.

Delagen:
@XhmikosR enabled for minified files

XhmikosR (dismissed):
We need `includeSources` and `url` for uglify-js. See https://github.com/mishoo/UglifyJS2#cli-source-map-options

Also, we need to do the same for `js-minify-docs`.

Delagen:
@XhmikosR updated

XhmikosR:
@Delagen: `url` isn't a boolean; you need to specify the filename.

Delagen:
@XhmikosR it use basefilename with postfix by default

XhmikosR:
Because the bundle already has sourcemaps specified in rollup step.

XhmikosR:
@Johann-S: any ideas?

Delagen:
@XhmikosR I understand. I don't use cli tools at all

XhmikosR:
The sourcemaps still don't show the sources for me. I'll try to find why, otherwise if anyone has any ideas please comment here.

XhmikosR:
Hmm, Chrome loads the source files fine. FF v57 beta doesn't here on Windows.

Delagen:
@XhmikosR I think it's due base&root source map settings. Many `../` in source paths.

XhmikosR:
@Delagen: that is with Chrome, right? Because with Chrome it works for me too but with FF doesn't work.

Delagen:
@XhmikosR Also in Firefox 
image

XhmikosR:
Ah, true. It's grouped under the `(no domain)` row.

All tight, all seems good now.

One last thing; should we include the map files for CSS too?

XhmikosR:
Scratch my last comment.

XhmikosR (approved):

Delagen:
@XhmikosR Yes. The problem is in base and root )

XhmikosR:
@Delagen: do you have a fix for that?

Delagen:
@XhmikosR I think rollup calculate paths relative to output, so I'll see how to tune this.

Delagen:
@XhmikosR I think if files required from node_modules, all be fine due long path to it, and it points to real files. If we change source mapping, we broke package debugging, so I think we must keep it unchanged.

XhmikosR:
@mdo: are you OK with adding sourcemaps for JS files? I know we didn't include them in v3, not sure what the reasons were.

They don't hurt being present plus we already include sourcemaps for CSS files.

XhmikosR:
@mdo: ping

mdo:
I've no preference on including or not—that's a better question for @Johann-S and @bardiharborow.

Johann-S:
Not against it because we already provide source map for our minified CSS

XhmikosR:
@mdo: when you do `npm run dist` next time, don't forget to include the new files.

PR #24305, opened 2017-10-09
package.json: Change repo URL to `git+https`.

XhmikosR:


Johann-S (approved):

bardiharborow:
Is there a good reason for this? I had a search through the git and npm docs am I'm not sure what you're trying to do.

XhmikosR:
I'm not "trying" to do anything; this is the default URL when the repo is https. Do `npm init` and you'll get the same.

bardiharborow:
@XhmikosR hmm, interesting that's not documented anywhere. I'm not sure if there's a protocol difference, or if they mean the same thing, but if you're just normalising configs then I guess that's all good. 😄

PR #24309, opened 2017-10-09
Use `$white` variable for `.navbar-dark`'s color.

XhmikosR:
Using `rgba($white,1)` results in `white` instead of the hex.


andresgalante (approved):
😁 yeap this works as it should

PR #24315, opened 2017-10-09
Reduce z-indexes in button-group, input-group, list-group, and pagination to the minimum necessary

rmacklin:
These were using `z-index: 2` to ["Place active items above their siblings for proper border styling"](https://github.com/twbs/bootstrap/blob/bab3246a1d7219e93ee12e09ff7608c5dd6bcbc2/scss/_list-group.scss#L73). However, using `z-index: 1` is sufficient for accomplishing that goal.

In input-group, there were also three `z-index: 3` rules for the hover/focus/active states. I reduced these to `z-index: 2` since they just needed to be "one more than normal" (i.e. one more than what is now `z-index: 1` after my changes).

These changes can be verified by viewing the documentation pages for [Button group], [Input group], [List group], and [Pagination] before and after this commit and observing that the active elements are still "above" their siblings, so their borders look correct. (In fact, the easiest way may just be to change these z-indexes in the dev tools style editor - that's what I did before opening this PR.)

I also added a brief paragraph in the documentation about these z-indexes.

[Button group]: https://getbootstrap.com/docs/4.0/components/button-group/
[Input group]: https://getbootstrap.com/docs/4.0/components/input-group/
[List group]: https://getbootstrap.com/docs/4.0/components/list-group/
[Pagination]: https://getbootstrap.com/docs/4.0/components/pagination/


andresgalante (commented):
@rmacklin Thanks for the PR, you are right there is no need to use `2` and `3` z-indexes.

I don't know if @mdo did it intentionally to leave space for a possible `z-index: 1` element in the future. 

I also don't think there is a need to document this. I don't see the relationship between the z-index of these elements and layouts.

mdo (approved):

mdo (approved):

rmacklin:
Thanks!!

PR #24317, opened 2017-10-10
Add information about failed tests

Johann-S:
Add some information about the number of failing tests

### DIFF
--- a/build/saucelabs-unit-test.js
+++ b/build/saucelabs-unit-test.js
@@ -70,7 +70,13 @@ const waitingCallback = (error, body, id) => {
       // Exit
       if (jobsDone === browsersFile.length - 1) {
         jsUnitSaucelabs.stop()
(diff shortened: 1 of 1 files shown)


build/saucelabs-unit-test.js
```diff
@@ -70,7 +70,13 @@ const waitingCallback = (error, body, id) => {
       // Exit
       if (jobsDone === browsersFile.length - 1) {
         jsUnitSaucelabs.stop()
-        process.exit(jobsDone === jobsSucceeded ? 0 : 1)
+        if (jobsDone > jobsSucceeded) {
```
XhmikosR: Any reason not to use `!==`?

Johann-S: Nope but for me it's more clear than using `!==` here because the line after we substract those two numbers

XhmikosR (approved):

Issue #23478, opened 2017-08-16
PreventDefault on callBack hide function bug


LuigiOnGitHub:
Hi
in Bootstrap (v4.0.0-beta): modal.js release I found that the following code in Modal.prototype.hide function (starting line 1982 of bootstrap.js)

```js
var transition = Util.supportsTransitionEnd() && $(this._element).hasClass(ClassName.FADE);
if (transition) {
  this._isTransitioning = true;
}
```

should be put after the hideEvent trigger otherwise the property isTransitioning will be true at the next call and therefore a not prevented callback on hide will be inhibited.

correct code starting at line 1982 should be:

```js
var hideEvent = $.Event(Event.HIDE);

$(this._element).trigger(hideEvent);

if (!this._isShown || hideEvent.isDefaultPrevented()) {
  return;
}

var transition = Util.supportsTransitionEnd() && $(this._element).hasClass(ClassName.FADE);

if (transition) {
    this._isTransitioning = true;
}
```

regards

prateekgoel:
@mdo I would like to make this change.

XhmikosR:
@Johann-S: thoughts?

Johann-S:
Seems a good change to me if we can provide a relevant unit test

PR #24318, opened 2017-10-10
Modal plugin set _isTransitioning after the trigger of hide event

Johann-S:
I added a visual test because on QUnit we disabled any transition (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/util.js#L49)

Fixes : #23478


XhmikosR (approved):
LGTM as far as I can tell :)

PR #24321, opened 2017-10-10
Adds a border radius to the image when it's the only child of a card group

andresgalante:
When a card is the only child of a card group, img-top and img-bottom don't have a border radius:
screen shot 2017-10-10 at 10 45 02 am

This PR is adding those border radius back:
screen shot 2017-10-10 at 10 45 44 am

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -215,6 +215,13 @@
 
         &:only-child {
           @include border-radius($card-border-radius);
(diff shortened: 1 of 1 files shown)


XhmikosR:
Sorry for the conflict @andresgalante :/

XhmikosR (approved):

PR #24323, opened 2017-10-10
Add Prettier (Fixes #22919)

### LINKED ISSUE #22919: Prettier
Is there any plan to use https://github.com/prettier/prettier for code formatting in the next version of bootstrap? Could really help for any contributions

lipis:
Fixes #22919 

I removed bunch of styling rules.. so feel free to comment if we can remove more stylistic rules since most of them are covered by Prettier.

/cc @XhmikosR


package.json
```diff
@@ -33,6 +33,7 @@
     "js": "npm-run-all js-lint* js-compile* js-minify*",
     "js-main": "npm-run-all js-lint js-compile js-minify",
     "js-docs": "npm-run-all js-lint-docs js-minify-docs",
+    "js-lint-fix": "eslint --fix js/ && eslint --fix --config js/tests/.eslintrc.json --env node build/",
```
lipis: I'm open for suggestions on how to name the command and if we can do some optimisations here.

XhmikosR:
First of all, thanks for the PR!

There are many things I personally don't like in this config. I think we are better off just tweaking our rules.

1. For example, removing the spaces around `=` and `:` in objects are OK with me, but I know Johann is using them.

2. We could look into the chaining methods breaking but this is too much IMO:

    ```diff
    fs.createReadStream(workboxSWSrcPath).pipe(fs.createWriteStream(workboxSWDestPath))
    -fs.createReadStream(workboxSWSrcMapPath).pipe(fs.createWriteStream(workboxSWDestMapPath))
    +fs
    +  .createReadStream(workboxSWSrcPath)
    +  .pipe(fs.createWriteStream(workboxSWDestPath))
    +fs
    +  .createReadStream(workboxSWSrcMapPath)
    +  .pipe(fs.createWriteStream(workboxSWDestMapPath))
    ```

3. Having the brackets in the same line is definitely something I'd like to have.

4. Removing the spaces before object `:` is definitely something I'd like to have.

4. Removing the space before a function's parentheses is something I personally don't like.

5. This kind of changes are also OK to me:

    ```diff
    -      const rootElement      = $(this._element).closest(
    -        Selector.DATA_TOGGLE
    -      )[0]
    +      const rootElement = $(this._element).closest(Selector.DATA_TOGGLE)[0]
    ```

6. `arrow-parens` I think I find it better to keep them all the time

7. The line width is just too much; screens are too big to have a limit of 80. Personally I'd try using 120, but I think we better have this off anyway.

-----

That being said, I still think we should first tweak our ESLint rules. And perhaps change the location of `.eslintrc.json`.

/CC @Johann-S for feedback

XhmikosR:
Also, don't commit dist files. Although Hound should ignore those, I'll have a look at this.

lipis:
As you might know already.. Prettier doesn't have many options and that's why we have to ignore/comply with how Prettier is doing it (point 5 and 7 - not much we can do here).

8. I increased the size to 120.. waiting for more comments.

XhmikosR:
That is why the points I made above are mostly so that we tweak our existent ESLint config and drop the whole idea of prettier.

FezVrasta:
You may use prettier-eslint to get the best of both worlds

FezVrasta:
Also, I think it would help the review process if you splitted the PR in two commits: 1 about the Prettier integration, 1 about the Prettier formatting of the files.

lipis:
I'll do that.. with the commits, because at the moment I'm using the the eslint and Prettier..

XhmikosR:
I think we are better off with using https://standardjs.com/ even though it's not a formatter.

Personally I don't like the changes prettier makes here judging from the files committed.

/CC @Johann-S

lipis:
Of course it's an opinionated code formatter and if there are some things that we don't like there is not much we can as earlier stated.

Having said that I have two commits in this PR now to separate the [config](https://github.com/twbs/bootstrap/pull/24323/commits/720f4632b24ba1c9ca906ae337c7d33e6edce941) from [formatting](https://github.com/twbs/bootstrap/pull/24323/commits/2f7d8e1c612ed01bca29320bfafa06c877184770).

FezVrasta:
@XhmikosR try to look beyond the formatting changes, think that you will never have to think about how your code looks like and you will never have to manually format a line of code in your life.

That's liberating, and boosts productivity. 

Apart from that, all the contributions will already follow the "convention" because Prettier will take care of it for you guys.

XhmikosR:
Like I said, I don't like the style it's enforcing. Like, a lot. It's a matter of personal taste.

That being said, wait for Johann to chime in too.

Johann-S:
Sorry for the waiting :smile: 

I understand how it's convenient to not have to think how to format your code, that being said as @XhmikosR they are few rules I don't like...

- The chaining methods breaking (because it's too much for me here)
- I prefer having the brackets in the same line.
- Removing the spaces before object, same I'd like to have that.
- Removing the space before a function's parentheses
- arrow-parens I prefer keep them all the time
- 120 in terms of line width is fine for me too

But that's a matter of preferences, here finally I'm not sure about adding prettier to our workflow or standard js too.
I would prefer to tweak our Eslint rules and add a pre-commit hook which will format our code (`eslint --fix`)

FezVrasta:
@Johann-S all your points can be covered by some custom eslint rule ran after Prettier.

Except for the chaining methods, because they don't just break on multi line always, but only when they are too long to live in a single line.

ESlint can't cover all the cases covered by Prettier, on the other hand.

lipis:
Plus Bootstrap will be featured here as well.. as the most starred library :) hehe

https://prettier.io

lipis:
Channing methods it is not the best solution, because it will be hard also for the editor to figure out what's the correct way to format things...! We either go full Prettier or not at all.. and rely on other things for the formatting.

bennyn:
> ESlint can't cover all the cases covered by Prettier, on the other hand.

@FezVrasta What are the cases (for JS & JSX code formatting) that [ESLint](https://eslint.org/) can not handle?

FezVrasta:
https://github.com/prettier/prettier-eslint/issues/101

bennyn:
> prettier/prettier-eslint#101

The only precise example in that issue is doing auto formatting for code that exceeds the maximum limit of characters per line. That's something modern IDEs have already built in.

I am also not convinced by the "Gives you an integrated experience" (https://github.com/prettier/prettier-eslint/issues/101#issuecomment-313239741) claim. I would argue that the integrated experience of ESLint in WebStorm is much more mature.

FezVrasta:
Yeah I mean, I'm not paid to convince you, whatever 😅

lipis:
WebStorm costs some 💶 💵

bennyn:
@FezVrasta No offense! I don't want you to be a sales person. 😀 

I am just looking for hard facts what makes Prettier better than ESLint (for JS & JSX code formatting) and thought that I can maybe find the answers here. 😃

Johann-S:
Thank you all for your feedbacks and @lipis for your PR but we will pass on adding prettier to our workflow.

But we will tweak our Eslint rules and  a git hook which will run `eslint --fix`

lipis:
Fair enough.. see you in a few months.. :)

PR #24324, opened 2017-10-10
Hound CI: Specify `ignore_file` for stylelint.

XhmikosR:
Specify `ignore_file` for stylelint.

I noticed that Hound complained on #24323, which it shouldn't have so this hopefully should fix it.


bardiharborow (approved):
Not sure why Hound doesn't pick that up automatically...

PR #24329, opened 2017-10-11
Update .eslintignore

graceandchen:


bardiharborow:
I'm not sure what software is generating those files. Are you able to clarify?

PR #24334, opened 2017-10-11
Vertically aligns labels on horizontal forms using flex instead of math

andresgalante:
Now I got it! This PR changes vertical alignment of labels on horizontal forms from paddings and `clac` to a more robust solution with flex. Let's let the computer do the math 😄

The visual result is the same.

This PR is related to #24328

@XhmikosR @mdo what do you guys think?

### DIFF
--- a/docs/4.0/components/forms.md
+++ b/docs/4.0/components/forms.md
@@ -314,36 +314,36 @@ More complex layouts can also be created with the grid system.
 <form>
   <div class="form-row">
     <div class="form-group col-md-6">
-      <label for="inputEmail4" class="col-form-label">Email</label>
+      <label for="inputEmail4">Email</label>
       <input type="email" class="form-control" id="inputEmail4" placeholder="Email">
     </div>
     <div class="form-group col-md-6">
-      <label for="inputPassword4" class="col-form-label">Password</label>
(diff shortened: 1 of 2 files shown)


XhmikosR:
LGTM but better wait for @mdo. Maybe we are missing something.

andresgalante:
@XhmikosR Agreed, we are either missing something or the docs for Form row are wrong.

wojtask9:
This breaks cases where `.col-form-label`  is on the left  `.form-control` and `.invalid/valid-feedback` is present

```
<div class="form-group row">
  <label class="col-2 col-form-label" for="inputError">With error</label>
  <div class="col-10">
     <input id="inputError" class="form-control is-invalid" type="text">
      <span class="invalid-feedback">Please correct the error</span>
   </div>
</div>
```
If we want replace `calc()` we can add transparent `border`. `calc` is used here because of different units.

andresgalante:
@wojtask9 good catch! you are absolutely right! The presence of a `invalid-feedback` span would cause this:

screen shot 2017-10-11 at 11 18 08 am

Thanks a lot for reviewing this.

@XhmikosR @wojtask9 I think this is a no go for the `align-self` part for now. But I'll open a new PR for the docs fix, that should still be relevant, right?

Issue #23824, opened 2017-09-04
Unclear docs: v3-v4 migration of horizontal forms, .col-form-label vs. .form-control-label


Herst:
[The example code for a horizontal form in v3](https://getbootstrap.com/docs/3.3/css/#forms-horizontal):

```html
<form class="form-horizontal">
  <div class="form-group">
    <label for="inputEmail3" class="col-sm-2 control-label">Email</label>
    <div class="col-sm-10">
      <input type="email" class="form-control" id="inputEmail3" placeholder="Email">
    </div>
  </div>
  […]
</form>
```
Let's say we wanted to port it to v4. The relevant parts from [the migration document](https://getbootstrap.com/docs/4.0/migration/#content):

> - **Renamed `.control-label` to `.form-control-label`.**
[…]
> - Horizontal forms overhauled:
>   - Dropped the `.form-horizontal` class requirement.
>   - `.form-group` no longer applies styles from the `.row` via mixin, so `.row` is now required for horizontal grid layouts (e.g., `<div class="form-group row">`).
>   - **Added new `.form-control-label` class to vertically center labels with `.form-control`s.**
>   - Added new `.form-row` for compact form layouts with the grid classes (swap your `.row` for a `.form-row` and go).

(Highlight by me.)

OK, so the code would turn into:
```html
<form>
  <div class="form-group row">
    <label for="inputEmail3" class="col-sm-2 form-control-label">Email</label>
    <div class="col-sm-10">
      <input type="email" class="form-control" id="inputEmail3" placeholder="Email">
    </div>
  </div>
  […]
</form>
```
But if we were to look at the [v4 example code for horizontal forms](https://getbootstrap.com/docs/4.0/components/forms/#horizontal-form) we see that it's different:

```html
<div class="container">
  <form>
    <div class="form-group row">
      <label for="inputEmail3" class="col-sm-2 col-form-label">Email</label>
      <div class="col-sm-10">
        <input type="email" class="form-control" id="inputEmail3" placeholder="Email">
      </div>
    </div>
    […]
  </form>
</div>
```
There are two discrepancies:
* `.container` around it, I think this one can be ignored, it's obvious that it's just for style.
* `.col-form-label` for the label instead of `.form-control-label`. Former is never mentioned in the migration doc and latter is not mentioned in the forms documentation, so what is it now?

Herst:
Perhaps related to #18573?

andresgalante:
Hi @Herst, as usual you are right.

[...]

PR #24335, opened 2017-10-11
Removes unnecessary `col-form-label` from form row docs

andresgalante:
This PR removes unnecessary `col-form-label` from form row docs

cc/ @mdo @XhmikosR


Herst:
Is this related to #23824? Does that mean `.form-control-label` would also be useless there?

andresgalante:
@Herst You are right on #23824. I left a note there.
 
I believe that this PR is still valid.

mdo (approved):
Ah yes, these aren't needed here because this isn't a horizontal grid layout. Thanks!

PR #24340, opened 2017-10-11
Minor link issue on 3.3 homepage

i1u5:
Fixed the examples link on homepage of 3.3 bootstrap


XhmikosR:
Thanks but the master branch is the one you should target.

@mdo: this is why I told you the other day you need to clean up the v3 branches to avoid confusion...

PR #24342, opened 2017-10-12
Update Hound config.

XhmikosR:
1. Explicitly disable ESLint in Hound for now; it was disabled all this time. Temporarily disabled until the ESLint plugins are updated upstream
2. Remove unneeded ignore from .houndignore


XhmikosR:
OK, something is very weird in Hound...

@gylaz: You can see here our .eslintignore file and .houndignore file. Running `npm run js-lint` correctly ignores the files but Hound is still complaining, for example about the dist/css files. I added a .stylelintignore file, but shouldn't Hound respect the ignores provided in .houndignore?

XhmikosR:
@gylaz: I give up. Anything I tried just fails and Hound doesn't respect our ignores.

Please let me know if we are doing something wrong or if it's a bug on your side.

gylaz:
@XhmikosR It looks like we currently don't support an ignorefile for stylelint -- an oversight on our part.

XhmikosR:
@gylaz: thanks for the reply! That at least makes sense :)

Is there any issue I can track?

XhmikosR:
@mdo: I haven't disabled stylelint, although it is ignoring our ignores until https://github.com/houndci/hound/issues/1454 is fixed. Also, we will be able to actually enable ESLint, which was disabled all this time, after https://github.com/houndci/linters/issues/160 is fixed.

XhmikosR:
How do you want me to proceed in the meantime?

Johann-S (approved):

XhmikosR:
@Johann-S: can you try this branch to make sure we don't have any false positives from Hound? I moved tests/.eslintrc.json to unit to work around some FP.

Test by making changes in unit files that violate eslint only there.

XhmikosR:
@Johann-S: can you look into the eslint changes? I didn't split those, 'cause time, but you get the idea.

@mdo: don't forget to nuke Hound from the repo integrations and I'll remove .hound.yml in this branch.

XhmikosR:
@Johann-S: bump

Johann-S (approved):
It adds a lot of new eslint files but it seems a better way to manage our rules, nice work ! 👍

XhmikosR:
@mdo ping for the Hound removal

mdo:
The Hound webhook has been removed.

PR #24360, opened 2017-10-13
Moves `box-sizing` setting on reboot to the html selector

### LINKED ISSUE #24357: box-sizing best practice
Reading [reboot.css](https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap-reboot.css), I saw you set box-sizing using
```css
*,
*::before,
*::after {
    box-sizing: border-box;
}
```
instead of
```css
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```

Even if in most cases the first version would works fine, the second one allows devs to reset box-sizing for components designed to work in default box-sizing just by setting `box-sizing : conten-box` on the parent element.

Related article: [Inheriting box-sizing Probably Slightly Better Best-Practice](https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/)

andresgalante:
This PR moves `box-sizing` to the `html` selector. 

Closes #24357 Thanks @Dunexus

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -21,10 +21,11 @@
 *,
 *::before,
 *::after {
-  box-sizing: border-box; // 1
+  box-sizing: inherit; // 1
 }
 
 html {
+  box-sizing: border-box; // 1
   font-family: sans-serif; // 2
   line-height: 1.15; // 3
   -webkit-text-size-adjust: 100%; // 4


XhmikosR:
LGTM but we definitely need @mdo's word.

XhmikosR:
@andresgalante: just make sure you saw https://github.com/twbs/bootstrap/commit/7b66aeb2f6c7f066a8d1f18dae839240886c2fa0#diff-7f9eec82312835f6155c8de839bf0e49

andresgalante:
@XhmikosR thanks for the heads up. Until Chrome fixes the `<details>` bug let's keep things are they are.

PR #24362, opened 2017-10-13
Add border-y-0 and border-x-0, because why not?

kevinruscoe:
We have `border-left-0`, `border-right-0`, `border-top-0` and `border-bottom-0`. This adds `border-x-0` and `border-y-0`. which sets either top/bottom or left/right to zero.


AdrianKuriata:
I suggesting, maybe this classes is so long? I mean paddings and margins has a classes like a px, py, pt, pb, pr, pl and this is really faster for writting code, maybe border should has it too? I mean bx-0 - what mean border hasnt left and right borders and bx-1 what mean has border?

kevdotbadger:
I kinda agree. It also seems a little backwards. Following the p and m utils, this makes more sense.

`bd-0` - border: none;
`bd-1` - border: 1px solid;
`brd1`- border-(top/bottom): 1px solid;
`brdx-1` - border-(left/right): 1px solid;
`brdl-1` - border-left: 1px solid;
`brdr-1` - border-right: 1px solid;
`brdt-1` - border-top: 1px solid;
`brdb-1` - border-bottom: 1px solid;
`brdy-0`- border-(top/bottom): none;
`brdx-0` - border-(left/right): none;
`brdl-0` - border-left: none;
`brdr-0` - border-right: none;
`brdt-0` - border-top: none;
`brdb-0` - border-bottom: none;
`brd-primary`: border-color: primary (then the rest of the theme options).

@mdo thoughts on this?

AdrianKuriata:
Why brd, not B? Like b-0 or by-0?

kevdotbadger:
Assumed `b-` was used :)

AdrianKuriata:
By what? :O

andresgalante:
Hi @kevdotbadger thanks a lot for sending the PR, and also hi @AdrianKuriata,

Although I understand you point about been consistent with margin and padding utility classes, when we start reducing classes to just one letter it gets hard to read (similar to SMACC, or OSS AtomicCSS). The impact of performance you get for a few more characters on a class name is insignificant.

About adding `border-y` and `border-x` it'll up to @mdo. I personally don't see the need.

kevdotbadger:
@andresgalante Completely agree RE the single letter classes, as for this suggestion I disagree. Adding borders to just top and bottom you need 3 classes (`border border-left-0 border-right-0`), in terms of readability having just `border border-y` just makes more sense.

kevdotbadger:
@mdo thoughts on this addition?

Kopyov:
I'd presume b-* is better for bottom property, as used in margin/padding utils. What about:

- .border-b-0
- .border-t-0
- .border-l-0
- .border-r-0
- .border-x-0
- .border-y-0

Looks more consistent to me.

andresgalante:
Yeap, I meant `brd` vs `border`

kevdotbadger:
@mdo: ping

kevdotbadger:
@mdo: ping

mdo:
`b` could easily be confused for `background` without first seeing `bg`. Let's stick with `border` for now. I'll swing back around to merging this after the betas.

mdo:
Revisiting this, I'm unsure if we should proceed after we added the _additive_ border classes in our stable v4 release. Doing so means we either don't need these, or if we do need them, we likely need additive x and y variants. These start to add up quickly—thoughts?

Also, these classes would need documentation, but let's wait on that until we know we need it.

kevdotbadger:
I personally prefer `border-y` to `border-top border-bottom`. However, I can see your point in have both `x` and `y`, with the additive and subtractive classes. I'm happy to drop this, see what others think.

mdo:
Closing for now, will see what folks think with these new ones. Utilities for utilities is something I generally want to avoid though :).

PR #24366, opened 2017-10-13
Removes margin for `pre` on _code.scss since it's already set on _reboot.scss

andresgalante:
Reboot defines margins for `pre` https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L247-L249

This PR removes the `pre` margin definition on _code.scss


mdo (approved):
Nice, slating for Beta 3 so we can slim down remaining Beta 2 todos.

XhmikosR (approved):

PR #24379, opened 2017-10-15
Restore CNAME

XhmikosR:
Since the v4-dev branch will be eventually merged to master.


mdo:
It's worth noting that even when restored via `master`, this won't have an impact on the `gh-pages` branch, as we don't merge from `master` for publishing the docs. Instead, it's a manual copy-pasta process, and the `CNAME` is already there.

Perhaps we nix it from the dev branch instead?

XhmikosR:
@mdo: if we delete it, then when the merge happens, it will be deleted in master too.

mdo (approved):

Issue #24373, opened 2017-10-14
V4 Documentation - Various Issues


saturdaywalkers:
The doco. needs to be reviewed by some people who:
a) have never used Bootstrap before
b) have only used v3 before

From the perspective of migrating a site from v3 to to v4, and so far...

I can do a pull request if you don't already have people working on it.

1) Navbar ( https://getbootstrap.com/docs/4.0/components/navbar/ )

Add : For an explanation of the utility classes used here, see:  colours (e.g. bg-light) and spacing (e.g. my-2, m is for margin)

It would be worth doing this for all the components, especially were its not obvious, i.e. text-danger you can guess, my-3 you'd have to know.

2) Display Utility

"Responsive variations also exist for every single utility mentioned above."
Add: These classes effect the breakpoint and above. e.g.
```
<span class="d-md-none">Not displayed on medium and above (i.e. display on small and below)"</span>
<span class="d-none d-md-inline">Only displayed on medium and above (i.e. not on small and below)</span>
```

"Display in print"

There are no d-screen classes, so:
```
<span class="d-none d-print-inline">print only</span>
<span class="d-print-none">screen only</span>
```

3) Cards

Add: Cards have a zero margin by default, so to mimic v3's panels (see spacing utility for an explanation of mb-3)
`<div class="card mb-3"> ... </div>`

xdvarpunen:
@saturdaywalkers Basically complete example would be best. Documentation assumes too much knowledge (:

saturdaywalkers:
"[The] Documentation assumes too much knowledge" - Yes! 
More examples - Yes!

XhmikosR:
You can improve the migration page by submitting a PR.

aavmurphy:
OK, so I did 1 page (navbar). 
There'll be a learning curve for me in doing PR's, e.g. as soon as I made the PR, I saw how to link to other website pages :)
Its a little hard to test without a local copy of the bootstrap website
Added
1) a reference to (not navbar) classes used, e.g. my-2, 
2) navbar's don't print, so said this, and added a 'print-only' example.
Hopefully, this'll make other think about things that are missing.

stephaniepurvis:
I'll work on adding to the utility class documentation.

aavmurphy:
@stephaniepurvis Have you started utility/display, I was going to do that next.

aavmurphy:
Started doing this as I'm migrating a site to V4, so going through its learning curve. 

[...]

PR #24380, opened 2017-10-15
#24373 : Documentation/Navbar : Refer to utility classes used. Hidden when printing. 

aavmurphy:
refer to the  (not navbar) classes used for newbies
add: navbars don't print
give the 1st example a drop-down (as its the template people will copy and adapt)


docs/4.0/components/navbar.md
```diff
@@ -29,7 +29,10 @@ Navbars come with built-in support for a handful of sub-components. Choose from
 - `.navbar-text` for adding vertically centered strings of text.
 - `.collapse.navbar-collapse` for grouping and hiding navbar contents by a parent breakpoint.
 
-Here's an example of all the sub-components included in a responsive light-themed navbar that automatically collapses at the `lg` (large) breakpoint.
+Here's an example of all the sub-components.
+- the navbar is light-themed
+- its responsive - it automatically collapses at the `lg` (large) breakpoint.
```
XhmikosR: `It's`

mdo (changes requested):
Rather than add an entire new section to this page, let's put this under the `How it works` list.

aavmurphy:
@mdo please re-check. have added 1 more thing (fixed-top/bottom uses position:fixed), so doesn't reserve a gap on the screen. couldn't thing of a nice way to phrase it though. thanks

mdo (approved):
I pushed another commit to fine tune some things here around the language and formatting of the content—more links to utilities pages, better formatting of paragraphs and lists, etc.

Thanks for the back and forth!

aavmurphy:
Thanks!

PR #24386, opened 2017-10-16
Update package-lock.json and lock babel to 7.0.0-beta.2.

XhmikosR:
@hzoo: opened this to track the progress.


XhmikosR:
@hzoo: I locked the babel deps to 7.0.0-beta.2 so that we are on the safe side. I'll try again in the next days and see if the issue is fixed.

hzoo:
I was commenting on https://github.com/twbs/bootstrap/commit/c09f280017f69797281cd6e427df3044d4920ce8

preset-env should be on beta.3 to be in sync with everything else? same as https://github.com/babel/babel/pull/6488/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L19

XhmikosR:
@hzoo: there's no beta.3 for preset env. So everything else is using beta.3.

XhmikosR:
@hzoo: I added back the patch for beta3 so that you can see the failures on Travis.

XhmikosR:
@Johann-S: do you have any idea about the failure so that we help fix the issue upstream?

Johann-S:
it seems this error come from `babel-cli` package, 
because if we keep `babel-plugin-external-helpers` at `7.0.0-beta.3` and `babel-preset-env` at `^2.0.0-beta.2` with `babel-cli` at `7.0.0-beta.2` everything works fine but not if we set `babel-cli` at `7.0.0-beta.3`

XhmikosR:
After looking into this more I think there's something wrong with rollup or its plugins. Not sure 100% so I guess we are stuck with 7.0.0-beta.2 for the time being.

If anyone has any idea please let us know so that we at least make an issue to the appropriate repo.

Johann-S (approved):

hzoo:
Ok I believe it's a breaking change in beta.3 that the babel rollup plugin was relying on which we will fix.

XhmikosR:
Thanks @hzoo. If you have any new updates later please do let us know.

PR #24404, opened 2017-10-17
Refactor a few selectors.

XhmikosR:
1. move a couple of them before the more specific ones.
2. change nesting to be under the same parent selector
3. use the ampersand more


assets/scss/_component-examples.scss
```diff
@@ -170,31 +170,36 @@
...
+    }
+  }
+
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    margin: 0;
```
mdo: I _think_ we can change this to `margin-top: 0;` and `margin-bottom: 0;`. Can't imagine we're adding horizontal margin to any headings.

scss/_modal.scss
```diff
@@ -26,16 +26,10 @@
   // gnarly iOS Safari bug: https://bugs.webkit.org/show_bug.cgi?id=158342
   // See also https://github.com/twbs/bootstrap/issues/17695
 
-  // When fading in the modal, animate it to slide down
-  &.fade .modal-dialog {
-    @include transition($modal-transition);
-    transform: translate(0, -25%);
+  .modal-open & {
```
mdo: In general I'm not a fan of this type of nesting—usually ends up more difficult to parse in more complicated rulesets.

scss/_code.scss
```diff
@@ -31,7 +31,7 @@ kbd {
   @include border-radius($border-radius-sm);
   @include box-shadow($kbd-box-shadow);
 
-  kbd {
+  & & {
```
mdo: This is weird at first, but probably fine.

scss/_modal.scss
XhmikosR: I think we need to keep this; it doesn't complicate things a lot and along with the complexity rule we should be fine. I plan to enable that later too.

I removed the `& &` instances.

mdo (approved):

andresgalante (approved):
@XhmikosR sorry that I couldn't help much with this one. Thanks for doing it. You made a linter happy :)

XhmikosR:
@andresgalante: this isn't quite ready since I haven't enabled the rule yet. Still waiting for your patch so that we enable the rule. :)

andresgalante:
:p lol I thought it was! ok I'll continue with it

XhmikosR:
@andresgalante: I'm gonna merge this as is, and we should tackle the 2 remaining stylelint rules later.

PR #24406, opened 2017-10-17
remove unused docs color variables

m5o:
remove unused docs color variables
  * `$bd-purple-lightest`
  * `$bd-graphite-light`


XhmikosR (approved):
@mdo: They are indeed unused. LGTY?

mdo (approved):

PR #24420, opened 2017-10-18
Using the word JavaScript consistently across the page.

kevinruscoe:
This document contains the word JavaScript everywhere apart from in one place where it's abbreviated to JS. This just makes sure it's written as JavaScript consistently across the page.


XhmikosR:
Doesn't make sense to change it here only. Either way, I don't think this improves anything.

@mdo

PR #24426, opened 2017-10-19
Add vars for yiq color contrast function

mdo:
Adds two vars instead of the hardcoded colors we had previously. This supersedes #23440, getting us close to what that implementation provides without all the additional Sass map merges.

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -57,9 +57,9 @@
   $yiq: (($r * 299) + ($g * 587) + ($b * 114)) / 1000;
 
   @if ($yiq >= 150) {
-    @return #111;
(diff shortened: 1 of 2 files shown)


andresgalante (approved):

mdo:
The function determines brightness for us and makes a call on whether a `#fff` or `#111` color passed color contrast against the background. A fixed percentage doesn't work here.

danijelGombac:
@mdo 

I understand how it works, but there isn't fixed percentage. Here is a demo if you have time.

https://www.sassmeister.com/gist/efc3b9eafddadb17dd4c0b0873fd47e5

PR #24428, opened 2017-10-19
Reindent _variables.scss.

XhmikosR:


mdo:
Eh, I meant to change those. Whatever y'all prefer though.

XhmikosR:
I don't mind either, this just makes it consistent with the rest of the file.

PR #24434, opened 2017-10-19
Make 1.9.1 the minimum jQuery version again.

XhmikosR:
~~We'll just show a warning message instead.~~

~~While we don't use 2.x nor test it, a lot of people might still use it.~~

~~**Do not merge yet.** I'm still not personally sold we should do this. I do see the point of people using various CMS and not being able to update jQuery, but on the other hand this feels very inconsistent to me...~~


Johann-S (approved):
Seems fine to me :+1:

Johann-S:
It feels inconsistent but we added a warning about that so things will change :smiley: and they are warning that's at their own risks so LGTM

Johann-S (approved):

mdo (approved):

PR #24441, opened 2017-10-19
v4 Beta 2 prep

mdo:
Trying out something different given my meeting schedule today. :sob: When we're ready, we'll merge this, cut the release, and do the rest of the shipping dance.

- Updates the version from `4.0.0-beta` to `4.0.0-beta.2`
- Generates new hashes for minified CSS and JS in `_config.yml`
- Update homepage and download page to use Jekyll var for RubyGems version

/cc @twbs/team


jackmu95:
@mdo The version number got corrupted. It changed from "4.0.0-beta" to "v4.0.0-beta.2" (notice the leading "v").

mdo:
@jackmu95 Yup, pushed additional commits for that.

docs/4.0/getting-started/download.md
```diff
@@ -68,13 +68,13 @@ Bootstrap's `package.json` contains some additional metadata under the following
...
 {% highlight ruby %}
-gem 'bootstrap', '~> 4.0.0.beta'
+gem 'bootstrap', '~> {{ site.current_version }}'
 {% endhighlight %}
 
 Alternatively, if you're not using Bundler, you can install the gem by running this command:
 
 {% highlight sh %}
-gem install bootstrap -v 4.0.0.beta
+gem install bootstrap -v {{ site.current_version }}
```
glebm: The format of the rubygem version is slightly different:

```
site.current_version: 4.0.0-beta.2
     rubygem version: 4.0.0.beta2
```

`-` replaced by `.`, and no `.` between "beta" and "2".

XhmikosR: Hmm, @mdo, I'd say move the rubygem to a new _config.yml variable; this at least should make it easier to change it.

XhmikosR: Or alternatively, @glebm could change the tag starting from beta 2 and on?

glebm: The rubygem version format is correct for rubygems and is not compatible with the tag here.

We could transform it with a Jekyll helper like this (or ERB code in _config.yml if Jekyll supports that):

```ruby
version.tr('-', '.').sub(/([a-z])\.(\d+)$/, '\1\2')
```

mdo: well crap, fixing

mdo: just pushed a new var for that

XhmikosR: OK, whatever I tried didn't work even with a custom Jekyll filter to pass a regex to gsub.

So I guess, the separate variable solution should do the job; we just need to document it so that we don't miss it.

PR #24456, opened 2017-10-20
 Fix broken URI fragment and a redirected link.

XhmikosR:


Johann-S (approved):

XhmikosR:
@mdo: are we OK with that new translation link? It does look very close to the official domain...

mdo (approved):
No issues with the URL at this time. If it becomes an issue, we can revisit.

PR #24459, opened 2017-10-20
Set autoprefixer's cascade option to false.

XhmikosR:
This saves a few bytes from the unminified dist files.


mdo (approved):

sam1010:
thanks everyone, lol "seven" years ago and i'm just now getting back to it ;) between gigs for "seven years."

PR #24465, opened 2017-10-20
Adds animation to color on outlined btns

### LINKED ISSUE #24447: [V4 Beta 2] Outline button mouseover animation is jarring
For a test case, the outline buttons in the docs exhibit the issue. It appears that the font color changes immediately, while the button color transitions. The effect is jarring when compared to the smooth/coordinated transition of V4 Beta 1. See: https://getbootstrap.com/docs/4.0/components/buttons/#outline-buttons

This is on macOS 10.13 and Safari 11.0.

Can the behavior be changed back to the pleasant experience from Beta 1?

andresgalante:
This PR fixes #24447 by adding an animation to `color` on outline buttons.


XhmikosR (approved):

XhmikosR:
LGTM. Although the properties we animate are not the ones that are hardware accelerated, but I don't think it matters a lot here.

Waiting for @mdo's approval.

andresgalante:
You are right, there is a repaint. 
We can change it to animate a `transition` probably, but it'd be a larger change.

XhmikosR:
Ideally, yeah, we should only animate the hardware accelerated properties only everywhere. But it won't be an easy task. Maybe we should have an issue for this.

mdo (approved):

PR #24466, opened 2017-10-20
Replace jekyll-toc with https://github.com/allejo/jekyll-toc

XhmikosR:
Similar results, one less gem and we can directly modify it if needed.

Please do not merge yet, but feel free to try it after checking out the branch and running `npm run css-docs` locally.

Only thing I want to fix is to exclude specific classes instead of us doing `display: none`.


XhmikosR:
Closing since I don't think this could solve our current issues.

PR #24467, opened 2017-10-20
Remove `thumbnail-transition` from `.img-thumbnail`.

XhmikosR:
We don't change anything so the transition is basically unused.


andresgalante (approved):
Maybe I am missing something too, but I don't see there is another state to transition them to.

mdo (commented):
Let's add a mention to the `migration.md` page, too.

PR #24470, opened 2017-10-20
Reduce whitespace liquid outputs.

XhmikosR:
@mdo: alternatively we can remove the whole comment block.


XhmikosR:
@mdo: I added one more patch which fixes the pages' description. Previously the generic site's description was being used.

_includes/social.html
```diff
@@ -1,34 +1,20 @@
...
-  <meta property="og:url" content="{{ site.url }}{{ page.url }}">
-  <meta property="og:title" content="{{ page.title }}">
-  <meta property="og:description" content="{{ page.description }}">
-  <meta property="og:type" content="website">
-{% else %}
-  <meta property="og:url" content="{{ site.url }}">
-  <meta property="og:title" content="{{ site.title }}">
-  <meta property="og:description" content="{{ site.description }}">
-{% endif %}
+<meta property="og:url" content="{% if page.title %}{{ site.url }}{{ page.url }}{% else %}{{ site.url | append: "/" }}{% endif %}">
```
XhmikosR: This might be possible to simplify more by just using `{{ site.url }}{{ page.url }}` since for the index page

mdo (approved):

PR #24471, opened 2017-10-20
Tighten htmllint's rules.

XhmikosR:


XhmikosR:
OK, I believe this is ready now. I got htmllint confused with htmlhint...

mdo (approved):

PR #24478, opened 2017-10-20
Dismissable alert close button padding

### LINKED ISSUE #24469: Dismissible alert - long text overlaps close button
Not sure i understand why close button now has absolute positioning in Beta2, but now if text in alert is too long, it overlaps close button. I think we need to add extra right padding to dismissible alert to ensure text and button are always visible.

Image

MattJLeach:
I have increased the y padding for the alert and made the padding for the close button the same all round.

Fixes #24469

### DIFF
--- a/scss/_alert.scss
+++ b/scss/_alert.scss
@@ -32,7 +32,7 @@
     position: absolute;
     top: 0;
     right: 0;
-    padding: $alert-padding-y $alert-padding-x;
+    padding: $alert-padding-y;
     color: inherit;
   }
 }
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -709,7 +709,7 @@ $modal-transition:                  transform .3s ease-out !default;
 // Define alert colors, border radius, and padding.
 
 $alert-padding-y:                   .75rem !default;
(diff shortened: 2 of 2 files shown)


andresgalante:
Hi @MattJLeach Thanks a lot for trying to fix this.

The solution you are proposing adds a huge padding on the left or the alert. I don't think this is valid solution.

MattJLeach:
Would you prefer to have increased padding on the right and keep the left the same as it is now? The only reason I increased both was to keep the text in the centre.

andresgalante:
@MattJLeach I think the padding should go on the parent, not on the close icon, as I describe here #24469

andresgalante:
Hi @MattJLeach, I am going to close this PR in favor of #24481

Thanks a lot for taking the time to make a contribution

MattJLeach:
Sounds good @andresgalante. Thanks for the feedback on my PR. I'll keep an eye out for any other issues I can help with but comment and discuss before doing an actual PR first.

andresgalante:
Thanks man, feel free to send as many PRs as you can, it'll be my pleasure to review them.

PR #24481, opened 2017-10-20
Fix for right padding in dismissible alerts

### LINKED ISSUE #24469: Dismissible alert - long text overlaps close button

Kopyov:
Added right padding to dismissible alerts to avoid button/text overlap.

Fixes #24469

### DIFF
--- a/scss/_alert.scss
+++ b/scss/_alert.scss
@@ -27,7 +27,8 @@
 // Expand the right padding and account for the close button's positioning.
 
 .alert-dismissible {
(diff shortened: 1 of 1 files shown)


andresgalante (approved):
I think this is a good solution. There is more information on whats the result of this PR here #24469

Thanks @Kopyov

Issue #24488, opened 2017-10-21
Card headers and footers maintain border radius inside a card group


onokazu:
The issue with card header/footer border radius of cards in a card group is still present in Beta 2 although the issue initially raised (#19159) has already been closed.

Try zooming into the footer section of the 2nd card-group example on the documentation site. https://getbootstrap.com/docs/4.0/components/card/#card-groups

PR #24491, opened 2017-10-21
fix card-group border radii

gijsbotje:
in a card group the border radius of the headers and footers aren't properly reset
the image caps do have these resets
simple fix was adding the classes at the correct spot to get them to work correctly

fixes: #24488


andresgalante (approved):
This PR sets the correct border radius on header and footer for card groups and fixes the issue. Thanks a lot @gijsbotje

mdo (approved):

Issue #24490, opened 2017-10-21
ServiceWorker registration failed


ghost:
When running **`npm run docs`**, the following happens:

1. **`bundle exec jekyll build`**: generates the documentation files.
2. **`node build/workbox.js`**: which, in turn, performs the following:
2.1 Obtains the location of the **`workbox-sw`** module currently installed in `node_modules`.
2.2 Copy the main file obtained in 2.1 to `gh_pages/assets/js/vendor`. With the current version of `workbox-sw`, the path to the copied file would look like this: `gh_pages/assets/js/vendor/workbox-sw.prod.v2.1.0.js`.
2.3 Read the file `sw.js` in the repository root, replace the `{fileName}` pattern with the file name copied in 2.2 (`workbox-sw.prod.v2.1.0.js`) and save it in the `gh_pages` root.

Then, when running the documentation server, the `index.html` page includes `<script src="/assets/js/src/pwa.js">`. And this script invokes, in turn, the file `sw.js` created in point 2.3.

So far everything would be OK, but there is a problem when running the test server using **`npm run docs-serve'**. As Jekyll's documentation mentions (see https://jekyllrb.com/docs/usage/): **Destination folders are cleaned on site builds**. Then, when executing "npm run docs-serve", when the documentation files are rebuilt, all the work done in point 2 is lost. And, for this reason, the service worker stops working, throwing errors in console.

Error in Windows shell:
![error-sw-shell](https://user-images.githubusercontent.com/29401668/31850671-96e48d9e-b62c-11e7-9f70-bbb95372d402.jpg)

Error in Firefox console:
![error-sw-console](https://user-images.githubusercontent.com/29401668/31850670-96b7bb98-b62c-11e7-9e29-431ff8956f66.jpg)

I am now working on proposing a solution for this. I have to find a way for **`docs-serve`** script not to overwrite or delete files generated during **`docs-compile`** and **`postdocs-compile`** scripts.

XhmikosR:
This is known and there is no way to work around it so just ignore it. It's just a local issue.

If you can come up with a better organization of the scripts, make a PR and CC me.

lucascono:
Hello @XhmikosR. I have already solved it, shortly I send the PR. I wanted to open this issue for reference later in the PR.

XhmikosR:
Personally I don't use the `docs-serve` script at all. I just do `bundle exec jekyll s`.

[...]

PR #24492, opened 2017-10-21
Fix ServiceWorker registration failed

ghost:
Fixes: #24490 

I explain what I'm doing here:

1. I change the destination of 'workbox-sw' from **'assets/js/vendor'** to **'assets/js/sw'**. The change of destination is because I want to include the folder in **'keep_files'** key of **'_config.yml'**.

2. I include './sw.js' in the files to exclude during build, since this file only serves as a template (it should not be copied to gh_pages, as it would overwrite the correctly generated file).

Then:

1. After `docs-compile`, the workbox files are copied and the manifest is generated in gh_pages.
2. During `docs-serve`, Jekyll deletes all files, but keeps the files needed for the service worker.

error-sw-fixed

As the current Bootstrap documentation mentions, `npm run docs` must be run at least once for the service worker files to be generated.


build/workbox.js
```diff
@@ -8,10 +8,12 @@ const buildPrefix = '_gh_pages/'
 
 const workboxSWSrcPath = require.resolve('workbox-sw')
 const wbFileName = path.basename(workboxSWSrcPath)
-const workboxSWDestPath = buildPrefix + 'assets/js/vendor/' + wbFileName
+const workboxSWDestDir = path.resolve(buildPrefix + 'assets/js/sw')
+const workboxSWDestPath = path.join(workboxSWDestDir, wbFileName)
 const workboxSWSrcMapPath = `${workboxSWSrcPath}.map`
 const workboxSWDestMapPath = `${workboxSWDestPath}.map`
 
+if (!fs.existsSync(workboxSWDestDir)) fs.mkdirSync(workboxSWDestDir)
```
XhmikosR: brackets and new lines

XhmikosR:
Why can't we just keep the file in vendor?

XhmikosR:
Also, with your solution, how will the precache files be updated during the serve command?

lucascono:
Look at the changes @XhmikosR . This works both during `npm run docs-compile` and during `npm run docs-serve`.

XhmikosR:
I'm not sure I like it. Tying things more to Jekyll. Plus the Travis build doesn't even run for your PR so I'll need to check it out locally.

At any rate, I think, I'd rather skip this kind of solution and just live with the local warning, until we have a proper solution.

Issue #24496, opened 2017-10-21
Vertical button style radios/checkboxes bottom label margin not removed


tmorehouse:
- Bootstrap V4.0.0.-beta.2
- Operating system and version (any)
- Browser and version All
- Reduced test case https://codepen.io/tmorehouse/pen/BwEpOr

When using button style radios or checkboxes in vertical mode, the label margin-bottom is not set to 0 causing spacing issues between the "buttons":

![image](https://user-images.githubusercontent.com/2781561/31856557-70a4a8bc-b69a-11e7-88f0-6541980e6ef7.png)

**Expected behavior:**
That the bottom margin be set to 0 for the labels, as would be for regular buttons in a vertical button group.

**Workaround:**
```css
.btn-group-vertical label.btn {
  margin-bottom: 0;
}
```
or manually setting class `mb-0` on each `label.btn`

gijsbotje:
working on a fix :)

PR #24498, opened 2017-10-22
fix vertical input button group margin

gijsbotje:
the label in a checkbox and radio version of a button group still has margins on the bottom
this breaks the vertical version of a button group and does not match the button group styling with normal buttons

fixes: #24496


mdo (approved):

printercu:
@gijsbotje @mdo  should this be changed to use `.btn-group` scope instead of `[data-toggle="buttons"]`? This way it'll work without js behaviour too.

PR #24505, opened 2017-10-22
Add fallback border-radius to .btn

### LINKED ISSUE #24503: Variable $enable-rounded doesn't disable border-radius in buttons
I don't know if this is the expected behavior, but in Beta.2 setting `$enable-rounded` to false doesn't change the border-radius value of `.btn` classes at all, instead it just left the browser's default value.

Shouldn't BS set it to zero like it does in `.form-control`? Otherwise I'll have to override CSS classes in my application.css anyway like `.btn { border-radius: 0; }` and I believe this is not the main BS4 goal.

screen shot 2017-10-22 at 19 03 45

Thanks!

mdo:
Fixes #24503 by manually calling the border-radius instead of using the mixin.

### DIFF
--- a/scss/mixins/_buttons.scss
+++ b/scss/mixins/_buttons.scss
@@ -90,5 +90,10 @@
   padding: $padding-y $padding-x;
   font-size: $font-size;
   line-height: $line-height;
-  @include border-radius($border-radius);
+  // Manually declare to provide an override to the browser default
+  @if $enable-rounded {
+    border-radius: $border-radius;
+  } @else {
+    border-radius: 0;
+  }
 }


scss/mixins/_buttons.scss
```diff
@@ -90,5 +90,10 @@
   padding: $padding-y $padding-x;
   font-size: $font-size;
   line-height: $line-height;
-  @include border-radius($border-radius);
+  // Manually declare to provide an override to the browser default
+  @if $enable-rounded {
```
XhmikosR: Maybe use the one-liner solution?

andresgalante (commented):
Hi @mdo, Maybe I am missing something, but we are already resetting the browser default on reboot here https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L342 

I can't reproduce this issue on `v4-dev` and I can set `$enable-rounded: false` and `.btn`s go square.

brunnopleffken:
@andresgalante This block removes `border-radius` from `button` elements only, but not `input[type=button]`, `input[type=submit]` or `input[type=reset]`. If this PR is denied then these selectors should be added to _reboot.scss to keep visual consistency.

andresgalante:
Hi @brunnopleffken Thanks for the clarification.
I've tested in every browser, even IE10 and I can't reproduce this issue.

Here are the reboot lines where we remove border radius from webkit browsers:
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L382

Firefox will remove the borders as soon as we define a (transparent) border.

Opera, Edge and IE just don't show rounded corners by default.

I am sure I am missing something since @mdo and you have been able to reproduce it, but I just can't reproduce it neither in Bootstrap with `$enable-rounded: false` or on this codepen I created to test: https://codepen.io/andresgalante/pen/jGjOKZ

@mdo if you think this should be merged please go ahead and do it, it doesn't hurt to have it. I can't approve it without reproducing it.

andresgalante:
@brunnopleffken thanks a lot for testing it. Nope, I am on a mac. As I mention before it's probably me and since the code doesn't hurt to ensure we remove border radius. I am ok to merge it.

PR #24510, opened 2017-10-23
Add `.modal-dialog-centered` for optional vertically centered modal

### LINKED ISSUE #23638: v4 - Modal align vertical center
I need to leave my modal aligned to the center of the screen vertically.

I tried with Flex and with Translate, however the 2 solutions cause bizarre behavior in the modal.

Is there any native form of bootstrap v4 doing this or some workround?

Several workround's found have problem when content is large.

mdo:
Fixes #23638

### DIFF
--- a/docs/4.0/components/modal.md
+++ b/docs/4.0/components/modal.md
@@ -210,7 +210,7 @@ When modals become too long for the user's viewport or device, they scroll indep
 
 ### Vertically centered
 
-Add `.modal-dialog-centered` to `.modal-dialog` to vertically center the modal.
+Add `.modal-dialog-centered` to `.modal-dialog` to vertically center the modal. **Do not use this with long modals**—it will overflow the viewport and potentially hide parts of your modal.
 
 <div id="exampleModalCenter" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="exampleModalCenter" aria-hidden="true">
   <div class="modal-dialog modal-dialog-centered" role="document">
--- a/docs/4.0/components/modal.md
+++ b/docs/4.0/components/modal.md
@@ -213,7 +213,7 @@ When modals become too long for the user's viewport or device, they scroll indep
 Add `.modal-dialog-centered` to `.modal-dialog` to vertically center the modal.
 
 <div id="exampleModalCenter" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="exampleModalCenter" aria-hidden="true">
-  <div class="modal-dialog" role="document">
+  <div class="modal-dialog modal-dialog-centered" role="document">
     <div class="modal-content">
       <div class="modal-header">
         <h5 class="modal-title" id="exampleModalCenterTitle">Modal title</h5>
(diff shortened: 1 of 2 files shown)


andresgalante:
@mdo I've added a commit to remove margin on responsive states too to remove vertical scrolls and added the centered class to the example on the docs.
@mdo @XhmikosR Can you please remove my changes?

XhmikosR:
@andresgalante: what do you mean remove? Why don't you revert your patch?

andresgalante:
Sorry :) my patch is correct, I wrote "review" and autocorrect changed it

docs/4.0/components/modal.md
```diff
@@ -208,6 +208,64 @@ When modals become too long for the user's viewport or device, they scroll indep
 </div>
 {% endhighlight %}
 
+### Vertically centered
+
+Add `.modal-dialog-centered` to `.modal-dialog` to vertically center the modal.
+
+<div id="exampleModalCenter" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="exampleModalCenter" aria-hidden="true">
```
tmorehouse: should `aria-labeledby` be set to `"exampleModalCenterTitle"`?

tmorehouse:
One other thing that might be worth noting is that `.modal-dialog-centered` is only suitable for short modals (i.e. one or two liner content).  Tall modals will have their header cut off (unable to click the top close button, and no visible `.modal-title`), although the footer is accessible via scrolling.

andresgalante:
Hey @tmorehouse you are right, once the modal is larger than the viewport it cut off the header.

I tried to find a way out of this, removing the 100% height on`modal-dialog-centered` does it but of course it doesn't center it.

To be honest I don't think it's an option to document it as suitable for short texts, a component should be robust enough to handle anything. You solution is smart, I like it! I want to try find a way to keep our scroll. So far I failed with every attempt I went for 😄  But I'll try a bit more. 

Thanks a lot for reviewing it and finding the gaps!

tmorehouse:
Yeah, I tried trying to get that header back into the viewport, but failed miserably as well trying to keep the right hand scrollbar at the same time.  My brain is not flex-y enough to figure it out (not without out adding additional wrapper elements). 

But the scrollable modal-body option might be a "simple" (air quotes) solution (for now) without requiring extra markup. And the existing modal JS code appears to work with it well. (it just doesn't add a scroll bar on the right of the viewport)

XhmikosR:
I wonder, can't we use the old `translateY` trick?

tmorehouse:
Would `translateY` interfere with the fade animation?

XhmikosR:
I haven't had a close look into this, I was just thinking out loud, in case we missed something obvious.

sesemaya:
@andresgalante 

Wouldn't `.modal-dialog-centered { min-height: 100%; }` instead of `height: 100%;` fix the header cut off problem while still keep smaller modals vertically centred?

tmorehouse:
Another option might be javascript.  Just before modal opens, set modal-content height to 0, then measure the scrollHeight, and if less than the viewport height use the centered CSS, and if taller than the viewport, revert back to non-vertically centered (and be sure not to forget to reset height to auto before starting the fade in transition.

PR #24511, opened 2017-10-23
Clean up transitionEnd properties.

XhmikosR:
We don't support Opera 12, and Firefox's property was the unprefixed one.


Johann-S (approved):

XhmikosR:
@Johann-S: so it seems FF on OS X fails with this. I guess I'll just remove the Opera property only.

PR #24513, opened 2017-10-23
Add Util.jQuery which will detect jQuery instead of relying on global $

### LINKED ISSUE #24500: Bootstrap 4 Beta v2 JS not working with WordPress jQuery
Similar to #24417 which is closed.

When including the latest Beta v2 JS in a WordPress environment the following error is in the console: ```TypeError: Cannot read property 'fn' of undefined``` on Chrome. Safari and Firefox also fail to load, but a slightly different error.

JS Bin: https://jsbin.com/sowahazege/1/edit?html,js,console,output

I have included jQuery from a current live WP install on my server.

Swapping out the Beta V2 JS for the Beta V1 JS corrects the issue. As does loading jQuery from a CDN instead of the bundled WP version.

Has something changed between Beta v1 and v2 on how the jQuery object is accessed?

Johann-S:
Fixes #24500

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -157,7 +157,7 @@ const Util = (() => {
     },
 
     get jQuery() {
-      return window.$ || window.jQuery
+      return window.jQuery || window.$
     }
   }
 
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -154,6 +154,10 @@ const Util = (() => {
           }
         }
       }
+    },
+
+    get jQuery() {
+      return window.$ || window.jQuery
     }
   }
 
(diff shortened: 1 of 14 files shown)


js/src/util.js
```diff
@@ -154,6 +154,10 @@ const Util = (() => {
           }
         }
       }
+    },
+
+    get jQuery() {
+      return window.$ || window.jQuery
```
XhmikosR: Maybe we should change the order since `jQuery` should always be available?

Johann-S: done :wink:

XhmikosR (approved):
LGTM as long as tests pass.

Issue #24527, opened 2017-10-23
jQuery 1 noconflict still not working.


nextgenthemes:
Following up on #24417. I saw #24513 And it looks like this is exactly what was needed.

However since the dist/ files are not updated to include that fix yet. I compiled myself and pasted the minified version into [my testing bin](https://jsbin.com/wizoxehihu/1/edit?html,js,output) for this. I get the same error I got b4 with the minified version.

That is `TypeError: e is undefined` in Firefox.

Johann-S:
You're right that's a compatibility issue see my Codepen(s) : 

Using your jQuery 1.12.4 : https://codepen.io/Johann-S/pen/PJvLMK
Using jQuery 3 : https://codepen.io/Johann-S/pen/MEdxZP

Johann-S:
I was wrong in my last comments, they are something which break or JS in Wordpress 

Using jQuery 1 from jQuery CDN : https://codepen.io/Johann-S/pen/PJvLMK (working)
Using jQuery 1 from your jQuery link : https://codepen.io/Johann-S/pen/rGgbOE (not working)

PR #24530, opened 2017-10-24
Ensure our plugins can work in jQuery.noConflict mode

Johann-S:
Fixes : #24527

@nextgenthemes can you confirm it works as wanted ? 
See my CodePen which use my last changes and jQuery used by Wordpress : https://codepen.io/Johann-S/pen/PJvrgj


nextgenthemes:
Thanks, this finally did it! [Tested it here.](https://jsbin.com/fewogugona/edit?html,js,output)

It would be helpful if fresh changes like this on separate branches would be build to the `dist/` dir. Then people could use rawgit to include those versions onto their JSbins codepens ... (JSBin even refuses "to big" if I not use the minified version). Also the need to copy paste them is probably error prone and there is no clear indication what version is use exactly.

I guess you do not do this because thats not wanted in the PRs and you only build when the next version is ready? Maybe some automation for this? Just a Idea.

Johann-S:
When they are a lot of changes @mdo do a `build` commit and it happens before our next release :+1:

XhmikosR (approved):

PR #24533, opened 2017-10-24
Alerts example code: Close button before message?

PiDelport:
The close button's float-right styling requires that its markup appear before the message;
otherwise the styling breaks once the content begins wrapping.

### DIFF
--- a/docs/4.0/components/alerts.md
+++ b/docs/4.0/components/alerts.md
@@ -59,10 +59,10 @@ You can see this in action with a live demo:
 
 {% example html %}
 <div class="alert alert-warning alert-dismissible fade show" role="alert">
-  <strong>Holy guacamole!</strong> You should check in on some of those fields below.
   <button type="button" class="close" data-dismiss="alert" aria-label="Close">
     <span aria-hidden="true">&times;</span>
   </button>
+  <strong>Holy guacamole!</strong> You should check in on some of those fields below.
 </div>
 {% endexample %}
 


XhmikosR:
@andresgalante: didn't we have a fix for this already in the CSS?

andresgalante:
Hi @pjdelport thanks for the PR. You are probably running beta1, on beta2 to improve accessibility we've changed the close btn to be position absolute here: https://github.com/twbs/bootstrap/pull/23783

This is how it looks on our docs http://getbootstrap.com/docs/4.0/components/alerts/#dismissing

screen shot 2017-10-24 at 9 46 29 am

@pjdelport let me know if I am missing something

pjdelport:
@andresgalante: Ah, okay! Yeah, we're on beta 1.

Hmm, the absolute positioning approach seems to regress in that it doesn't make the text flow around the close button anymore, like the float-right approach does?

getbootstrap com_docs_4 0_components_alerts_

Should this be reported as a separate bug?

XhmikosR:
@pjdelport: with which version is that? If it's beta-2, it was fixed in https://github.com/twbs/bootstrap/commit/0a82cd9f3c44904d931614dba267295231e85f90

pjdelport:
@XhmikosR: That screenshot is from what's live right now on <https://getbootstrap.com/docs/4.0/components/alerts/#dismissing>, which appears to be `v4.0.0-beta.2`?

andresgalante:
@pjdelport the fix was made in `v4-dev` it will be available on beta 3

pjdelport:
Cool, thanks! :)

Issue #24538, opened 2017-10-24
Progress bar doesn't handle 0% case


jacroe:
This is a duplicate of #12848 but for v4. It looks like it was reintroduced.

What happens is that when it's 0%, the text is pushed off-screen and is back to sitting on the edge. Compare the same HTML with v3.3.7 (https://jsfiddle.net/pz0dj4eo/) with v4.0.0-beta.2 (https://jsfiddle.net/dy5mecan/).

This is happening on Windows with Chrome 61, Firefox 55, and Edge 38.

gijsbotje:
this is caused by replacing `text-align: center;` by `justify-content: center;`
a fix is switching these back again
i'll make a pr for it

PR #24544, opened 2017-10-25
Progress bar text

gijsbotje:
the label was cutoff at 0% due to the flex properties
this restores the `text-align: center;` and changes the flex direction and changes align-items to justify-content

fixes: #24538


andresgalante (approved):
@gijsbotje thanks!
Just a thought: although `width` works to set the width of the bar, since now the direction is set to `columns` would it be cleaner to use `flex-basis` instead?

gijsbotje:
@andresgalante i guess you can do it either way, doesn't really make a difference.
as the width or flex-basis would be set using inline styling i would maybe mention it in the documentation so people know they can do it either way
do you want me to add that to the docs?

btw: wasn't that allready possible?

andresgalante:
@gijsbotje nope, leave it as is. Not worth mentioning since it just works either way. Thanks a lot for fixing this one.

gijsbotje:
fine with me :)
so it's up to @mdo ?

mdo (approved):

PR #24545, opened 2017-10-25
Add unit tests for util.js

Johann-S:
Util.js is wild used in our plugins so we should be sure if it works as expected or not


XhmikosR (approved):
LGTM. How about testing all of our util functions?

Johann-S:
I added more unit tests, the other methods are not very relevant (_reflow_ which just return the element's offsetHeight, _supportTransitionEnd_ always false on QUnit and _triggerTransitionEnd_ which just trigger an event)

PR #24550, opened 2017-10-25
Update colors in `_root.scss`

codecalm:
Jekyll compilator breaks on this file:

```
Error in node_modules/bootstrap/scss/_root.scss:4 Invalid CSS after "...lor}: #{$value}": expected "{", was ";" 
  Liquid Exception: Invalid CSS after "...lor}: #{$value}": expected "{", was ";" in /_layouts/base.html
jekyll 3.5.2 | Error:  Invalid CSS after "...lor}: #{$value}": expected "{", was ";"
```

This fix repair this error

### DIFF
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -1,11 +1,11 @@
 :root {
   // Custom variable values only support SassScript inside `#{}`.
   @each $color, $value in $colors {
-    --#{$color}: #{$value};
+    --color-#{$color}: #{$value};
   }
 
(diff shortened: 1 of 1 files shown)


glebm:
This only happens if you use an older version of the Ruby Sass gem. Simply update Sass to v3.5.2+.

morrow95:
Also came across this problem.  Is there a manual fix for this?  I am using Koala for compiling locally so updating Sass is not an option right now.

Issue #22104, opened 2017-03-03
Responsive Display Headings


sunyang713:
I really like these larger, thinner headings that are provided:

```html
<h1 class="display-1">Display 1</h1>
<h1 class="display-2">Display 2</h1>
<h1 class="display-3">Display 3</h1>
<h1 class="display-4">Display 4</h1>
```
As shown here: https://v4-alpha.getbootstrap.com/content/typography/#display-headings

But I've find that on smaller screens, these are almost always way too large. Could you consider adding responsive breakpoints like so:

```html
<h1 class="display-l-1">Display 1</h1>
<h1 class="display-l-2">Display 2</h1>
<h1 class="display-md-3">Display 3</h1>
<h1 class="display-md-4">Display 4</h1>
```

so that the text can return to the standard h1 h2 etc font style on smaller screens?

MartijnCuppens:
This could be solved with #23816 
Result: http://martijncuppens.github.io/rfs/docs/4.0/content/typography/#display-headings

knalette:
This is still an issue. Any plans to fix?

prashaddey:
I'm still facing the same issue. They haven't fixed it yet, I guess.

knalette:
I just added this in my custom.scss, which worked for the sites I've been working on:

@media (max-width: 575.98px) {
    //Manage Display Headers for Mobile Devices
        .display-1 {
            font-size: 3rem;
            font-weight: 300;
            line-height: 1.0;
        }

        .display-2 {
            font-size: 2.75rem;
            font-weight: 300;
            line-height: 1.0;
        }

        .display-3 {
            font-size: 2.25rem;
            font-weight: 300;
            line-height: 1.0;
        }

        .display-4 {
            font-size: 1.75rem;
            font-weight: 300;
            line-height: 1.0;
        }
}

XhmikosR:
Closing this due to the Responsive Font sizes PR.

anandchakru:
am curious, why [this](https://github.com/twbs/bootstrap/issues/22104#issuecomment-387431795) is **not** incorporated into the framework itself yet?

ck9sky:
@anandchakru, kind of wonder why, seems like a good fix??

PR #24554, opened 2017-10-25
add responsive sizing

sabrown84:
Responsive Display Headings #22104

Is this what you mean @sunyang713 ?


sunyang713:
not quite, your change decreases the sizes of the headings overall. I like how large they are when I’m on a desktop, but I’d like them to be smaller on mobile.

sunyang713:
actually sorry I might be completely wrong. does ‘$display-size-base’ change with the size of the screen?

sabrown84:
Yea I did change the overall size and place a base-sizing that changes with the screen sizes.

mdo:
This doesn't make them responsive, it just makes them based on a particular variable. Responsive typography would mean we change the particular values of those variables based the viewport (e.g., `3rem` for `lg` and `2rem` for `md`).

For now, I'm going to say we pass on this change. I do think we need to improve the display headings, but probably not like this.

sabrown84:
Ok I understand. Thank you.

abdussami-tayyab:
What's the recommended way to fix this, for now?

Issue #21366, opened 2016-12-19
Dropdown: don't ignore keydown from input or textarea


gpakosz:
I'm reopening an issue for v4, as #15757 got [closed without resolution for v3](https://github.com/twbs/bootstrap/issues/15757#event-778679826).

At the time I opened the issue, [v4 was still filtering key down events for input or textarea](https://github.com/twbs/bootstrap/blob/3c12d73203bad27edcaa1072e263fdab2a2c14f7/js/src/dropdown.js#L214-L218):
```
static _dataApiKeydownHandler(event) {
  if (!/(38|40|27|32)/.test(event.which) ||
     /input|textarea/i.test(event.target.tagName)) {
    return
  }
```

In my opinion, a common use of an `<input type="text">` and dropdown combination would be displaying search results as a dropdown ala http://developer.github.com:
![image](https://cloud.githubusercontent.com/assets/553208/21317711/9c30916a-c606-11e6-8166-9e50e16ac71f.png)

In that respect, can you please make the behavior configurable (if you don't want to revert the default behavior).

Thank you.

gpakosz:
About the original issue discussion, I disagree with @fat's assertion:
> ... and most people aren't putting inputs in their dropdowns to begin with) ...

A search box with dropdown results is not a rare pattern. This is what the search box does on https://twitter.com btw.

@hnrch02 I had a look at the [original comment](https://github.com/twbs/bootstrap/issues/15757#issuecomment-85457384) but in the corresponding JSBin, clicking the input element in the `.dropdown-menu` closes the dropdown right away and I don't get a chance to hit space. So I'm not sure what it was trying to demonstrate with spaces.

As far as key down/up are concerned: imho when `<input type="text">` lives inside `.dropdown-menu`, default behavior for key down/up should be to select next/previous item. So yes I believe this would solve it.

hnrch02:
> A search box with dropdown results is not a rare pattern. This is what the search box does on https://twitter.com btw.

This pattern is not rare, I agree. However, this concerns input fields/textareas **inside** the dropdown menu itself, which I think is rather uncommon usage.

> @hnrch02 I had a look at the original comment but in the corresponding JSBin, clicking the input element in the .dropdown-menu closes the dropdown right away and I don't get a chance to hit space. So I'm not sure what it was trying to demonstrate with spaces.

[...]

PR #21802, opened 2017-01-21
Dropdown: refine handling of keydown in input and textarea

pvdlg:
Fixes #21366

This PR handle better the key inputs in `input` and `textarea` within `dropdown` and `dropdown-menu`.
As a side note, keyboard navigation is current broken and the fix is here #21535. 

**This PR will be fully working only when #21535 lands.**

Here is the new behaviors.

#### `input` and `textarea` within a `dropdown-menu`:
- **Escape key:** dropdown behavior => close the menu
- **Arrow down:** default behavior (move to the cursor to the end of the text in the `input` or `textarea`)
- **Arrow up:** default behavior => move to the cursor to the beginning of the text in the `input` or `textarea`
- **Space key:** default behavior => enter a space character in the `input` or `textarea`
- **Other key:** default behavior => enter the character in the `input` or `textarea`

#### `input` and `textarea` within a `dropdown` but outside of the `dropdown-menu`:
- **Escape key:** dropdown behavior => close the menu
- **Arrow down:** dropdown behavior => put the focus on the first/next `dropdown-item` in the menu
- **Arrow up:** dropdown behavior => put the focus on the first/previous `dropdown-item` in the menu
- **Space key:** default behavior => enter a space character in the `input` or `textarea`
- **Other key:** default behavior => enter the character in the `input` or `textarea`

PR #24557, opened 2017-10-26
Dropdown - refine handling of keydown in input and textarea thanks to @pvdlg

Johann-S:
Add changes from this PR #21802 and avoid to modify our old unit tests by adding new ones, plus it fix the errors.

Again it's thanks to @pvdlg :+1: 

Fixes : #21366

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -62,11 +62,6 @@ const Dropdown = (($) => {
   const Selector = {
     DATA_TOGGLE   : '[data-toggle="dropdown"]',
     FORM_CHILD    : '.dropdown form',
-<<<<<<< HEAD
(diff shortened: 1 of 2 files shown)


XhmikosR:
@Johann-S: why didn't you add 2 commits, one with @pvdlg's changes and the second with your fixes. That way the author would be properly attributed.

Johann-S:
because I was lazy to pull his repository and update his branch but I let him the credits by adding his name on the commit message and on my PR

XhmikosR:
Still, he should be the author. Don't get me wrong, but we need to give credit where credit's due.

I can do it if you want.

XhmikosR (approved):

PR #24571, opened 2017-10-27
Add information about the fact we do not use Popper.js for Dropdown c…

Johann-S:
Add information about the fact we do not use Popper.js for Dropdown contained in navbars

Related to this comment : https://github.com/twbs/bootstrap/pull/24092#issuecomment-339861457


XhmikosR:
Not sure my wording is right either, @mdo please confirm <3

mdo (approved):

PR #24572, opened 2017-10-27
Do not create a Popper.js instance when we don't need it

Johann-S:
Why did I change that ? Because maybe if we find a solution to just warn user instead of throwing an error a breaking our entire JS, folks will be able to use our Dropdown in a navbar without Popper.js


XhmikosR (approved):

mdo:
@Johann-S Can you resolve the conflict?

PR #24580, opened 2017-10-27
Custom select border color

sabrown84:


mdo:
This isn't the right fix I'm afraid. None of the selectors that apply the styles to the `.custom-select` use `!important` themselves, so an `!important` here isn't necessary. Something else needs to be addressed instead. Let me have a look and then I'll comment in the referenced issue so you can submit a new PR :).

Issue #23381, opened 2017-08-13
Bootstrap dropdown require Popper.js (https://popper.js.org)


zspitzer:
I'm having trouble getting started with bootstrap v4 via requirejs due to Popper

**Bootstrap dropdown require Popper.js (https://popper.js.org)**

https://jsbin.com/xululo/edit?js,console,output

this maybe related? https://github.com/FezVrasta/popper.js/issues/287

zspitzer:
other people are also having problems with this as well
https://stackoverflow.com/questions/45650583/how-to-import-popper-js

dushkostanoeski:
Have a look at this https://getbootstrap.com/docs/4.0/getting-started/webpack/

liquidvisual:
Popper.js is proving to be a massive headache for me and I don't even know why. When ever I pull the minified file into my build, it throws a console error - something about not having the  **.js.map**? Seems to work fine though.

But then when I concatenate and uglify my assets - Bootstrap's Javascript just stops working. I know it's something to do with Popper, no idea how to solve it, such a fiddly thing. Arg.

Johann-S:
I'm not an expert on how to use `requirejs` but the only thing you have to be sure it's to include Popper.js before Bootstrap and to ensure `Popper` is available in the global scope of your application

Johann-S:
Maybe this page of our documentation can help you : https://getbootstrap.com/docs/4.0/getting-started/webpack/

fran-worley:
For anyone trying (and failing/ struggling) to use Popper js with Ember Js. I got it to work by grab popper from npm, and import the umd version of the dist from node_modules (either using ember-browserify or the latest beta of ember-cli 2.15)

zspitzer:
any ideas @jrburke?

Skipvanthof:
@fran-worley Thanks, that seems to work if you're not using webpack for your project.

36864:
For laravel projects using laravel mix, `window.Popper = require('popper.js').default;` does the trick.
Same concept should apply for anyone using webpack, as that's what mix uses under the hood.

sjdeere:
i notice  that if you  npm popper.js  the  dist/popper.min.js code in that is different  from the popper.min.js  used in the the bootstrap  examples?  http://getbootstrap.com/docs/4.0/examples/jumbotron/

do we have to do something to popper.js  to   export it or something for  bootstrap?  
If i cheat and use the js  from the bootrsap  examples  my dropdown works.

[...]

PR #24601, opened 2017-10-30
Explain contents of bundled JS files in docs (#23735)

vannitotaro:
See https://github.com/twbs/bootstrap/issues/23381#issuecomment-340402261 and https://github.com/twbs/bootstrap/issues/23381#issuecomment-340403219.

### DIFF
--- a/docs/4.0/getting-started/contents.md
+++ b/docs/4.0/getting-started/contents.md
@@ -34,7 +34,7 @@ bootstrap/
     └── bootstrap.min.js
 {% endhighlight %}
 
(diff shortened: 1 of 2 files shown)


vannitotaro:
@Johann-S here it is the PR, as per your request. :wink:

XhmikosR (dismissed):
Please add the trailing slashes in the URLs

docs/4.0/getting-started/contents.md
```diff
@@ -82,6 +82,36 @@ This is the most basic form of Bootstrap: precompiled files for quick drop-in us
   </tbody>
 </table>
 
+### Comparison of JS files
```
Johann-S: Not sure if this table, isn't too much here for just two files

XhmikosR: I believe it's just to match the style we have for the CSS above.

vannitotaro: Yeah, maybe overkill but it matches the other table style. And I think it's easier to be modified, should other dependencies be added in the future.

vannitotaro:
@XhmikosR I've added trailing slashes in the URLs as per your request. Thank you. :smiley:

mdo (approved):

XhmikosR:
@mdo: you should have squashed this.

PR #24604, opened 2017-10-30
Fix bad title for our dispose method in our documentation

Johann-S:
I made a small mistake when I added `dispose` method... 

image

image


XhmikosR (approved):

XhmikosR:
These still show up with Hugo in TOC, so we'll see again then :)

PR #24606, opened 2017-10-30
Revisit vnu-jar suppressions and fix the script to work on 32-bit OS

XhmikosR:
Opening this so that we can tackle any redundant suppressions, with @patrickhlauke

### DIFF
--- a/js/tests/visual/tooltip.html
+++ b/js/tests/visual/tooltip.html
@@ -46,9 +46,6 @@
         <button type="button" class="btn btn-secondary" data-toggle="tooltip" data-html="true" title="<em>Tooltip</em> <u>with</u> <b>HTML</b>">
           Tooltip with HTML
         </button>
-        <svg width="30" height="20">
-          <circle cx="15" cy="10" r="10" fill="#62448F" data-toggle="tooltip" data-placement="top" title="Tooltip on SVG" />
-        </svg>
       </p>
       <div id="target" title="Test tooltip on transformed element"></div>
     </div>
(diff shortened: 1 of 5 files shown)


mdo (approved):

js/tests/visual/tooltip.html
```diff
@@ -46,9 +46,6 @@
         <button type="button" class="btn btn-secondary" data-toggle="tooltip" data-html="true" title="<em>Tooltip</em> <u>with</u> <b>HTML</b>">
           Tooltip with HTML
         </button>
-        <svg width="30" height="20">
```
Johann-S: This SVG is present because it's a visual unit test

XhmikosR: Which is what I had told you the other day... we need to make this work properly and not on the circle element.

Johann-S: I know but by doing that we remove a unit test previously added since a long time

XhmikosR: Yes, that is why I say, we need to make sure things work with valid stuff and not invalid, random code...

Johann-S: Can you open an issue to track that please ?

XhmikosR: I checked the git log and the original example was added to test an invalid code from the beginning.

So, I'm not sure what to say on the issue, so feel free to make one.

patrickhlauke:
a labour of love, this. thanks @XhmikosR for patiently coaching me through this collaborative work

PR #24607, opened 2017-10-30
Accessibility : manage aria-describedby attribute for feedback state

Lausselloic:
Provide a way to associate input fields with the error feedback for screen-reader

Accessibility improvement, manage aria-describedby attribute when valid / invalid state is update for an input field with feedback state

Open for discussion, don't hesitate to adapat it.


XhmikosR (changes requested):
This whole snippet should match our JS style, with spaces after keywords, no semicolons etc.

patrickhlauke:
Surprised/annoyed that I missed this new custom validation stuff (should really sit down and go over the whole of the docs again to see if there's anything else that's new but problematic).

In principle yes, agree that `aria-describedby` is needed here. Maybe also programmatically add/remove `aria-invalid="true"` ?

docs/4.0/components/forms.md
```diff
@@ -765,13 +765,37 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
+        if(inputs[i].parentElement.getElementsByClassName('invalid-feedback').length > 0
+          && inputs[i].parentElement.getElementsByClassName('invalid-feedback')[0].id) {
+          inputs[i].setAttribute('aria-describedby',inputs[i].parentElement.getElementsByClassName('invalid-feedback')[0].id);          
+        }
+      } else {
+        inputs[i].removeAttribute('aria-describedBy');
+      }
+
+      // Listen to input update, and check again the validity
+      inputs[i].addEventListener('input', function() {
```
Johann-S: so much event listeners added in case of `input` event triggered, not a good practice in terms of perf :worried: 
this should be done before the first call of `updateAriaDescribedAttribute`
with : 
```javascript
inputs.forEach(function (input) {
  input.addEventListener('input', function () {
    updateAriaDescribedAttribute([this]);
  })
})
```
and no more listeners added when `updateAriaDescribedAttribute` is called

Johann-S (changes requested):

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,43 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
+        inputs[i].removeAttribute('aria-describedBy')
+      }
+    }
+  }
 
-  window.addEventListener('load', function() {
-    var form = document.getElementById('needs-validation');
-    form.addEventListener('submit', function(event) {
+  window.addEventListener('load', () => {
+    const form = document.getElementById('needs-validation')
```
Johann-S: Not sure if we should use ES6 in our examples :thinking:

Johann-S (approved):

XhmikosR (approved):

docs/4.0/components/forms.md
```diff
@@ -700,7 +700,7 @@ While Bootstrap will apply these styles in all browsers, Internet Explorer 11 an
 Provide valuable, actionable feedback to your users with HTML5 form validation–[available in all our supported browsers](https://caniuse.com/#feat=form-validation). Choose from the browser default validation feedback, or implement custom messages with our built-in classes and starter JavaScript.
 
 {% callout warning %}
-We **highly recommend** custom validation styles as native browser defaults are not announced to screen readers.
+We **highly recommend** custom validation styles as native browser defaults are not announced to screen readers. Please ensure to manage `aria-describedby` attribute on `input` element according to the validation state.
```
patrickhlauke: i'd add a comma after "validation styles" to make it easier to read.

patrickhlauke: "ensure to" probably better changed to "ensure you" (my preference). Otherwise "make sure to manage" sounds more natural. also good place to include the `aria-invalid` thing in the prose, e.g. "Please ensure you manage the `aria-invalid` and `aria-describedby` attributes on the `input` element..."

patrickhlauke (changes requested):

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,45 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
 // Example starter JavaScript for disabling form submissions if there are invalid fields
 (function() {
   'use strict';
+  // Generic function to manage aria-describedby attribute value on input validity update
+  // @param array of inputs to check
+  // Works only if the invalid-feedback element have an ID
+  function updateAriaDescribedAttribute(inputs) {
+    for (var i = 0; i < inputs.length; i++) {
+      // Check element validity, and set aria-described attribute
```
patrickhlauke: `and set aria-describedby and aria-invalid attributes`

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,45 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
   window.addEventListener('load', function() {
-    var form = document.getElementById('needs-validation');
+    var form = document.getElementById('needs-validation')
     form.addEventListener('submit', function(event) {
       if (form.checkValidity() === false) {
-        event.preventDefault();
-        event.stopPropagation();
+        event.preventDefault()
+        event.stopPropagation()
+        // Check each form elements validity and add an aria-describedby to associate properly feedback with input
```
patrickhlauke: `Check each form element's validity and set appropriate aria-invalid / aria-describedby attributes`

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,45 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
+        var inputs = form.getElementsByTagName('input')
+        updateAriaDescribedAttribute(inputs)
```
patrickhlauke: change function name as per above comment

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,45 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
+
+        // Listen to input update, and check again the validity
+        for (var i = 0; i < inputs.length; i++) {
+          inputs[i].addEventListener('input', function () {
+            updateAriaDescribedAttribute([this])
```
patrickhlauke: ditto

docs/4.0/components/forms.md
```diff
@@ -765,17 +765,45 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
...
       if (form.checkValidity() === false) {
-        event.preventDefault();
-        event.stopPropagation();
+        event.preventDefault()
+        event.stopPropagation()
+        // Check each form elements validity and add an aria-describedby to associate properly feedback with input
+        var inputs = form.getElementsByTagName('input')
+        updateAriaDescribedAttribute(inputs)
+
+        // Listen to input update, and check again the validity
```
patrickhlauke: `and check validity again`

docs/4.0/components/forms.md
```diff
@@ -700,7 +700,7 @@ While Bootstrap will apply these styles in all browsers, Internet Explorer 11 an
 Provide valuable, actionable feedback to your users with HTML5 form validation–[available in all our supported browsers](https://caniuse.com/#feat=form-validation). Choose from the browser default validation feedback, or implement custom messages with our built-in classes and starter JavaScript.
 
 {% callout warning %}
-We **highly recommend** custom validation styles as native browser defaults are not announced to screen readers.
+We **highly recommend** custom validation styles, as native browser defaults are not announced to screen readers. Please ensure you manage the `aria-invalid` and `aria-describedby` attribute on `input` element according to the validation state.
```
patrickhlauke: may actually be good to soften this warning a bit, as some browsers now seem to work ok. suggest `We recommend using custom validation styles, as native browser validation errors are not consistently announced by all browser and screen reader combinations.`

patrickhlauke (changes requested):
Sorry, few more changes I just spotted

docs/4.0/components/forms.md
Lausselloic: yes, will finish to test that. Actually doesn't work properly with FFESR52 and NVDA 2017 even with the aria-invalid false, the browser still announce field as invalid when the user first get in the input field

Lausselloic: Wording updated. But also a bug filled to Mozilla. https://bugzilla.mozilla.org/show_bug.cgi?id=1413193
By my tests, aria-invalid status works well on Chrome and IE11, but don't on Firefox with NVDA or Jaws

mdo:
This is too much JS to casually put into a code snippet. Should this be a plugin of some sort instead?

danijelGombac:
I saw this PR and have a question about aria-describedby. What if aria-describedby is already used because form-text. Then should be changed to invalid-feedback id when is invalid and back when is valid?

patrickhlauke:
@mdo 

> This is too much JS to casually put into a code snippet. Should this be a plugin of some sort instead?

generalising this into a plugin of sorts would probably only work if we then also mandate exactly how authors code/add the extra validation/error message, or make the plugin super-configurable in terms of being able to say on each input how/where any error message is in the document in order to then add/remove the `aria-describedby` reference to it.

I'm almost tempted to say this goes beyond what Bootstrap itself would/should tackle directly...

But agree that it's quite a chunk of JS just thrown in there in passing.

patrickhlauke (approved):

patrickhlauke:
@danijelGombac 

>I saw this PR and have a question about aria-describedby. What if aria-describedby is already used because help text (form-text). Then should be changed to invalid-feedback id when is invalid and back when is valid?

`aria-describedby` takes one or more IDs (space-separated list of IDs), so if you're already using `aria-describedby` to point to additional info about the form field, then when there's a related error/validation message, you'd want BOTH the id for the additional info AND the id of the error message

tmorehouse:
@patrickhlauke on a related note about `aria-describedby`.  Popovers and Tooltips add `aria-describedby` to the trigger element when the tooltip or popover is opened (and removes it when closed), but they don't respect any existing `aria-describedby` ids that may already be on the trigger elements (they overwrite and completely remove).

patrickhlauke:
@tmorehouse could you file separate issue(s) for those?

tmorehouse:
@patrickhlauke issue #24640 has been created.

XhmikosR:
@patrickhlauke: LGTY?

XhmikosR:
Not sure what to do with this. It has conflicts.

XhmikosR:
Closing since they are conflicts. Feel free to make a new PR referencing this one.

PR #24610, opened 2017-10-30

XhmikosR:
@mdo: please make sure you don't have [such changes](https://github.com/twbs/bootstrap/commit/2fec9f13cd044cb46060ebecfb10d410c3d97c19#commitcomment-25293500) sneak into dist commits; it's hard to pinpoint any breakage.

Make sure you have the latest node.js for your OS installed. Otherwise if we keep hitting this, it should be a bug in npm.


XhmikosR:
@Johann-S: can you confirm you get the same results like I do without those fsevents entries?

mdo (approved):

Issue #22944, opened 2017-06-30
Add new grid gutter options


mdo:
With #22940 I unshipped the Sass map and code used to loop over those values. It generated the wrong code and isn't the approach I'd like to use moving forward. Looking at `.no-gutters`, I think this is the method I'd like to use—modifier classes on the row that can be customized by breakpoint.

IamManchanda:
Very Right decision @mdo... Best of luck in your progress dude :+1:

Tyutyesz:
Sorry for asking, but is this option will be available in the final release? 
BTW keep up the good work!

x-christian-romeni:
Otherwise, please explain how to add custom gutters for mobile devices.
I was searching for a way to have 5px gutter on mobile, 15 on tablet and 30 on desktop (just an example).
What is the right way foreward?

Tyutyesz:
@x-christian-romeni I made a seperate scss file with custom gutters, but not tested it, so it might brake something. Not sure if it will work fine and reading this repo, maybe there was a **good** reason why this feature has been unshipped

` $my-gutters: (
        xs: 5px,
        sm: 5px,
        md: 15px,
        lg: 30px
);
.row.custom-gutter {

  @media (min-width: map-get($grid-breakpoints, lg)) and (max-width: map-get($grid-breakpoints, xl) - 1px) {
    margin-left: map-get($my-gutters, lg) *(-1);
    margin-right: map-get($my-gutters, lg) *(-1);
  }
  @media (min-width: map-get($grid-breakpoints, md)) and (max-width: map-get($grid-breakpoints, lg) - 1px) {
    margin-left: map-get($my-gutters, md) *(-1);
    margin-right: map-get($my-gutters, md) *(-1);
  }
  margin-left: map-get($my-gutters, sm) *(-1);
  margin-right: map-get($my-gutters, sm) *(-1);
}

[class^='col'].custom-gutter,
[class*=' col'].custom-gutter {

  @media (min-width: map-get($grid-breakpoints, lg)) and (max-width: map-get($grid-breakpoints, xl) - 1px) {
    padding-left: map-get($my-gutters, lg);
    padding-right: map-get($my-gutters, lg);
  }
  @media (min-width: map-get($grid-breakpoints, md)) and (max-width: map-get($grid-breakpoints, lg) - 1px) {
    padding-left: map-get($my-gutters, md);
    padding-right: map-get($my-gutters, md);
  }
  padding-left: map-get($my-gutters, sm);
[...]

PR #24620, opened 2017-10-31
Support $grid-gutter-widths for responsive gutter widths

gloaec:
This is actually a missing feature that appears on the doc: 

Variable : `$grid-gutter-widths`
https://v4-alpha.getbootstrap.com/layout/grid/#variables

This approach aims to define different gutter widths depending on the defined breakpoints.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -191,6 +191,15 @@ $container-max-widths: (
 $grid-columns:                12 !default;
 $grid-gutter-width:           30px !default;
 
+$grid-gutter-widths: (
+  xs: $grid-gutter-width,     // 30px
+  sm: $grid-gutter-width,     // 30px
+  md: $grid-gutter-width,     // 30px
+  lg: $grid-gutter-width,     // 30px
+  xl: $grid-gutter-width      // 30px
+) !default;
+
+
 // Components
 //
 // Define common padding and border radius sizes and more.
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -25,10 +25,20 @@
       @extend %grid-column;
     }
 
+    // Allow gutter width to be different depending on breakpoints
     @include media-breakpoint-only($breakpoint, $breakpoints) {
       $gutter: map-get($gutters, $breakpoint);
 
-      .col {
+      @for $i from 1 through $columns {
+        .col#{$infix}-#{$i} {
+          padding-left: ($gutter / 2);
+          padding-right: ($gutter / 2);
+        }
+      }
+
+      .col,
+      .col#{$infix},
+      .col#{$infix}-auto {
         padding-left: ($gutter / 2);
         padding-right: ($gutter / 2);
       }
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -39,13 +39,13 @@
       .col,
       .col#{$infix},
       .col#{$infix}-auto {
-        padding-left: ($gutter / 2);
         padding-right: ($gutter / 2);
+        padding-left: ($gutter / 2);
       }
 
       .row {
-        margin-left: ($gutter / -2);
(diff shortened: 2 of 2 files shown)


gloaec:
I read from other PRs that this feature was removed because it generates to much code (~5k). Maybe there is a way to get less code using instead :

    @include media-breakpoint-only($breakpoint, $breakpoints) {
      $gutter: map-get($gutters, $breakpoint);

      [class^='col'],
      [class*=' col'] {
        padding-right: ($gutter / 2);
        padding-left: ($gutter / 2);
      }

      .row {
        margin-right: ($gutter / -2);
        margin-left: ($gutter / -2);
      }
    }

andresgalante:
Hi @gloaec thanks a lot for your contribution.

The reference to the docs you provide is for an alpha release, beta docs are now here:
http://getbootstrap.com/docs/4.0/getting-started/introduction/

As you mention, the responsive gutters were removed here: https://github.com/twbs/bootstrap/commit/b013b987b0b770434b3bc1ec61a7b9971a73aedb

If we want to prosigue with getting them back, it should also be introduce back to the docs

And although I honestly don't think this approach solves the problem, and it's missing how to handle the `make-` mixns. http://getbootstrap.com/docs/4.0/layout/grid/#mixins I do think that 
there is value is having responsive gutters.

dkozickis:
Seems that @mdo idea is to implement responsive gutter width via implementtion similar to `.no-gutters` as per this https://github.com/twbs/bootstrap/issues/22944

hrvojegolcic:
Here in documentation it still stands 
"Grid gutters now have a Sass map so you can specify specific gutter widths at each breakpoint."

https://getbootstrap.com/docs/4.0/migration/
https://getbootstrap.com/docs/4.1/migration/

I'd assume this is wrong?

gloaec:
@hrvojegolcic yup, from my knowledge, responsive gutters were definitely removed. I ended up adding custom sass in my projects (and honestly, since "css flexboxes", there's no real need for the grid system anymore). 

> I do think that there is value having responsive gutters.

@andresgalante I think it makes sense to a lot of people to have smaller gutters on smaller screens. No hard feelings if the goal is now to have the lightest library, but yeah, the docs needs to be updated ;)

Cheers !

joshwand:
Can someone please update the docs to remove this feature if it's not implemented! Went down a rabbit hole today trying to get this to work :(

hrvojegolcic:
While the solution from @gloaec works fine I have noticed that this breaks up `no-gutters` class. So in my case this worked:

``` sass
    .container,
    [class^='col'],
    [class*=' col'] {
        padding-right: floor($cust-grid-gutter-width-lg-up / 2);
        padding-left: ceil($cust-grid-gutter-width-lg-up / 2);
    }

    .row {
        margin-right: ceil($cust-grid-gutter-width-lg-up / -2);
        margin-left: floor($cust-grid-gutter-width-lg-up / -2);
    }
}

// Since we overwrite gutters in each viewports we need to repeat this class here
.no-gutters {
    margin-right: 0;
    margin-left: 0;

    > .col,
    > [class*="col-"] {
        padding-right: 0;
        padding-left: 0;
    }
}
```

lucasklaassen:
@hrvojegolcic nice! Here's the mixin version I used from your example:
```sass
@mixin responsive-gutter($size, $gutter-width) {
  @include media-breakpoint-up($size) {
    .container,
    [class^='col'],
    [class*=' col'] {
      padding-right: floor($gutter-width / 2);
      padding-left: ceil($gutter-width / 2);
    }

    .row {
      margin-right: ceil($gutter-width / -2);
      margin-left: floor($gutter-width / -2);
    }
  }
}

@include responsive-gutter(xs, 16px);
@include responsive-gutter(sm, 20px);
@include responsive-gutter(md, 28px);
```

mdo:
We're skipping on new grid gutters given our newly added negative margins. See #26825.

Issue #24632, opened 2017-11-01
z-index of custom file input


ysds:
Custom file inputs has `z-index: 5` and `6`. I'm not sure why `1` to `4` are not being used, but I think we can reduce the `z-index`.

Related code:

https://github.com/twbs/bootstrap/blob/06641ca0b3fe28021907f30c6f297d40f789f2fa/scss/_custom-forms.scss#L213-L218

https://github.com/twbs/bootstrap/blob/06641ca0b3fe28021907f30c6f297d40f789f2fa/scss/_custom-forms.scss#L236-L241

PR #24633, opened 2017-11-01
Reduce custom file z-index

ysds:
Fixes: #24632


arajay (approved):

mdo (approved):
Can't think of a single reason it'd be set to those values.

PR #24639, opened 2017-11-01
Nav Tabs use single variable to style `$nav-tabs-link-hover-border-co…

pat270:
…lor` and `$nav-tabs-link-active-border-color`

I think it's easier to customize Nav Tabs hover and active state if we can pass in the border colors through one variable instead of a few.

### DIFF
--- a/scss/_nav.scss
+++ b/scss/_nav.scss
@@ -41,7 +41,7 @@
     @include border-top-radius($nav-tabs-border-radius);
 
     @include hover-focus {
-      border-color: $nav-tabs-link-hover-border-color $nav-tabs-link-hover-border-color $nav-tabs-border-color;
+      border-color: $nav-tabs-link-hover-border-color;
     }
 
     &.disabled {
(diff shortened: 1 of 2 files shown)


andresgalante (commented):
Hi @pat270 Good point. I understand why mdo wanted to hide the option to change the border bottom, but sometimes users want to change the bottom border to be for example white and make the tab look open on hover. This PR gives more control to the user.

Would it make it easier to understand if we make the variables:

```scss
$nav-tabs-link-hover-border-color: $gray-200;
$nav-tabs-link-hover-border-bottom-color: $nav-tabs-border-color;
```

@mdo @pat270 What do you think?

pat270:
Hey @andresgalante, I still favor the single variable because designers tend to change every aspect of a component. Personally, I don't think I will encounter many situations where I will only need to change the `border-bottom-color`, but I'm good with whatever you guys decide. Either fix makes it a lot easier to style via variables. Thanks!

andresgalante (approved):
@pat270 I agree, either way is good. Unless @mdo thinks a border bottom variable is better, I am ok to merge it this way.
Thanks a lot 😄

mdo (approved):
I can dig it

PR #24644, opened 2017-11-02
update README, use next tag for npm, yarn installation

sibiraj-s:
use next tag to always install the latest version of bootstrap published to the registry

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 <p align="center">
   <a href="https://getbootstrap.com/">
-    <img src="https://getbootstrap.com/assets/brand/bootstrap-solid.svg" alt="" width=72 height=72>
+    <img src="https://getbootstrap.com/assets/brand/bootstrap-solid.svg" alt="Bootstrap" width=150 height=150>
   </a>
-
   <h3 align="center">Bootstrap</h3>
+  <h5 align="center">(v4.0.0-beta.2)</h5>
 
   <p align="center">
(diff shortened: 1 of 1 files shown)


Zain3:
Mute the thread

mdo:
I feel like I'd rather be specific in our versions. Thoughts @XhmikosR?

XhmikosR:
Yup, agreed.

README.md
```diff
@@ -40,8 +40,8 @@ Several quick start options are available:
 
 - [Download the latest release.](https://github.com/twbs/bootstrap/archive/v4.0.0-beta.2.zip)
```
bkeeton12: Download

bkeeton12 (approved):

Sibiraj-S:
I felt the same @mdo when I used to see the versions get updated manually. 

But this will be those installations, however, points to latest(beta) one. then why not make use of npm's @next tag? I am not sure how the composer works though.

Sibiraj-S:
something like in the last [commit](https://github.com/twbs/bootstrap/pull/24644/commits/e71df55b378ab4739b75ba04f552a13882a9acbc).

but if it is GTG with the current format. it is A-Ok then

XhmikosR:
Thanks, but we'd rather be explicit currently. Remember, v4-dev will be merged to master eventually.

Sibiraj-S:
Cool then

Issue #24638, opened 2017-11-01
table-responsive class does not take full width


davidism:
This was previously reported in #24256, which was closed after #24438 was merged. However, that merge was only documentation, and the issue doesn't actually appear to be fixed in Beta 2.

A table with the `table-responsive` class will not render at full width. I'm not sure why it works in the example in the docs, but this [codepen](https://codepen.io/anon/pen/EbVvRb) demonstrates that the issue still persists.

```html
<table class="table table-responsive">
  <thead>
    <tr>
      <th>Name</th>
      <th>Description</th>
      <th>Link</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>David</td>
      <td>Developer</td>
      <td>None</td>
    </tr>
  </tbody>
</table>
```

![screenshot_2017-11-01_12-00-12](https://user-images.githubusercontent.com/1242887/32292098-3ddc184c-befc-11e7-8824-7cff68857ebc.png)

![screenshot_2017-11-01_12-01-01](https://user-images.githubusercontent.com/1242887/32292134-5985d3c6-befc-11e7-9d55-c05841d8b2bb.png)

The issue occurs on both Firefox 58 and Chrome 62.

browner12:
simple solution, use a wrapping `<div>` with the `table-responsive` class.

```html
<div class="table-responsive">
    <table class="table"></table>
</div>
```

davidism:
OK, but why doesn't the documentation example need that?

jipexu:
this is the display block in .table-responsive  ...

tmorehouse:
In the documentation examples, the reason it appears to "work" is that there is enough content in the table to force it to full width inside the container..

jipexu:
true

matthewmalone:
+1 to get the documentation updated, please!

FTWinston:
Is it definitely the intention that the table-responsive class _not_ be put on a table any more?

Or is that just a workaround? Cos it seems a bit weird.

shtse8:
It seems we should have to wait for the next release.

tmorehouse:
One other caveat with placing `table-responsive` on the table element  is for screen readers.

When a `<table>` has a CSS `display` of any value other than `table`, the table is no longer presented to screen readers as being a table structure, and requires placing an explicit `role` on each element within the table (i.e. `table`, `rowgroup`, `row` and 'cell`)

mdo:
[...]

PR #24645, opened 2017-11-02
fix: block display of tables

sibiraj-s:
I am really not sure of this. But is it necessary for a table to have block property for tables?

As mentioned here in [docs](https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables)

```
<table class="table table-responsive">
  ...
</table>
```

`display: block` seems to override the table property


Sibiraj-S:
closed as it is already open in #24638.

PR #24654, opened 2017-11-03
Expand bg-* classes with $colors and $grays elements.

### LINKED ISSUE #24655: Expansion for bg- classes
When reading the docs for color utils (http://getbootstrap.com/docs/4.0/utilities/colors/#background-color) I expected the bg- classes would be generated for all colors, not just for theme-colors.

Is it useful to add those classes or is this there a reason for this decision?

I added a pull requst for explanation:  #24654

sisao:
Fixes #24655

### DIFF
--- a/scss/utilities/_background.scss
+++ b/scss/utilities/_background.scss
@@ -1,5 +1,13 @@
 // stylelint-disable declaration-no-important
 
+@each $color, $value in $grays {
+  @include bg-variant(".bg-gray-#{$color}", $value);
+}
+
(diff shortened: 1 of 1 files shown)


andresgalante:
Hi @sisao thanks for you contribution.

I don't think we need to add all those options, we have the theme map to scope the amount of colors. If we want to add more colors we should add it to "theme" and not loop grays and colors

mdo:
We'll pass on these classes for now, thanks though!

PR #24660, opened 2017-11-03
WIP: Simplifies and improves table backgrounds

### LINKED ISSUE #24529: Inconsistant .table-active color

I would simply use:
> .table-active

andresgalante:
This PR is needs more clean up and better organization, but bare with me.

The way we use selectors for background on tables doesn't allow to have a different colors within a row, for example:
```html
    <tr class="table-danger">
        <th scope="row">9</th>
        <td>Column content</td>
        <td class="table-warning">Column content</td>
        <td>Column content</td>
      </tr>
```
It also doesn't allow to mark something as active like this:
```html
      <tr class="table-danger">
        <th scope="row">9</th>
        <td>Column content</td>
        <td class="table-active">Column content</td>
        <td>Column content</td>
      </tr>
```
or to mark something as active within an active row like this:
```html
      <tr class="table-active">
        <th scope="row">9</th>
        <td>Column content</td>
        <td class="table-active">Column content</td>
        <td>Column content</td>
      </tr>
```
... and it produces issues like the one described on #24529

By simplifying the selectors it resolves it and I don't see conflicts with nested tables or `.table-striped`. I tried to digg up the history of this selector but I couldn't find the root.

@mdo @XhmikosR Can you please point me out what I am missing so I can continue with this PR?

Thanks a lot!

This is a WIP but when it's done it will close #24529


XhmikosR:
I'm definitely in favor of this, but I always thought there was some specificity reason that led to the current solution. Only @mdo can confirm :)

mdo:
I think the move from Less to Sass is what broke these selectors. I can't recall any other background on why it's coded the way it is, so I'm definitely open to changing it.

mdo:
Ah, just read the deleted comment again—these don't work with `.table-striped` now.

screen shot 2017-11-06 at 2 02 47 pm

andresgalante:
@mdo right! I'll work on it

andresgalante:
@mdo This is a horrible solution, but it fixes the problem.

At the moment a stripe table will overide active states:
screen shot 2017-11-07 at 12 11 15 pm

And it has all the problems described above.

This horrible selector is needed to get our way around tables when `table-stripe` is on.

@XhmikosR @mdo can you please take another spin on this one and see what I am missing, or if we can improve the selector

scss/_tables.scss
```diff
@@ -70,6 +70,13 @@
 .table-striped {
   tbody tr:nth-of-type(odd) {
     background-color: $table-accent-bg;
+    /* stylelint-disable */
```
XhmikosR: @andresgalante: please be more explicit with what you disable here.

scss/_tables.scss
```diff
@@ -139,6 +144,13 @@
   &.table-striped {
     tbody tr:nth-of-type(odd) {
       background-color: $table-dark-accent-bg;
+      /* stylelint-disable */
```
XhmikosR: ...and here.

mdo:
Quick update from me—I've been trying to look into this here and there to see what else can be done with our selectors. Still trying, but not a top priority for me or block to v4 stable right now.

mdo:
Closing as stale and as something to revisit in v5.

PR #24665, opened 2017-11-03
Object spread : less jQuery more ES6

Johann-S:
This PR will remove the use of `$.extend` in favor of this awesome ES6 feature : https://babeljs.io/docs/plugins/transform-object-rest-spread/ which will do the same as `$.extend` but without jQuery :smile: 

Currently I only made the change for our Tooltips but if everybody is ok I'll do the same for our other plugins :+1:

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -10,6 +10,7 @@ module.exports = {
     ]
   ],
   plugins: [
-    process.env.PLUGINS && 'transform-es2015-modules-strip'
+    process.env.PLUGINS && 'transform-es2015-modules-strip',
+    '@babel/proposal-object-rest-spread'
   ].filter(Boolean)
 };
--- a/build/rollup.config.js
+++ b/build/rollup.config.js
@@ -15,7 +15,8 @@ const plugins = [
     externalHelpersWhitelist: [ // include only required helpers
       'defineProperties',
       'createClass',
-      'inheritsLoose'
+      'inheritsLoose',
+      'extends'
     ]
   })
 ]
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -223,7 +223,10 @@ const Carousel = (($) => {
     // private
 
     _getConfig(config) {
-      config = $.extend({}, Default, config)
+      config = {
+        ...Default,
+        ...config
+      }
       Util.typeCheckConfig(NAME, config, DefaultType)
       return config
     }
@@ -428,10 +431,16 @@ const Carousel = (($) => {
     static _jQueryInterface(config) {
       return this.each(function () {
         let data      = $(this).data(DATA_KEY)
-        const _config = $.extend({}, Default, $(this).data())
+        let _config = {
+          ...Default,
+          ...$(this).data()
+        }
 
         if (typeof config === 'object') {
-          $.extend(_config, config)
+          _config = {
+            ..._config,
+            ...config
+          }
         }
 
         const action = typeof config === 'string' ? config : _config.slide
@@ -468,7 +477,10 @@ const Carousel = (($) => {
         return
       }
 
-      const config     = $.extend({}, $(target).data(), $(this).data())
+      const config = {
+        ...$(target).data(),
+        ...$(this).data()
+      }
       const slideIndex = this.getAttribute('data-slide-to')
 
       if (slideIndex) {
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -277,7 +277,10 @@ const Collapse = (($) => {
     // private
 
     _getConfig(config) {
-      config = $.extend({}, Default, config)
+      config = {
+        ...Default,
+        ...config
+      }
       config.toggle = Boolean(config.toggle) // coerce string values
       Util.typeCheckConfig(NAME, config, DefaultType)
       return config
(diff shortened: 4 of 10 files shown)


package.json
```diff
@@ -85,6 +85,7 @@
     "babel-eslint": "^8.0.0",
     "babel-plugin-external-helpers": "7.0.0-beta.2",
     "babel-plugin-transform-es2015-modules-strip": "^0.1.1",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
```
XhmikosR: Why not 7.0.0-beta.2 like the rest?

BTW I see now there's 7.0.0-beta.5 out, so I'll give it another go in another branch.

Johann-S: I didn't choose the version number, I did : `npm i -D babel-plugin-transform-object-rest-spread` and that's all :smile:

Johann-S: It seems it's the latest release : https://www.npmjs.com/package/babel-plugin-transform-object-rest-spread

XhmikosR: It's the latest stable. For every other plugin we are using the 7.x versions so we should do the same here.

XhmikosR: I'm having trouble updating to beta.5 anyway. I think we'll need to bug @hzoo so that we have all the plugins in sync...

XhmikosR:
Regarding the changes, LGTM, so yeah, I'd say go ahead and make the changes in the rest of the code.

package.json
Johann-S: I found the 7.x release of this plugin :+1:

package.json
```diff
@@ -80,6 +80,7 @@
     "popper.js": "^1.12.6"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.2",
```
hzoo: Uhh you should make sure you are using the same version plugin and without the `^`?

`"babel-plugin-transform-object-rest-spread": "7.0.0-beta.2",` (or just update everything to scoped)

XhmikosR: @hzoo: can you please make a PR against v4-dev to update all babel packages to the latest versions and CC me? I tried with scoped packages and 7.0.0-beta.5 but then I got some error in ESLint.

I'd like to pinpoint any potential issues and report them upstream so that they get fixed.

Thanks!

XhmikosR: @Johann-S: for this branch just remove `^` like @hzoo says above and like we do with the other babel plugins already.

Johann-S: done 👍

Johann-S:
It seems they are no beta 2 for this plugin @XhmikosR see : https://libraries.io/npm/@babel%2Fplugin-proposal-object-rest-spread
Which one should I choose ?

XhmikosR:
beta.5? I still hope @hzoo will help us update to the latest so it should be OK if the transpiled code is OK :)

XhmikosR:
@Johann-S: rebased. I don't get any changes in package-lock.json; is something wrong? Maybe we don't need the extra package?

hzoo:
@XhmikosR oh yeah this is because we added a new option in preset-env in v7 called `shippedProposals` https://github.com/babel/babel/tree/master/experimental/babel-preset-env#shippedproposals

anything that's Stage 3 in browsers including object rest spread

Johann-S:
I removed the extra package and added `shippedProposals: true` in our `.babelrc.js`
is it what you suggested to us @hzoo ?

hzoo:
If you add `shippedProposals` then you don't need the `   '@babel/proposal-object-rest-spread'` in the plugins. Although if you are **only** going to use object rest spread I would just explicitly add that as a plugin and not use `shippedProposals`, just depends on what is easier to understand

Johann-S:
I think it's ok to use `shippedProposals` because it will only add what we need on our dist files so as we don't use the other feature that's ok because no extra size will be added

hzoo:
Yeah totally I agree, it's just that it allows someone to use new syntax and it won't error (but I guess since you commit the dist files you will see the change anyway then 👍)

Johann-S:
@XhmikosR ping

XhmikosR (approved):

Johann-S:
Special thanks to @hzoo for his awesome feedbacks :+1: and thank you @XhmikosR as always :+1:

los93sol:
Does the move to ES6 mean IE11 is no longer compatible?

XhmikosR:
Everything is currently transpiled via Babel so it works just fine for our
supported browsers.

pfx2:
Hello, after upgrading from bootstrap beta 2 to bootstrap 4 final, I can't compile with webpack anymore (using Encore from Symfony)

```
Error  in ./node_modules/bootstrap/js/src/tooltip.js

Module parse failed: Unexpected token (497:10)
You may need an appropriate loader to handle this file type.
|       if (this.config.selector) {
|         this.config = {
|           ...this.config,
|           trigger: 'manual',
|           selector: ''
```

Do I need to do something ?

Johann-S:
It's uncommon to compile an external lib in a different project, so you should avoid to do that, if you need that please take a look to our `package.json` to see which babel packages we use

pfx2:
Sorry compile is the wrong word, I'm just importing tooltip in my project's main JS file.
`require('bootstrap/js/src/tooltip.js');`

When I launch "yarn run encore" (encore is the Symfony wrapper for webpack), the above error happens. With bootstrap beta 2 it works.

I'll check the package.json.

Issue #24673, opened 2017-11-04
Bootstrap's font-family style reduces SVG performance in Google Chrome


Herst:
In Google Chrome there seems to be an issue where the value of the `font-family` has an effect on the performance when scaling `<text>` elements inside a SVG. It seems like the values are parsed anew every time the effective font size changes. With Bootstrap 4 the value of `font-family` is `-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol"` and this had a noticeable negative impact on a SVG interaction heavy web application which I ported from Bootstrap 3 to Bootstrap 4

It could be that the effect does not show on macOS, since the `-apple-system` generic font family matches right away, while on Windows/Linux/Android several font families need to be dismissed before arriving at one which exists.

See <https://crbug.com/781344> for details. I would also welcome contributions to the bug report which would help in getting it confirmed and fixed. I reproduced it on Desktop Linux and Windows for now, no idea about macOS or Android.

Setting the `font-family` style of `<svg>` elements to something simple seems to work.

PR #24674, opened 2017-11-04
Separate font-family setting for SVG (#24673)

Herst:
Workaround for #24673. See <https://crbug.com/781344> for details.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -276,6 +276,9 @@ img {
 
 svg:not(:root) {
   overflow: hidden; // Hide the overflow in IE
+  // Workaround Chrome performance issue,
+  // see: https://github.com/twbs/bootstrap/issues/24673
+  font-family: $font-family-svg;
 }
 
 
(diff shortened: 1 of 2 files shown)


XhmikosR:
I'm not sure we should do this. I mean, it's clearly a browser bug, isn't it? If the bug report is closed for whatever reason without a fix, then it would make sense to do this.

Herst:
Fair enough and in any case the users making use of SVGs with a lot of text it in can add such a fix to their CSS.

That said, should this not be fixed soon though then it might be worth adding a note about it to the documentation. Others might be surprised like I with by a sudden performance drop in their application after switching to Bootstrap 4. I thought at first that Bootstrap's new button styling, flex box or `calc()` usage might be to blame.

Herst:
@mdo, @cvrebert should you decline this patch, would <https://crbug.com/781344> be a possible candidate for the Wall of Bugs and [the Bootstrap meta bug](https://crbug.com/536263)?

patrickhlauke:
i'd say yes to the wall of bugs

Herst:
Closed in favor of #24751.

Still the workaround is worth to keep in the back of one's head if dealing with dynamic animated/interactive SVG content because even if a fix were to land in three weeks then it's not going to reach general Chrome users until March 2018.

PR #24682, opened 2017-11-05
Allow users to customize focus and active button states.

flachware:
This enhancement of the buttons mixin allows the user the following adjustments in `_variables.scss`:

- Adjust the percentages of the `:focus` and `:active` state color function
- Either 'lighten' or 'darken' the states by using a positive or negative value
- Mix 'lighten' and 'darken' states

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -357,7 +357,11 @@ $btn-font-weight:             $font-weight-normal !default;
 $btn-box-shadow:              inset 0 1px 0 rgba($white, .15), 0 1px 1px rgba($black, .075) !default;
 $btn-focus-width:             $input-btn-focus-width !default;
 $btn-focus-box-shadow:        $input-btn-focus-box-shadow !default;
+$btn-focus-bg-lightness:      -7.5% !default;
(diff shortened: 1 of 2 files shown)


andresgalante:
Hi @johanneskrtek Thanks for sending this PR and giving it another shot.

To me having variables that get inserted into mixins are hard to understand.

@mdo mentioned that the focus stays as is, and I agree with him, I don't see the need to change it. But I'll let him decide on this PR.

johanneskrtek:
Hi @andresgalante, thanks for your review!

We could also pass `$btn-focus-bg-lightness` etc. as regular arguments to the mixin and do the color manipulation later, if that helps. This way we would not have to rely on global variables. Does that sound better to you?

If you think the focus and hover states should not be configurable at all, well, that’s another story. Imho users will either patch the Bootstrap mixin code then or move the button/button mixin code into their own source code and run the mixin with modified values there (in case they need to modify the states).

mdo:
Going to punt on this for now. We'll wait for more feedback on the mixins before adding all the params.

sekmo:
I think it would be a great addition! What's the current state? :-)

PR #24686, opened 2017-11-06
Switch to cssnano.

XhmikosR:
OK, this is something I've been meaning to try out for a long time.

With the first patch, results are pretty much the same except for CSS  pseudo elements reduced to pseudo classes like `::after` -> `:after`. Browsers shouldn't have any issue with that.

The second patch is using the default preset which does selector merging (should be OK), transforms reducing (`translate3d` to `transform{x,y,z)` if shorter; this one we need to verify), completely removes quotes from font names (this should be OK based on the specs, although traditionally we use quotes for spaces in names).

The gain isn't huge, especially if we go with the first patch. Still the fact that the sourcemaps are OS independent is one plus currently compared to clean-css.

/C @ben-eb for any thoughts/suggestions.

PS. The reason I went with the postcss plugin solution is that there is no v4.x of the cssnano available yet.


wolfy1339:
I don't think `:after` vs `::after`
 change is a good one in my opinion, users had requested it be changed in v4 to comply with specs

XhmikosR:
That's just in the minified file which isn't supposed to be readable. That
being said you are welcome to make an issue upstream to make it
configurable.

andresgalante:
@XhmikosR I wouldn't be able to review this PR, but I must say that I had excellent experiences with css nano on my projects. Good move, thanks a lot 😄

mdo:
> That's just in the minified file which isn't supposed to be readable. That being said you are welcome to make an issue upstream to make it configurable.

It's not about readability though—it's about adhering to spec (which I admit I don't always do) and showing the differences between pseudo classes and pseudo elements.

mdo:
> completely removes quotes from font names (this should be OK based on the specs, although traditionally we use quotes for spaces in names)

Also this might cause trouble... I don't know for sure if browsers are all okay with no quotes around names with spaces in them.

XhmikosR:
https://mathiasbynens.be/notes/unquoted-font-family

We can change this if we want. I also think we might be better disabling selector merging since it seems to hurt gzip compression.

I'll make a few more tests with this in the next days and come up with the optimal config for our use case.

XhmikosR:
>It's not about readability though—it's about adhering to spec (which I admit I don't always do) and showing the differences between pseudo classes and pseudo elements.

Like I said, we shouldn't care about this in the minified files. Do you care about .min.js files? No, as long as it works.

That being said, we can ask @ben-eb for an option to disable this optimization.

wolfy1339:
That was already requested, and was not accepted
Ref: ben-eb/cssnano#24

PR #24691, opened 2017-11-06
Display Headings - add variable and breakpoints

stevenkaspar:
- Add breakpoint mappings for display-x
- Use $display-headings map to allow for extending like $theme-colors


mdo:
Thanks, but I don't see much value in having a nested Sass map for these right now. I agree they should be responsive, but not in the sense that we have modifiers for each grid breakpoint—more so that they each adapt to the relative viewport size.

PR #24696, opened 2017-11-07
Clean up the reset code for button's rounded corners in button group and input button

ysds:
A `.dropdown-toggle` selector is used to reset a button's rounded corner in button group and input button. But should not use a `dropdown-toggle` selector to reset it, because `dropdown-toggle` does not have rounded corners by default.

This PR including the follwings:
- Use `last-of-type` selector instead of `.dropdown-toggle`
- Remove `border-raduis :0` because it is enough by each reset of the left and right side

I hope that this PR will help improve code. Thanks.

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -54,8 +54,7 @@
     float: left;
   }
 
-  // Reset rounded corners individual because sometimes a single button can be
-  // in a .btn-group and we need :first-child and :last-child to both match.
+  // Reset rounded corners
   > .btn:not(:last-child):not(.dropdown-toggle),
   > .btn-group:not(:last-child) > .btn {
     @include border-right-radius(0);
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -44,46 +44,27 @@
 }
 
 .btn-group {
-  > .btn:not(:first-child):not(:last-child):not(.dropdown-toggle) {
-    border-radius: 0;
-  }
-
-  // Set corners individual because sometimes a single button can be in a .btn-group
-  // and we need :first-child and :last-child to both match
   > .btn:first-child {
     margin-left: 0;
-
-    &:not(:last-child):not(.dropdown-toggle) {
-      @include border-right-radius(0);
-    }
   }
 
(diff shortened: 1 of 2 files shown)


andresgalante:
Hi @ysds Thanks for the PR. I found that there is still a rounded corner on the last btn before the drop menu. Look a the first example under "nesting":

screen shot 2017-11-28 at 8 57 34 am

It's really challenging to remove or simplify code, thanks a lot for giving it a try.

ysds:
Hi @andresgalante Thank you for your review! and sorry for my insufficient check. Using `last-of-type` instead of `.dropdown-toggle` seems does not seem to be appropriate.
I modified and restored some selector, please double check it.

andresgalante (approved):
Thanks a lot @ysds, I have tested and review the code and I it looks ok to me.

But in case I am missing something I'll assign this PR to @mdo to take another look. 

The comment on line 57 is confusing to me, maybe we can improve the language.

scss/_button-group.scss
```diff
@@ -44,46 +44,25 @@
...
-  > .btn:last-child:not(:first-child),
-  > .dropdown-toggle:not(:first-child) {
-    @include border-left-radius(0);
   }
 
-  // Custom edits for including btn-groups within btn-groups (useful for including
+  // Alignment for including btn-groups within btn-groups (useful for including
   // dropdown buttons within a btn-group)
   > .btn-group {
     float: left;
```
andresgalante: Do we actually need this float here? What is it for?

ysds:
@andresgalante 

I updated the comment more concisely.

> Do we actually need this float here? What is it for?

I think so too. it will not be needed. and there seems to be some other code that can be deleted/clean-up.

e.g: L47-
```scss
.btn-group {
  > .btn:first-child {
    margin-left: 0;
  }
```
and more. I can additional cleanup with another PR if this PR is approved. Thanks.

mdo:
Changes to the input group here will need to be removed once #25020 lands.

ysds:
Please follow #25073 😄

PR #24697, opened 2017-11-07
Rewrite custom form check backgrounds

### LINKED ISSUE #24598: Custom Radio and Checkbox broken when $enable-gradients is true
unbenannt-2

mdo:
Previously, this was all just a `background-color` and `background-image`. When we restored the gradients though, we had two `background-image`s competing on the same element, causing rendering glitches. This refactors that code, creating a mixin to simplify things, so we can we easily use two `background-image`s (SVG icon and gradient) when `$enable-gradients` is set to true.

Fixes #24598.

---

#### Todo
- [x] Remove variable changes
- [x] Double check validation states of custom checks


jonnysp:
@mdo thanks looks good. maybe change the validation status?

scss/_variables.scss
```diff
@@ -98,8 +98,8 @@ $yiq-text-light: $white !default;
 
 $enable-caret:              true !default;
 $enable-rounded:            true !default;
-$enable-shadows:            false !default;
-$enable-gradients:          false !default;
+$enable-shadows:            true !default;
```
andresgalante: Do we want to leave shadows and gradients on by default?

mdo: Nope! Not yet at least :).

mdo:
Okay, decently sized update here:

I've had to change how we apply the `background-color` and multiple `background-image`s for custom checkboxes and radios. Previously, we had multiple comma-separated `background-image`s on the same `.custom-control-indicator` element. This presents a problem though for when you want to customize the background color or gradient without having to redeclare the icon, too. Now, the gradient and color are on `.custom-control-indicator` and the icon is on `.custom-control-indicator::before`.

I'll be adding a note to the migration docs for this here as well.

mdo:
@jonnysp With the changes I mentioned in my previous comment, you should be good to go with green and red background gradients on those custom forms.

jonnysp:
looks great, many thanks. can merge!

jonnysp (approved):
looks great, many thanks. can merge!

Issue #23058, opened 2017-07-12
Move .custom-select styles to select.form-control


pat270:
Is it possible to move `.custom-select` styles to `select.form-control`? As it stands, it's only really useful if you don't plan on changing the border-radius on form-controls. By adding the `form-control` class to the select element we are already committing to Bootstrap's contract to style the element, doesn't make sense why we would only have partial control.

Edit: Also, `.custom-select` is the only custom element that doesn't need an extra element to achieve custom styles.

mdo:
Let's slate this for a follow up minor release; we can deprecate it later.

alecpl:
That would be nice. Now .custom-select-* does not play nice with .form-control. And if I'd want to override some general controls style it would be simpler and smaller to do this for .form-control only.

romieleo:
Pada tanggal Sel, 22 Mei 2018 15.40, Aleksander Machniak <
notifications@github.com> menulis:

mdo:
This was fixed in #28450 for v5.

PR #24699, opened 2017-11-07
Custom select updates

### LINKED ISSUE #20316: .c-select-sm and .c-select-lg missing

### LINKED ISSUE #23823: Add a mention and/or example of .custom-select-sm to Custom Select section
This would be a trivial update, but: can the [Select Menu](https://getbootstrap.com/docs/4.0/components/forms/#select-menu) section of the custom form controls be updated to include a mention (or example) showing that there is a sizing class for a smaller version? My last attempt to issue a pull-request against the docs directory was auto-closed, or I'd just add it myself...

I could not find a mention of `.custom-select-sm` anywhere on that page, so I had to check the bootstrap.css file to find that there was, in fact, a sizing option.

**On a related tangent:** is there a reason not to also have a `-lg` version, or am I simply missing something?

### LINKED ISSUE #24504: Custom-select doesn't support single select with size attribute set above 1
- Bootstrap V4.0.0.-beta.2
- Operating system and version (any)
- Browser and version All
- Reduced test case https://codepen.io/tmorehouse/pen/boJQLL

The `size` attribute controls the number of visible "rows" the select displays. It is commonly used in conjunction with the `multiple` attribute, but can be used on single selects to produce a "list-box" style select (except on some mobile browsers).

When used on a `.custom-select`, not in `multiple` mode, the `size` attribute is not respected and only  single option (and bits of the next option) are shown (as well as bits of the custom select indicator):

image

As an aside note, when in multiple mode, there is also excess padding on the right (where the custom indicator would be on a single select)

Workaround CSS:
```css
.custom-select[multiple],
.custom-select[size]:not([size="1"])
{
    height: auto;
    background-image: none;
    /* Fix for excess right padding */
    padding-right: 0.75rem;
}
```

### LINKED ISSUE #24597: Custom Select Border and Custom File Browser when active or focus not displaying properly
only the normal 
the validation variant works

testet on FF and Chrome win 10

unbenannt-1

mdo:
Few fixes in one PR coming here:

- Fixes #24504: Add support for size attribute on custom selects.
- Fixes #20316 and fixes #23823:  Add large custom select, document it and the small variant.
- Fixes #24597 where custom file inputs had an inconsistent focus shadow.

Also need to look into #23058.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -149,7 +149,7 @@
   &:focus {
     border-color: $custom-select-focus-border-color;
     outline: 0;
-    @include box-shadow($custom-select-focus-box-shadow);
+    box-shadow: $custom-select-focus-box-shadow;
 
     &::-ms-value {
       // For visual consistency with other platforms/browsers,
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -216,7 +216,12 @@
   opacity: 0;
 
   &:focus ~ .custom-file-control {
+    border-color: $custom-file-focus-border-color;
     box-shadow: $custom-file-focus-box-shadow;
+
+    &::before {
+      border-color: $custom-file-focus-border-color;
+    }
   }
 }
 
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -162,9 +162,12 @@
     }
   }
 
-  &[multiple] {
+  &[multiple],
+  &[size]:not([size="1"]) {
     height: auto;
+    padding-right: $custom-select-padding-x;
     background-image: none;
+
   }
 
   &:disabled {
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -167,7 +167,6 @@
     height: auto;
     padding-right: $custom-select-padding-x;
     background-image: none;
-
   }
 
   &:disabled {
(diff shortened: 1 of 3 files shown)


andresgalante (approved):

scss/_custom-forms.scss
```diff
@@ -149,7 +149,7 @@
   &:focus {
     border-color: $custom-select-focus-border-color;
     outline: 0;
-    @include box-shadow($custom-select-focus-box-shadow);
+    box-shadow: $custom-select-focus-box-shadow;
```
XhmikosR: Is this intentional?

scss/_custom-forms.scss
```diff
@@ -162,9 +162,12 @@
     }
   }
 
-  &[multiple] {
+  &[multiple],
+  &[size]:not([size="1"]) {
     height: auto;
+    padding-right: $custom-select-padding-x;
     background-image: none;
+
```
XhmikosR: This doesn't seem intentional :P

scss/_custom-forms.scss
mdo: Yes, we want the focus state to always be there regardless of the global shadow variable. This matches our other inputs.

mdo:
Going to get this merged in a few and then tackle #23058.

Issue #22060, opened 2017-02-24
Button Element not updated if target ID contains colon


svenhaag:
Bootstrap Version 4.0.0-alpha.6 and 3.3.7

Target ID = "collapse:Example"
data-target = "data-target="#collapse\\:Example""
aria-controls="collapse:Example"

Only the Target attributes get updated, not the button (toggle element) attributes.
IDs with a colon exist especially in the JSF world.

[Bug JS Fiddle](https://jsfiddle.net/DTcHh/30179/)

[Related Bug](https://github.com/twbs/bootstrap/issues/22041)

bkdotcom:
This works

```
<button role="button" type="button"
        data-toggle="collapse"
        data-target="#collapse\:\:Example"
        aria-expanded="true"
        aria-controls="collapse:Example">
  Button with data-target
</button>
<div class="collapse in" id="collapse::Example">
  bla bla bla
</div>
```

apparently one needs to escape the colons in the data attribute

svenhaag:
Hi, in the JSFiddle the data-target ID is already escaped with a single backslash, i.e data-target="#collapse\\:Example". This is an already valid jQuery selector. Though why doubling it to "\:\:" ? This does not work for me in the JSFiddle example. Maybe you can provide a running JSFiddle? Thanks.

bkdotcom:
[double colon fiddle](https://jsfiddle.net/DTcHh/30182/)

svenhaag:
Hi bkdotcom, adding a second double-colon is not a solution. JSF will render the ID attribute like "`a:b:c`".

bkdotcom:
> "second double-colon"

?

> JSF will render the ID attribute like "a:&#8203;b:c".

?

I inspected my fiddle's html and it's
```
<div class="collapse in" id="collapse::Example" aria-expanded="true">
  bla bla bla
</div>
```

> Button Element not updated if target ID contains double-colon

Fiddle has working double-colon

Am I missing something?

svenhaag:
Hi bkdotcom, sorry for the messup. I did not mean two subsequent colons but instead a single one. I fixed the heading and description of this issue. However the first JSFiddle still shows the bug. The toggling elements aria attributes are not updated if the ID contains a colon. Cheers.

peterdesmet:
[...]

PR #24700, opened 2017-11-07
Escape ID in Util.getSelectorFromElement

Johann-S:
Related to #22060

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -89,7 +89,7 @@ const Util = (($) => {
 
   function escapeId(selector) {
     // we escape IDs in case of special selectors (selector = '#myId:something')
-    // $.escapeSelector do not exist for jQuery release < 3
+    // $.escapeSelector does not exist in jQuery < 3
     selector = typeof $.escapeSelector === 'function' ? $.escapeSelector(selector).substr(1) :
       selector.replace(/(:|\.|\[|\]|,|=|@)/g, '\\$1')
 
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -87,6 +87,14 @@ const Util = (($) => {
     }
   }
 
+  function escapeId(selector) {
+    // we escape IDs in case of special selectors (selector = '#myId:something')
+    // $.escapeSelector do not exist for jQuery release < 3
+    selector = typeof $.escapeSelector === 'function' ? $.escapeSelector(selector).substr(1) :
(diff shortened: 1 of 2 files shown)


js/src/util.js
```diff
@@ -87,6 +87,10 @@ const Util = (($) => {
     }
   }
 
+  // Thanks to learn.jquery.com see : http://bit.ly/1Tyr4SZ
+  function escapeId(selector) {
```
Johann-S: Maybe we can avoid to create a new function for that :thinking:

XhmikosR:
If we required v3 there's already a function.

I'll review when I'm back in an hour or so.

Johann-S:
it seems it wasn't working in v3 too

XhmikosR:
We didn't call the function. How do you expect it to work ?

Johann-S:
Sorry I don't get it of which function do you talk about ?

XhmikosR:
https://api.jquery.com/jQuery.escapeSelector/

XhmikosR (approved):

Johann-S:
`$.escapeSelector` don't produce the same result for example : 
```html
<div id="test.id"></div>
```
```js
$($.escapeSelector('#test.id')).length // => 0
$('#test.id'.replace(/(:|\.|\[|\]|,|=|@)/g, '\\$1')).length // => 1
```

Johann-S:
Finally I found a way to use `$.escapeSelector` :smile:

XhmikosR:
We could add a comment about `escapeSelector`. Or we could use that if available maybe?

Johann-S:
I added a comment :+1:

js/src/util.js
```diff
@@ -112,6 +111,12 @@ const Util = (($) => {
         selector = element.getAttribute('href') || ''
       }
 
+      // if it's an ID
+      if (selector.charAt(0) === '#') {
+        // we escape IDs in case of special selectors (selector = '#myId:something')
+        selector = $.escapeSelector(selector).substr(1)
```
XhmikosR: This will fail if the user uses jQuery < 3.

Johann-S: Nice catch ! :+1:

XhmikosR: Just keep our function like in your first patch, add a comment that this is required for jQuery < 3 compatibility and call it conditionally maybe if it's defined.

Johann-S: done :+1:

PR #24707, opened 2017-11-07
Update _buttons.scss

deezin:
prevent text-color on visited links (especially buttons) from having the same color as the button's background color


mdo:
We don't style visited links at all, thanks though.

Issue #23674, opened 2017-08-25
Consistency on component paddings and margins


andresgalante:
On `_varibles.scss` there are several definitions of margins and paddings, and although the numbers are consistant, it is really hard to change one and keep consistency across all components.

I want to suggest that we introduce a set of component spacer variables with different sizes that will rule all components paddings and margins. That way by changing the variables in one place we'd have full control of all components and keep them consistant.

For example we have now:

```scss
$input-btn-padding-y:       .5rem !default;
$pagination-padding-y:      .5rem !default;
```

And I want to change it to:

```scss
$component-spacer-xs:   .5rem !default;

$input-btn-padding-y:    $component-spacer-xs !default;
$pagination-padding-y:   $component-spacer-xs !default;
```
We can even use the same $spacer mapping to generate them so utility classes follow the same meassurements we use on components.

You can read more about it here:
https://medium.com/eightshapes-llc/space-in-design-systems-188bcbae0d62

This is a large change that requieres a lot of testing, if you guys are ok with it, I can send a PR to see how it looks,

What do you think?

mdo:
This feels like an unnecessary abstraction right now. The example here at least doesn't seem to buy us more, and adds one more layer to the input-button variables. Passing for now I think.

PR #24710, opened 2017-11-07
$component-active-bg and $component-active-color are partially useful

pat270:
Changing `$component-active-bg` and `$component-active-color` only changes the style on some components, it looks like custom checkbox, custom radio, and pagination should also use these variables.

The second commit bases focus colors for Form elements and Buttons on `$component-active-bg`. This seemed logical to me, but wasn't sure if this behavior fits with Bootstrap 4's roadmap.


andresgalante:
Hey @pat270 😄  I think we should add them in places that have a direct relationship, like pagination. But I wouldn't add them to variables that don't have a tie to the `component-` variable.

To me, yes to you first commit, no to the second one. But I'd definitely add new component variables to standardise the way we treat components.

Thanks a lot!

pat270:
@andresgalante, An idea that I had that I didn't mention is doing away the `$component-active-*` variables (a breaking change that you might not want to introduce in beta). One size fits all global color variables are hard to implement because of the number of components and colors per component. I like the idea of Bootstrap's color system minimizing the number of global colors and taking it on a per component basis.

andresgalante:
I am for any kind of `.component-*` variable to standardise components (I bring one up here  #23674). So I'd say yes, but since it adds a layer of abstraction I'd wait for @mdo to comment.

tmorehouse:
Would this include focus styling for the active link in pagination? Currently for keyboard users it is not possible to tell if the active page link is focused.  I'm currently uising hte following CSS to add the focus styling:

```css
.pagination > .page-item > .page-link.active:focus {
    box-shadow: 0 0 0 3px rgba(0,123,255,.5);
    z-index: 1;
}
```

pat270:
@tmorehouse This just lets you change the `background-color` and `color` of `.page-link.active` through the `$component-active-bg` and `$component-active-color` variables.

For the focus style, it looks like Bootstrap just uses the browser default focus outline if `.page-link` is an `<a>` tag. Seems like the only way to customize it is to do what you did.

mdo:
Updated to resolve conflicts in `_variables.scss` from consolidating input/btn styles. Should be good to go, but if you could re-review, that'd be awesome!

mdo (approved):

pat270:
This LGTM.

PR #24713, opened 2017-11-07
Use native pseudo-selectors instead of jQuery selectors

vsn4ik:
https://www.w3.org/TR/css3-selectors/
https://caniuse.com/#feat=css-sel3


Johann-S (commented):
Not sure about that change, I know selector are processed threw Sizzle in jQuery but those CSS selectors are not compatible with IE 10 :worried: 

see : 
- https://caniuse.com/#search=first-child
- https://caniuse.com/#search=last-child
- https://caniuse.com/#feat=css-nth-child-of

vsn4ik:
@Johann-S, no, is compatible with IE 10.
See https://caniuse.com/#feat=css-sel3 and click on `Show all` button.

Johann-S (approved):
You're right my bad :smile: 

No concerns about merging this PR @XhmikosR :+1:

mdo (approved):

PR #24722, opened 2017-11-08
Update devDependencies

Johann-S:
A lots of improvements have been done in Rollup since 0.50.0 so we should update our used version see : https://github.com/rollup/rollup/blob/master/CHANGELOG.md


XhmikosR (approved):
Changed the title to a generic one since more things are updated

XhmikosR:
Also, see @mdo the issue with fsevents :/

PR #24725, opened 2017-11-08
v4 docs Collapse trigger using anchor tags should have `role="button"`

pat270:


patrickhlauke (approved):

patrickhlauke:
Good spot. May also be worth adding something to the accessibility bit later on in that page?

patrickhlauke:
Good spot. May also be worth adding something to the accessibility bit later on in that page?

pat270:
I added a little blurb to the Accessibility section. I wasn't sure if the role button thing was an oversight, Thanks!

patrickhlauke:
@pat270 lovely, thank you

Issue #24705, opened 2017-11-07
JS tests: test jQuery 1.9.1 too


XhmikosR:
Currently, we have no way of making sure things actually work on 1.9.1, which is our minimum supported jQuery version.

We should test against 1.9.1 and latest 3.x.

/CC @Johann-S

PR #24749, opened 2017-11-10
Run our unit tests on different jQuery versions

Johann-S:
~I added a new devDeps because `phantomjs` do not allow to pass parameters in URL~
~BTW if this PR is accepted I'll change `js-test-cloud` to use our `server` task :+1:~

Add Karma to run our unit tests on ChromeHeadless instead of PhantomJS + it's allow us to run our unit tests on two jQuery versions

Fixes: https://github.com/twbs/bootstrap/issues/24705


package.json
```diff
@@ -61,7 +63,8 @@
     "test": "npm-run-all dist js-test docs-compile docs-lint bundlesize",
     "watch": "npm-run-all --parallel watch-css watch-js",
     "watch-css": "nodemon --ignore js/ --ignore dist/ -e scss -x \"npm run css\"",
-    "watch-js": "nodemon --ignore scss/ --ignore js/dist/ --ignore dist/ -e js -x \"npm run js-compile\""
+    "watch-js": "nodemon --ignore scss/ --ignore js/dist/ --ignore dist/ -e js -x \"npm run js-compile\"",
+    "server": "http-server -s -p 3000"
```
XhmikosR: Not sure this makes sense in this branch since we already need Ruby.

I did this in my Hugo branch because the Ruby dep is removed.

js/tests/vendor/jqueryloader.js
```diff
@@ -0,0 +1,11 @@
+(function () {
+  var path = '../../assets/js/vendor/jquery-slim.min.js'
```
XhmikosR: Maybe we don't need this new file at all? It's a small snippet so perhaps we could have it in the HTML file?

XhmikosR:
IMO you should revert the http-server switch here; it doesn't make things better since we already need Ruby for Jekyll.

Also, the js-tests will probably need to be run sequentially; we can't use the same port simultaneously.

Johann-S:
My main issue here is how to launch a npm task in background and lauch our qunit tests :worried:

package.json
XhmikosR: OK, now I read you comment about phantom.

Why do we need phantom anyway? Can't we use jsdom or Chrome/Firefox headless?

Johann-S: Yep that's a possible switch because `phantomjs` won't receive any fixes

js/tests/vendor/jqueryloader.js
Johann-S: yep sure :+1:

Johann-S:
I added Karma to run our test on ChromeHeadless it's possible to add FirefoxHeadless too so if it's wanted do not hesitate :+1: 

With this implementation we can run our unit tests on jQuery 1.9.1 and the latest release of jQuery 3

js/tests/unit/modal.js
```diff
@@ -29,6 +29,8 @@ $(function () {
       $.fn.bootstrapModal = $.fn.modal.noConflict()
     },
     afterEach: function () {
+      $('.modal-backdrop, #modal-test').remove()
+      $(document.body).removeClass('modal-open')
```
Johann-S: I added that because in our `index.html` which run our unit test we did the same see : https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/index.html#L82

XhmikosR: But do we need to do it in modal.js if we do it in index.html?

karma.conf.js
```diff
@@ -1,13 +1,11 @@
...
     frameworks: ['qunit'],
     plugins: ['karma-chrome-launcher', 'karma-qunit'],
-
-
     // list of files / patterns to load in the browser
     files: [
-      jQueryVersion,
+      jqueryFile,
       'assets/js/vendor/popper.min.js',
       'js/dist/util.js',
```
XhmikosR: Would it work if we did `js/dist*.js` and just excluded `index.js`? Or even just include everything?

Johann-S: Nope because we have to be sure `util.js` is the first included script

js/tests/unit/modal.js
Johann-S: Karma do not use our index.html it generate a new index.html

Johann-S:
I made a test using `jsdom` and it's not possible because `jsdom` do not have `document.createRange` function which is used in Popper.js

Johann-S:
@XhmikosR I changed the reporter, and I used wildcard now to include our dist files and I added Firefox

karma.conf.js
```diff
@@ -0,0 +1,39 @@
+module.exports = function (config) {
+  const jqueryFile = process.env.USE_OLD_JQUERY === 'true' ? 'js/tests/vendor/jquery-1.9.1.min.js' : 'assets/js/vendor/jquery-slim.min.js'
```
XhmikosR: Maybe `process.env.USE_OLD_JQUERY` would be enough?

Johann-S: I don't understand what you mean here 😟

XhmikosR: I was wondering, wouldn't this be enough?

```js
const jqueryFile = process.env.USE_OLD_JQUERY ? 'js/tests/vendor/jquery-1.9.1.min.js' : 'assets/js/vendor/jquery-slim.min.js'
```

Oh, and perhaps we should only change the filename and keep the path since it's common?

Johann-S:
Do you have any concerns about that PR @mdo ? Or can we merge ?

FezVrasta:
About Popper.js on jsdom, the suggested way is to mock Popper.js away directly since it can't work on a jsdom environment even if you polyfill everything that's required.

Johann-S:
Seems better in parallel @XhmikosR 👍

XhmikosR (approved):

PR #24751, opened 2017-11-10
Wall of Browser Bugs: Chrome SVG perf bug (#24673)

### LINKED ISSUE #24673: Bootstrap's font-family style reduces SVG performance in Google Chrome

Herst:
Fixes #24673 and <https://crbug.com/781344> for details.


XhmikosR:
I saw a Chromium dev replied; we shouldn't hurry with this.

Herst:
It has been a month and the bug is still not assigned to anyone and there have been no more comments from Chromium developers. Off to the wall with it, I say. Please unhold.

XhmikosR (approved):

PR #24754, opened 2017-11-10
Use color-yiq for outline button text.

agriffis:
Replace a hard-coded #fff with color-yiq for text color.


andresgalante:
Hi @agriffis thanks for the contribution. I don't think we need a default value for `$color-hover` in this case since we are defining for every hover color but it doesn't hurt to remove the hardcoded.

agriffis:
Hi @andresgalante, thanks for the reply.

Regarding the default, I think customizers like myself would appreciate having it, otherwise I'll just be calling `color-yiq` on the way. It seems to make sense for `button-outline-variant` considering `button-variant` uses `color-yiq` for its text color. Do you disagree?

andresgalante (approved):
@agriffis you are right, it's good to have it to extend bootstrap

mdo:
Resolved the conflict here and put the `$color-hover` variable to use in the mixin. Turns out we weren't even using it there; we were overriding that value with an existing `color-yiq` function.

mdo:
Actually, to track that right, I cherry-picked your commit into #25339.

PR #24762, opened 2017-11-11
Update devDependencies.

XhmikosR:
@Johann-S: should we update popper.js to 1.12.7 too?


FezVrasta:
1.12.7 is just a bundle size reduction of the library. Safe to use

XhmikosR:
Updated popper.js to 1.12.7 too, thanks @FezVrasta.

Johann-S (approved):

PR #24767, opened 2017-11-13
Adds variables for form validation messages

### LINKED ISSUE #24763: Hardcoded font-size in form-validation-state
I use `$font-size-base: 0.8125rem;`.

Because the `form-validation-state` mixin has a fixed font-size of `.875rem`, I got validation messages of an unexpectedly bigger font.

Could this be changed to percent values or variables?

### LINKED ISSUE #24768: Different font size for "form-text text-muted small" and "invalid-text small"
I don't know if this is intended or what, but in my opinion those should have the same font sizes.

[Help Text](https://getbootstrap.com/docs/4.0/components/forms/#help-text) can be display with either

```
<input type="password" id="inputPassword5" class="form-control" aria-describedby="passwordHelpBlock">
<small id="passwordHelpBlock" class="form-text text-muted">My Help Text</small>
```

or if you want to omit the semantics

```
<input type="password" id="inputPassword5" class="form-control" aria-describedby="passwordHelpBlock">
<p id="passwordHelpBlock" class="form-text text-muted small">My Help Text</small>
```

---

Error messages in forms can be display using [Server Side Validation](https://getbootstrap.com/docs/4.0/components/forms/#server-side)

```
<input type="text" class="form-control is-invalid" id="validationServer03" placeholder="City" required>
<div class="invalid-feedback">Please provide a valid city.</div>
```

The issue I am seeing is that both, help text and error text are smaller than regular text, but still have different text sizes. For example, in my browser, help text has 12.8 px while error text has 14 px and normal text has 16 px. If I add `.small` to my error text class (`class="invalid-feedback small"`) it stays 14 px.

If you want to verify that, just compare the font sizes using the developer console in the linked documentation page. Maybe add class `.small` to one of the `.invalid-feedback` div's

When I noticed it, I had to display a list of errors (password too short, password same as username). In my case it would be beneficial if error text (`.invalid-feedback`) could the same size as the help text by default. Or at least having the option to make it the same size as help text by adding `.small`

andresgalante:
This PR closes #24763 by adding variables for margin top and font size for the validation feedback message.

Fixes #24768

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -515,6 +515,8 @@ $custom-file-text: (
 
 
 // Form validation
+$form-feedback-margin-top:          .25rem !default;
+$form-feedback-font-size:           ($font-size-base * .875) !default;
 $form-feedback-valid-color:         theme-color("success") !default;
 $form-feedback-invalid-color:       theme-color("danger") !default;
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -515,8 +515,8 @@ $custom-file-text: (
 
 
 // Form validation
-$form-feedback-margin-top:          .25rem !default;
-$form-feedback-font-size:           ($font-size-base * .875) !default;
+$form-feedback-margin-top:          $form-text-margin-top !default;
+$form-feedback-font-size:           $small-font-size !default;
 $form-feedback-valid-color:         theme-color("success") !default;
 $form-feedback-invalid-color:       theme-color("danger") !default;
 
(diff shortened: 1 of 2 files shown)


andresgalante:
I've added a new commit to fix #24768. Thanks @Borkason!

@mdo is it intentional to have no `font-size` set for `.form-text`. What if the user wants to use something else rather than a `small`? 
If not, I suggest we define a `font-size` for `.form-text` so feedback and form-text can share the same size regardless of the DOM structure.

mdo (approved):

PR #24795, opened 2017-11-15
Custom file focus styling for firefox keyboard users

tmorehouse:
Firefox does not correctly work with the `input[type="file"]:focus ~ .custom-file-control` sibling selector, as it considers the browse button inside the file input to have focus and not the actual `input[type="file"`), when either clicking the browse button or tabbing into the input.

See the answer for question https://stackoverflow.com/questions/20095105/input-type-file-focus-adjacent-selector-in-css (its an old question from 2015, but it is still applicable).

**Proposed solution:**
The `<label>` wrapper does make focus styling work for mouse users, but does not work for keyboard users (the inner browse button of the file input gets focus, not the actual file input when tabbing into the file input).

Adding a `focusin`/`focusout` handler (in `helpers.js`) that adds/removes the class `.focus`, and adding the class selector in `_custom-forms.scss` addresses this issue.

### DIFF
--- a/js/src/helper.js
+++ b/js/src/helper.js
@@ -70,7 +70,7 @@ const Helper = (() => {
   $(document).on(Event.INPUTFOCUSIN, Selector.CUSTOMFILE, handleFocusin)
   $(document).on(Event.INPUTFOCUSOUT, Selector.CUSTOMFILE, handleFocusout)
   $(document).on(Event.FORMRESET, Selector.FORM, handleFormReset)
-  
+
 })($)
 
 export default Helper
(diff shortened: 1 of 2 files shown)


Johann-S:
I'm not against it but my branch `add-js-helper` don't have any review from @mdo so we cannot be sure it will end as a merge of my branch :smile:

tmorehouse:
Heheh.  I didn't want to PR against v4-dev, but your branch has the right spot to add this hook in 😄

Johann-S:
if you can fix this error : https://travis-ci.org/twbs/bootstrap/jobs/302734501#L1093 I can merge your PR in my branch and let's see what the other members think about that :wink:

tmorehouse:
That silly space that got in there should be gone now 😉

Issue #24807, opened 2017-11-17
Feature: direction RTL support


dharmeshpipariya-zz:

mdo:
We definitely want to do this soon. The approach should be to generate a separate CSS file, built from `bootstrap-rtl.scss` with an ability to generate our other builds using the same approach if folks want it (no need to provide RTL builds for each one).

We'll need to hear from @Johann-S and @bardiharborow though to see if we need to adjust our JS plugins to account for the direction change. I dunno if Popper supports it off the top of my head.

@twbs/team I'd like this for v4.1 or v4.1—thoughts?

XhmikosR:
Sounds good to me. Feel free to ping me if there are new tools we need to add.

Superpencil:
Popper does not support it currently cc @FezVrasta

FezVrasta:
@Superpencil what do you mean? What should Popper.js to to support RTL?

kishorsathe:
Any tentative date , when it will be release ?

PerseusTheGreat:
Hi,
@mdo, please consider these:
- [Unofficial RTL-ized Bootstrap 4.x ](https://github.com/PerseusTheGreat/bootstrap-rtl)
- [RTL-ized Bootstrap 4.x Examples](https://perseusthegreat.github.io/bs4rtl/)

prateem:
Hi, @mdo 
Given that v4.1 is coming out soon and the ship list for it (#25375) doesn't include this issue, is there any update that can be provided about rtl support?
Thanks!

mdo:
No update to give other than this won't be in v4.1 :).

DJafari:
@mdo rtl support is coming for v4.2 ? :-(

nabeel-alalmai:
@mdo we need rtl support, is it coming in 4.2?

mohamedragaey:
Hello @mdo 
I have an approach for rtl-ltr support that i used on Bootstrap3 and now on bootstrap4 
here in this repo you will find that i took only the scss core files for bootstrap 4 and updated for direction and added my direction.scss file 
and in the build you will find 2 files bootstrap.css & bootstrap-rtl.css 
https://mohamedragaey.github.io/bootstrap-scss-multi-dir/
and here is my build tool included in bootstrap-rtl as npm package 
https://mohamedragaey.github.io/bootstrap4-playground/
if you find it interesting please let me know then i will make a pull request and then go for it .

MERamadan:
[...]

PR #24808, opened 2017-11-17
feat: RTL support

dharmeshpipariya-zz:
feature for support dir="rtl" for latest version of bootstrap #24807


andresgalante:
hi @dharmeshpipariya thanks for you PR. This has been discussed many times, the last one here: #23703

mdo:
The implementation approach here isn't something we'd move forward with. We should replace the values instead of overriding them each time.

PR #24813, opened 2017-11-17
adds hover state to the `button` element for background color utilities

### LINKED ISSUE #24806: Button in List group not getting hover state on focus
In my card I added a list group with `<a>` elements and one `<button>`. With normal background the items and also the button will get into hover state on focus but with bg-dark the button will no longer support hover state.

photo_2017-11-16_22-05-31
photo_2017-11-16_22-05-36

andresgalante:
This PR and it has 2 commits:

The first one adds hover state to the `button` element for background color utilities and fixes #24806 [Codepen with a demo of the problem](https://codepen.io/anon/pen/GOOojM).

The second one adds a transition to be consistent with other hovers like btns

This is what this PR generates:
https://codepen.io/andresgalante/pen/pddapy


XhmikosR (approved):

scss/mixins/_background-variant.scss
```diff
@@ -6,10 +6,12 @@
   #{$parent} {
     background-color: $color !important;
   }
-  a#{$parent} {
+  a#{$parent},
+  button#{$parent} {
     @include hover-focus {
       background-color: darken($color, 10%) !important;
     }
+    @include transition($btn-transition);
```
mdo: Let's ditch the transition—there's no need to match this against buttons because `.bg-` utilities shouldn't be used with our buttons.

andresgalante:
@mdo 👍  I've removed the last commit.

mdo (approved):

tomagladiator:
Tell me, I tested your codepen, and for the button.bg-dark the focus state doesn't works for me...

tmorehouse:
@andresgalante There also appears to be a related issue when the button (with or without a color variant) has the `.active` class as well. it lacks any hover/focus styling>

> image

Also, when focused (on non active items, that aren't the last item) the focus ring is cut off on the bottom.  perhaps adding `z-index: 1` when focused will help as well.

> image

mdo:
Active `.list-group-action`s get a browser default focus style:

screen shot 2017-12-22 at 6 39 58 pm

Testing the disabled ones in at least latest Safari and Chrome on macOS show no hover styles for me.

Note that while we can add the `z-index`, the outline will still be below the active one. Doing it any other way would cause the top border of the next item to overlap the blue background of the active one. It's an improvement, but not 100% perfect.

PR #24814, opened 2017-11-17
Add a simple script to generate SRI hashes.

XhmikosR:


mdo:
I take it this will generate the hashes and write them to the `_config.yml`?

XhmikosR:
Yup, exactly.

build/generate-sri.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+]
+
+files.forEach((file) => {
+  fs.readFile(file.file, 'utf8', (err, data) => {
+    if (err) {
+      throw err
+    }
+
+    const integrity = sriToolbox.generate({
```
bardiharborow: Something like this would be a lighter weight approach than adding another dependancy, but it's very much up to you.

```js
const integrity = 'sha384-' + require('crypto').createHash('sha384').update(data, 'utf8').digest('base64')
```

bardiharborow (approved):

Johann-S (approved):

build/generate-sri.js
XhmikosR: True. Not sure with which way to go since the overhead is small anyway...

XhmikosR: I tend to prefer the sri-toolbox approach.

PR #24854, opened 2017-11-23
Upgrade node-sass and qunit-phantomjs-runner

lipis:


XhmikosR:
Thanks but we update the dependencies ourselves periodically.

Let alone you are on on macOS and due to a bug in npm you add stuff to package-lock.json that are going to be removed if anyone runs `npm i` on another platform.

lipis:
np

PR #24855, opened 2017-11-23
Reformat all markdown files using Prettier

lipis:
[Prettier](https://prettier.io) is now supporting Markdown files and used it with the [`proseWrap`](https://prettier.io/docs/en/options.html#prose-wrap) option set to `false` if you want otherwise we can change it.. but I think most of the editors now-days support the word wrap and especially for writing documentation it's a bit annoying when we have to manually think about the theoretical line width.

I can also add some commit hooks so every time something is ending up in the repo will be properly formatted.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,89 +2,60 @@
 
 Looking to contribute something to Bootstrap? **Here's how you can help.**
 
-Please take a moment to review this document in order to make the contribution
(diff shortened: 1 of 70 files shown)


lipis:
If you notice that something is broken.. we can add `<--! prettier-ignore -->` in more places..

XhmikosR:
I don't see the gain from this. Plus it breaks formatting.

PR #24866, opened 2017-11-24
Use Jekyll's smartify filter when possible.

XhmikosR:
Example before:

```html
<p class="bd-lead">Get started with Bootstrap, the world's most popular framework for building responsive, mobile-first sites, with the Bootstrap CDN and a template starter page.</p>
```

After:

```html
<p class="bd-lead">Get started with Bootstrap, the world’s most popular framework for building responsive, mobile-first sites, with the Bootstrap CDN and a template starter page.</p>
```

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,14 +1,14 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-<meta name="description" content="{{ page.description | default: site.description }}">
+<meta name="description" content="{{ page.description | default: site.description | smartify }}">
 <meta name="author" content="{{ site.authors }}">
 <meta name="generator" content="Jekyll v{{ jekyll.version }}">
 
 <title>
   {%- if page.title -%}
-    {{ page.title }} · {{ site.title }}
+    {{ page.title | smartify }} · {{ site.title | smartify }}
   {%- else -%}
-    {{ site.title }} · {{ site.description }}
+    {{ site.title | smartify }} · {{ site.description | smartify }}
   {%- endif -%}
 </title>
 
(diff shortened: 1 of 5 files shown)


patrickhlauke:
is docs/4.0/content/tables.md the *only* place where single quotes/apostrophes need to be manually changed? i somehow doubt that (i for one never bother with smart quotes myself, and i'm sure there's more places in the md files where they're present).

this feels more like something that should be hooked into build step also for the `{{ content }}` but i guess that would result in far too many false positives when it hits markup bits?

XhmikosR:
There are no other cases. Rule of thumb: just don't use them in Markdown files.

XhmikosR:
Note that inside HTML code, even if that is in a markdown file, the quotes aren't converted. In those cases, one needs to explicitly use them.

patrickhlauke:
> There are no other cases. Rule of thumb: just don't use them in Markdown files.

not sure i follow here. so you're saying one shouldn't straight `'` quotes in markdown, but author using actual smart quotes always? it's a pain to do in Windows fwiw, but i'm more concerned if from now on we then need to make sure any PR that touches docs must be checked for appropriate use of smart quotes? (and if so that all reviewers be made aware of this)

XhmikosR:
I wouldn't say we need to do anything. Using smart quotes doesn't hurt
anything assuming the proper file encoding is used.

So the changes I made here is so that we are consistent and secondly so
that we use smart quotes where possible.

We don't need to overcomplicate things. The rules is what I said above.
Pure Markdown code doesn't need smart quotes because they are automatically
converted. HTML code anywhere needs it but there only a few places which I
plan to change later.

patrickhlauke:
> Pure Markdown code doesn't need smart quotes because they are automatically
converted.

ah, i don't think you actually said that above. in that case fair enough.

patrickhlauke (approved):

PR #24871, opened 2017-11-25
Tab.js: do not remove fade class if there's no initial active pane

alekitto:
A small issue was discovered while rewriting tab js without jQuery for #24094 ([discussion here](https://github.com/twbs/bootstrap/pull/24094#discussion_r152955270)):
The component does not trigger the fade effect if no active pane is present initially, subsequently removing the `fade` class from all the tab elements.

### DIFF
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -166,7 +166,6 @@ const Tab = (($) => {
       const complete = () => this._transitionComplete(
         element,
         active,
-        isTransitioning,
         callback
       )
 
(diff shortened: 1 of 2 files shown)


js/src/tab.js
```diff
@@ -166,26 +166,21 @@ const Tab = (($) => {
...
-
-      if (active) {
-        $(active).removeClass(ClassName.SHOW)
-      }
     }
 
-    _transitionComplete(element, active, isTransitioning, callback) {
+    _transitionComplete(element, active, callback) {
       if (active) {
+        $(active).removeClass(ClassName.SHOW)
```
Johann-S: Should be changed to : 
```javascript
$(active).removeClass(`${ClassName.SHOW} ${ClassName.ACTIVE}`)

Johann-S (changes requested):
LGTM after you made that change 👍

js/src/tab.js
alekitto: Done 👍

Johann-S (approved):

PR #24872, opened 2017-11-25
Remove unused CSS.

XhmikosR:
There is probably more, but let's leave it for later.


mdo (approved):
Nice, thanks for the cleanup!

PR #24876, opened 2017-11-26

XhmikosR:


bardiharborow (approved):

XhmikosR:
My only concern here is, the new stylelint version depends on postcss-html which in turn depends on postcss-sass. Should we install that too? The build seems to be fine without it.

Johann-S:
If it works without the add of `postcss-sass` then we don't have to add it 👍

Johann-S (approved):

Issue #24877, opened 2017-11-26
Docs: horizontal scroll on small screen/mobile


patrickhlauke:
There are 5 instances in our current documentation where, on a mobile device, there's an unsightly horizontal scroll due to elements being wider than 100% of the viewport.

Although these appear in the documentation, some of them may point to things that need fixing in core bootstrap regarding width/margins/responsiveness.

The below were found testing the docs as they stand just now in repo, using Chrome DevTools device emulation set to iPhone 6 (375px width)

> ![1-components-buttons](https://user-images.githubusercontent.com/895831/33241386-6ca4a288-d2bc-11e7-80f3-2e877ec6b5b1.png)
Components > Buttons

> ![2-components-input-group](https://user-images.githubusercontent.com/895831/33241389-780ae6d2-d2bc-11e7-9fe0-e71bbcc1e473.png)
Components > Input groups

> ![3-components-modal](https://user-images.githubusercontent.com/895831/33241394-838ff448-d2bc-11e7-9185-1c7ac6b65907.png)
Components > Modal (addressed by https://github.com/twbs/bootstrap/pull/24878)

> ![4-components-pagination](https://user-images.githubusercontent.com/895831/33241397-8b1863f8-d2bc-11e7-8702-6668d1263047.png)
Components > Pagination

> ![5-utilities-text](https://user-images.githubusercontent.com/895831/33241400-943b130e-d2bc-11e7-8216-068a1bc0a84c.png)
Utilities > Text

patrickhlauke:
To purely address the "don't make docs scroll horizontally" issue, added all necessary fixes to https://github.com/twbs/bootstrap/pull/24878

However, the root issue of why button groups, custom file inputs and large pagination cause an overflow in the first place / don't adapt to the correct width should likely still be investigated properly.

PR #24878, opened 2017-11-26
Fix docs horizontal scrollbar

patrickhlauke:
Addresses the scrolling issues in our docs

Note: this only deals with the scrolling of the docs caused by certain examples and the way code is presented in the docs. For the former cases, a more fundamental fix is likely still necessary to address the overflowing problem at the root.

Closes https://github.com/twbs/bootstrap/issues/24877


patrickhlauke:
Before:
> 3-components-modal

After:
> 3-components-modal-fix

patrickhlauke:
Added further fix that adds `overflow:scroll` to the examples class in the docs. This papers over all remaining issues where the docs have horizontal scrollbar (though the root causes still persist of course).

patrickhlauke:
Fixes look like this (horizontal overflow scrollbar)

> 1-components-buttons-fixed

> 2-components-input-group-fixed

> 4-components-pagination-fixed

> 5-utilities-text-fixed

assets/scss/_component-examples.scss
```diff
@@ -118,6 +118,7 @@
   position: relative;
   padding: 1rem;
   margin: 1rem (-$grid-gutter-width / 2);
+  overflow: scroll;
```
mdo: `overflow: auto` is a better option here—it'll only show a scrollbar when necessary vs always showing one.

docs/4.0/components/modal.md
```diff
@@ -459,7 +459,11 @@ For modals that simply appear rather than fade in to view, remove the `.fade` cl
 
 ### Dynamic heights
 
-If the height of a modal changes while it is open, you should call `$('#myModal').data('bs.modal').handleUpdate()` or `$('#myModal').modal('handleUpdate')` to readjust the modal's position in case a scrollbar appears.
+If the height of a modal changes while it is open, you should call
+{% highlight js %}$('#myModal').data('bs.modal').handleUpdate(){% endhighlight %}
```
mdo: Rather than introduce inline highlighting here (which we don't do anywhere else), let's rewrite the paragraph.

> If you change the contents of a modal while it's open, you may need to manually account for changes to the modal's height.
> 
> {% highlight js %}
> // Option 1
> $('#myModal').modal('handleUpdate')
>
> // Option 2
> $('#myModal').data('bs.modal').handleUpdate()
> {% endhighlight %}

Might need better comments than `Option 1/2` though?

mdo (changes requested):

assets/scss/_component-examples.scss
patrickhlauke: DOH! good catch, feel stupid for using `scroll`. my CSS must be rusty. changed.

docs/4.0/components/modal.md
patrickhlauke: Any hint as to when to use one option over the other? I admit that this is the part that's currently not explained at all in the docs, and I never grokked it myself (as i never had a need to dynamically alter modal content before)

patrickhlauke: @mdo digging a bit further, it seems the variant with `data(...)` is related to our programmatic API, but it's just a preference (some authors may use this)? as we don't seem to do this consistently anywhere else (we seem to just stick to the documented API methods), I'm actually more inclined to drop this here (as it then avoids having to go off on a tangent on why authors may or may not choose one over the other, which is best kept to the programmatic API docs page anyway).

mdo (approved):
Nice, thanks! Appreciate the shorter content.

mdo:
Also, I agree on the component behaviors here, but I've never found a good responsive variation for these many of our components that didn't require extra code to make them work.

patrickhlauke:
> Nice, thanks! Appreciate the shorter content.

de nada. it scratches an itch that i had for a while (the scrolling issue, as well as "there's something in the docs that I don't have a clue about...")

> Also, I agree on the component behaviors here, but I've never found a good responsive variation for these many of our components that didn't require extra code to make them work.

I'll have a look if at least some of these can't be tackled in a reasonably simple way (perhaps with flex) soon...

patrickhlauke:
@mdo do these issues (moved via automation etc) still need to be manually noted in the ship list?

mdo:
Not every change needs to be there. This one is small enough that I’m not overly concerned about noting it in the ship list. Automation only applies to the project board, not the ship list. I need to find an integration for automating that part :).

PR #24880, opened 2017-11-27
Update bootstrap-theme.css

wangjinju:
test


mdo:
Don't test on public repos! Your one PR just notified over 7,000 people of your test :(.

PR #24881, opened 2017-11-27
enhance sizing

storyn26383:
Add `notation` and `auto size` to Sizing (width and height), like **Spacing**, so we can use `<div  class="w-sm-100 w-md-50 w-xl-auto">` to do something amazing.

I think that will be more convenient and friendly :)


mdo:
This has been proposed before and I'm not in favor of it. Thanks though!

PR #24884, opened 2017-11-27
Update carousel controls. Fixes #24798

### LINKED ISSUE #24798: Increase visual contrast of carousel controls
Depending on the content of a carousel, the control icons and indicators can get lost against the background, especially when the background content is light.

Changing colors of the controls is not a good solution if you don't know what the background will be in advance (e.g. an image album).

By using the SVG `stroke` property for the `carousel-control-*-icons` and using a box-shadow for the `carousel-indictator`, we can get carousel controls that work well against multiple backgrounds.  This would solve some of what was brought up in #21985

[Codepen of default & enhanced controls](https://codepen.io/anon/pen/WXZprY)

If something like this is a good, idea, I'll be happy to submit a PR.

exige81:
Fixes #24798.  Adds a stroke to the carousel icon svg, and box shadows to carousel indicators for enhanced contrast against light and dark backgrounds.

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -146,6 +146,7 @@
     margin-left: $carousel-indicator-spacer;
     text-indent: -999px;
     background-color: rgba($carousel-indicator-active-bg, .5);
+    box-shadow: 1px 1px 1px darken($carousel-indicator-active-bg, 40%);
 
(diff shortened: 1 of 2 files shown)


mdo:
Closing per https://github.com/twbs/bootstrap/issues/24798#issuecomment-347237100.

PR #24860, opened 2017-11-23
Replace lint-vars.sh script with a shell.js script. 

XhmikosR:

PR #24891, opened 2017-11-27
Restore currently unused variables

mdo:
The new script in #24860 showed we weren't using two variables that should be in place; this PR restores them to their former glory.


XhmikosR (approved):

mdo:
Thanks for that fix!

PR #24898, opened 2017-11-28
Examples updates

mdo:
Fixes bugs, updates to newer techniques, and adds some new examples. In addition, I've replaced the old `examples.md` file with an `examples.html` that loops over a new data file, `examples.yml`, to generate our list of examples.

- **Added four new examples—Pricing, Checkout, Product, and Floating Labels.** Closes #18296.

- **Overhauled Dashboard** with a brand new navbar and layout, large Chartjs example instead of placeholder circles, icons from Feather, and more.

- **Redesigned Offcanvas**  completely to be a mobile-focused example that integrates a drawer menu into the Bootstrap navbar.

- **Redesigned Blog** to give it the appearance of popular blogging platforms and magazines, featuring horizontal cards, Google Web Fonts, and more.

- **Updated Sign-in** to vertically center the form via flexbox and add more copy, images, and footer.

- **Refreshed Album** to remove the bulk of the `.card` overrides, add more components to the albums, and improve the visual styles.

- **Updated Cover** to replace the table hack for vertical centering with flexbox, simplifying the HTML and CSS. Also fixed a bug where the nav would be left aligned on mobile devices.

Still have some more to do here, so leaving as WIP for v4 stable.

- [x] Responsive Product
- [x] Spacing on responsive Checkout
- [x] Medium responsive on Pricing
- [x] Responsive Blog (cards, scrolling nav)
- [ ] Retake screenshots
- [ ] Compress screenshots
- [ ] Squash useless commits

---

Closes #22847, takes a first pass implementation for #21312.

PR #24904, opened 2017-11-28
Only validate required form controls

mdo:
This changes the form validation styles to only apply to elements with an explicit `required` attribute. While working on a new form example for #24898, it caught me off guard that we were saying _optional_ inputs were `valid` when in fact there's nothing to actually validate. This change ensures that only required inputs, selects, textareas, and custom form controls are validated and effectively treats all non-`required` form controls as opt-in for form validation styles.

Here's the before and after to show you what I mean. Note how the second address line and two checkboxes shouldn't be green as it implies we performed validation on `value`s that don't exist.

| Before | After |
| --- | --- |
| screen shot 2017-11-28 at 2 04 25 pm | screen shot 2017-11-28 at 2 05 01 pm |

### DIFF
--- a/docs/4.0/components/forms.md
+++ b/docs/4.0/components/forms.md
@@ -707,7 +707,7 @@ We **highly recommend** custom validation styles as native browser defaults are 
 
 Here's how form validation works with Bootstrap:
 
-- HTML form validation is applied via CSS's two pseudo-classes, `:invalid` and `:valid`. It applies to `<input>`, `<select>`, and `<textarea>` elements.
+- HTML form validation is applied via CSS's two pseudo-classes, `:invalid` and `:valid`. It applies to `<input>`, `<select>`, and `<textarea>` elements with the `required` attribute.
 - Bootstrap scopes the `:invalid` and `:valid` styles to parent `.was-validated` class, usually applied to the `<form>`. Otherwise, any required field without a value shows up as invalid on page load. This way, you may choose when to activate them (typically after form submission is attempted).
 - As a fallback, `.is-invalid` and `.is-valid` classes may be used instead of the pseudo-classes for [server side validation](#server-side). They do not require a `.was-validated` parent class.
 - Due to constraints in how CSS works, we cannot (at present) apply styles to a `<label>` that comes before a form control in the DOM without the help of custom JavaScript.
--- a/scss/mixins/_forms.scss
+++ b/scss/mixins/_forms.scss
@@ -50,8 +50,8 @@
     border-radius: .2rem;
   }
 
-  .form-control,
-  .custom-select {
+  .form-control:required,
+  .custom-select:required {
     .was-validated &:#{$state},
     &.is-#{$state} {
       border-color: $color;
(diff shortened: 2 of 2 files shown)


patrickhlauke (approved):

tmorehouse:
Does restricting it to only `:required` inputs prevent it from validating elements that may not be required, but do have constraints (such as `pattern`, `min`, `max`, input type, etc), when the input has a value which doesn't match the constraint?

patrickhlauke:
hmm, good point @tmorehouse ... worth testing

tmorehouse:
@patrickhlauke & @mdo 

Here is a test that lets one validate "optional;" form fields with constraints. If the form field is empty, no validation applies:

https://codepen.io/tmorehouse/pen/QOZZYd

The following CSS is used:
```css
input:optional:invalid:not([value=""]),
input:required:invalid {
  background: hsla(0, 90%, 70%, 1);
}
input:required:valid {
  background: hsla(100, 90%, 70%, 1);
}
```

This appears to work well for textual style inputs and selects, and will style invalid non-empty constrained inputs.

Remove the selector `input:optional:invalid:not([value=""])` from the first CSS rule, and you will see that the constrained optional inputs (two by type, and one by pattern) will not show the invalid styling.

I was unable to get a "valid" optional non empty input to show green though.

mdo:
Testing this out, the problem with the `:optional:not([value=""])` is that optional fields without any input from the user appear validated. This is undesirable in my opinion for the same reason why I opened this PR—we should appear to have validated a value that doesn't exist. It makes sense why it works this way, but I'm anticipating we'll hear reports of confusion around that.

mdo:
Trying to play with this more: https://codepen.io/anon/pen/MOLEzK?editors=1100. Downside still is optional inputs that are valid with a non-zero value don't show green, but do when the input is invalid. Note the email address here:

screen shot 2017-12-02 at 1 57 29 pm

mdo:
Perhaps related to https://stackoverflow.com/questions/16952526/detect-if-an-input-has-text-in-it-using-css-on-a-page-i-am-visiting-and-do-no?

tmorehouse:
Yeah, I was unable to make valid optional inputs green, but they do show red when invalid, and assuming the input has not preset initial value.

patrickhlauke:
Arguably optional fields, even with constraints met, would be ok to show unstyled I'd say. Tricky conundrum...

tmorehouse:
And optional fields that do not meet constraints should be styled invalid.

For me, styled `invalid` inputs are more important that styled `valid` inputs.

tmorehouse:
May be valid inputs shouldn't be styled at all?  or make the valid styling opt-in, and only style invalid inputs?

mdo:
Going to wait on this and see what folks think after v4 goes stable.

ludovicofischer:
Maybe it’s worth considering some usability studies, for example this one https://baymard.com/blog/inline-form-validation

mdo:
Closing as stale.

PR #24905, opened 2017-11-28
Disable linter on .text-hide to prevent error

mdo:
Was causing our npm scripts to fail.


XhmikosR (changes requested):
This is already done in v4-dev. I don't see why you need this.

XhmikosR:
This is already done in v4-dev. I don't see why you need this.

mdo:
Doh! Thanks, I thought I had merged `v4-dev` into my branch.

PR #24915, opened 2017-11-29
Add order-last grid class

cahbb:
**This PR adds a new `.order-last` grid class.**

It does the exact opposite thing of `.order-first`. It reorders the target element to be the last using `order: $columns + 1;`

```html
<div class="container">
  <div class="row">
    <div class="col order-sm-last">
      First by default, but last on sm
    </div>
    <div class="col">
      Second by default, and still second on sm
    </div>
    <div class="col order-sm-first">
      Last by default, but first on sm
    </div>
  </div>
</div>
```
This gives the user a more intuitive way of understanding how the columns are ordered. Saying "I want this to be the last child on sm" and not begin to figure out how many columns you have in your grid and so on. The child is going to be what you say - the last child. 

As it is now you would use the class `.order-sm-12` instead of `.order-sm-last` in the above example.

 🚀 💪


vudaltsov (approved):

mdo:
I resolved the conflict here and tweaked the white-space of the source code. Should be good to go once tests pass.

cahbb:
Cool 🚀
I appreciate you including this PR in Bootstrap 👍🏼

PR #24932, opened 2017-12-01
Remove .col-form-legend in favor of .col-form-label

### LINKED ISSUE #24917: Feature request: col-form-label should have optional size classes
- Operating system and version: Any
- Browser and version: Any
- Bootstrap V4

Currently, `.col-form-label` and two helper classes for sizing `.col-form-label-sm` and `.col-form-label-lg`, so that the label can be matched with the form-control's sizing.

When using a `<fieldset>` to group related controls in a form-group, one uses the `<legend>` element to label the group of inputs, and one can use the corresponding class `.col-form-legend` to style the legend. 

Unfortunately there are no helper classes `.col-form-legend-sm` or `.col-form-legend-lg` to control the legend text size so that it matches up with the form-control size or other `col-form-labels` that may be in the same  form.

I have tried applying the `.col-form-label-sm` (or `-lg)` classes to the `<legend>` along with `.col-form-legend`, but it does not work, due to the order of the definition of the CSS rules which make `.col-form-legend`'s font size overrule `.col-form-label-sm`

Maybe just simply moving the `.col-form-legend` CSS rule before the `.col-form-label-{size}` rules would suffice?

tmorehouse:
Makes `.col-form-label` and `.col-form-label-(sm|lg)` work on `<legend>` elements, negating the need for the separate class `.col-form-legend`

Fixes #24917


mdo (approved):
I'll add a note to the migration docs after merging.

PR #24946, opened 2017-12-04
Enable floats to work with flex

pierreneter:


andresgalante:
Hi @pierreneter thanks for the PR. 
I am sorry but I don't follow, can you please explain why we'd want to add an `order` to the float mixins?

shtse8:
@andresgalante because flex only concern order, not float

mdo:
Don't mix floats with flex—choose one method and stick with that.

pierreneter:
@shtse8 set an `order` greater than 10 is a deliberate to custom css, oryou have more than 10 columns. So I think the value of 10 is reasonable. Or maybe, set it to `13`?

pierreneter:
@mdo do you have any suggestions? The `floats` don't work with `flex` now. Create a new class like `flex-right`?

pierreneter:
Oh, I see the relevance with #24915, I agreed with it. Thanks. Do not care about this PR (#24946) anymore.

PR #24947, opened 2017-12-04
add a link to an examplary implementation of the tooltip wrapper

mareksuscak:


andresgalante:
Hi @mareksuscak thanks a lot for taking the time to send a PR 😄 

Please change the .md file instead of the html here: https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/components/tooltips.md

I wouldn't add a link to stackoverflow. If you think that `disabled` elements needs more explanation on how to use the wrapper element, please improve the text directly on the docs.

mareksuscak:
I knew there must have been a source file lying somewhere. Thanks for pointing me to it. I can incorporate it into the doc page then.

PR #24952, opened 2017-12-05
Add Eyeglass Support

giordanovi:


Johann-S:
Maybe a bit of context here ? What is it Eyeglass ? What's the reason we need that ?

giordanovi:
I'm sorry,
eyeglass is a node-sass (github) extension manager built on top of npm. Using eyeglass, you can bring the power of node modules to your Sass files. You can find more information here:
https://github.com/sass-eyeglass/eyeglass
Thanks

mdo:
Thanks, but there's still far too little context here to do anything with this PR.

PR #24956, opened 2017-12-06
Carousel fade

supergyz:


Johann-S:
it seems you're not the author of those commits

Issue #24958, opened 2017-12-06
Transitionend with bootstrap 4


719media:
When looking into the necessity of $.support.transition.end, it seems to be that this is no longer needed.

I see that some properties were removed here:

https://github.com/twbs/bootstrap/pull/24511

But `webkitTransitionEnd` remains. However, according to https://caniuse.com/#feat=css-transitions, all browsers that bootstrap 4 supports use `transitionend` just fine. Is this even needed anymore?

Johann-S:
It seems you're right @719media :+1: 
/CC @XhmikosR

XhmikosR:
True, we can remove it. I'll make a PR later or if someone else does just
CC me.

On Dec 7, 2017 10:37, "Johann-S" <notifications@github.com> wrote:

It seems you're right @719media <https://github.com/719media> 👍
/CC @XhmikosR <https://github.com/xhmikosr>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/twbs/bootstrap/issues/24958#issuecomment-349898627>,
or mute
the thread

kamlekar:
@XhmikosR I did the PR. and cc'ed you. Not sure whether I have done it correct. Please check.

PR #24962, opened 2017-12-07
Changes related to #24958

kamlekar:
cc @XhmikosR 

Related Pull request: https://github.com/twbs/bootstrap/pull/24511

Fixes: #24958

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -20,10 +20,6 @@ const Util = (($) => {
 
   const MAX_UID = 1000000
 
-  const TransitionEndEvent = {
-    transition       : 'transitionend'
-  }
-
   // shoutout AngusCroll (https://goo.gl/pxwQGp)
   function toType(obj) {
     return {}.toString.call(obj).match(/\s([a-zA-Z]+)/)[1].toLowerCase()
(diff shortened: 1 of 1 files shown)


XhmikosR:
IMO we should remove the whole object and use the unprefixed property
directly.

kamlekar:
@XhmikosR  Yes, that would be far better. Let me try it out. Please feel free to close if someone do it already.

kamlekar:
@XhmikosR done.

js/src/util.js
```diff
@@ -50,11 +45,9 @@ const Util = (($) => {
 
     const el = document.createElement('bootstrap')
```
Johann-S: they are no needs now to create an element here just return : 
```js
return {
  end: 'transitionend'
}
```

that's all

Johann-S (changes requested):

js/src/util.js
kamlekar: @Johann-S done and committed.

Johann-S:
You have to fix that : https://travis-ci.org/twbs/bootstrap/jobs/312883636#L1063

Johann-S (approved):

PR #24974, opened 2017-12-08
Specify host 0.0.0.0 for Jekyll

patrickhlauke:
Allows for remote connections on same network (used to be the default behavior, but Jekyll now seems to default to 127.0.0.1 which means it does not react to incoming external requests)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -11,6 +11,7 @@ permalink:      pretty
 # Server
 source:         "."
 destination:    ./_gh_pages
+host:           0.0.0.0
 port:           9001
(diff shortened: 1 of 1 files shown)


patrickhlauke:
The need for this seems related to Jekyll changing its default configuration a while back. Wondering if it kept working until recently here because we hadn't updated dependencies... https://stackoverflow.com/questions/24108302/jekyll-serve-suddenly-switched-from-localhost-to-0-0-0-0-any-fix

mdo (approved):

PR #24976, opened 2017-12-08
feat(dropdown): add boundary config option

### LINKED ISSUE #24251: Big dropdowns are unusable when used inside an `overflow:scroll` container, due to being constrained inside it

tmorehouse:
Adds a new config option to specify the Popper.js `preventOverflow.boundariesElement`, which will allow a _non navbar_ dropdown to "overflow" its scrollParent (when the dropdown is inside an element with style `overflow: scroll`, or similar), without the need to _re-parent_ (#24257) the `.dropdown-menu` (maintaining document tab order).

It also, when `boundary` is **not** `'scrollParent'`, adds `position: static;` to the `.dropdown` (parent), which allows the actual overflow "escape" to happen.

The `boundary` config option accepts either a string (`'scrollParent'`, `'viewport'` or '`window'`), or a reference to an `HTMLElement`. The default value is `'scrollParent'` (the Popper.js default value). A reference to an HTMLElement can only be passed via JavaScript options and not via the data API `data-boundary=".."`.

Using `position: static` only (as suggested by @FezVrasta https://github.com/twbs/bootstrap/issues/24251#issuecomment-341413786) doesn't always completely address the issue unless Popper's `modifiers.preventOverflow.boundariesElement` is set to something other than `'scrollParent'` (specifically for taller dropdowns in short scroll containers)

Popovers and tooltips have a similar issue (when inside an element with `overflow: scroll`). for which I have created a separate PR #24979 (which doesn't need the `position: static` "fix")

/cc @Johann-S @patrickhlauke @FezVrasta 

Edit: Link to demo codepen: https://codepen.io/tmorehouse/pen/GyKYZe?editors=1010

Fixes #24251


tmorehouse:
Before hte "fix":
> image

With the fix and option `boundary` set to `'viewport`:
> image

tmorehouse:
This PR would negate the need for PR #24257 (re-parenting on dropdown open)

tmorehouse:
If class `.dropdown` always had `position: static`, then line 167 of this PR could be removed (https://github.com/twbs/bootstrap/pull/24976/files#diff-bfe9dc603f82b0c51ba7430c1fe4c558R167), although I am not sure if that would break anything else layout-wise.

Johann-S:
Your code looks good to me :+1: but it's seems your fork is a bit outadated maybe can you add a rebase ? And maybe a Codepen where we can try your PR would be nice please

js/src/dropdown.js
```diff
@@ -159,6 +161,10 @@ const Dropdown = (($) => {
             element = parent
           }
         }
+        // If boundary is not `scrollParent`, then set position to `static` to allow the
+        // menu to "escape" the scroll parent's boundaries
+        // https://github.com/twbs/bootstrap/issues/24251
+        $(parent).css('position', this._config.boundary !== 'scrollParent' ? 'static' : '')
```
XhmikosR: I think this would be better.

```js
if (this._config.boundary !== 'scrollParent') {
  $(parent).css('position', 'static')
}
```

Johann-S: agree with @XhmikosR :+1:

Johann-S: should we prefer using inline CSS instead of using one of our CSS classes ? (`.position-static` https://getbootstrap.com/docs/4.0/utilities/position/#common-values)

XhmikosR:
Shouldn't we have a test for this?

Johann-S:
Yep a Unit test would be rad :+1:

js/src/dropdown.js
tmorehouse: Oooh, i forgot that the positioning utilities has a .`position-static`.  it would be better than using an inline style.

tmorehouse:
I'll see if I can get a codepen set up for this.

For tests, not sure if will be easy to test this visually, but can definitely test to check for the .`position-static` class being applied when `boundary` is not `'scrollParent'`

tmorehouse:
@Johann-S @XhmikosR here is a codepen which demonstrates the default behavior and how it works with the new option: https://codepen.io/tmorehouse/pen/GyKYZe?editors=1010

tmorehouse:
@XhmikosR I've added in a basic unit test for checking for the presence (and absence) of the `position-static` class, based on the `boundary` setting.

XhmikosR:
When I view your pen with 360px width device, I see the text is still being cut off, but I guess that is normal, right? Because in normal resolution I see the impact of your patch.

js/src/dropdown.js
```diff
@@ -159,6 +161,12 @@ const Dropdown = (($) => {
             element = parent
           }
         }
+        // If boundary is not `scrollParent`, then set position to `static`
+        // to allow the menu to "escape" the scroll parent's boundaries
+        // https://github.com/twbs/bootstrap/issues/24251
+        if (this._config.boundary !== 'scrollParent') {
+          $(parent).addClass('position-static')
```
Johann-S: Maybe you can add `position-static` to our const var `ClassName` and replace that line by : 
```js
$(parent).addClass(ClassName.POSITION_STATIC)

Johann-S:
LGTM :+1:

Johann-S:
@XhmikosR do you know if they are a way to clean his git history ? Because they are unrelated commits :thinking:

FezVrasta:
Just squash them together

XhmikosR:
Either you do it manually and push to his fork, or *I think* everything will be squashed into one commit when you hit squash and merge.

XhmikosR (approved):

XhmikosR:
@Johann-S: so, LGTY?

Johann-S:
Yep it looks good to me if squash fix everything but I'm not able to merge his PR : 

image

Johann-S (approved):

XhmikosR:
You didn't try to squash the changes; you were trying to rebase hence the error.

Johann-S:
But I wasn't able to change to squash :worried: anyway thank you :+1:

tmorehouse:
Squash is good with maple syrup, butter and pepper ;-)

PR #24979, opened 2017-12-08
feat(tooltip+popover): add boundary config option

tmorehouse:
Related to issue #24251, larger popovers and tooltips which have the trigger element placed inside an element with `overflow: scroll` (or similar) will end up with weird placements that cover the trigger element (note the trigger button is completely hidden by the popover):

> image

This PR, similar to PR #24976, adds a `boundary` config option, that allows the user to specify what element constrains the positioning of the tooltip or popover. The  `boundary` option accepts a string - either `'scrollParent'` (the Popper.js default), `'viewport'` or `'window'` - or a reference to an HTMLElement (which can only be set via the JavaScript option, not `data-boundary="..."`).  The new config option `boundary` is passed directly to the [Popper.js config `modifiers.preventOverflow.boundariesElement`](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow) as is.

Popover with `boundary` set to `'viewport'`, which now can properly _breakout_ of the scrollable parent:

> image

This PR doesn't include the `position: static` fix required in PR #24976 (dropdowns), as tooltips and popover are typically appended to `<body>` and not the scrollable container, nor is there an obvious parent element to place `position: static` on.

/cc @Johann-S


Johann-S (approved):

Johann-S:
But same as your other PR, you have 15 commits in this PR, and I don't think we need all of those commits so please do a rebase so we can have only what's needed :+1:

tmorehouse:
Yeah, not sure how to clear out those older commits from the previous PR that was merged.

Johann-S:
We cannot :worried: 

image

PR #24986, opened 2017-12-11
Add map-name param to _assert-starts-at-zero

zacharywelch:
The `_assert-starts-at-zero` warning assumes map name is `$grid-breakpoints`. This PR adds a `$map-name` param similar to `_assert-ascending`.

```sass
@include _assert-ascending($nav-breakpoints, "$nav-breakpoints");
@include _assert-starts-at-zero($nav-breakpoints, "$nav-breakpoints");
```


mdo:
Would we consider this a breaking change at all?

scss/_functions.scss
```diff
@@ -21,12 +21,12 @@
 }
 
 // Starts at zero
-// Another grid mixin that ensures the min-width of the lowest breakpoint starts at 0.
-@mixin _assert-starts-at-zero($map) {
+// Used to ensure the min-width of the lowest breakpoint starts at 0.
+@mixin _assert-starts-at-zero($map, $map-name) {
```
MartijnCuppens: @mdo, we're completely covered if we add this, just in case developers also use this function in their custom code:
```suggestion
@mixin _assert-starts-at-zero($map, $map-name: "$grid-breakpoints") {
```

MartijnCuppens (approved):

PR #25001, opened 2017-12-13
Updated Quick Start Link

JCA122204:


mdo (approved):

XhmikosR:
Why do we need this? It's redirected.

mdo:
We've updated URLs elsewhere to avoid the redirect, so figured this one was fine, too. Better to skip the redirect whenever possible IMO.

PR #25015, opened 2017-12-15
document how to add a tooltip on disabled elements

mareksuscak:
as requested in #24947 i've added a hint on how to add a tooltip on disabled elements in the source file

the original text was taken from v3 docs as seen below and enriched with the information from [this popular SO answer](https://stackoverflow.com/a/19938049/596625)

image

i think it's worth adding this to the docs


mdo:
Docs section for both tooltip and popovers coming in #25095. Thanks for the suggestion!

PR #25020, opened 2017-12-16
Rewrite input group component

### LINKED ISSUE #23454: .invalid-feedback doesn't show for .input-group with an invalid control
Example:

```html
<div class="form-group col-md-4">
  <label class="form-control-label is-invalid" for="valuation_value">Estimated Value</label>
  <div class="input-group is-invalid">
    <span class="input-group-addon">£</span>
    <input class="form-control is-invalid" min="0" type="number" value="" name="valuation[value]">
  </div>
  <div class="invalid-feedback">can't be blank, is not a number</div>
</div>
```

Results in:

screen shot 2017-08-15 at 11 59 15

I would expect to see `can't be blank, is not a number` under the input-group.

### LINKED ISSUE #25005: Different way are used to control overlap border of `input-group-addon` between` input-group-btn`
Regarding overlapping of borders of a grouped element, @mdo commented in #24453 that it is not good way to control the border color. As an alternative, we can use `border-width: 0` or negative margin.

In input group, `.input-group-addon` uses **border width to 0**, and ` input-group-btn` uses **negative margin**. I think we should use a same way (negative margin), since ` .input-group-addon` and `input-group-btn` are similar in principle.

https://github.com/twbs/bootstrap/blob/9de141f42819f3ad374bf4774ddbefea7df8f950/scss/_input-group.scss#L145-L151

https://github.com/twbs/bootstrap/blob/9de141f42819f3ad374bf4774ddbefea7df8f950/scss/_input-group.scss#L165-L172

mdo:
I've rewritten the input group HTML (slightly) and CSS (mostly). This was primarily motivated by a need to support our HTML5 form validation and an urge to improve the readability of this component's CSS. Prior to this PR, adding form validation feedback and styles wasn't possible without visually breaking the component.

**This is a breaking change from Beta 2.**

### Changes

- Input group addons are now specific to their placement relative to an input. We've dropped `.input-group-addon` and `.input-group-btn` for two new classes, `.input-group-prepend` and `.input-group-append`. You must explicitly use an append or a prepend now, simplifying much of our CSS.

- Multiple inputs are now supported, though I'm unsure on the accessibility angle here (/cc @patrickhlauke). I'm also unsure how this affects form validation. Fixes #23454.

- Added support for form validation feedback in input groups. `.input-group` now has `flex-wrap: wrap;` on by default, allowing you to place validation feedback within the `.input-group` and style it based on the input's overall state. This isn't applicable to file inputs though.

- Buttons now must all be of the same element type for `:last-of-type` to work. We use `:last-of-type` to avoid `border-radius` issues from HTML5 form validation and dropdowns being the true `:last-child` elements.

- We no longer allow `.form-control-sm` or `.form-control-lg` on the input group elements. Instead, use `.input-group-sm` and `.input-group-lg` on the parent `.input-group`. Currently I'm unclear how to handle custom file and select sizing.

- There are a few more direct children selectors. While we strive to avoid direct children selectors in Bootstrap 4 (e.g., `.parent > .child`), input groups require them to avoid cascading styles into additional components. For example, we have to target only the children buttons in a prepend or append otherwise we could affect dropdown menu content.

- More consistently handles double borders, fixes #25005.

- We've dropped ~50 lines of source Sass in the process.

I'll update this list as I work through todos listed below.

### Todos

- [x] Double check files and selects
- [x] Form validation example added to form docs (verifying multiple inputs? file inputs?)
- [x] Update input groups everywhere else in docs and examples
- [x] Migration docs added w/ breaking changes detailed
- [ ] Explore a custom input group button, styled to match the input with a white background and same color border
- [ ] Do we try to do file input and custom select sizes?

### DIFF
--- a/docs/4.0/components/button-group.md
+++ b/docs/4.0/components/button-group.md
@@ -60,7 +60,9 @@ Feel free to mix input groups with button groups in your toolbars. Similar to th
     <button type="button" class="btn btn-secondary">4</button>
   </div>
   <div class="input-group">
-    <span class="input-group-addon" id="btnGroupAddon">@</span>
+    <div class="input-group-prepend">
+      <div class="input-group-text" id="btnGroupAddon">@</div>
+    </div>
     <input type="text" class="form-control" placeholder="Input group example" aria-label="Input group example" aria-describedby="btnGroupAddon">
   </div>
 </div>
@@ -73,7 +75,9 @@ Feel free to mix input groups with button groups in your toolbars. Similar to th
     <button type="button" class="btn btn-secondary">4</button>
   </div>
   <div class="input-group">
-    <span class="input-group-addon" id="btnGroupAddon2">@</span>
+    <div class="input-group-prepend">
+      <div class="input-group-text" id="btnGroupAddon2">@</div>
+    </div>
     <input type="text" class="form-control" placeholder="Input group example" aria-label="Input group example" aria-describedby="btnGroupAddon2">
   </div>
 </div>
--- a/docs/4.0/components/forms.md
+++ b/docs/4.0/components/forms.md
@@ -731,14 +731,26 @@ When attempting to submit, you'll see the `:invalid` and `:valid` styles applied
 
 {% example html %}
 <form id="needs-validation" novalidate>
-    <div class="col-md-6 mb-3">
   <div class="form-row">
+    <div class="col-md-4 mb-3">
       <label for="validationCustom01">First name</label>
       <input type="text" class="form-control" id="validationCustom01" placeholder="First name" value="Mark" required>
     </div>
-    <div class="col-md-6 mb-3">
+    <div class="col-md-4 mb-3">
       <label for="validationCustom02">Last name</label>
       <input type="text" class="form-control" id="validationCustom02" placeholder="Last name" value="Otto" required>
+    </div>
+    <div class="col-md-4 mb-3">
+      <label for="validationCustomUsername">Username</label>
+      <div class="input-group">
+        <div class="input-group-prepend">
+          <span class="input-group-text" id="inputGroupPrepend">@</span>
+        </div>
+        <input type="text" class="form-control" id="validationCustomUsername" placeholder="Username" aria-describedby="inputGroupPrepend" required>
+        <div class="invalid-feedback">
+          Please choose a username.
+        </div>
+      </div>
     </div>
   </div>
   <div class="form-row">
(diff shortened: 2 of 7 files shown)


mdo:
Screenshots of the new component in action...

**Custom file:**
screen shot 2017-12-15 at 7 48 13 pm

**Custom select:**
screen shot 2017-12-15 at 7 48 05 pm

**Segemented dropdowns:**
screen shot 2017-12-15 at 7 29 17 pm

**Multiple inputs:**
screen shot 2017-12-15 at 7 28 52 pm

**Validation styles:**
screen shot 2017-12-15 at 7 27 14 pm

alecpl:
What if I want two inputs and a text label between? E.g. timespan input group. Another case, I'm using multiple inputs in a wrapper div as a main element of input-group, surrounded by input-group-addons, will that work?

mdo:
The component can't and shouldn't solve everything, @alecpl. I'd suggest something of a hybrid for that, mixing this and/or our utils for something custom.

ysds:
How about add `.btn-static` class as a modifier for a button instead of `.input-group-text`? The SCSS will be more concise and modulable.

```html
<div class="input-group mb-3">
    <span class="btn btn-static" id="basic-addon1">@</span>
  </div>
  <input type="text" class="form-control" placeholder="Username" aria-label="Username" aria-describedby="basic-addon1">
</div>
```

SCSS: (This is an example, there are many code where similar things can be done.)
```scss
.btn + .btn,
.btn + .input-group-text,
.input-group-text + .input-group-text,
.input-group-text + .btn {
  margin-left: -$input-border-width;
}
```
to
```scss
.btn + .btn {
  margin-left: -$input-border-width;
}
```

mdo:
> How about add .btn-static class as a modifier for a button instead of .input-group-text? The SCSS will be more concise and modulable.

Doesn't feel right to extend the button class into something that isn't actually a button.

ngbrown:
If the `btn-input-group` is in a row, and one of the columns has texts that wraps at certain widths, then the `<input />` element stays at 100% height, while the `input-group-append` div does not, causing weird height issues.

This worked fine with beta 2's `input-group-append`.

Here's an example:

screenshot - 2018-01-03 6_22_00 pm

I can had `h-100` class to the `<span/>` but that still shows the text mis-aligned.  Adding vertical text alignment classes or styles doesn't change that within a `<span/>`:

```html
<div class="col-sm-9 form-row">
    <div class="input-group col mb-3">
        <input type="search" class="form-control" placeholder="Search" id="searchAdvanced.qManual" name="searchAdvanced.qManual"
            value="">
            <span class="input-group-text h-100" id="searchAdvanced.qAspects">involving:someone</span>
        </div>
    </div>
    <button type="submit" class="btn btn-outline-success mb-3">Search</button>
</div>
```

screenshot - 2018-01-03 6_29_15 pm

br3nt:
How do you add invalid feedback to the custom files?

PR #25044, opened 2017-12-20
feat: adds `.btn-circle`

zalog:
It would be nice for us to have circle buttons, but I think we have to make sure that all the child icons align fine.
Demo: https://codepen.io/zalog/pen/YYWraX

Perhaps we should use flex `.btn-circle { display: inline-flex; justify-content: center; align-items: center; }`, and make some tunings on the icon itself?
And some for child svgs to take the inherit color: `svg { fill: currentColor; }`?

This is a wip temp branch because I'm not sure if this is the right abordation. Especially on the (16 / 2) part to get that 2px border from the normal buttons.


MartijnCuppens:
Thanks, but we'll pass on this for now.

PR #25047, opened 2017-12-21
Input Groups don't need to jankily prevent input button groups from w…

pat270:
…rapping anymore with flexbox

This looks like leftovers from Bootstrap 3 when input-groups were table based.

https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_input-groups.scss#L137-L139


mdo:
#25020 will address this. Thanks though!

PR #25049, opened 2017-12-21
Customize lint vars output, improve messaging

mdo:
Originally added in #24860, the output of the messaging was confusing to me at first, stating that a variable was _being used_ once rather than _it's unused_. The former sent me down a tailspin of thinking my editor's find function was broken :D.

While I was in there, I stylized the output a bit, too.

screen shot 2017-12-21 at 2 11 05 pm


build/lint-vars.js
```diff
@@ -27,7 +27,7 @@ function findUnusedVars(dir) {
     process.exit(1)
   }
 
-  console.log(`Finding unused variables in "${dir}"...`)
+  console.log(`\x1b[1m%s\x1b[0m`, `\n➜ Finding unused variables in "${dir}"...`)
```
XhmikosR: I'd rather not add those unicode stuff because it's not cross-platform.

Also, note that I have made this script a module https://github.com/XhmikosR/find-unused-sass-variables. I just need to publish it after some tweaks and then I'll switch to that here too.

XhmikosR: So, if you did only the word changes, I'd appreciate it.

mdo: > I'd rather not add those unicode stuff because it's not cross-platform.

Is there a cross-platform solution you can recommend? I assume you mean Mac and Windows, correct?

XhmikosR:
Please look at the package I've made and submit patches there. This script
is obsolete and I will replace it.

I have a few ideas for the symbols yeah. And I already have added colors
there. But feel free to submit patches so that we are done sooner and
switch to it faster here :)

PR #25050, opened 2017-12-21
Form check markup v2

### LINKED ISSUE #23426: Non-custom checkboxes don't change with .is-invalid

### LINKED ISSUE #23449: Normalize Custom Checkbox and Radio with Form Check

### LINKED ISSUE #24624: What’s the correct mark-up for custom radios and server-side validation?
I’m having trouble rendering stacked custom radios with server-side validation.

My raw mark-up looks like this (a Laravel Blade template):

```blade
<div class="form-group">
    <label class="form-control-label">Category</label>
    <div class="custom-controls-stacked">
        @foreach($categories as $id => $name)
            <label class="custom-control custom-radio">
                <input type="radio" id="category-{{ $id }}" name="category_id" value="{{ $id }}" class="custom-control-input" {{ old('category_Id') == $id ? 'checked ' : '' }}/>
                <span class="custom-control-indicator"></span>
                <span class="custom-control-description">{{ $name }}</span>
            </label>
        @endforeach
    </div>
    @foreach($errors->get('category_id') as $error)
        <div class="invalid-feedback">{{ $error }}</div>
    @endforeach
</div>
```

Where is the appropriate place to put the `.is-invalid` class that both makes the radio inputs red, _and_ displays the `<div class="invalid-feedback">` elements?

### LINKED ISSUE #24888: Custom stacked checkbox label stretches to full width
**Operating system and version:** Any
**Browser and version:** Any
**Test case:** https://getbootstrap.com/docs/4.0/components/forms/#stacked

bootstrap_v4_custom_checkbox_stacked_old

As seen in the picture above, the label of a stacked checkbox stretches to the full with of the parent container. This can cause issues when a user clicks far right from the visible label and accidentally toggles a checkbox or radio.

A simple fix would be to set `align-items` for the `custom-controls-stacked` container to `flex-start` (see picture below).

bootstrap_v4_custom_checkbox_stacked_new

mdo:
I opened #23444 several weeks ago to gather feedback, but I haven't seen or heard much. This past week I put the changes to the test in this Codepen and found a simpler, more flexible path forward. Thus, I've once again rewritten the native and custom checks.

Now, both native and custom checks share the same HTML, but different classes:

```html
<!-- Default -->
<div class="form-check">
  <input type="checkbox" class="form-check-input" id="exampleCheck1">
  <label class="form-check-label" for="exampleCheck1">Browser default checkbox</label>
</div>

<!-- Custom -->
<div class="custom-control custom-checkbox">
  <input type="checkbox" class="custom-control-input" id="customCheck1">
  <label class="custom-control-label" for="customCheck1">Custom checkbox</label>
</div>
```

Here's why things have changed:

- This structure allows sibling selectors (`~`) to style the `<label>` based on the `<input>` state.
- For form validation feedback, we can also use sibling selectors to show valid or invalid text.
- Previously, default and custom checks were vastly different in markup. They're now the same, but different classes. For custom checks, the custom indicator was previously a `<span>`—it's now generated `::before` and `::after` pseudo-elements.
- Default and custom checks layout previously behaved differently—default being stacked, custom being inline. Now, both are the same and `.custom-controls-stacked` is no more. Inline custom checks need the `.custom-control-inline` modifier, matching the defaults.

Altogether, these changes fix #23426, fix #23449, close #23503, fix #24624, fix #24888.


hiddewie:
I think this is a good step forward, but my original comment (https://github.com/twbs/bootstrap/pull/23444#issuecomment-322407624) has not been solved with the new styling for the new markup.

In case a whole `<div>` or `<fieldset>` is valid/invalid and marked with `.is-valid` or `.is-invalid`, then the labels will not change their color.

mdo:
Semantically, neither the `<div>` or `<fieldset>` elements have any sense of valid or invalid. Our validation styles are predicated on being attached to the inputs directly, not a parent element. This has been the case since I first overhauled these styles to use `:valid`/`:invalid`.

hiddewie:
I think I disagree that a fieldset has no semantic meaning. Especially for groups of checkboxes or radio buttons, they are used to group. Screen readers may also use this information for users with visual disabilities. See also https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/How_to_structure_an_HTML_form.

As an example where a user must select at least one item in a group of checkboxes. Then the group is invalid if none has been checked, while the individual checkboxes may be valid if it is or if it is not checked.

mdo:
I didn't say there's no semantic meaning at all to those elements, only that they have no `:valid`/`:invalid` semantics. A fieldset cannot be valid or invalid—it has no support for those states. It can only be in it's default state or `disabled`.

A group of radio buttons sharing the same name will automatically validate together as a group. Checkboxes will not though, as they don't behave like radios.

hiddewie:
Yes, I agree.

But this only holds true for client side validation. For server side validation, would you recommend adding the is-valid (or invalid) class to each checkbox of a group rather than the div or fieldset they live in for showing the validity to the user?

mdo:
Ah, sometimes I get ahead of myself and forget we have _way_ more flexibility with the server-side options. Definitely opens the doors for more flexible styling of parent groups and more.

Can you open an issue with a reference to this thread, your original comment, and any demos? I'd like to see if this is something we could tackle in a subsequent release. I know a lot of folks would like more flexibility for these classes. We could add onto them as feedback rolls in.

For now, you'll want to add-on to Bootstrap's classes to accomplish that. It'd be decent scope creep on this PR to try to tackle that here. To be flexible enough for multiple types of inputs and layouts, I imagine you'd want a parent and child class pairing—that way you could selectively include which inputs are valid/invalid. But, that's without building anything myself to test that out.

Sorry I can't be of more help there yet, and thanks for the back and forth!

pat270:
@mdo, I'm late to the party, but the old markup pattern for checkboxes/radios was nice because you could stuff other elements inside the checkbox/radio label and essentially make the hit area for checkboxes larger. One example is a row of checkable thumbnails, clicking on any part of the thumbnail would check the checkbox.

mdo:
@pat270 I feel like you can still build things like that, you just might need to customize a bit more. I'd be open to hear more and see examples so we can get this on the docket for whenever a v5 needs to hit for deprecating older browsers, fixing legacy issues, etc.

pat270:
@mdo You are right, I was able to figure it out with help from https://github.com/twbs/bootstrap/issues/25179#issuecomment-354909891. The only suggestion I would have is to separate the checkbox and label elements instead of bundling it together using `::before` and `::after` pseudo elements. We can position the label on the left or right based on markup instead of relying on CSS to position the checkbox on the other side (I think this is generally easier). It also makes removing the space between the text and checkbox trivial when no label text is present. http://pat270.github.io/lexicon/content/test-table-view-template/ is an example of how we are using Bootstrap's checkboxes.

PR #25051, opened 2017-12-21
Move from headings to divs to fix #24975

### LINKED ISSUE #24975: List group item heading link is dead
https://getbootstrap.com/docs/4.0/components/list-group/

In the side menu to the right you can see three anchor links named "List group item heading" under "Custom content", none of these works.

Dead anchor links:
https://getbootstrap.com/docs/4.0/components/list-group/#list-group-item-heading
https://getbootstrap.com/docs/4.0/components/list-group/#list-group-item-heading-1
https://getbootstrap.com/docs/4.0/components/list-group/#list-group-item-heading-2

mdo:
Janky fix that we're using elsewhere to get around the Jekyll ToC shortcomings. Will need to revisit down the line.

### DIFF
--- a/docs/4.0/components/list-group.md
+++ b/docs/4.0/components/list-group.md
@@ -136,7 +136,7 @@ Add nearly any HTML within, even for linked list groups like the one below, with
 <div class="list-group">
   <a href="#" class="list-group-item list-group-item-action flex-column align-items-start active">
     <div class="d-flex w-100 justify-content-between">
-      <h5 class="mb-1">List group item heading</h5>
+      <div class="h5 mb-1">List group item heading</div>
       <small>3 days ago</small>
     </div>
     <p class="mb-1">Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit.</p>
@@ -144,7 +144,7 @@ Add nearly any HTML within, even for linked list groups like the one below, with
   </a>
   <a href="#" class="list-group-item list-group-item-action flex-column align-items-start">
     <div class="d-flex w-100 justify-content-between">
-      <h5 class="mb-1">List group item heading</h5>
+      <div class="h5 mb-1">List group item heading</div>
       <small class="text-muted">3 days ago</small>
     </div>
     <p class="mb-1">Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit.</p>
@@ -152,7 +152,7 @@ Add nearly any HTML within, even for linked list groups like the one below, with
   </a>
   <a href="#" class="list-group-item list-group-item-action flex-column align-items-start">
     <div class="d-flex w-100 justify-content-between">
-      <h5 class="mb-1">List group item heading</h5>
+      <div class="h5 mb-1">List group item heading</div>
       <small class="text-muted">3 days ago</small>
     </div>
     <p class="mb-1">Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit.</p>


XhmikosR:
Note that the Hugo branch doesn't need these workarounds and works fine.

XhmikosR:
Also, @mdo a workaround we can use, assuming you don't mind going through all the docs, is to limit the depth for toc.

https://github.com/twbs/bootstrap/blob/5cede314698a382fb9d7fddacce0a990bc2940d9/_config.yml#L79-L81

Which I already have set its max to 4, so this fixes nothing.

So, as an intermediate workaround, we could be consistent with examples' headers (say use an h5 or h6) and then jekyll-toc shouldn't generate those.

patrickhlauke:
have to admit i'm not a fan of removing semantic markup. is there no clean way to exclude example markup (e.g. headings, regardless of level) from Jekyll TOC?

XhmikosR:
@patrickhlauke: like I said above, this PR fixes nothing since I have limited jekyll-toc to h2-h4. So any h5, or h6 don't show up in toc.

So, the best way with what we have now, is to make sure what we want to exclude isn't included in jekyll-toc's limits.

patrickhlauke:
ok, but is there a way to also explicitly tell jekyll not to include certain things? otherwise, there may be situations where a heading should really be an `<h4>` or something, but we have to bump it (illogically) to `<h5>` just because that skirts around the TOC issue.

(not that it's a big deal, more that it would be nice to know for possible future needs)

XhmikosR:
Nope there's no way, at least not currently.

The nice thing about my Hugo branch is that I just works without any hassle
:)

mdo:
Definitely agreed on the problematic usage of this—wasn't something I was stoked about. That said, I assumed because @XhmikosR didn't mention anything in #24975 that the issue was unresolved. Looking at the docs again, it's clear the config settings are being applied, so this is unnecessary. I'll need to go back and audit the docs for any prior usage, too.

> The nice thing about my Hugo branch is that I just works without any hassle

At risk of being a bit of a jackass, please stop passively aggressively reminding me that you have a huge PR to be reviewed to move us to Hugo. You want me to sign off on it, but I honestly cannot give you the time to not only properly review those changes, but to fully understand Hugo like I do Jekyll. I'm not going to switch site generators right now—there's enough to do in just getting v4 done.

XhmikosR:
Well, you are, no offense. Because I keep bugging you for more than a month
for an answer, which I just now get from you.

The site generator doesn't affect the end users. And

1. I have already done the work
2. I was planning to do any more work to get it finished
3. Everyone will learn the basics like I also do, it's not so hard nor the
learning curve is huge

That being said, suit yourself, and stay with a clearly inferior project
which clearly slows down development.

mdo:
> Well, you are, no offense. Because I keep bugging you for more than a month for an answer, which I just now get from you.

No offense taken, I shouldn’t have taken that tone with you here. You’ve pinged me about it numerous times, in Slack and in issues, and I’ve replied that I’ve not had time. I've also mentioned in one of our team issues a month ago that this isn’t a blocker for v4 and I wouldn’t be actively working on it.

That said, I should’ve been clearer with you. It’s only today that I've phrased it in such a way, and I should've said no earlier to a site generator change before v4 goes stable. That stance hasn’t changed, though. I can look at it in a minor release, but no sooner.

> The site generator doesn't affect the end users.

We both know that isn't true. The Ruby dependency, build tooling for pushing to Pages, ToC problems, redirects, docs and repo structure, etc all affect the folks that use Bootstrap. It clearly affects us more, but the choice does have an impact on every person using, forking, or contributing to Bootstrap.

> 1. I have already done the work

Simply because the work is done doesn't meant it should ship. I appreciate how passionate you are about the change and am very fortunate you took on the challenge, but it's no guarantee we'll move forward.

> 3. Everyone will learn the basics like I also do, it's not so hard nor the learning curve is huge

Something I’ve learned through years of maintaining this project—don’t assume everyone else will learn like you. What’s not so hard for you might be monumental for someone else. I've never used Hugo and found the docs slightly frustrating to follow at first glance. It'll take me some time to familiarize myself, thoroughly test, etc.

> That being said, suit yourself, and stay with a clearly inferior project which clearly slows down development.

Again, you’re being passive aggressive here. It’s not necessary and it hurts us all. Jekyll is great; it’s not perfect, but it’s worked well enough for me as someone who’s not a super experienced developer. I’d love to be rid of it—all I want is HTML, CSS, and JS—but that’ll take time.

Relatedly, and I think I mentioned this in Slack, what are our other options? I know Jekyll and Hugo aren't the only site generators out there. I haven't had much time to dive deeply into them though. You seem to love Hugo and believe it superior, but what else has been thoroughly evaluated?

Long story short, we cannot have this block v4’s release. I don't want to get into additional dev and debugging time for something so essential to the project's development. I want—hell I need—this thing to be a stable.

Hope that helps, and apologies I came at you with a shit attitude.

XhmikosR:
Well, I offered a solution apart from Jekyll, Hugo, because I prefer it. It's faster than most solutions and easy to learn. The Hugo devs even offered to improve their code to help us.

No, the site generator does not affect the end users. The end users, aren't the website visitors.

That being said, my solution here has nothing to do with the solution itself. The approaches here are these:

1. you test my branch, you see what you don't like, come back with specific comments, so  that we can get them sorted and move forward with Hugo
2. stay with Jekyll and all of its issues; toc isn't the only one. Bad performance is another one. The Ruby dep another one. And so on.

I don't think this is the right place to discuss the pros and cons of Hugo. Nor do I have something to gain from switching to Hugo.

I've been using Hugo already, and I honestly believe it will help us, otherwise I wouldn't have made any patch in the first place.

If you want to wait a little bit, that's OK. But in the meantime you can try the branch and comment on the Hugo issue what bothers you so that we can get it implemented. If you try it, I'm pretty sure you will love its speed and simplicity, along with the other pros I have already mentioned.

So, feel free to bug me on Slack, but seriously, the lack of communication is an issue and should not happen between team members.

mdo:
Feedback heard on the communication. Been burnt out from work once again and haven't had time to put into things as much as I'd have liked. I can evaluate the Hugo branch shortly after we get v4 stable out and leave some comments.

PR #25062, opened 2017-12-23
Fix tooltip/popover arrow size and position

### LINKED ISSUE #23793: Tooltip arrows not centered perfectly
The issue is visible on the demo page here: https://getbootstrap.com/docs/4.0/components/tooltips/

If you look closer at the left / right positioning you will notice that arrows are not centered (vertically) perfectly:

issue

### LINKED ISSUE #23846: Arrow out of popover

### LINKED ISSUE #25045: Misaligned popover arrows
Seems similar to #23793, although I could not reproduce the issue for tooltips, but indeed for popovers.

Here is a demo: https://codepen.io/simonihmig/pen/WdxXja. The first button triggers a popover, where you can clearly see that the arrow is misaligned. 

It has `left: 132px` applied by popper.js. However the popover itself has a width of `276px`, so the center should rather be `137px`. And indeed if I set this manually to this value or `left: 50%` in Chrome Dev Tools, then it is perfectly aligned! So somewhere this 5px offset is causing the misalignment.

simonihmig:
This overhauls the tooltip and popover arrows implementation, fixing a few problems:

**1. arrows were not properly aligned to the trigger.**

Fixes #25045 and fixes #23793

**2. the `.arrow` element did not match the size and position of the actual visible arrow.**

Before:
image

After:
image

Same for tooltips. This possibly also fixes #23846, see https://github.com/twbs/bootstrap/issues/23846#issuecomment-327465549.

**3. Fixed `$popover-arrow-height` / `$tooltip-arrow-height`**

These Sass variables did not have any effect on the visual appearance. They were just used for the (wrong, see above) sizing of the `.arrow` element, which has absolutely no visible effect. Only `$...-arrow-width` had a visual effect, but it always determined implicitly both the width and height (height = width/2, so a 45° angle). 

Also the value for width was wrongly used, AFAICT. Setting it to `10px` would lead to an arrow `20px` wide. My understanding of these dimensions is that `$popover-arrow-width` should specify the width of the `.arrow` (for top or bottom placement), as can be seen on the second screenshot above.

This has been fixed here. But for anyone who has customized the arrows with these variables, this would be a somewhat breaking change (the width will be the half of it as it has been before). But still I think this has to be done, as this restores that the assigned values now exactly match the visual dimensions.

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -26,96 +26,88 @@
     display: block;
     width: $popover-arrow-width;
     height: $popover-arrow-height;
+    margin: 0 $border-radius-lg;
   }
 
   .arrow::before,
   .arrow::after {
     position: absolute;
     display: block;
+    content: "";
     border-color: transparent;
     border-style: solid;
-  }
-
-  .arrow::before {
-    content: "";
-    border-width: $popover-arrow-width;
-  }
-  .arrow::after {
-    content: "";
-    border-width: $popover-arrow-width;
   }
 
   // Popover directions
 
   &.bs-popover-top {
-    margin-bottom: $popover-arrow-width;
+    margin-bottom: $popover-arrow-height;
 
     .arrow {
-      bottom: 0;
+      bottom: calc((#{$popover-arrow-height} + #{$popover-border-width}) * -1);
     }
 
     .arrow::before,
     .arrow::after {
-      border-bottom-width: 0;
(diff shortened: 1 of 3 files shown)


Johann-S:
Thank you so much @simonihmig 👍

PR #25063, opened 2017-12-23
Undo #22740, removing break-word from .card

### LINKED ISSUE #24990: Select wrap bug
Building an internal database interface and came across a word-wrap bug for certain select options that are large. Only appears to be an issue with Safari/Chrome. Using v4.0-beta.

This is the code the wsgi/Jinja app puts out, the profiler is a little too long and Safari wraps the \'r\'

```
<select name="type" class="form-control">
  <option value="1">Aquadopp Meter</option>
  <option value="2" selected="">Aquadopp Profiler</option>
  <option value="3">RDI Long Ranger</option>
  <option value="4">RDI Sentinel</option>
</select>
```
I am able to fix it live by adding `word-wrap: normal;` to the `.form-control` class.

### LINKED ISSUE #25235: Progress bar text is wonky when <3% and in .card
When a progress bar is included within a card and its value is less than 3%, the text becomes stacked with the number above the percentage sign. https://jsfiddle.net/gseffeqr/

image

This is occuring in Chrome, Firefox, and Edge.

mdo:
Fixes #24990 and fixes #25235 by undoing #22740.

This needs more thought in regards to what replace this with, but for now I want to at least have a fix for the identified problem. Consider this WIP until we have an alternative.

/cc @pat270


pat270:
Hey @mdo,
I was able to look into this today and it might be a browser bug. Safari doesn't seem to follow the definition of `white-space: pre` when a parent element has `word-wrap: break-word` https://www.w3.org/TR/css-text-3/#white-space-property, specifically "Lines only break at forced line breaks; content that does not fit within the block container overflows it".

`word-break: break-all` with child element `white-space: pre` works by the way.

PR #25069, opened 2017-12-24
Update README.md

sagarp0810:
Twitter Inc. seems more correct!!!!!
Right...


patrickhlauke:
And yet, it's how it's written normally.

PR #25073, opened 2017-12-25
Clean up button group 

ysds:
This PR is reproduction of #24696, includes the followings:
- Remove unnecessary float
    - Since parent `.btn-group` is displayed as inline-flex, the child elements are placed left to right by default.
- Cleanup of reset rounded corners
    - Remove `border-radius :0` since it is enough by each reset of left and right side
    - A `.dropdown-toggle` selector should not be used for reset rounded corner since it does not have `border-radius` property. Instead use `.dropdown-toggle` as negation pseudo-class.

Not include changes to input group. Thanks.

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -68,10 +68,6 @@
 
   // Custom edits for including btn-groups within btn-groups (useful for including
   // dropdown buttons within a btn-group)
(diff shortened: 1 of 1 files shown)


mdo:
You're killing it with these PRs, thanks so much!

PR #25082, opened 2017-12-25
Opinionated: reintroduce `cursor:pointer`

### LINKED ISSUE #23224: Actions vs Links inside of Dropdowns
With the idea that cursors now convey meaning (https://github.com/twbs/bootstrap/pull/21812), the `dropdown` documentation seems to need some love.  The examples in the documentation refer to "actions" and "links", but each `dropdown-item` is just ultimately just an anchor (a link).

I'd be happy to submit a PR to improve the documentation (using `a` for links and `button` for actions with a small paragraph describing intent).

** Edited out link/button font nonsense that's already fixed.

### LINKED ISSUE #24156: Remove hand-cursor on links displayed as button
As explained quite clearly [on this article](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b), hand (pointer) cursor should be limited to links on text and images.

The rule seems to be: for a button, there is no need to use a pointer cursor. for a text link, a pointer cursor should be used.

Currently, the underlying element is not altered: if you render a classic link as a button (`<a>` with `class="btn"`) the cursor is still a hand cursor.
For example, [in the documentation](https://getbootstrap.com/docs/4.0/components/buttons/#button-tags) for the first button. For this button only, the cursor is a pointer.

Also note that the same is true for button elements rendered as links (`<button>` with `class="btn-link"`). The link should be rendered with a hand cursor. See [this example on the doc](https://getbootstrap.com/docs/4.0/components/buttons/#examples).

A naïve implementation could be:
```
.btn-link { cursor: pointer; }
a.btn { cursor: default; }
```

If the solution is as simple as this, I can submit a PR.

Regards,

### LINKED ISSUE #24954: Set `cursor: pointer` for interactive elements
In #21812 @mdo provided links which says that pointer cursor should be used only for links. 

Both [Microsoft's design guidelines](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742466%28v=vs.85%29.aspx) and [Apple's HIG](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/Pointers.html) are related to desktop applications. While [W3C spec](https://www.w3.org/TR/CSS21/ui.html#propdef-cursor) is related to web, it does not follow it's recommendation: all provided sources use cursor pointer for buttons on their web pages, some of them for checkboxes (ex.: [Apple](https://appleid.apple.com/account#!&page=create), [Microsoft](https://login.microsoftonline.com), [W3C](https://www.w3.org/)).

There are some issues with having `cursor: pointer` for links but not for buttons.

1. It's usual to have controls in forms like this:
```html
<button class='btn btn-primary' type='submit'>Save</button>
<a class='btn btn-secondary' href='/back'>Cancel</a>
```

This way both controls looks the same, but have different cursors. This looks inconsistent.

2. `.navbar-toggler` is good example when it's not clear if element is interactive or not: https://getbootstrap.com/docs/4.0/components/navbar/#external-content

### Why desktop application design guides should not be used for web pages?

The main difference is that web-page contains mostly static content (not-interactive) but desktop app usually almost totally interactive (this should be considered in global scale, not for single web-site that can be fully interactive: users use browser to visit different web-sites). So it has become usual to use pointer cursor to help users distinguish interactive elements on webpages. Especially with custom-styled inputs.

Please consider to set back pointer cursor for interactive elements.

patrickhlauke:
Explicitly re-adds the "hand" `cursor:pointer` for non-disabled `.btn`, `.close`, `.navbar-toggler` elements, as well as forcing `page-link` pagination controls to always have the "hand" cursor (even if an author uses it on, say, `<button>` elements for a dynamically updating in-page pagination).

Controversial, as everybody jumped on the bandwagon following this article https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b - which does have its merits of course, but there are also counter-arguments like http://kizu.ru/en/issues/cursor-pointer/

And seeing the amount of issues we've seen following the change, and the potential complexity needed to consistently address the cursor issue (see https://github.com/twbs/bootstrap/issues/24156 where i explore how the cursor needs to be based on "intent", and how that's tough to determine), I'm favouring an opinionated take here of just reintroducing the `cursor:pointer`.

Closes #24954, closes #24156, and closes #23224.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -32,6 +32,11 @@
     @include box-shadow(none);
   }
 
+  // Opinionated: add "hand" cursor to non-disabled .btn elements
+  &:not([disabled]):not(.disabled) {
+    cursor: pointer;
+  }
+
   &:not([disabled]):not(.disabled):active,
   &:not([disabled]):not(.disabled).active {
     background-image: none;
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -12,6 +12,11 @@
     text-decoration: none;
     opacity: .75;
   }
+
+  // Opinionated: add "hand" cursor to non-disabled .close elements
+  &:not([disabled]):not(.disabled) {
+    cursor: pointer;
+  }
 }
 
 // Additional properties for button version
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -116,6 +116,11 @@
   @include hover-focus {
     text-decoration: none;
   }
+
+  // Opinionated: add "hand" cursor to non-disabled .navbar-toggler elements
+  &:not([disabled]):not(.disabled) {
+    cursor: pointer;
+  }
 }
 
 // Keep as a separate element so folks can easily override it with another icon
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -20,6 +20,12 @@
     background-color: $pagination-hover-bg;
     border-color: $pagination-hover-border-color;
   }
+
+
+  // Opinionated: add "hand" cursor to non-disabled .page-link elements
+  &:not([disabled]):not(.disabled) {
+    cursor: pointer;
+  }
 }
 
 .page-item {
@@ -45,6 +51,8 @@
   &.disabled .page-link {
     color: $pagination-disabled-color;
     pointer-events: none;
+    // Opinionated: remove the "hand" cursor set previously for .page-link
+    cursor: auto;
     background-color: $pagination-disabled-bg;
     border-color: $pagination-disabled-border-color;
   }


XhmikosR:
Please fix the linting issues.

Other than that, I'm personally in favor of this, mostly because I'm used to it.

patrickhlauke:
Yeah, linting issue slipped me by. Fixed now :)

mdo (approved):

patrickhlauke:
cheers boss @mdo

johnunclesam:
In Bootsrap 4.1 still doesn't work? Any change?

PR #25087, opened 2017-12-26
Linked card

### LINKED ISSUE #18294: Link entire card v4
I think it would be very useful to be able to link entire cards. We tried doing this in our application, but realised all text got blue which is not ok for us. 

I would expect them to behave similar to linked list group items, and think a card view is a nice alternative to plain lists.

MartijnCuppens:
Fixes #18294:
- Possibility to add `.card-full-link` class to links inside card. This adds a ::before pseudo-element to the link which is stretched over the card.
- Possibility to wrap the content of the card in a link with `.card-full-link` class. No styling is influenced by doing this.


andresgalante:
Hi @MartijnCuppens this is an intresting concept, and I think a full clickable card is a common use case.

To be honest I don't like the class name `.card-full-link` but I'll let @mdo decide on it.

scss/_card.scss
```diff
@@ -61,6 +61,28 @@
...
+  .card > & {
+    display: inherit;
+    flex-direction: inherit;
+    min-width: 0;
+    color: inherit;
+    text-decoration: inherit;
+  }
+
+  :not(.card) > & {
+    &::after{
```
andresgalante: there is a space missing after the `after`

MartijnCuppens:
Thx for the review @andresgalante. The name of the class could indeed be improved.

eugen0188:
Nice work. How about cursor:pointer on the whole card?

MartijnCuppens:
Closing this, #26184 covers this

PR #25095, opened 2017-12-28
Add examples of disabled tooltip and popover triggers

mdo:
Closes #25015 with a dedicated section to both tooltip and popover docs.

screen shot 2017-12-27 at 5 08 07 pm


patrickhlauke:
Except that this way, there's no way for keyboard users to trigger that popover...

patrickhlauke:
xref https://github.com/twbs/bootstrap/pull/25163

PR #25099, opened 2017-12-28
Wall of Browser Bugs: position:sticky support landed in MS Edge

Herst:
Patch to remove it from "Most wanted features" now that `position: sticky` support is in Microsoft Edge 16.

~~(Also includes #24751.)~~

### DIFF
--- a/_data/browser-features.yml
+++ b/_data/browser-features.yml
@@ -22,16 +22,6 @@
   browser: >
     Edge
   summary: >
-    Implement [sticky positioning](http://html5please.com/#position:sticky) from CSS Positioned Layout Level 3
-  upstream_bug: >
-    UserVoice#6263621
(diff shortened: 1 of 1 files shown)


XhmikosR (changes requested):
I don't see why you include the other issue when it has its own PR. Please remove it from this PR.

Herst (commented):
I assumed the changes would be in the same file which is why I took the other PR as a basis.

Herst:
I assumed the changes would be in the same file which is why I took the other PR as a basis.

Herst:
OK, done.

BTW, I would be glad if #24751 got merged in before the final release (assuming there is still no further reaction from the Chromium team's side).

XhmikosR (approved):

PR #25101, opened 2017-12-28
Extend spacers and sizes

unrevised6419:
Extend `$spacers` and `$sizes` like `$theme-colors` and other maps.  

Used `// stylelint-disable` because `stylelint` complains about map on single line.

ps: Can `$grid-breakpoints` and `$container-max-widths` also be extended like this?


mdo:
I suppose all our Sass maps should behave the same way—it'd make documentation easier. The only problem is removing values from the maps this way—you have to use `map-remove`. Not the end of the world, but worth noting here.

scss/_variables.scss
```diff
@@ -114,23 +114,27 @@ $enable-print-styles:       true !default;
 // variables. Mostly focused on spacing.
 // You can add more entries to the $spacers map, should you need more variation.
 
+// stylelint-disable
```
XhmikosR: Can't we be more explicit with what we disable here?

iamandrewluca: As in case for other empty maps. Stylelint does not allow empty maps ex:
```sass
// stylelint-disable
$grays: () !default;
// ...
$colors: () !default;
// ...
$theme-colors: () !default;
// ...
// stylelint-enable
```
https://github.com/twbs/bootstrap/pull/25101/files/17a992f414d1c6963320cd1d877d205c85e8b82c#diff-d8ee409a461718bfb6240710c8c73382R11

XhmikosR: No worries, I remembered I had to do the same there when we switched to stylelint.

mdo (approved):

karengrigoryan:
To add/modify values in `$grid-breakpoints` and `$container-max-widths` we still need to copy the whole map to `custom.scss` in **4.0.0**. Any chance to extend them like other maps?

iamandrewluca:
For now you can `map-merge` after default vars import.

iamandrewluca:
@karengrigoryan https://github.com/twbs/bootstrap/pull/25392

PR #25104, opened 2017-12-28
Expand all docs sidebar links

mdo:
Revamps docs left sidebar to show all groups' nav items for easier docs browsing.

screen shot 2017-12-28 at 11 34 05 am


XhmikosR:
Are you sure this is better? I find the current way simpler to my eyes, at least.

mdo:
@XhmikosR I was navigating around fixing some stuff up yesterday and was annoyed when I had to click one section, wait for the wrong page to load, and then click another. For example, I click Components, am brought to Alerts, and then have to click Navbar.

Dunno if it's best yet, and the search does take care of most of this, but figured I'd see what the reason was to something like this.

XhmikosR:
That is a valid point indeed. How about we just make the root category
expandable with a click but collapsed by default?

It's something I was thinking the other day.

XhmikosR:
Or maybe just show the current group only and the rest just collapsed
instead of a tags.

mdo:
Punting on this.

PR #25106, opened 2017-12-28
Docs: explicitly apply CSS `::before` styles for syntax highlighting just to `<code>` element

patrickhlauke:
Currently, the classes are added to both `<pre>` and `<code>`, resulting in a doubling-up of the `::before` content.

See for instance http://getbootstrap.com/docs/4.0/getting-started/download/

capture

### DIFF
--- a/assets/scss/_syntax.scss
+++ b/assets/scss/_syntax.scss
@@ -64,14 +64,14 @@
 .css .o + .nt,
 .css .nt + .nt { color: #999; }
 
-.language-bash::before,
-.language-sh::before {
+code.language-bash::before,
+code.language-sh::before {
   color: #009;
   content: "$ ";
   user-select: none;
 }
 
-.language-powershell::before {
+code.language-powershell::before {
   color: #009;
   content: "PM> ";
   user-select: none;


mdo:
https://cl.ly/2F3c3q3s0u3H/Screen%20Shot%202017-12-28%20at%201.06.39%20PM.png

mdo:
Is your, or my, cache stale? I don't see this.

screen shot 2017-12-28 at 1 06 39 pm

patrickhlauke:
@mdo strange. i only see it happening in chrome. looking at devtools, there the `language-sh` is applied to both the `<pre>` and the nested `<code>`. in edge and firefox that isn't the case.

patrickhlauke:
this also seems to mess up code highlighting in other instances. only happens in chrome stable, not canary...very odd

patrickhlauke:
can't pinpoint what it is, but appears to be something at my end. in the same stable chrome instance, in an incognito window, it behaves correctly. could be one of my extensions misbehaving, perhaps.

patrickhlauke:
turns out it was the aXe chrome extension causing it for some reason https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd - the highlighting code seemed to run twice (slowing the connection i could see it being applied once, correctly, and then the second time, breaking things and adding the doubled-up `::before`)

/cc @marcysutton

marcysutton:
@patrickhlauke there's a new version out today that should fix that issue, can you see if it persists? I'm not seeing the issue original on my machine with it installed.

PR #25108, opened 2017-12-28
Update Theming docs for removing map items

### LINKED ISSUE #24849: Overriding $theme-colors
Using v4.0.0-beta.2, I found out something changed with map of theme colors.

Now default `$theme-colors` is merged with custom one, no matter what.
But what if I don't need _success_, _info_, _danger_, etc… for background or text utility classes?

mdo:
Fixes #24849


XhmikosR (approved):

jbmikk:
I don't understand how to remove a variant.
If can't remove them before importing bootstrap, and if I remove them after importing the variants get compiled anyway. What's the point of remove a variant for the map if all variants have already been generated?

PR #25109, opened 2017-12-28
Docs: mark up contents ASCII trees as semantically meaningful nested lists

patrickhlauke:
Possibly a work in progress, but: currently, we use ASCII art trees to show directory structures. This is completely nonsensical/meaningless to non-visual (e.g. screenreader) users. We either need to provide an alternative, or try and provide some semantic structure. This PR formalizes an idea I had a while ago https://codepen.io/patrickhlauke/pen/azbYWZ to use nested lists and generated CSS content to mimic the current look, while having proper structure.

It could probably do with more styling to make it a block/have proper background, and to get the correct monospace font applied to, say, `<pre>` ... but putting this here as an initial exploration.

### DIFF
--- a/assets/scss/_ascii-tree.scss
+++ b/assets/scss/_ascii-tree.scss
@@ -1,0 +1,40 @@
+// stylelint-disable selector-max-type, selector-max-combinators, selector-max-compound-selectors
+
+// ASCII tree
+//
+// Styles to allow for traditional ASCII trees to be marked up as semantically meaningful nested lists.
+
+.bd-ascii,
+.bd-ascii ul {
+  padding-left: 0;
+  margin-left: 0;
+  font-family: monospace;
+  list-style: none;
+}
+.bd-ascii li {
+  padding: 0;
+  margin: 0;
+}
+
+/* level 1 */
+.bd-ascii > li::before { content: ""; }
+
+/* level 2 */
+.bd-ascii > li > ul > li::before { content: "├──\00a0"; }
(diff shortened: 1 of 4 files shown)


patrickhlauke:
Comparison of the ASCII approach (left) and the new approach (right):

semanticascii

mdo:
Huh, I've always wondered about this. 😆 Props for digging in and finding a solution. Looking at the amount of CSS though, maybe we switch to a table and describe the folder contents that way?

patrickhlauke:
@mdo 
> Looking at the amount of CSS
I think we can drop the "level 4" part of the CSS (left it in there to be a bit more future-proof just in case). In the grand scheme of things it doesn't look too excessive to me.

> maybe we switch to a table and describe the folder contents that way?
how would you convey the nested hierarchy in a data table?

mdo:
> how would you convey the nested hierarchy in a data table?

I think the only thing we need to convey is that there is `css/` and `js/`, and there are multiple bundles in each of those. For example:

| Directory | Description |
| --- | --- |
| `css/` | Compiled CSS, minified CSS, and source maps for our default, grid, and Reboot bundles |
| `js/` | Compiled and minified JavaScript plugins with an optional bundle that includes our dependencies   |

Descriptions would need work mind you.

patrickhlauke:
table like that would start getting difficult for anything with more levels, like http://getbootstrap.com/docs/4.0/getting-started/contents/#bootstrap-source-code and http://getbootstrap.com/docs/4.0/getting-started/theming/#file-structure though

PR #25116, opened 2017-12-29
Remove unnecessary `align-middle` class from Navbar docs

andresgalante:
The [forms docs on Navbar](http://getbootstrap.com/docs/4.0/components/navbar/#forms) has an example of buttons with an `align-middle` utility class.

I think that a `vertical-align` in this context is unnecessary.


mdo (approved):
Correct!

PR #25117, opened 2017-12-29
Removes hardcoded value for .nav-link padding

andresgalante:
`.nav-link` padding within navbar is hardcoded to `.5rem`.

@mdo I've created a new variable for this value, but I was unsure if we should use `$nav-link-padding-x / 2` What do you think?


mdo (approved):
No need for basing this on `$nav-link-*` variables.

PR #25121, opened 2017-12-29
Allow nested structure for accordions

### LINKED ISSUE #25119: Collapse and two columns layout
I am trying to use one accordion for two columns layout and this is only working in bootstrap 4 alpha and is NOT working since any beta release (tried even with the latest beta 3).

## Steps to reproduce with CodePen examples
Take a look at this [codepen](https://codepen.io/starkskalle/pen/YYVPpw) - once you click on any collapsible element it stays collapsed forever unless you click on it again. However the expected behaviour is that only one element should be collapsed at a time.

Take a look at the very same [HTML markup](https://codepen.io/starkskalle/pen/mpmymO) using Bootstrap Alpha 6. Everything is working as expected. You are not able to have more than a one element collapsed.

## Related issues
There is a similar issue #24810 but I am not sure whether it is related or not.

## Environment

- Windows 7 64 bit
- Chrome version 63 (64 bit)

MartijnCuppens:
Fixes #25119


Johann-S (changes requested):
Please add a unit test

MartijnCuppens:
- Simplified the documentation (data-children attribute was not used in js and is not needed anymore)
- Added a unit test for nested elements
- Fixed existing tests where the data-parent was set on the link and not the collapsible content

docs/4.0/components/collapse.md
```diff
@@ -63,7 +63,7 @@ Multiple `<button>` or `<a>` can show and hide an element if they each reference
 
 ## Accordion example
 
-Using the [card]({{ site.baseurl }}/docs/{{ site.docs_version }}/components/card/) component, you can extend the default collapse behavior to create an accordion.
+You can extend the default collapse behavior to create an accordion.
```
XhmikosR: Hmm, this doesn't seem right.

Johann-S: Agree with @XhmikosR

Johann-S (approved):

PR #25123, opened 2017-12-29
Darken $gray-600 by 10% for AA contrast

### LINKED ISSUE #23319: Inadequate color contrast
We have many issues even with WCAG2AA. I think we should try to fix some if not all of the issues.

Some examples I found in our docs homepage:

`.text-muted`
`.btn-outline-secondary`

/CC @patrickhlauke

mdo:
Closes #23319. Both `.text-muted` and `.btn-outline-secondary` (and indeed all secondary items) make use of `$gray-600`. New value provides a contrast ratio of 4.69.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -15,7 +15,7 @@ $gray-200: #e9ecef !default;
 $gray-300: #dee2e6 !default;
 $gray-400: #ced4da !default;
 $gray-500: #adb5bd !default;
-$gray-600: #868e96 !default;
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

patrickhlauke:
for reference, see https://github.com/twbs/bootstrap/issues/25126

patrickhlauke:
While this addresses the most egregious (and widely used) color, note that there are still many combinations in use that fail AA 4.5:1 ratio.

PR #25139, opened 2017-12-30
Add `docs-github-serve` script for local serving the github target.

XhmikosR:
This is mostly so that we can quickly test the github build without having to use other ways.

It assumes `docs-github` has been run, and it doesn't watch for changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "docs-compile": "bundle exec jekyll build",
     "postdocs-compile": "npm run docs-workbox-precache",
     "docs-github": "shx echo \"github: true\" > twbsconfig.yml && npm run docs-compile -- --config _config.yml,twbsconfig.yml && shx rm ./twbsconfig.yml",
+    "docs-github-serve": "bundle exec jekyll serve --skip-initial-build --no-watch",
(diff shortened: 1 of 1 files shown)


mdo:
> This is mostly so that we can quickly test the github build without having to use other ways.

Is this referring to my ship checklist which mentions starting a Python server?

XhmikosR:
I don't recall what you are using but if you are firing up a Python server
to test the GitHub build then yeah this should take care of that.

mdo:
Yeah, because of our plugins and what not, we can't do a straight `git push origin gh-pages`—we need to copy-pasta the code over. My `build/ship` script accomplishes this with a `cp` from `bootstrap` to a sibling clone of the repo, `bs-docs`. I switch over to that, run a Python server because there's only compiled files here, and verify some changes.

This wouldn't change that process for me though since the `_config.yml` and `build/` aren't present in the `gh-pages` branch. I don't think I personally have a need for this one right now—is there another situation where this is required?

XhmikosR:
Yeah I like to test the production build without having to resort to other
tools locally.

Plus I might put something together later with browser sync instead of
jekyll's watch server.

mdo (approved):

PR #25141, opened 2017-12-30
Remove Entypo from "Preferred Icon Sets".

raajtram:
The download link for Entypo (on its website) is invalid. It's a Dropbox link which now results in a 403 error. The website was last updated in 2015 (as seen in the footer).


raajtram:
Here's Entypo - http://www.entypo.com/.

XhmikosR (approved):

PR #25142, opened 2017-12-30
Overflow issue IE

### LINKED ISSUE #23172: overflow on Internet Explorer 10
So I was trying v4-dev docs locally with Internet Explorer 10 and I noticed this:

28716442-f6cf1aa4-73a5-11e7-9dc1-c05c7a5fcf48

I narrowed down to `.bd-docs` column. Also, the issue goes away if I remove `.flex-xl-nowrap` from the parent row.

On a project of mine I have switched to v4, I also have this issue in a row like this:

```html
.container-fluid.mt-6
  .row.justify-content-center
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
    .col-12 col-sm-6 col-lg-3 mx-auto card-deck
```

Probably not the same issue since this one goes away if I remove `justify-content-center`. If that's confirmed I should make a new issue for this to track it.

/CC @Johann-S @mdo

MartijnCuppens:
Removes scrollbar in IE10 and IE11.

Fixes #23172

Max-width of 320px is not respected in IE10 on screens wider as 1920px, but imo this is an edge case and the difference is hardly noticeable.


XhmikosR (approved):

XhmikosR:
@mdo: I can confirm this fixes the issue.

mdo (approved):

PR #25143, opened 2017-12-30
Validation tooltip example

### LINKED ISSUE #24811: Add example of .valid-feedback, .valid-tooltip and .invalid-tooltip
The example of `.valid-feedback`, `.valid-tooltip` and `.invalid-tooltip` does not seem to be in the document. Should we add these examples? especially it would be useful to clarify how to use invalid/valid-tooltip. Thanks.

Related SCSS:
https://github.com/twbs/bootstrap/blob/8df34b075a1ffffcc90d37fc44fac07c4c5d5791/scss/mixins/_forms.scss#L31-L51

mdo:
Fixes #24811.

This adds `.valid-feedback` to our custom styles and server side examples; previously we ommitted this to suggest you don't always need valid feedback. In addition, this adds examples of the `.{valid|invalid}-tooltip` classes with a new subsection in the Validation docs.

This also update the validation tooltip styles to remove the fixed `width`; instead it should retain itself to the parent element.

/cc @ysds @tmorehouse

### DIFF
--- a/docs/4.0/components/forms.md
+++ b/docs/4.0/components/forms.md
@@ -724,15 +724,21 @@ For custom Bootstrap form validation messages, you'll need to add the `novalidat
 When attempting to submit, you'll see the `:invalid` and `:valid` styles applied to your form controls.
 
 {% example html %}
-<form id="needs-validation" novalidate>
+<form class="needs-validation" novalidate>
   <div class="form-row">
     <div class="col-md-4 mb-3">
       <label for="validationCustom01">First name</label>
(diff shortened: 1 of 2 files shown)


ysds:
Thanks. It is more helpful if documented that parent of `.{valid|invalid}-tooltip` needs to have `position: relative` like http://localhost:9001/docs/4.0/components/dropdowns/#examples
> Wrap the dropdown’s toggle (your button or link) and the dropdown menu within .dropdown, or another element that declares position: relative;.

mdo:
@ysds Added!

PR #25146, opened 2017-12-31
Improve ESLint rules

XhmikosR:
Not sure about the new files, but this way one can run `eslint .` and it will just work.


js/src/dropdown.js
```diff
@@ -324,10 +318,10 @@ const Dropdown = (($) => {
 
       const toggles = $.makeArray($(Selector.DATA_TOGGLE))
       for (let i = 0; i < toggles.length; i++) {
-        const parent        = Dropdown._getParentFromElement(toggles[i])
-        const context       = $(toggles[i]).data(DATA_KEY)
+        const parent = Dropdown._getParentFromElement(toggles[i])
```
Johann-S: Not big fan with this change 😟 but that's my personal feeling I think it's more readable like it was before that change

Johann-S (approved):

js/src/dropdown.js
XhmikosR: Hmm, I didn't enforce any rules about spacing after `=`, so let me have a look again. I had enabled it at first but then dropped it.

Issue #24918, opened 2017-11-30
Tooltips do not work on elements inside Shadow DOM


ericcarino:
I'm trying to add a tooltip to a button that resides inside a shadow DOM.  Unfortunately, does not work.  I traced the problem down to this snippet of code inside bootstrap.js.

```js
Tooltip.prototype.show = function show() {
      var _this22 = this;

      if ($(this.element).css('display') === 'none') {
        throw new Error('Please use show on visible elements');
      }

      var showEvent = $.Event(this.constructor.Event.SHOW);
      if (this.isWithContent() && this._isEnabled) {
        if (this._isTransitioning) {
          throw new Error('Tooltip is transitioning');
        }
        $(this.element).trigger(showEvent);

        var isInTheDom = $.contains(this.element.ownerDocument.documentElement, this.element);

        if (showEvent.isDefaultPrevented() || !isInTheDom) {
          return;
        }
```

_isInTheDom_ is false because jquery.contains doesn't find the element, which I assume is because it resides inside the shadow DOM.  If I manually set this value to true, the tooltip works perfectly.  See my screenshot below.

![Example](https://puu.sh/ywkBk/b4307b369b.png)

To access my JSBin example, use the link below.
[http://jsbin.com/qukeye/2/edit?html,js,output](http://jsbin.com/qukeye/2/edit?html,js,output)

Johann-S:
That's true as we use jQuery and jQuery access to the DOM, Currently Bootstrap do not handle Shadow DOM.

I made a CodePen with our latest release (beta 2) : https://codepen.io/Johann-S/pen/YERLxV

Johann-S:
It's totally doable to fix our JS to allow folk to use Bootstrap in shadow DOM, but it seems our CSS do not work 😟 

/CC @mdo @andresgalante @XhmikosR @bardiharborow

andresgalante:
The web is going toward Webcomponents, I think it's important to be able to implement bootstrap with the ShadowDom

ericcarino:
@Johann-S What is the CodePen supposed to demonstrate?  Were you able to get the tooltip to work?  I've tried on Chrome and Safari and I don't see it working.

Johann-S:
Nothing but it's better to use the latest release which do not use Tether 😂

LMOkino:
Will this problem be fixed? If so, when do you think it will be fixed? Our product would like to use this feature if it gets fixed soon.

Johann-S:
[...]

PR #25150, opened 2017-12-31
Allow to use Tooltips/Popovers in Shadow DOM

Johann-S:
Fixes : #24918

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -244,8 +244,9 @@ class Tooltip {
     if (this.isWithContent() && this._isEnabled) {
       $(this.element).trigger(showEvent)
 
+      const shadowRoot = Util.findShadowRoot(this.element)
       const isInTheDom = $.contains(
-        this.element.ownerDocument.documentElement,
+        shadowRoot !== null ? shadowRoot : this.element.ownerDocument.documentElement,
         this.element
       )
 
--- a/js/tests/visual/tooltip.html
+++ b/js/tests/visual/tooltip.html
@@ -62,7 +62,12 @@
           </button>
         </p>
       </div>
-      <div id="target" title="Test tooltip on transformed element"></div>
+      <div class="row">
+        <div class="col-sm-3">
+          <div id="target" title="Test tooltip on transformed element"></div>
+        </div>
+        <div id="shadow" class="pt-5"></div>
+      </div>
       <div id="customContainer"></div>
     </div>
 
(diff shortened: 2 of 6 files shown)


Johann-S:
ping @XhmikosR

js/tests/visual/tooltip.html
```diff
@@ -57,6 +62,21 @@ <h1>Tooltip <small>Bootstrap Visual Test</small></h1>
...
       $(function () {
+        if (typeof document.body.attachShadow === 'function') {
+          var shadowRoot = $('#shadow')[0].attachShadow({ mode: 'open' })
+          shadowRoot.innerHTML = [
+            '<button type="button" class="btn btn-secondary" data-toggle="tooltip" data-placement="top" title="Tooltip on top in a shadow dom">',
+              'Tooltip on top in a shadow dom',
+            '</button>',
+            '<button id="secondTooltip" type="button" class="btn btn-secondary" data-toggle="tooltip" data-placement="top" title="Tooltip on top in a shadow dom with container option">',
+              'Tooltip on top in a shadow dom',
+            '</button>'
```
XhmikosR: Any reason you use an Array instead of a string?

XhmikosR:
@Johann-S: I'm not familiar with shadow DOM at all, so I'm not sure I can be of any help...

js/tests/visual/tooltip.html
Johann-S: For me it's more readable but that's a matter of preference

XhmikosR:
I'd rather we are consistent. Plus eventually we should use template
literals.

Johann-S:
This is a live example : https://codepen.io/Johann-S/pen/ZvxgLZ

if @ericcarino or @LMOkino can take a look and see if it works as you wanted it would be perfect 👍

LMOkino:
Johann, it looks good. Please let us know when it is in the bootstrap code.

Johann-S:
@mdo if you have time to check that change, it seems it works pretty well

mdo:
Is this intentional?

screen shot 2018-01-15 at 9 34 22 pm

Johann-S:
Yep it's to show that if you use `container` option to append into the Shadow DOM you'll loose Bootstrap's style but if you include Bootstrap CSS in your Shadow DOM it should be fine

js/src/tooltip.js
```diff
@@ -254,8 +254,9 @@ const Tooltip = (($) => {
       if (this.isWithContent() && this._isEnabled) {
         $(this.element).trigger(showEvent)
 
+        const shadowRoot = Util.findShadowRoot(this.element)
         const isInTheDom = $.contains(
-          this.element.ownerDocument.documentElement,
+          shadowRoot !== null ? shadowRoot : this.element.ownerDocument.documentElement,
```
thejohnfreeman: This code is trying to find the "root" `Element`, either `document.documentElement` or a `ShadowRoot`, but it is split into two pieces: the `findShadowRoot` function, which may return `null`, and this ternary, which will handle `null`. Why not just fold these two pieces into one function whose purpose is to find the root element?

```js
// Returns the root Element, either document.documentElement or a ShadowRoot.
findRootElement (element) {
  // Go straight to the root node if the browser supports this method.
  if (typeof element.getRootNode === 'function') {
    const node = element.getRootNode()
    return (node instanceof ShadowRoot) ? node : node.documentElement
  }

  // Walk the parent chain to the root node.
  while (element.parentNode) {
    // Check for shadowRoot on the way up.
    if (element instanceof ShadowRoot) {
      return element
    }
    element = element.parentNode
  }
  // At this point, element has no parentNode. Thus, it is the root element.

  return element
}
```

Johann-S: I agree that's better like the way you suggested 👍

Johann-S:
@XhmikosR I rebased here, and made some changes, maybe we should add information in our docs ?

XhmikosR:
Yup, we should mention it in docs. Can you add some basic info?

Johann-S:
Done 👍

XhmikosR (approved):

PR #25152, opened 2017-12-31
input group addon support for textareas

### LINKED ISSUE #25102: Input-group-text doesn't expand
Input-group-text doesn't expand when form-control is higher

https://codepen.io/danijelGombac/pen/MrpXzj

gijsbotje:
The append and prepend classes vertically aligned items in the center
For textareas this meant the addon would aling vertically instead of stretch and have it's contents vertically aligned
these changes fix that so everything is aligned how it should be

Fixes #25102

### DIFF
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -60,7 +60,6 @@
 .input-group-prepend,
 .input-group-append {
   display: flex;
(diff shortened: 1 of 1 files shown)


gjp3384:
Good

PR #25159, opened 2018-01-01
Fixed color-yiq variable in button mixin

### LINKED ISSUE #25158: 4.0.0-beta.2 => 4.0.0-beta.3 - outline buttons active state incorrect
I can't seem to find any documentation as to why the active state color was changed for buttons such as `.btn-outline-secondary`.  The contrast on active state is horrible and this seems recently changed between beta version 2 and 3.

https://getbootstrap.com/docs/4.0/components/buttons/#outline-buttons (press in / trigger active state on the buttons)

the desired result should be what was in beta version 2 where the text color is white on active state/press.

the only change seems to be https://github.com/twbs/bootstrap/blob/v4.0.0-beta.2/scss/mixins/_buttons.scss#L80 and https://github.com/twbs/bootstrap/blob/v4.0.0-beta.3/scss/mixins/_buttons.scss#L82

I think there's a typo, shouldn't `color: color-yiq($color-hover)` actually be `color: color-yiq($active-background)`?

niftylettuce:
Fixes #25158.


andresgalante (approved):
Hi @niftylettuce Thanks for the contribution

You are right it should be based on `$active-background` but notice that `$active-background: $color` so it wont make any difference in the resulting color

niftylettuce:
I think you are incorrect @andresgalante - the `color-yiq` will make this text white because the value is different for outlined button variants (e.g. `btn-outline-xyz` - there are different values passed for these variables).

Can this please be merged?  This was a breaking CSS change afaik.

niftylettuce:
Hmm..  This is odd -- where is the actual breaking change at?  You're right @andresgalante it seems those variables aren't ever actually explicitly passed to this...

```scss
@each $color, $value in $theme-colors {
  .btn-outline-#{$color} {
    @if $color == "light" {
      @include button-outline-variant($value, $gray-900); // <--- Note it only passes 2 args
    } @else {
      @include button-outline-variant($value, $white); // <--- Note it only passes 2 args
    }
  }
}
```

niftylettuce:
To clarify what this bug is visually, here's a little animated GIF recording.  This is a comparison of https://v4-alpha.getbootstrap.com/components/buttons/ vs. https://getbootstrap.com/docs/4.0/components/buttons/#outline-buttons.  The focus state causes the text to have the incorrect contrast text color.

kapture 2018-01-04 at 11 10 26

kapture 2018-01-04 at 11 09 50

jnizet:
@niftylettuce unless I'm missing something, won't this change make things even worse? Your demo shows that the hover color is correct (white on dark gray background) whereas the active color is not (black on dark gray background), but your PR changes the :hover color, and not the active one.

Shouldn't the color on line 82 be set to $color instead of $active-background, just like it is in the :hover state?

jnizet:
Anyway, the issue seems to have been fixed already anyway (see https://github.com/twbs/bootstrap/commit/43c551e97d78f0145e4eaec2ffda09e49dbeb2a8).

niftylettuce:
isn't that commit you linked fot dropdowns?  I don't think its fixed yet.

jnizet:
No. It's a change in the `@mixin button-outline-variant`. And building and deploying the docs from master locally, I don't reproduce the issue anymore.

andresgalante:
@jnizet is right, it is fixed on `v4-dev` via https://github.com/twbs/bootstrap/commit/43c551e97d78f0145e4eaec2ffda09e49dbeb2a8

@niftylettuce Thanks!

PR #25161, opened 2018-01-01
Multi accordion support + unit test

### LINKED ISSUE #24810: Collapse: Multiple Targets Accordion only works once, then fails
I created a collapse with multiple targets in an accordion-style setup (set data-parent).
The switching works well, exactly once.
The second group of targets then has the `show` class removed and never added again (only after page reload

## Steps to reproduce

Check the codepen sample below and click  the numbered collapse links from 2-3, then click on 1 again.
The second group switches perfectly when clicking 2&3, when clicking on 1 again it fails.

# environment

- Operating system and version Linux (Manjaro, Arch-Based
- Chrome Version 62.0.3202.94 (Official Build) (64-bit)
- Example, based on he official docs: https://codepen.io/sweisgerber-dev/pen/mqqBbK

MartijnCuppens:
Fixes #24810.


Johann-S:
Nice PR @MartijnCuppens I'll take time to review it asap

XhmikosR (approved):
LGTM from a quick look. @Johann-S please verify when you can.

Johann-S:
Is it possible to see a CodePen with your feature @MartijnCuppens ? please

MartijnCuppens:
Her you go @Johann-S: https://codepen.io/MartijnCuppens/pen/YYxpqW?editors=1000

Shall I also include an example in the docs?

Johann-S:
I picked up this code from our documentation and your change break the current behavior see : https://codepen.io/Johann-S/pen/PEKbdR?editors=1000

The example is here in our documentation : https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example
(the second one)

An example or a quick note would be rad 👍

MartijnCuppens:
@Johann-S , doesn't `data-parent` need to be set on the element with the `.collapse` class?

XhmikosR:
If it's breaking the current code we should make sure we add related tests
and second postpone merging this for the stable...

Johann-S:
Nope my bad this example is incorrect we should remove that example which isn't relevant anymore https://github.com/twbs/bootstrap/pull/22251 (since we removed the use or `data-children`)

Johann-S (approved):

MartijnCuppens:
@Johann-S, I already took care of that ;) : https://github.com/twbs/bootstrap/pull/25121/files

Johann-S:
Nice thank you @MartijnCuppens 👍

PR #25162, opened 2018-01-01
Remove incorrect roles from accordion, tweak/expand accessibility info, use buttons for accordion

patrickhlauke:


patrickhlauke:
Before/after with changing the `<a>`s to actual `<button>` elements. As this is the more correct thing to do, we might as well lead with a good example (rather than links with `role="button"` all the time)

accordion

mdo:
Nice!

@andresgalante I know you have a PR for improving accordion borders—can you check your changes against this PR to see if that one is still good to go? I'd like to tackle that one in 4.0 or 4.1.

andresgalante:
@mdo I've just rebased #23839 with this branch, thanks for the heads up!

patrickhlauke:
cool, so assume it's good to merge, right?

andresgalante (approved):

PR #25164, opened 2018-01-02
Print tweaks

### LINKED ISSUE #23864: Problem printing pages on IE vs Chrome
I'm having trouble designing my pages so that they can be printed consistently accross browsers. I noticed that bootstrap homepage (in v3) itself does not print the same way for Chrome and IE.

I'm on Windows 10, using Chrome 60 and IE 11. When I print Bootstrap home page, i get the following results:

Chrome:
image

IE:
image

As it is possible to see, when printing in Chrome it is possible to fit a lot more into the page. This happens consistently for all pages built in Bootstrap. I've tested with v4 homepage and the issue still happens.

Is there a way to overcome this, or is this an issue?

### LINKED ISSUE #23992: Printing Problems
Having a variety of issues with printing and created a test case to help explain it, with some PDF samples.  Using the v4 beta as can be seen in my print-demo.txt (rename to .html and you can view this in any browser).  Designed to be a test case which spans a number of pages for ease of debugging.

[print-demo.txt](https://github.com/twbs/bootstrap/files/1312376/print-demo.txt)

-SOME- of these can be fixed with uncommenting the media print stylesheet in the heading, but this causes other display problems of course with various browsers.

**Firefox v55:**
 Mostly Blank page 2
 Three full flex boxes
 Three cut off and missing headers
[print-demo firefox.pdf](https://github.com/twbs/bootstrap/files/1312318/print-demo.firefox.pdf)

**Chrome v61 (printing from full width):**
 Correctly displays the six flex boxes
 Stops after the first of the 3 cards
 Does not display two of the boxes
 Does not display the footer at all
[print-demo chrome.pdf](https://github.com/twbs/bootstrap/files/1312319/print-demo.chrome.pdf)

**Chrome v61 printing from a narrow width (by resizing the browser very narrow and then printing the page):**
 Seems to do everything correctly
[print-demo chrome (narrow).pdf](https://github.com/twbs/bootstrap/files/1312317/print-demo.chrome.narrow.pdf)

mdo:
Adds two new variables for printing—page `size` and `min-width` for the `<body>` and `.container`. Using the docs homepage as my example, this printed the most similar in macOS Chrome and Safari. I'll check in on Windows browsers before merging.

While working on this, a few things came to light:

- Safari doesn't respect `@page`'s `size`, hence the `min-width`s.
- Chrome is super finicky with `@page`. Adding `margin` or `!important` broke the `size` property.

I'm expecting even more issues across other browsers as I get into them.

Hopefully fixes #23864 and fixes #23992.

/cc @tigermarques @michaelkrieger

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -82,6 +82,11 @@
 
     // Bootstrap specific changes start
 
+    // Specify a min-width to make printing more consistent across browsers
+    body {
+      min-width: $print-body-min-width;
+    }
+
     // Bootstrap components
     .navbar {
       display: none;
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -83,9 +83,17 @@
 
     // Bootstrap specific changes start
 
-    // Specify a min-width to make printing more consistent across browsers
+    // Specify a size and min-width to make printing closer across browsers.
+    // We don't set margin here because it breaks `size` in Chrome. We also
+    // don't use `!important` on `size` as it breaks in Chrome.
+    @page {
+      size: $print-page-size;
+    }
     body {
-      min-width: $print-body-min-width;
+      min-width: $print-body-min-width !important;
+    }
+    .container {
+      min-width: $print-body-min-width !important;
     }
 
     // Bootstrap components
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -20,9 +20,10 @@
       box-shadow: none !important;
     }
 
-    a,
-    a:visited {
-      text-decoration: underline;
+    a {
+      &:not(.btn) {
+        text-decoration: underline;
+      }
     }
 
     // Bootstrap specific; comment the following selector out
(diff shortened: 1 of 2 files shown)


tigermarques:
Hi,

Great news! You mention that this should make Chrome and Safari look similar. The problem I had was with Chrome vs IE. Is it likely that this issue will be fixed too?

Thank you!

mdo:
Just verified this improves things in IE11 and Edge.

nafg:
To clarify, it's supposed to be a smaller paper, so I had `width: 5.5in` on the `.container`. Now it gets disregarded because of the min-width.

browner12:
I'll add here that this caused Chrome's print dialog to remove the "Orientation" option for me.

coorasse:
Chrome margins are wrong. why is this set to `a3` and not `auto` by default? I really appreciate to have this variable now, but the default should simply be auto. we fix IE and break Chrome? 😕

PR #25165, opened 2018-01-02
Add new Approach docs page

mdo:
Super WIP, but the page exists with no outline at all. Adding one here to at least capture a few things.


mdo:
Pushed an extensive update here after the first outline. Would love feedback from @twbs/team on what else we should put here.

XhmikosR:
I think it's a pretty good first approach 👍

docs/4.0/extend/approach.md
```diff
@@ -1,5 +1,70 @@
...
+- Whenever possible, avoid enforcing strict HTML requirements (children selectors)
+
+## Responsive-up
+
+- sometimes called mobile-first, but that can be a misnomer
+- styles apply to a min-width and up
+
+## Base and modifier classes
+
+- Components should be built with a base class that houses common, not-to-be overridden property-value pairs.
```
andresgalante: Should we add a comment about when to create a modifier class and when to just create a new component. I'd say that if we need to modify the color (like primary, warning, etc) or the scale (like lg or sm) of the component then we generate a modifier, if it's anything else then it's a new component.

docs/4.0/extend/approach.md
```diff
@@ -1,5 +1,70 @@
...
+The same goes for more complex components. While we *could* write our own form validation plugin to add classes to a parent element based on an input’s state, thereby allowing us to style the text say red, we prefer using the `:valid`/`:invalid` pseudo-elements every browser provides us.
+
+## Utilities
+
+Utility classes—formerly helpers in Bootstrap 3—are a powerful ally in combatting CSS bloat and poor page performance. A utility class is typically a single, immutable property-value pairing expressed as a class (e.g., `.d-block` represents `display: block;`). Their primary appeal is speed of use while writing HTML and limiting the amount of custom CSS you have to write.
+
+Specifically regarding custom CSS, utilities can help combat increasing file size by reducing your most commonly repeated property-value pairs into single classes. This can have a dramatic effect at scale in your projects.
+
+## Flexible HTML
+
```
andresgalante: Should we also mention that we never do type selectors, and elements are styled through classes.

XhmikosR:
@mdo: is this ready to merge?

XhmikosR (approved):

patrickhlauke (approved):
In future, perhaps something generic about Bootstrap trying to be a generic foundation, rather than trying to solve every possible use case. That authors need to still make sure they use good semantic markup. And cross-reference to the accessibility page? But as a starting point, this is great.

XhmikosR:
@mdo: anyway, I'll leave this for you to merge, just please make sure you squash when merging.

mdo:
Was reviewing last night and fell asleep without hitting submit. 😆 This should be good to go now. I'm reviewing some issues and PRs and going to get some things merged.

PR #25167, opened 2018-01-02
Fixed color for disabled buttons

wolfgangGoedel:
missing reset for disabled button color could cause hover effect on disabled buttons when hover color changed i.e. from light to dark


andresgalante (approved):
Hi @wolfgangGoedel thanks for the contribution.

To test this PR, increase the darken function for hover background on the button mixin to for example 50% and try to hover over a disabled light btn.

mdo (approved):

PR #25169, opened 2018-01-02
Using the expected variable to set the custom select height

straube:
In the current `v4-dev` branch, the variable `$custom-select-height` is not being used in the `_custom-forms.scss` file. Instead, its default value is set there: `$input-height`. So, when one tries to change that variable's value it doesn't lead to the expected result. To fix that I'm simply using the variable from `_variables.scss` to set the `.custom-select` height.


straube:
After the last time I updated this pull request branch the build failed. [The output](https://travis-ci.org/twbs/bootstrap/jobs/327189472) shows an error that doesn't seem to be related to the code. I'm wondering if re-building it would make the checks to pass. Is that even possible?

mdo:
@straube It was unrelated I imagine—Travis had some issues after patching those major vulnerabilities.

iamandrewluca:
Encountered right now. Thx for fix. :smiley:

iamandrewluca:
As `$custom-select-height` is `$input-height` int `_variables.scss`
https://github.com/twbs/bootstrap/pull/25331

PR #25180, opened 2018-01-02
descriptive docs brand variable naming

m5o:
* use descriptive variable naming for brand color
* rename `purple` to `brand`
* align docs variables


mdo:
Thanks, but I don't think this needs a change given it's our docs and very few folks work on it. Our docs Sass is meant to be rather simplistic, and ideally I remove even more from it in the future.

m5o:
Ok, just stumble on this minor naming. Who knows how long the purple naming last. 😉

PR #25175, opened 2018-01-02
Update _custom-forms.scss

719media:
Example pull request for https://github.com/twbs/bootstrap/issues/25174

Issue #25174, opened 2018-01-02
Simpler custom control markup


719media:
I was running into issues with the custom-control (input and checkbox) padding and generally just not being as friendly as I would have liked.

1) Instead of padding-left:1.5rem, why not padding-left:1rem, and then create margin-left:0.5, which allows for the checkbox itself to be rendered cleanly. Useful for me when putting a checkbox into a container (say, a `<td>` of a `<table>`) and want it to look clean and centered. Obviously I could just create my own class here, but I feel that this type of styling is more clean.

2) Where possible, I like to stay away from position: absolute when creating something meant to be extending/used globally (e.g., the bootstrap library :) ). One benefit of avoiding absolute is that sizing the custom input becomes a little easier, because all you need to worry about at that point is the top offset (instead of having to worry about the padding-left spacing to make room for the absolute)

Anyway, I've put my ideas here:

https://jsfiddle.net/fqxcdp8h/

I hope you would at least consider implementing the fix for `#1`.

719media:
For example:

https://github.com/twbs/bootstrap/pull/25175

Please note that the pull request changes the semantic meaning of `$custom-control-gutter` to mean the actual "gutter" between the custom-control checkbox/radio instead of the total space reserved for it. I think this actually happens to also be more accurate of what a "gutter" is.

The default in _variables.css would be updated to 0.5rem instead of 1.5rem if this request was applied I imagine.

mdo:
Hmm, this might be doable. I played with flexbox styles on the checkboxes before I rewrote them for Beta 3, but I couldn't make it work well. With the reduced markup, this gets easier. Lemme try this out on both the default and custom checks.

mdo:
Only downside I see to this is, within the same form there being inconsistent spacing should your radio and checkbox be different dimensions. I don't know off the top of my head if they are yet, but I'll take a look shortly.

mdo:
Fresh pass for both input types at #25184.

tmorehouse:
I was playing around with making sizing available to custom radio/checkboxes, and figured that rather than using `rem` units for the `::before` and `::after` pseudo elements, and for the left padding and margin, using `em` units allows the indicator to scale with the size of the label text (which can be left in `rem` units).

mdo:
I started playing with some new stuff again for v5, but I don't think we'll be significantly changing the markup here. [See the Codepen I put together.](https://codepen.io/anon/pen/xmOEvx)

[...]

PR #25184, opened 2018-01-03
Updates the native and custom checks to use flexbox

mdo:
- Removes absolute positioning and relies on flexbox for input+label layout and spacing
- Adjusts the gutter variables to be true gutters, not gutters plus space for input
- Updates inline variants accordingly

This is an implementation of #25174, and a more complete attempt than #25175 at implementing it across all our checks.


ysds:
Please add `flex-shrink: 0` to `.custom-control-label::before`  for long label.

image

scss/_custom-forms.scss
```diff
@@ -9,9 +9,7 @@
 
 .custom-control {
   position: relative;
-  display: block;
-  min-height: (1rem * $line-height-base);
-  padding-left: $custom-control-gutter;
+  display: flex;
```
719media: I had the same display: flex in my pull request, but now I'm trying to remember the reason for even having it. Does removing this cause any negative effect? Or does it have a purpose for inclusion? I fiddled around a bit and didn't see a reason for just leaving the display property off... what do you think?

scss/_forms.scss
```diff
@@ -228,15 +226,12 @@ select.form-control-lg {
...
   display: inline-flex;
   align-items: center;
-  padding-left: 0; // Override base .form-check
   margin-right: $form-check-inline-margin-x;
 
   // Undo .form-check-input defaults and add some `margin-right`.
   .form-check-input {
-    position: static;
     margin-top: 0;
     margin-right: $form-check-inline-input-margin-x;
```
719media: should $form-check-inline-input-margin-x be consolidated with $form-check-input-gutter?
If not, should the default value in _variables be changed so that they are equal?

scss/_custom-forms.scss
```diff
@@ -87,6 +84,10 @@
     background-position: center center;
     background-size: $custom-control-indicator-bg-size;
   }
+
+  &:not(:empty)::before {
+    margin-right: $custom-control-gutter;
```
pat270: `.custom-control-label` will never be empty because of the before and after pseudo-elements. It will always have `margin-right`.

mdo:
Note to myself, check #25337.

tmorehouse:
@mdo, To allow the indicator to scale with the text size of the label, I would suggest that the indicator width/height (and possibly padding/gutter) be switched to `em` units instead of `rem` units, as it would then inherit the size from the `label` text size, and scale quite nicely. 

This would then allow easy support for `custom-control-sm` and `custom-control-lg` sizing classes (with no special handling for the indicators needed).

mdo:
Punting on this per https://github.com/twbs/bootstrap/issues/25174#issuecomment-447691388.

Using flex on these checks isn't really doable without more complications—we'd need `row` direction for the input and label, but for any nested content that's a sibling of the input and label (like help text), it'd require `column` direction.

PR #25185, opened 2018-01-03
Add pure css button (group) toggle

### LINKED ISSUE #25122: Discrepancy between .active and :checked in radio button groups when going back to a page via browser history
I have noticed that when I press the "previous" button in my browser, the page is reloaded from cache and the item shown as `.active` is not the one I clicked last, but the one initially defined as `.active` in the DOM.
This might not match with the actual radio buttons state as the `:checked` one is the last one clicked.
Checkboxes have the same behavior in a similar setup.

I figured a way to overcome this by not using the `.active` class in the DOM and setting it according to the `:checked` states of radio buttons at page load.

See this fiddle for an example of the bug, and the temporary fix: https://jsfiddle.net/pr_shadoko/saymu7wa/7/

Tested on :
Windows 10 64bit
Chrome 63.0.3239.84 / Firefox 57.0 / Opera 49.0
jQuery 3.2.1
BS 4.0.0-beta.3
popper 1.12.9

### LINKED ISSUE #27600: Button toggling checkbox is always left un-checked on spacebar "click"
- Operating system and version: Ubuntu 18.10
- Browser and version: Chrome 70.0.3538.77 and Firefox 63.0
- Example: https://jsbin.com/gaxuwobiku/edit?html,output

Problem:
When I'm filling out forms using a physical keyboard, I typically tab from input to input, and use the spacebar for checking/unchecking checkboxes.  Unfortunately, using the [example code for a label button with "embedded" checkbox](https://getbootstrap.com/docs/4.0/components/buttons/#checkbox-and-radio-buttons) does not work when the spacebar initiates the "click".  A mouse click works as expected.  It appears to do something with the timing of the event, as the "checked" property is properly set immediately after the event, but is always removed (read as false) sometime after.

ysds:
This PR add pure CSS toggle button.

The button plugin will be no longer need, but this allows available both JS and pure css.

Button Plugin:
```html
<button type="button" class="btn btn-secondary" data-toggle="button" aria-pressed="false" autocomplete="off">
  Single toggle
</button>
```

Pure CSS (Required new `.btn-group-input` class):
```html
<div class="btn-group">
  <input type="checkbox" class="btn-group-input" id="checkbox1">
  <label class="btn btn-primary" for="checkbox1">Single toggle</label>
</div>
```

Demo: https://codepen.io/fellows3/pen/bQydEm

Fixes #25122
Fixes #27600


mdo:
> Required new `.btn-group-input` class

That's likely a deal breaker on this being in v4 as an update to the existing button group. We'd have to deprecate the existing one and provide this as a new component. Breaking changes are a no-go heading into stable v4.0.

ysds:
I added `.btn-group-input` as a class similar to` .custom-control-input`, but It can be realized without `.btn-group-input` by complicating CSS.

```css
.btn-group-input { ... }
```
to
```css
.btn-group input[type="radio"],
.btn-group input[type="checkbox"] { ... }
```
```html
<div class="btn-group">
  <input type="checkbox" id="checkbox1">
  <label class="btn btn-primary" for="checkbox1">Single toggle</label>
</div>
```

patrickhlauke:
i'd like to see this addressed in 4.1 perhaps though. toggle groups are extremely brittle due to their reliance on JS which makes them easily go out of sync with things like keyboard behavior, autofill behavior, etc

ysds:
x-ref: #25281

MartijnCuppens:
I like this, but maybe we better move this to v5? Otherwise we'll have 2 implementations of the same component. @mdo, what do you think?

mdo:
Superseded by #28463.

Issue #25182, opened 2018-01-03
hover-media-query


719media:
I was looking into hover-media-query, which landed in bootstrap 4 at some point in the past. There is even a little snippet of documentation about it here:

https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/#sticky-hoverfocus-on-mobile

However, it appears that this `$enable-hover-media-query` is commented out in the src scss file and has no effect.

I was hopeful that the hover classes could be controlled with something like `.touch` on the `<body>` element, but it doesn't appear that this class works that way, as the mixin is inside of the element and can't be escaped...

Aside from this, as you probably know, the hover media query has two problems:
1) It isn't supported in Safari (solved by using your shim https://github.com/twbs/mq4-hover-shim)
2) It doesn't really "work" for multiple-input devices, where a user could touch and mouse

After reading https://codepen.io/MillerTime/post/prepend-body-classes-to-nested-selectors-sass-mixin, I realized the same thing as the author: That you can use the `&` anywhere in the selector, and easily create body .no-touch classes using the existing hover mixin within bootstrap.

I propose that this could land in bootstrap easily by means of a simple selector like:

https://github.com/twbs/bootstrap/pull/25181

Now all I need to do is set the variable in my own personal custom variables file, e.g.:

`$enable-hover-selector: ':not(body.touch)'`

Works great!

patrickhlauke:
so what's the actual scenario this tries to solve? skimming over the prepend classes article, it seems that all the author achieved is to achieve a DRYer way of having styles for "touch" and "non-touch". That still leaves the fundamental problem of multi-input devices, just in a terser way?

719media:
I think I misunderstood that the shim runs during postcss phase. My approach is merely to use a containing class, which after further consideration is probably not a great approach.

So, perhaps in the code the commented out `$enable-hover-media-query` can just be uncommented?

patrickhlauke:
my question really is: what are you actually trying to achieve? what's the end goal?

719media:
The end goal is to disable hover classes during touch, but not during actual mouse hover.

This was a worse problem than I initially supposed. Getting an element touched on a mobile device to "unhover" is problematic, so merely setting a body class during touch and then unsetting during touchend doesn't really solve the problem, since the element is still "hovered" on the mobile device even after the touchend. So, fixing the issue for **multi-input** devices where the hover works during mouse, but not during touch, seems like a goal that I cannot seem to achieve with any reasonable code.

[...]

PR #25194, opened 2018-01-03
Update _hover.scss with ability to have custom wrapper class

719media:
Re-enable the media hover query selector with a slight tweak to allow `$enable-hover-media-query` to be set like:

`$enable-hover-media-query: 'body.my-custom-class'`

for some reason, sass doesn't work with the variable set like:

`$enable-hover-media-query: '@media (hover: hover)'`

I'm not really sure why. Any ideas? It would be nice to have this working with just one variable, instead of having to create two (one for selector, one for media query)


patrickhlauke:
Noted this in the discussion https://github.com/twbs/bootstrap/issues/25182 but for reference: i'd like to spend some time exploring some of the more fundamental issues here and if there's a way to avoid having to do any naive "hover-capable primary input" based generalisation, as it'll still fail in multi-input scenarios.

719media:
Closing this pull request as it no longer applies

PR #25196, opened 2018-01-04
Distribute Fancy CSS

bangzek:
CSS with optional feature

- $enable-shadows
- $enable-gradients


mdo:
Thanks, but we likely won't be doing this.

PR #25199, opened 2018-01-04
Check for window before window.QUnit

lukaszzdanikowski:
This fixes error ```window is undefined``` on env where the window is not available.

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -39,7 +39,7 @@ const Util = (($) => {
   }
 
   function transitionEndTest() {
-    if (window.QUnit) {
+    if (window && window.QUnit) {
       return false
     }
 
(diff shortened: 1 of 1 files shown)


js/src/util.js
```diff
@@ -39,7 +39,7 @@ const Util = (($) => {
   }
 
   function transitionEndTest() {
-    if (window.QUnit) {
+    if (window && window.QUnit) {
```
Johann-S: it would be better like that : 
```js
if (typeof window !== 'undefined' && window.QUnit) {

Johann-S (approved):

Johann-S:
Thank you @lukaszzdanikowski 👍

PR #25223, opened 2018-01-07
feat: improvements of `.navbar-expand .dropdown-menu`

zalog:
I think we need some improvements on the `.navbar-expand .dropdown-menu` on mobile.
Imho it must feels like a submenu, not a dropdown like on desktop vieports.

### This is my proposal:

after

Later edit: and for desktop, with a contextual background:

dropdown-bg

### This was before:

before

You can also take a look at my [pen](https://codepen.io/zalog/full/dJZxWe).

Later edit: [pen v2](https://codepen.io/zalog/full/jYxYRJ)

What do you think?

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -149,6 +149,62 @@
           padding-right: 0;
           padding-left: 0;
         }
+
+        .dropdown-menu {
+          padding-top: 0;
+          padding-bottom: 0;
+          margin-top: 0;
+          background-color: transparent;
+          border: 0;
+          border-radius: 0;
+        }
+
+        .dropdown-header,
+        .dropdown-item {
+          padding-right: $navbar-nav-link-padding-x;
+          padding-left: $navbar-nav-link-padding-x;
+          white-space: normal;
+        }
+
+        .dropdown-item {
+          background-color: transparent;
+        }
+
+        &.navbar-light {
+          .dropdown-item {
+            color: $navbar-light-color;
+
+            @include hover-focus {
+              color: $navbar-light-active-color;
+            }
+
+            &.active {
+              color: $navbar-light-active-color;
+            }
+          }
+
+          .dropdown-divider {
+            border-color: $navbar-light-toggler-border-color;
+          }
+        }
+
+        &.navbar-dark {
+          .dropdown-item {
+            color: $navbar-dark-color;
+
+            @include hover-focus {
+              color: $navbar-dark-active-color;
+            }
+
+            &.active {
+              color: $navbar-dark-active-color;
+            }
+          }
+
+          .dropdown-divider {
+            border-color: $navbar-dark-toggler-border-color;
+          }
+        }
       }
 
       @include media-breakpoint-up($next) {
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -32,6 +32,31 @@
     align-items: center;
     justify-content: space-between;
   }
+
+  .dropdown-menu {
+    position: static;
+    float: none;
+    margin-top: 0;
+    border: 0;
+    border-radius: 0;
+  }
+
+  @each $color, $value in $theme-colors {
+    &.bg-#{$color} .dropdown-menu {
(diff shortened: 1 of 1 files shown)


andresgalante:
Hi @zalog Thanks for the contribution, a few month ago I've taken a try to fix this on this PR #24055

Your solution is a bit different, maybe you can take a look at that one and let me know what you think.

zalog:
Oh! Sorry, didn't saw that PR... Perhaps because diversity, it's a win/win for the repo :)
Can you please rebase your branch? I'd like to take a look there.

zalog:
I made some tests on the min file size, bootstrap.min.css:
- actual `twbs:v4-dev` - **141.0k**
- `zalog:zalog-navbar-dropdown` - **145.8k**
- `andresgalante:drop-menu-navbar` rebased onto v4-dev - **147.5**

`andresgalante:drop-menu-navbar` does not have styling for `.dropdown-header` and it needs some extra vars `$navbar-*-dropdown-*` - [code here](https://github.com/andresgalante/bootstrap/commit/44843412711c019debcc8bbbabd3e0a8cc5a6a2e#diff-d8ee409a461718bfb6240710c8c73382R542).

I think `zalog:zalog-navbar-dropdown` it's a better approach.

And, if we consider #25274, we can win 0.5k, and the final size would be 145.3 :).

But, I think I can make some other improvements.

zalog:
Done! I think this is it, or we are very close.
We now have only 1.8kb increase, **142.8kb**.
It's the same aproach as here #25274, mobile first.

Check out the last commit, and of course the [pen v2](https://codepen.io/zalog/pen/jYxYRJ) :)

andresgalante (approved):
Hi @zalog , you are right, this is a leaner solution and takes headers into consideration.

@mdo I am ok merging this one and closing #24055

scss/_navbar.scss
```diff
@@ -32,6 +32,31 @@
     align-items: center;
     justify-content: space-between;
   }
+
+  .dropdown-menu {
+    position: static;
+    float: none;
+    margin-top: 0;
+    border: 0;
+    border-radius: 0;
```
andresgalante: @zalog sorry, I missed 2 this:
First:
You are removing `border` and `border-radius` when the drop menu is within a navbar always, even in large view ports. Remember that border radius is an opt in and we should keep them if the user chooses to do it.

scss/_navbar.scss
```diff
@@ -32,6 +32,31 @@
...
+  }
+
+  @each $color, $value in $theme-colors {
+    &.bg-#{$color} .dropdown-menu {
```
andresgalante: Second: this is changing the design. the drop menus would follow the color of the menu. I think I like it, but it might be a visual breaking change.

andresgalante (changes requested):

scss/_navbar.scss
```diff
@@ -158,21 +178,29 @@
...
+        .dropdown-menu-right {
+          right: 0;
+          left: auto; // Reset the default from `.dropdown-menu`
+        }
+
+        .dropdown-header,
+        .dropdown-item {
+          padding-right: $dropdown-item-padding-x;
+          padding-left: $dropdown-item-padding-x;
+          white-space: nowrap;
```
andresgalante: what happens if the string is too long, would it overflow the screen?

scss/_navbar.scss
```diff
@@ -32,6 +32,31 @@
     align-items: center;
     justify-content: space-between;
   }
+
+  .dropdown-menu {
```
andresgalante: Why did you moved the `.dropdown-menu` selector under `.navbar` instead of `.navbar-nav`?

zalog:
Hi @andresgalante, I'll come back shortly on this branch.
Thanks for your time&feedback, you have a point there.

In the mean while, if you guys consider that we should not commit that contextual background on the dropdown for desktop (the breaking change), we can reset hard the last commit. But with this reset, we go back at 145.8kb :(

Issue #23358, opened 2017-08-12
Disabled state for links inside navbar is not consistent with buttons


ghost:
According to https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state

> The .disabled class uses pointer-events: none to try to disable the link functionality of &lt;a&gt;

It's not true for a `.nav-link` in navbar https://getbootstrap.com/docs/4.0/components/navbar/#toggler

![](https://i.imgur.com/dUKPei6.png)

mdo:
Yeah, the docs you're mentioning are for for buttons, not nav links. Do we want to disable `pointer-events` for all our links?

maxkorz:
@mdo I believe the styles must be consistent among all components. What I found out:

`pointer-events` is disabled for:
* [Buttons](https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state)
* [Pagination](https://getbootstrap.com/docs/4.0/components/pagination/#disabled-and-active-states)

`pointer-events` is not disabled for:
* [List group](https://getbootstrap.com/docs/4.0/components/list-group/#links-and-buttons)
* [Dropdown menu](https://getbootstrap.com/docs/4.0/components/dropdowns/#disabled-menu-items)
* [Navs](https://getbootstrap.com/docs/4.0/components/navs/)

It's important to point out that `.disabled` `<a>`s are still clickable without `pointer-events: none`

mdo:
@patrickhlauke Any strong feelings here?

patrickhlauke:
i *think* that shouldn't have any adverse side effects

MartijnCuppens:
Fixed in the meanwhile: https://twbs-bootstrap.netlify.com/docs/4.3/components/navbar/

PR #25224, opened 2018-01-07
Disabled nav link with pointer-events

melih:
**bootstrap/scss/_navbar.scss** is providing only **color** property for disabled nav links. When there is an href attribute, HTML keeps the tag inactive "as visual" but keeps it active and clickable "as functional".

```
  .navbar-nav {
    .nav-link {
      color: $navbar-light-color;

      @include hover-focus {
        color: $navbar-light-hover-color;
      }

      &.disabled {
        color: $navbar-light-disabled-color;
      }
    }
```

To make these lines more useful and meaningful, pointer-events property should also be added so the link with disabled class would really be inactive .

```
      &.disabled {
        color: $navbar-light-disabled-color;
        pointer-events: none;
      }
```

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -226,6 +226,7 @@
 
       &.disabled {
         color: $navbar-light-disabled-color;
+        pointer-events: none;
       }
     }
 


mdo:
Should this instead be on all `.nav-link`s in `_nav.scss`?

melihsaricam:
@mdo Well, I think. Should I open a new issue for this?

mdo:
This starts to get at #23358, but I'd rather see a holistic approach—applying this to all components where possible and clearly documenting it. Closing this out for now, so feel free to open a new PR if you want to try to tackle that. See the referenced issue for more components that might need this.

PR #25226, opened 2018-01-08
Fix link to constraint validation API

bryik:
I was reading your lovely docs and noticed [this link](https://www.w3.org/TR/html5/forms.html#the-constraint-validation-api) has died. It seems W3C moved their documentation; the new address is [here](https://w3c.github.io/html/sec-forms.html#the-constraint-validation-api).


mdo:
Unsure about the `github.io` address as `w3.org` still works for me as the primary domain. What about https://dev.w3.org/html5/spec-preview/constraints.html?

bryik:
I found a few links to W3C constraint validation API docs and wasn't sure which to pick. WHATWG has a [version](https://html.spec.whatwg.org/#the-constraint-validation-api) as well--confusing!

@mdo if that is the link you want, I'll update the PR

wolfy1339:
The github.io seems to be more up to date then the other addresses

XhmikosR:
@mdo: I say you just edit the file on this PR and switch to the link @wolfy1339 suggested above.

PS. I will run the broken link checker branch which should show if we have other such errors.

mdo:
Closing for #25244 to do just that.

PR #25248, opened 2018-01-09
Prevent word-wrap

### LINKED ISSUE #25235: Progress bar text is wonky when <3% and in .card

MartijnCuppens:
Fixes #25235


mdo (approved):
Approving with the side note that this wasn't an issue in v3—or at least we didn't have this property-value pair there.

Issue #24670, opened 2017-11-04
Input slow and high CPU on large forms


eeaquino:
When having a large table with several inputs there is a noticeable lag in the input box. Also CPU usage stays at 12.5% table is about 1500 rows.

Commenting out in line 250 bootstrap.css stopped the issue : 
```
a,
area,
button,
[role="button"],
/*input:not([type="range"]),*/
label,
select,
summary,
textarea {
  -ms-touch-action: manipulation;
      touch-action: manipulation;
}
```

patrickhlauke:
In which OS/browser do you see this happening?

patrickhlauke:
related: as most browsers now do the 300ms optimization for `width=device-width` (see https://patrickhlauke.github.io/touch/tests/results/#suppressing-300ms-delay) we can probably consider dropping this directive.

Herst:
Is that a browser bug?

kamelkev:
Above issue is affecting us as well.

We have some legacy administrative forms that can contain up to 5000 divs, each containing an input and a button element. Loading this page brings Chrome to a crawl. A full pageload in firefox take approximately 6 seconds, in Chrome it takes approximately 50s.

Commenting out the above rule solved the problem, similar to what @eeaquino noted.

The pageload behavior for Chrome is pretty interesting. When the above rule is in place the 'Content Download' portion of the waterfall is approximately 45s. However when the rule is commented out that drops to 241ms.

I did not know that CSS rules could affect document transport? This behavior can only be seen on Chrome. Most notably I don't see it on Safari.

Please let me know if you have any additional questions relating to this. We are happy to help however we can.

Herst:
Please someone report this to the Chromium team: https://crbug.com/

patrickhlauke:
we'd need a reproducible test case. i just tested a document with 5000 `<div><input><button>foo</button></div>` with and without the `touch-action` directive, using a local server on my machine, and chrome loaded both the same with no noticeable delay nor difference.

eeaquino:
https://codepen.io/anon/pen/aEqReQ

patrickhlauke:
filed https://bugs.chromium.org/p/chromium/issues/detail?id=800522 /cc @mustaqahmed @RByers @NavidZ

[...]

PR #25250, opened 2018-01-09
Remove `touch-action:manipulation` click optimization

patrickhlauke:
Due to a severe performance bug in Chrome (see https://bugs.chromium.org/p/chromium/issues/detail?id=800522) and due to the fact that (with the exception of IE/Edge on Windows desktop with touchscreen) all modern browsers now optimize the click delay away anyway, this optimization is not needed.

Closes https://github.com/twbs/bootstrap/issues/24670


XhmikosR:
@mdo: you moved this into Ready to merge, but you didn't approve. Does it look good to you?

Personally, LGTM.

XhmikosR (approved):

PR #25253, opened 2018-01-10
Improve album example on mobile

taylorbryant:
Last night, I was looking through Bootstrap's documentation on my phone, and I noticed that the album example looked rather odd. My first thought was that maybe the examples are not meant to be optimized for mobile. After looking at the other examples, which look just as good on mobile, I concluded otherwise.

In these commits, I did the following:
* Removed float and width properties from the .card class
* Added .col-md-4 class to each card
* Refactored the footer to use Bootstrap's own grid system instead of relying on the .float-right utility class

In my opinion, the end result is a much better experience on mobile. The only thing that sticks out now is that the buttons at the top hug each other on mobile. Anyways, I figured this is a decent start.

### DIFF
--- a/docs/4.0/examples/album/album.css
+++ b/docs/4.0/examples/album/album.css
@@ -33,8 +33,6 @@ body {
 }
 
 .card {
-  float: left;
-  width: 33.333%;
   padding: .75rem;
   margin-bottom: 2rem;
   border: 0;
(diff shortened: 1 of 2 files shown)


mdo:
Haven't looked at the code changes here, but nearly all the examples are getting updated and I'm adding new ones in #24898.

PR #25255, opened 2018-01-10
Remove `position: relative` from grid columns

### LINKED ISSUE #25254: [v4] Why are all col classes 'position: relative'?
Is there a reason for making all columns relative?

Typically I wrap my rows and cols inside 'sections' which have their own padding top and bottom. I sometimes choose to make elements inside those columns stick absolutely to the edges of the section. With cols as relative now, it's less trivial to do so.

Applying an override of..

```
[class*='col'] {
    position: static;
}
```

does the trick, and I can't seem to find any drawbacks. Although I'm wondering if there's a good reason for relative in the first place? Please correct me if I'm wrong or missing something.

Cheers!

mdo:
Prompted by #25254, I dug into our Git history to see why we add `position: relative` to our grid classes and mixins. Turns out, I have no idea and added it to v3 in a commit that fixed an unrelated issue—see https://github.com/twbs/bootstrap/commit/0f17f9b7f7fe15bb4ce6301caa2600bfd30e63d7. This change would restore a default, and implicit, `position: static`, allowing you to opt into whatever positioning you need.

I don't see this as a breaking change—we change properties and values as needed—but I could see this being confusing and unnecessary either way.

Thoughts from folks?

Closes #25254.

---

##### Todos if moving forward
- [ ] Update docs snippets around mixins
- [ ] Verify docs don't mention relative position


ysds:
I prefer remove it because it is essentially unnecessary for grid system (not only bootstrap). I do not care compatibility from v3.

This change will affect a wrapper of custom validation tooltip.

> If your form layout allows it, you can swap the .{valid|invalid}-feedback classes for .{valid|invalid}-tooltip classes to display validation feedback in a styled tooltip. Be sure to have a parent with position: relative on it for tooltip positioning. In the example below, our column classes have this already, but your project may require an alternative setup.
https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/components/forms.md#tooltips

slavanga:
It was required for the dropped push and pull modifier classes.

mdo:
Ah good points @ysds and @slavanga. That definitely makes sense for the push/pull classes given those used `right` and `left`. Sounds like the validation is the only remaining thing. For that, we'll need to add `.position-relative` or add `position: relative` to our `.form-group` and add those. Plus, update the docs text.

mdo:
Going to have to punt this to v5 because it breaks form validation tooltip positioning. We currently rely on that to align the form feedback message.

PR #25264, opened 2018-01-10
Fix breakpoint mixins for custom breakpoint maps

sandydoo:
When using the `media-breakpoint-between` or `media-breakpoint-only`
mixins with a custom breakpoint map – not defined as `$grid-breakpoints`
– the mixin incorrectly defaults back to the global `$grid-breakpoints`
map in certain situations. This commit correctly passes on the
`$breakpoints` argument to the other mixins called within the block.


mdo:
Can you run this through an example in Sassmeister by chance? Hoping we can see it not working and working as intended with the changes if possible.

sandydoo:
@mdo Done! 👍 

Sassmeister: https://www.sassmeister.com/gist/cda75c97d781be99f370d1a3768d0f57
Gist: https://gist.github.com/sandydoo/cda75c97d781be99f370d1a3768d0f57

mdo (approved):

PR #25274, opened 2018-01-11
A better `.navbar` align and a small optimization

zalog:
On `.navbar`, I think we should use the same padding like on `.container`.
If we use a bigger `.$grid-gutter-width`, or smaller, the differences are visible.

#### Before
Check [this pen](https://codepen.io/zalog/full/XVqMLX), or:

before

#### After
Check [this pen](https://codepen.io/zalog/full/baMedp), or:

after

### Commit 2

A mobile first approach and with that, a small optimisation on the min css file. About 0.5kb :)

### Commit 3

The third commit, well I'm not sure, but I think it's about Star Trek :)

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -171,7 +171,7 @@
         }
 
         .navbar-collapse {
-          display: flex !important;  // stylelint-disable-line declaration-no-important
+          display: flex !important; // stylelint-disable-line declaration-no-important
 
           // Changes flex-bases to auto because of an IE10 bug
           flex-basis: auto;
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -31,6 +31,8 @@
     flex-wrap: wrap;
     align-items: center;
     justify-content: space-between;
+    padding-right: 0;
+    padding-left: 0;
   }
 }
 
@@ -143,14 +145,6 @@
     $infix: breakpoint-infix($next, $grid-breakpoints);
 
     &#{$infix} {
-      @include media-breakpoint-down($breakpoint) {
-        > .container,
-        > .container-fluid {
-          padding-right: 0;
-          padding-left: 0;
-        }
-      }
-
       @include media-breakpoint-up($next) {
         flex-flow: row nowrap;
         justify-content: flex-start;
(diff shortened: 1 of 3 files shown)


scss/_navbar.scss
```diff
@@ -176,7 +176,7 @@
         }
 
         .navbar-collapse {
-          display: flex !important;  // stylelint-disable-line declaration-no-important
+          display: flex !important; // stylelint-disable-line declaration-no-important
```
XhmikosR: This improves nothing; it depends on your editor. I don't mind having though.

zalog: We need that comment for stylelint, `npm run css-lint` sequence.
It comes from #23572, [here](https://github.com/twbs/bootstrap/pull/23572/files#diff-df1f74faf207a474f1d78c86393221c8R178) :)

XhmikosR:
I'm obviously aware of that since I added it. What I'm saying is your
whitespace change is useless.

zalog:
I thought you mean the entire comment.
Right, it's not a big deal. I just verified all the *.scss files and deleted the extra spaces, excepting [this cases](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_print.scss#L44).
So, as you guys wish :) I can delete the last commit anytime.

zalog:
Hmm... I'm not sure what you mean by live demo.
There are 2 pen links above each screenshot.

If by live, you mean using the css/sass file from the actual twbs/v4-dev or my temp branch, I can try to do that.
I'm not sure if I can use sass on codepen, because there are some limitation, but I'll figure it out.

mdo:
Shit, I missed those! Sorry :)

zalog:
I rebased branch if we needed.

mdo:
Punting this to v5 just so we can proceed without interruptions and make any potential breaking changes. Will revisit with @MartijnCuppens I think soon.

zalog:
Ok, thanks.

MartijnCuppens:
I think we should remove the `padding-left` and `padding-right` from the navbar completely and make the `.container` or `.container-fluid` required. Overriding paddings based on where a container is placed doesn't feel right. Also if we want to continue with https://github.com/twbs/bootstrap/pull/25631, we will need to generate a lot of css to get this right.

If for some reason someone wants to use the navbar in the page, the padding utilities can be used to add . 

What do you think @mdo?

PR #25282, opened 2018-01-12
Use the SRI hashes for the live site's assets.

XhmikosR:


Johann-S (approved):

mdo (approved):

XhmikosR:
@mdo: just so that we are on the same boat, this assumes you don't change the dist files on the gh-page without updating the HTML files for v4. If you do that, the CSS files won't load. Which IMO is a good thing.

mdo:
I've never needed to update our dist files on `gh-pages` outside a release—only our docs CSS. So as long as those assets don't change, I'm good.

PR #25312, opened 2018-01-15
Update _custom-forms.scss

Smilefounder:
Fix issue displaying select with default arrow for chrome.


XhmikosR:
We use autoprefixer for this task.

PR #25315, opened 2018-01-15
Use the bundle version instead of the normal version and Popper.js

ghost:
With Beta 2 a bundle version of the Bootstrap plugins js with Popper.js included was created.
Using this version instead of the Popper.js _and_ the Bootstrap plugins js increases the speed and simplicity.

This PR changes the index and examples to the bundled Bootstrap js plugins version.

### DIFF
--- a/docs/4.0/getting-started/introduction.md
+++ b/docs/4.0/getting-started/introduction.md
@@ -25,14 +25,13 @@ Copy-paste the stylesheet `<link>` into your `<head>` before all other styleshee
 
 ### JS
 
-Many of our components require the use of JavaScript to function. Specifically, they require [jQuery](https://jquery.com), [Popper.js](https://popper.js.org/), and our own JavaScript plugins. Place the following `<script>`s near the end of your pages, right before the closing `</body>` tag, to enable them. jQuery must come first, then Popper.js, and then our JavaScript plugins.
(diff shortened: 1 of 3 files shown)


docs/4.0/getting-started/introduction.md
```diff
@@ -25,14 +25,13 @@ Copy-paste the stylesheet `<link>` into your `<head>` before all other styleshee
 
 ### JS
 
-Many of our components require the use of JavaScript to function. Specifically, they require [jQuery](https://jquery.com), [Popper.js](https://popper.js.org/), and our own JavaScript plugins. Place the following `<script>`s near the end of your pages, right before the closing `</body>` tag, to enable them. jQuery must come first, then Popper.js, and then our JavaScript plugins.
+Many of our components require the use of JavaScript to function. Specifically, they require [jQuery](https://jquery.com), [Popper.js](https://popper.js.org/), and our own JavaScript plugins. Place the following `<script>`s near the end of your pages, right before the closing `</body>` tag, to enable them. jQuery must come first, then our JavaScript plugins which include Popper.js in the bundle version already.
```
Johann-S: Not sure about that change

index.html
```diff
@@ -59,11 +59,10 @@ <h5>CSS only</h5>
 <link rel="stylesheet" href="{{ site.cdn.css }}" integrity="{{ site.cdn.css_hash }}" crossorigin="anonymous">
 {% endhighlight %}
 
-<h5>JS, Popper.js, and jQuery</h5>
+<h5>JS (with Popper.js included), and jQuery</h5>
```
Johann-S: same here

Johann-S:
I'm not against using our bundled file in our documentation but it seems you included unrelated changes here

mdo:
I'd rather we leave it as-is. Not every component needs Popper; introducing it this way should encourage folks to remove it or avoid it entirely if they don't need tooltips, popovers, or dropdowns.

Issue #25316, opened 2018-01-15
Inconsistent use of :disabled and [disabled]


MartijnCuppens:
As well `:disabled` as `[disabled]` are used. 

https://github.com/twbs/bootstrap/blob/fa3de5cbd8f73d2cbffc0c722c996b8a13b44c5d/scss/mixins/_buttons.scss#L29-L37

It would be a bit more consistent to use `:disabled` everywhere.

patrickhlauke:
the `:not()` pseudo selector cannot contain another pseudo-selector https://developer.mozilla.org/en-US/docs/Web/CSS/:not#Syntax (unless SCC allows for some fancy rewriting of the output to account for that?)

MartijnCuppens:
As far as I know, `:disabled` is not a pseudo-element but a pseudo-class which is allowed in the `not()` selector.

patrickhlauke:
ah, right you are, my mistake.

patrickhlauke:
@MartijnCuppens https://github.com/twbs/bootstrap/pull/25317

patrickhlauke:
agree that settling on `:disabled` looks cleaner, and it may remove the idea that you can just arbitrarily throw a `disabled` attribute on any old random element to have the styles kick in.

PR #25317, opened 2018-01-15
Consistently use `:disabled` rather than `[disabled]` selector

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/25316


mdo (approved):
Looks so much simpler now! :D

PR #25318, opened 2018-01-15
Adjust examples index

### LINKED ISSUE #25311: Resize examples screenshots
The old ones were 1600x1200, resized in the HTML to 800x600 (**ratio 1.33**).

The new ones after #24898 are 1440x900 so with **a ratio of 1.6**, yet we still resize them in the HTML to 800x600 (**1.33**).

IMO, we should resize the PNGs to whatever dimensions we show the thumbnails. Sure, they won't look as crisp as they can on retina displays but it's so many bytes to ignore.

/CC @mdo: ping me if you resize them and make a PR so that I can compress them before we merge.

mdo:
Corrects dimension attributes and improves rendering of columns at smaller viewports.

Fixes #25311.


XhmikosR:
@mdo: I finished compressing the screenshots. This should have been made before you merged; the more we commit binary files, the more the git size increases.

Just FYI so that you bug me in the future before merging :)

PS. The screenshots don't have a consistent size; others are bigger, others smaller.

mdo:
Ah, indeed. I'll take a look at the dimensions here before we merge to avoid that once again :). Looking at them, we have a one-off on the new offcanvas one and there are some full retina ones in there at 2880x1800 (wayyyyy too big). I'll resize, then we can recompress again.

XhmikosR:
All right, just ping me when you are done and I'll compress the new ones.

mdo:
Just pushed new screenshots for everything, sized at 960x600 (1/3 their original as I took them).

XhmikosR:
@mdo: I rebased and cleaned up the branch. I also switched to 960x600 for the HTML dimensions.

Have a look and let me know.

Issue #25304, opened 2018-01-15
Dropright/dropleft is broken on navbar or with static position


ysds:
I wondered why `position: relative` is not given to dropright and dropleft, so examined any influence. 

Related code:

https://github.com/twbs/bootstrap/blob/ed286e0f3d7a376087bcb990f5721dff73153d50/scss/_dropdown.scss#L2-L5

Then, I found a position of dropright or dropleft is not as expected when using it on a navbar. [Demo](https://codepen.io/anon/pen/ppZrjW) Please see on a large screen.

Also, a dropright/dropleft outside navbar works well since it is positioned by popper, but if disable poper (#24092) it will break position.

Johann-S:
Yep we made the choice a few release before to not enable Popper.js in our navbar component but with `.dropleft` or `.dropright` I'm not sure if we'll have the choice to not use Popper.js

mdo:
Perhaps we limit left/right to everything but navbars? Or do we try to do some basic positioning for them inside navbars?

ysds:
Resolved in #25400

PR #25324, opened 2018-01-16
Add basic position to dropup/right/left

ysds:
Fixes: #25304

This changes fix a position of dropright/left menu on navbar.  And give a basic position for dropup/right/left for disable popper support in the future enhancement (#24092).

**Note**:
- The direction of carets on navbar is not responsive. (e.g. a caret of dropright is turning to the right even on mobile.)
- `display` is not defined for **dropright**. In many cases it should be use in/on an inline element (`.btn-group` or `.d-inline-block` etc). Otherwise the dropright menu will jump to the right end (`left: 100%`) of the `div`.

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -1,6 +1,8 @@
 // The dropdown wrapper (`<div>`)
 .dropup,
-.dropdown {
+.dropright,
+.dropdown,
+.dropleft {
   position: relative;
 }
 
@@ -35,6 +37,8 @@
 // Just add .dropup after the standard .dropdown class and you're set.
 .dropup {
   .dropdown-menu {
+    top: auto;
+    bottom: 100%;
     margin-top: 0;
     margin-bottom: $dropdown-spacer;
   }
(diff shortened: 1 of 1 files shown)


Johann-S:
is it possible to see a live example in a CodePen ?

ysds:
While creating a demo, I noticed a bug in this PR. The dropleft/dropup menu is broken by `right`, `bottom` property when not disabled popper (actually by default).

Demo: https://codepen.io/anon/pen/MrqGBV

ysds:
Closing since has a bug

Johann-S:
Dropup works pretty well currently (with Popper.js disabled) : https://getbootstrap.com/docs/4.0/examples/navbar-bottom/

Johann-S:
I think it's doable to fix that issue @ysds , `bottom: 100%` should be applied only for Dropdown and not for Dropup

ysds:
> Dropup works pretty well currently (with Popper.js disabled) : 

Yes. because it has the spesific scss:

https://github.com/twbs/bootstrap/blob/1f490ec165be0271b739053f8f5c7af3a75e63e9/scss/_navbar.scss#L193-L197

but dropup will not work the outside of navbar after the future enhancement (#24092).

PR #25326, opened 2018-01-16
Adjust examples

liuxiaolong2011:
adsf


mdo:
Do not use public repos to experiment, and do not open PRs for branches that belong to other people.

PR #25331, opened 2018-01-16
Fix input height border

unrevised6419:
When changing `$input-border-width`, `$input-height-border` does not change and `select.form-control` `custom-select` `custom-file` will still have `$input-height` which is resolved from `$input-btn-border-width`

This will work in cases when want controls with border, and buttons without.

Temporary fix will be to override 2 variables

```scss
$input-border-width: 2px !default;

// which is
// $input-height-border: $input-btn-border-width * 2 !default;
$input-height-border: $input-border-width * 2 !default;
```


mdo (approved):
Good catch and well explained succinct PR—thanks!

iamandrewluca:
Doing my best to contribute to open source 😃

PR #25319, opened 2018-01-15
use jekyll markdown converter

m5o:
* use build-in jekyll markdown converter
* remove custom markdown plugin
* constant usage of `markdownify`

Why? Custom Jekyll plugins are a hurdle if you want to use [github-pages](https://pages.github.com).
Bootstrap docs are featured here. It's possible to refactor the other plugins. I'll volunteer…

Hope you like it.

PR #25334, opened 2018-01-16
use bugify without custom jekyll plugin

m5o:
* use bugify without custom jekyll plugin

:link: https://github.com/twbs/bootstrap/pull/25319


XhmikosR:
@m5o: please add a comment in the new file with usage instructions.

Also, it would be nice to see if all the changes from plugins to this affect build time.

XhmikosR (approved):

PR #25338, opened 2018-01-17
Use .list-group-item-action for list group item variant

ysds:
`a.list-group-item` has been replaced to `.list-group-item-action` in v4.

Related:
- http://getbootstrap.com/docs/4.0/components/list-group/#links-and-buttons
- http://getbootstrap.com/docs/4.0/migration/#list-groups

This PR applies this approach to the mixin.


scss/mixins/_list-group.scss
```diff
@@ -6,10 +6,7 @@
     background-color: $background;
   }
 
-  a.list-group-item-#{$state},
-  button.list-group-item-#{$state} {
-    color: $color;
-
+  .list-group-item-action.list-group-item-#{$state} {
```
mdo: Maybe nest this in the above selector?

```scss
.list-group-item-#{$state} {
  // styles

  &.list-group-item-action {
    // styles
  }
}
```

ysds: Agree!

mdo (approved):

PR #25351, opened 2018-01-17
Explictly add `cursor:pointer` to dropdown menu items

### LINKED ISSUE #25350: <button> menu items inside dropdown have cursor: default
With beta.3 Bootstrap brought back `cursor: pointer` for buttons. However when having menu items using `<button>` tags inside dropdown they still use `cursor: default` (see this [example from docs](https://getbootstrap.com/docs/4.0/components/dropdowns/#menu-items)). I believe it is a bug and menu items should also have `cursor: pointer`.

patrickhlauke:
Follow-up to #25082

Closes #25350

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -89,6 +89,7 @@
   color: $dropdown-link-color;
   text-align: inherit; // For `<button>`s
   white-space: nowrap; // prevent links from randomly breaking onto new lines
+  cursor: pointer; // opinionated: add "hand" cursor to non-disabled items
   background-color: transparent; // For `<button>`s
   border: 0; // For `<button>`s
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
@patrickhlauke: Isn't there any way to inherit this?

mdo:
Our v4 release is tomorrow morning, punting this to our next release.

patrickhlauke:
> @patrickhlauke: Isn't there any way to inherit this?

inherit it from where? unless we add opinionated `button { cursor: pointer; }` or similar, i think explicitly defining it makes more sense. would also cover weirder scenarios where authors use some arbitrary element like `<span role="button" tabindex="0">` or whatever in their menus

XhmikosR:
I thought we set this at a higher level, yeah. If not, your PR seems good to me. But the final decision is on @mdo :)

MaPePeR:
The same problem exists with Buttons in List Groups ([example from docs](https://getbootstrap.com/docs/4.0/components/list-group/#links-and-buttons)).

mdo:
I think we'll go for #25598.

Issue #24838, opened 2017-11-20
Pagination active page-link lacks noticeable focus styling in Firefox


tmorehouse:
- Operating system and version : Any
- Browser and version: Firefox (and maybe others)
- Reduced test case: https://codepen.io/tmorehouse/pen/MOQXyd

The current CSS for pagination `.page-item.active .page-link` does not include any noticeable focus styling when using Firefox, which makes it difficult for keyboard only users to know if the `active` link is focused or not.

Tab through the page links in the codepen to see the issue (and one possible solution) using Firefox

I've used the following workaround CSS to provide focus styling for the active page-link:

```css
.pagination .page-item.active .page-link:focus {
    box-shadow: 0 0 0 3px rgba(0,123,255,.5);
    z-index: 2;
}
```
Currently the active page link looks like this with focused and not focused (the change in outline is negligible):
![image](https://user-images.githubusercontent.com/2781561/33040193-b44eff94-ce10-11e7-9365-3b78298130cb.png)

With the proposed focus styling it looks as follows (which uses similar styling as the primary button focus styling):
![image](https://user-images.githubusercontent.com/2781561/33040268-ea269f46-ce10-11e7-878b-bb1399363784.png)

Another quick workaround is to place the class `.btn-primary` (without the `.btn` class) on the active page-link (the `<a>` tag), which produces the same effect when focused.

andresgalante:
Hi @tmorehouse, thanks for reporting this.

The focus on anchors is the default focus of the browser, for example: on a mac on chrome it looks like this:

![img](http://g.recordit.co/GXGeiOAlJA.gif)

I see what you mean about the focus ring not been consistent with `.btn` and I am for adding the shadow on pagination. @patrickhlauke What do you think?

tmorehouse:
I should have mentioned that I was also using Firefox, which has a very subtle focus ring

andresgalante:
@tmorehouse I agree, different browsers will show less noticeable focus rings. Let's wait for @patrickhlauke comment. As I mention before I think we should add the same focus ring as in btns for consistency.

PR #25363, opened 2018-01-19
Pagination move `.page-link:focus` outline value to variable `$pagina…

pat270:
…tion-focus-outline` so we can unset it via Sass and don't have to re-implement browser defaults

Previous to https://github.com/twbs/bootstrap/issues/24838, using `.page-link` with an anchor tag didn't apply a focus outline when clicked with mouse. This lets those that don't want the focus outline on click unset it via Sass.

Demo: http://getbootstrap.com/docs/4.0/components/pagination/

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -23,7 +23,7 @@
 
   &:focus {
     z-index: 2;
-    outline: 0;
+    outline: $pagination-focus-outline;
     box-shadow: $pagination-focus-box-shadow;
   }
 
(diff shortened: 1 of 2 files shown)


mdo:
We don't do this anywhere else—why here?

pat270:
The main reason for this is the focus box shadow on mouse click. Browsers didn't apply focus styles on click when using an anchor tag in pagination before https://github.com/twbs/bootstrap/commit/2dc5762f96bd5c84f775a986861a852dd619ea79. Moving the outline value to a variable lets us prevent it from being output by Sass with `$pagination-focus-outline: !default;`, otherwise we would need to reset the outline value to browser defaults (kind of a pain). The biggest issue is being unable to reset it back in IE (If you know how I would love to know).

mdo (approved):

Issue #25273, opened 2018-01-11
Collapse multi-target fails with "#id1, #id2" data-target selector


bpierson:
[Fiddle](https://jsfiddle.net/bpierson/k0qtL2to/4/)

With a comma-separated CSS selector, the targeted collapsible elements are not toggled, for example:
data-target="#coll1, #coll2"

No error is raised.

MartijnCuppens:
An `aria-expanded` attribute is added to the toggle button to indicate if the collapsible element is collapsed or not. If you target multiple collapsible elements which have different states (collapsed and shown), it's impossible to determine the `aria-expanded` state of the toggle button.

I don't know if there are (accessible) solutions for this situation.

patrickhlauke:
this is a valid bug related to http://getbootstrap.com/docs/4.0/components/collapse/#multiple-targets

@MartijnCuppens as for the `aria-expanded`, you're right that if the multiple targets aren't all either expanded or collapsed, the `aria-expanded` will become meaningless. interesting edge case, as i'm not sure it's intended. might be worth filing a separate issue for it.

bpierson:
@MartijnCuppens , @patrickhlauke that's true about the `aria-expanded`, but wouldn't it be true in any case using the example from the docs that takes a class selector (and works correctly)? Thanks.

patrickhlauke:
@bpierson yes, which is why i'm saying that particular thing should be a separate issue. the problem you reported is valid as per your description (the valid jquery selector you provide, with the two ids, isn't working as it should)

bpierson:
Won't SR's honor the `display` style of the collapsible element in question, irrespective of which element triggered it? Maybe `aria-expanded` on the trigger is overkill here. Don't want to get too far into the weeds and away from the OP, but using `aria-expanded` seems to be a non-starter if the "any selector" feature stays in place since the value can only be true or false.

XhmikosR:
Ping @Johann-S

Johann-S:
Possibly related to this line https://github.com/twbs/bootstrap/blob/v4-dev/js/src/util.js#L109

bpierson:
@Johann-S I just confirmed that to be the case by just changing the value to `" #coll1, #coll2"`, which worked as expected (note the space at index 0).

[...]

PR #25390, opened 2018-01-21
Remove escaping selector and add a warning to inform folks to escape …

Johann-S:
…their selectors

Fixes : #25273


XhmikosR (approved):

Johann-S:
Thank you @XhmikosR 😄

PR #25392, opened 2018-01-21
Sort ascending `$grid-breakpoints` and `$container-max-widths` maps

unrevised6419:
~~Extend `$grid-breakpoints` and `$container-max-widths` maps, as `$theme-colors`~~ see #26714 

When extending `$grid-breakpoints` and `$container-max-widths` we must
be sure that they are sorted ascending, added a new function that sorts maps by their values.
Now `_assert-ascending` call it seems to be obsolete if we sort maps before that

If user extends mentioned vars after `variables` import
`map-sort-by-values` must be called again to be sure they are sorted

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -48,6 +48,57 @@
   @return $string;
 }
 
+// map-sort-by-values
+// Sorts a map ascending by values
+// Used for grid-breakpoints and container-max-widths sorting
+//
+// @param {Map} $map - Initial map
+// @return {Map} - Sorted map
+@function map-sort-by-values($map) {
+  // Transform map to zipped list
+  $keys: ();
+  $values: ();
+
+  @each $key, $val in $map {
+    $keys: append($keys, $key);
+    $values: append($values, $val);
+  }
+
+  $list: zip($keys, $values);
+
+  // Sort zipped list and create sorted map
+  $sortedMap: ();
+  @while length($list) > 0 {
+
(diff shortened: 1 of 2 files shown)


mdo:
Nice, thanks for this. Give me some time to review and test those new functions :).

scss/_functions.scss
```diff
@@ -48,6 +48,58 @@
   @return $string;
 }
 
+// map-sort-by-values
+// Sorts a map ascending by values
+// Used for grid-breakpoints and container-max-widths sorting
+//
+// @author Andrew Luca
```
MartijnCuppens: No need for the `@` comments here. You'll get the credits in the git history if this gets merged 😉

MartijnCuppens (dismissed):

MartijnCuppens:
@mdo, we didn't think about this when we merged https://github.com/twbs/bootstrap/pull/26714.

scss/_functions.scss
MartijnCuppens: Thanks a lot!

iamandrewluca:
@MartijnCuppens any feedback here? As I see merging maps has been reverted.

MartijnCuppens:
Hi @iamandrewluca 
#26714 was indeed reverted because it was a breaking change. We restored the original behaviour because this is the easiest behavior to understand for everyone, so we'll pass on this for now.

That being said, I'm quite impressed by the function you made here!

PR #25395, opened 2018-01-22
Button group refactoring and fixes

ysds:
The negative margin codes can be refactoring. And children selectors should be added similar as #25279.


mdo (approved):

XhmikosR:
@mdo: are you sure about this? Personally I try to not use `:not`, see also https://bitsofco.de/on-not-and-specificity/

ysds:
umm I read the article, but I didn't understand why you dislike in this case. Rather, I feel the article describes that ‘:not’ is useful for this case.

XhmikosR:
Read the part about specificity.

ysds:
I understand that the specificity of ":not" is complicated. Also the author said in the last section:

> I will definitely keep using :not because, in many circumstances, it is still the cleaner way to write styles. 

Don't afraid to use ":not". It allows CSS can stay clean and readable 👍

XhmikosR:
That is not my point here. My question was @mdo anyway.

scss/_button-group.scss
```diff
@@ -119,17 +113,14 @@
...
+  > .btn-group {
     width: 100%;
   }
 
-  > .btn + .btn,
-  > .btn + .btn-group,
-  > .btn-group + .btn,
-  > .btn-group + .btn-group {
+  > .btn:not(:first-child),
+  > .btn-group:not(:first-child), {
```
MartijnCuppens: This doesn't look ok

ysds: My all test case is here. There is no visual difference 👍 
* Before: https://codepen.io/anon/pen/RqbRRW
* After: https://codepen.io/anon/pen/QJLNPv

mdo: Is there a linter for stuff like that in the future?

XhmikosR:
I'd expect it to fail already :/
That being said I will try to see if there's something we can use.

mdo:
Missed the conversation here around `:not`, apologies for that.

Overall I have no issue using `:not()`s in our CSS overall. I don't like them being clever selectors at times, but  that's all in _how_ you use it. The only potential concern I see here is around existing problems we face—tooltips on buttons in a button/input group. However, that affects sibling/adjacent selectors as much as it does these selectors.

PR #25414, opened 2018-01-23
Override .custom-select width in .form-inline

### LINKED ISSUE #25413: Custom Selects Now Fill Entire Form Width
In v4.0, custom selects now have "width: 100%", causing them to fill up an entire line of a form.  I assume that this is a bug, as it looks pretty weird.  A perfect test case is the Bootstrap documentation.  In the "Components > Forms" section, under "Inline forms", there is an example under "Custom form controls and selects are also supported" that includes a custom select.  Back in v4 alpha 6, this select was only as wide as its widest text; now in the v4 full release, the select takes up the entire width of the form, pushing all following controls onto the next line, as seen below:

- v4 Alpha 6: https://v4-alpha.getbootstrap.com/components/forms/#inline-forms
image

- v4.0: https://getbootstrap.com/docs/4.0/components/forms/#inline-forms
image

I observed this behavior in Windows 10, on Chrome 63 (64-bit), Edge 41, and IE 11.

mdo:
Fixes #25413


BibbyChung (approved):

XhmikosR (approved):

sesemaya:
A quick question regarding this change: shouldn't this fix be applied to `.custom-file`, too?

Issue #25425, opened 2018-01-24
Unable to use .is-valid or .is-invalid on nested .form-control(s)


leaski:
Using a simple layout as the below:
```
<div class="form-group">
  <input type="text" class="form-control">
  <span>.</span>
  <input type="text" class="form-control">
</div>
```
the .is-valid or .is-invalid class must currently be applied to every form-control element

I would suggest that the validation selector should include `.is-valid` or `.is-invalid` on a `form-group` and apply the styling to nested `.form-control`(s)

leaski:
#25426 - PR

mdo:
We're not styling parent elements in v4's form validation—this was discussed during our alphas and betas. We're basing everything around the `:valid` and `:invalid` states for inputs.

PR #25426, opened 2018-01-24
Add form validation styling to form-controls when .is-${state} is app…

leaski:
…lied to the form-group element they are nested inside

Fixes - #25425


mdo:
Thanks, but closing per https://github.com/twbs/bootstrap/issues/25425#event-1439741149.

PR #25435, opened 2018-01-25
Removed menu item from navbar example

fraserhart:
The number/size of menu items in the navbar example was causing the navbar-right element to wrap on to two lines when the browser window width was 768px - 991px


mdo:
Bootstrap 3 is essentially in maintenance mode. As such, we're only accepting changes to v3's code on a case-by-case basis, usually only for critical bug fixes or docs improvements.

---

_This is a saved reply._

PR #25437, opened 2018-01-25
text-capitalize would capitalize The First T in Text

coliff:
text should be Text

Test: https://www.bootply.com/DD2KdMDkdc#


SharkyKZ:
Source text should not be capitalized. Otherwise .text-capitalize makes no difference.

coliff:
Hi @SharkyKZ - I just re-reviewed my PR - and you're right. I made a mistake reading the example... thanks for spotting that. I'll close this now.

PR #25444, opened 2018-01-26
add Micon

xtoolkit:
hi
i add Micon link

### DIFF
--- a/docs/4.0/extend/icons.md
+++ b/docs/4.0/extend/icons.md
@@ -25,3 +25,4 @@ While we haven't tried these out, they do look promising and provide multiple fo
 - [Dripicons](http://demo.amitjakhu.com/dripicons/)
 - [Ikons](http://ikons.piotrkwiatkowski.co.uk/)
 - [Glyph](http://glyph.smarticons.co/)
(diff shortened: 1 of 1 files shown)


coliff:
Sorry, but I don't think this project is legal and don't think this PR should be merged:
https://github.com/xtoolkit/Micon/issues/9

Furthermore, I think the documentation on your site would need to be up-to-date and HTTPS CDN links available.

rafalp:
Micon seems to be Microsoft's Segoe MDL2 icon font repacked for web use, which is direct violation of Microsoft's license, as they explain on their [fonts redistribution FAQ](https://www.microsoft.com/typography/RedistributionFAQ.mspx):

> **Q.** What can I do with the fonts supplied with Microsoft products?
> **A.** The fonts are governed by the same restrictions as the products they are supplied with. **You are not allowed to copy, redistribute or reverse engineer the font files.** For full details see the license agreement supplied with the product.

And on [font's page](https://docs.microsoft.com/en-us/windows/uwp/design/style/segoe-ui-symbol-font#how-do-i-get-this-font):

> **Q.** How do I get this font?
> **A.** To obtain Segoe MDL2 Assets, you must install Windows 10.

PR #25500, opened 2018-01-29
Move styles out of make-col-ready, in favour of make-row 

lunelson:
Since `.col-` elements are required to be direct children of `.row` elements, the styles in the `make-col-ready` mixin can be more simply and efficiently applied (bytes saved), and IMO be more in keeping with the dynamics of flexbox (i.e. `container > item` structure), if they are applied as part of the `make-row` mixin using a direct-child universal selector.

As a further optimisation, I suggest removing `position: relative` and adding explicit `flex` values, since flex-item defaults are [not cross-browser consistent](https://github.com/philipwalton/flexbugs#flexbug-6). Also, an alternative way of writing the direct-child selector, rather than using the universal selector, would be `.col, [class^="col-"], [class*=" col-"]`; this might be preferable for clarity, but I'm not sure if it creates the possibility of a specificity conflict. With the universal selector, the `.col` classes breakpoint- and variant-specific styles clearly override the ones they inherit by being direct-children of `.row`

EDITS: 

* restored the `make-col-ready` mixin with comment to the effect of deprecation, in order to not break dependent code

### DIFF
--- a/assets/scss/_component-examples.scss
+++ b/assets/scss/_component-examples.scss
@@ -48,7 +48,6 @@
 }
 
 .example-content-main {
-  @include make-col-ready();
 
   @include media-breakpoint-up(sm) {
     @include make-col(6);
@@ -60,7 +59,6 @@
 }
 
 .example-content-secondary {
-  @include make-col-ready();
 
   @include media-breakpoint-up(sm) {
     @include make-col(6);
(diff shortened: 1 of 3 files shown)


lunelson:
As a further optimisation, I suggest removing `position: relative` and adding explicit `flex` values, since flex-item defaults are consistent in all browsers

mdo:
Thanks, but we can't make this change. First, it'd break backward compatibility. Second, yes we do require columns to be immediate children—some folks can and do work around that—but we take a class first approach to our selectors. That also carries over to mixins to match.

Thanks though!

PR #25503, opened 2018-01-29
Replace unicode character, fixes #25391

### LINKED ISSUE #25391: v4.0.0 Scss Compilation Error: Invalid US-ASCII character “\xE2” (_hover mixin partial)
Good day, 

Was trying to update projects from beta 3 to v4.0.0 but kept receiving the following error when compiling some modules.
screen shot 2018-01-21 at 4 42 36 pm

After isolating the affected file _**mixins/_hover.scss**_ and going through it, seems to be the `-` character from _iOS-an._. in top comments (line 6). 
Replacing the `-` character with your own or setting `@charset: 'UTF-8';` at the top of the file does the trick.

scssboot

Thank you!

rvanlaak:
This will replace an incompatible Unicode character with it's ASCII variant.

Having a `4.0.1` hotfix would be preferable, as SCSS compilers like `sass 3.5.5` now stopped working.

```
Error: Invalid US-ASCII character "\xE2"
on line 6 of wwwroot/node_modules/bootstrap/scss/mixins/_hover.scss
from line 7 of wwwroot/node_modules/bootstrap/scss/_mixins.scss
from line 12 of wwwroot/styles/bootstrap.scss
```

Fixes #25391


XhmikosR:
Weird. Doesn't the Sass compiler work with UTF-8? node-sass is fine on the other hand, that is why I have said @mdo we shouldn't support anything that we don't test...

rvanlaak:
My guess would be that `sass` doesn't explicitly define all files as `utf8`? Explicitly adding `@charset "UTF-8";` on top of the file would indeed also fix compilation. So, `sass` does support `utf8`, but does not do that automatically. Don't think that line should be added on top of each file though.

What about having an "allowed to fail" check on Travis that tries to compile using `sass`? Are there actually any usage stats on what compiler is used most?

XhmikosR:
We don't use it at all officially. There's the gem project which does use it and they should test things already. But they just get releases so this would be caught with the next release I guess.

/CC @glebm

glebm (approved):

glebm:
This is not actually a Sass gem issue per se, because bootstrap-rubygem uses the Sass gem and the tests do pass. I guess it's an environment configuration issue. Nevertheless, there is no reason to keep that Unicode dash in that comment if it reduces compatibility.

PR #25548, opened 2018-02-04
Added Viewport Height & Width helpers

ItaloBC:
This allows the user to make a container (ideally) to use viewport height and width and allow better vertical/horizontal alignments of elements.


mdo:
I previously punted on these being responsive, but these are good to go. Just want to call that out and explicitly support this direction over that one.

mdo (approved):

guledali:
@mdo @XhmikosR  I'm curious why this PR over #27397

XhmikosR:
@guledali:

>I previously punted on these being responsive, but these are good to go. Just want to call that out and explicitly support this direction over that one.

PR #25553, opened 2018-02-05
Ensure modal scrolling is contained

mbardelmeijer:
Make use of the new `overscroll-behavior` tag to contain the scrolling in the modal itself.

Supported on Chrome 63 and above: https://caniuse.com/#feat=css-overscroll-behavior

More info: https://developers.google.com/web/updates/2017/11/overscroll-behavior#overlay

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -19,6 +19,7 @@
   z-index: $zindex-modal;
   display: none;
   overflow: hidden;
+  overscroll-behavior: contain;
   // Prevent Chrome on Windows from adding a focus outline. For details, see
   // https://github.com/twbs/bootstrap/pull/10951.
   outline: 0;


andresgalante (commented):
Hi @mbardelmeijer this is sweet. I didn't knew about `overscroll-behavior` :)

Since Bootstrap doesn't have an inner scroll within the modal I don't think it's necessary:
http://localhost:9001/docs/4.0/components/modal/#scrolling-long-content

Can you provide a codepen or a live example to show this in action with Bootstrap modal please?

mbardelmeijer:
Client reported some bugs with scrolling on Android devices so thought this would fix it, but i'm unable to replicate the bugs in the first place now on Chrome 64 emulator. Will close this for now and if i have any more info and/or a replication case, will reopen it :)

PR #25557, opened 2018-02-05
Make dark tables readable for print

### LINKED ISSUE #25453: Tables are not printing correctly
Hello,
today I've added feature to print tables from my website.
Then I have noticed that tables with classes below are **not printing correctly**, although they are displayed correctly on screen.

- `<thead class="thead-dark">` - Header and font are both white so header is not readable

- `<table class="table table-dark">` - Like header, whole table is white, including the font.

- `<table class="table table-striped">` - There are no striped rows when printing. Class has no effect when printing. Print result is the same with and without it.

At first I suspected that something is wrong with my website (that my CSS is overriding Bootstrap CSS because it is linked after it), but then I tried to print tables from [official Bootstrap website](https://getbootstrap.com/docs/4.0/content/tables/) and got the same issue.

I'm using Bootstrap v4.0

Hope this will be resolved soon.
Thanks.

MartijnCuppens:
Fixes #25453.

The text of dark tables/theads is white and invisible if the page is printed (some browsers don't print backgrounds). Wrapping the theming for dark tables in a `@media not print` media query will disable theming for dark tables for print.

Also I removed the lines which forced the `td`s and `th`s to have a white background. If a table/row/cell has an accented background, you would probably also want this when you print the page (if the browser supports coloured backgrounds).


mdo:
I think I'd rather see us override these styles in our `_print.scss` file. Thoughts?

PR #25565, opened 2018-02-06
set color to form-control-plaintext

m5o:
* set color to `.form-control-plaintext`
* add `$input-plaintext-color` default to `$body-color`
* fixes readability on dark themes like e.g. 'darkly'
  * https://bootswatch.com/darkly/index.html#forms
  * https://bootswatch.com/slate/index.html#forms
  * https://bootswatch.com/superhero/index.html#forms
  * https://bootswatch.com/solar/index.html#forms

/cc @thomaspark

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -121,6 +121,7 @@ select.form-control {
   padding-bottom: $input-padding-y;
   margin-bottom: 0; // match inputs if this class comes on inputs with default margins
   line-height: $input-line-height;
+  color: $input-plaintext-color;
   background-color: transparent;
   border: solid transparent;
   border-width: $input-border-width 0;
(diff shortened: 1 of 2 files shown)


andresgalante (dismissed):
Hi @m5o good catch! bare with me with this use case: 
If you want to keep the background of the inputs white with dark font color, but use the form on a dark background, this solution will force to have dark plaintext making it invisible.

What if we create a variable specific for `$plaintext-color` and we make it equal to `$input-color` by default?
```
$plaintext-color: $input-color
```

What do you think?

m5o:
> Hi @m5o good catch! bare with me with this use case:
> If you want to keep the background of the inputs white with dark font color, but use the form on a dark background, this solution will force to have dark plaintext making it invisible.

valid point. I'm ok with `$plaintext-color` this can be `$input-color` or `$body-color` by default. 👍 
Should I change it? Or wait for a comment from @mdo

andresgalante:
yes, `$body-color` makes more sense than `$input-color`. Thanks!

m5o:
🛎 @andresgalante Ping

mdo (approved):

PR #25572, opened 2018-02-07
Update stylelint config to add missing properties

in-in:
list of new properties:

```
align-content
animation-fill-mode
appearance
column-count
column-fill
column-gap
column-rule
column-rule-color
column-rule-style
column-rule-width
column-span
column-width
columns
direction
fill
flex-shrink
orphans
page-break-after
page-break-before
page-break-inside
perspective
src
stroke
unicode-bidi
widows
```


XhmikosR:
I don't have an order preference myself; I used the previous config from scss-lint which I think it was @mdo's choice.

Also, this breaks linting currently so it should be fixed. Or change the order of the offending properties.

mdo:
I'll review this and see about making the necessary changes for tests to pass.

XhmikosR:
@mdo: I changed the order so that tests pass. Don't forget to squash.

PR #25575, opened 2018-02-07
Variable darken percentage for emphasized links

### LINKED ISSUE #25574: Emphasized link
The darken percentage of a link is not the same as the darken percentage of an emphasized link (link with a class like `.text-success`). If we use a variable for this, it would also be possible to disable the darkened links on hover (sometimes you only want an underline on hover).
 
https://github.com/twbs/bootstrap/blob/9057d8da164bed683f49d30c6414753dad41de3e/scss/_variables.scss#L153

https://github.com/twbs/bootstrap/blob/9057d8da164bed683f49d30c6414753dad41de3e/scss/mixins/_text-emphasis.scss#L11

PR incomming

MartijnCuppens:
Closes #25574.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -166,11 +166,12 @@ $body-color:                $gray-900 !default;
 //
 // Style anchor elements.
 
-$link-color:                $primary !default;
-$link-decoration:           none !default;
-$link-hover-color:          darken($link-color, 15%) !default;
-$link-hover-decoration:     underline !default;
-
+$link-color:                              $primary !default;
+$link-decoration:                         none !default;
+$link-hover-color:                        darken($link-color, 15%) !default;
+$link-hover-decoration:                   underline !default;
+// Darken percentage for links with `.text-*` class (e.g. `.text-success`)
+$emphasized-link-hover-darken-percentage: 15% !default;
 
 // Paragraphs
 //
--- a/scss/mixins/_text-emphasis.scss
+++ b/scss/mixins/_text-emphasis.scss
@@ -8,7 +8,7 @@
   }
   a#{$parent} {
     @include hover-focus {
-      color: darken($color, 10%) !important;
+      color: darken($color, $emphasized-link-hover-darken-percentage) !important;
     }
   }
 }


andresgalante:
Hi @MartijnCuppens you are right they both should be 15%.

I don't think the implementation is correct though.

`$link-hover-color:` is allowing users to change the way they do hover, they can even change it to be lighten or any other function.

As you mention the `text-emphasized` mixin is hardcoding it.

By generating a variable to change the percentage we'd be allowing users to customize how dark the link color can be, but it's not flexible enough to produce the same results they'd get from changing `$link-hover-color:`.

Therefore I suggest a new variable like: 
```
$text-hover-color:                        darken($color, 15%) 
```

What do you think?

MartijnCuppens:
And how would you implement this `$text-hover-color`? Wouldn't this make all the hover colors of all links with a `text-*` color class the same?

MartijnCuppens:
I've got an idea. We could calculate the difference in lightness between `$link-color` and `$link-hover-color` with the [lightness()](http://sass-lang.com/documentation/Sass/Script/Functions.html#lightness-instance_method) function and than darken (or even lighten) the hover color of emphasized links.

MartijnCuppens:
@andresgalante, what do you thing of the latest changes? This doesn't require an extra variable and feels more flexible.

andresgalante:
Hi @MartijnCuppens there are other color functions apart from lighten and darken. Let me think about this one and get back to you.

@mdo any good idea of how to make this flexible and consistent?

mdo:
This one's tough. There's no real way to do what we do elsewhere—a variable for the entire thing like `$link-hover-color`. In this case, I think it's ultimately okay to limit things or have some constraints in an effort to avoid complexity. I think I'm fine with just a variable.

MartijnCuppens:
Jup, this was getting too complex indeed. I've reverted my changes to the initial commit. Is `$emphasized-link-hover-darken-percentage` a good name for the variable or do you have another suggestion, @mdo?

XhmikosR (approved):

mdo (approved):

XhmikosR:
@MartijnCuppens: this needs a rebase.

MartijnCuppens:
@XhmikosR, done.

PR #25578, opened 2018-02-07
Changing color of the .primary-button

ghost:
Change #007bff for #1660af


mdo:
Sorry, but we don't accept arbitrary changes to our color system. Next time either customize this on your own, or explain your changes better.

PR #25581, opened 2018-02-07
Allow non-string content as breadcrumb divider

### LINKED ISSUE #25579: SVG as Breadcrumb Divider
I am creating a custom theme for Bootstrap and would like to utilize a custom variables file as much as possible. However, I have an issue with the $breadcrumb-divider variable. By default it is set to 

`$breadcrumb-divider: "/";`

But, I want to use an SVG, which seems to be supported by the content property. For example something like this:

`$breadcrumb-divider: url(data:image/svg+xml,%3C%3Fxml%20version%...);`

The issue is that in `breadcrumb.scss` quotes are added around `$breadcrumb-divider` preventing this level of customization:

`content: "#{$breadcrumb-divider}";`

I know I can easily create my own breadcrumb.scss and override. But, it would certainly be cleaner and more maintainable just to change the variable. Figured it was worth asking about.

Thanks,
Eric

MartijnCuppens:
Fixes #25579

Allow non string content as breadcrumb divider. `$breadcrumb-divider` can also be set to `none` , that's why `padding-left: $breadcrumb-item-padding;` is set on `.breadcrumb-item` and not the `::before` pseudo element.


mdo:
Can we see a before/after example? And do we need to make any docs updates here to guide folks through customizing this?

MartijnCuppens:
There is no visual difference for this implementation. The static quotes in `_breadcrumb.scss` prevented other content than strings to be used as separator, this is now solved by using the `quote()` function in `_variables.scss`.

https://github.com/twbs/bootstrap/blob/0205fc8bdbd66c5c9101a95cd15c648a134a59a8/scss/_breadcrumb.scss#L18

I just updated the docs a bit to where I explained how the divider can be changed. It's also possible to use the `url()` function now, but this would in most cases require extra styling (as mentioned in #25579), so I didn't mention that in the docs.

andresgalante (commented):
Hi @MartijnCuppens good job, this fixes the issue and makes dividers more flexible.
I would add and example on the docs of how to use an svg by replacing:
```
$breadcrumb-divider:                quote("/")
```

width:
```
$breadcrumb-divider:                url("data:image/svg+xml,%3Csvg xmlns=...")
```

XhmikosR (approved):

mdo (approved):

mackelito:
Sorry for commenting on a old PR but I just was this...
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_breadcrumb.scss

What is the point of the hack selectors? (line: 30 and 34)
Looks as a misstake.. or?

The last selector will always be used.. 
Also there is comments to suppress the linting error.. ?

MartijnCuppens:
@mackelito
https://github.com/twbs/bootstrap/blob/47ef0a877e6fc18c1e516f7a0855e0531c77af96/scss/_breadcrumb.scss#L24-L29

mackelito:
@MartijnCuppens Sorry for this.. we used the breadcrumb css/html in a non bootstrap project and realised that this is just a BS bug and not a "native hack thing".. so the hack was not needed for us 👍

PR #25590, opened 2018-02-09
Docs: Fix the incorrect id in the vertically centered modal snippet

ysds:
Make it equal to the value of `aria-labelledby`.


patrickhlauke (approved):

patrickhlauke:
Good spot, thank you.

PR #25598, opened 2018-02-11
Consolidate cursor resets in Reboot

mdo:
After one of the comments in #25351, I think we can consolidate our `cursor` resets on non-disabled anchors and buttons into Reboot. 

- Removes pagination link, `.navbar-toggler`, `.close`, and `.btn` `cursor: pointer` resets.
- Adds new ones to Reboot for links and buttons.

/cc @XhmikosR @patrickhlauke


patrickhlauke:
wondering if this doesn't get us back where we started with regards to authors using random elements, for whatever reason, with, say `.btn` class. like (assuming they did it at least accessibly) `<span tabindex="0" role="button" class="btn">Foo</span>` or something.

mdo:
With something like that, there's still no reason to not use a `<button>` element, correct?

XhmikosR:
I too think it sort of brings us back to the previous situation but it's simpler for us this way.

Now, guys, it's your call, but personally I'd go with this and with the docs stating to prefer the proper element.

mdo:
> I'd go with this and with the docs stating to prefer the proper element.

Yeah, noted. I can get some docs updates in here.

scss/_reboot.scss
```diff
@@ -205,6 +205,11 @@ a {
     color: $link-hover-color;
     text-decoration: $link-hover-decoration;
   }
+
+  // Reset the `cursor` on non-disabled links. Also applies to buttons.
+  &:not(:disabled):not(.disabled) {
+    cursor: pointer;
```
patrickhlauke: this would also force `cursor:pointer` for any anchors (`<a name="blah">`, or more generally `<a>` elements without an `href`). this will likely make react people happy (as they're usually the ones complaining, if memory serves), but will end up with confusing experience on sites that do use static anchors that are not links?

patrickhlauke: solution if we did want to just do this for real links would be to change selector to something like `&[href]:not(disabled):not(.disabled)`

patrickhlauke:
i'll admit i'm conflicted. on the one hand, sure, using semantically correct elements should be the preferred approach. on the other hand we know for a fact not all our authors do it, and then come here and complain that we're not adding the `cursor:pointer` as part of the class.

but i guess if we actually document this, advise authors to manually add their own `cursor:pointer` if for whatever reason they can't use the correct element, and strictly point people to that part of the docs when they come here and open issues...that'd probably be ok?

scss/_reboot.scss
ysds: Is this code necessary? When `<a>` is a link, browser makes the cursor to pointer by default.

patrickhlauke: ah, good point. yes, it seems at that point, it's redundant (and nicely circumvents the whole "what about anchors" dilemma)

XhmikosR:
Should we drop this in favor of #27021?

PR #25600, opened 2018-02-12
Custom range input

mdo:
Builds on #23380 with additional commits:

- Adds some docs notes about the track/thumb
- Splits out the active and focus styles to each browser's specific rulesets so they apply
- Updates the Firefox inner focus override—old one didn't work
- Add note for `padding-left` override for which browser it affects

/cc @gijsbotje


docs/4.0/components/forms.md
```diff
@@ -1171,6 +1171,20 @@ As is the `size` attribute:
...
 
+### Range
+
+Create custom `<input type="range">` controls with `.custom-range`. The track (the background) and thumb (the value) are both styled to appear the same across browsers. As only IE and Firefox support "filling" their track from the left or right of the thumb as a means to visually indicate progress, we do not currently support it.
+
+{% example html %}
+<input type="range" class="custom-range">
+{% endexample %}
+
+Additionally, you may specify steps by adding `min` and `max` attributes.
```
patrickhlauke: To be a bit more specific: if you don't add a `min`/`max` to a range input, it defaults to `min=0` and `max=100`. if you don't add a specific `step`, the step value is 1.

So you're not really "specifying steps" by adding `min` and `max`. You're going from a default of 101 steps to a more specific number of steps. you could set `min=0 max=5 step=0.05` and have 101 steps again.

patrickhlauke: also, for both examples, you should really have either a `<label for="...">` or `aria-label` or similar...

mdo: Noted. I'll take a pass at updating the docs.

mdo:
Have a look @patrickhlauke at the latest commit, please.

docs/4.0/components/forms.md
```diff
@@ -1176,13 +1176,15 @@ As is the `size` attribute:
 Create custom `<input type="range">` controls with `.custom-range`. The track (the background) and thumb (the value) are both styled to appear the same across browsers. As only IE and Firefox support "filling" their track from the left or right of the thumb as a means to visually indicate progress, we do not currently support it.
 
 {% example html %}
-<input type="range" class="custom-range">
+<label for="customRange1">Example range</label>
+<input type="range" class="custom-range" id="customRange1">
 {% endexample %}
 
-Additionally, you may specify steps by adding `min` and `max` attributes.
+Range inputs have implicit values for `min` and `max`—`0` and `100`, respectively. You may specify new values for those using the `min` and `max` attributes.
```
patrickhlauke: Maybe also include at the end:
```
By default, range inputs "snap" to integer values. To change this, you can specify a `step` value.
```
and then perhaps add `step="0.5"` to the following example? Or is that getting too specific?

patrickhlauke (approved):
changes look good. maybe add the extra mention of `step`, but not essential if you feel it gets too specific...

patrickhlauke (approved):

patrickhlauke:
Perfect 👍

mdo:
I've added new variables for everything. I also have some comment notes here—need to document answers to the questions in my comments as some of the values seem arbitrary or limited to a specific browser. Would be nice to describe those things.

PR #25602, opened 2018-02-12
Fix `custom-forms` heights and position

unrevised6419:
Now `.custom-control` will have height equal to `$font-size-base`.
Fixed custom checkbox and radio top position (we can't subtract rem from unit)
`$line-height-base - $custom-control-indicator-size` is not valid
Because `$line-height-base` is a unitless variable and `$custom-control-indicator-size`
is a unit (rem) variable


andresgalante (approved):
thanks!

Issue #25611, opened 2018-02-13
Grammar issue in Alerts section of documentation


ChrisCallanJr:
This issue can be found [here](https://getbootstrap.com/docs/4.0/components/alerts/).

I found this on Chrome, but I am  fairly certain this can be replicated on any browser.

![screen shot 2018-02-13 at 9 17 37 am](https://user-images.githubusercontent.com/15125965/36154628-095f6114-10a0-11e8-8440-8bede44679ba.png)
This is one of 4 of this issue found on the page. It should be "This is **an** info alert", not "This is **a** info alert".

patrickhlauke:
that part of the documentation is generated programmatically
```
{% for color in site.data.theme-colors %}
<div class="alert alert-{{ color.name }}" role="alert">
  This is a {{ color.name }} alert with <a href="#" class="alert-link">an example link</a>. Give it a click if you like.
</div>{% endfor %}
```
and there's no straightforward way (that I know of) to do any "if first letter is a vocal, write 'an', otherwise 'a'"

we can either device some programmatic way of doing this, live with it, or find another wording that avoids this (e.g. just dropping "This is a" part). @mdo ?

patrickhlauke:
took at stab at rewording - #25614

PR #25614, opened 2018-02-13
[Docs]: Reword alert and list group text to avoid "a/an" problems

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/25611


mdo (approved):

patrickhlauke:
@mdo do we add these sorts of small trivial things to ship list?

PR #25624, opened 2018-02-14
Be more explicit where to find variables 

pgiraud:
The documentation already tells that variables are in `_variables.scss`. However it's done in specific sections only (such as global options or colors). Some users may not read this.

With this PR the reference to the `_variables.scss` file is given earlier in the documentation. I hope it's not too redondant.

### DIFF
--- a/docs/4.0/getting-started/theming.md
+++ b/docs/4.0/getting-started/theming.md
@@ -76,6 +76,8 @@ With that setup in place, you can begin to modify any of the Sass variables and 
 
 Every Sass variable in Bootstrap 4 includes the `!default` flag allowing you to override the variable's default value in your own Sass without modifying Bootstrap's source code. Copy and paste variables as needed, modify their values, and remove the `!default` flag. If a variable has already been assigned, then it won't be re-assigned by the default values in Bootstrap.
 
+You will find the complete list of Bootstrap's variables in `scss/_variables.scss`.
+
 Variable overrides within the same Sass file can come before or after the default variables. However, when overriding across Sass files, your overrides must come before you import Bootstrap's Sass files.
 
 Here's an example that changes the `background-color` and `color` for the `<body>` when importing and compiling Bootstrap via npm:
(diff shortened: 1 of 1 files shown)


docs/4.0/getting-started/theming.md
```diff
@@ -76,6 +76,8 @@ With that setup in place, you can begin to modify any of the Sass variables and
 
 Every Sass variable in Bootstrap 4 includes the `!default` flag allowing you to override the variable's default value in your own Sass without modifying Bootstrap's source code. Copy and paste variables as needed, modify their values, and remove the `!default` flag. If a variable has already been assigned, then it won't be re-assigned by the default values in Bootstrap.
 
+You will find the complete list of variables that can be overriden and their default values in our `scss/_variables.scss` file.
```
andresgalante: Hi @pgiraud thanks for the contribution, I this is makes it more clear. I'd change the wording from "our" to "Bootstrap's". For example:
"You will find the complete list of Bootstrap's variables and their default values in "

andresgalante (changes requested):

docs/4.0/getting-started/theming.md
pgiraud: I agree with you. I don't like the 'our' either. I just tried to be consistent.
Changed.

docs/4.0/getting-started/theming.md
```diff
@@ -226,7 +228,7 @@ You can also specify a base color with our color map functions:
 
 Customize Bootstrap 4 with our built-in custom variables file and easily toggle global CSS preferences with new `$enable-*` Sass variables. Override a variable's value and recompile with `npm run test` as needed.
 
-You can find and customize these variables for key global options in our `_variables.scss` file.
+You can find and customize these variables for key global options in our `scss/_variables.scss` file.
```
andresgalante: same here, just remove "our"

docs/4.0/getting-started/theming.md
```diff
@@ -298,7 +300,7 @@ An expansive set of gray variables and a Sass map in `scss/_variables.scss` for
   </div>
 </div>
 
-Within `_variables.scss`, you'll find our color variables and Sass map. Here's an example of the `$colors` Sass map:
+Within `scss/_variables.scss`, you'll find our color variables and Sass map. Here's an example of the `$colors` Sass map:
```
andresgalante: and here replace "our" with "Bootstrap's"

andresgalante (changes requested):

pgiraud:
Updated.

andresgalante (approved):
Thanks a lot @pgiraud

mdo (approved):

Issue #24903, opened 2017-11-28
Responsive container-fluid


fulldecent:
FEATURE REQUEST: `.container-*-fluid`

DESCRIPTION: Switch to fluid layout for viewports larger than a certain size.

Use case: https://codepen.io/anon/pen/WXaoyo?editors=1100#0

Implementation for `.container-xl-fluid`:

```css
@media (min-width: 1200px) {
  .container-xl-fluid {
    width: 100%;
    margin-right: auto;
    margin-left: auto;
    padding-right: 15px;
    padding-left: 15px;
    width: 100%;
    max-width: none;
  }
}
```

exige81:
You can basically do this with the existing responsive [flex direction utilities](https://getbootstrap.com/docs/4.0/utilities/flex/#direction)

example: https://codepen.io/anon/pen/RjeBMG

fulldecent:
Isn't that the same as just normal responsive rows?

https://codepen.io/anon/pen/zPmVGJ

```html
<div class="container-fluid">
  <div class="row">
    <div class="col-xl">
      <p class="alert alert-success">Left</p>
    </div>
    <div class="col-xl">
      <p class="alert alert-primary">Right</p>
    </div>
  </div>
</div>
```

The problem I am trying to solve is for breakpoints `-sm`, `-md` and `-lg`with a single column, there are the nice margins that `.container` has.

![screen shot 2017-11-29 at 8 51 32 am](https://user-images.githubusercontent.com/382183/33378390-a0eea782-d4e2-11e7-923a-0084ae79da1f.png)
![screen shot 2017-11-29 at 8 51 40 am](https://user-images.githubusercontent.com/382183/33378396-a2fee172-d4e2-11e7-9e9b-25a0bc7df19c.png)
![screen shot 2017-11-29 at 8 51 48 am](https://user-images.githubusercontent.com/382183/33378403-a4ef0f84-d4e2-11e7-8658-dc3a2ae424a6.png)
![screen shot 2017-11-29 at 8 51 56 am](https://user-images.githubusercontent.com/382183/33378407-a723e266-d4e2-11e7-9e05-a6029c906767.png)
![screen shot 2017-11-29 at 8 52 07 am](https://user-images.githubusercontent.com/382183/33378412-aa54a826-d4e2-11e7-8a38-b9a16be9fb98.png)

mdo:
[...]

PR #25631, opened 2018-02-15
create responsive containers

browner12:
### Motivation

Currently with `.container` and `.container-fluid` we have an all or nothing approach to either a full width or fixed width layout.  I would like to make these containers responsive to give the end user full control over when the switch to a fixed width layout occurs.

It was brought up in PR #24903 , and discussed on the Slack channel. The use case I am addressing in this PR is actually the opposite of what's being asked in the other one, because I see this as the more 99% use case.

### Example

The most common use case I see is the user who wants their **extra small** and **small** devices to maintain a fluid layout since real estate is at a premium, and only switch to a fixed layout on **medium** devices and larger where you have much more width.

### Approach

If you dig into the source on containers you realize that on **extra small** devices, `.container` and `.container-fluid` are the same. It's not until you hit the larger sizes that `max-width`s start getting applied.  The approach I use is similar to all the other responsive classes where styling is applied to the breakpoint and up.  This PR introduces 4 new classes which can be used in place of existing container classes.

- `.container-sm`
- `.container-md`
- `.container-lg`
- `.container-xl`

As an example,

```html
<div class="container-md"></div>
```

would mean the container stays fluid until you hit the **medium** breakpoint, and from there on up it would be fixed width.  This table gives you a full breakdown.

| | XS Screen | SM Screen | MD Screen | LG Screen | XL Screen |
| ------------ | ------------- | ------------- | ------------- | ------------- | ------------- |
| `.container` | **fixed** * | **fixed**| **fixed** | **fixed** | **fixed**
| `.container-sm` | fluid | **fixed**| **fixed** | **fixed** | **fixed**
| `.container-md` | fluid | fluid| **fixed** | **fixed** | **fixed**
| `.container-lg` |  fluid | fluid| fluid | **fixed** | **fixed**
| `.container-xl` | fluid | fluid| fluid | fluid | **fixed**
| `.container-fluid` |  fluid | fluid| fluid | fluid | fluid

As you can see, the `.container` and `.container-fluid` maintain the same functionality so backwards compatibility is maintained.

### Next Steps

I imagine this PR will require a decent amount of discussion. One point to address is if it is worth it to be able to go the other direction as well, and switch from a fixed layout to a fluid layout after a certain breakpoint.

Also, I'm sure I could also use some help on optimizing the SCSS.


browner12:
It might be easier to comprehend if you can see the (desired) compiled code, so here is the CSS:

```css
.container,
.container-fluid,
.container-sm,
.container-md,
.container-lg,
.container-xl {
  width: 100%;
  padding-right: ($grid-gutter-width / 2);
  padding-left: ($grid-gutter-width / 2);
  margin-right: auto;
  margin-left: auto;
}

@media (min-width: 576px) {
  .container,
  .container-sm {
    max-width: 540px;
  }
}

@media (min-width: 768px) {
  .container,
  .container-sm,
  .container-md {
    max-width: 720px;
  }
}

@media (min-width: 992px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg {
    max-width: 960px;
  }
}

@media (min-width: 1200px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg,
  .container-xl {
    max-width: 1140px;
  }
}

```

browner12:
ping @chasepalmer

andresgalante:
Hi @browner12 Thanks a lot for your contribution, the idea is clear and intresting, and I see how this can be useful.

I got your branch and test it out and it works as it should.

I don't know if the naming `container-*` represent what it does, but naming is really hard and I can't come up with a better option at the moment.

I'll be up to @mdo to decide on this one 😄

XhmikosR:
This is something I always meant to look into myself since I needed it for a project, but worked around it locally and forgot about it :)

I'd like to see this merged too, so 👍 from me.

fulldecent:
Why not use the same approach here as is already being done with rows/columns:

```css
.container,
.container-fluid {
  width: 100%;
  padding-right: ($grid-gutter-width / 2);
  padding-left: ($grid-gutter-width / 2);
  margin-right: auto;
  margin-left: auto;
}

@media (min-width: 576px) {
  .container,
  .container-sm {
    max-width: 540px;
  }
  .container-sm-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}

@media (min-width: 768px) {
  .container,
  .container-sm,
  .container-md {
    max-width: 720px;
  }
  .container-sm-fluid,
  .container-md-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}

@media (min-width: 992px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg {
    max-width: 960px;
  }
  .container-sm-fluid,
  .container-md-fluid,
  .container-lg-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}

@media (min-width: 1200px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg,
  .container-xl {
    max-width: 1140px;
  }
  .container-sm-fluid,
  .container-md-fluid,
  .container-lg-fluid,
  .container-xl-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}
```

browner12:
If we can give the user this full control for whether their breakpoint is fluid or fixed, I'm all for that.  Even though I feel like switching from a fixed container on a smallish device to a fluid container on a largish device will be a very uncommon use case.

My one concern with your code is we will need to reset the `max-width` property for all higher breakpoints, since our media queries to do have an upper bound, but simply a **min**-width.

For example:

```html
<div class="container-fluid container-sm container-md-fluid"></div>
```

The relevant media queries would be:

```css
@media (min-width: 576px) {
  .container,
  .container-sm {
    max-width: 540px;
  }
  .container-sm-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}

@media (min-width: 768px) {
  .container,
  .container-sm,
  .container-md {
    max-width: 720px;
  }
  .container-sm-fluid,
  .container-md-fluid {
    width: 100%;
    padding-right: ($grid-gutter-width / 2);
    padding-left: ($grid-gutter-width / 2);
    margin-right: auto;
    margin-left: auto;
  }
}
```

This would mean on a device >768px even though it has a `width: 100%` from the `container-md-fluid`, it still has the `max-width:540px` from the `.container-sm`.  Therefore it would still (I believe) behave as a fixed width container at this point.   It may be as simple as resetting it with `max-width:none` for those responsive fluid classes.

browner12:
any movement on this? I'm not sure what the turnaround time is like on new features in this repo.

patrickhlauke:
> I'm not sure what the turnaround time is like on new features in this repo

"hey, i know you're all volunteers, doing this for free in your spare time to give away for free...but any idea when you'll have this done or what?"

browner12:
sorry @patrickhlauke, the way you read it is not at all the way I meant it.

just legitimately curious if we like to take the new features a little slower to flesh them out more. For the projects I contribute to, I have a better understanding of the pacing of things, and was just trying to get a feel for BS.

I absolutely share your sentiment on people feeling entitled to OSS, and thanks for all you do in this project.

patrickhlauke:
yeah sorry, flew off the handle there...hence deleted the comment. missed the context that this was a PR as well.

mdo:
@browner12 Sorry for the delay! I'm going to pull this down and write some docs for it.

mdo:
Checking it out locally, the code doesn't generate what you've put into your desired output comment.

browner12:
yah, i think i need some help with the Sass.

i'll give it a shot this week

mdo:
Yeah, I'll see about trying to take another look, too :). These layers of loops, maps, and what not always trip me up.

browner12:
@mdo I was able to clean up the top part with an `@extend` directive.

I'm running into an issue in the media queries where `@extend` returns the error 

> you may not extend an outer selector from within media

which I feel shouldn't apply because I'm trying to extend a selector from within the media query.

Any thoughts on that part?

browner12:
bump

browner12:
could we put this on either the 4.2 or 4.3 project board?

mdo:
@browner12 Haven't forgotten about this. v4.2 is up after v4.1.2, so stay tuned!

browner12:
thanks @mdo ! 4.1 and 4.2 are looking great!

mdo:
Checked this out locally—`.container-` specific changes look good. However, I realized that this may require additional changes for the navbar. There are `.navbar .container` and `.navbar .container-fluid` selectors in there. Not sure the best way forward on that just yet...

browner12:
hey @mdo. made the necessary change to make the containers work in the navbars. would love a second set of eyes on it.

I don't love how I had to add the global scoped variable to the top of `_navbar.scss`. Was wondering if it would be good to move this into `_variables.scss` since I'm now using it in more than one file.

mdo:
Thanks @browner12! Looking at the diff, we've lost `.container-fluid` outright now.

Also, curious if we can get away from some of the complexity with `.navbar [class^="container-"] { ... }` instead of listing all the responsive container classes out again?

browner12:
what do you mean by, "lost `.container-fluid` outright now"?

for listing them all out again, are you talking about this part:

```scss
$combined-container-max-widths: map-merge(
  (
    xs: 0,
    fluid: 9999px,
  ),
  $container-max-widths
) !default;
```

I was actually going to ask about this, because as I was writing it, it was feeling a little messy.  Since we use this same code on both `_grid.scss` and `_navbar.scss`, would it be appropriate to extract it to `_variables.scss`?

rafalp:
Quick look at current diff shows `.container-fluid` class is gone from the sass code, that may be what @mdo meant by "lost `.container-fluid` outright now".

browner12:
correct, we remove `.container-fluid` and add it the the map that we loop over.

could we extract the `map-merge` to `_variables.scss`?

rafalp:
I'm not sass pro, but if your change means that `class="container-fluid"` is gone, then this is breaking change that can't be merged to 4.2

browner12:
it doesn't remove it from the compiled code, it's just doesn't explicitly declare it separately in the SASS anymore.

pauloavelar:
> What is the purpose of this PR?
> Afaik `container` class already scales for different viewports perfectly. Can OP olaborate why to add additional class for each viewport?

Just to give a real use case of this PR, I am developing a responsive form and it's currently _broken-ish_ between 768px and 991px (sm).

The base layout (sm+) uses two columns (8/4) on a container. I don't want to switch to our mobile layout for sm resolutions, There is _just enough_ width to display both columns pretty well up to 780px, but the container's width is fixed to 720px on sm. I get that we need to support 768px anyway, so might as well apply media queries all the way up to 991px, but I'd rather sacrifice consistency over a good UX for most of sm resolutions.

In this case, I would use container-fluid up to sm and then switch to container for md+. In the past, I have implemented this behavior using container-fluid all the way and applying offsets on md+, but the downside is that the content keeps resizing.

browner12:
@MartijnCuppens anything I can do to help get this in 4.2? really don't want to wait until 4.3 for this.

MartijnCuppens:
@browner12, 

I understand, but we've decided to postpone all new features to `v4.3`. There are a lot of features I personally would love to have in `v4.2`, but we're at a point we can no longer postpone the release of `v4.3` anymore.

browner12:
gotcha, thanks for the response.

browner12:
With your first commit, is there a benefit to removing the `map-merge`? My one concern here is we're now duplicating the `@extends %containers;` code, which could lead to divergence if people don't realize they're doing the same thing.

It looks like your second commit is the same thing, but for the `_navbar.scss`.

This is why I asked/suggested we move the map merge to `_variables.scss`. Then we could refactor that duplication to just one spot, and then still keep my original loops here.

Thoughts?

It looks like what you have here will still compile the same, BTW.

MartijnCuppens:
The values of the map weren't used, didn't make sense to merge maps here. The generated code should indeed be the same.

We 'll need some docs for this. I guess a separate examples page will be most interesting to demonstrate this?

browner12:
Sorry, I don't follow that. What do you mean "the values weren't used"?  Are you referring to the '0' and '9999px' values? because the keys were definitely being used in the loop.

MartijnCuppens:
Yup, was talking about the `0` and `9999px`

browner12:
I guess I preferred the original code, because it avoided the duplication, but it's not a hill I'm going to die on.

They both seem to generate the same output, so I'm okay with it.

I think the documentation could go on this page: https://getbootstrap.com/docs/4.1/layout/overview/#containers

dillix:
@browner12 how to make resp. containers in v4.2 release?

browner12:
it was not merged in yet, so you'll have to wait until 4.3

mdo:
I might be overthinking this after coming back to the PR a few times over the last several months, but this feels a little off still. The latest changes from @MartijnCuppens output multiple container classes per media query when I think we'd want a single container per breakpoint.

For example, here's what @browner12's earliest comment shows for compiled CSS, and this is matched by @MartijnCuppens's latest changes.

```css
.container,
.container-fluid,
.container-sm,
.container-md,
.container-lg,
.container-xl {
  width: 100%;
  padding-right: ($grid-gutter-width / 2);
  padding-left: ($grid-gutter-width / 2);
  margin-right: auto;
  margin-left: auto;
}

@media (min-width: 576px) {
  .container,
  .container-sm {
    max-width: 540px;
  }
}

@media (min-width: 768px) {
  .container,
  .container-sm,
  .container-md {
    max-width: 720px;
  }
}

@media (min-width: 992px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg {
    max-width: 960px;
  }
}

@media (min-width: 1200px) {
  .container,
  .container-sm,
  .container-md,
  .container-lg,
  .container-xl {
    max-width: 1140px;
  }
}
```

But wouldn't we want this instead? [See the CodePen demo.](https://codepen.io/emdeoh/pen/KbQGxx?editors=1100) (Forgive the shorthand attribute selector, and note `.container-fluid` is absent as it's covered by that selector.) 

```css
[class^="container"] {
  padding-right: 1.5rem;
  padding-left: 1.5rem;
  margin-right: auto;
  margin-left: auto;
}

.container {
  max-width: 980px;
}

@media (min-width: 576px) {
  .container-sm {
    max-width: 540px;
  }
}

@media (min-width: 768px) {
  .container-md {
    max-width: 720px;
  }
}

@media (min-width: 992px) {
  .container-lg {
    max-width: 960px;
  }
}

@media (min-width: 1200px) {
  .container-xl {
    max-width: 1140px;
  }
}
```

browner12:
I don't think that behaves the same.

For example, let's assume 

```html
<div class="container-sm"></div>
```

and a viewport width of 800px.

With my original code the `<div>` would fall under the following media query

```css
@media (min-width: 768px) {
  .container,
  .container-sm,
  .container-md {
    max-width: 720px;
  }
}
```

and have a width of 720px.

With your new code, it would fall under the following media query

```css
@media (min-width: 576px) {
  .container-sm {
    max-width: 540px;
  }
}
```

and have a width of 540px.

With that new code the container doesn't grow as the screen size grows.

MartijnCuppens:
It all depends on what we want. The behaviour of this PR and [the codepen](https://codepen.io/emdeoh/pen/KbQGxx?editors=1100) of @mdo is indeed different.
As a comparison, this is the behaviour of the current implementation: https://codepen.io/MartijnCuppens/pen/pqaGzd?editors=1100

I personally think case of @mdo would be used more. @andresgalante, what do you think?

browner12:
I strongly disagree.

Let's take a more extreme example. Let's again assume HTML of

```html
<div class="container-sm"></div>
```

but this time a screen size of 2000px. With the new code we would **still** have a container width of 540px, meaning it's taking up approximately 1/4 of the screen width. Even if we continue growing the screen size, 2100px, 2200px, the container is still *stuck* at 540px.

With this new code, you are changing the fundamental purpose behind containers. Previously we essentially had 2 states, a `.container-fluid` state (1) where the element was always 100% of its parent, and a `.container` state (2) where the element was approximately between 70-95% of its parent. This was good because it allowed us to effectively use our screen width.

Now we've got 5 distinct states:

1. `.container-sm` 540px
1. `.container-md` 720px
1. `.container-lg` 960px
1. `.container` 980px <-- where did this one come from?
1. `.container-xl` 1140px

and a very inefficient use of our screen width.

My original PR maintains those 2 original states, but gives us the ability to decide at which breakpoint we want to switch from state 1 to state 2.

You've also broken all of the backwards compatibility. A `.container` is now stuck permanently at 980px rather than adapting to the screen width.

browner12:
Your codepen is not accurate.  Try this one instead:

https://codepen.io/anon/pen/YdawZJ?editors=1100

rafalp:
`.container-xx` that never gets wider than specified for it seems more usable (speaking from experience). I am catching myself plenty of times when I want to limit width of some content (eg. message). @mdo proposition appears more clean than what I am currently doing (`.container .row .col-md-8`).

browner12:
A good chunk of the classes in Bootstrap use `sm`, `md`, `lg`, and `xl` to indicate the breakpoint it applies at.  look at all of the utility classes such as `m-md-5`. We are not saying "a medium margin", we are saying "a margin of 5 at the medium breakpoint".  Or look at a column class like `col-lg-5`. We are not saying "a large column", we are saying "5 columns at the large breakpoint".

This PR stays consistent to that convention.

I understand classes like `modal-sm` and `modal-lg` exist where 'sm' and 'lg' are actually referring to sizes. We also have this convention with `.btn`s, `.btn-group`s, and `.form-control`s.  I can see the concern for the confusion this could cause. However, the overlap already exists in v4, so I don't think the worry of confusion should prevent us from adding it to v4.  People will just have to read the documentation to use it correctly, just like they do currently.

If we want to address this concern as a whole in v5, I'm down for that.  I think people pointed out Tailwind's convention in one of @mdo's recent tweaks asking for v5 suggestions, where Tailwind prefixes the class for breakpoints.  For example `.md:font-bold`.

andresgalante:
It doesn't matter what we do, @browner12 is right:

> People will just have to read the documentation to use it correctly, just like they do currently.

And as @MartijnCuppens pointed out, the behaviors are different. As an author, I'd expect it to behave like the original proposal by @browner1, but I can see how @mdo's behavior is more useful.

If I had to choose one out of a gut feeling and not a real use-case, I'd go with @mdo's.

Having said that why don't we offer both? Similar to how we do `media-breakpoint-up` or `media-breakpoint-down`

markfenner:
@browner12 stick to your idea - you are getting a too much stick for a positive enhancement and a good implementation imo.

Naming convention - using the suggestion solution by @browner12 follows current grid (col, offset) and layout (margin, padding, display) formats and is intuitive - keep it.

Switching to fluid layout on larger screens is uncessessary. The goal is to maximise usage of the real estate on small screens. Start with that as per previous PRs and if it becomes a common request to do it the other way add this in the future.

As a working example iPhone 6/7/8 (a significant proportion of our visitors) has a display of 750px which means we waste available space above 576px on fixed width containers. If I switch to fluid layout it looks great on the iPhone but becomes terrible on XL monitors.

Please @mdo consider this for v4.3 it’s a positive improvement :)

Well done @browner12

mhlozek:
> The issue for me is the naming relative to the behavior, and the inconsistencies it could bring.
> 
> For example, with this PR we'd have a `.container-sm` that could be the width of the standard `.container` at its `xl` breakpoint. That seems ripe for confusion, especially when our [modals behave as my demo is built](https://getbootstrap.com/docs/4.2/components/modal/#optional-sizes)—the `.modal-sm` is the same width from the `sm` breakpoint and up.
> 
> I understand this PR's intended behavior better now, but I don't know how to refer to it yet without confusing myself.

I have some kind of my implementation is scss and we use naming like this:
.container-md-8 (max width of eight columns in pixels for resolution from md up)
.container-lg-10 (max width of 10 columns in pixels for resolution from lg up)
...

We use this mainly for text blocks, where text in "full" container is too long for comfortable reading.  
And just my two cents: consider if is really necessarily to add margin-left: auto and margin-right: auto for this elements. It is easy to do this in html with class mx-auto:
<div class="container-md-8 mx-auto">...</div>

mrlife:
Really looking forward to this being baked in! It’s something I use on every website I make.

marko-avlijas:
Please merge this

Soviut:
My use case for wanting `.container-*-fluid` is to only turn on fluidity _below_ a certain breakpoint.

I have a layout that uses non-fluid containers at desktop sizes so the site, and most importantly, it's header/navbar, lines up with the rest of the page.

However, once you get to tablet and mobile sizes, I want that header/navbar to become fluid, so that the brand/logo sits as far left as possible and the nav menu sits as far right as possible.

I've solved this using my own custom class and media query

```scss
  .container-md-fluid {
    @include media-breakpoint-down(md) {
      max-width: 100%;
    }
  }
```

browner12:
@Soviut if I'm understanding you correctly, that's exactly what my PR solves.  You would use `container-lg` and all screen sizes MD and down would be fluid, while LG and up would be fixed.

Soviut:
@browner12 yeah, I was just reinforcing that use case since there were some proposals that seemed like the opposite (fluid on larger screens).

My only issue is the name; I feel like the `container-*-fluid` is more in keeping with the convention set by other breakpoint classes. eg: `d-lg-flex`, `col-md-6`, `text-sm-right`, etc.

Tylerian:
> Can someone who likes the alternate behavior please explain to me your reasoning and a possible use case? I am honestly struggling to see the benefit of that approach. Why would people want Bootstrap predefined `max-width`s when they could just add their own specific ones as needed?
> 
> ```
> <div class="container-sm"></div>
> 
> <div style="max-width:600px"></div>
> ```
> 
> Back in my original comment, I laid out a (IMO) very plausible use case, and one that I've run into **many** times
> 
> > The most common use case I see is the user who wants their extra small and small devices to maintain a fluid layout since real estate is at a premium, and only switch to a fixed layout on medium devices and larger where you have much more width.
> 
> The purpose of the `.container`, as it currently stands is to be _responsive_ and adapt to the growing screen width to use as much real estate as possible. We are trying to add to this ability to effectively use screen width by giving the user the ability to choose where the container switches from fluid to fixed.
> 
> The alternate behavior takes the responsive aspect of the `.container` away. I think it will also add to the confusion, because `.container` will behave differently than all of the others. With this approach, I can also see in the future people saying "Well, I want the `<div>` to be a `.container-sm` on small and medium devices, but I want it to be a `.container-lg` on large and extra large devices." and then we're basically back to where we started.

I found myself on the same boat once and my solution matched the one from @mdo.

In my experience, when you build a mobile-first website, you want your layout to take the full width of the mobile device, usually up to a small laptop.
Extrapolating to bootstrap named breakpoints, you want your layout to grow in width up to the `xl` breakpoint.

With @mdo approach, you would set the `.container-xl` class in your container and it would do the job seamlessly, starting fluid and constraining on the `xl` breakpoint. Similarly, if you wanted your layout to grow up to the `md` breakpoint, you would set your container class to `.container-md` and you are done. On the other hand, with your approach, the container grows fluid to the desired breakpoint, and then it adapts the behaviour of the default `.container` class, increasing in width at each grid breakpoint. For me, this feels wrong and not the natural way to do, but hey, again, I'm talking based on my own experiences and of course, I might be wrong.

browner12:
@Tylerian so if I'm understanding you correctly, you would want `.container.md` to be fluid up til the MD breakpoint, and then stick at a defined width for all wider sizes?  That would mean on a 2000px wide screen, the container would be 720px wide.

IMO, this looks silly, and is not a common use case.  As I stated before, this takes away the responsiveness of the container, and people should write their own code for this.

Do you have a live example where you've used it this way?  It might be helpful for me to see.

browner12:
If you liked the original functionality of this PR, please voice your concern on the new one. It has been merged, and completely changes how this feature will work.

https://github.com/twbs/bootstrap/pull/29095

browner12:
hallelujah, thanks guys!

jandante:
Just for clarification. This will go into 4.4 as described by @browner12 in the initial post?

MartijnCuppens:
@jandante: Yes, you can view the ship list here: https://github.com/twbs/bootstrap/issues/29096

The `v4.4.0` project board can be found here: https://github.com/twbs/bootstrap/projects/18

scss/_navbar.scss
```diff
@@ -25,8 +25,7 @@
 
   // Because flex properties aren't inherited, we need to redeclare these first
   // few properties so that content nested within behave properly.
-  > .container,
-  > .container-fluid {
+  > [class^="container"] {
```
antonstsk: You better select items with possible multiple classes and container
```css
> [class^="container"],
> [class*=" container"] {
```

scss/_navbar.scss
```diff
@@ -140,8 +139,7 @@
 
     &#{$infix} {
       @include media-breakpoint-down($breakpoint) {
-        > .container,
-        > .container-fluid {
+        > [class^=".container"] {
```
antonstsk: Same as above but the dot has to be removed:
```css
> [class^="container"],
> [class*=" container"] {
```

scss/_navbar.scss
```diff
@@ -165,8 +163,7 @@
         }
 
         // For nesting containers, have to redeclare for alignment purposes
-        > .container,
-        > .container-fluid {
+        > [class^=".container"] {
```
antonstsk: 👆👆👆 same...

PR #25637, opened 2018-02-15
New lines to the end of JS dist files for better concatenating compatibility

hdwebpros:
Many tools misbehave if the last line of data in a text file is not terminated with a newline or carriage return / new line combination. They ignore that line as it is terminated with ^Z (eof) instead.

I ran into an issue using NPM to install Bootstrap and cherrypicking a few components to use on a project. When I concatenated with Grunt, I ran into all sorts of issues. Mainly, a comment like `//# sourceMappingURL=util.js.map` would be the last line of the utils.js file for example, and then the function from the first line of the next file in the concatenation process would end up as a comment as well. 

Here's an example of the combined comment and function:
`//# sourceMappingURL=util.js.map;function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }`

Adding a new line to the end of the JS files fixes this issue. 

I added these manually. If there is a process that generates these files, then I'd assume the fix has to go there instead. I'm kind of a novice at this pull request business, but hopefully, I helped.


wolfy1339:
Yeah, those are generated files.

hdwebpros:
Shoot. Sorry. What generates them, or what should I edit to include this fix?

wolfy1339:
I'm not sure how to add it to the build system.
They use npm scripts and the build components directly from cli

Johann-S:
For me it doesn't look like a Bootstrap build issue, but more an issue in your build system (Grunt, concat etc...).

To bundle our files we use RollupJS and to convert our ES6 files to ES5 we use Babel.

Our Rollup config file : https://github.com/twbs/bootstrap/blob/v4-dev/build/rollup.config.js
And how we build our dist files in `js/dist` directory : https://github.com/twbs/bootstrap/blob/v4-dev/package.json#L41

I close this issue because you edited generated files.

hdwebpros:
I'm sorry I edited generated files. This PR is good as closed.

I'd like to push a little bit on the concept of leaving empty lines at the end of your JS files.

First, if you want to force all people to use RollupJS and Babel, sure, just do whatever. In the world of web, frameworks should be flexible and allow users to compile their resources how they see fit. "It's an issue with your system" shouldn't be the cop-out.

Here are 3 good resources on why empty lines are recommended at the end of files (one straight from NPM):
https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file
https://evanhahn.com/newline-necessary-at-the-end-of-javascript-files/
https://www.npmjs.com/package/javascript-style-guide

Johann-S:
I said it's an issue related to your build system that's a bit different 😉

If you find a way to ensure a new line at the end of our build files please open a PR 👍

hdwebpros:
I wish I knew how to do that. I have 4+ hours trying into trying to make RollupJS bundle the src files with Grunt and then use Babel to convert the bundled files, with no luck. I've tried different ES6 concatenators and uglifiers, with no luck.

I tried for a while too to see how I can get a new line added to the end of src the ES5 files.

I'm utterly stuck. Even Babel itself, which is what you use, gets me the "Warning: source/js/site.js: Unexpected token (3850:1) Use --force to continue."

Here are the lines around 3850

```
  return Popover;
}($);
//# sourceMappingURL=popover.js.map;/**
 * --------------------------------------------------------------------------
 * Boo
```

with that sourcemap messing up that comment. 

So, if the tools I am using is what BS is using, and I still have issues, I'm not sure what to do. I could bypass Grunt altogether, but that seems drastic. 

Do you have any other clues on how I can solve this?

Johann-S:
Maybe you can try to work with Webpack which is a bundler like RollupJS with some extra features :thinking:

Issue #25249, opened 2018-01-09
Make use of the `prefers-reduced-motion` media query


MikeRogers0:
Are there any plans to add support for the `prefers-reduced-motion` media query? Here is the [CSS Tricks](https://css-tricks.com/introduction-reduced-motion-media-query/) article about it.

Pretty much, if a user has set in their browser/system preferences they'd like reduce the amount of motion they see, we can know about.

What I'd purpose (if everyone thinks this is a good idea), is adjusting [scss/mixins/_transition.scss](https://github.com/twbs/bootstrap/blob/9a93671b75dda49421cb20e43eaa7281ceac61b0/scss/mixins/_transition.scss) to look more like:

```
@mixin transition($transition...) {
  @if $enable-transitions {
    @if length($transition) == 0 {
      transition: $transition-base;
    } @else {
      transition: $transition;
    }
    @media screen and (prefers-reduced-motion: reduce) {
        transition: none;
    }
  }
}
```

mdo:
What's the browser support? I only see an issue regarding adding it to `caniuse.com`.

alecpl:
It looks to me as Safari-only (?) feature for now. https://bugzilla.mozilla.org/show_bug.cgi?id=1365045

patrickhlauke:
Yes, Safari only. I'd see this as a nice-to-have, future-proof addition though. Perhaps for 4.1?

MikeRogers0:
Awesome! Am I ok to open a PR with the change?

patrickhlauke:
Depends if @mdo and rest of the team feel it's a worthwhile addition, but yes a PR is always nice :)

brianteeman:
@MikeRogers0 definite thumbs up from me - although its a AAA level issue in WCAG2.1 and currently only 100% supported on ios and safari that doesnt mean we shouldnt strive to make everything as accessible as possible. 

I came up with a similar css change for **Joomla** https://github.com/joomla/joomla-cms/issues/19694 which is based on a non bootstrap one for **Drupal** https://www.drupal.org/project/drupal/issues/2940012 

As I am not even remotely a css wizard I am sure whatever you propose will be much better and will get a big thumbs up from me.

patrickhlauke:
@MikeRogers0 if you want to make a PR, i'd be happy to test it. @mdo i'd be in favour of adding this even if Safari-only at the moment, as it's low effort / nice to have / does not adversely impact non-Safari browsers that don't support it.

[...]

PR #25641, opened 2018-02-16
Making use of `prefers-reduced-motion` media query

MikeRogers0:
As discussed in #25249 - if a user (Who is using Safari / iOS) requests
reduced motion in their system settings, we should avoid transitions.


MikeRogers0:
It looks like Travis is failing because stylelint doesn't know about `prefers-reduced-motion`. 

Is there a prefered way to [whitelist](https://github.com/stylelint/stylelint/blob/c85372c07239e2d0688abcda1eb1b390e649bfba/lib/rules/media-feature-name-case/README.md) media features or should I open a PR to `stylelint` ( I think the LOC is https://github.com/stylelint/stylelint/blob/f6fbad37e1ffa84aed4a996b3829b4275e7320c5/lib/reference/keywordSets.js#L553 )

patrickhlauke:
no time to test just now, but i believe editing `build\.stylelintrc` to include the following (or something very similar to it) should work

```
"media-feature-name-no-unknown": [true, {
  "ignoreMediaFeatureNames": ["prefers-reduced-motion"]
}]
```

MikeRogers0:
@patrickhlauke That fixed it! Tests are now passing 🎉

patrickhlauke:
just to check, did you compile this locally and test it actually works as expected?

MikeRogers0:
> just to check, did you compile this locally and test it actually works as expected?

@patrickhlauke I did indeed 👍 If you'd like I can record a little demo video of it working?

brianteeman:
Looks great to me - thanks

mdo:
We should document this, too.

MikeRogers0:
@mdo I'm happy to do that :)

Is after the "Visually hidden content" section in [docs/4.0/getting-started/accessibility.md](https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/getting-started/accessibility.md#visually-hidden-content) a suitable place? 

Copy wise I'm thinking:
```
## Reduced Motion
iOS & MacOS users have a system preference to Reduce Motion which is passed onto the browser. If a user has selected this option, Bootstrap will attempt to reduce the the amount of transitions on a page by defaulting them to `transition: none;`.
```

brianteeman:
@MikeRogers0 iirc the reduce motion preference is not limited to the browser so maybe delete "which is passed onto the browser"

patrickhlauke:
I think documentation-wise I'd lean more towards the following (but yes, the accessibility markdown seems the most appropriate place):

```
### Reduced motion

Bootstrap includes support for the [`prefers-reduced-motion` media feature](https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion). In browsers/environments that allow the user to specify their preference for reduced motion, most CSS transition effects in Bootstrap (for instance, when a modal dialog is opened or closed) will be disabled. Currently, support is limited to Safari on macOS and iOS.
```

@mdo ?

patrickhlauke:
incidentally (testing this PR) it seems `.carousel-fade` and animated progress bars always use transitions/animations, regardless of the setting (and presumably regardless of `$enable-transitions` ?)

MikeRogers0:
@patrickhlauke I'm happy to update [scss/_carousel.scss](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_carousel.scss#L76) and [scss/_progress.scss](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_progress.scss#L33) to remove their animation, though I'm not sure if that would reduce animation to much & make things to clunky. What do you think?

Also updated `docs/4.0/getting-started/accessibility.md` :)

brianteeman:
Updating the progress bar to remove the animation on the candy strip was my initial aim - sadly cant test it right now as am away from my apple

MikeRogers0:
Updated the progress bar & carousel now have reduced animation :)

Here is a demo:
[Screenshot of reduce motion demo](https://www.youtube.com/watch?v=Ksg7WOwHG5A)

patrickhlauke:
ah sorry @MikeRogers0 i didn't intend for you to have to do the carousel and progressbar thing. was more of a passing note.

wondering if those two should (and should have been from the start?) be using the transition mixin instead. that'd then also stop them when `$enable-transitions` is changed. @mdo, thoughts?

patrickhlauke:
@MikeRogers0 would it be possible for you to omit the last commit (and we'll tackle the carousel and progress bar animations separately, probably tying them to the mixin/`$enable-transitions` instead). once that's done, I think we're good to merge :)

MikeRogers0:
@patrickhlauke Reverted that commit 👍

patrickhlauke (approved):

patrickhlauke:
filed a separate issue here https://github.com/twbs/bootstrap/issues/25709

@mdo, good to merge?

MikeRogers0:
@mdo @patrickhlauke Any plans to merge this? :)

patrickhlauke:
@MikeRogers0 went ahead and merged. Think @mdo was happy in principle, but with volume of issues/PRs we get, it's sometimes easy for things to fall between the cracks :)

MikeRogers0:
Ah wonderful! Thank you so much!

brianteeman:
Thank you

PR #25651, opened 2018-02-18
modal input autofocus attr & keyboard nav

### LINKED ISSUE #22402: Feature Request:  Modals:  input autofocus attr, and keyboard focus (has-PR)
Spinoff of #22362

**a)** Bootstrap should automatically exhibit this behavior:
```
$(document).on("shown.bs.modal", function (e) {
    $("[autofocus]:first", e.target).focus();
});
```
(ie, when opening a modal, find the first input with the `autofocus` attribute and focus it 
The `autofocus` attribute is currently ineffective in modals without the user implementing something like the above.
This would be a non-breaking change.

**b)** new modal option/data-attr

`<div class="modal" data-keyboard-bnav="true"></div>`

Common use case:   confirmation dialog.  Enabled, this would allow the user to arrow between "Cancel" and "OK"  (just like a native confirm dialog).
I know I always audibly \*sigh* when I open a non-native confirmation dialog and a) nothing's focused by default, and b) I can't select an option via keyboard/arrow-keys

(Keyboard functionality is a requirement for [Accessibility Compliance](https://www.w3.org/TR/WCAG20/))

---

**See the documentation demo modals**
https://v4-alpha.getbootstrap.com/components/modal/#live-demo

There are 9 "live demos"

4 of the demos are true confirmation dialogs and would benefit from ` data-keyboard-nav="true"` and/or having a default/autofocused action

3 of the demo modals are  email forms which would benefit from having `autofocus` on the message textarea

(The remaining 2 demos have no content)

**Edit:** It appears that autofocus has been suggested before (#12525)  but poo-poo'd out of consideration for mobile devices (don't want to trigger virtual keyboard).    Does focusing a button trigger virtual keyboard?  Potential solution: if touch device and element to focus is of the text/textarea variety: don't autofocus?   My opinion: the accessibility benefits gained by autofocusing are worth the challenge of dealing with the virtual-keyboard exception.

bkdotcom:
### Two new modal accessibility options

**autofocus**:
  * notTouch: (default) gives focus to input/button with autofocus attribute when modal is shown (if not a touch device - to not trigger virtual keyboard)
   * true - always honor autofocus attribute
   * false - don't honor autofocus attribute

**keyboardNav**: (true) | false
    keyboard left/right arrow keys can be used to focus buttons / move between buttons
    (does not take focus from input, select, or textarea)

Check out the updated modal component docs to experience the enhancements

Fixes #22402

### notes:
* had to ignore "A document must not include more than one “autofocus” attribute."  (modal doc contains multiple)...


docs/4.0/components/modal.md
```diff
@@ -67,8 +60,8 @@ Below is a _static_ modal example (meaning its `position` and `display` have bee
         <p>Modal body text goes here.</p>
```
bkdotcom: Static modal example's highlight html didn't match the inline-html (close button should come first)

docs/4.0/components/modal.md
```diff
@@ -67,8 +60,8 @@ Below is a _static_ modal example (meaning its `position` and `display` have bee
...
       </div>
       <div class="modal-footer">
-        <button type="button" class="btn btn-primary">Save changes</button>
```
bkdotcom: Static modal example's highlight html didn't match the inline-html (close button should come first)

XhmikosR:
@bkdotcom: can you fetch upstream and rebase please?

bkdotcom:
> @bkdotcom: can you fetch upstream and rebase please?

done deal

XhmikosR:
@bkdotcom: that's not a rebase.

Johann-S (changes requested):

js/src/modal.js
```diff
@@ -21,17 +21,23 @@ const EVENT_KEY          = `.${DATA_KEY}`
...
 const JQUERY_NO_CONFLICT = $.fn[NAME]
 const ESCAPE_KEYCODE     = 27 // KeyboardEvent.which value for Escape (Esc) key
+const LEFT_KEYCODE       = 37
+const RIGHT_KEYCODE      = 39
 
 const Default = {
+  autofocus: 'notTouch',      // true|false|notTouch
   backdrop : true,
   keyboard : true,
+  keyboardBtnNav: true,  // ability to use arrows to nav button focus
```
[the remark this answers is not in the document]
bkdotcom: renamed to `keyboardNav`..   wanting to keep option to use "arrow-key button nav" separate from <kbd>esc</kbd> key..   perhaps boostrap 5.0 can use less generic `keyboardEsc` (instead of `keyboard`) along with `keyboardNav`

js/src/modal.js
```diff
@@ -280,6 +297,10 @@ class Modal {
     }
   }
 
+  _autofocus() {
+    $(this._element).find(':input[autofocus]:not(:hidden)').eq(0).trigger('focus')
```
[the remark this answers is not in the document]
bkdotcom: changed to 
```
    const selector = 'input[autofocus], select[autofocus], textarea[autofocus], button[autofocus]'
    $(this._element.querySelectorAll(selector)).filter(':not(:hidden)').eq(0).trigger('focus')
```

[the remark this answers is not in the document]
bkdotcom: ... and moved `selector` to `Selectors.AUTOFOCUSABLE`

js/src/modal.js
```diff
@@ -325,6 +360,29 @@ class Modal {
     })
   }
 
+  _keyboardBtnNav(prevNext) {
+    const $focusable = $(this._element).find('.btn')
```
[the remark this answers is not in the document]
bkdotcom: done

js/src/modal.js
```diff
@@ -234,6 +245,11 @@ class Modal {
     return config
   }
 
+  // Util worthy?
+  _isTouchDevice() {
+    return 'ontouchstart' in window || window.DocumentTouch && document instanceof window.DocumentTouch
```
Johann-S: Can you use the same way we check touch device in our carousel ? 
https://github.com/twbs/bootstrap/blob/v4-dev/js/src/carousel.js#L119

Johann-S (changes requested):

js/src/modal.js
bkdotcom: done 👍

bkdotcom:
@Johann-S I'm following up to confirm that all requested changes have been addressed / not waiting on me.

js/src/modal.js
```diff
@@ -83,6 +92,7 @@ class Modal {
     this._ignoreBackdropClick = false
     this._isTransitioning     = false
     this._scrollbarWidth      = 0
+    this._touchSupported      = 'ontouchstart' in document.documentElement || navigator.maxTouchPoints > 0
```
XhmikosR: @Johann-S: maybe we should have a common function in util about this?

bkdotcom: @XhmikosR & @Johann-S 
I can add a util method if you like...
```
    this._touchSupported      = Util.isTouchSupported()
```

or would a "constant" (like `Util.TRANSITION_END` be preferred ?
```
    this._touchSupported      = Util.TOUCH_SUPPORTED
```

if going with the "constant" route, would we even need the local `this._touchSupported` assignment and instead simply check `Util.TOUCH_SUPPORTED` vs `this._touchSupported`?

Johann-S: A constant should be enough 👍

bkdotcom: done..  dropdown.js & tooltip.js also do a touch test... now also using Util.isTouchEnabled()
I was getting unit test errors when going the full constant route..  I don't know why...  but seemed to be carousel unit test related

XhmikosR:
@bkdotcom please don't change the bundle size yourself, thanks.

mdo:
Closing as stale and needing to revisit for now as pure JS for v5. See #28481.

bkdotcom:
If I were to want to contribute/migrate this to v5, which branch would I create a pull request on?

MartijnCuppens:
@bkdotcom, the `master` branch

PR #25662, opened 2018-02-19
Variable transition durations

### LINKED ISSUE #25448: Make collapse transition duration dynamic
Collapse.js transition duration is taken from a constant of value of 600.

However in some cases the transition needs to be altered with CSS, for example:

```
.collapsing {
    transition: none;
}
```

This prevents the collapsing trigger to work properly as it always waits for the 600ms timeout.

The collapse.js transition duration should take its value dynamically from css.

Demo: https://codepen.io/artifex404/pen/LewydR
(Note: auto updating is disabled, rerun codepen to see changes)

MartijnCuppens:
Based on #25218. All constant `TRANSITION_DURATION`s are replaced by the `transition-duration` of the element in question. This allows transition duration manipulation with css.

Closes #25448.


js/src/util.js
```diff
@@ -104,6 +105,24 @@ const Util = (($) => {
       }
     },
 
+    getTransitionDurationFromElement(element) {
```
Johann-S: Please add a unit test for this method

Johann-S (changes requested):

js/src/util.js
MartijnCuppens: See https://github.com/twbs/bootstrap/pull/25218#pullrequestreview-87160376

Johann-S: You can test when you have multi durations and when you have `ms`

MartijnCuppens: Unit tests are added: https://github.com/twbs/bootstrap/pull/25662/commits/05cef0a4ec9122ee7eb1b6e1d940809884881df2

Johann-S (approved):

XhmikosR:
@Johann-S: how do you want to proceed with this?

Johann-S:
it works perfectly I tested it locally 👍

js/tests/unit/util.js
```diff
@@ -42,6 +42,50 @@ $(function () {
     assert.strictEqual(typeof Util.isElement({}) === 'undefined', true)
   })
 
+  QUnit.test('Util.getTransitionDurationFromElement should accept transition durations in milliseconds', function (assert) {
+    assert.expect(1)
+    var $div = $('<div style="transition: all 300ms ease-out;"></div>').appendTo($('#qunit-fixture'))
```
Johann-S: This test do not work, I don't know how but your `300ms` is converted into seconds in your JS, see our coverage : 

image

Johann-S (changes requested):

js/tests/unit/util.js
MartijnCuppens: What did you do to detect this? `npm test` doesn't return any errors and the visual tests look fine?

XhmikosR: It's the coverage, see https://github.com/twbs/bootstrap/pull/25783

artifex404: My 2 cents are, that the condition it should test should be: `transitionDuration.indexOf('ms') != -1`

XhmikosR: Indeed, `indexOf` returns -1 if not found. So `!== -1` would be better.

MartijnCuppens: Changed it to `transitionDuration.indexOf('ms') !== -1`, problem with coverage still exists.

Johann-S: As I said on my first message, this line : 
```js
let transitionDuration = $(element).css('transition-duration')
```

auto convert `transition-duration` into seconds.

IMO jQuery use `getComputedStyle` which do that

MartijnCuppens: jQuery takes care of this indeed (Also tested this in Chrome, FF, IE).

IMO it's safe to remove this check as long as we depend on jQuery.

Johann-S: I don't think it's jQuery which do that, IMO it's more `getComputedStyle` but that's OK to me 👍

Johann-S (approved):

jrochkind:
Trying to figure out how to use this.. it isn't documented great. It seems to be minimally documented on carousel, but isn't documented at all on all the components where it is also in effect. 

Not sure what the situation in in bootstrap 5 prereleases either. Very confused about supported ways of changing transition durations on a per-componet or even per-instance basis.

PR #25677, opened 2018-02-21
Add color value to table-row-variant mixin

### LINKED ISSUE #25675: Table row-variants does not respect font-color

tomsommer:
Fixes #25675


MartijnCuppens (changes requested):
The `.table-active` doesn't look ok. I don't know if we can easily fix this because a semi transparent color is used. 
https://deploy-preview-25677--twbs-bootstrap.netlify.com/docs/4.2/content/tables/#contextual-classes

tomsommer:
Could one not argue that the fault in `.table-active` lies i `color-yiq()`?

MartijnCuppens:
We can't merge this because it introduces unexpected results for `.table-active` and this will break existing implementations.

PR #25678, opened 2018-02-21
Update list-group-flush top/bottom item border setting

kvlsrg:
Change `*-child` to `*-of-type` selector.

This fixes an issue with appearing of top/bottom border of first/last item if flush list group is inside the content.

list

PR #25688, opened 2018-02-22
Fix one-by-one flush lists items borders

kvlsrg:
Here is a [pen](https://codepen.io/kvlsrg/pen/bLKzQw) with a problem.

Also this PR fixes issue with first/last item top/bottom border if list used with other content (closes #25678)


mdo:
We're going to need to see every variant of the list group before we can tweak these all again :).

mdo:
Haven't seen any more movement here, so closing this out as stale.

PR #25690, opened 2018-02-22
Clarify the necessary `data-target`/`href` attribute for carousel controls

balbuf:
I couldn't figure out why my controls stopped working; I didn't initially realize the connection between the `href` value and the `id` of the carousel (in the controls example). The other method of using `data-target` for non-link elements (in the indicator example) is a little more clear, but I also didn't notice that at first. Perhaps additional clarification that `data-target` takes precedence over `href` would be wise?


XhmikosR:
@Johann-S: seems good to you?

Johann-S (approved):
LGTM 👍

XhmikosR (approved):

PR #25708, opened 2018-02-25

XhmikosR:
**We need to verify that the prefixes removed is something we want, otherwise we need to tweak our browserslist config.**


Johann-S (approved):

XhmikosR:
Ping @mdo about the prefixes removal.

XhmikosR:
Can someone verify the removed prefixes? Not sure if I missed something, but the webkit prefixes are mostly removed.

XhmikosR:
Nevermind, I checked and it seems it's normal. If any issue arise please ping me.

PR #25724, opened 2018-02-26
Fix nested calc

### LINKED ISSUE #25717: Height property of ::after pseudo-element of custom-file-label class is an error in IE11.
`height` property of `::after` pseudo-element of `custom-file-label` class is an error.  
Because "calc" is nested.
https://getbootstrap.com/docs/4.0/components/forms/#file-browser  
```
<div class="custom-file">
  <input type="file" class="custom-file-input" id="customFileLang" lang="es">
  <label class="custom-file-label" for="customFileLang">Seleccionar Archivo</label>
</div>
```
image

In IE 11, "calc" can not be described in calc()...
`calc((2.25rem + 2px) - 1px * 2)`

Operating system and version: Windows8.1 / Internet Explorer 11

MartijnCuppens:
Fixes #25717


XhmikosR:
@mdo: this should be in the next release I think.

MartijnCuppens:
#26328 also fixes this and doesn't require an extra variable.

PR #25730, opened 2018-02-27
Added overflow hidden to resolve border radius conflicts

emaildano:
# Issue

Modal by default does not include `overflow:hidden`. This can cause a conflict with the border radius on `modal-body` if any part of the modal background color is changed.

# Example

When `.modal-content` background is white, changing any other part of the modals background color such as Header or Footer it disabled the border radius, visually.

# Fix

Adding `overflow:hidden` to `.modal-content` does not create any conflicts with existing code but it does resolve instances when the modal header, footer, and body are not the same color.

# Screenshots

screenshot 2018-02-27 10 29 18
screenshot 2018-02-27 10 28 59
screenshot 2018-02-27 10 28 42

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -67,6 +67,7 @@
   background-color: $modal-content-bg;
   background-clip: padding-box;
   border: $modal-content-border-width solid $modal-content-border-color;
+  overflow: hidden;
   @include border-radius($border-radius-lg);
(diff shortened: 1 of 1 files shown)


mdo:
This would crop any tooltips, popovers, or dropdowns within the modal footer. Thanks, but we'll pass on this suggested solution.

PR #25737, opened 2018-02-28
Shadows Classes

ayoubkhan558-zz:
I added some shadows classes which are missing.


caiquecastro:
Shouldn't you add them on the raw scss files instead of on the compiled css?

Johann-S:
Please do not edit our generated dist files

mdo:
These are random classes and values, thanks but we'll pass. I have something else in mind.

PR #25738, opened 2018-02-28
Improve order utilities doc

weberlance:
Mention about .order-first and .order-last. Replace long list of order-* 
classes with the pattern.

### DIFF
--- a/docs/4.0/utilities/flex.md
+++ b/docs/4.0/utilities/flex.md
@@ -356,20 +356,50 @@ Responsive variations also exist for `flex-wrap`.
 
 ## Order
 
-Change the _visual_ order of specific flex items with a handful of `order` utilities. We only provide options for making an item first or last, as well as a reset to use the DOM order. As `order` takes any integer value (e.g., `5`), add custom CSS for any additional values needed.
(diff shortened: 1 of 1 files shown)


mdo:
You're mixing flex utilities with grid modifiers. They can mention one another, but they shouldn't conflate one another.

weberlance:
[It doesn't look like different entities](https://github.com/twbs/bootstrap/blob/6431d746bccab270812a841be44f5e7829f7f367/scss/mixins/_grid-framework.scss#L49-L55). In Alpha there were "flex-first" class but now only "order-..." classes remain. So, I think, it's good idea to separate description of order classes but mention of order-first, -last is better to place here too([like here](https://github.com/twbs/bootstrap/blob/v4-dev/docs/4.0/layout/grid.md#reordering)).

mdo:
What I mean is those numbered `order` classes are specifically built for the grid system (hence their range of `-1` through `13`). Note where you're finding those classes—`_grid-framework.scss` vs `utilities/_flex.scss`. That's the mix I'm referring to, not their usage or relation :).

Lemme pull this down and review. Might end up punting to 4.2 given limited timing on my part.

Thanks!

mdo:
Forgot about this PR. Coming back to it, I'm still sticking to my previous stance—this conflates grid classes with the utilities based on the flex properties. We can refer folks over to the grid docs, but that's all I'd suggest here.

Issue #25417, opened 2018-01-24
Modal shifts navbar-toggler on fixed navbar


ysds:
![modal](https://user-images.githubusercontent.com/4065765/35313650-e2c4113a-0104-11e8-832f-9f716a2d0b69.gif)

Demo: https://codepen.io/anon/pen/WdBvpB

I've seen a similar issue before, but I could not find it...

Johann-S:
Closed thanks to: #25745

PR #25745, opened 2018-03-01
fix(js): removes .modal .navbar-toggler margin

zalog:
The problem is that `.navbar-toggler` flips left when we open a bs modal dialog (gif or [pen](https://codepen.io/zalog/pen/LQqdJe) below).

I don't think we need to adjust `.navbar-toggler` in this case.
We'll always have a parent that manage to adjust the missing scroll width:

- in case of a `.fixed-*` we have padding-right on that
- in case of a no fixed navbar, we have padding-right on body

Here you have a [pen](https://codepen.io/zalog/pen/LQqdJe) and a demo:

demo

What do you think?

### DIFF
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -431,7 +431,6 @@ const Modal = (($) => {
         //   while $(DOMNode).css('padding-right') returns the calculated value or 0 if not set
         const fixedContent = [].slice.call(document.querySelectorAll(Selector.FIXED_CONTENT))
         const stickyContent = [].slice.call(document.querySelectorAll(Selector.STICKY_CONTENT))
-        const navbarToggler = [].slice.call(document.querySelectorAll(Selector.NAVBAR_TOGGLER))
 
         // Adjust fixed content padding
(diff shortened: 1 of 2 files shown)


ysds:
@zalog Thanks! (x-ref: #25417)

Johann-S:
Please fix unit tests and it would be rad 👍

zalog:
Yep, done. We're on green now.

Johann-S:
Deleted our unit test, isn't a good solution, unless this case is already covered by a present unit test

zalog:
I deleted the test about `$('<div class="navbar-toggler"></div>')` because we don't need to test for margin-right. Now, after the first commit from this pr, we are not manipulating `.navbar-toggler` from js anymore.

Hmm... or is something that I'm missing? Sorry

Johann-S:
But you made some changes, so we should test that

twbs-closer:
Hey there!

Thanks!

Johann-S (approved):

PR #25747, opened 2018-03-01
Fixes validation tooltip position on inline forms

### LINKED ISSUE #25694: Validation Tooltip inside horizontal form not just below input when label multiline
## Problem Description ##
When using validation tooltip whith horizontals forms, the tooltip position is not just below the input when label is display on multiple lines : 

image

## Expected result ##

image

## Reproduce ##

Source to reproduce the problem : [CodePen Source](https://codepen.io/Outlivier/pen/oEMPYY) (need a resolution at minimum of 1280*1024 to display the form horizontally)

## Question ##

I'm not sure if it's a bug or a design decision, because it's due to `top:100%` on invalid-tooltip class.

If it's a design decision, can i safely remove top 100% whithout side effects ?

andresgalante:
This PR fixes #25694 

Change the position of the validation tool tip by scoping the `top` property only to `input-group`.

@mdo can you please review this one? I am unsure if I am missing any corner case.


mdo:
This PR solves this one situation, but doesn't solve the underlying issue that we likely need to be way more specific in the structure of the forms to enable proper positioning. For example, both this and the current implementation would break visually if there's help text below the field.

For now, closing this and slating for a v5 revisit with the removal of `position: relative` on the columns.

PR #25755, opened 2018-03-01
fix(table): Add border color relative to theme for accessibility

### LINKED ISSUE #25685: Contrast issue: Border between two table-{success,danger,…} rows is gray
Currently, when several rows of the same `table-<context>` class follow each other, they are separated by a gray border which looks out of place and in some cases also have contrast issues:

screen shot 2018-02-22 at 4 21 23 pm

(Note: in this screenshot the leftmost column has `rowspan="3"`)

I propose a rule similar to the following to all classes that affect table row backgrounds:

```css
tr.table-success ~ tr.table-success td, tr.table-success ~ tr.table-success th {
    border-top-color: hsla(133, 42%, 74%, 1);
}
```

It would look like this instead:

screen shot 2018-02-22 at 4 15 36 pm

bastienmoulia:
Related to #25685

Add a border color on table with theme classes to better accessibility contrast.

Not sure about the `-6` in `theme-color-level($color, -6)` for the border color, but it seems good to me.

capture d ecran 2018-03-01 a 22 58 32

Fixes #25685


zwagoth:
As somebody who is currently fighting to make any sort of dark theme work properly, you should make the levels a variable or you force people to override them later on. This was fixed for alerts, but not for tables.

bastienmoulia:
@zwagoth: Done

zwagoth:
Appreciate it.

andresgalante (approved):
@bastienmoulia Good move, thanks!
This is following the same structure as alerts, I think we should merge it.

XhmikosR:
@mdo: do you have any objections against this this?

XhmikosR:
Rebased @mdo

scss/mixins/_table-row.scss
```diff
@@ -9,6 +9,15 @@
     > td {
       background-color: $background;
     }
+
+    @if $border != null {
```
mdo: Why wrap this around a `null` conditional?

XhmikosR: Probably because `$border` is initialized to `null` above. And this way it's an optional parameter.

PR #25756, opened 2018-03-02
Update package.js

DcyLsy:


mdo:
Don't test Git or GitHub on public repositories.

Issue #23775, opened 2017-08-31
Create component shadow variables to keep consistency across components


andresgalante:
We are defining different shadows in different components. There are about 25 shadow definitions on variables.

To keep consistency, I think we unify all of them in 4 shadow variables (sm, md, lg and inset) and then reuse them across components.

For example, instead of:
```scss
$popover-box-shadow:                  0 .3rem .6rem rgba($black,.2) 
```

We would have something like:
```scss
$component-box-shadow-md:                  0 .3rem .6rem rgba($black,.2) 
$popover-box-shadow:                  $component-box-shadow-md 
```
This follows the same idea as #23674 

Before I send a PR I'd like to know what do you guys think?

mdo:
I'm down!

Herst:
Accomplished with #24429?

mdo:
Revisiting this, I've put some existing `box-shadow` variables that we added awhile ago better use. See #29023.

PR #25758, opened 2018-03-02
Add box-shadow utilities

### LINKED ISSUE #23437: Add shadow into utilities
It would nice to quickly add shadows to components, please add this feature into utilities

mdo:
First pass at adding `box-shadow` utilities with `.shadow-*` classes. Right now these are pulled from other components and modified slightly. ~~They're also conditionally included based on the `$enable-shadows` boolean variable. Still need to finish docs and update existing shadow values using these new variables.~~

**Update:** I've removed the `$enable-shadows` conditional, meaning these variables are always available now. I've also added a `.shadow-none` option.

Fixes #23437.


andresgalante (approved):
I'd also be great if we tie it to #23775 probably on another PR

mdo:
@mixinix No, but that's a good idea! Can you open a new issue for that?

Issue #25748, opened 2018-03-01
Code coverage


Johann-S:
We have a lot's of unit test, and it's difficult to see which case our units tests covered or not.

We should add a something to see our code coverage

/CC @XhmikosR, @bardiharborow

XhmikosR:
@Johann-S: which solution do you have in your mind? Using nyc?

Generally speaking it's a good idea, so feel free to CC me if you have a working branch.

Johann-S:
Something like this : https://github.com/marketplace/coveralls

PR #25760, opened 2018-03-02
add code coverage

Johann-S:
We this PR we will be sure our JS is covered at 80% but we are free to grow this percentage 😉 

BTW it should throw an error if we move under 80% of covered code

Fixes : #25748 

/CC @XhmikosR, @bardiharborow


package.json
```diff
@@ -39,6 +39,7 @@
     "js-compile-standalone": "rollup --environment BUNDLE:false --config build/rollup.config.js --sourcemap",
     "js-compile-bundle": "rollup --environment BUNDLE:true --config build/rollup.config.js --sourcemap",
     "js-compile-plugins": "cross-env PLUGINS=true babel js/src/ --out-dir js/dist/ --source-maps",
+    "js-compile-plugins-coverage": "cross-env PLUGINS=true NODE_ENV=test babel js/src/ --out-dir js/dist/coverage/ --source-maps",
```
Johann-S: maybe I should change the output dir to : `js/coverage/dist` which is not distributed through npm 🤔

Johann-S:
it's give us something like this in `js/coverage/lcov-report/index.html`

image

XhmikosR:
Shouldn't we have a service like Travis so that we can check quickly things? I'm pretty sure there are various services we could integrate.

Johann-S:
Travis execute it for us see : https://travis-ci.org/twbs/bootstrap/jobs/348736259#L704

XhmikosR:
OK, I'm just used to coverage as a repo hook. But this works too.

Johann-S:
I thinks it's a good first step and it's pretty convenient to be able to run the coverage locally , but I'm not against a hook in an other PR

Johann-S:
Would you like that PR @XhmikosR ? Or do you have any concerns ? If not maybe can you approve my PR 😄

XhmikosR (approved):

XhmikosR:
@Johann-S; just make sure you resolve the conflicts right :)

Issue #17118, opened 2015-08-20
v4 - Touch gesture support for carousel


nodkz:
This is must be introduced in v3. 
If bootstrap goes adaptive and mobile friendly - carousel must support swipe on touch devices!
http://v4-alpha.getbootstrap.com/components/carousel/#example

kkirsche:
I'm not sure I know what you mean by "this is must be introduced in v3" — v3 is essentially feature frozen so new features most likely won't be added to it as we are preparing for only critical bug fixes to it.

surjithctly:
He meant, This feature should have been introduced in v3, But its already late :)

Now, The requested feature should add in v4 not v3 :P

rustd:
+1 for this. This will be useful now that bootstrap-touch-carousel is not being maintained https://github.com/ixisio/bootstrap-touch-carousel

fat:
yep agreed, it's not ~that~ hard to do… maybe for thanksgiving or xmas i'll write it if ya'll haven't already

patrickhlauke:
would this include both touch events and pointer events (the latter for IE/Edge on Windows touchscreen laptops, where touch events are not enabled by default)? if so, it can start to get quite dense without any third-party library like hammer.js that papers over the differences in event models

nodkz:
Nice vanilla slider https://github.com/gilbitron/Ideal-Image-Slider

I already migrated to React. But may be it would be helpful - reactive sliders for React are:
http://kenwheeler.github.io/nuka-carousel/
http://oliviertassinari.github.io/react-swipeable-views/

GianlucaMaggio:
Did you plan it in next alpha? Or when will be available?

Johann-S:
No date for this feature but it's planned for V4 (see : https://github.com/twbs/bootstrap/pull/17021) and we are open for PRs 😄

zberno:
There is a lot of options, events and who know what strange feature needs for a project working on a touch/drag carousel.
Maybe it's not the right method but.. I'm here to suggest something like this:
http://idangero.us/swiper/

marchy:
Any status update on this?
Would be great to know rough timeline (ie: even which quarter/season) as to plan accordingly.

Cheers!

endcore:
+1

kaungmyatlwin:
+1

Nodws:
years later, we still need to hack the events in

patrickhlauke:
PRs welcome...

[...]

PR #25776, opened 2018-03-03
Add touch support in our carousel

Johann-S:
- [x] Add unit test
- [x] Documentation

Fixes: #17118


XhmikosR:
Its quite a big addition... Can't we do the basic stuff ourselves?

_includes/scripts.html
```diff
@@ -2,6 +2,8 @@
 <script>window.jQuery || document.write('<script src="{{ site.baseurl }}/assets/js/vendor/jquery-slim.min.js"><\/script>')</script>
 
 <script src="{{ site.baseurl }}/assets/js/vendor/popper.min.js"{% if site.github %} integrity="{{ site.cdn.popper_hash }}" crossorigin="anonymous"{% endif %}></script>
+<script src="{{ site.baseurl }}/assets/js/vendor/hammer.min.js"{% if site.github %} integrity="{{ site.cdn.hammer_hash }}" crossorigin="anonymous"{% endif %}></script>
```
XhmikosR: You need to add it in package.json too for the docs package.

Johann-S:
I don't think it's a big change here, we add an optionnal lib, if folks need a touch support that's all

XhmikosR:
Sure, but can't we have the basic functionality without a 20 KB file? I recall you had a branch in the past.

Johann-S:
This won't be added in our dist files, so we won't add 20 kb

I'm not in favor of adding logics about touch detection and swipe detection when they are libs which can do that better than me

XhmikosR:
I agree on that, but then one would say why don't you support jQuery mobile or whatever other lib there's out there.

Anyway, I'm not against this, I just want us to think it through.

Johann-S:
I think HammerJS is a good choice, Angular chose HammerJS too

_includes/scripts.html
Johann-S: It's strange but Popper isn't in our `doc-minify` task 🤔

XhmikosR: True, indeed. Just leave it as is and I'll give it a go in the next days and let you know. :)

js/src/carousel.js
```diff
@@ -244,16 +261,22 @@ const Carousel = (($) => {
...
 
     _addEventListeners() {
+      const touchSupported = 'ontouchstart' in document.documentElement
       if (this._config.keyboard) {
         $(this._element)
           .on(Event.KEYDOWN, (event) => this._keydown(event))
       }
 
+      if (touchSupported) {
+        this.hammer.on(Event.SWIPELEFT, () => this.next())
```
jackmu95: Isn't here a check quired if hammer support is enabled or the hammer reference exists to prevent a reference error? 🤔

Johann-S: Yep that's true nice catch 👍

js/tests/index.html
```diff
@@ -20,11 +20,15 @@
       }())
     </script>
     <script src="../../assets/js/vendor/popper.min.js"></script>
+    <script src="../../assets/js/vendor/hammer.min.js"></script>
```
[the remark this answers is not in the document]
Johann-S: yep sure !

js/tests/karma.conf.js
```diff
@@ -17,6 +17,8 @@ module.exports = (config) => {
     files: [
       jqueryFile,
       'assets/js/vendor/popper.min.js',
+      'assets/js/vendor/hammer.min.js',
```
XhmikosR: This shouldn't be needed.

XhmikosR:
@Johann-S can you rebase this?

Johann-S:
Done @XhmikosR 👍

Johann-S:
I cannot validate our test because on SauceLabs we run our test on Linux Firefox 45, Linux Chrome 48 and Mac OS X 10.11 Safari 10 which are obsolete version

XhmikosR:
@Johann-S: so we can't merge this now if we can't test it, right? What can we do about this?

XhmikosR:
@mdo: can this be done in 4.2?

Johann-S:
@XhmikosR everything is done here and works, thanks to BrowserStack 😄 

/CC @bardiharborow for a review

js/tests/unit/carousel.js
```diff
@@ -1,6 +1,8 @@
 $(function () {
   'use strict'
 
+  var touchSupported = 'ontouchstart' in document.documentElement
+
```
patrickhlauke: for testing purposes, this is of course fine. i'll just note here that to detect pointer events touch support you'd want to also check for `navigator.maxTouchPoints > 0` (in production on a live site, not for automated unit tests)

site/docs/4.1/components/carousel.md
```diff
@@ -12,6 +12,8 @@ The carousel is a slideshow for cycling through a series of content, built with
 
 In browsers where the [Page Visibility API](https://www.w3.org/TR/page-visibility/) is supported, the carousel will avoid sliding when the webpage is not visible to the user (such as when the browser tab is inactive, the browser window is minimized, etc.).
 
+The carousel can handle swipe events (left and right) only if you include [HammerJS]({{ site.cdn.hammer }} before Bootstrap.
```
patrickhlauke: perhaps reword this a tad:

"The carousel supports swipe gestures (left and right) using <a href="https://hammerjs.github.io/">HammerJS</a>. For this to function correctly you need to include [HammerJS]...."

js/tests/unit/carousel.js
Johann-S: yep, I added that line because on Safari our unit tests with touch events failed 😄

site/docs/4.1/components/carousel.md
Johann-S: I changed that, thanks Patrick !

XhmikosR:
@Johann-S: can you rebase this one please? :)

Johann-S:
Done @XhmikosR 😉

XhmikosR:
All good with this @Johann-S. Only thing I was thinking is if we should have an option to enable/disable touch support.

Johann-S:
If they don't include HammerJS there is no touch support

XhmikosR:
Yes, but the bundle already includes it.

js/src/carousel.js
```diff
@@ -96,11 +101,23 @@ const Carousel = (($) => {
       this._isSliding          = false
 
       this.touchTimeout        = null
+      this.hammer              = null
 
       this._config             = this._getConfig(config)
       this._element            = $(element)[0]
       this._indicatorsElement  = this._element.querySelector(Selector.INDICATORS)
 
+      if (HAMMER_ENABLED) {
```
Johann-S: I should use hammer only if touch events supported

ben29:
So when it will be supported?

XhmikosR:
@Johann-S: would you be so kind and rebase this one last time (hopefully)?

Johann-S:
Done @XhmikosR 😉 Maybe someday it'll be merged 🙏

XhmikosR (approved):

XhmikosR:
OK, I think this is ready to be merged :)

site/docs/4.1/components/carousel.md
```diff
@@ -281,6 +283,12 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>true</td>
       <td>Whether the carousel should cycle continuously or have hard stops.</td>
     </tr>
+    <tr>
+      <td>touch</td>
+      <td>boolean</td>
+      <td>true</td>
```
XhmikosR: One thing. This is true, but maybe we should mention that it will be disabled gracefully if HammerJs isn't present?

Johann-S: You're right we should add that

patrickhlauke (dismissed):
the documentation part may need some more clarity, as it's not quite obvious to me what that even means...

site/docs/4.1/components/carousel.md
```diff
@@ -287,7 +285,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>touch</td>
       <td>boolean</td>
       <td>true</td>
-      <td>Whether the carousel should handle touch event and allow swipe left/right.</td>
+      <td>Whether the carousel should handle touch events and allow swipe left/right (only for mobile browsers).</td>
```
patrickhlauke: no, not really "only mobile browsers" - there are desktop computers with touch (e.g. surface). also, hammer uses pointer events where available, so technically not "touch events". suggest changing it to what i said originally in my follow-up comment :)

patrickhlauke (changes requested):

patrickhlauke:
do any of the demo carousels on https://deploy-preview-25776--twbs-bootstrap4.netlify.com/docs/4.1/components/carousel/ have touch enabled? if not, would be good to do it and note it

Johann-S:
All the carousels have touch enabled 👍

site/docs/4.1/components/carousel.md
Johann-S: Oh ok I'll change that. We removed HammerJS we use a Vanilla JS solution

patrickhlauke: oh, missed that you're using vanilla JS (as i saw the hammer simulator being loaded as well). in which case could I'd say it would be good to also support pointer events...but that's probably a more involved addition...

js/src/carousel.js
```diff
@@ -84,21 +89,21 @@ const Selector = {
...
+    this._isPaused      = false
+    this._isSliding     = false
+    this.touchTimeout   = null
+    this.touchStartX    = 0
+    this.touchDeltaX    = 0
+
+    this._config            = this._getConfig(config)
+    this._element           = element
+    this._indicatorsElement = this._element.querySelector(Selector.INDICATORS)
+    this._touchSupported    = 'ontouchstart' in document.documentElement || navigator.maxTouchPoints > 0
```
patrickhlauke: if you're not using pointer events, why are you testing for `navigator.maxTouchPoints` (which is part of the pointer events spec) ?

site/docs/4.1/components/carousel.md
patrickhlauke: for reference: supporting pointer events will also make this work on touch-enabled desktop/laptop devices in Edge (which, by default, does not expose touch events by default, but does support touch interactions via pointer events)

patrickhlauke:
swiping left/right in Chrome on desktop/laptop with touchscreen and on android also triggers back/forward navigation. it likely needs a preventDefault() in there somewhere (this is the sort of stuff that i think Hammer would have handled by default already, as well as pointer events support)

XhmikosR:
>the swipe direction seems counterintuitive. if i swipe right, i would want the carousel to go to the previous slide, not the next slide, and vice-versa

Yeah, that was one of the things that didn't feel right to me either. I asked Johann to keep the HammerJS patch just in case we need to revert. So if things tend to be too broken, we can still just go with Hammer.

patrickhlauke (approved):

patrickhlauke (changes requested):
- revert the swipe direction (so swiping left goes to next slide, not previous, and vice-versa)
- prevent swipes from triggering browser back/forward navigation (e.g. Chrome on Android / on Surface)

Johann-S:
it should be better @patrickhlauke can you confirm it ? Please

patrickhlauke (approved):

patrickhlauke:
strangely, it still doesn't work in Edge on Surface (while the hammer.js demo itself http://hammerjs.github.io/ does). i'll have a look later tonight to see if i can work out why. other than than, the swipe direction now feels natural, and back/forward in Chrome is suppressed when swiping. :+1:

Johann-S:
I tested (thanks to BrowserStack) on a Lumia 930 with Windows Phone 8.1 and it works fine 🤔

patrickhlauke:
Win Phone 8 behaves differently from a desktop Windows 10 on a touchscreen.

patrickhlauke:
ah, i see you had added hammer, but then removed it again, hence my confusion. to be clear: Edge/Win 10 on a touch-enabled desktop/laptop does not have touch events enabled, but does allow touch interaction via pointer events (filtered to `type = "touch"`, and with addition of relevant `touch-action: pan-x` to the element/container with the event handlers). as you reverted to only touch events again after removing hammer, it's not surprising it's not working in Edge on Surface. i'd say although it's an edge case (hah) at this point, since pointer events are the more modern way of handling things (and finally even Apple have them "in development" for future Safari), it's worth future-proofing this code here as well. i can give it a shot if you don't mind me fuzzing around with your branch?

patrickhlauke:
should ideally use both, with preference for PE where available. i'll have a go either later tonight or tomorrow

Johann-S:
I implemented a solution with PointerEvent (finally I found the time 😄) but I'm not sure if I'm fully compatible 😟 
If you can help me on that @patrickhlauke or improve my code do not hesitate 👍

js/src/carousel.js
```diff
@@ -230,23 +261,74 @@ class Carousel {
       $(this._element)
         .on(Event.MOUSEENTER, (event) => this.pause(event))
         .on(Event.MOUSELEAVE, (event) => this.cycle(event))
-      if ('ontouchstart' in document.documentElement) {
+    }
+
+    this._addTouchEventListeners()
```
patrickhlauke: should this not only be added if the configuration for touch is true?

js/src/carousel.js
```diff
@@ -230,23 +261,74 @@ class Carousel {
...
+  }
+
+  _addTouchEventListeners() {
```
patrickhlauke: beyond adding the event listener, this code should also add `touch-action: pan-x` to the carousel when pointer events are used ... `preventDefault()` doesn't stop the forward/back browser navigation for pointer events, it's done declaratively with `touch-action` in the PE model. either doing it dynamically (adding a `style`? seems a bit dirty) or adding it to the definition of carousel CSS styles (though then it would "swallow" horizontal swipes in a PE browser even if the author DIDN'T want touch to be handled on a carousel)

js/src/carousel.js
```diff
@@ -230,23 +261,74 @@ class Carousel {
...
+    if (!this._touchSupported) {
+      return
+    }
+
+    const start = (event) => {
+      event.preventDefault()
+
+      if (this._pointerEvent) {
+        this.touchStartX = event.originalEvent.clientX
```
patrickhlauke: ideally also check first that the pointer event is of `pointerType=="touch"` - and possibly even `pointerType="pen"` if we want to allow pen/stylus swiping - as otherwise the swipe will also happen for mouse users. see https://w3c.github.io/pointerevents/#dom-pointerevent-pointertype

js/tests/unit/carousel.js
```diff
@@ -989,4 +1018,189 @@ $(function () {
       }, 80)
     }, 80)
   })
+
+  QUnit.test('should allow swiperight and call prev with pointer events', function (assert) {
```
patrickhlauke: does this only simulate/test touch events? or does it also test if the code works correctly for pointer events?

js/src/carousel.js
Johann-S: I added a class when we use PointerEvents but I don't understand why it doesn't work on IE 10/11, do you any ideas why @patrickhlauke ? 🤔

patrickhlauke:
urgh, i'm an idiot . the touch action that's needed isn't `touch-action: pan-x` ... i seem to always forget that `touch-action` really means "which actions should the *browser* handle", and horizontal panning is exactly what we *don't* want the browser to handle...

long story short, changing to `touch-action: pan-y pinch-zoom` makes this work in Edge/Surface and IE11/Surface as well

js/src/carousel.js
patrickhlauke: i'm an idiot...and suggested the wrong `touch-action` value...i should go and have a word with the co-editor and chair of the PE spec, it seems...

XhmikosR:
So in Firefox 62 dev tools emulation now it throws this:

```
TypeError: originEvent.touches is undefined carousel.js:287:8
start
carousel.js:287:8
_addTouchEventListeners/<
carousel.js:329:56
```

Johann-S:
It seems not @patrickhlauke our unit tests do not pass on IE 10/11 😟

Johann-S:
@patrickhlauke I cleaned a bit the git history and squashed some of your commits

Johann-S:
@patrickhlauke everything is fine now, can you confirm it ? Thanks 👍

js/src/carousel.js
```diff
@@ -114,8 +116,8 @@ class Carousel {
     this._config            = this._getConfig(config)
     this._element           = element
     this._indicatorsElement = this._element.querySelector(Selector.INDICATORS)
-    this._touchSupported    = 'ontouchstart' in document.documentElement || navigator.maxTouchPoints > 0
     this._pointerEvent      = Boolean(window.PointerEvent || window.MSPointerEvent)
+    this._touchSupported    = 'ontouchstart' in document.documentElement || this._pointerEvent
```
patrickhlauke: the original check for `navigator.maxTouchPoints > 0` was, i'd say, more appropriate. as pointer events (in modern browsers) are supported/present even if there's just a mouse, so you end up with unnecessary code being executed...unless you're doing this on purpose to cover the scenario where an (external) touch-capable display is attached at a later point after the page has already loaded, which is a bit niche...

patrickhlauke: i guess if you did revert back to the `navigator.maxTouchPoints` check, you'd then need to also fake it in the unit tests so that that code gets run

Johann-S: The unit tests wasn't in success because when you swipe on a Imagine with IE it start a drag event 😬 I'll to check maxTouchPoints and see if everything is fine

Johann-S: Ok I restored the use of `navigator.maxTouchPoints`, it should be fine now 👍 I'll make some tests on mobile devices

Johann-S:
It works fine for me 👍 

/CC @XhmikosR if you want to do some tests too

patrickhlauke:
https://deploy-preview-25776--twbs-bootstrap4.netlify.com/docs/4.1/components/carousel/ seems to be working fine now (tested there as i've not got my build/test environment all set up again at the moment) ... but man, the aggressive caching/service worker really gets in the way of quickly checking changes out sometimes...

possibly more related to netlify here: on IE11 and Edge, I found that the carousels are unresponsive for quite a long time after page load (unless you kick them into action if there are the "<" / ">" controls). it seems the browser is waiting for some file to load from the network, and only once it gives up does it run the onload stuff.

XhmikosR:
There is actually no service worker on netlify; it doesn't run the production scripts on purpose.

patrickhlauke:
> There is actually no service worker on netlify; it doesn't run the production scripts on purpose.

i found i had to force refresh (and in the case of Firefox, go in and delete session storage etc) while testing. not quite sure why then...aggressive caching? i see pwa.js is loaded as well if that means anything...i'm actually not sure if this is related to service worker or not.

in any case, it seems to be the carbon ads json that is never loaded (shows as "(pending)" in the F12 tools' network tab). only once the browser gives up trying to load it is the JS on page load actually executed.

anyway, that's all tangential. the carousel swipe thing now seems to work nicely

XhmikosR:
This no longer works for me with Firefox 63 devtools and say S9 emulation. It works fine with Chrome though.

Johann-S:
@XhmikosR that's strange it works perfectly on my Android with Firefox

XhmikosR:
Well a real touch device is not the same as the emulator :) That being said I don't know if we should even care about it. I just thought I'd report it since Firefox is my main browser and this worked in the previous patches.

Johann-S:
I think we should test using BrowserStack here

XhmikosR:
All right, let's merge this then. Great work everyone!

PR #25782, opened 2018-03-05
Improve accessibility.md's clarity

jblang94:
This PR attempts to improve accessiblity.md's clarity by making some phrases more concise. 
A sentence in "Interactive components" is re-structured.

### DIFF
--- a/docs/4.0/getting-started/accessibility.md
+++ b/docs/4.0/getting-started/accessibility.md
@@ -10,25 +10,25 @@ Bootstrap provides an easy-to-use framework of ready-made styles, layout tools, 
 
 ## Overview and Limitations
 
-The overall accessibility of any project built with Bootstrap depends in large part on the author's markup, additional styling, and scripting they've included. However, provided that these have been implemented correctly, it should be perfectly possible to create websites and applications with Bootstrap that fulfill [<abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.0](https://www.w3.org/TR/WCAG20/) (A/AA/AAA), [Section 508](https://www.section508.gov/) and similar accessibility standards and requirements.
+The overall accessibility of any project built with Bootstrap depends in large part on the author's markup, additional styling, and included scripts. However, provided that these have been implemented correctly, it should be perfectly possible to create websites and applications with Bootstrap that fulfill [<abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.0](https://www.w3.org/TR/WCAG20/) (A/AA/AAA), [Section 508](https://www.section508.gov/) and similar accessibility standards and requirements.
 
 ### Structural markup
 
-Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples to demonstrate the use of Bootstrap itself and illustrate appropriate semantic markup, including ways in which potential accessibility concerns can be addressed.
+Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples to demonstrate Bootstrap's use and illustrate appropriate semantic markup, including ways to address potential accessibility concerns.
 
 ### Interactive components
 
-Bootstrap's interactive components—such as modal dialogs, dropdown menus and custom tooltips—are designed to work for touch, mouse and keyboard users. Through the use of relevant [<abbr title="Web Accessibility Initiative">WAI</abbr>-<abbr title="Accessible Rich Internet Applications">ARIA</abbr>](https://www.w3.org/WAI/intro/aria) roles and attributes, these components should also be understandable and operable using assistive technologies (such as screen readers).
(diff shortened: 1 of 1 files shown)


docs/4.0/getting-started/accessibility.md
```diff
@@ -10,25 +10,25 @@ Bootstrap provides an easy-to-use framework of ready-made styles, layout tools,
 
 ## Overview and Limitations
 
-The overall accessibility of any project built with Bootstrap depends in large part on the author's markup, additional styling, and scripting they've included. However, provided that these have been implemented correctly, it should be perfectly possible to create websites and applications with Bootstrap that fulfill [<abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.0](https://www.w3.org/TR/WCAG20/) (A/AA/AAA), [Section 508](https://www.section508.gov/) and similar accessibility standards and requirements.
+The overall accessibility of any project built with Bootstrap depends in large part on the author's markup, additional styling, and included scripts. However, provided that these have been implemented correctly, it should be perfectly possible to create websites and applications with Bootstrap that fulfill [<abbr title="Web Content Accessibility Guidelines">WCAG</abbr> 2.0](https://www.w3.org/TR/WCAG20/) (A/AA/AAA), [Section 508](https://www.section508.gov/) and similar accessibility standards and requirements.
```
patrickhlauke: changing from "scripting they've included" to "included scripts" changes the meaning, i'd say. "included scripts" may be mistaken for scripts included by Bootstrap, while we tried to convey here that it's about scripts that are not part of Bootstrap that authors have added on top of what's provided here.

docs/4.0/getting-started/accessibility.md
```diff
@@ -10,25 +10,25 @@ Bootstrap provides an easy-to-use framework of ready-made styles, layout tools,
...
 
 ### Structural markup
 
-Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples to demonstrate the use of Bootstrap itself and illustrate appropriate semantic markup, including ways in which potential accessibility concerns can be addressed.
+Bootstrap's styling and layout can be applied to a wide range of markup structures. This documentation aims to provide developers with best practice examples to demonstrate Bootstrap's use and illustrate appropriate semantic markup, including ways to address potential accessibility concerns.
```
patrickhlauke: "can be address" > "to address" change feels a bit too absolute. there are always many different ways to address things for accessibility, and what the documentation shows may not always be the best way for all authors...which is why we wanted to keep it a bit more open.

docs/4.0/getting-started/accessibility.md
```diff
@@ -10,25 +10,25 @@ Bootstrap provides an easy-to-use framework of ready-made styles, layout tools,
...
 
 ### Interactive components
 
-Bootstrap's interactive components—such as modal dialogs, dropdown menus and custom tooltips—are designed to work for touch, mouse and keyboard users. Through the use of relevant [<abbr title="Web Accessibility Initiative">WAI</abbr>-<abbr title="Accessible Rich Internet Applications">ARIA</abbr>](https://www.w3.org/WAI/intro/aria) roles and attributes, these components should also be understandable and operable using assistive technologies (such as screen readers).
+Bootstrap's interactive components—such as modal dialogs, dropdown menus and custom tooltips—are designed to work for touch, mouse and keyboard users. Through the use of relevant [<abbr title="Web Accessibility Initiative">WAI</abbr>-<abbr title="Accessible Rich Internet Applications">ARIA</abbr>](https://www.w3.org/WAI/intro/aria) roles and attributes, assistive technologies (ex. screen readers) should be able to understand and use Bootstrap's interactive components.
```
patrickhlauke: this switches from an implied "users who use assistive technologies" who will be able to understand what's going on to the ATs themselves understanding what the components are/do. ATs may not be able to determine what something is, but convey it in a reasonable enough way so that the *users* of the AT will be able to.

docs/4.0/getting-started/accessibility.md
```diff
@@ -10,25 +10,25 @@ Bootstrap provides an easy-to-use framework of ready-made styles, layout tools,
...
 
-Because Bootstrap's components are purposely designed to be fairly generic, authors may need to include further <abbr title="Accessible Rich Internet Applications">ARIA</abbr> roles and attributes, as well as JavaScript behavior, to more accurately convey the precise nature and functionality of their component. This is usually noted in the documentation.
+Because Bootstrap's components are purposely designed to be fairly generic, authors may need to include further <abbr title="Accessible Rich Internet Applications">ARIA</abbr> roles and attributes, as well as JavaScript behavior, to convey the precise nature and functionality of their component accurately. This is usually noted in the documentation.
```
patrickhlauke: "more accurately" is less absolute than "accurately". there are scenarios where things simply can't be fully/accurately conveyed to AT users, but an effort can be made to make them be conveyed *more* accurately than without any ARIA etc.

patrickhlauke (commented):
Appreciate the effort to make things clearer. However, I feel some of the changes may make this too absolute (when accessibility is often about weighing up different options, being only able to mitigate rather than fully solve certain scenarios, etc.) and lose some of the connotations (e.g. the difference between something being understandable to users of ATs versus being understandable to ATs themselves). For this reason, I'd pass on this I'm afraid, but I do appreciate the effort and hope this won't discourage you from filing suggestions in future.

PR #25320, opened 2018-01-15
use callout without custom jekyll plugin

m5o:
* use callout without custom jekyll plugin

:link: https://github.com/twbs/bootstrap/pull/25319

PR #25784, opened 2018-03-05
use example without custom jekyll plugin

m5o:
* use example without custom jekyll plugin
* remove custom example plugin

🔗 https://github.com/twbs/bootstrap/pull/25320
🔗 https://github.com/twbs/bootstrap/pull/25319

Why? Custom Jekyll plugins are a hurdle if you want to use [github-pages](https://pages.github.com).
Bootstrap docs are featured here.

Hope you like it.


XhmikosR:
OK, this is huge, but very promising.

I'll try to get to review the Jekyll PRs soon hopefully, thanks!

_includes/example.html
```diff
@@ -0,0 +1,15 @@
+{%- comment -%}
+Usage: {% include example.html content=markup %},
+where content is a capture with the HTML content
+optional: hide_preview - disabled (default)
+optional: hide_markup - disabled (default)
+{%- endcomment -%}
+
+{% if include.hide_preview == null %}
```
XhmikosR: Doesn't `!include.hide_markup` work?

m5o: @XhmikosR Could change it to `unless` if you find this easier to read:
```erb
{% unless include.hide_markup %}
{% highlight html %}{{ include.content }}{% endhighlight %}
{% endunless %}
```

XhmikosR: nah, it's OK then as it is :)

XhmikosR: Actually, now that I look at it, `hide_preview` and `hide_markup` aren't used anywhere.

So we can remove the checks, or reverse their values, like require it to be `true`/`false`.

What do you think?

docs/4.0/components/alerts.md
```diff
@@ -10,38 +10,38 @@ toc: true
...
 Alerts are available for any length of text, as well as an optional dismiss button. For proper styling, use one of the eight **required** contextual classes (e.g., `.alert-success`). For inline dismissal, use the [alerts jQuery plugin](#dismissing).
 
-{% example html %}
+{% capture example %}
 {% for color in site.data.theme-colors %}
 <div class="alert alert-{{ color.name }}" role="alert">
   A simple {{ color.name }} alert—check it out!
 </div>{% endfor %}
-{% endexample %}
+{% endcapture %}{% include example.html content=example %}
```
XhmikosR: Is there any reason you have added the `include`s to the same line?

m5o: No reason, just easier the first search-and-replace batch - I've change it if you want

_includes/example.html
m5o: Currently they are not used, but there are so many `<div class="bd-example">[…]</div>` blocks like [here for example](https://github.com/twbs/bootstrap/blame/v4-dev/docs/4.0/components/button-group.md#L91) that can be refactored in the future.

XhmikosR:
OK, so there are things missing from your implementation compared to the plugin. Please check out what it could do like:

1. removing holder from src
2. remove example classes
3. the functionality of passing an ID and a class

m5o:
1. removing the [holder.js](https://github.com/imsky/holder) is not necessary anymore, the [`data-src`](https://github.com/imsky/holder/blame/master/README.md#L44) is used everywhere.
1. there are actually no left `bd-` classes inside of the example captures, which requires no modification
1. Done - https://github.com/twbs/bootstrap/pull/25784/commits/93e11fab81677f24d45562e2e6e3b48b8254382f

XhmikosR:
1. it is like before: we don't want to show it in the highlight
2. OK...
3. Thanks!

m5o:
> To avoid console 404 errors, you can use `data-src` instead of `src`.

🔖 https://github.com/imsky/holder#usage

There is not a single holder.js reference in a `src`. Therefore no action required.

XhmikosR:
Please see what we were doing with the plugin since you don't understand it.

m5o:
Sorry, just don’t get it you only meant the markup part. Regular expression are not supported by [Liquid](https://help.shopify.com/themes/liquid/filters/string-filters#replace). https://github.com/twbs/bootstrap/pull/25784/commits/7d16bf8d52529a73c7e48b54d5a4919f4dc4ccec removes holder.js functionality.

_includes/example.html
```diff
@@ -0,0 +1,21 @@
...
+id - null (default)
+class - "bd-example" (default)
+optional: hide_preview - disabled (default)
+optional: hide_markup - disabled (default)
+{%- endcomment -%}
+
+{%- assign preview_id = include.id -%}
+{%- assign preview_class = include.class | default: "bd-example" -%}
+
+{%- if include.hide_preview == null -%}
```
XhmikosR: OK, I think we should change the check to `if include.hide_preview != false`. Can you try it and change the two checks?

_includes/example.html
```diff
@@ -0,0 +1,23 @@
...
+Usage: {% include example.html content=markup %},
+where content is a capture with the HTML content
+id - null (default)
+class - "bd-example" (default)
+optional: hide_preview - disabled (default)
+optional: hide_markup - disabled (default)
+{%- endcomment -%}
+
+{%- assign preview_id = include.id -%}
+{%- assign preview_class = include.class | default: "bd-example" -%}
```
XhmikosR: I'm not sure this does what we want. We want to append a class we pass and always have `bd-example`. So maybe try using append with a space.

_includes/example.html
m5o: Does not work, with `!= false`. Guess it does a [`not equals`](https://help.shopify.com/themes/liquid/basics/operators) check

XhmikosR (approved):

PR #25803, opened 2018-03-07
Enable docsearch in development env

m5o:
* enables docsearch in development environment
  * previously a search hard redirect to `https://getbootstrap.com/…`

### DIFF
--- a/assets/js/src/application.js
+++ b/assets/js/src/application.js
@@ -115,8 +115,13 @@
         },
         transformData: function (hits) {
           return hits.map(function (hit) {
-            var baseurl = document.getElementById('search-input').getAttribute('data-baseurl')
-            hit.url = hit.url.replace('https://getbootstrap.com' + baseurl, baseurl)
+            // When in production, return the result as is,
+            // otherwise remove our url from it.
+            var siteurl = document.getElementById('search-input').getAttribute('data-siteurl')
+            var urlRE = /^https?:\/\/getbootstrap\.com/
+
+            hit.url = siteurl.match(urlRE) ? hit.url : hit.url.replace(urlRE, '')
+
             return hit
           })
         },
--- a/assets/js/src/application.js
+++ b/assets/js/src/application.js
@@ -115,7 +115,8 @@
         },
         transformData: function (hits) {
           return hits.map(function (hit) {
-            hit.url = hit.url.replace('https://v4-alpha.getbootstrap.com', '/docs/4.0')
+            var baseurl = document.getElementById('search-input').getAttribute('data-baseurl')
+            hit.url = hit.url.replace('https://getbootstrap.com' + baseurl, baseurl)
             return hit
           })
         },
--- a/_includes/docs-sidebar.html
+++ b/_includes/docs-sidebar.html
@@ -1,5 +1,5 @@
 <form class="bd-search d-flex align-items-center">
-  <input type="search" class="form-control" id="search-input" placeholder="Search..." aria-label="Search for..." autocomplete="off" data-baseurl="{{ site.baseurl }}/docs/{{ site.docs_version }}">
+  <input type="search" class="form-control" id="search-input" placeholder="Search..." aria-label="Search for..." autocomplete="off" data-siteurl="{{ site.url }}">
   <button class="btn btn-link bd-search-docs-toggle d-md-none p-0 ml-3" type="button" data-toggle="collapse" data-target="#bd-docs-nav" aria-controls="bd-docs-nav" aria-expanded="false" aria-label="Toggle docs navigation">
     {%- include icons/menu.svg width="30" height="30" -%}
   </button>
--- a/_includes/docs-sidebar.html
+++ b/_includes/docs-sidebar.html
@@ -1,5 +1,5 @@
 <form class="bd-search d-flex align-items-center">
-  <input type="search" class="form-control" id="search-input" placeholder="Search..." aria-label="Search for..." autocomplete="off">
+  <input type="search" class="form-control" id="search-input" placeholder="Search..." aria-label="Search for..." autocomplete="off" data-baseurl="{{ site.baseurl }}/docs/{{ site.docs_version }}">
   <button class="btn btn-link bd-search-docs-toggle d-md-none p-0 ml-3" type="button" data-toggle="collapse" data-target="#bd-docs-nav" aria-controls="bd-docs-nav" aria-expanded="false" aria-label="Toggle docs navigation">
     {%- include icons/menu.svg width="30" height="30" -%}
   </button>


assets/js/src/application.js
```diff
@@ -115,7 +115,8 @@
         },
         transformData: function (hits) {
           return hits.map(function (hit) {
-            hit.url = hit.url.replace('https://v4-alpha.getbootstrap.com', '/docs/4.0')
+            var baseurl = $('#search-input').data('baseurl')
+            hit.url = hit.url.replace('https://getbootstrap.com' + baseurl, baseurl)
```
XhmikosR: @mdo: any reason you had the old URL there?

XhmikosR:
@m5o: can you keep both `replace` calls? We can remove the old one later if it's not needed.

m5o:
> @m5o: can you keep both replace calls? We can remove the old one later if it's not needed.

Sorry, I don't get it

XhmikosR:
I mean the replacement of `https://v4-alpha.getbootstrap.com`.

assets/js/src/application.js
```diff
@@ -116,6 +116,8 @@
         transformData: function (hits) {
           return hits.map(function (hit) {
             hit.url = hit.url.replace('https://v4-alpha.getbootstrap.com', '/docs/4.0')
+            var baseurl = $('#search-input').data('baseurl')
```
XhmikosR: Can you try moving this above and chain the two calls?

Also, can you use vanilla JS? Like getElementById and getAttribute? Maybe like `document.getElementById('search-input').getAttribute(data-baseyrl)` (untested).

assets/js/src/application.js
```diff
@@ -115,7 +115,9 @@
         },
         transformData: function (hits) {
           return hits.map(function (hit) {
+            var baseurl = document.getElementById('search-input').dataset.baseurl
```
XhmikosR: Since IE 10 is still supported, please use `getAttribute` because it doesn't support `dataset`.

Also, did you verify everything works locally? I haven't had the time to test it yet.

m5o: Yes, it still works.

XhmikosR:
@maxiloc: looks good to you? Any suggestions, maybe?

s-pace:
👋 I am working on DocSearch, I will take over from @maxiloc .

From what I get you want to change the URL from the hit we have scrapped within you DocSearch index (`bootstrap-v4`).

Example for`query=changes`:
**Current result from your DocSearch index:**
`hit.url=https://getbootstrap.com/docs/4.0/migration/#stable-changes`
From what I get:
```js
data-baseurl={{ site.baseurl }}/docs/{{ site.docs_version }}="https://getbootstrap.com/docs/4.0/"
```

What you are doing:
```js
hit.url.replace('https://v4-alpha.getbootstrap.com', '/docs/4.0')
``` This replacement is useless since 'https://v4-alpha.getbootstrap.com' doesn't match the URL `hit.url`.

Furthermore, you want to replace the pattern:
`'https://getbootstrap.com' + baseurl=https://getbootstrap.comhttps://getbootstrap.com/docs/4.0/`
This pattern will not be matched either.

After all, the url will be the same.

Let me know if I missed something.

**Comment by XhmikosR** [comment]
@s-pace: first of all, thanks for the reply!

Now, what we want to do is replace the baseurl for local development where `{{ site.baseurl }}` is basically empty.

For production it's set to `https://getbootstrap.com`. The `v4-alpha.getbootstrap.com` part must be from the alpha time, I'm not sure we need it either, so we can probably indeed remove it.

**Comment by XhmikosR** [comment]
Also, note that if you try this PR locally, it indeed works fine. I just feel something is not right in the implementation.

**Comment by s-pace** [comment]
@XhmikosR 

You are welcome.

This should work if the URL & the anchor (so the layout) are the same between your dev and production environment.

**Comment by m5o** [comment]
> For production it's set to `https://getbootstrap.com`.

🙅‍♂️ Not true, you can validate by compare: – Or in't this valid proof? https://github.com/twbs/bootstrap/blob/d01b4eb02560ba67b61b1b2c0496e9f8281c154d/index.html#L9
with the generated markup from http://getbootstrap.com/

```html
<div class="col-6 mx-auto col-md-6 order-md-2">
  <img class="img-fluid mb-3 mb-md-0" src="/assets/img/bootstrap-stack.png" alt="" width="1024" height="860">
</div>
```
---
History wise these changes are important https://github.com/twbs/bootstrap/pull/22731 and https://github.com/algolia/docsearch-configs/commit/ce46787bb133ed6f0833a6c601683ce386f2afed . 
`v4-alpha.getbootstrap.com` is definitely a leftover and can be removed. I've tried it.

The main question is, maybe @s-pace can help. The [returned `hit.url`](https://github.com/algolia/docsearch-configs/blob/d4f85e117008450881fac8f56c2ea5fbaad9587f/configs/bootstrap-v4.json#L4) is an absolute url, can this a relative url instead? A relative url would work in `development` and `production` environment.

```diff
+https://getbootstrap.com/docs/4.0/migration/#stable-changes
-/docs/4.0/migration/#stable-changes
```
[algolia/docsearch-configs/configs/bootstrap-v4.json](https://github.com/algolia/docsearch-configs/blob/d4f85e117008450881fac8f56c2ea5fbaad9587f/configs/bootstrap-v4.json#L4)

XhmikosR:
True, `baseurl` is always empty. `site.url` changes which is why I mixed those two.

XhmikosR:
So, basically, all this PR does is this:

1. passes a data attribute with the `baseurl` which is translated to `/docs/4.0` for both dev and production
2. It then replaces this from the search results: `https://getbootstrap.com/docs/4.0` with `/docs/4.0`

So, it seems that we can simplify this and not rely to the empty `baseurl`.

XhmikosR:
Please check out the last patch I pushed. Basically I rely on `site.url` which is different for production and development. When in development, I just remove the url from the search results, making them relative.

Example searching for `tooltip`:
```
baseurl: http://0.0.0.0:9001
hit.url: /docs/4.0/components/tooltips/#overview
```

assets/js/src/application.js
```diff
@@ -115,7 +115,13 @@
         },
         transformData: function (hits) {
           return hits.map(function (hit) {
-            hit.url = hit.url.replace('https://v4-alpha.getbootstrap.com', '/docs/4.0')
+            // When in production, return the result as is,
+            // otherwise remove our url from it.
+            var baseurl = document.getElementById('search-input').getAttribute('data-baseurl')
```
m5o: 💅 `data-siteurl` to be consistent with the value

m5o:
👍 thank you

XhmikosR (approved):

Issue #25817, opened 2018-03-09
Spies, Mocks and stub for our Unit test


Johann-S:
We should add a libs which will allow us to use Spies, Mocks and Stubs, I thought about Sinon : http://sinonjs.org/

It would be easier for us to test our plugins

Because, I work on improving our code coverage but it's very difficult to test everything with QUnit which do not provide enough tools for that...

/CC @XhmikosR, @bardiharborow

PR #25820, opened 2018-03-09
Add Sinon to do better unit test

Johann-S:
Fixes: #25817

/CC @XhmikosR , @bardiharborow


js/tests/karma.conf.js
```diff
@@ -19,6 +19,7 @@ module.exports = (config) => {
     // list of files / patterns to load in the browser
     files: [
       jqueryFile,
+      'js/tests/vendor/sinon.min.js',
```
Johann-S: Maybe should I add `karma-sinon` instead of adding this file 🤔

AshikurRahman-Peters:
4f2a137a484c5740ad03b294a51d0791

js/tests/karma.conf.js
XhmikosR: If we can achieve the same without the manually copied dependency, then go for it 👍

Johann-S: We'll still need the local file to run our unit test without Karma

XhmikosR: Your call. I'm not very familiar with the karma config.

Also, I don't quite get this whole change, care to elaborate a bit?

Johann-S: Sure 😃 QUnit possess just a few methods compare to newer Framework of unit testing (Karma, Mocha), they are a lot of new concepts (Mocks, Spies, Stubs) which provide a better way to test our methods, because for example you can be sure a method have been called, or fake user data, or change temporary the behavior of our methods.

See : http://sinonjs.org/#try-it-out

XhmikosR (approved):
Looks good as far as I can review this.

Issue #24285, opened 2017-10-07
Card with .card-img-overlay and overflowing text


klor:
Cards with lots of text do not keep overflowing text inside the card box (Bootstrap 4).

In the example below, the card text overflows the H1 and div.row below it. The example uses an [image overlay](https://getbootstrap.com/docs/4.0/components/card/#image-overlays) example from the docs.

Screenshot:
![image](https://user-images.githubusercontent.com/129573/31305598-c61c2d1e-ab3d-11e7-9aaf-561719a423b2.png)

HTML:

```html
<!DOCTYPE html>
<html>
<head>
    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <title>Cards</title>
    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.min.css" integrity="sha384-/Y6pD6FV/Vv2HJnA6t+vslU6fwYXjCFtcEpHbNJ0lyAFsXTsjBbfaDjzALeQsN6M" crossorigin="anonymous">
</head>
<body>


<div class="container">
    <div class="card bg-dark text-white">
      <img class="card-img" src="http://lorempixel.com/1822/937/sports/" alt="Card image">
      <div class="card-img-overlay">
        <h4 class="card-title">Card title</h4>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
        <p class="card-text">This is a wider card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
[...]

PR #25822, opened 2018-03-10
feat(css): add `.card-bg`

zalog:
I think this is a feature opposite to `.card-img-overlay` because the way that our `.card` height is computed.

So, in this pr, our card has:
- his parent height relative to content, not to image like `.card-img-overlay`
- a 100% width&height background on the back of the card body content

But, because our `.card` has more childrens than `.card-body`, I think we should introduce a new `+ .card-bg-overlay` or something like that.

What do you guys think?

Here we have a simple demo, and a [pen](https://codepen.io/zalog/full/oqgPJv/), of course :).

bootstrap card-bg


andresgalante:
sorry :( Miss clicked on close while reviewing it

andresgalante (commented):
Hey @zalog, thanks for the PR.

I think that what happens with the overflowing text on `.card-img-overlay` is a problem.

@mdo I know that on #24285 we decided not to take actions, but should we change `card-img-overlay` instead of introducing a new feature?

@zalog this PR will need docs if it gets merge

zalog:
I'd vote too for not introducing a new feature, but I think that both cases are common. Perhaps, parent height computed by the childrens with text, are more common?

I can tackle whatever we decide.

mdo:
Coming back to this real late. Any chance we can get a demo added to the docs? No sweat if not, I can add one myself after I review this. I agree that the overlay behavior is less than ideal currently.

zalog:
Ok, I'll come back to this soon.

XhmikosR:
@zalog: ping

MartijnCuppens:
I think it's better to just use a background image for these scenarios (like @mdo mentioned here: https://github.com/twbs/bootstrap/issues/24285#issuecomment-337945556). We will still have the issue that the image will be stretched as long as we don't use `object-fit` (which we can't use because of IE we support).

PR #25823, opened 2018-03-10
Fix scrollspy test on firefox (v4 without jquery)

alekitto:
Related to https://github.com/twbs/bootstrap/pull/23586#issuecomment-371472645
Seems that jQuery `position()` method calculates position including sub-pixels on firefox and returning floating point numbers as offsets, while native offsetTop property does not.
Maybe should we remove jQuery from our unit tests and add some compatibility tests using the jQuery interface?


Johann-S:
Yep it's in the roadmap to remove jQuery from our unit test 😉 
thanks ! 👍

PR #25827, opened 2018-03-12
Add extra columns to table docs

### LINKED ISSUE #25813: Clarify docs of breakpoint responsive table
I'm not fully understood those use cases sorry, but there is almost no difference in [the demo of the breakpoint specific responsive tables](http://getbootstrap.com/docs/4.0/content/tables/#breakpoint-specific).

table

Could someone update the demo so that the effective usage of these breakpoint options will become clear? Thanks.

Varunram:
Partially address #25813.


mdo:
Closing since this doesn't address the docs problem as originally reported.

mdo:
Actually, reopening to at least partially address the docs.

PR #25834, opened 2018-03-13
fix get the transition duration parent

Johann-S:
Fix our build: https://travis-ci.org/twbs/bootstrap/jobs/352733509

/CC @XhmikosR to fix our build and @MartijnCuppens

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -106,19 +106,22 @@ const Util = (($) => {
     },
 
     getTransitionDurationFromElement(element) {
+      if (!element) {
+        return 0
+      }
+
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
Why would you get the transition duration from the parent if the element has no transition duration?

Johann-S:
Because as I linked, you cannot get the inherited transition duration on Safari, and in Bootstrap sometimes we use `Util.getTransitionDurationFromElement` with a child element which do not contain any transition duration

XhmikosR (approved):

Issue #25658, opened 2018-02-19
Card width excesively narrow


usuarionuevor:
`card` get too shrunk before suddenly turn into one card instead of making a more progressively transition like 3cards > 2cards > 1card.

 

Example: https://getbootstrap.com/docs/4.0/components/card/

 
![card](https://user-images.githubusercontent.com/35457371/36371970-ecfad706-1563-11e8-8caf-5ae72100d190.png)

 

and then:

 

![1](https://user-images.githubusercontent.com/35457371/36372020-114af4d8-1564-11e8-907c-3d32cbcd7335.png)

mdo:
Going to close as a won't fix for v4. We'll end up deleting this card layout come v5 given it's lack of control over horizontal card layouts like Masonry can provide.

PR #25841, opened 2018-03-13
#25658 More responsive card columns

bakedbird:
Replace $columns-count with $columns-sm-count and $columns-md-count that responsively span cards in 2 and 3 columns respectively. 
Include media-breakpoint mixin for sm and md devices to use the new variables


mdo:
This variable change breaks backward compatibility.

PR #25847, opened 2018-03-14
Remove img-retina mixin

### LINKED ISSUE #25754: img-retina mixin
The `img-retina` mixin is not documented in the docs.
https://github.com/twbs/bootstrap/blob/29f178d89129732d35ae812667709c64cdb28a6f/scss/mixins/_image.scss#L19-L36

Shall we:
- document this
- or remove the mixin (anno 2018 it might be better to use `.svg`s as background images)

MartijnCuppens:
Closes #25754.

### DIFF
--- a/scss/mixins/_image.scss
+++ b/scss/mixins/_image.scss
@@ -1,6 +1,4 @@
 // Image Mixins
-// - Responsive image
-// - Retina image
 
 
 // Responsive image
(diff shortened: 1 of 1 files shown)


mdo:
This is a breaking change as it removes functionality, regardless of it's usefulness today :). We'd need to deprecate and not remove outright.

MartijnCuppens:
I'm closing this.
Discussion about mixins documentation here: #25977
Once we have an overview of all the mixins, we can figure out which mixins need to be deprecated in the future.

PR #25855, opened 2018-03-15
Fix order of Save/Close buttons in modal code example

### LINKED ISSUE #25853: In docs, .modal-footer code is not the same as output
I am referring to the [modal components](https://getbootstrap.com/docs/4.0/components/modal/#modal-components) section of the Bootstrap documentation for v4.

The code example shows:
```html
<div class="modal-footer">
  <button type="button" class="btn btn-primary">Save changes</button>
  <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
</div>
```

But the output rendered is:
image

And upon inspecting the rendered HTML content, its:
```html
<div class="modal-footer">
  <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
  <button type="button" class="btn btn-primary">Save changes</button>
</div>
```

Looks like a small mistake on the order of the buttons in the code example?

Bennet11:
Hi, Im a first time contributor and I'm submitting this pull request to resolve the issue submitted by the author of the post. Please let me know if there are problems arising from this pull request. Thank you!

EDIT: Fixes #25853

### DIFF
--- a/docs/4.0/components/modal.md
+++ b/docs/4.0/components/modal.md
@@ -45,8 +45,8 @@ Below is a _static_ modal example (meaning its `position` and `display` have bee
           <p>Modal body text goes here.</p>
         </div>
         <div class="modal-footer">
-          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
           <button type="button" class="btn btn-primary">Save changes</button>
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
         </div>
       </div>
     </div>
--- a/docs/4.0/components/modal.md
+++ b/docs/4.0/components/modal.md
@@ -45,8 +45,8 @@ Below is a _static_ modal example (meaning its `position` and `display` have bee
           <p>Modal body text goes here.</p>
         </div>
         <div class="modal-footer">
-          <button type="button" class="btn btn-primary">Save changes</button>
           <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+          <button type="button" class="btn btn-primary">Save changes</button>
         </div>
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


kevdotbadger:
Am I losing my shit, or is there no typo change at all in this commit? Just looks like a re-ordering of the button

patrickhlauke:
Indeed. This just looks like you're changing the order of the Save/Close buttons? if you were proposing changing this, you should also really propose the change for all instances on that page.

patrickhlauke:
ah, having now found https://github.com/twbs/bootstrap/issues/25853 i see what you meant. I'd suggest the real fix here is to edit the code sample, not the live modal, to keep it consistent with the rest of the examples

docs/4.0/components/modal.md
```diff
@@ -45,8 +45,8 @@ Below is a _static_ modal example (meaning its `position` and `display` have bee
           <p>Modal body text goes here.</p>
         </div>
         <div class="modal-footer">
-          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
           <button type="button" class="btn btn-primary">Save changes</button>
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
```
patrickhlauke: leave the order here as it is originally, but change the order in the `{% highlight html %}` that comes immediately after instead. the live example is right/consistent with the rest of the page, it's the visible highlighted html code that's wrong

patrickhlauke (changes requested):
if you could push a new commit to this, which reverts your original change and then makes the change in the highlight html, that'd be super! i can squash the PR in the end so it'll just result in a single commit, no need to try and rewrite history or open a new PR

XhmikosR:
@patrickhlauke: anyone with push rights can edit the file too :)

Try https://github.com/Bennet11/bootstrap/edit/johnB_contri/docs/4.0/components/modal.md?pr=/twbs/bootstrap/pull/25855

And then squash as usual via the button.

This is the UI solution; the manual one applies too, but it's more troublesome for simple stuff.

docs/4.0/components/modal.md
Bennet11: Thank you for the input! I'll try to do it now.

Bennet11: The test seems to be failing though.

patrickhlauke (approved):

patrickhlauke:
@Bennet11 thank you 👍

PR #25864, opened 2018-03-15
#25842 Remove max-width from table class

### LINKED ISSUE #25842: The background color of responsive table with .table-dark is not applied
When scrolling the responsive table with `.table-dark`, the background color of the hidden part is not displayed correctly.

image

Delete `max-width: 100%;` property of `.table` will display correctly.

image

- OS: iOS 10.3.3
- Browser: Safari
- Sample Code: https://codepen.io/aoi70/pen/ZxQRer

bakedbird:
max-width: 100% was causing issues on 
macOs safari
iOs safari

Removed max-width and tested for possible side-effects on:
os: macOs
browsers: Safari, Chrome, Opera, Firefox

and iPhoneX safari (on browserStack)

Fixes #25842

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -4,7 +4,6 @@
 
 .table {
   width: 100%;
-  max-width: 100%;
   margin-bottom: $spacer;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
 


mdo:
Without a demo reproducing these problems you mention, we cannot accept the PR.

bakedbird:
@mdo what do you mean? The PR was the fix for #25842

andresgalante:
@mdo I'll reopen this one since there is [an example](https://codepen.io/aoi70/pen/ZxQRer) in this issue #25842

I haven't review the PR yet.

aledbrown:
I can confirm this bug happens on iOS Safari with class="table table-dark" and a responsive table div surrounding the table. The dark table only renders to the initial view box. Scroll the table horizontally and the background goes white again.

The fix also seems to work fine. So I've done this to fix it for now on the specific table I'm having the issue with:

```
<div class="table-responsive-xl">
<table style="max-width: none !important;" class="table table-dark table-striped table-hover table-bordered">
...
```

mdo (approved):

PR #25893, opened 2018-03-19
Move card border on pseudo element

dahfazz:
On cards with border and image inside, we can see a 1px blank area between border and img.
Moving the border on pseudo element fix this.

+ z-index to put border over inner elements

+ pointer-events: none to let click event go through the pseudo element


mdo:
Thanks, but we'll pass on this. It's a stylistic choice and the `pointer-events` property isn't available across every browser.

PR #25896, opened 2018-03-20
Use transitionEnd in QUnit since we moved away from PhantomJS

Johann-S:
Remove that code because it's useless since we moved away from PhantomJS, plus by doing that our coverage increased a lot 😍

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -14,8 +14,7 @@ const Util = (($) => {
    * ------------------------------------------------------------------------
    */
 
-  let transition = false
-
+  const TRANSITION_END = 'transitionend'
   const MAX_UID = 1000000
   const MILLISECONDS_MULTIPLIER = 1000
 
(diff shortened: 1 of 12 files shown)


js/src/util.js
```diff
@@ -14,7 +14,9 @@ const Util = (($) => {
    * ------------------------------------------------------------------------
    */
 
-  let transition = false
+  const transition = {
```
XhmikosR: Do we really need it to be an object now?

Johann-S:
Done @XhmikosR, plus I removed `Util.supportsTransitionEnd` which is useless now 👍

XhmikosR:
I wonder, won't this break stuff for people using `Util.supportsTransitionEnd()`?

Johann-S:
it can be a breaking change but if we let this method it will be the same as : 

```js
supportsTransitionEnd() {
  return true
}
```

because we are compatible with browsers whose support transition end.

XhmikosR:
Yeah, I think we should keep compatibility and add a todo comment that this should be removed in v5

Johann-S:
I did that 😉 `Util.supportsTransitionEnd` is back with a TODO

XhmikosR:
Thanks!

How about `transitionEndTest()`? Maybe we should do the same for this too?

Johann-S:
`transitionEndTest` isn't callable currently, so it's not a breaking change

XhmikosR (approved):

PR #25898, opened 2018-03-20
throw error when coverage is under our thresholds

Johann-S:
Before that, we just had warnings, now it'll throw an error on Travis if our coverage is too low

/CC @XhmikosR @bardiharborow

### DIFF
--- a/js/tests/karma.conf.js
+++ b/js/tests/karma.conf.js
@@ -57,14 +57,13 @@ module.exports = (config) => {
     coverageIstanbulReporter: {
       dir: jsCoveragePath,
       reports: ['lcov', 'text-summary'],
-      fixWebpackSourcePaths: true,
       thresholds: {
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

js/tests/karma.conf.js
```diff
@@ -57,14 +57,13 @@ module.exports = (config) => {
...
-      fixWebpackSourcePaths: true,
       thresholds: {
-        emitWarning: true,
+        emitWarning: false,
         global: {
-          statements: 80,
-          lines: 80,
-          branches: 80,
-          functions: 80
+          statements: 89,
```
XhmikosR: Are you sure about the new numbers? Maybe we should be less strict?

Johann-S: For me it's pretty low we should be at least at 95% on those numbers

PR #25919, opened 2018-03-22
Fixes Edge and IE label overlapping issue (Floating Labels Example)

coliff:
The 'Floating labels' example (https://getbootstrap.com/docs/4.1/examples/floating-labels/) is only supported on Chrome, Safari and Firefox. On Edge and IE the label is displayed always and any text input overlaps the label which makes it unreadable.

The fix here targets Edge and IE and makes the form behave normally, the labels are hidden and the placeholder color is standard.


docs/4.0/examples/floating-labels/floating-labels.css
```diff
@@ -82,3 +82,25 @@ body {
...
 }
+
+/* Fallback for Edge
+-------------------------------------------------- */
+@supports (-ms-ime-align: auto) {
+  .form-label-group > label {
+    display: none;
+  }
+  .form-label-group input::-ms-input-placeholder {
+    color: #777 !important;
```
XhmikosR: Why is `!important` needed? Isn't this pseudo selector IE only?

This is the reason tests fail BTW.

coliff:
@XhmikosR so basically make this change on:
github.com/twbs/bootstrap/blob/v4-dev/docs/4.1/examples/floating-labels/floating-labels.css
and remove the commit for 4.0 docs?
(I don't usually use Git via command line so wasn't sure)

XhmikosR:
@coliff: if you use a recent git version, after you fetch upstream and do a rebase against upstream/v4-dev, it should follow the file rename automatically.

If you don't succeed, just ping me and I'll try to rebase this for you.

XhmikosR:
Nvm, I rebased it for you.

coliff:
thanks @XhmikosR !

mdo (approved):

PR #25921, opened 2018-03-23
Fieldset double rule

massiws:
In version 3.3.7 there are two declarations on `<fieldset>` and `<label>` tags, one coming from `bootstrap.less` and another from `normalize.less`.
They apply different styles but, obviously, the first (from Normalize) are overridden by the latter (from Bootstrap), so they can be removed.


wolfy1339:
This rule is there for a reason.
Another point: don't edit the dist files directly, second: you're targeting the wrong branch here

Johann-S:
You're trying to merge v3 in our v4, so I'll close your PR please make a new one correctly

PR #25924, opened 2018-03-23
Add note: floats don't mix with miscellaneous modules

### LINKED ISSUE #25829: .float-right doesn't work inside a .row
The button in the HTML below ([codepen.io](https://codepen.io/anon/pen/aYOQxL)) remains left-aligned despite the `float-right` class.

``` html
<!DOCTYPE html>
<html>
<head>
  <title>.float-right issue</title>
  <link rel="stylesheet" href="css/bootstrap.css">
  <style>.row { background-color: #ff8621; }</style>
</head>
<body>
  <div class="container">
    <div class="row">
      <button class="btn btn-primary float-right">To the right!</button>
    </div>
  </div>
</body>
</html>
```

Firefox 58.0.2, linux/amd64

Varunram:
Closes #25829. Could change where we place the sentence, open to suggestions (I just thought creating a new section with caveats would be bad).

### DIFF
--- a/docs/4.0/utilities/float.md
+++ b/docs/4.0/utilities/float.md
@@ -8,7 +8,7 @@ toc: true
 
 ## Overview
 
-These utility classes float an element to the left or right, or disable floating, based on the current viewport size using the [CSS `float` property](https://developer.mozilla.org/en-US/docs/Web/CSS/float). `!important` is included to avoid specificity issues. These use the same viewport breakpoints as our grid system.
+These utility classes float an element to the left or right, or disable floating, based on the current viewport size using the [CSS `float` property](https://developer.mozilla.org/en-US/docs/Web/CSS/float). `!important` is included to avoid specificity issues. These however, don't mix with miscellaneous layout modules such as flex or grid. These use the same viewport breakpoints as our grid system.
 
 ## Classes
 


danijelGombac:
I use it with grid system, but i need to add d-block clearfix to row. It's seems to work fine. But i can't use flex utility.

mdo:
> I use it with grid system, but i need to add d-block clearfix to row. It's seems to work fine. But i can't use flex utility.

You shouldn't be doing that. If you're reverting to floats, you should just avoid the grid entirely and roll your own custom stuff. Try out the flexbox approach more before jumping to that.

danijelGombac:
@mdo 

I am trying few days play with flexbox, but without a result. The problem is that flex item can't be in two flexbox line.

https://codepen.io/danijelGombac/pen/yKoXGB

mdo:
Yeah, for stuff like that, you're relying on float's ability to collapse against content around it. If you changed your markup a bit, it wouldn't be an issue. Also, you could (depending on browser support you need) use grid layout.

danijelGombac:
@mdo 

I tried different markup, but without a good result. On mobile i need the sequence of card 1,2,3. For now i leave floats, I don't want to use a grid layout (browser support), thanks.

PR #25925, opened 2018-03-23
Remove justify-content from navbar

### LINKED ISSUE #25697: Navbar style includes justify content property
The .navbar class includes `justify-content: space-between; // space out brand from logo` which means alignment styles are applied to the navbar by default, which is not desired. 

The BS4 documentation does not reflect this; it indicates you should add the "justify-content-space-between" class when this styling is desired. This seems like a bug/oversight to me. It is not desirable to override this style every time .navbar is used.

Varunram:
Closes #25697. I fi understood the issue properly, this property wasn't desired

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -20,7 +20,6 @@
   display: flex;
   flex-wrap: wrap; // allow us to do the line break for collapsing content
   align-items: center;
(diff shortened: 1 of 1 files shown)


mdo:
This is an incorrect fix for a docs error. Thanks, though! See #26171.

PR #25933, opened 2018-03-24
Add .text-decoration-none utility class

swrobel:
ref: https://github.com/twbs/bootstrap/pull/23881#issuecomment-337789596


swrobel:
Come on guys, this totally could've made 4.1!

PR #25934, opened 2018-03-24
Fix css-docs scripts

patrik-csak:
- In `css-compile-docs` script, change `docs.min.css` to `docs.css` because it's not minified
- In `css-prefix-docs` script:
  - Point PostCSS at `docs.css` instead of `docs.min.css`
  - Exclude `docs.min.css` so it's consistent with the `css-prefix` script
- In `css-minify-docs` script, change clean-css input file from `docs.min.css` to `docs.css`
- Add `docs.css` and `docs.css.map`, which is consitent with the tracked built files in `dist/css`


ptrkcsk:
I included the updates to `docs/` as they are the result of the `css-prefix-docs` script, but I'm not sure if they're in the scope of this PR.

mdo:
This isn't necessary IMO. Our build scripts are always used to create minified docs CSS—we'll have have a non-minified one when developing our docs. Thus, there's no need to create and have Git track another file in the repo. Thanks, though!

ptrkcsk:
Ah, got it. Thanks for maintaining Bootstrap, @mdo.

PR #25954, opened 2018-03-27
docs: Switch to StackPath's URL for the CDN.

XhmikosR:
While the old one will keep working, better switch to the new one.

@jdorfman: I'm thinking we should first do the related changes on BootstrapCDN and then start using the new URL. It might be confusing otherwise to some people.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -67,9 +67,9 @@ download:
 
 cdn:
   # See https://www.srihash.org for info on how to generate the hashes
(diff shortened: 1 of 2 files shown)


XhmikosR:
All right, I made https://github.com/MaxCDN/bootstrapcdn/pull/988, so when we merge and deploy, we can merge this one here too.

bardiharborow (approved):

mdo (approved):

XhmikosR:
@mdo: this shouldn't have been merged yet.

PR #25958, opened 2018-03-27
add cursor

dotku:


afuno:
Maybe then there should be added all cursors?

mdo:
Thanks, but we'll pass on this for now.

PR #25968, opened 2018-03-29
remove default background-color from breadcrumb

m5o:
* remove default background-color from breadcrumb
* re-use contextual background classes for background styling

### DIFF
--- a/docs/4.1/components/breadcrumb.md
+++ b/docs/4.1/components/breadcrumb.md
@@ -9,20 +9,20 @@ group: components
 
 {% capture example %}
 <nav aria-label="breadcrumb">
-  <ol class="breadcrumb">
+  <ol class="breadcrumb bg-light">
     <li class="breadcrumb-item active" aria-current="page">Home</li>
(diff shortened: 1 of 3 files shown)


mdo:
Removing variables breaks backward compatibility. We'll want to leave the defaults and add docs for how to override those with utilities. Come v5, we can revisit the default styles.

m5o:
@mdo UIkit has a `.[uk-]text-reset` / `.[uk-]link-reset` util class which reset `color` and `text-decoration`. This would be super handy.

```css
a.uk-link-reset, a.uk-link-reset:hover, .uk-link-reset a, .uk-link-reset a:hover {
  color: inherit !important;
  text-decoration: none !important;
}
```

📖 https://getuikit.com/docs/link#link-reset

mdo:
Closing this out since it's stale and cannot be merged as-is into v4. Happy to revisit in v5, and to revisit the link reset in a subsequent minor release if there's a PR for it.

PR #25969, opened 2018-03-29
inherit alert-link from parent

m5o:
* inherit `.alert-link` from parent

Before | After
--- | ---
screen shot 2018-03-29 at 15 51 07 | screen shot 2018-03-29 at 15 50 37

### DIFF
--- a/scss/mixins/_alert.scss
+++ b/scss/mixins/_alert.scss
@@ -8,6 +8,6 @@
   }
 
   .alert-link {
-    color: darken($color, 10%);
+    color: inherit;
   }
 }


andresgalante (changes requested):
Hi @m5o, this change might be breaking change to authors customizing the color of the alert links.

Even though it simplifies things a bit, it gives less control to author and I don't see that we gain anything by merging it.

Is there a good use case where you'd want to have it?

m5o:
@andresgalante – I just wanted to set awareness, that there exist these one-liners which adjust just a color nuance. An author can always setup his own `alert-[highlight]` styles with whatever color modifier he wants.

My point is, the `.alert-link` pattern isn't scaling, if we continue to define these utility class to "quickly provide matching colored links [within any alert]" to different components like e.g. Cards, Jumbotron, Nav,and Navbar.

UIkit, for example, has a `.[uk-]text-reset` / `.[uk-]link-reset` util class which reset `color` and `text-decoration`. This would be super handy and globally reusable in different components.

```css
a.uk-link-reset, a.uk-link-reset:hover, .uk-link-reset a, .uk-link-reset a:hover {
  color: inherit !important;
  text-decoration: none !important;
}
```

📖 https://getuikit.com/docs/link#link-reset

m5o:
@andresgalante 🔔

mdo:
Thanks @m5o, but we'll leave these links as they are for the darker contrast.

PR #25972, opened 2018-03-29
Ship v4.1

mdo:
- Updates the docs directory from `docs/4.0/` to `docs/4.1/`
- Runs the `build/ship.sh` script to updated version from `4.0.0` to `4.1.0`

We need to check some of this out though—I'm seeing some `-webkit` prefixes being dropped and I can't recall if those are intentional.


XhmikosR (dismissed):

js/src/index.js
```diff
@@ -31,7 +31,7 @@ import Util from './util'
   const maxMajor = 4
 
   if (version[0] < ltMajor && version[1] < minMinor || version[0] === minMajor && version[1] === minMinor && version[2] < minPatch || version[0] >= maxMajor) {
-    throw new Error('Bootstrap\'s JavaScript requires at least jQuery v1.9.1 but less than v4.0.0')
+    throw new Error('Bootstrap\'s JavaScript requires at least jQuery v1.9.1 but less than v4.1.0')
```
razh: This also doesn't need to be changed.

XhmikosR:
@mdo: after you review let me know so that I rebase and clean up the branch before the final merge.

vsn4ik (approved):

XhmikosR:
Obviously hence why I reverted it.

mdo:
@XhmikosR Fwiw, I'm guessing @vsn4ik's comment was from looking at a different commit or not the full range of the diff. It's admittedly all too easy on GitHub to click into a notification and see a subset of the overall diff :).

XhmikosR:
@mdo: when you are back to this,

```
git fetch --prune
git checkout ship-v4.1
git reset --hard
```

Then run the build script as normal. If you run the `release-version` script again, we'll need to review the changes again because it will change stuff that shouldn't change like in your first patch here.

mdo:
New PR coming in #26218.

PR #26178, opened 2018-04-01
Tiny grammar and punctuation tweaks to a comment.

rec:
1. It's == "it is"

2. The subject of the sentence is "Only one" (not "these") so the verb must agree with it.

Feel free to do nothing with this change, use it, etc - I just saw this while reading through the code.


XhmikosR (approved):

XhmikosR:
Thanks!

PR #26182, opened 2018-04-02
Update alert.js

wangxiaoxaio:
just a idiot try


mdo:
**Do not** test out Git or GitHub on public repos. You're notifying 7,000+ people every time you do this.

PR #26184, opened 2018-04-02
Feature request: stretched link

### LINKED ISSUE #18294: Link entire card v4

MartijnCuppens:
Fixes #18294

Stretched links can make whole cards or other elements clickable.  This PR is based on #25087, but can be used in multiple situations, not only cards.

Demo: https://deploy-preview-26184--twbs-bootstrap4.netlify.com/docs/4.1/utilities/stretched-link/

### DIFF
--- a/scss/utilities/_stretched-link.scss
+++ b/scss/utilities/_stretched-link.scss
@@ -1,0 +1,19 @@
+//
+// Stretched link
+//
+
+.stretched-link {
+  &::after {
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 1;
+    // Just in case `pointer-events: none` is set on a parent
+    pointer-events: auto;
+    content: "";
+    // IE10 bugfix, see https://stackoverflow.com/questions/16947967/ie10-hover-pseudo-class-doesnt-work-without-background-color
+    background-color: rgba(0, 0, 0, 0);
+  }
+}
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -11,6 +11,7 @@
 @import "utilities/screenreaders";
 @import "utilities/shadows";
 @import "utilities/sizing";
+@import "utilities/stretched-link";
 @import "utilities/spacing";
 @import "utilities/text";
 @import "utilities/visibility";
--- a/site/_data/nav.yml
+++ b/site/_data/nav.yml
@@ -70,6 +70,7 @@
     - title: Shadows
     - title: Sizing
     - title: Spacing
+    - title: Stretched link
     - title: Text
     - title: Vertical align
     - title: Visibility
--- a/site/docs/4.2/utilities/stretched-link.md
+++ b/site/docs/4.2/utilities/stretched-link.md
@@ -1,0 +1,100 @@
+---
+layout: docs
+title: Stretched link
+description: Easily make an element which contains a stretched link clickable.
+group: utilities
+---
+
+Adding the `.stretched-link` class to a link will make the [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block) of the `::after` pseudo element of the link clickable. In most cases, this means that an element with `position: relative;` which contains a link with the `.stretched-link` class is clickable.
+
+Cards are relative by default, so in this case you can safely add the `.stretched-link` class to a link in the card.
+
+{% capture example %}
+<div class="card" style="width: 18rem;">
+  {% include icons/placeholder.svg width="100%" height="180" class="card-img-top" text=" " title="Card image cap" %}
+  <div class="card-body">
+    <h5 class="card-title">Card with stretched link</h5>
+    <p class="card-text">Some quick example text to build on the card title and make up the bulk of the card's content.</p>
+    <a href="#" class="btn btn-primary stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Media objects are not relative by default, we need to add the `.position-relative` here to prevent the link from stretching outside the media object.
+
+{% capture example %}
+<div class="media position-relative">
+  {% include icons/placeholder.svg width="144" height="144" class="mr-3" text=" " title="Generic placeholder image" %}
+  <div class="media-body">
+    <h5 class="mt-0">Media with stretched link</h5>
+    <p>Cras sit amet nibh libero, in gravida nulla. Nulla vel metus scelerisque ante sollicitudin. Cras purus odio, vestibulum in vulputate at, tempus viverra turpis. Fusce condimentum nunc ac nisi vulputate fringilla. Donec lacinia congue felis in faucibus.</p>
+    <a href="#" class="stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Columns are relative by default, so if we want to stretch the link over a row, we need to add the `.position-static` class to column and add the `.position-relative` to the row.
+
+{% capture example %}
+<div class="row no-gutters bg-light position-relative">
+  <div class="col-md-6 mb-md-0 p-md-4">
+    {% include icons/placeholder.svg width="100%" height="200" class="w-100" text=" " title="Generic placeholder image" %}
+  </div>
+  <div class="col-md-6 position-static p-4 pl-md-0">
+    <h5 class="mt-0">Columns with stretched link</h5>
+    <p>Cras sit amet nibh libero, in gravida nulla. Nulla vel metus scelerisque ante sollicitudin. Cras purus odio, vestibulum in vulputate at, tempus viverra turpis. Fusce condimentum nunc ac nisi vulputate fringilla. Donec lacinia congue felis in faucibus.</p>
+    <a href="#" class="stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Multiple links can be added to elements with a stretched link, but their `z-index` must be increased to be become clickable.
+
+{% capture example %}
+<div class="card" style="width: 18rem;">
+  {% include icons/placeholder.svg width="100%" height="180" class="card-img-top" text=" " title="Card image cap" %}
+  <div class="card-body">
+    <h5 class="card-title">Card with multiple links</h5>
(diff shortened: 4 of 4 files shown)


docs/4.0/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+layout: docs
+title: Stretched link
+description: Easily make an element which contains a stretched link clickable.
+group: utilities
+toc: true
+---
+
+Adding the `.stretched-link` class to a link will make the [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block) of the `::after` pseudo element of the link clickable. In most cases, this means that an element with `position: relative;` which contains a link with the `.stretched-link` class is clickable.  
+
+Cards have are relative by default, so in this case you can safely add the `.stretched-link` class to a link in the card.
```
XhmikosR: small typo here

docs/4.0/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+    <p class="card-text">
+      <a href="#card-link-with-position-relative-and-increased-z-index" style="position: relative; z-index: 2;" class="text-success">This link will be clickable because <code>position: relative</code> is added and the z-index is higher than the z-index of the <code>::after</code> pseudo element of the stretched link</a>
+    </p>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+## Identifying the containing block
+If the stretched link doesn't seem to work, the [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#Identifying_the_containing_block) will probably be the cause. The following css-properties will make an element the containing block:
```
XhmikosR: Please add a newline after headers.

XhmikosR:
About the sidebar menu, you need to add it in data/nav.yml.

scss/utilities/_stretched-link.scss
```diff
@@ -11,5 +11,8 @@
     left: 0;
     z-index: 1;
     content: "";
+    // IE10 bugfix (see https://stackoverflow.com/questions/16947967/ie10-hover-pseudo-class-doesnt-work-without-background-color)
+    // rgba(0, 0, 0, 0) is converted to transparent by sass, so we use rgba(1, 1, 1, 0)
+    background-color: rgba(1, 1, 1, 0);
```
XhmikosR: We need to verify clean-css doesn't do this either. See https://github.com/twbs/bootstrap/pull/26400#issuecomment-385379463

MartijnCuppens: I just checked this with the latest Bootstrap build, it stayed `rgba(1,1,1,0)`

XhmikosR: No, I meant `rgba(0, 0, 0, 0)` -> `transparent`, because it's definitely one of the optimization clean-css applies. So, if we need this to be kept in more places, we'd need to verify and either disable the optimization or find another workaround.

MartijnCuppens: I merged with `v4-dev` today and it looks like `rgba(0, 0, 0, 0)` is not converted to `transparent` anymore (minified and not minified). Shall I change it to `rgba(0, 0, 0, 0)` then?

MartijnCuppens: Here you go:
```
index 8622f280e..53dbac68e 100644
--- a/dist/css/bootstrap.css
+++ b/dist/css/bootstrap.css
@@ -7424,6 +7424,17 @@ button.bg-dark:focus {
   max-height: 100% !important;
 }
 
+.stretched-link::after {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: 1;
+  content: "";
+  background-color: rgba(0, 0, 0, 0);
+}
+
 .m-0 {
   margin: 0 !important;
 }
```

XhmikosR:
@MartijnCuppens: I rebased your branch; it should fix the failures, but make sure you don't overwrite the changes.

florianlacreuse:
@MartijnCuppens Just to let you know that we can't select text in element containing a stretched link. Is this a known restriction?

MartijnCuppens:
@florianlacreuse, that's a known restricting. You can't select text in a link either (eg. see https://codepen.io/MartijnCuppens/pen/GwbWjp)

mdo:
I'd like to bring this back for v4.2. Looking at it though, I wonder if we should do more to discourage multiple links and modifying their `z-index`. For inline anchors, including the example you've included here, if I hover over the space between each line of text, the stretched link will be clicked. This feels like it'd backfire for users who aren't watching the hover styles close enough.

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
+---
+layout: docs
+title: Stretched link
+description: Easily make an element which contains a stretched link clickable.
```
mdo:
```suggestion
description: Make any HTML element or Bootstrap component clickable by "stretching" a nested link via CSS.
```

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+group: utilities
+---
+
+Adding the `.stretched-link` class to a link will make the [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block) of the `::after` pseudo element of the link clickable. In most cases, this means that an element with `position: relative;` which contains a link with the `.stretched-link` class is clickable.
```
mdo:
```suggestion
Add `.stretched-link` to a link to make it's [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block) clickable via a `::after` pseudo element. In most cases, this means that an element with `position: relative;` that contains a link with the `.stretched-link` class is clickable.
```

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+
+Cards are relative by default, so in this case you can safely add the `.stretched-link` class to a link in the card.
```
mdo:
```suggestion
Cards have `position: relative` by default in Bootstrap, so in this case you can safely add the `.stretched-link` class to a link in the card without any other HTML changes.
```

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+  <div class="card-body">
+    <h5 class="card-title">Card with stretched link</h5>
+    <p class="card-text">Some quick example text to build on the card title and make up the bulk of the card's content.</p>
+    <a href="#" class="btn btn-primary stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Media objects are not relative by default, we need to add the `.position-relative` here to prevent the link from stretching outside the media object.
```
mdo:
```suggestion
Media objects do not have `position: relative` by default, so we need to add the `.position-relative` here to prevent the link from stretching outside the media object.
```

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+  <div class="media-body">
+    <h5 class="mt-0">Media with stretched link</h5>
+    <p>Cras sit amet nibh libero, in gravida nulla. Nulla vel metus scelerisque ante sollicitudin. Cras purus odio, vestibulum in vulputate at, tempus viverra turpis. Fusce condimentum nunc ac nisi vulputate fringilla. Donec lacinia congue felis in faucibus.</p>
+    <a href="#" class="stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Columns are relative by default, so if we want to stretch the link over a row, we need to add the `.position-static` class to column and add the `.position-relative` to the row.
```
mdo:
```suggestion
Columns are `position: relative` by default, so clickable columns only require the `.stretched-link` class on a link. However, stretching a link over an entire `.row` requires `.position-static` on the column and `.position-relative` on the row.
```

site/docs/4.1/utilities/stretched-link.md
```diff
@@ -0,0 +1,100 @@
...
+  <div class="col-md-6 position-static p-4 pl-md-0">
+    <h5 class="mt-0">Columns with stretched link</h5>
+    <p>Cras sit amet nibh libero, in gravida nulla. Nulla vel metus scelerisque ante sollicitudin. Cras purus odio, vestibulum in vulputate at, tempus viverra turpis. Fusce condimentum nunc ac nisi vulputate fringilla. Donec lacinia congue felis in faucibus.</p>
+    <a href="#" class="stretched-link">Go somewhere</a>
+  </div>
+</div>
+{% endcapture %}
+{% include example.html content=example %}
+
+Multiple links can be added to elements with a stretched link, but their `z-index` must be increased to be become clickable.
```
mdo: Per my other comment, I feel like we should suggest folks avoid this approach because it could easily perpetuate incorrect clicks and user frustration.

mdo (commented):
Some copy suggestions here to be more specific.

MartijnCuppens:
Thanks for the copy changes, @mdo. Feel free to add an appropriate warning, I don't know if we should keep the example with the multiple links, because people might just copy past this bad example?

mdo:
Build is failing because this is stuck in the `4.1` docs folder. I'll pull the branch down and tackle it.

XhmikosR:
This was merged directly apparently. I don't agree with that solution since GH allows us to push to forks and pushing directly does **NOT** run the tests. Let alone you didn't squash.

XhmikosR:
`git rebase -i` is so easy, I don't get why you don't use it.

MartijnCuppens:
I agree, the `v4-dev` branch now contains some merge commits (like this one https://github.com/twbs/bootstrap/commit/de7a38250bf3c823c7b53a4f033237bad55d2627) which will make it unclear what the code is needed for.

XhmikosR:
@mdo: please disable merge commits for the whole repo. I'd also suggest that you disable direct pushes for admins too, no offense but you know :P

XhmikosR:
Joking aside, I don't know, you can leave the merge option although we don't really use it. I'm more concerned about the direct pushes.

PR #26192, opened 2018-04-03
Fix incorrect value of Stylelint option

ruedap:
https://stylelint.io/user-guide/rules/at-rule-empty-line-before/


XhmikosR:
What is wrong exactly? I have just set it to null to disable it, but the except part is in case we want to enable it.

XhmikosR:
@Johann-S: thoughts?

XhmikosR (approved):
I guess it doesn't hurt since the rule is disabled anyway.

PR #26194, opened 2018-04-03
Remove note about multiple inputs

jquense:
This note seems wrong given the example for multiple inputs further down?


andresgalante (approved):
Hi @jquense thanks for your contribution. You are right, it seems that this comment wasn't removed when Multiple inputs were added.

I'd remove the word "Also" on "Also remember" so we don't repeat it twice on the same paragraph.

PR #26218, opened 2018-04-08
Ship v4.1.0

mdo:
Supersedes #25972 which was still slightly off. **Here's what's changed for the release:**

- Moves docs directory from `docs/4.0/` to `docs/4.1/`. As a result, the v4.0.0 homepage no longer exists. Visiting `getbootstrap.com/docs/4.0/` will redirect right to `getbootstrap.com/docs/4.0/getting-started/introduction/`.

- Runs `build/ship.sh 4.1.0` to change version number on only our files.

- Modifies the docs page redirects as they're intended to be localized to each minor version (in most cases, we shouldn't redirect from v4.0 to v4.1 docs).

- Updates docs version dropdown with new release and link to v4.0 docs.

Alongside this, I've also prepped the compiled `docs/4.0/` directory in `gh-pages` with an updated release dropdown in the main nav. This adds the `docs/4.1/` link and dropdown divider to match our v4.1.0 release changes.

Still not as clean as I'd like—had to regenerate things after manually removing some changes caught by the build script. I'll be squashing on merge anyway, so should be fine. Here's how tomorrow morning will go for the release:

- [ ] Squash and merge this PR
- [ ] Publish the release with the highlights
- [ ] Push release to npm
- [ ] Wait for CDN to be updated
- [ ] Publish blog post
- [ ] Tweet tweet!

##

/cc @twbs/team


XhmikosR:
@mdo I pushed two small patches which you should squash too.:

1. Updated version in package-lock.json
2. Changed the navbar URL to v4.0.0 docs to the full URL; this URL doesn't exists locally. See how we do it with 3.3, 2.3.2 etc.

PR #26242, opened 2018-04-10
Suppress text-hide deprecation warning

### LINKED ISSUE #26237: text-hide() warning message
When building the new 4.1 scss source code, I get the following warning message:

```bash
alex@ubuntuVM:~/code/bootstrap-41/scss$ scss bootstrap.scss  bootstrap.css
WARNING: The `text-hide()` mixin has been deprecated as of v4.1.0. It will be removed entirely in v5.
         on line 10 of mixins/_text-hide.scss, in `text-hide'
         from line 57 of utilities/_text.scss
         from line 14 of _utilities.scss
         from line 41 of bootstrap.scss
```

It seems text-hide() ist still used in https://github.com/twbs/bootstrap/blob/v4.1.0/scss/utilities/_text.scss#L57

arechsteiner:
Adds an option to suppress the deprecation warning for the mixin `text-hide()` and sets the flag where the mixin is used internally. 

This avoids that the vanilla Bootstrap SCSS build spawns a warning by default, while maintaining the warning when `text-hide()` is used normally.

/CC: @XhmikosR 

Fixes #26237


XhmikosR:
Thanks for the PR!

Looks good to me, let's wait for a couple more approvals.

scss/mixins/_text-hide.scss
```diff
@@ -1,11 +1,13 @@
...
+@mixin text-hide($ignore-warning: false) {
   // stylelint-disable-next-line font-family-no-missing-generic-family-keyword
   font: 0/0 a;
   color: transparent;
   text-shadow: none;
   background-color: transparent;
   border: 0;
 
-  @warn "The `text-hide()` mixin has been deprecated as of v4.1.0. It will be removed entirely in v5.";
+  @if ($ignore-warning == false) {
```
XhmikosR: I'm thinking maybe checking against `true` would be better? Since that's what we use, and we can't know what the user might use.

XhmikosR:
@arechsteiner: also note that you can use `--quiet` or a similar option when you build and make the warnings go away.

scss/mixins/_text-hide.scss
XhmikosR: I always prefer to be as strict as possible.

mdo:
Building on your last comment here @XhmikosR, this is a build config choice. It's the first warning we've ever done for a deprecation, so I'm open to improving it, but this is a warning and not an error. Builds shouldn't fail because of it.

arechsteiner:
@mdo My builds based on Bootstrap don't fail because of this, they just throw warnings. From my perspective it's simply a bit odd that the Bootstrap core is warning me about stuff it is doing by itself (i.e. calling it's own deprecated mixin).

jleider:
@arechsteiner Thanks for this! This is the one thing holding us back from upgrading ourselves. 

@mdo If marking things as deprecated going forward is going to be more common, perhaps it would be best to add a config variable to the enable/disable section of https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L108. The default would be to show the warning intended for library maintainers and could still be disabled by library users.

jpickwell:
I don't have a problem with deprecation warnings. But, what I don't like is seeing warnings from just upgrading. I'm not using the `text-hide` mix-in directly, so I shouldn't see warnings. Suggesting the use of the `--quiet` CLI flag is not a good suggestion. When your own code uses something that's deprecated, it should not spit out anything.

mdo:
Alright, y'all convinced me. I'm down.

mdo (approved):
Let's also make it a follow-up issue please to document this optional mixin flag somewhere.

XhmikosR:
It's not supposed to be documented; that's the whole point of this change.
We still show the warning if one is using the mixin, but not when we use it
:)

XhmikosR:
Also, I want to change this to check for not true before merging.

ryancastle:
We would love to see this merged.

morgs32:
@mdo Did you ever make an issue to do something about an optional mixin flag? Maybe a global var I can set at the top to ignore warnings?
Meanwhile I'm using 4.1.0 and I see that the _text.scss file uses text-hide() even though it's deprecated?

iPazooki:
Awesome

PR #26255, opened 2018-04-11
Add dropdown responsive alignment

### LINKED ISSUE #25019: Responsive dropdown-menu-right class
It'd be great if we could have responsive control over the dropdown menu opening either left or right aligned (as opposed to writing custom CSS).

Here's a quick demo of what the following sass does:

kapture 2017-12-15 at 18 10 07

```scss
// since .dropdown-menu-right is not responsive
// we need to make sure that the language dropdown
// is right aligned on < md screen sizes
// <https://github.com/twbs/bootstrap/blob/34d745540d690086cdfc4ce648963f9acdc56bdb/scss/_navbar.scss#L160-L163>
#navbar-dropdown-language-ul {
  right: 0;
  left: auto; // Reset the default from `.dropdown-menu`
}
@include media-breakpoint-up(md) {
  #navbar-dropdown-language-ul {
    left: 0;
    right: auto; // Reset the default from `.dropdown-menu`
  }
}
```

Not sure if it should be called `dropdown-menu-right-$size-up` or `dropdown-menu-right-$size-down`?

Thoughts @mdo?

ysds:
Fixes #25019 

Add responsive alignment classes:

* Left to Right: `.dropdown-menu{-sm|-md|-lg|-xl}-right`
* Right to Left: `.dropdown-menu{-sm|-md|-lg|-xl}-left`

Basically Popper.js automatically adjust the position of dropdown menus, but its feature is disabled in navbars. These new responsive classes allow responsive alignment when use dropdown in navbars or disabling Popper.js manualy by `data-display="static"` option.

responsive-align

Demo: https://codepen.io/anon/pen/LdaxXq

Sorry I'm not a native English seaker, please let me know if the document is wrong.


andresgalante:
Hi @ysds , don't worry your english is pretty good.

I like this feature, I've build your branch locally and works as it should and the CSS is sound.

I have a question out of ignorance of how popper works. I can't understand why you need to change the position to static. From a pure CSS point of view you'd want the container to be relative so the absolute positioned menu has a point of reference.

So my questions are:
1- Do we need to make the container static?
1- If we do, can;'t we use the class `.position-static` instead of a data attribute?

cc: @Johann-S

Johann-S:
The `data-display` is used to enable or disable Popper.js on our Dropdown plugin, so I'm not sure where do you see a position static @andresgalante 🤔

ysds:
Yes, CSS `potision: static` can not be used in this case.

andresgalante:
I guess my question is, why do we need to make it static? is this a popper thing or a CSS thing?

ysds:
Umm... `data-display` is our option: 
https://getbootstrap.com/docs/4.1/components/dropdowns/#options

> By default, we use Popper.js for dynamic positioning. Disable this with `static`.

andresgalante (approved):
@ysds yeap, thanks for pointing that out.

My question was regarding CSS... I though data-display static was doing position static. Sorry I got confused.

Thanks for the patience with me :)

mdo (approved):

PR #26257, opened 2018-04-11
Replace reboot monospace hack with system font stack issue #26228

### LINKED ISSUE #26228: replace reboot/normalize monospace hack with monospace system font stack
There are two stanzas (selectors) for `code, kbd, pre, samp` in the main bootstrap distribution

The first in reboot:

```css
code, kbd, pre, samp {
  font-family: monospace,monospace;
  font-size: 1em;
}
```

And then later in bootstrap:

```css
code, kbd, pre, samp {
  font-family: SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace;
}
```

The proposal would be to move the system monospace font stack into reboot and get rid of the monospace, monospace hack (more details of which can be found here: https://github.com/necolas/normalize.css/issues/695 ).  I think this would make reboot more consistent (since sans-serif font uses system stack), be more clear, and make bootstrap smaller.

thoughts?

regards,

n

client9:
All test pass.  Spot check on my personal site seems to be fine.  Saves a whole 50 bytes minified, and 10 bytes compressed.  🔢 

I left the odd "font-size: 1em" since I'm unclear exactly what this does.  It's likely wrong (want rem) or un-necessarly or both, but this is enough change for this commit.

Comments welcome.

Fixes #26228

### DIFF
--- a/scss/_code.scss
+++ b/scss/_code.scss
@@ -1,11 +1,3 @@
-// Inline and block code styles
-code,
-kbd,
-pre,
-samp {
-  font-family: $font-family-monospace;
-}
-
 // Inline code
 code {
   font-size: $code-font-size;
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -232,15 +232,13 @@ a:not([href]):not([tabindex]) {
 // Code
 //
 
-// stylelint-disable font-family-no-duplicate-names
 pre,
 code,
 kbd,
 samp {
-  font-family: monospace, monospace; // Correct the inheritance and scaling of font size in all browsers.
+  font-family: $font-family-monospace;
   font-size: 1em; // Correct the odd `em` font sizing in all browsers.
 }
-// stylelint-enable font-family-no-duplicate-names
 
 pre {
   // Remove browser default top margin


XhmikosR:
Only issue I see with this, is that the standalone reboot won't have this fix any longer. And it seems it's still required by some browsers.

Alternatively, maybe we could do this conditionally.

I'll leave the final decision to @mdo.

client9:
hi @XhmikosR  -- my understanding of the monospace hack is that some browser/os combos have a hardwired "if and only if the font-family is exactly "monospace" then do something stupid with sizing".   This is why "monospace, monospace" prevents the issue.   Using something other than `monospace` such as the system stack will also work correctly.  Does this make sense? Or did I misunderstand your concern?  regards, --nickg

client9:
hi @XhmikosR Oh i see.  For bootstrap is this a non-change (except smaller, less redundant rules), but for reboot it would change the font used (assuming it's not changed downstream).   So that would be a @mdo decision on what to do.  got it!  thanks for your time guys.  Let me know if you need any additional details.  regards --nickg

mdo (approved):

PR #26273, opened 2018-04-11
fix `.modal-dialog-centered` on IE10/11

### LINKED ISSUE #25569: Class modal-dialog-centered does not work on IE 11
This might an issue with Flexbox and IE 11.  In any case, opening up the modal does not vertically center it.

Easy to Reproduce:  In IE11, go to https://getbootstrap.com/docs/4.0/components/modal/  and click on the "Launch demo modal" button in the section "Vertically centered".

IF there is a workaround class or code, please advise.  Thanks.

MohammadYounes:
Fixes #25569

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -54,6 +54,13 @@
   display: flex;
   align-items: center;
   min-height: calc(100% - (#{$modal-dialog-margin} * 2));
+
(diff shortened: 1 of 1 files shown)


ysds:
This fixes is the best workaround as far as I can see.

Demo: https://codepen.io/anon/pen/EEJKZj

xcartmods:
Thanks for this, all good now in Win7/IE11

PR #26278, opened 2018-04-12
HTTPS to more links

Swardu:
Added HTTPS to sass-lang.com. While I was at it, I also added HTTPS to example.com because why not.


patrickhlauke:
lovely, thanks

patrickhlauke (approved):

Swardu:
It looks like I'm the 1000th contributor 😊

PR #26286, opened 2018-04-13
Updated modal component documentation

sunpietro:
Added information about existing modal events that weren't exposed in the documentation.
Added information about:
- focusin.bs.modal,
- resize.bs.modal,
- click.dismiss.bs.modal,
- keydown.dismiss.bs.modal.

### DIFF
--- a/docs/4.1/components/modal.md
+++ b/docs/4.1/components/modal.md
@@ -665,6 +665,22 @@ Bootstrap's modal class exposes a few events for hooking into modal functionalit
       <td>hidden.bs.modal</td>
       <td>This event is fired when the modal has finished being hidden from the user (will wait for CSS transitions to complete).</td>
     </tr>
+    <tr>
+      <td>focusin.bs.modal</td>
+      <td>This event is fired when a modal gets focus inside.</td>
+    </tr>
+    <tr>
+      <td>resize.bs.modal</td>
+      <td>This event is fired when a modal is resized.</td>
+    </tr>
+    <tr>
+      <td>click.dismiss.bs.modal</td>
+      <td>This event is fired when a modal is closed by clicking on a backdrop.</td>
+    </tr>
+    <tr>
+      <td>keydown.dismiss.bs.modal</td>
+      <td>This event is fired when a modal is closed by pressing the Escape key on a keyboard.</td>
+    </tr>
   </tbody>
 </table>
 


Johann-S:
Sorry but we only expose events which are triggered by our plugins, not the used by our plugins

sunpietro:
But having information about these events is also very important in terms of detection whether some specific action happened, like pressing **Esc** key. Without digging into the plugin code, I wouldn't be able to detect such action the right way.
I'm very disappointed.

sunpietro:
@Johann-S of course it can be detected that way, but use case is like following:

I'm building React component that reuses Bootstrap4 features, that are already loaded on a website.
I'm opening a popup on demand clicking on a specific button. When I want to close it, I'm clicking on either **X** button or **Cancel** button. 
For these buttons `click` event I've attached an event handler, that performs some actions and at the end I'm calling `$(element).modal('hide');`. But when **Esc** key is pressed then I have no controlled way to detect that the popup was closed using this specific event: `keyup` and that this actions closes a popup.
The thing is that popup has a dynamic content, that is reloaded everytime it's opened.
When someone chooses to close the popup with **Esc** key press, then I'm not able to update the state of component without rendering it twice when a popup is displayed again.

Now I'm listening to `keydown.dismiss.bs.modal` event with jQuery `once` method and it updates the component state correctly.

In short, it's all about unexpected visual state changes, when you try to combine Bootstrap4 and ReactJS in the same component, that are not reflected in the component state.

Johann-S:
I'm a bit mismatched with your use case, because from the beginning of Bootstrap we just show the events we triggered, and you're free to add listeners on your own without using those added by Bootstrap... 
If we accept that PR, we'll have to edit all of our plugins documentation 🤔 

@XhmikosR or @bardiharborow any opinion on that ?

patrickhlauke:
my personal take: this comes down to our plugins having, to a certain extent, "private" variables/functions/events. of course they're not private in the sense that yes they CAN be seen/intercepted, but are not part of the official interface of those plugins/components.

i suspect that for the majority of users, they'll just want to use bootstrap and its components out of the box. once you get to the point where you want to use bootstrap, but hijack/modify the behavior of its core components...then yeah you'll have to dig deeper than what our documentation documents.

sunpietro:
@patrickhlauke It's more like reacting on some specific events, when the ones exposed _officially_ are not enough in some cases or they come too late.

XhmikosR:
This is a complicated matter and both parties have valid points.

Perhaps for this specific case you should reconsider @Johann-S? Can this use case be covered by the existent methods we document?

Johann-S:
Nope but this use case can be handled by simple events listeners: 
```js
$('#modal').on('keydown', (event) => {
  if (event.which === 27) { // escape
    // Your code
  }
})
```

sunpietro:
I'll try to find another way or I'll keep listening to so called private
events. Thanks for sharing your opinions.

pt., 13 kwi 2018, 20:50 użytkownik XhmikosR <notifications@github.com>
napisał:

Issue #26266, opened 2018-04-11
Collapses not working on IE11


adrigm:
SCRIPT5007: Object.keys: the argument is not an object.

![dd](https://user-images.githubusercontent.com/246042/38625125-17579d04-3daa-11e8-89f4-78d786daaa4f.PNG)

razh:
This is because Babel improved its compliance with spec behavior:

https://github.com/babel/babel/pull/7034

And the `_objectSpread` helper function uses `Object.keys`.

https://github.com/babel/babel/blob/v7.0.0-beta.44/packages/babel-helpers/src/helpers.js#L373-L391

To resolve this, `loose` mode could be used:

https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-object-rest-spread#loose

Herst:
Is this not what Babel is supposed to prevent? :roll_eyes: I thought the idea was: You just write your code newest features and we do the rest to make it work in older browsers or warn you if it is not possible… 

Anyway, alternatively users targeting browsers without `Object.keys` support could be asked to add a shim. What else would need to be shimmed in order to support IE11? `Object.assign` perhaps? In that case this note would have to be added to the general browser support page and not just the sub pages of the affected features.

razh:
Wait a minute, [IE 9 and above support `Object.keys`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Browser_compatibility
), but it appears their implementation is the ES5 version.

> In ES5, if the argument to this method is not an object (a primitive), then it will cause a `TypeError`. In ES2015, a non-object argument will be coerced to an object.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Notes

This error is occurring because of this line:

https://github.com/twbs/bootstrap/blob/8f7bd419935adfcd53c471a0202083464800619e/js/src/collapse.js#L334

If `config` is not an object, the value is `false`. In IE, `Object.keys(false)` results in an error, while modern browsers return `[]`.

japrogramer:
IE .. always slowing developers down .. it's just  a shame thats all.

PR #26291, opened 2018-04-13
fix issue related to Object.keys and Dropdown issue

Johann-S:
Fixes: #26266

/CC @XhmikosR

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -331,7 +331,7 @@ const Collapse = (($) => {
         const _config = {
           ...Default,
           ...$this.data(),
-          ...typeof config === 'object' && config
+          ...typeof config === 'object' && config ? config : {}
         }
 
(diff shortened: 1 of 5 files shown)


js/src/collapse.js
```diff
@@ -331,7 +331,7 @@ const Collapse = (($) => {
         const _config = {
           ...Default,
           ...$this.data(),
-          ...typeof config === 'object' && config
+          ...typeof config === 'object' && config ? config : {}
```
vsn4ik:
```js
typeof config === 'object' ? config : {}
```
?

Johann-S: yep that can be enough thanks 👍

Johann-S: Finally I won't make this change because it'll decrease our code coverage

XhmikosR (approved):

PR #26292, opened 2018-04-13
Resolve #26262: Add deprecated warning

### LINKED ISSUE #26262: Remove .text-hide mixin from documentation or mark as deprecated
Shouldn't be this page http://getbootstrap.com/docs/4.1/utilities/image-replacement/ removed or marked as deprecated? I think it could be confusing for new users.

Varunram:
I could modify the text if needed, dunno what other content should be in here apart from "This class is deprecated".

Fixes #26262.

### DIFF
--- a/docs/4.1/utilities/image-replacement.md
+++ b/docs/4.1/utilities/image-replacement.md
@@ -8,7 +8,7 @@ toc: true
 
 {% capture callout %}
 ##### Warning
-This class has been deprecated
+This component is deprecated and will be removed in a future version of bootstrap
 {% endcapture %}
 {% include callout.html content=callout type="warning" %}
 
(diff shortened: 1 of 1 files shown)


wolfy1339:
You will want to add a `@warn This component is deprecated and will be removed in a future version of bootstrap`

XhmikosR:
@wolfy1339: it's already there. https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_text-hide.scss#L10

Although arguably we shouldn't show it like we do, see #26237 and #26242

mdo:
@XhmikosR I think @wolfy1339 means the text should match in the docs to explain why/when/etc, which I agree with. @Varunram, can you make that copy edit to match the Sass deprecation warning?

Varunram:
@mdo Done, thanks!

XhmikosR (approved):

PR #26302, opened 2018-04-15
Don't zero container padding with navbar-expand

### LINKED ISSUE #26301: bare navbar-expand incorrectly removes padding for all widths
There's a bug in the loop that generates `navbar-expand`, resulting in removing container padding for all viewport widths. Here is the live jsbin: https://jsbin.com/zugezewati/edit?html,output

I understand the purpose of removing padding. This report is not about that. Rather there's a specific bug in the scss loop that causes `navbar-expand` to remove padding where it shouldn't, while `navbar-expand-sm` and similar work as intended.

How it's [intended to work](https://getbootstrap.com/docs/4.1/components/navbar/#containers):

> When the container is within your navbar, its horizontal padding is removed at breakpoints lower than your specified .navbar-expand{-sm|-md|-lg|-xl} class. This ensures we’re not doubling up on padding unnecessarily on lower viewports when your navbar is collapsed.

The `navbar-expand` class is synonymous with the fictional `navbar-expand-xs`, meaning that it should remove horizontal padding at breakpoints lower than 0px. Of course there are no such breakpoints, so it shouldn't remove padding anywhere. Instead it removes padding everywhere, as you can see in the jsbin linked above.

The bug is in the loop [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_navbar.scss#L138-L152)

```scss
// Generate series of `.navbar-expand-*` responsive classes for configuring
// where your navbar collapses.
.navbar-expand {
    $next: breakpoint-next($breakpoint, $grid-breakpoints);
    $infix: breakpoint-infix($next, $grid-breakpoints);

    &#{$infix} {
        > .container,
        > .container-fluid {
          padding-right: 0;
          padding-left: 0;
        }
      }
```

In this loop, there is a relationship: `$next` should always be the
breakpoint immediately larger than `$breakpoint`

However this breaks (no pun intended) when `$breakpoint = "xl"` -- in that case there is no larger breakpoint, so `$next = null` and `$infix = ""`

The result is this unintended css:

```css
.navbar-expand > .container,
.navbar-expand > .container-fluid {
    padding-right: 0px;
    padding-left: 0px;
}
```

There are some similar reports, but #22471 is closed and I'm not sure #24726 is exactly the same.

agriffis:
Fixes #26301

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -143,11 +143,18 @@
     $infix: breakpoint-infix($next, $grid-breakpoints);
 
     &#{$infix} {
-      @include media-breakpoint-down($breakpoint) {
-        > .container,
-        > .container-fluid {
-          padding-right: 0;
-          padding-left: 0;
+      // On the final pass through the loop,
+      // $breakpoint = xl, $next = null and $infix = "",
+      // effectively wrapping around from xl to xs. This final pass
+      // generates rules for bare navbar-expand, and we shouldn't zero
+      // container padding since there are no smaller breakpoints.
+      @if $next {
+        @include media-breakpoint-down($breakpoint) {
+          > .container,
+          > .container-fluid {
+            padding-right: 0;
+            padding-left: 0;
+          }
         }
       }
 


agriffis:
This is not a complete fix. In fact, this is impossible to fix properly with the current code, because the problematic padding is on navbar not container, but we can't conditionalize the navbar padding on "has a container inside it."

PR #26305, opened 2018-04-15
Add docs/3 and docs/4 pages that link to the correct version

Kwpolska:
This adds two more redirections that make it slightly easier to get to a page with just link editing (one wouldn’t need to remember that 3.3 and 4.1 are the latest versions in their respective branches).

(I hope I’m not the only person who would do link editing like this.)

### DIFF
--- a/docs/3/index.html
+++ b/docs/3/index.html
@@ -1,0 +1,38 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


mdo:
`https://getbootstrap.com/docs/` automatically redirects to the latest stable version, and we'll ensure that it always does. I don't see any point in adding another redirect for v3, even with a v3.4.0 being proposed.

PR #26316, opened 2018-04-16
Doc update - remove media-breakpoint-up(xs)

DavidDeSloovere:
If media-breakpoint-down(xl) is not listed, then media-breakpoint-up(xs) does not need to be listed either.
The above 4 media queries are now aligned with the 4 media-breakpoint-up mixins (just like their media-breakpoint-down counterparts).

Also added example for media-breakpoint-down.

This PR is a result of questions from a colleague that misinterpreted the existing docs. This should clarify it.


docs/4.1/layout/overview.md
```diff
@@ -70,13 +70,12 @@ Bootstrap primarily uses the following media query ranges—or breakpoints—in
 Since we write our source CSS in Sass, all our media queries are available via Sass mixins:
 
 {% highlight scss %}
-@include media-breakpoint-up(xs) { ... }
```
mdo: Let's actually swap this with something that explicitly states there's no `xs` query?

```scss
// No media query necessary for `xs` breakpoint as it's effectively `@media (min-width: 0) { ... }
```

docs/4.1/layout/overview.md
```diff
@@ -70,13 +70,12 @@ Bootstrap primarily uses the following media query ranges—or breakpoints—in
...
 @include media-breakpoint-up(sm) { ... }
 @include media-breakpoint-up(md) { ... }
 @include media-breakpoint-up(lg) { ... }
 @include media-breakpoint-up(xl) { ... }
```
mdo: Let's insert an example then to back up that `xs` comment I suggested earlier.

docs/4.1/layout/overview.md
DavidDeSloovere: Good point. People might think 'where is xs'.

docs/4.1/layout/overview.md
DavidDeSloovere: Added example, hope I interpreted the comment correctly.

DavidDeSloovere:
I don't know why the build is failing for just a docs update.

wolfy1339:
Maybe you should check. Because even docs have linting and builds too 😃 
Anyways, here's the error `docs/4.1/layout/overview/index.html: line 463, col 1, indenting spaces must be used in groups of 2`

DavidDeSloovere:
🤦‍♂️  Well at least I learned something.
I thought was related to the branch being out of date and not being able to update. Was more other way around.

PR #26322, opened 2018-04-17
Update devDependencies.

XhmikosR:
We should do `npm run dist` locally and make sure the bundle works with `npm run docs-github && npm run docs-github-serve` with IE 10/11.


Johann-S:
or just run `js/tests/index.html` on your browser 😉 (after you ran `npm run dist`)

Johann-S (approved):
LGTM 👍 and it works on IE 10 and 11

PR #26328, opened 2018-04-17
fix issues with calc

### LINKED ISSUE #25717: Height property of ::after pseudo-element of custom-file-label class is an error in IE11.

### LINKED ISSUE #26206: Consider rem-calc() instead of calc()
In some cases, calc() thing breaks CSS: if i set $card-border-width to 0, compiled CSS for card header top border radius will look like calc(0.25rem - 0), which is incorrect. Same happens if $input-height-border variable is set to 0.
We either need to document $border-width and other px values used in calc() function properly (value greater than 0 is required) or implement custom rem-calc() function, which will convert all px units to rem and get rid of calc() where possible.

MartijnCuppens:
Fixes: #26206.

[postcss-calc](https://github.com/postcss/postcss-calc) uses [reduce-css-calc](https://github.com/MoOx/reduce-css-calc) to reduce CSS `calc()` function. This also removes zero-values thus fixes #26206.

This also seems to solve the issue of nested calc functions. (Fixes #25717)

### DIFF
--- a/build/postcss.config.js
+++ b/build/postcss.config.js
@@ -9,6 +9,7 @@ module.exports = (ctx) => ({
   plugins: {
     autoprefixer: {
       cascade: false
-    }
+    },
+    'postcss-calc': {}
   }
 })
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "nodemon": "^1.17.3",
     "npm-run-all": "^4.1.2",
     "popper.js": "^1.14.3",
+    "postcss-calc": "^6.0.1",
     "postcss-cli": "^5.0.0",
     "qunitjs": "^2.4.1",
     "rollup": "^0.58.0",


XhmikosR:
We've already tried this approach and @mdo didn't want to use postcss back then.

mdo:
Deferring to #25724

PR #26329, opened 2018-04-17
Inaccurate note about color utils

travisdahl:
The docs say:
 **Note that the `.text-white` and `.text-muted` class has no link styling.** 
but that isn't true, they still have link styling. Either the docs are wrong, or the functionality is wrong. If it's the docs, here is a PR. :)

Hovered over the muted link...
image

Hovered over the white link...
image

### DIFF
--- a/docs/4.1/utilities/colors.md
+++ b/docs/4.1/utilities/colors.md
@@ -19,7 +19,7 @@ toc: true
 {% endcapture %}
 {% include example.html content=example %}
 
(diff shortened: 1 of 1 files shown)


razh:
This is in reference to the fact that they do not use the `text-emphasis-variant` mixin, which adds a hover-color.

https://github.com/twbs/bootstrap/blob/8f7bd419935adfcd53c471a0202083464800619e/scss/utilities/_text.scss#L42-L49

travisdahl:
@razh ah. I guess its just a bit unclear then. I consider the underline "link styling". You can close if its a non-issue

PR #26338, opened 2018-04-18
Detaches button padding-x from inputs on input group btns

### LINKED ISSUE #26334: Buttons in input group have incorrect padding
Large and small buttons in input group extend form controls, not buttons. So if i want to change horizontal padding for input fields, buttons in 2 sizes adjust paddings accordingly. Is this an expected behaviour or a bug?

andresgalante:
This PR fixes #26334

When you have a btn on an input group the padding x is attached to the input padding, making it impossible to change the padding of the input without affecting the btn.

This PR detaches it.

@mdo same happens with `.input-group-text` do you think we should have a variable to control that too?


mdo:
Do we need to review this after the last input group updates? We no longer `@extend` those styles.

tangchau:
good！

blackc03r:
Yes

Issue #26311, opened 2018-04-16
Dashboard example, Diagram title & buttons, xs & sm devices


PerseusTheGreat:
Hi,
In **Dashboard example**, there are a title text & 3 buttons top of the diagram (canvas).

![1](https://user-images.githubusercontent.com/5496125/38796837-e419e444-4171-11e8-9fc3-df47745a299e.png)

When screen size is less than 768px (xs & sm devices), **Nav bar** will covers (hides) those elements.

![2](https://user-images.githubusercontent.com/5496125/38796959-4e119888-4172-11e8-808a-a12e3590a519.png)

I have added below codes to **dashboard.css**, line 72, to fix the issue.
```
  @media (max-width: 768px) {
      [role="main"] {
          padding-top: 133px; /* Height of navbar structure (fixed navbar + search input + sign out link) */
      }
  }
```

![3](https://user-images.githubusercontent.com/5496125/38797079-a221281c-4172-11e8-9612-3119f0c864b2.png)

So, please consider.
Thanks in advance.

mdo:
Browser and OS versions?

PR #26345, opened 2018-04-19
fix dashboard issue in responsive #26311

chiraggmodi:
fixed issue #26311


XhmikosR (changes requested):
Please revert the unrelated changes.

XhmikosR:
Thanks but #26332 was already open it seems.

PR #26348, opened 2018-04-19
Add workaround for not respected Navbar-nav list type on Edge

### LINKED ISSUE #26287: Navbar-nav list type not being respected on refresh in Edge
So if you load up this bootstrap 4 example in Edge with the browser window in md or above

https://getbootstrap.com/docs/4.1/examples/sticky-footer-navbar

Shrink the window so you see the navbar-toggeler and expand.

Refresh the window and expand again.

You'll see the that list-style: none; isn't being respected.

example

i've checked this in chrome, firefox and IE11 so this looks to be a edge only bug.

fatmaby:
Hi, This is a pull request fixes #26287


Johann-S:
You have to fix this lint error: https://travis-ci.org/twbs/bootstrap/jobs/368902835#L591

MartijnCuppens:
It looks like this bug was fixed in the latest version of Edge (Edge 18).

XhmikosR:
Yeah but IE I doubt it :)

MartijnCuppens:
This was only an issue in Edge, IE looks fine, I've updated the title.

XhmikosR:
https://github.com/twbs/bootstrap/issues/26287#issuecomment-381372876

MartijnCuppens:
I can't reproduce this in IE though

XhmikosR:
Not sure then. We do support older Edge versions so idk.

patrickhlauke:
confirming that i can't repro the problem in IE11 on the specific https://getbootstrap.com/docs/4.1/examples/sticky-footer-navbar case. 

https://github.com/twbs/bootstrap/issues/26287#issuecomment-381372876 points to a stackoverflow issue around suckerfish menus, and second-level menus ... which likely wouldn't affect bootstrap.

patrickhlauke:
> Not sure then. We do support older Edge versions so idk.

the wording in our browser support prose is a bit vague

> Bootstrap supports the latest, stable releases of all major browsers and platforms. On Windows, we support Internet Explorer 10-11 / Microsoft Edge

you could read that as meaning the latest stable release of Microsoft Edge (17, currently, which still has the problem though)

XhmikosR:
So... what's the verdict on this?

patrickhlauke:
Seems this bug is fixed now in Edge. Suggest closing this as (now) unnecessary.

PR #26349, opened 2018-04-20
Optimised Google Analytics Loading

coliff:
The Google Analytics script currently on getbootstrap.com is loaded and executed asynchronously on all browsers but has the disadvantage of not allowing modern browsers to preload the script.

The alternative async tracking snippet in this PR adds support for preloading, which will provide a small performance boost on modern browsers.

This script implementation is provided and supported by Google:
https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

(Very old browsers that do not recognize the `async` script attribute will  degrade to synchronous loading and execution - [but none of those browsers are supported by Bootstrap 4 anyway](https://caniuse.com/#search=async))

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -30,10 +30,8 @@
 {% include social.html %}
 
 <script>
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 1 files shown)


XhmikosR:
/CC @mdo. Personally I wouldn't change it, well, because, I see no reason yet.

mdo:
Cherry-picked into 26732. Thanks!

PR #24671, opened 2017-11-04
Remove precss plugin from webpack.md documentation

pmendelski:
AFAIK `precss` is not needed. Moreover I have tested it in a [side proejct](https://github.com/mendlik/webpack-starter/tree/master/webpack-starter-bootstrap-spa) and it broke some styling. 

For example when you add `precss` to [plugins](https://github.com/mendlik/webpack-starter/blob/7f32bfdca59aa0efa43e52d855a37f6e518edc2d/webpack-starter-bootstrap-spa/lib/webpack/config/styles.js#L43) the there is no `:root` styling in the result css file.

Fixes: #24949

PR #26357, opened 2018-04-22
Precss not needed on minimal setup

lmuzquiz:
Precss not needed on minimal setup


XhmikosR:
Thanks but we already have #24671.

PR #26383, opened 2018-04-25
change test runner

browner12:
I can't find grunt anywhere in the dependencies anymore. I believe this should be updated to NPM, correct?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -259,7 +259,7 @@ includes code changes) and under the terms of the
 
 ### Checking coding style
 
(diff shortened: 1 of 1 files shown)


wolfy1339:
Your patch is for the wrong branch.,master is v3. v3 uses Grunt, v4 uses NPM

Johann-S:
What @wolfy1339 said 👍

browner12:
I think we need to update the links on https://github.com/twbs/bootstrap/blob/v4-dev/README.md then because they are linking the master branch, which can be confusing for users.

PR #26385, opened 2018-04-26
Fix custom range thumb style

### LINKED ISSUE #26379: Custom range focus style issue
There is a bug in the focus style of the custom range input. Related on 6c620de01fdbfed50c34849d3d17d601d0bc48d3

https://github.com/twbs/bootstrap/blob/8f7bd419935adfcd53c471a0202083464800619e/dist/css/bootstrap.css#L3740-L3743

It seems that the focus shadow is defined, but this does not apply:
image

Demo: http://getbootstrap.com/docs/4.1/components/forms/#range

The `:focus` selector must be put after the  `.custom-range`: Also, `outline: none;` before the `box-shadow` is unnecessary.

Therefore, it can be fixed as follows:
```css
.custom-range:focus::-webkit-slider-thumb {
   box-shadow: 0 0 0 1px #fff, 0 0 0 0.2rem rgba(0, 123, 255, 0.25);
}
```

This fixes add focus shadow, but IE and Edge have new issue of cutting focus shadow.

Chrome:
image

IE11/Edge
image

Demo: https://codepen.io/anon/pen/BxzyNB

So we should:
+ Remove the wrong focus shadow CSS? (Current style, but has accessibility issue)
+ Fix the focus shadow CSS as above? (IE/Edge has issue)
+ Add an alternative focus style?

ysds:
Fixes #26379 and thumb positions.

Demo: https://codepen.io/anon/pen/GdjKop

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -378,8 +378,8 @@
     width: $custom-range-thumb-width;
     height: $custom-range-thumb-height;
     margin-top: 0; // Edge specific
-    margin-right: $custom-range-thumb-focus-box-shadow-total-width; // Workaround that overflowed box-shadow is hidden.
-    margin-left: $custom-range-thumb-focus-box-shadow-total-width;  // Workaround that overflowed box-shadow is hidden.
+    margin-right: $custom-range-thumb-focus-box-shadow-width; // Workaround that overflowed box-shadow is hidden.
+    margin-left: $custom-range-thumb-focus-box-shadow-width;  // Workaround that overflowed box-shadow is hidden.
     @include gradient-bg($custom-range-thumb-bg);
     border: $custom-range-thumb-border;
     @include border-radius($custom-range-thumb-border-radius);
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -326,7 +326,7 @@
   &::-webkit-slider-thumb {
     width: $custom-range-thumb-width;
     height: $custom-range-thumb-height;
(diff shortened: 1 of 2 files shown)


mdo (approved):

scss/_custom-forms.scss
```diff
@@ -311,6 +311,12 @@
 
   &:focus {
     outline: none;
+
+    // Pseudo-elements must be split across multiple rulesets to have an affect.
+    // No box-shadow() mixin for focus accessibility.
+    &::-webkit-slider-thumb { box-shadow: $custom-range-thumb-focus-box-shadow; }
```
XhmikosR: @ysds Doesn't autoprefixer take care of the prefixes?

mdo: Not in the selectors, no. All these need to be split as the comment says or no browser picks it up. Strange, but it's common across most of these form styles.

ysds: As @mdo said. Also I found @ai's comment in https://github.com/postcss/autoprefixer/issues/279

> Main problem here, that we absolutly didn’t know unprefixed version of this selectors. Autoprefixer is developed to be able to remove Autoprefixer and future and your CSS will be worked.

> Try PostCSS and create polyfill.

ai: Autoprefixer can add prefixes in selectors (like it support `:: placeholder`). But Autoprefixer need to have W3C draft to have unprefixed version.

XhmikosR: Yeah, I was mostly wondering :)

PR #26400, opened 2018-04-27
Update devDependencies.

XhmikosR:
@Johann-S: this is after 4.1.1 is released.

We need to make sure nothing is broken due to the babel updates; they changed some things regarding the config files, so we need to read the release notes.


Johann-S:
This is the change which lead our unit test to fail: 
image

XhmikosR:
OK, so I added a couple more updates. Notes:

1. in reboot we have this change in the non-minified CSS

```diff
-  -webkit-tap-highlight-color: transparent;
+  -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```

But the minifier changes this to `transparent` already. @mdo not sure if this reverts the effect you want

2. We need to fix Babel and perhaps update to the new config format?
3. sinon is updated too
4. karma-qunit is still the old version

XhmikosR:
Oh, and 4. I had to disable a few rules for htmllint. IMO we should drop htmllint completely unless we plan to enforce class/id styles and stuff.

XhmikosR:
@Johann-S: I reverted the babel updates and added karma-qunit. Tests still fail though.

@mdo: https://github.com/twbs/bootstrap/pull/26400#issuecomment-385379463

Johann-S:
We should delete this unit test: https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/tooltip.js#L525

Because it's not our job to be sure tooltips are correctly positionned, it's Popper.js's job

XhmikosR:
You know better the popper stuff 👍 

PS. I don't know why your modification to package-lock.json had so many unrelated changes... I just did `npm i` using node 10.x and npm 5.6.0.

XhmikosR:
@mdo: can you check my comment above? We might need to change clean-css' settings if we need it to keep this.

XhmikosR:
@Johann-S: do you think you could look into updating babel too?

XhmikosR:
@Johann-S: ping ^^

Johann-S:
I don't have time to work on Babel upgrade sorry 😟

XhmikosR:
All right, then please approve so that we merge this as is. You can tackle the Babel upgrade later, or maybe @hzoo can help us out with that.

Johann-S (approved):

hzoo:
@Johann-S all good, this like this broke IE and we have a PR to fix in https://github.com/babel/babel/pull/7969

hzoo:
Depends on how it's used; if you mean `babel.config.js` it's mostly for monorepo/compiling node_modules setup

XhmikosR:
@hzoo: yeah, I remember I read about that file in the release notes. So, does that mean we don't need to change something for our case, right?

Issue #26419, opened 2018-04-30
Performance with large number of Collapse in v4.1


chtheis:
I have a page with 6 collapsibles on it, each one with more than 20 in them, so I end up with about 140 collapsibles in total. The idea is to click on one of the outside collapsibles to see an overview list of collapsible items, click on one of the inner ones to see details of that item. 

Unfortunately the performance is very bad: to open one the first time takes about 10 seconds!

I dug down to the constructor of the class Collapse:
```js
        var tabToggles = $$$1(Selector.DATA_TOGGLE);

        for (var i = 0; i < tabToggles.length; i++) {
          var elem = tabToggles[i];
          var selector = Util.getSelectorFromElement(elem);

          if (selector !== null && $$$1(selector).filter(element).length > 0) {
            this._selector = selector;

            this._triggerArray.push(elem);
          }
        }
```
Nearly all time is spent in the for loop.

`tabToggles` is an array of all 140 elements with `[data-target="collapse"]`.

First question: why do I need all of them when I click on one? 

Then for each one it will get the selector and filter the list of all matching elements filter them for the element parameter. And that is where all the time is spent, almost equally distributed between `getSelectorFromElement` and `$$$1(selector).filter(element)`

`getSelectorFromElement` will select a list of matching elements and return either the selector, if some elements where found, or null if not. And the next statement in the code will do the same, then apply the filter. 

Second question: Could we save time if we had a function like `getSelectorFromElement` which could do the filter(element), too ?
I tried to pass the element to `getSelectorFromElement` and, if present, do the filtering there. This cut the time in half. But still, very slow.

Johann-S:
Can you provide to us a CodePen ? 

And you can try that : 
```js
      var tabToggles = $.makeArray(document.querySelectorAll(Selector.DATA_TOGGLE))
        .filter(function (elem) { return !!elem.getAttribute('data-target') })

chtheis:
I'm trying to cut it down on JSfiddle.
So far the (almost) original data are at 
https://jsfiddle.net/chtheis/8mxntvLj/

An edited version, where I try to simulate number of collapsibles and elements is at 
https://jsfiddle.net/chtheis/zou2st3e/1/

On the original, click on a time and be patient. A table will open a table with around 20 matches, each match is a <tbody>.
Click on a match and after some match details are shown,

I have other data where the matches don't have details and there opening a time is fast(er).

[...]

PR #26422, opened 2018-04-30
refactor(plugins): improve how we query elements

Johann-S:
It should help us to be more performant on how we get our DOM elements

Should help a bit here: #26419 but we should improve collapse with multi targets


js/src/collapse.js
```diff
@@ -72,11 +72,14 @@ const Collapse = (($) => {
         `[data-toggle="collapse"][href="#${element.id}"],` +
         `[data-toggle="collapse"][data-target="#${element.id}"]`
       ))
-      const tabToggles = $(Selector.DATA_TOGGLE)
-      for (let i = 0; i < tabToggles.length; i++) {
-        const elem = tabToggles[i]
+      const toggleList = [].slice.call(document.querySelectorAll(Selector.DATA_TOGGLE))
+      for (let i = 0; i < toggleList.length; i++) {
```
XhmikosR: Maybe cache the length in the loops

Johann-S: it was interesting in old browsers but not anymore as far as I know

XhmikosR: Doesn't hurt to do it either.

Johann-S: in that case I'll target that in an other PR, to fix all at once 😉

zhanyirong:
nice

XhmikosR (approved):

mibby:
Apologies for commenting on a pull-request over a year old, but I believe this change may be causing a permanent breaking change for me with newer versions of bootstrap. Perhaps you could enlighten me if I'm missing something obvious?

I updated from Bootstrap v4.1.1 to v4.3.1 when I noticed a page stopped functioning. Testing every version in-between, the breakage seems to occur with v4.1.2. I didn't notice any outstanding breaking changes in the changelog outside of this possible change.

Relevant site script.
```javascript
var all_member_id = 3187407;
function tab_setup() {
    var counter=0;
    $('.m_members').each(function() {
        counter++;
        var title = $(this).parents('.container').find(
            '.header_text_text').text();
        var tab_id = $(this).attr('class').replace(/m_members\s+/g,
            '');
        $(this).attr('data-tab', tab_id);
        if (counter == 1) {
            $('.roster_page .nav-tabs').append(
                '<li class="active"><a data-toggle="tab" href="#' +
                tab_id + '">' + title + '</a>');
            $('.roster_page .tab-content').append(
                '<div class="tab-pane active" id="' + tab_id +
                '"></div>');
            $('#' + tab_id).append($('.m_members[data-tab="' +
                tab_id + '"]'));
        } else {
            $('.roster_page .nav-tabs').append(
                '<li><a data-toggle="tab"  href="#' + tab_id +
                '">' + title + '</a>');
            $('.roster_page .tab-content').append(
                '<div class="tab-pane" id="' + tab_id +
                '"></div>');
            $('#' + tab_id).append($('.m_members[data-tab="' +
                tab_id + '"]'));
        }
    });
};
```


[Live page functioning fine with bootstrap v4.1.1.](https://www.eoscraftmc.com/community)

[Test page failing to toggle data from other elements with bootstrap v4.1.2.](https://www.eoscraftmc.com/communityclone)

Johann-S:
Hi @mibby that's happen because you use selector which contains only numbers, and it's not a valid selector. We added information about that here: https://getbootstrap.com/docs/4.3/getting-started/javascript/#selectors

mibby:
Ah, I see. That indeed would be the problem. Thanks for the info!

Is there any way to perhaps override the selector validation? As I'm not able to change the div ids for the container modules hosting the data itself, which is where the numbers are coming from.

Johann-S:
@mibby  You can change the `href` attribute by something like that: `.tab-content #<your id>` it'll work

PR #26433, opened 2018-05-01
Fix heading levels in grid example

brandon-d-logan:
H3 was used without preceding H2, breaking document flow.


patrickhlauke:
Good spot, thanks

patrickhlauke (approved):

PR #26439, opened 2018-05-02
IE8 uglifyjs compatity

namiltd:
IE8 compatibility for project like https://github.com/coliff/bootstrap-ie8


Johann-S:
Sorry but we won't be compatible with IE 8

/CC @mdo and @XhmikosR

namiltd:
This is not about bootstrap IE8 compatibility, only about compatible 'min' compression.

Johann-S:
UglifyJS docs said: 
>Support non-standard Internet Explorer 8. Equivalent to setting `ie8: true` in `minify()` for `compress`, `mangle` and `output` options.  By default UglifyJS will not try to be IE-proof.

X-Ref: https://github.com/mishoo/UglifyJS2#command-line-options

XhmikosR:
There's no point in this PR or discussion. We just don't want to even try to support an ancient browser.

You are of course free to do what you want in your project.

PR #26452, opened 2018-05-04
Fix double-transition bug

### LINKED ISSUE #26451: [v4] Spurious animation after the carousel has slid
Because the CSS `transition:` property is applied too broadly, a (generally invisible) animation happens after the classes have been set to quiescent state again by the carousel JavaScript code.

This doesn't matter, as the second animation is mostly invisible, except in some cases — For instance, since "transform" breaks z-index, the spurious animation needlessly doubles the time where z-index doesn't work in a parallax carousel (see https://codepen.io/domq/pen/ZoJrQZ )

Narrowing down the application of the `transition:` property to just the active sliding classes, helps (also demonstrated in the codepen, see comments under "CSS").

domq:
Fixes #26451

By applying the transition: CSS property only to classes that are
active during sliding, we avoid an unnecessary, non-zero-time
animation that although mostly invisible, does interfere with e.g.
z-index based parallax carousels (see
https://codepen.io/domq/pen/ZoJrQZ)

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -23,7 +23,6 @@
   display: none;
   align-items: center;
   width: 100%;
-  @include transition($carousel-transition);
   backface-visibility: hidden;
   perspective: 1000px;
 }
@@ -32,6 +31,7 @@
 .carousel-item-next,
 .carousel-item-prev {
   display: block;
+  @include transition($carousel-transition);
 }
 
 .carousel-item-next,


XhmikosR (approved):
LGTM

andresgalante (changes requested):
Hi @domq thanks for the contribution.
I built your branch and the slide animation is not working.
Am I missing something?

domq:
@andreasgalante I'm sorry that the slide animation is not working for you; could you please let me know the steps you took for trying it out? Here is what I did try:

- `npm i`
- `npm test` (I did get failures, but they are unrelated to my change afaict)
- opened `js/tests/visual/carousel.html` in browser — everything seems OK

domq:
sliding

andresgalante:
Sorry, you are right it is working when I open `js/tests/visual/carousel.html` directly.

But when I run the docs `npm run docs-serve` it's not working for me on  `http://localhost:9001/docs/4.1/components/carousel/`

It might be just me, but just wanted to make sure.

XhmikosR:
@andresgalante: unless you run `npm run dist` or `js` the dist files used by the docs are the old ones.

domq:
Thanks @andresgalante for the review. Just to be sure, is the defect you are witnessing the same as in the gif below? If so, I fixed it and rebased onto today's `v4-dev`, kindly take another look.

actual

I.e. no smooth transition at http://localhost:9001/docs/4.1/components/carousel/, as opposed to the online docs (https://getbootstrap.com/docs/4.1/components/carousel/) which work as expected:

expected

andresgalante (approved):

andresgalante:
I am sorry @domq it was my build 😞

andresgalante:
oh! I just read your last comment, thanks for fixing it.

I though it was my build but your last commit is doing the trick :)

PR #26462, opened 2018-05-07
chore(node): use node 6 as it required in our package.json

Johann-S:
it does not fix all of our build issues on this v3 but it helps 😆 

fixed: https://travis-ci.org/twbs/bootstrap/jobs/375776929#L1668
new build issue: https://travis-ci.org/twbs/bootstrap/jobs/375780714#L1540

/CC @mdo, @XhmikosR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: node_js
 git:
   depth: 10
 node_js:
-  - "0.12"
+  - "6"
 before_install:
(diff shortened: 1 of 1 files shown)


XhmikosR:
What is it with the `(crap)` commit messages?

XhmikosR (approved):

Johann-S:
it's a well known convention of commit see: https://conventionalcommits.org/
it's used for example by Angular, Material and so and

XhmikosR:
Yes, I know that. But why would you use it here, when we don't follow that.

Anyway, that convention is so bad IMO.

PR #26471, opened 2018-05-09
ability to generate warning form state

wojtask9:
Currently there is no way to generate warning state for form inputs because css rules with ``state`` other than `valid` and `invalid` is dropped by browsers.

With this PR we are able to generate other states than ``valid`` and ``invalid`` for example

```
@include form-validation-state("warning", $warning);
```
or
```
@include form-validation-state("info", $info);
```

Edit: This is useful if we want Server Side validation

### DIFF
--- a/scss/mixins/_forms.scss
+++ b/scss/mixins/_forms.scss
@@ -25,6 +25,20 @@
   }
 }
 
+@mixin form-valid-state($state) {
+
+  @if $state == "valid" or $state == "invalid" {
+    .was-validated &:#{$state},
+    &.is-#{$state} {
+      @content;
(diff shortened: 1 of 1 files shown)


mdo:
This has been shot down at least once before, but suffice to say we're not going to add it as our validation states are based on what the browser provides.

wojtask9:
@mdo 
I understand your concerns. This PR is not about adding or support other states than `valid` and `invalid` by default.
This is about utilize existing mixin to generate other states in `custom` builds.
Currently mixin `form-validation-state` and reusing in `custom` builds will be great :)

PR #26481, opened 2018-05-10
Add Steppers/Multi Step Form

ayoubkhan558-zz:
Add Steppers,/Multi Step Form in bootstrap as
02_stepwizard

So we do not have to include any external library for that.
screenshot 81


wolfy1339:
First off, you're targeting the wrong branch, second I don't see any changes

Johann-S:
as @wolfy1339 said, please target the right branch

PR #26488, opened 2018-05-11
Add disabled bg color to disabled custom file

### LINKED ISSUE #25537: Custom file input does not have disabled style
There is no appearance difference when add a `disabled`.

cap

Demo: https://codepen.io/anon/pen/NyxRMO

Should be add the disabled style to https://github.com/twbs/bootstrap/blob/v4-dev/scss/_custom-forms.scss#L242 ?

ysds:
Fixes #25537

image

Demo: https://codepen.io/anon/pen/odqrEg


andresgalante (approved):
Thanks @ysds

PR #26500, opened 2018-05-12
Mixins documentation page

### LINKED ISSUE #25977: Mixins documentation
Not every mixin is documented in the [Bootstrap docs](https://getbootstrap.com/docs/4.0/getting-started/introduction/). Maybe we should add a dedicated place for mixins in the docs as @kevdotbadger suggested [here](https://github.com/twbs/bootstrap/issues/25752#issuecomment-374306145).

MartijnCuppens:
Fixes #25977.

All mixins are generated by the `mixins_documentation` jekyll plugin under the `_plugins` folder. This plugins executes the `npm run mixins-documentation --silent` command. The `--silent` parameter prevents the filename to be outputted.

`npm run mixins-documentation --silent` runs `build/mixins-documentation.js`. This script reads all the `.scss` files in the `scss/mixins` folder and returns a list of the mixin names, the file location and the mixin itself in code block.

It's also possible to add a description to a mixin. Therefore you can add a comment block just above the mixin (no blank lines between the mixin and the comment block). Markdown is supported in these blocks. Stylelint comments are ignored. 

Demo: <https://deploy-preview-26500--twbs-bootstrap4.netlify.com/docs/4.2/extend/mixins/>

TODO:

- [x] fix stylelint comments detection
- [x] switch to shell.js's find
- [x] link to the actual mixin file
- [x] automatically convert links
- [x] move `Mixins` page under `Extend` nav
- [ ] enable stylelint rule to check for newlines between mixins, and no newlines before mixins when there's a comment
- [ ] maybe sort mixins alphabetically; currently they are sorted by their filename and then alphabetically
- [ ] use smaller headers for mixins in the same file so that ToC detects them properly
- [ ] decouple task from Jekyll maybe (remove the Jekyll plugin)
- [ ] show a callout for deprecated mixins
- [ ] fix link to file line bug
- [ ] Check if https://github.com/twbs/bootstrap/issues/25977#issuecomment-455917335 could be useful to build the docs


XhmikosR:
Wow I totally missed this patch :/

From a quick look I like it. But I still see stylelint comments in the output.

XhmikosR:
Also, I wouldn't enforce Markdown in the comments. We could just use a lib maybe?

MartijnCuppens:
@XhmikosR, should'nt we leave the stylelint comments? And do you have any suggestions for a lib we can use to transform the links in clickable links?

XhmikosR:
I think the Stylelint comments don't offer any help here where we just want the code and we shouldn't assume everyone is using Stylelint.

For libs there are plenty, not sure which one to use, maybe check them out with https://packagephobia.now.sh/

* https://www.npmjs.com/package/marked
* https://www.npmjs.com/package/remark
* https://www.npmjs.com/package/markdown-it
* https://www.npmjs.com/package/showdown

and probably even more :)

XhmikosR:
I rebased this and squashed it. Not sure about the plugin which runs the npm script though. I don't like mixing the two environments a lot.

build/mixins-documentation.js
```diff
@@ -0,0 +1,76 @@
...
+const execFile = require('child_process').execFile
+
+let mixinName
+let mixin
+let mixinDocumentation = ''
+let inMixin = false
+let output = ''
+
+// Find all files in mixins folder
+execFile('find', ['./scss/mixins'], (err, stdout) => {
```
XhmikosR: Better change this to shelljs's find.

scss/mixins/_screen-reader.scss
```diff
@@ -16,10 +15,9 @@
 
 // Use in conjunction with .sr-only to only display content when it's focused.
 //
-// Useful for "Skip to main content" links; see https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1
+// Useful for "Skip to main content" links; see [https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1](https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1)
```
XhmikosR: This kind of changes is what a Markdown lib could help :) Or we could change the script to look for `/https?/` or something so that links are automatically converted.

XhmikosR:
It seems just the check for stylelint comments needs to be tweaked. The `startsWith` doesn't cover the cases when there are spaces before `//`.

Something like `/^\s*?\/\/\s?stylelint/` should do the job.

scss/mixins/_float.scss
```diff
@@ -3,9 +3,11 @@
 @mixin float-left {
   float: left !important;
 }
+
```
XhmikosR: BTW this reminds me, we should have stylelint check for this newline.

build/mixins-documentation.js
```diff
@@ -6,9 +6,11 @@ const Stream = require('stream')
 sh.config.fatal = true
 
 const stylelintLineRegex = /\s*?\/\/\s?stylelint.*/
+const fileUrl = 'https://github.com/twbs/bootstrap/blob/master/'
```
XhmikosR: Not master :P

MartijnCuppens:
> link to the actual mixin file

I've added this, but then I realised we'll need to change this link everytime a new release is released (eg https://github.com/twbs/bootstrap/tree/3b558734382ce58b51e5fc676453bfd53bba9201 for v4.1.3) because the mixin could get removed or moved to another file. 

@XhmikosR, do you have any idea how we could automate this?

XhmikosR:
Yeah, use the tag :)

XhmikosR:
https://github.com/twbs/bootstrap/tree/v4.1.3

I'd say you could even use the version from package.json instead of hardcoding it too.

MartijnCuppens:
Oh, didn't kwow you could also link to a tag, nice. Getting the version from `package.json` is the best idea than!

build/mixins-documentation.js
MartijnCuppens: I used master as placeholder before asking this: https://github.com/twbs/bootstrap/pull/26500#issuecomment-432535399

But I underestimated your speed of reviewing 😀

MartijnCuppens:
I used `require('./../package.json')` to get the package file and version number, but I noticed some other files are using another method:
https://github.com/twbs/bootstrap/blob/47ef0a877e6fc18c1e516f7a0855e0531c77af96/build/banner.js#L1-L2

Another option is to use `process.env.npm_package_version` to get the version number.

Which approach would be the best?

XhmikosR:
@MartijnCuppens: just updated deps with fusv with the fix so feel free to rebase and adapt the commands for this branch.

MartijnCuppens:
@XhmikosR,
> enable stylelint rule to check for newlines between mixins, and no newlines before mixins when there's a comment

Don't think there is a stylelint rule for this at the moment

>  maybe sort mixins alphabetically; currently they are sorted by their filename

I tried to store the output in an array, but the streams happen async, so I cant just sort all mixins after the `scss.forEach((file) => {...})`. I don't know the appropriate way to handle this?

> use smaller headers for mixins in the same file so that ToC detects them properly

Can you give an example of what you mean?

> decouple task from Jekyll maybe (remove the Jekyll plugin)

Any other suggestions here?

XhmikosR:
1, 2 I'll check it out later.

3. When a file has multiple mixins those are listed with the same header level thus ToC can't distinguish them.

4. None, so let's leave it as TODO for now and we revisit before the merge.

mdo:
Sounds good, you okay getting this into our v4.3 release (recently changed from v4.2.1)?

MartijnCuppens:
Not sure if we can tackle all TODOs, but I'll check with @XhmikosR what we can do

XhmikosR:
The line counter seems wrong, most noticeable in mixins that are not the only ones in a file.

MartijnCuppens:
@mdo, we're going to postpone this to `v5` since we'll need to fix some issues and add more documentation. I'm probably going to create a new branch under `twbs` to make it easier to work on this.

mdo:
Thoughts on the status of this @MartijnCuppens?

MartijnCuppens:
I think we now better use `scss-docs` we recently introduced to document code fragments

frank101594 (commented):
file:///private/var/mobile/Containers/Shared/AppGroup/CF621F42-9601-44F7-8BFF-C5990641C86D/File%20Provider%20Storage/88219381/1IzQCUzzb1E84lB6JR47zMDg5wAFb9ZGa/memories_history.html

Issue #26479, opened 2018-05-10
New overflow-hidden utility class?


sts-ryan-holton:
Hi, I'm suggesting another simple feature here for Bootstrap. Often I find in some projects I'm needing to set `overflow` to hidden to maintain consistent sizing or as part of jQuery functionality, I often find either having to set overflow on either the horizontal axis or vertical axis.

I was thinking we could possibly create some helper classes like this - `overflow-*`

Such as:

```
.overflow-y {
     overflow-y: hidden
}
```

Any possibility of getting some base functionality in place for this? We could possibly make use of SCSS variables to choose either hidden or auto for example.

kevdotbadger:
To my `.overflow-y` implies there IS overflow. The other show/hide visible/invisible classes describe their function quite clearly.

So, I think, to maintain consistency, it should be `.overflow-y-hide` and `.overflow-x-hide`.

sts-ryan-holton:
@kevdotbadger  Understandable, I was just expressing one particular method. I agree with your comment, could we possibly allow for changing the `hide` to `auto` as well? You could have a few classes such as the following:

`.overflow-y-hide`
`.overflow-y-auto`
`.overflow-y-scroll`
`.overflow-y-visible`

`.overflow-x-hide`
`.overflow-x-auto`
`.overflow-x-scroll`
`.overflow-x-visible`

And possibly some that do both x & y

`.overflow-hide`
`.overflow-auto`
`.overflow-scroll`
`.overflow-visible`

MartijnCuppens:
IMO there is no need for all these classes. A helper class for `overflow: hidden` maybe is something I might use, but I don't see me using classes like `.overflow-x-visible`.

kevdotbadger:
Thoughts on responsive versions for these helpers? I can't really think of a use case where you'd need to responsively screw with overflow, so it may be overkill.

PR #26509, opened 2018-05-13
Add overflow helpers

kevinruscoe:
Simply adds a bunch of overflow helpers as per https://github.com/twbs/bootstrap/issues/26479


scss/utilities/_overflow.scss
```diff
@@ -0,0 +1,16 @@
+// stylelint-disable scss/dollar-variable-default
+$directions: x, y;
+$name-map: (
+  hide: hidden,
```
mdo: Make this one `hidden: hidden` so that it's 1:1 with the property-value pair.

scss/utilities/_overflow.scss
```diff
@@ -0,0 +1,16 @@
...
+  auto: auto,
+  scroll: scroll,
+  visible: visible,
+);
+
+@each $name, $value in $name-map {
+  .overflow-#{$name} { overflow: $value; }
+
+  @each $direction in $directions {
+    .overflow-#{$direction}-#{$name} { overflow-#{$direction}: $value; }
```
mdo: Wondering if we start with just the standard overflow instead of directional options? Easier to add than it is to remove :).

kevdotbadger: Huh? So you're thinking .overflow-hidden, .overflow-auto, .overflow-visible, then do the x and y versions?

scss/utilities/_overflow.scss
kevdotbadger: Noted.

mdo:
Given the deleted fork here and additional comments in the original issue, closing this for #27268.

PR #26510, opened 2018-05-14
Make navbar download the same as homepage download

tbodt:
By far the most common reason to download bootstrap would be to get the minified source, which the download button in the navbar currently does not do.

It might be a good idea to directly link to the minified source instead. Not sure.


romieleo (approved):

mdo:
Cherry-picked this into #26732. Thanks!

Issue #25832, opened 2018-03-13
Type scale map and type utility classes. (Font)


sanchez2210:
Hi! First of all, sorry for broken english. Please admins feel free to
edit / ask for clarification is something doesn't make sense.

So, as I understand it, bootstrap has some options for the use of
typography. Like $font-size-lg and $font-size-sm variables and
the `.small` and `<small />` tags. Also, there's the possibility to 
use the `.h1 ... .h6` classes to change the font size of any element.

Still I find that `lg` and `sm` font-sizes are insufficient for every
case. And even though the `.hX` could serve this purpose, they
have some inherent semantic sense that is inappropriate in many
cases. The docs even intruct to use them: 

> for when you want to match the font styling of a heading, but
cannot use the associated HTML element.

It'd be useful to have dedicated: `size-1` to `size-6` (or better name)
that are only supposed to modify the element's font size. These
classes would be dependent on a potential `$type-scale`  variable like:

```
$type-scale: (
  1:  2.5,
  2:  2,
  3:  1.75,
  4:  1.5,
  5:  1.25,
  6:  1,
  7:  .8,
  8:  .7
);
```

This could be implemented with a function called `font-size($scale-number)`.

I understand one of the counterpoints to this feature is that it can
be implemented with custom, per-project css, and it's true. But I
can see the benefit of having it by default in the framework. One
way it could be extended is having another `line-height-for($scale-number)`
function that calculates a line height that is a multiple of the base line
spacing `($line-height-base * $font-size-base)`, as an example.
So `size-1` could, for example, apply:

```
font-size: font-size(1);
line-height: line-height-for(1);
```

Notice that there's space for additional font sizes smaller than the
base, (7 & 8). This is more practic than having
`$font-size-{sm/lg/base/smaller/larger/super-large}`.

In summary, the things why I'd like to see this implemented are:

 1. Consistency / One source of truth. Font size units not  
     scattered around / duplicated.
 2. Versatility. `size-1` would allow developers to just modify the
     sizing aspect of an element's typography. Leaving `.hX`
     classes for proper title-replication.
 3. Example value. By implementing a good-practice, well defined
     type scale.
 4. Extensibility. A couple of issues in the past were raised, in
     my opinion partly, because of a lack of font related features.
     Some of them: #24856 ,  #18137, #23119 . 
 5. Easy to implement.

I'd love to hear your opinion, cheers!

mdo:
[...]

PR #26533, opened 2018-05-18
Easily add display headings with sass maps

### LINKED ISSUE #26515: More display-* headings - .display-5
### New .display-5 class

**Something like:**

```
.display-5 {
    font-size: 2.5rem;
    font-weight: 300;
    line-height: 1.2;
}
```

display-5-class

Many Thanks!

:-)

MartijnCuppens:
Fixes #26515.

Provides a way to easily add display headings with a sass map.


sanchez2210:
Will this be possible with normal text utilities? #25832

scss/_variables.scss
```diff
@@ -260,16 +260,32 @@ $headings-font-weight:        500 !default;
...
-$display2-size:               5.5rem !default;
-$display3-size:               4.5rem !default;
-$display4-size:               3.5rem !default;
-
-$display1-weight:             300 !default;
-$display2-weight:             300 !default;
-$display3-weight:             300 !default;
-$display4-weight:             300 !default;
-$display-line-height:         $headings-line-height !default;
+$displays: (
```
ocarreterom:
```scss
$displays: () !default;
$displays: map-merge($displays, ());
```

andresgalante:
Hi @MartijnCuppens, this solution would gives more settings and it works.

My problem is that by having the definition for headings scattered into different files, instead of scoped to type.scss we would be making the codebase harder to contribute, code and maintain.

What do you think?

MartijnCuppens:
Hi @andresgalante,  

Would you move the `$displays` sass map to the `_type.scss`? All bootstrap variables are defined in `_variables.sccs`, I think it's better to keep it there?

andresgalante:
Yes, you are right, all variables are define on variables and putting them within a map wouldn't make a difference than having them as variables outside the map.

mdo (dismissed):
You've removed several variables—that's a breaking change. You can likely rewrite the Sass map to include those variables rather than remove them outright.

MartijnCuppens:
I re-added the original display variables

MartijnCuppens:
I'm not convinced anymore that this is a good idea (especially for v4). This technique is quite different from the way we handle things in other places, we're practically putting every property of the generated css in a sass map.

I'm considering closing this, unless someone has a good argument to merge this.

XhmikosR:
CC @andresgalante @mdo

mdo:
Let's punt on it and reimagine the typography in v5 to do it right. v5 shouldn't be sweeping changes, but an easy-ish upgrade path. I feel like stuff like this can be tackled then. Feel free to close if you still agree! <3

PR #26538, opened 2018-05-19
Custom forms transition

MartijnCuppens:
Currently there is no transition applied to custom form elements. This PR adds transition to checkboxes, radios, custom range sliders, custom selects and custom file inputs.

`background-color` transition is needed to transition the checkboxes and radio buttons. This is why `$custom-forms-transition` is not inherited from `$input-transition`

Demo: https://codepen.io/MartijnCuppens/pen/OZrBYj


patrickhlauke:
just to check: does this also obey the "reduced motion" thing (where transitions are disabled) https://getbootstrap.com/docs/4.1/getting-started/accessibility/#reduced-motion ?

MartijnCuppens:
@patrickhlauke, 
Yes, the `transition` mixin is used for each transition.

andresgalante (approved):
@MartijnCuppens cool! I like it

Issue #26546, opened 2018-05-21
Improving documentation


ItsPugle:
Hey,

I'm looking through the [contributing guidelines](https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md), and I've noticed that it directs future-contributors to edit documentation on the master branch, under the `/bootstrap/docs/` directory.

> Similarly, when contributing to Bootstrap's documentation, you should edit the documentation source files in the /bootstrap/docs/ directory of the master branch. Do not edit the gh-pages branch. That branch is generated from the documentation source files and is managed separately by the Bootstrap Core Team.

However:

1. There is no `/bootstrap` directory (there is a `/docs` though)
2. The documentation in `/docs/_includes` hasn't been updated in ~2 years (apart from #23808)

So, should where should I be targeting my PRs against?

coliff:
Hi. For changes to the Bootstrap 4 documentation you should make PRs against the `v4-dev` branch (which is the default). https://github.com/twbs/bootstrap/tree/v4-dev/docs/4.1
The docs includes are at: https://github.com/twbs/bootstrap/tree/v4-dev/_includes

Agree that the contributing guidelines could be updated to make that clearer though I don't think it's incorrect to specify that the root directory is `/bootstrap`. That is what shows in GitHub's web interface and it's referenced as such consistently in the documentation e.g.  `/bootstrap/scss/` and `/bootstrap/js/`

ItsPugle:
Ah, gotcha! I'll make a PR with the changes that I was going to do originally, then also one for clarifying about Documentation PRs 😄 Thanks for your help!

PR #26552, opened 2018-05-22
Adding Bower to download options

ghost:
Hey,

Since a Bootstrap package is available through [Bower](https://bower.io/), it's probably a good idea to add it as a download option since:

1. It promotes using the same software for developers, providing a more native experience and in turn giving less reasons for people to not use Bootstrap (don't we want more people to use Bootstrap, anyways?)
2. It acknowledges the package on the Bower registry, and it looks as though it is an official one, since it's owned by `twbs`

This PR just simply adds Bower as a download option :^)

#### P.S.
Sorry if this is the wrong branch/location to target, I was just taking some advice and agreed with it on #26546

### DIFF
--- a/docs/4.1/getting-started/download.md
+++ b/docs/4.1/getting-started/download.md
@@ -87,6 +87,14 @@ You can also install and manage Bootstrap's Sass and JavaScript using [Composer]
 composer require twbs/bootstrap:{{ site.current_version }}
 {% endhighlight %}
 
+### Bower
+
+A Bootstrap package is also available through [Bower](https://bower.io/):
+
+{% highlight sh %}
+bower install bootstrap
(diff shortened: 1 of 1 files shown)


jgacuca567:
@ItsPugle 
**Bower** is no longer maintained.

ItsPugle:
Hey, @jgacuca567!

As far as I know, Bower is actually still maintained. Their is a notice on the website saying so, too:
> [...] While Bower is maintained, we recommend using Yarn and Webpack for front-end projects read how to migrate!

It's still used a lot, too.

mdo:
It's basically dead at this point and I don't see any reason to bring it back. We made of a point of dropping it with v4 given their encouragement of alternatives. Thanks though.

PR #26561, opened 2018-05-23
fix branch for coveralls badge

Johann-S:
/CC @XhmikosR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
   - npm install
 after_success:
   - if [ "$TRAVIS_REPO_SLUG" = twbs-savage/bootstrap ]; then npm run docs-upload-preview; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "8" ]; then npm run coveralls; fi
 stages:
   - test
   - name: browser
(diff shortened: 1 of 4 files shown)


XhmikosR:
@Johann-S: can you change it to shields.io?

Johann-S:
Wow we should just run coverall for one build not for all 😆 and done @XhmikosR 😉

Johann-S:
Oh yes you're right, I wasn't sure about that 😄

XhmikosR:
@Johann-S: not sure why, but every time you change the package-lock.json you make many changes. Please make sure you are using `npm` that ships with node 8 or 10 and not another version. I'm gonna revert the changes, but we need to make sure it doesn't happen again.

Johann-S:
I use npm `v6.0.0` and node `v9.7.1` 😟

XhmikosR:
Yeah, please move to 8 or 10.

XhmikosR:
OK, so this runs coveralls after success, but for both builds. It doesn't seem to affect anything.

Johann-S:
and no more comments 🎆

XhmikosR:
I could move the `after_success` call to run only after node.js 8. What do you think about that?

Johann-S:
Much better to just run it once 👍

XhmikosR:
Not sure this would run if `browser` stage doesn't run though :/

We need some help here.

Johann-S:
Or you can revert that change and let `coveralls` command in our package.json as we do not have Coveralls comments

XhmikosR:
OK, I think we just need  a little shell help. See my last push.

Johann-S:
I made a change, which run `coveralls` just for node 8, can you take a look @XhmikosR ?

XhmikosR:
Does the job indeed!

But I think we should not export another variable and just do the version check in the if check.

Something like this: https://stackoverflow.com/questions/48250235/how-to-setup-travis-to-skip-script-on-particular-nodejs-version

XhmikosR (approved):

XhmikosR:
LGTM @Johann-S, feel free to rebase and merge :)

XhmikosR:
@Johann-S: I just become aware there's a `TRAVIS_NODE_VERSION`. Might be worth checking if this variable is something we can just use.

XhmikosR:
OK, it works :) I'm gonna merge when the build finishes.

PR #26573, opened 2018-05-23
Card header color theming

### LINKED ISSUE #26571: Card header color is not themable with Sass variable
Since the `.card-header` `background-color` is able to be themed using `$card-cap-bg`, the `color` should also be themable, otherwise you're stuck with the default color for whatever style text you put there. This could lead to low contrast between the background and text color, like so: 

image

This would ideally be reconciled by adding a new variable, but could perhaps be fixed by using the `color-yiq` function to automatically set the color based on the value for `$card-cap-bg`.

andrewpucci:
Closes #26571 

Adds a new variable `$card-cap-color` with a default value identical to that of `$bg-color`. This default value is an acceptable contrast ratio to the default value for `$card-cap-bg`, well over 7.

This variable is then used to assign a `color` to `.card-header`.


mdo:
Rather than add a unnecessary `color` property by default, why not use our utilities?

andrewpucci:
Yes, you could use a color utility like `.text-white` to change the color of the text in a card header, and that's great for one offs, but when building a theme, having to use a color utility for every card header seems a little misguided.

Please point me in the right direction if I'm missing something, but my interpretation is that the Sass variables are used when you want to make a change to a theme.

To give a counter example, if `$card-cap-bg` was not available, I'd have to use something like `.bg-dark` on every card header. The variable I'm adding here is meant to serve a purpose similar to the existing `$card-cap-bg` variable.

andresgalante:
I see @andrewpucci's point, if you are creating a theme you'd want to set a default color for the header. We define a background-color and we could use a utility too, right?

scss/_variables.scss
```diff
@@ -716,6 +716,7 @@ $card-border-radius:                $border-radius !default;
 $card-border-color:                 rgba($black, .125) !default;
 $card-inner-border-radius:          calc(#{$card-border-radius} - #{$card-border-width}) !default;
 $card-cap-bg:                       rgba($black, .03) !default;
+$card-cap-color:                    $gray-900 !default;
```
mdo: Let's make this `$body-color` since that's what it was before.

andrewpucci: Done, thanks!

ayoubkhan558-zz (approved):
Its a good option to have.

mdo (approved):

Issue #23969, opened 2017-09-16
Feature Request: Additional Font Weight Classes


planetoftheweb:
I think it would be great to include some more font-weight options in bootstrap. Even when you look at the documentation, you can see that the weights have been modified because current design sensibilities call for a wider range of font weights that the default. I realized that when I'm creating bootstrap projects, I end up modifying the font-weights a lot, so maybe this could be added into the classes.

![screen shot 2017-09-16 at 7 22 29 pm](https://user-images.githubusercontent.com/216426/30516697-657091d2-9b14-11e7-9faf-a0866a469639.png)

Although `font-weight-bold` and `font-weight-normal` exist, I feel like there needs to be a bit more broad.  I'm proposing adding some font-weight classes.

`font-weight-light`
`font-weight-medium`
`font-weight-normal`
`font-weight-bold`
`font-weight-heavy`

These could be tied to sass variables. We could also add numbered font-weight classes.

`font-weight-100`
`font-weight-300`
...etc

andresgalante:
Hi @planetoftheweb reading your issue I noticed that `font-weight-light` is defined on the code base but not surfaced on the docs. I've sent a PR to add it #23970

As for adding more utility classes I think it's a good idea, in fact I am doing it at my own project. 

From a bootstrap perspective I'd make sure to test it in different browsers and OS. Since Bootstrap has a System Font Stack, it might be the case that some font families don't have a specific weight (in numbers).

Using keywords is a safer bet and users can always define them on their projects.

anyway, good idea 👍

XhmikosR:
We could add them, but we must have a note that based on the font used, the result might not be the expected one, and it's the developer's job to make sure the font weight they are using exists.

See for example an issue with our docs, #24419.

polys:
Could you add `font-weight-medium` as well please?

It would set `font-weight: 500 !important;` similar to how `font-weight-light` is defined.

I find it very useful and keep adding it to my projects; it would be nice if it was part of bootstrap. I understand that it may not work with all fonts but at least it sets the _intent_. How's `font-weight-light` any different?

mdo:
Closing since we had the lighter/bolder added. For any other weights and utilities, please feel free to open PRs to suggest :).

PR #26580, opened 2018-05-24
add 'lighter' and 'bolder' font weight classes

browner12:
these are 2 special values for font weight, that will give their content a font-weight value of 100 more or less than their inherited font-weight.

probably doesn't fully fulfill this issue, https://github.com/twbs/bootstrap/issues/23969, but it's a start

### DIFF
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -32,10 +32,12 @@
 
 // Weight and italics
 
-.font-weight-light  { font-weight: $font-weight-light !important; }
-.font-weight-normal { font-weight: $font-weight-normal !important; }
-.font-weight-bold   { font-weight: $font-weight-bold !important; }
-.font-italic        { font-style: italic !important; }
+.font-weight-light   { font-weight: $font-weight-light !important; }
+.font-weight-lighter { font-weight: lighter !important; }
+.font-weight-normal  { font-weight: $font-weight-normal !important; }
+.font-weight-bold    { font-weight: $font-weight-bold !important; }
+.font-weight-bolder  { font-weight: bolder !important; }
+.font-italic         { font-style: italic !important; }
 
 // Contextual colors
 
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -33,10 +33,10 @@
 // Weight and italics
 
 .font-weight-light   { font-weight: $font-weight-light !important; }
-.font-weight-lighter { font-weight: lighter !important; }
+.font-weight-lighter { font-weight: $font-weight-lighter !important; }
 .font-weight-normal  { font-weight: $font-weight-normal !important; }
 .font-weight-bold    { font-weight: $font-weight-bold !important; }
-.font-weight-bolder  { font-weight: bolder !important; }
+.font-weight-bolder  { font-weight: $font-weight-bolder !important; }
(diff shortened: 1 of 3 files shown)


browner12:
to fix these errors I guess we'd have to update

https://github.com/twbs/bootstrap/blob/v4-dev/.stylelintrc#L21

I can't speak to the original intent behind using this rule, so I'm not sure what the appropriate answer is.

XhmikosR:
Don't bolder and lighter have numeric equivalents?

patrickhlauke:
> Don't bolder and lighter have numeric equivalents?

they are relative to the inherited/current value, so no

browner12:
'bold' is equivalent to 700.

'lighter' and 'bolder' are variable, and are either 100 more or less than the value they inherit.

```html
<div style="font-weight:600; ">
     <div style="font-weight:lighter; ">This text will be 500 weight.</div>
     <div style="font-weight:bolder; ">This text will be 700 weight.</div>
</div>
```

http://htmldog.com/references/css/properties/font-weight/

XhmikosR:
Ok, it's passing, so it's just a matter if we want this in or not. /CC @mdo @andresgalante

MartijnCuppens:
These utility classes might need more documentation. There will be more people who think these values have numeric equivalents.

browner12:
sounds good. if it get's accepted i'll add some docs.

XhmikosR:
It's already been accepted for v4.2 :)

browner12:
really? link?

scss/utilities/_text.scss
```diff
@@ -32,10 +32,12 @@
 
 // Weight and italics
 
-.font-weight-light  { font-weight: $font-weight-light !important; }
-.font-weight-normal { font-weight: $font-weight-normal !important; }
-.font-weight-bold   { font-weight: $font-weight-bold !important; }
-.font-italic        { font-style: italic !important; }
+.font-weight-light   { font-weight: $font-weight-light !important; }
+.font-weight-lighter { font-weight: lighter !important; }
```
ahtee: would this need to stay uniform with the other styles and be 
lighter -> $font-weight-lighter,
bolder -> $font-weight-bolder, etc?

would there need to be additions to make this work?

XhmikosR: Those are CSS values.

mdo (changes requested):
Per the conversation at https://github.com/twbs/bootstrap/pull/26580/files#r193721911, let's make these variables so that folks can change them if they so choose. I know these values are inherently relative, but they could easily be remapped to `100` and `900` depending on the situation.

mdo:
I understand their intended usage, but people do whatever they want with CSS like this. 😆

My point is, if we include two new variables here, we'll be helping folks who want to have explicit options here. As soon as these become utilities and variables, it's an opportunity to move away from browser default values.

Take this example below: both are valid in their own contexts.

```scss
.font-weight-lighter { font-weight-lighter: lighter; }
.font-weight-light { font-weight-lighter: light; }

.font-weight-lighter { font-weight-lighter: 100; }
.font-weight-light { font-weight-lighter: 300; }
```

browner12:
alright, the variables have been added.

mdo (approved):
Thank you!

PR #26618, opened 2018-05-28
"above image" -> "image above"

ogu2:
grammar fix: "above image" -> "image above"


patrickhlauke:
both ways of saying/writing this are valid. thanks for the suggestion though.

PR #26622, opened 2018-05-29
appearance need -moz- and -webkit- prefix

miau715:


patrickhlauke:
No, as Bootstrap uses autoprefixer. https://getbootstrap.com/docs/4.1/getting-started/build-tools/#autoprefixer

PR #26638, opened 2018-05-31
Changed Node.js download URL

marcosfreitas:
I think this link have a better understanding to download Node.js


patrickhlauke (approved):

patrickhlauke:
good spot, thanks

Issue #26643, opened 2018-06-01
Improve how we traverse the DOM


XhmikosR:
I think we need to eventually do the following across the whole codebase:

1. use `querySelector` when we just need the first match
2. use `node.querySelectorAll` instead of `document`
3. **always** cache the result to a var

These might be micro-optimizations, but they should definitely help with large DOMs.

arunikayadav42:
Definitely :)

arunikayadav42:
but exactly how do I start , can you let me know please.I am a beginner to open source

Arvind-Suthar:
Can i know the problem with large DOM
I can help minimize the problem?

PR #26645, opened 2018-06-01
refactor(plugins): query elements without jquery

Johann-S:
Fixes: #26643

/CC @XhmikosR

### DIFF
--- a/js/src/collapse.js
+++ b/js/src/collapse.js
@@ -68,7 +68,7 @@ const Collapse = (($) => {
       this._isTransitioning = false
       this._element         = element
       this._config          = this._getConfig(config)
-      this._triggerArray    = $.makeArray($(
+      this._triggerArray    = $.makeArray(document.querySelectorAll(
         `[data-toggle="collapse"][href="#${element.id}"],` +
         `[data-toggle="collapse"][data-target="#${element.id}"]`
       ))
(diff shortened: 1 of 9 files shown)


Johann-S:
I have some issues to fix for IE 😆

js/src/collapse.js
```diff
@@ -162,7 +162,7 @@ const Collapse = (($) => {
 
       this._element.style[dimension] = 0
 
-      if (this._triggerArray.length > 0) {
+      if (this._triggerArray.length) {
```
XhmikosR: I'd keep the length checks

Johann-S: That's useless because 0 is a false value

Johann-S: Yep but that rule isn't activated

XhmikosR: I never said it was :P I remember I made the length changes some time ago, so I'd say keep them even though we don't have that check.

Johann-S: I prefer to remove them, it's part of the JavaScript power 💪

XhmikosR: I don't agree, because I'd rather be explicit when possible. It's actually one of the problems of JavaScript, not strict types etc.

Johann-S: That's a well known property of JavaScript that 0 is a false value, be explicit isn't a good thing when it's hide a language particularity because other languages are different

XhmikosR: Sorry, I don't agree with the length removal. Even if we know that, I always prefer to be explicit.

The rest of the patch looks good, though. 👍

XhmikosR (approved):
Forgot to approve for the rest.

PR #26647, opened 2018-06-01
Stricter stylelint rules

MartijnCuppens:
Stylelint is disabled and enabled if sass maps are merged with their defaults in `_variables.scss`. It's better to just disable the `scss/dollar-variable-default` rule and keep the other rules enabled. Also did some indention fixes in this PR to comply with other stylelint rules.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -8,7 +8,6 @@
 // Color system
 //
 
-// stylelint-disable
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

XhmikosR:
Thanks!

PR #26650, opened 2018-06-02
Allows tab function to be used with OL and UL

tiesont:
There are times when an `<ol>` is the correct node to use for a list, but the current tab() function will only work with either `<ul>` or `<nav>`. 

This PR changes `ACTIVE_UL` to `ACTIVE_LIST` and adds `nodeName === 'OL'` to the two checks for the tab button's parent node.


Johann-S (changes requested):
Please add a unit test which cover this case

tiesont:
Actually curious to know how the current unit tests pass (assuming `v4-dev / js / tests / unit / tab.js` is the correct file to update). There are QUnit tests for ol-based markup, but I could not get tabs to work correctly without the changes I've made. I *think* it's because these tests only check for `active` and not `show`, which is how the active tab/pill is indicated. Do I have that right?

Johann-S:
I think you're right but can you add or change a unit test with a valid example ?

tiesont:
Sure. Won't be today, but I'll try to add something within a day or two.

tiesont:
@Johann-S Should the `assets/vendor/ ` folder contain a copy of QUnit (seems so from the README)? Seems to be missing, assuming patch-2 is the correct branch to be working from. Trying to run the unit tests in a browser, since I don't have NPM installed (and would rather not).

Johann-S:
No we made the decision to remove in our vendors because it's in our dev deps, so you have to install it with NPM, and please do not commit dist files

tiesont:
Okay, it'd probably be easiest if I close this, delete my fork, and create a new one with just the required updates. _Is_ patch-2 the correct branch to be working from?

Johann-S:
you can call your branch whatever you want 😉

PR #26653, opened 2018-06-02
Fix node_modules path

MartijnCuppens:
On the getting started with theming page http://getbootstrap.com/docs/4.1/getting-started/theming/#importing, the path to the `node_modules` folder is wrong.

This is the file structure:
```
your-project/
├── scss
│   └── custom.scss
└── node_modules/
    └── bootstrap
        ├── js
        └── scss
```

So if we start from `scss/custom.scss`, we need to add `../` to get to the `node_modules` folder.

### DIFF
--- a/docs/4.1/getting-started/theming.md
+++ b/docs/4.1/getting-started/theming.md
@@ -50,7 +50,7 @@ In your `custom.scss`, you'll import Bootstrap's source Sass files. You have two
 // Custom.scss
 // Option A: Include all of Bootstrap
 
-@import "node_modules/bootstrap/scss/bootstrap";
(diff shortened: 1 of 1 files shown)


XhmikosR:
Well, that's not always true because one might have added the parent folder in the includes.

That being said, I'm OK I guess.

MartijnCuppens:
@XhmikosR, what do you mean? Is there a way this works with the proposed [file structure](http://getbootstrap.com/docs/4.1/getting-started/theming/#file-structure)?

XhmikosR:
Yes it can work if you add the parent folder in the compiler's include
path. But anyway, the common thing would be your way for most people.

Johann-S (approved):
No concerns about that change

PR #26674, opened 2018-06-06
Flexbox based sticky footer example

### LINKED ISSUE #26656: Bootstrap 4 sticky footer overlaps content
I am using Bootstrap 4 for adding sticky footer which have two components : bottom bar(using container-fluid) with copyright year and website name and sub-footer navigation menu(using container element) which have sitemap and other links.

Here is my css
```css

html {
  position: relative;
  min-height: 100%;
}

body {
  margin-bottom: 40px;
  height: 100%;
}

.footer {
  position: absolute;
  bottom: 0;
  width: 100%;
}

.footer .container-fluid {
  height: 40px;
}

.footer .container {
  padding-top: 20px;
}
```
Whenever I remove those two last footer sub classes and add height:40px; to footer, there is no overlapping of content but a page shows vertical scrollbar even with less content on page.

ysds:
Closes #26656

The sticky footer examples are using specific the height of the footer (60px). I've rewritten it to flexbox based for remove the specific values. The benefitis that the CSS will be more general and more responsive design.

Demo: https://codepen.io/anon/pen/WyxRKp

..I'm worried about using `main` element selector. Please review it too :)


MartijnCuppens:
I was thinking about making a similar PR 😉 Using the `main` selector might indeed not be the brightest idea.

Maybe we can take this one step further and add some classes which we can use in bootstrap? One class to put on the `body` and one to put on the `main`?

ysds:
Of course. The following custom CSS can be replaced by bootstrap utility classes.

```scss
body {
  display: -ms-flexbox;
  display: flex;
  -ms-flex-direction: column;
  flex-direction: column;
  height: 100vh; /* Use height insted of min-height for IE 10/11. */
}
main {
  /* Prevent shrinking the main area in IE 10/11. */
  -ms-flex-negative: 0;
  flex-shrink: 0;
}
.footer {
  padding: 1rem 0;
  margin-top: auto;
  background-color: #f5f5f5;
}
```

to

```html
<html class="h-100">
<body class="d-flex flex-column h-100">
<main class="flex-shrink-0">
<footer class="py-3 mt-auto bg-light">
```

In many current examples, it seems inconsistent to use utility classes and custom CSS usage. For example, in the sign-in example, `body` has just `text-center` class.

https://github.com/twbs/bootstrap/blob/9c3146d0dd1320d9b1e902005d5d6efe7d0e09d1/docs/4.1/examples/sign-in/index.html#L19

https://github.com/twbs/bootstrap/blob/9c3146d0dd1320d9b1e902005d5d6efe7d0e09d1/docs/4.1/examples/sign-in/signin.css#L2-L14

This can be rewritten as follows:

```html
<body class="d-flex align-items-center h-100 py-5 bg-light text-center">
```

So should we reduce the custom CSS and replace it with a utility class?

MartijnCuppens:
Interesting, I'm more in favour of the utility classes. It stimulates developers to use the build in stuff instead of writing own custom css.

ysds:
> I'm more in favour of the utility classes.

I've updated it :)

andresgalante (approved):
Hi @ysds thanks a lot for this PR, and thanks @MartijnCuppens for helping reviewing it

PR #26689, opened 2018-06-08
Ignore percentage values for compare in maps

### LINKED ISSUE #24061: WARNING Potentially invalid value for $container-max-widths
Hello, I have warning when trying compile styles with this variables:
```
$container-max-widths: (
				sm: 540px,
				md: 720px,
				lg: 960px,
				xl: 92%
);
@include _assert-ascending($container-max-widths, "$container-max-widths");
```
Message:
```WARNING: Potentially invalid value for $container-max-widths: This map must be in ascending order, but key 'xl' has value 92% whose unit makes it incomparable to 960px, the value of the previous key 'lg' !```

I think what this block is wrong and should ignore the specified values in percentage
```
@each $key, $num in $map {
    @if $prev-num == null {
      // Do nothing
    } @else if not comparable($prev-num, $num) {
      @warn "Potentially invalid value for #{$map-name}: This map must be in ascending order, but key '#{$key}' has value #{$num} whose unit makes it incomparable to #{$prev-num}, the value of the previous key '#{$prev-key}' !";
    } @else if $prev-num >= $num {
      @warn "Invalid value for #{$map-name}: This map must be in ascending order, but key '#{$key}' has value #{$num} which isn't greater than #{$prev-num}, the value of the previous key '#{$prev-key}' !";
    }
    $prev-key: $key;
    $prev-num: $num;
  }
```

### PR DESCRIPTION
**Pull request by romanlex**
All conversation about it in https://github.com/twbs/bootstrap/issues/24061#issuecomment-395745350
I think this is the only solution

Short:
We cant compare percantage values with absolute values in map, but many people want use `%` in `$container-max-widths`

Fixes #24061

### CODE REVIEW

**Review by andresgalante — APPROVED** [review]
Thanks a lot @romanlex, this works and it'll give the chance to author to pick `%` or `px`.

There has been back and forwards and discussions for a very long time about the units of the breakpoints and the fact that we don't allow them to be `rem` or `em`.

I'll like @patrickhlauke and @mdo to decide what are the implications of allowing `%` for breakpoints.

Also if we merge this, there should be also a change in the docs here:
http://getbootstrap.com/docs/4.1/layout/grid/#grid-options

**Comment by Paladinium** [comment]
@andresgalante , @patrickhlauke @mdo : Would be really nice to see this change in one of the next minor releases...

File: scss/_functions.scss [hunk 1]
```diff
@@ -8,7 +8,7 @@
   $prev-key: null;
   $prev-num: null;
   @each $key, $num in $map {
-    @if $prev-num == null {
+    @if $prev-num == null or unit($num) == "%" {
```
errge: If I'm not mistaken, this only fixed % warning, if you use percentage at the biggest unit.

E.g. the bug report was: sm 500, md 700, lg 850, xl 100%.

But I have a use case, where I have only two categories: small: 80%, big: 720px.  So big is never too wide for reading, but small always has a nice padding from the screen (80% width).  This causes a warning still, because not the current unit is "%", but the previous unit.  Can we also ignore the warning in case of `unit($prev-num) == "%"`?

XhmikosR: @errge you should make a new issue

PR #26700, opened 2018-06-11
Prevent pointer-events on disabled anchor dropdown-item

### LINKED ISSUE #26695: Inconsistency between disabled anchors
Hi, I've noticed and was wondering whether there was a conscious decisions for:

`a.btn.disabled` having `pointer-events: none;`

https://getbootstrap.com/docs/4.1/components/buttons/#disabled-state

But anchors inside a Menu Item / split-button do not, and still have a pointer.

https://getbootstrap.com/docs/4.1/components/dropdowns/#disabled

I've also noticed that if we remove the `href` from an anchor (which makes the browser treat the anchor as non-interactable), the `disabled` style is actually overridden by:

```
a:not([href]):not([tabindex]) {
    color: inherit;
    text-decoration: none;
}
```

That's quite annoying.

I'd be happy to submit a PR to change both of these, if they weren't intentional decisions, otherwise I'll just override locally.

howardcox:
Fixes #26695

### DIFF
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -136,6 +136,7 @@
   &.disabled,
   &:disabled {
     color: $dropdown-link-disabled-color;
+    pointer-events: none;
     background-color: transparent;
     // Remove CSS gradients if they're enabled
     @if $enable-gradients {
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -66,6 +66,7 @@
   &.disabled,
   &:disabled {
     color: $list-group-disabled-color;
+    pointer-events: none;
     background-color: $list-group-disabled-bg;
   }
 


andresgalante:
Why are you qualifying the selector?

You could add this under:

```scss
&.disabled,
  &:disabled {
```
on line 137. Or am I missing something?

andresgalante (approved):
Hi @howardcox I've just patched your PR to have the desired result without the qualifying the selector.

Thanks a lot!

andresgalante:
Right, but even if `dropdown-item` is a btn I don't how it hurts to have the pointer events none there.

cc/ @patrickhlauke am I missing something?

howardcox:
Can this get merged at some point? It's missed 4.1.2 and 4.1.3

XhmikosR:
Ping @patrickhlauke

jquense:
I'd expect `list-group-item` to also have this for consistency yeah?

howardcox:
Ok e8ac9ca adds this to `_list-group.scss` too

XhmikosR:
Ping @patrickhlauke

patrickhlauke (approved):

PR #26702, opened 2018-06-11
Allows both OL and UL lists for tab buttons

tiesont:
Modifies tab.js to allow OL as a parent element for the tab list items.

Duplicates UL-based visual tests to provide OL-based versions.

Adds check for "show" class to tabs unit tests.

### DIFF
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -80,7 +80,7 @@ class Tab {
     const selector = Util.getSelectorFromElement(this._element)
 
     if (listElement) {
-      const itemSelector = listElement.nodeName === 'UL' ? Selector.ACTIVE_UL : Selector.ACTIVE
+      const itemSelector = listElement.nodeName === 'UL' || listElement.nodeName === 'OL' ? Selector.ACTIVE_UL : Selector.ACTIVE
       previous = $.makeArray($(listElement).find(itemSelector))
       previous = previous[previous.length - 1]
     }
@@ -141,7 +141,7 @@ class Tab {
   // Private
 
   _activate(element, container, callback) {
-    const activeElements = container && container.nodeName === 'UL'
+    const activeElements = container && (container.nodeName === 'UL' || container.nodeName === 'OL')
       ? $(container).find(Selector.ACTIVE_UL)
       : $(container).children(Selector.ACTIVE)
 
--- a/site/docs/4.1/components/navs.md
+++ b/site/docs/4.1/components/navs.md
@@ -35,7 +35,7 @@ The base `.nav` component does not include any `.active` state. The following ex
 {% endcapture %}
 {% include example.html content=example %}
 
-Classes are used throughout, so your markup can be super flexible. Use `<ul>`s like above, or roll your own with say a `<nav>` element. Because the `.nav` uses `display: flex`, the nav links behave the same as nav items would, but without the extra markup.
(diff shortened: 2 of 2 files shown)


js/tests/unit/tab.js
```diff
@@ -71,13 +73,15 @@ $(function () {
...
     $(pillsHTML).find('li:last-child a').bootstrapTab('show')
     assert.strictEqual($('#qunit-fixture').find('.active').attr('id'), 'profile')
+    assert.strictEqual($('#qunit-fixture').find('.show').attr('id'), 'profile')
 
     $(pillsHTML).find('li:first-child a').bootstrapTab('show')
     assert.strictEqual($('#qunit-fixture').find('.active').attr('id'), 'home')
+    assert.strictEqual($('#qunit-fixture').find('.show').attr('id'), 'home')
   })
 
   QUnit.test('should activate element by tab id in ordered list', function (assert) {
```
Johann-S: it's the only unit test needed here, the other aren't related

Johann-S (changes requested):
Some of your changes are irrelevant

XhmikosR:
@tiesont: can you take care of @Johann-S's comment?

XhmikosR:
Closing due to lack of response. You can comment here and we can re-open it if you decide to resume working on the requested changes.

tiesont:
I can fix this, I suppose, but I'm not sure what exactly is the issue here - I added those tests to make sure the plugin actually works with both UL and OL markup, which was the whole point of the original issue. Should I assume what you want is just the 

```javascript
assert.strictEqual($('#qunit-fixture').find('.show').attr('id'), 'home')
```

test? Or, do you only want the tests in tab.js?

Johann-S:
That's right, you just have to create one new test, with `<ol>` list, which show it works 👍

tiesont:
One _unit_ test (which should already be there) or one _visual_ test (which is also already there)? Is the issue that I modified the existing visual test to indicate which was UL-based and which was OL-based?

Johann-S:
@tiesont the is, you have unrelated changes.
Your fix should be:

image

Because ordered list works, but with your fix we use the correct selector, so the use case is already covered

tiesont:
Okay, I'll admit that I'm not all that great at using Git (or GitHub, for that matter). I pulled down the 4.1.3 source, merged the updated tab.js file, and then made the updates to allow OL as a selector. I reverted the changes I made to the unit test and visual test files to match what is in this branch.

Johann-S:
I updated your branch and removed unrelated changes.

For me we should add one thing, it's a little note on our documentation to show we accept `<ol>` too 🤔 
You can edit that file to do that: https://github.com/twbs/bootstrap/blob/v4-dev/site/docs/4.1/components/navs.md#base-nav

Johann-S (approved):

Johann-S:
@XhmikosR do you know why Github tells me there are conflicts here ? I resolved all the conflicts 😟

XhmikosR:
I don't see any conflicts here.

XhmikosR:
Ah, I see it now. It's shown if you switch Rebase and merge. Because there's a merge commit.

Johann-S:
Thank you @tiesont 👍

tiesont:
@Johann-S No problem. Thank you for being patient with me.

PR #26708, opened 2018-06-12
Remove inkscape data from brand svgs

MartijnCuppens:
Inkscape has added some redundant data to the brand svgs.
The attributes that are deleted are:
- `sodipodi:docname`
- `inkscape:version`
- `xmlns:svg`
- `xmlns:sodipodi`
- `xmlns:rdf`
- `xmlns:dc`
- `xmlns:cc`
- `xmlns:inkscape`

Also the groups are removed and the `fill` attribute is moved to the root element when the group had a fill attribute.


XhmikosR:
This has to be reproducible with svgomg for example.

MartijnCuppens:
[svgomg](https://jakearchibald.github.io/svgomg/) uses [svgo](https://github.com/svg/svgo) which only removes these attributes from the svg. It's probably also safe te remove these attributes from groups, but just to make sure, I asked it here svg/svgo#983.

I'll keep you posted.

XhmikosR:
I think I'm gonna commit my svgo config so that we can reproduce this all the time.

XhmikosR:
Oh, and thanks for opening the issue, I'm tracking it :)

XhmikosR:
OK, I think we ought to merge this, because we have no idea when if ever this will be sorted upstream...

XhmikosR (approved):

PR #25352, opened 2018-01-18
Input group tests and tweaks

mdo:
This PR is adding a new `docs/4.0/examples/tests/` folder to house more complex and dense kitchen sink examples of all the supported variations of a component. Perhaps there's a better location, so I'm open to moving this in the future.

For now, I'm using this to test additional input group tweaks in hopes of fixing our rounded corner issue from #25110. I'll split those commits out after verifying what we support is functioning as intended.

---

screen shot 2018-01-17 at 9 10 11 pm

Issue #25120, opened 2017-12-29
Feature Request: .input-group-between


rkieru:
With Beta 3 bringing `.input-group-append` and `.input-group-prepend` it seems like an opportune moment for one additional group structure: `-between`.  This pattern would ideally be for situations where multiple input fields need to be more visually connected.  Such patterns might include:

- Inputting dimensions: `[width][x][height]`
- Inputting time: `[hour][:][minutes] or [mm][/][yyyy]`

Currently this pattern can be achieved with existing CSS; with the exception that you get either a double border on the left or right of the input field depending on whether you used `-prepend` or `-append`.  Adding this additional class would be as simple as:

- Modifying class declarations for `..input-group-append, .input-group-prepend { }` to include the new wrapper.
- Removing the `border-radius` as well as the left and right `border` of `.input-group-between .input-group-text { }`

I have created an example on CodePen to illustrate the current implementation (using both `.input-group-prepend` and `.input-group-append`) which results in the aforementioned double-border, as well as the proposed `.input-group-between`.

https://codepen.io/rkieru/pen/Mrmagb

Kramb:
Combining both `.input-group-append` and `.input-group-prepend` on the `.input-group-text` wrapper produces the same result.

[input-group-text Between Two Inputs](https://codepen.io/Kramb/pen/eyWpPM)

rkieru:
@Kramb - I had not considered simply applying both classes to the wrapper; that's a great solution within the existing CSS... but do you think it is a little confusing in terms of comprehension?  Defining something as both an append and a prepend seems less coherent than offering the clarity of a distinct class.

mdo:
Moving to a single class for that is dependent on how complex the selectors get. We might end up with that dual-class situation, but hopefully not :). Will look into it more after v4.0 ships.

vanillajonathan:
Can we add a example that uses both `.input-group-prepend` and `.input-group-append` to the documentation?

vanillajonathan:
> Moving to a single class for that is dependent on how complex the selectors get. We might end up with that dual-class situation, but hopefully not :). Will look into it more after v4.0 ships.

@mdo Now that v4.0 have shipped, have you looked into it?

vanillajonathan:
@mdo, Now with Bootstrap 4.5.0 released that updated the documentation and with pull request #30180 merged maybe it is time to revisit this?

mdo:
Shipped with #31984 it seems! :)

PR #26710, opened 2018-06-14
Input group - fix margins

vchebotarev:
Hi.
With the help of input-group i want to make something like this (perfect variant already fixed)
joxi_screenshot_1528964290145
But there is a problem with double borders
joxi_screenshot_1528963219807
So i've fixed it, and now it looks like this - without double margins
joxi_screenshot_1528965389632

### DIFF
--- a/docs/4.1/components/input-group.md
+++ b/docs/4.1/components/input-group.md
@@ -25,6 +25,14 @@ Place one add-on or button on either side of an input. You may also place one on
   </div>
 </div>
 
+<div class="input-group mb-3">
+  <input type="text" class="form-control" aria-label="Min value" placeholder="Min value">
+  <div class="input-group-prepend">
+    <span class="input-group-text">to</span>
+  </div>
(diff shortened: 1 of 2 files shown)


TheJoppley:
Hi, where are you from?

andresgalante (approved):
Thanks @vchebotarev

andresgalante:
@vchebotarev I've patch your PR to add an example of an add on between 2 inputs on the docs.

ysds:
This change will cause a problem when there is validation feedback.

@andresgalante Please check with the test cases in #25352

ysds:
Please see also #25120

andresgalante (changes requested):
Thanks @ysds I'll change my approval for this PR, sorry @vchebotarev It will cause problems for feedback validation described here #25352 and #25120

Thanks again @ysds for keeping an eye on it

mdo:
Thanks, but we'll pass on this for now as it starts to take us in an even crazier direction with input group support for multiple components.

PR #26714, opened 2018-06-15
Easy merging of new grid breakpoints and containers

omnichronous:
This change follows exactly what was done for the `$grays`, `$colors`, and `$theme-colors` lists in #23260. This allows for easy addition of new breakpoints or grid containers also.


MartijnCuppens:
I don't think this is a good idea. With this implementation it's not possible easily remove breakpoints or place a breakpoint somewhere in between (see https://codepen.io/MartijnCuppens/pen/NzyNaL).

mdo:
@MartijnCuppens We use this approach elsewhere and have content in the docs for how to add, modify, or remove items from a Sass map. http://getbootstrap.com/docs/4.1/getting-started/theming/#maps-and-loops

MartijnCuppens:
Yes indeed, changed my mind about this in the meantime. This LGTM.

mdo (approved):

PR #26722, opened 2018-06-19
Adding RTL support.

mohamedragaey:
Adding RTL support for bootstrap4.


andresgalante:
Hi @mohamedragaey, thanks for this PR. We've talked about RTL many times, for example #24807 #24808

This solution is smart but it forces contributors to know that they should replace `left` and `right`. To me it's a big barrier. On the other hand I don't see a good solution until [CSS Logical Properties and Values](https://drafts.csswg.org/css-logical/) are implemented in browsers.

@mdo what do you think about this one?

mohamedragaey:
Hi @andresgalante, thanks for your reply.
I did have a look on both of the two examples that you mentioned above before I start this pull request, and the duplication of every rule is not a good solution.
I think the value added here is worth that contributer would know to replace the `left` with `$default` and `right` with `$opposite` in the contribution guide.

If we waited untill CSS logical properties and values to be implemented in browsers we won't use bootstrap RTL out of the box not in the near future
@mdo please let us know what you think.

Aeon:
👍

mohamedragaey:
@andresgalante @mdo 
any news on this PR ??

DJafari:
@andresgalante you say right, but when browsers support it, another time is required to update users of their browsers, i think it will take a year ... and i hope bootstrap version is 6 :-D

i think this solution best solution until  browsers support `start` and `end`

mohamedragaey:
@andresgalante @mdo
any news on this PR ??

XhmikosR (changes requested):
Please fetch upstream, fix the conflicts and drop any dist files from your commits.

XhmikosR:
@mohamedragaey: that's not a rebase.

mohamedragaey:
@XhmikosR 

> @mohamedragaey: that's not a rebase.
Is every thing is ok now ??

XhmikosR:
Nope. That is not a rebase nor you have dropped the dist file changes.

mohamedragaey:
@XhmikosR sorry but do you mean i should add dist files to gitignore or delete them and not commit them

XhmikosR:
Please read https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md#pull-requests

No dist files, and a proper rebase is needed for this to be reviewed eventually.

XhmikosR:
@twbs/css-review: how do you want to proceed with this? It's unmergeable as it is, also we have #27027

MartijnCuppens:
We should definitely have a look how to support RTL in the future, but at the moment we have these issues:
- RTL support will slow down our development, every future PR that's going to be made will need to be checked for RTL support
- If we're going to continue with RTL, we'll need a team of people that will review the PRs, at the moment we don't have any Bootstrap team members that are familiar with RTL (I think).
- This PR adds a layer of complexity to Bootstrap, if someone wants to look into the code and sees for example `padding-#{$default}`, it's not immediately clear what `$default` means.

MartijnCuppens:
I'm going to close this because this approach is going to be way too difficult to maintain. RTL is definitely something we should look in to in the future, but we need to find a simpler solution for it.

PR #26723, opened 2018-06-19
remove reference to breakpoint-specific gutter widths which are not implemented

joshwand:
see #22944 and #24620


mdo:
Cherry-picked this into #26732. Thanks!

PR #26658, opened 2018-06-03
Update dependencies.

XhmikosR:
@Johann-S: the babel ones is what I haven't tested thoroughly. We need to read the release notes in case we need to update our config or something. Also, make sure you are using node 10.3.0 which ships with npm 6.3.1.

PR #26739, opened 2018-06-21
Updates

### LINKED ISSUE #26401: Make our browserslist more specific
https://github.com/babel/babel/pull/7790/files

/CC @twbs/team for thoughts

XhmikosR:
I know I should split this but I don't have the time so bear with me.

/CC @Johann-S @mdo for the `not dead` browserslist addition; this in fact saves us quite some CSS size.

Fixes #26401 along the way.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,15 +5,13 @@
     "browser": true,
     "es6": true
   },
-  "extends": "eslint:recommended",
-  "plugins": ["compat"],
+  "extends": ["eslint:recommended", "plugin:compat/recommended"],
   "rules": {
     // Possible Errors
     "no-await-in-loop": "error",
     "no-extra-parens": "error",
     "no-prototype-builtins": "error",
     "no-template-curly-in-string": "error",
-    "compat/compat": "error",
     "valid-jsdoc": "error",
 
     // Best Practices


XhmikosR:
This might require #26658 to be merged first.

XhmikosR:
IIRC it's IE mobile not desktop. Did you verify?

mdo:
> IIRC it's IE mobile not desktop. Did you verify?

I have not—I don't have my VMs setup on the computer I have access to this weekend. I assumed you would test it before suggesting it. 😜 Everything I see in my research just says IE10 requires `-ms` prefix. Couple more places just speak of IE10:

- http://the-echoplex.net/flexyboxes/
- https://medium.com/css-mine/flexbox-how-to-deal-with-older-browsers-fbf6eb8c7a65

Tried setting up some VMs on my Windows 10 PC, but ran into error after error with Virtualbox.

XhmikosR:
But if IE10 **is** dead, why do we need to support it still?

I will play with browserslist in another PR, I'll remove the related patch from here. Maybe the order of entried in browserslist affects the results.

But I still don't think we should support dead browsers. I know we can't do it in a minor release, but I;m throwing it out there.

mdo (approved):

PR #26744, opened 2018-06-22
Fix color in nested list group

someden:
Example with problem: https://codepen.io/someden/full/EbJWQd


mdo:
This is the wrong solution to the referenced problem. If we do end up resetting this `color`, it'd be on the `list-group-item` and `list-group-action`, and it'd be with a variable that sets a specific value and not `initial`.

You're welcome to try again in this PR, or you can open an issue with the demo and we can revisit for a v4.1.3 release.

mdo:
Closing out for now.

PR #26785, opened 2018-07-01
Update gems.

XhmikosR:


XhmikosR:
I'm thinking, we could get rid of the redirect layout. It's only supposed to be seen when JavaScript is disabled.

Johann-S (approved):
I'm not the best in Ruby and you know our docs better than me 😄

XhmikosR:
@mdo: what do you think about removing the redirect layout?

mdo:
@XhmikosR If it's unused, remove it.

PR #26800, opened 2018-07-03
Added display heading responsive class

ItaloBC:
This allows to user display headings that will automatically shrink it's size (if the user opts to) using `.display-responsive`. This will only work on small screens (xs, sm).

### DIFF
--- a/docs/4.0/content/typography.md
+++ b/docs/4.0/content/typography.md
@@ -109,7 +109,7 @@ Use the included utility classes to recreate the small secondary heading text fr
 
 ## Display headings
 
-Traditional heading elements are designed to work best in the meat of your page content. When you need a heading to stand out, consider using a **display heading**—a larger, slightly more opinionated heading style.
+Traditional heading elements are designed to work best in the meat of your page content. When you need a heading to stand out, consider using a **display heading**—a larger, slightly more opinionated heading style
 
 <div class="bd-example bd-example-type">
   <table class="table">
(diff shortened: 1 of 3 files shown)


andresgalante:
Hi @ItaloBC Thanks a lot for your PR. 

This idea is interesting but really opinionated. By adding the `display-responsive` class a title would automatically change itself on smaller viewports. When dealing with responsive states Bootstrap usually adds responsive namespaces like `-md` `-xs` etc so we let the author decide what he wants in any given breakpoint. Maybe they don't want to make it smaller on small devices but larger. 

Having said that, the display headers does not support different sizes at different breakpoints at this moment and to be honest I would argue that they are unnecessary.

andresgalante (commented):
Hi @ItaloBC Thanks a lot for your PR.

This idea is interesting but really opinionated. By adding the display-responsive class a title would automatically change itself on smaller viewports. When dealing with responsive states Bootstrap usually adds responsive namespaces like -md -xs etc so we let the author decide what he wants in any given breakpoint. Maybe they don't want to make it smaller on small devices but larger.

Having said that, the display headers does not support different sizes at different breakpoints at this moment and to be honest I would argue that they are unnecessary.

ItaloBC:
Yeah, I missed the point @andresgalante. At least, the implementation should be responsive first, but from what I perceive `.display` sizing it's not a mobile first experience.

While it could be implemented the other way around, then it could collision with the `h1-h6` headings. And, you could replicate the `display` by just adding `font-size-light` as a class.

Maybe a helper class to make a heading responsive from the small screens to large would be better instead of a static `display` heading. That way the `h1-h2` headings would work as always on small screens, bu adding `display-sm/md/lg` class would make it larger on that viewport and higher.

If out shoot me out I'll stop.

andresgalante:
Hi @ItaloBC as I mentioned before, I don't think I'd like to add responsive headers, but if @mdo likes the idea, I think the solution would be to add responsive prefixes to each display heading.

Issue #26760, opened 2018-06-26
Modal positioning issue


szabbenjamin:
Open the page which contains a link to bootstrap modal.
Scroll down until the Chrome address bar goes to hide.
Click the link which open modal.
Try to inspect an element in this modal - you have a great chance to see the issue.

I attach the result: https://imgur.com/sNvztc4

It's an existing issue with bs 3 and 4.
I use with Chrome 67.0 on Huawei P20 Pro, Android 8.1

I tried this on Chrome Canary (version 69) - Canary can't solve this problem.

jvitela:
I experience the same issue, specially when the modal has long content as the one in the docs
https://getbootstrap.com/docs/4.1/components/modal/#scrolling-long-content

szabbenjamin:
Thank you, it seems to be better

PR #26811, opened 2018-07-06
Fixes modal positioning on android (#26760)

jvitela:
When address bar is hidden, sometimes the visual position of the controls is out of sync with its logical position.

### DIFF
--- a/site/docs/4.1/getting-started/browsers-devices.md
+++ b/site/docs/4.1/getting-started/browsers-devices.md
@@ -121,9 +121,7 @@ For a list of some of the browser bugs that Bootstrap has to grapple with, see o
 
 ## Internet Explorer
 
-Internet Explorer 10+ is supported; IE9 and down is not. Please be aware that some CSS3 properties and HTML5 elements are not fully supported in IE10, or require prefixed properties for full functionality. Visit [Can I use...](https://caniuse.com/) for details on browser support of CSS3 and HTML5 features.
-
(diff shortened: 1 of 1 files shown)


jvitela:
@MartijnCuppens, @mdo Can somebody review this changes? They've been sitting here for a while :-(

mdo:
I don't have an Android on me at this time to test unfortunately. Anyone else?

XhmikosR:
@mdo: You can use Browser Stack.

@MartijnCuppens: can you review please?

MartijnCuppens:
@jvitela, can you also add this fix for `.modal-backdrop`?

MartijnCuppens:
This is better, but now I see the backdrop isn't stretching at the bottom when address bar is hidden. Can you check if you can fix this @jvitela?

jvitela:
Using **viewport units** for sizing `.modal-backdrop` solves the resizing issues and also eliminates the undesired effect of the modal jumping up and down after the address bar appears/disappears.

mdo:
@MartijnCuppens If you confirmed latest changes fix your issues, this looks good to me. I don't have an Android phone right now to test myself.

MartijnCuppens:
Nope, `height: 100vh` will make the modal scrollable if the address bar is shown. I'll check tomorrow if there's something we can do about this.

MartijnCuppens:
Something went wrong here, we're looking into it

PR #26823, opened 2018-07-08
Dismissable badges

mdo:
Updated version of #23784 with conflicts resolved and some docs adjustments.

/cc @andresgalante


docs/4.1/components/badge.md
```diff
@@ -80,3 +80,23 @@ Using the contextual `.badge-*` classes on an `<a>` element quickly provide _act
...
+{% capture example %}
+<span class="badge badge-primary">
+  Dismissable badge
+  <button type="button" class="close" aria-label="Dismiss">
+    <span aria-hidden="true">&times;</span>
+  </button>
+</span>
+<a href="#" class="badge badge-secondary">
+  Dismissable link badge
+  <span role="button" class="close" aria-label="Dismiss">
```
andresgalante: @mdo Thanks for pulling this one in. The error on the build is "An element with the attribute "role=button" must not appear as a descendant of the "a" element."

Since we need to bring the semantics of a button to the span here, can't we skip this example and leave just the one that has a button?

mdo: @andresgalante I changed to the `<span>` to try to get around the same error for having a `<button>` in a `<a>`. Either way, it errors out for us.

scss/_badge.scss
```diff
@@ -24,6 +24,15 @@
   &:empty {
     display: none;
   }
+
+  .close {
+    padding-right: $badge-padding-x;
+    padding-left: $badge-padding-x;
+    margin-right: -$badge-padding-x;
+    font-size: inherit;
+    color: inherit;
```
XhmikosR: My only comment is that this doesn't respect the color for `focus`/`hover`.

XhmikosR (changes requested):

site/docs/4.1/components/badge.md
```diff
@@ -72,3 +72,24 @@ Using the contextual `.badge-*` classes on an `<a>` element quickly provide _act
...
+{% capture example %}
+<span class="badge badge-primary">
+  Dismissable badge
+  <button type="button" class="close" aria-label="Dismiss">
+    <span aria-hidden="true">&times;</span>
+  </button>
+</span>
+<a href="#" class="badge badge-secondary">
+  Dismissable link badge
+  <span role="button" class="close" aria-label="Dismiss">
```
patrickhlauke: this...looks pointless. why have a role="button" on a span that's not focusable? would need to do testing, but gut feeling is that no AT will announce/do anything sensible with it when it's part of a link - at best, you'd hear something like "dismissable link badge...dismiss...link". in some cases, the aria-label won't be announced at all, so just "dismissable link badge...link"

patrickhlauke (changes requested):

site/docs/4.1/components/badge.md
patrickhlauke: maybe i'm missing what the whole point of that thing is...why would you want the whole badge to be a link? what's the expected behavior here?

PR #26831, opened 2018-07-09
Add missing role="button" in .navbar's .dropdown-toggle

### LINKED ISSUE #26830: [Docs] Role "button" in Navbar dropdowns
hi,
in the first example
https://getbootstrap.com/docs/4.1/components/navbar/#supported-content
.dropdown-toggle has a "role="button"
in the second example
https://getbootstrap.com/docs/4.1/components/navbar/#nav
"role=button" is missing

which is the correct version?

XhmikosR:
Fixes #26830

### DIFF
--- a/docs/4.1/components/navbar.md
+++ b/docs/4.1/components/navbar.md
@@ -185,7 +185,7 @@ You may also utilize dropdowns in your navbar nav. Dropdown menus require a wrap
         <a class="nav-link" href="#">Pricing</a>
       </li>
       <li class="nav-item dropdown">
-        <a class="nav-link dropdown-toggle" href="#" id="navbarDropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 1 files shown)


patrickhlauke (approved):

patrickhlauke:
oops, just realised perhaps intent was not to merge due to closeness of new release...sorry if that was the case/if that creates issues now :(

XhmikosR:
Nah, this is good I believe; it's a trivial change, don't worry.

thgh:
Why does Bootstrap use `<a>` instead of `<button>` for a button functionality?

patrickhlauke:
mainly, for ease/consistency of styling. as it's then still exposed correctly thanks to the `role="button"`, it's not a major issue.

of course, authors can use an actual `<button>`, e.g. something like

```
<button class="btn btn-link nav-link dropdown-toggle" href="#" id="navbarDropdownMenuLink" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Dropdown link</button>
```

but currently the baseline for the text in the links and the text in the button don't always correctly line up.

Issue #26836, opened 2018-07-09
Tooltip flickers while inside .table-responsive


simPod:
When using Tooltips inside .table-responsive, they flicker as can be seen on this [codepen example](https://codepen.io/anon/pen/oMvdpY).

It can be fixed by placing `overflow: visible` on `.table-reponsive` but that breaks table's responsivity.

simPod:
I solved it by changing `{ boundary: 'window' }` but shouldn't the default value then be reconsidered?

alecpl:
The default value change is also a solution for #25428.

Johann-S:
You found the solution and it's not a bug so I'll close this issue too

simPod:
But it is a bug when used with table responsive. 

The thing that I was able to solve it doesn't mean it is not a bug. 

The bug is something that behaves not according to the specification. And the specification states that tooltip is created and steadily visible when the underlying object is hovered. But in this case tooltip flickers. So it is a bug.

Johann-S:
Yep but under the hood if you use our Tooltips/Popovers or Dropdown, you use Popper.js, so you have to dig in Popper.js documentation and find a solution to your problem, which isn't the common use case.

At least we can add a callout in the `.table-responsive` to help folks

FezVrasta:
There's an open issue on Popper.js to make the default value even smarter, that should also fix this annoyance, but there's not any ETA unfortunately.

warmwhisky:
> I solved it by changing `{ boundary: 'window' }` but shouldn't the default value then be reconsidered?

Hey would you let me know where I use `{ boundary: 'window' }` I am currently having this issue and absolutely need responsive tables.

simPod:
@warmwhisky check my PR https://github.com/twbs/bootstrap/pull/26844

warmwhisky:
> @warmwhisky check my PR #26844
Thank you :)

cjsteven:
Hi guys I tried to put some media query in table-responsive. Like table-responsive-lg. It worked for me. I do hope also it might help you.

Issue #25428, opened 2018-01-24
Popover positioning regression


alecpl:
I built multi-level menu using Bootstrap 4 popovers. It worked like a harm with Popper 1.12.6 (Bootstrap 4-beta2), but since 1.12.7 menu positioning is broken. Tested with all versions including 1.13.0-next, and Bootstrap 4.0.

Looks like the second-level popover is placed incorrectly. It works if I use Bootstrap 4.0 with Popper 1.12.6.

https://codepen.io/anon/pen/mpYmVX

I've found out that removing my `.popover-body { overflow-x: hidden; }` fixes the issue, but I need the popover scrollable.

I created also https://github.com/FezVrasta/popper.js/issues/532, but I can't provide Popper-only test case.

mdo:
We'll have to look into why we're including that overflow.

alecpl:
Sorry if I wasn't clear. That overflow style is added by my custom css not Bootstrap. However, I think it's a valid use-case even if you didn't think about popovers containing buttons that open another popovers before.

tompat:
I also have a popover placement issue from popper 1.12.7
Still broken with popper 1.12.9
1.12.6 works well

After few investigation, it seems that if the element triggering the popover has a container with an overflow set to scroll or hidden and the popover doesn't have enough space to position itself in the container with the right placement then the popover rollback to an unexpected position.

Bootstrap 4.0.0 + popper 1.12.9 + container with overflow set to scroll > unexpected position
https://codepen.io/anon/pen/RxXKgY

Bootstrap 4.0.0 + popper 1.12.6 + container without overflow > expected position
https://codepen.io/anon/pen/opKBQM

alecpl:
Create a new Popper ticket with simpler test case https://github.com/FezVrasta/popper.js/issues/535

alecpl:
A workaround is to set `boundary: 'window'` (or 'viewport') in popover config. Maybe this should become a default for popover if Popper.js does not fix the issue on their side. For more see https://github.com/FezVrasta/popper.js/issues/535

@Johann-S you think you can fix that, or you rather wait for @FezVrasta for an opinion?

Yacie:
@ tompat you helped me bro.
With the sroll issue, working as expected

Johann-S:
@alecpl we have boundary option in our Popover/Tooltip plugins see : https://getbootstrap.com/docs/4.0/components/popovers/#options

Is it still an issue ?

[...]

PR #26842, opened 2018-07-10
Change default boundary value to `window`

simPod:
Trying to address these
https://github.com/twbs/bootstrap/issues/25428
https://github.com/FezVrasta/popper.js/issues/535
https://github.com/twbs/bootstrap/issues/26836


simPod:
Needs fixing tests, but not sure if it's viable. It's BC break

Johann-S:
The default value is `scrollParent` because it's the default value specified by Popper.js so we won't change that, unless Popper.js change their default value.

You have the hability to change this value, so there are no need to change that

simPod:
Only the thing that it is the bug 🐛

PR #26848, opened 2018-07-10
Ensure calling dispose on a popover will not set the title of the popover next time it is created

### LINKED ISSUE #26847: Calling dispose on a popover may set the title of the popover next time it is created
If the element a popover is attached to has a title, manually calling .popover('dispose') causes the pop over to have a title in the pop over which matches the title attribute of the element.

This can actually replace the title which is specified in the options. See codepen: https://codepen.io/jarstelfox/pen/EpxdrX

BS Version: 4.1.1

jarstelfox:
Fixes #26847

After doing some digging it looks like popover was previously checking for destroy, but this has been renamed to dispose in v4


js/src/popover.js
```diff
@@ -150,7 +150,7 @@ const Popover = (($) => {
         let data = $(this).data(DATA_KEY)
         const _config = typeof config === 'object' ? config : null
 
-        if (!data && /destroy|hide/.test(config)) {
+        if (!data && /dispose|destroy|hide/.test(config)) {
```
Johann-S: Can you remove `destroy` because it doesn't exist anymore

Johann-S (changes requested):

jarstelfox:
I've addressed your requested changes

vsn4ik:
@Johann-S Review please!

Johann-S (approved):

bensamin:
good

PR #26861, opened 2018-07-12
Use text-emphasis-variant mixin for all contextual colors

zce:
Let all the contextual colors can be applied on the link.


mdo:
Thanks, but we'll pass on this. I want to keep as many utilities as simple as possible.

PR #26893, opened 2018-07-16
Add query string to the start_url to track how often app is launched

coliff:
Useful suggestion from Google in the Web App Manifest documentation.
https://developers.google.com/web/fundamentals/web-app-manifest/#start-url


XhmikosR (approved):

XhmikosR:
I guess it doesn't hurt to have this.

XhmikosR:
Although we are going to remove the service worker so A2HM won't be available probably.

PR #26922, opened 2018-07-19
Add transition to property-blacklist

MartijnCuppens:
In each PR, we need to check manually if the `@transition` mixin is used instead of the `transition` property. By adding this stylelint rule, the `css-lint` command will take care of this.


XhmikosR (approved):

XhmikosR:
@MartijnCuppens: there are more instances of transition in docs.

MartijnCuppens:
Changed it.

The stylelint check also happens on the custom css files. There are several ways we can handle this:
1. Leave it as it is now (Add `/* stylelint-disable-next-line property-blacklist */` above the transition declaration)

2. Also add this query in the css files:
```css
@media screen and (prefers-reduced-motion: reduce) {
  .selector {
    transition: none;
  }
}
```
3. Change stylelint config in `site/docs/4.1/examples/.stylelint`

Which way do you want to go with this?

XhmikosR:
Hmmm... not sure to be honest.

@andresgalante @mdo  thoughts?

XhmikosR:
Oh, and @MartijnCuppens jump on Slack if you want :)

mdo (approved):

XhmikosR (approved):
Thanks!

PR #26948, opened 2018-07-24
feat(Dropdown): dropdown menu fixed positioning

vepanimas:
PopperJS can position poppers in a "fixed" mode. I think this feature could be useful and it would be great to support it.


js/src/dropdown.js
```diff
@@ -297,7 +298,8 @@ const Dropdown = (($) => {
           preventOverflow: {
             boundariesElement: this._config.boundary
           }
-        }
+        },
+        positionFixed: $(this._menu).hasClass(ClassName.POSITION_FIXED)
```
Johann-S: I'm a bit mix-matched because with that Popper.js will apply `position: fixed` but that's already applied by our CSS class `.position-fixed`, what's the difference between ?

vepanimas: This param affects `top` and `left` inline style calculataion. If we simply add .position-fixed to a dropdown menu, it would be rendered somewhere in the top left corner of the window, because popper won't know anything about it. 
So, we don't need any classes or styles, we need to configure the PopperJS instance correctly. I'm not sure, what's the best way to do it. Do you think it might be better to use `data-position-fixed' instead of classes?

Johann-S: Yep according to your answer it's better to use a parameter instead of a CSS class, because it'll add style

vepanimas:
@Johann-S, I reworked it. Please, could you review it again?

Johann-S:
LGTM 👍 Can you make a live demo please ?

vepanimas:
@Johann-S, yep, here it is https://z6lr2rw3rx.codesandbox.io/. It is very simple.

vepanimas:
@XhmikosR, @Johann-S any updates? can we merge that?

XhmikosR:
@vepanimas: can you rebase one more time?

XhmikosR:
@Johann-S can you decide about this? I retargetted to master BTW.

XhmikosR:
Closing since the fork is deleted.

PR #26775, opened 2018-06-28
Use local workbox libraries

MikeyBeLike:
cc @XhmikosR 

https://github.com/twbs/bootstrap/issues/26600

I couldn't get tests to pass locally, dunno why?

PR #26958, opened 2018-07-25
Remove service worker

### LINKED ISSUE #26600: docs service worker improvements
1. load workbox locally and not from Google's CDN
2. switch to `generateSW()` maybe, if we can find a way to fix the previous point; IIRC you can't configure the libs location with `generateSW()`
3. update mechanism?

/CC @Lyricalz  @gauntface

### LINKED ISSUE #26868: Nuke service worker
/CC @mdo

mdo:
Starts the process to delete this code, which fixes #26868. This is only have the battle though—we will need to unregister the service worker as well (perhaps even before this?).

References:
- Originally added in #23101
- Updated w/ workbox in #23533

Closes #26868, closes #26600 (nullifies it at least) and its related PR, closes #26775.

Solution to the unregistering it likely at https://stackoverflow.com/questions/33704791/how-do-i-uninstall-a-service-worker?

TODO:

* [x] unregister service worker


XhmikosR:
We definitely need to unregister the service worker.

XhmikosR:
@gauntface @jeffposnick: can you guys help us out here, please? :)

What is the proper way to remove the service worker? Or any SW we have registered for that matter.

jeffposnick:
https://stackoverflow.com/a/38980776/385997 has some info on the best practices for deploying a no-op service worker. That's usually the best first step.

And if there are specific issues that you ran into with your service worker implementation, we're happy to try to help if we can. Deploying the no-op in the meantime to get back to a "base" state could be preferable to making incremental tweaks.

XhmikosR:
@jeffposnick: thanks for the reply!

The problem is that we refactored some files and the old service worker file (sw.js) couldn't be found.

I'm also having issues with html being cached on a project I'm working on, and it takes like 1 day until the service worker shows the new HTML content.

I could definitely use some help with this but I can't mention all the details in public. I'm just using `generateSW()`, so was expecting this to work out of the box.

jeffposnick:
Okay, well, using the advice in https://stackoverflow.com/a/38980776/385997 to deploy a new `sw.js` should definitely help in the short term.

XhmikosR:
Thanks for the reply @jeffposnick. I was just hitting the issue where I had to close the browser window for the service worker to update. I'm not confident if I should use `skipWaiting` though. But anyway this is sort of unrelated to this PR, which is not the right way to remove the service worker anyway.

I'll try to fix this branch when I have some time.

XhmikosR:
@jeffposnick OK, I made some progress here, but I'm not sure about this: the cached service worker will look for `sw.js`; should I explicitly include the unregister code in that file and not bundle it with the rest?

jeffposnick:
> @jeffposnick OK, I made some progress here, but I'm not sure about this: the cached service worker will look for domain/sw.js; should I explicitly include the unregister code in that file and not bundle it with the rest?

If you're goal is to ship a no-op SW to effectively neutralize any existing SWs that were previously registered, then I'd

a) Update the contents of the existing SW URL (e.g. `https://example.com/sw.js`) to contain the no-op code found at https://stackoverflow.com/questions/33986976/how-can-i-remove-a-buggy-service-worker-or-implement-a-kill-switch/38980776#38980776. Don't change this URL or anything—it's really important that you keep the URL of the service worker consistent to make sure that updates happen for the existing registrations.

b) Just stop registering new SWs by updating the code in your client pages to remove `navigator.serviceWorker.register()`.

XhmikosR:
Thanks for the reply!

I think point "b" is sorted now, so I just need to take care of point "a".

XhmikosR (approved):

XhmikosR:
@mdo: I'm gonna merge this. For v4.0 docs I'd say you can replace the existent sw.js with this one and should work in theory, but I haven't tested it so we might need to run some tests and even rebuild the docs JS for 4.0.

PR #26959, opened 2018-07-25
Add new variables for form feedback tooltips

### LINKED ISSUE #26886: Add new variables for form validation tooltips
With v4.1.3, we'll be updating the source Sass to use tooltip vars, but these should be their own and remapped for improved customization.

See #26816.

mdo:
Fixes #26886


mdo:
Ah, build is failing because tooltip variables are coming later in the file. Could move, or we could punt on doing this.

PR #26963, opened 2018-07-25
add .rtl files for supporting right-to-left languages like persian

thezahedi:


Johann-S:
There is a PR which handle that: https://github.com/twbs/bootstrap/pull/26722

So close, but thank you 👍

PR #26976, opened 2018-07-26
Allow multiple ways of padding declaration for modal header

### LINKED ISSUE #26967: Modal header padding
It is not possible to assign a different vertical and horizontal padding in the modal header, this breaks the margin of the ".close"

.close {
    padding: $modal-header-padding;
    // auto on the left force icon to the right even when there is no .modal-title
    **margin: (-$modal-header-padding) (-$modal-header-padding) (-$modal-header-padding) auto;**
}

If you put "padding: .5rem 1rem" the result of the margin would be:
`margin: -0.5rem 1rem -0.5rem 1rem -0.5rem 1rem auto;`

A simple solution would be to split the padding vertically and horizontally. It would look like this:

_variables.scss
change:
$modal-header-padding: 1rem;
for:
$modal-header-padding-y: 1rem; => .5rem;
$modal-header-padding-x: 1rem;

_modal.scss
.modal-header {
    ...
    **padding: $modal-header-padding-y $modal-header-padding-x;**
    ...
}

.close {
    **padding: $modal-header-padding-y $modal-header-padding-x;**
    // auto on the left force icon to the right even when there is no .modal-title
    **margin: (-$modal-header-padding-y) (-$modal-header-padding-x) (-$modal-header-padding-y) auto;**
 }

MartijnCuppens:
Fixes #26967 

I've made a new function `split-value` which return the `top/right/bottom/left` values from `margin` or `padding`. I used this function to split up the negative margins for `.modal-header .close`.

cc @proactivo


mdo:
This feels over-engineered perhaps? The solution I suggested in the issue was something like this:

```scss
$modal-header-padding-y: 1rem !default;
$modal-header-padding-x: 1rem !default;
$modal-header-padding: $modal-header-padding-y $modal-header-padding-x !default

.modal-header {
  padding: $modal-header-padding;
}
```

Skips the need for a function altogether while matching the approach of many other components.

MartijnCuppens:
Jup, changed it.

mdo (approved):
Thanks!

PR #26987, opened 2018-07-29
Example Cover: Update cover.css - allow extended content

kwaaui:
This change must go along with removing width and height classes (w-100 h-100) from wrapper div.container on index.html (line 21).
It allows longer content to fit seamlessly inside MAIN element, stretching body shadow accordingly.
I know this example is for building "simple and beautiful home pages", but it might be useful to have them longer.


wolfy1339:
They should probably been one PR then

ysds:
This and #26986 doesn't work in IE

PR #26989, opened 2018-07-29
Add new docs Versions page

### LINKED ISSUE #26883: Feature Request: Add 'All Versions' link to dropdown docs menu for better UX/accessibility
This is a feature request, possibly for Bootstrap 4.2 and onwards. Whilst the [https://getbootstrap.com/](https://getbootstrap.com/ ) website **does have a dropdown in the top right** for the latest versions of each version of Bootstrap, this isn't always going to facilitate for older versions of Bootstrap, or versions which are being used in some themes these days (see [here as an example](https://themeforest.net/category/ecommerce/magento?compatible_with=Bootstrap&sort=price-asc)).

**The current docs selector:**

screen shot 2018-07-15 at 16 05 29

There are plenty of themes out there which I've used in the past which use **Bootstrap v3.0.0**, and whilst this is older than 3.3.7, it's not always possibly to upgrade and we shouldn't be limited to not being able to access documentation for each of these.

**A suggestion, and something to consider**, once **Bootstrap 4.2** is released I'm not sure whether another link in the dropdown for versions will be added, or whether the 4.1 link will be replaced, but either way, whatever option that is chosen could clutter and make it hard to find older versions if you're less technologically advanced.

What you could do, is something very similar to what Bulma are doing, I think it's simple, straightforward and allows you to link to all of your versions, and allows us developers and designers to quickly find the **exact** version that we need, see the example below:

**Example: https://versions.bulma.io/**

screen shot 2018-07-15 at 16 04 32

**Maybe consider adding an 'All Versions' link to the current dropdown, e.g-**

screen shot 2018-07-15 at 16 08 52

mdo:
screen shot 2018-07-29 at 3 45 41 pm

Fixes #26883.

### DIFF
--- a/site/_data/docs-versions.yml
+++ b/site/_data/docs-versions.yml
@@ -1,0 +1,41 @@
+- group: v1.x
+  baseurl: https://getbootstrap.com
+  description: Every minor and patch release from v1 is listed below.
+  versions:
+    - v: 1.0.0
+    - v: 1.1.0
+    - v: 1.1.1
+    - v: 1.2.0
+    - v: 1.3.0
+    - v: 1.4.0
+
+- group: v2.x
+  baseurl: https://getbootstrap.com
+  description: Every minor and patch release from v2 is listed below.
+  versions:
+    - v: 2.0.0
+    - v: 2.0.1
+    - v: 2.0.2
+    - v: 2.0.3
+    - v: 2.0.4
+    - v: 2.1.0
+    - v: 2.1.1
+    - v: 2.2.0
+    - v: 2.2.1
+    - v: 2.2.2
+    - v: 2.3.0
+    - v: 2.3.1
+    - v: 2.3.2
+
+- group: v3.x
+  baseurl: https://getbootstrap.com/docs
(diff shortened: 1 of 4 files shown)


sts-ryan-holton:
Looks Good!

XhmikosR:
Tests seem to fail.

XhmikosR (approved):

XhmikosR:
@mdo: is this ready from your side?

mdo:
@XhmikosR Thanks for the squash and the ping! <3

XhmikosR:
@mdo: shouldn't the link [here](https://github.com/twbs/bootstrap/pull/26989/files#diff-1bad4ad0daa5f9a70db514ede5361f85R42) be `{{ site.baseurl }}/docs/{{ site.docs_version }}/versions/`? Or you plan to move this to the root? I just have a feeling this isn't right.

mdo:
@XhmikosR If can be that. I was copy-pasting stuff across multiple versions and was being lazy with the hard coding.

XhmikosR:
All right, I thought so too. I added a patch in #27408

PR #26996, opened 2018-07-30
Simpler carousel indicators css

MartijnCuppens:
Pseudo elements are used to increase the hit area by 10px on top and bottom, but this can also be done with transparent borders which saves a bit css.

Demo: https://codepen.io/MartijnCuppens/pen/ajNXZj


XhmikosR:
I definitely like this, but better wait for one more approval.

XhmikosR:
Ping @andresgalante @mdo

scss/_carousel.scss
```diff
@@ -182,37 +182,20 @@
...
-    height: $carousel-indicator-height;
+    height: ($carousel-indicator-height + 20px);
     margin-right: $carousel-indicator-spacer;
     margin-left: $carousel-indicator-spacer;
     text-indent: -999px;
     cursor: pointer;
     background-color: $carousel-indicator-active-bg;
+    background-clip: padding-box;
+    // Use transparent borders to increase the hit area by 10px on top and bottom.
+    border-top: 10px solid transparent;
```
mdo: Can we move to a variable for the `10px` value, and then use that here, on L196 and L187?

MartijnCuppens: Changed it to a variable and used `box-sizing: content-box` instead of adding the hit area height to indicator height (this makes it possible to define the hit area height in another unit).

XhmikosR (approved):

PR #27005, opened 2018-07-31
Tweak modal transform

### LINKED ISSUE #27004: Tweak modal transform
The modal with scrolling long content flies in quite fast:
https://getbootstrap.com/docs/4.1/components/modal/#scrolling-long-content

Imo we should use a fixed transform value instead of the relative value of 25%.

MartijnCuppens:
Make the modal transform variable and tweak the default value.

Closes #27004


andresgalante (changes requested):
Hi @MartijnCuppens I think there is a need for 
```scss
$modal-show-transform:              translate(0px, 0) !default;
```
Since if the author wants to change any other transform that's not `translate` they'd also need an ending point.

MartijnCuppens:
Hi @andresgalante,

`transform: translate(0px, 0)` would be the same as `transform: none`, so I changed it like that. I've also included some documentation.

What do you think about this?

andresgalante (approved):
👍 yeap, thanks @MartijnCuppens

PR #27010, opened 2018-08-01
Remove nowrap from .btn

### LINKED ISSUE #26954: 4.1 Accordion Responsive Bug
https://getbootstrap.com/docs/4.1/components/collapse/

The example accordion is not responsive. When the accordion title is longer than the viewport it doesn't break. Instead it overflows and increases the page width.

You can see the bug here:
http://cloud.codecowboy.com/1Q3X2w2Y2S40

**Suggested Fix**
`
.accordion h5 .btn {white-space: normal;}
`

- Windows 10
- Chrome Version 67.0.3396.99 (Official Build) (64-bit)

rowantran:
Remove the white-space property from .btn, which prevents accordion titles from overflowing when they are wider than the viewport.

Closes #26954


rowantran:
After removing nowrap:
accordion-wrap

andresgalante (approved):
Thanks @rowantran

PR #27021, opened 2018-08-05
Move hand cursor for buttons to reboot

### LINKED ISSUE #26998: Cursor: pointer
`cursor: pointer;` is used for `.btn`, `.close` and some other buttons, but not for `button` and `input[type="button"]` in `reboot.scss`. Shouldn't we also add it there?

Something related: when `cursor: pointer;` is used, the following comment is shown:
```
// Opinionated: add "hand" cursor to non-disabled .navbar-toggler elements
```
I was thinking about making a `$enable-` variable and a mixin instead of this comment, because now people who don't want the hand still have got to override every case.

What's people's opinion on this?

MartijnCuppens:
Fixes #26998, fixes #22693

- This PR adds a hand cursor for all buttons
- Adds `$enable-hand-cursor-for-buttons` variable which allows developers to still use the default cursor for all buttons.
- Adds a reset button example to the [reboot page](http://getbootstrap.com/docs/4.1/content/reboot/#forms)
- Moved all the separated declarations to `scss/_reboot.scss`.

Closes #25598

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -341,6 +341,18 @@ button,
   -webkit-appearance: button; // 2
 }
 
+// Opinionated: add "hand" cursor to non-disabled button elements.
+@if $enable-pointer-cursor-for-buttons {
+  button,
+  [type="button"],
+  [type="reset"],
+  [type="submit"] {
+    &:not(:disabled) {
+      cursor: pointer;
+    }
+  }
+}
+
 // Remove inner border and padding from Firefox, but don't restore the outline like Normalize.
 button::-moz-focus-inner,
 [type="button"]::-moz-focus-inner,
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -114,6 +114,7 @@ $enable-transitions:                          true !default;
 $enable-prefers-reduced-motion-media-query:   true !default;
 $enable-hover-media-query:                    false !default; // Deprecated, no longer affects any compiled CSS
 $enable-grid-classes:                         true !default;
+$enable-pointer-cursor-for-buttons:           true !default;
 $enable-print-styles:                         true !default;
 $enable-validation-icons:                     true !default;
 
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -34,11 +34,6 @@
     @include box-shadow(none);
   }
 
-  // Opinionated: add "hand" cursor to non-disabled .btn elements
-  &:not(:disabled):not(.disabled) {
-    cursor: pointer;
-  }
-
   &:not(:disabled):not(.disabled):active,
   &:not(:disabled):not(.disabled).active {
     @include box-shadow($btn-active-box-shadow);
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -17,9 +17,6 @@
     @include hover-focus {
       opacity: .75;
     }
-
-    // Opinionated: add "hand" cursor to non-disabled .close elements
-    cursor: pointer;
   }
 }
 
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -116,11 +116,6 @@
   @include hover-focus {
     text-decoration: none;
   }
-
-  // Opinionated: add "hand" cursor to non-disabled .navbar-toggler elements
-  &:not(:disabled):not(.disabled) {
-    cursor: pointer;
-  }
 }
 
 // Keep as a separate element so folks can easily override it with another icon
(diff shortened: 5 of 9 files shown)


patrickhlauke (approved):
This looks tidy to me (though I bet people still won't be happy, or start asking why we're not doing it for generic `.btn` applied to arbitrary elements as well...)

ysds:
similar PR and discussion in #25598

andresgalante (approved):
@MartijnCuppens Thanks a lot, this is a good move IMO. 
I'd change `enable-hand-cursor-for-buttons` to `enable-pointer-cursor-for-buttons`, what do you think?

scss/_reboot.scss
```diff
@@ -360,6 +360,18 @@ html [type="button"], // 1
   -webkit-appearance: button; // 2
 }
 
+// Opinionated: add "hand" cursor to non-disabled button elements.
+@if $enable-hand-cursor-for-buttons {
```
patrickhlauke: as per @andresgalante's suggestion, maybe rename to `enable-pointer-cursor-for-buttons` - the preceding comment should make it clear what it means in practice

scss/_variables.scss
```diff
@@ -109,15 +109,16 @@ $yiq-text-light:            $white !default;
...
-$enable-hover-media-query:  false !default; // Deprecated, no longer affects any compiled CSS
-$enable-grid-classes:       true !default;
-$enable-print-styles:       true !default;
-$enable-validation-icons:   true !default;
+$enable-caret:                   true !default;
+$enable-rounded:                 true !default;
+$enable-shadows:                 false !default;
+$enable-gradients:               false !default;
+$enable-transitions:             true !default;
+$enable-hand-cursor-for-buttons: true !default;
```
patrickhlauke: see above

site/docs/4.1/getting-started/theming.md
```diff
@@ -230,18 +230,19 @@ Customize Bootstrap 4 with our built-in custom variables file and easily toggle
...
+| --------------------------------- | ---------------------------------- | -------------------------------------------------------------------------------------- |
+| `$spacer`                         | `1rem` (default), or any value > 0 | Specifies the default spacer value to programmatically generate our [spacer utilities]({{ site.baseurl }}/docs/{{ site.docs_version }}/utilities/spacing/). |
+| `$enable-rounded`                 | `true` (default) or `false`        | Enables predefined `border-radius` styles on various components. |
+| `$enable-shadows`                 | `true` or `false` (default)        | Enables predefined `box-shadow` styles on various components. |
+| `$enable-gradients`               | `true` or `false` (default)        | Enables predefined gradients via `background-image` styles on various components. |
+| `$enable-transitions`             | `true` (default) or `false`        | Enables predefined `transition`s on various components. |
+| `$enable-hand-cursor-for-buttons` | `true` (default) or `false`        | Add "hand" cursor to non-disabled button elements. |
```
patrickhlauke: see above

MartijnCuppens:
No problem, just changed it to `$enable-pointer-cursor-for-buttons`.

patrickhlauke:
:+1: ... @mdo what say you?

scss/_reboot.scss
```diff
@@ -360,6 +360,18 @@ html [type="button"], // 1
   -webkit-appearance: button; // 2
 }
 
+// Opinionated: add "hand" cursor to non-disabled button elements.
+@if $enable-pointer-cursor-for-buttons {
+  button,
+  [type="button"],
+  [type="reset"],
+  [type="submit"] {
+    &:not(:disabled):not(.disabled) {
```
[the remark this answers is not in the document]
MartijnCuppens: That's correct, I removed it. Thanks @ysds!

XhmikosR:
@mdo: ping for review

mdo (commented):
Current selectors include `:not(.disabled)` as well as the `disabled` attribute. I don't think we can consolidate all these under a single attribute selector, nor can we move away from the component classes completely I'd think since many, if not most, of these apply to anchors as well.

jackmu95:
Isn't this going to be available in v4.2?

MartijnCuppens:
> Current selectors include :not(.disabled) as well as the disabled attribute. I don't think we can consolidate all these under a single attribute selector, nor can we move away from the component classes completely I'd think since many, if not most, of these apply to anchors as well.

@mdo , this change won't affect the links, they will still have `cursor: pointer` unless a class `.disabled` is added to `a.btn`. The only thing this will change is that the `button` (and input buttons) will have `cursor:pointer` by default.

andresgalante (approved):
I think we should move this forward

mdo (approved):

PR #27027, opened 2018-08-06
users/adamsiv/rtl-support

DigitalKrony:
On behalf of Microsoft, I'm submitting this pull-request.

We have taken Bootstrap 4.x on as a dependency for a project, however, we have requirements for our websites to work in an international arena as well as maintain perf goals and an easy dev story. To that end, the best case in order to align these things is to integrate internationalization directly into Bootstrap. You'll find the appropriate update within. 

The chages in the files below simply add in the `dir` attribute to the HTML tag to allow ease of testing. While testing the RTL story wihtin the doc site, these must be set to `"rtl"`.
- site > _layouts >
    - default.html
    - docs.html
    - home.html
    - redirect.html

The update to the SCSS is realativly simple. I've added three variables to the `_variables.scss`. 
- $dir:String = ltr (default) || rtl
- $left:String = left (default) || right (is set based on $dir)
- $right:String = right (default) || left (is set based on $dir)

These three variables work in conjunction with eachother. By `!default`  `$dir` is set to `ltr`. Following this varaible, there's `$left` and `$right`. These two variables are set via a function (`set-dir(left|right)` ) base on the value of `$dir` for ease of use. The `$dir` variable also sets a global `direction: $dir` style decloration to capture finer style requirements for things like inputs and sudo elements.


wolfy1339:
This is a duplicate, there is already a PR that deals with LTR and RTL functionality

wolfy1339:
Please don't commit the dist files as per the Contribution guidelines of this repository

XhmikosR:
/CC @mdo: we really need to look into RTL support...

XhmikosR:
@DigitalKrony: there are still dist files here (js/dist)

XhmikosR:
There are still more dist files like the docs CSS and https://github.com/twbs/bootstrap/pull/27027/files#diff-2f2d399f0ea8844859fe5514b304733b

js/src/util.js
```diff
@@ -67,6 +67,13 @@ const Util = (($) => {
 
     TRANSITION_END: 'bsTransitionEnd',
 
+    getPageDirection() {
+      const htmlElement = document.getElementsByTagName('html')[0]
```
XhmikosR: any reason why you don't use `document.documentElement` here?

DigitalKrony: I was aiming for coverage and the dev docs around documentElement seemed unsure of IE and Edge support. If there's info surrounding coverage, I'd rather use it over the current implementation and would update accordingly.

XhmikosR: You work for Microsoft, you should be able to learn about it from the inside :P

XhmikosR: But joking aside, works fine for me on latest Edge that ships with 10 stable and IE 11 and IE 10 (emulated).

DigitalKrony: Those worked for me as well. And Emulated doesn't always do the trick. #trustMe.

However, I ran it through some rough and quick tests on Browser Stack, seemed to work as expected. At least on Win10. If someone else wants to roll back to win7 and Vista and test it, please send me your findings. Making the update now.

andresgalante:
Thanks a lot @DigitalKrony, as @wolfy1339 mentions there are other rtl PRs, including this one open: 
https://github.com/twbs/bootstrap/pull/26722

@mdo, I agree with @XhmikosR we do need to address rtl. If we are following this solution I'd rather use `#{start}` and `#{end}` instead of `#{left}` and `#{right}` following the standard for [CSS logical propierties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties).

DigitalKrony:
Thanks, @andresgalante, for dropping in. 

Were we consolidating into one PR or am I updating this one to align to guidance? Happy to do either. Just hoping to get this landed sooner rather than later.

andresgalante:
@DigitalKrony it'll be up to @mdo to decide if this goes in and when. The thing to consider is that if this is creating a breaking change, it'll have to land in v5.

DigitalKrony:
So @mdo and @XhmikosR, what do we need to do to get this in? I would really appreciate a look-see (other than resolving the conflicts that have worked their way in).

XhmikosR:
Unfortunately, it's not my decision to make. I know for sure this should have been addressed a long time ago for sure. For reference https://github.com/twbs/bootstrap/pulls?utf8=%E2%9C%93&q=is%3Apr+rtl

MartijnCuppens:
RTL support would indeed be nice! I'm not familiar with making rtl websites, but was curious how this was done, so I have set up this PR locally to check some things.

I think in some cases the changes don't really change anything. Like here:
https://github.com/twbs/bootstrap/blob/c4245556b2a123fa0a504f2ab39e9b2c3c8130ac/scss/_badge.scss#L34-L35

or here:
https://github.com/twbs/bootstrap/blob/c4245556b2a123fa0a504f2ab39e9b2c3c8130ac/scss/_card.scss#L108-L109
and some other places.

About the carousel, I don't know if this is the behaviour we want:
image

And some dropdowns are broken:
image

@DigitalKrony, I assume you did a _find-and-replace_ of the `right` and `left` appearances, combined with some extra tweaks but I think every component will need to be reviewed if we want a solid solution for rtl support.

DigitalKrony:
@MartijnCuppens _ I actually didn't do a find replace, I went file by file and corrected the issues I saw for each component by hand. However, this has been sitting, waiting to merge in for such a long time, that errors have started to work their way in. The screenshots that you've posted above, I remember having to massage out of the code when I first did my integration. 

I'll re-merge and do the work again.

DigitalKrony:
So I've gone back and reviewed the branch for the comments and inconsistencies mentioned above.

  - For the `_badge.scs` and `_card.scss` edits, I was aiming for consistency with regards to how padding and margin was handled. While, no, it's not really value after the files compile, it is adding value from a consistency standpoint as well as informing other on a proper way to utilize it in other places. It's also not hurting anything having these as variables.
  - Conserning the carousel, the glyphs for `next` and `previous` are varaible sets. However, I would agree that if a user built BS out of the box, it should look as expected. Those chevrons are no exception. So, I've added in a funciton to set a value based on the `$dir` variable. This resolved this issue and allows an author to do likewise in the future.

DigitalKrony:
@MartijnCuppens - Thanks for taking the time. Can you do a pull and let me know if you see anything else. I'd like to make sure we have this right when it's some to get it merged in. I have a lot of folks on my end REALLY looking forward to this being merged and release.

MartijnCuppens:
@DigitalKrony, I'm glad you're so driven to help Bootstrap with the RTL support but I'm afraid I can only review a limited part of this PR. I'm not familiar with rlt languages myself and I think we'll need reviews from several people who are used to make rtl websites before a PR can be merged.

powerchelle:
@mdo hey there, any idea when this might be integrated?

superware:
> @DigitalKrony, I'm glad you're so driven to help Bootstrap with the RTL support but I'm afraid I can only review a limited part of this PR. I'm not familiar with rlt languages myself and I think we'll need reviews from several people who are used to make rtl websites before a PR can be merged.

RTL is easy, in principal think of it as a mirror for LTR content, anything going from left-to-right should go from right-to-left (so right is "preceding" and left is "following"), anything aligned to the left should be aligned to the right, any left-arrow (indicating "previous") should be a right-arrow, etc. "Mirrored" is the key, it's pretty straight forward. See [bootstrap.rtlcss.com](http://bootstrap.rtlcss.com/) for example.

BoFFire:
Crazy threads about RTL in Bootstrap :))

superware:
> Crazy threads about RTL in Bootstrap :))

There is no technical reason for Bootstrap's lack of RTL support, with around 540 million Arabic/Persian/Hebrew speakers worldwide - no question regarding the necessity.

I was just pointing out the obvious, that RTL is a visual mirror of LTR (alignment, floating, margin, padding wise etc). This has nothing to do with the languages themselves and can be fully supported and tested with English.

Rux77:
[Zurb Foundation for Sites 6](https://foundation.zurb.com/sites.html) has a very long experience in RTL support. And they enable RTL by one `$global-text-direction` variable and then let SASS decide the right direction for `$right` and `$left` based on `$global-text-direction` variable .

```scss
$global-text-direction: ltr !default;
$global-left: if($global-text-direction == rtl, right, left);
$global-right: if($global-text-direction == rtl, left, right);
```

Bootstrap team is waiting for the perfect solution to support RTL, which will not come in the near future or maybe ever?.

powerchelle:
@mdo hi there, we chatted a few months ago about seeing if this could be included in the next minor release. Could you please provide an update?

powerchelle:
@mdo will this be merged in the future, or should we consider this dead?

mdo:
We're looking at tackling RTL in v5! See #30980 for details. Sorry we didn't get to do this for v4, maybe we'll still be able to bridge these major versions once we tackle this in v5. Please weigh in on the new PR and referenced issues with any feedback moving forward! <3

PR #27034, opened 2018-08-07
Change if statement negative margin utilities

MartijnCuppens:
I was testing the negative margin utilities but noticed they weren't generated. Found the cause, `@if not $size == 0` doesn't work.


MartijnCuppens:
Hmm, now that the classes are generated, the filesizes are exceeded.

```
 FAIL  ./dist/css/bootstrap-grid.css: 6.75KB > maxSize 6KB gzip 
 FAIL  ./dist/css/bootstrap.min.css: 21.63KB > maxSize 21KB gzip 
```

XhmikosR (approved):

PR #27035, opened 2018-08-07
Side modals

### LINKED ISSUE #24718: Drawer, sidebar, offcanvas nav

### LINKED ISSUE #27033: Feature request: side Modals

ezgif-2-21a3fc9652

PR incomming...

MartijnCuppens:
Closes #27033, closes #24718

Implementation "drawer modals". 

I used this hack to make sure the `.modal-footer` sticks to the bottom in IE. I wrapped this in a media query to prevent strange behaviour on mobile browsers (see screenshot here https://css-tricks.com/the-trick-to-viewport-units-on-mobile/#post-274409)
```css
// IE fix:
@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
  min-height: 100vh;
}
```

Side modals can be positioned on the left and right side of the screen and are compatible with the sizing classes.

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -63,6 +63,58 @@
   }
 }
 
+.modal-dialog-left-side {
+  position: absolute;
+  top: 0;
+  left: 0;
+  display: flex;
+  width: 100%;
+  min-height: 100%;
+  // stylelint-disable-next-line declaration-no-important
+  margin: 0 !important;
+
+  .modal-content {
+    // IE fix:
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+      min-height: 100vh;
+    }
+    border-width: 0;
+    border-radius: 0;
+  }
+
+  .modal.fade & {
+    transform: translate(-$modal-side-animation-offset, 0);
+  }
+}
+
+.modal-dialog-right-side {
+  position: absolute;
+  top: 0;
+  right: 0;
+  display: flex;
+  width: 100%;
+  min-height: 100%;
+  // stylelint-disable-next-line declaration-no-important
+  margin: 0 !important;
+
+  .modal-content {
+    // IE fix:
+    @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+      min-height: 100vh;
+    }
+    border-width: 0;
+    border-radius: 0;
+  }
+
+  .modal.fade & {
+    transform: translate($modal-side-animation-offset, 0);
+  }
+}
+
+.modal.show .modal-dialog {
+  transform: translate(0, 0);
+}
+
 // Actual modal
 .modal-content {
   position: relative;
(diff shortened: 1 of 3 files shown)


andresgalante:
Hi @MartijnCuppens should this be a new component instead of a reposition of the modal? Maybe something like a drawer?

MartijnCuppens:
Thought about that, but I think there's a lot of functionality that is the same like:
- The header (with title & close button), body & footer
- The backdrop
- The whole js-functionality (keyboard navigation, events, prevent page scrolling,...)

Apart from that, I've made these modals compatible with the sizing classes of the modals.

But maybe I can rename them to _Drawer modals_ in stead of _Side modals_, this is maybe a more frequently used term? Or do you have another suggestion, @andresgalante?

Johann-S:
I'm more in favor of a new composant with some refactoring

MartijnCuppens:
Hmm, I'll need some help with the refactoring (especially the js part) if we 're going for a new component.

wolfy1339:
Wasn't this already done before in another PR?

Johann-S:
@MartijnCuppens I would love to work with you on a separate plugin 😉

XhmikosR:
Rebased, hopefully nothing broke :)

LGTM BTW.

Johann-S:
As it would said before your comment @XhmikosR, we'll work on a separate plugin, we won't tweak our modal to do that

XhmikosR:
If you can refactor the code to be shared between both components, sure. Otherwise there will be a lot of duplication and this solution seems pretty good as a first step.

Johann-S:
I prefer to have two separate files, because if we have to evolve the drawer (and it'll happen) we will be scared to break things in the Modal, it's important to separate component which do not achieve the same things

XhmikosR:
Well, sure but there's a balance. There's no point in reinventing the wheel and increasing our JS code just yet when this works just fine.

Johann-S:
That's not because it works fine, we should accept that. That's not the responsability of the Modal component to show/hide a Drawer, or if it's the case we should use our Modal component to show/hide everything...

XhmikosR:
True that's a valid point. But I care about sharing the common code so that we don't bloat the JS code needlessly

Johann-S:
But I agree we should work on refactoring things if we add this component, I want to work on that with @MartijnCuppens

MartijnCuppens:
Fine by me 👍

viktorsmari:
Where is work happening on this now? I cannot find a reference to it.

madflow:
Maybe some inspiration from https://material-ui.com/demos/drawers/ can help? I really like the functionality persistent/permanent and the "mini" variant. Keep up the good work :) !

viktorsmari:
Is anyone working on this?
This is what I think is most needed in Bootstrap today.
Can I contribute or give a bounty directly to this issue via opencollective or similar services?

Issue #26950, opened 2018-07-24
Carousel transition between first -> last and last -> first has a visual abnormality (mobile Google Chrome)


Ruffio:
Using latest Google Chrome on Mobile (Samsung S9), if you go to the Carousel example page (http://getbootstrap.com/docs/4.1/examples/carousel/) and use the right and left buttons between first and last item, then during transition some of the text/area is missing fromt the first item.

During transition the left part of the first item is hidden/covered with background color and then when the transition ends it come back.
This is the area that disappears:
![image](https://user-images.githubusercontent.com/1877612/43125760-35a90be4-8f2b-11e8-953d-a07097923ef1.png)

The issue is only during transition.
I have looked for if this issue has already been submitted but didn't find any. If there is already an issue submitted regarding this issue then just close this. Thanks.

nitishchopra92:
@Ruffio , removing the div with `container` class from the items will resolve this issue. I have create a sample of this at https://jsfiddle.net/25nitish/pxt8a7jw

Will see if I can generate a PR to get this included in main site.

Ruffio:
Cool. I'm not using the carousel myself, just stumbled upon it...

fre. 10. aug. 2018 14.32 skrev Nitish Chopra <notifications@github.com>:

MartijnCuppens:
I could reproduce this i n `v4.1`, but not in `4.2`. @Ruffio, can you check if you still have this with https://getbootstrap.com/docs/4.2/examples/carousel/?

XhmikosR:
Closing due to lack of response. You can comment here and we can re-open the issue.

*This is an automated reply*

PR #27048, opened 2018-08-10
Carousel transition issue fix

nitishchopra92:
This is in relation to #26950 where some HTML changes will fix the transition issue seen in mobile Google Chrome.


mdo:
Without understanding _why_ the container interferes, we shouldn't make this change. Commenting real late on this, but any ideas from when you looked into this?

XhmikosR:
Maybe it was/is a browser bug?

twbs-closer:
Hey there!

Thanks!

PR #27053, opened 2018-08-11
fix(util): use querySelector for ids

Johann-S:
We should not use `getElementById` for selectors with multi ids

/CC @XhmikosR

### DIFF
--- a/js/src/util.js
+++ b/js/src/util.js
@@ -84,7 +84,7 @@ const Util = (($) => {
       }
 
       const validSelector = selector
-      if (selector.charAt(0) === '#') {
+      if (selector.charAt(0) === '#' && selector.indexOf(',') === -1) {
         selector = selector.substr(1)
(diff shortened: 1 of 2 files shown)


XhmikosR (approved):

js/src/util.js
```diff
@@ -84,7 +84,7 @@ const Util = (($) => {
       }
 
       const validSelector = selector
-      if (selector.charAt(0) === '#') {
+      if (selector.charAt(0) === '#' && selector.indexOf(',') === -1) {
```
glebm: How about:

1. Spaces, e.g. `#my-id div`
2. Combinators like `+` and `>`, e.g. `#my-id>div`, `#my-id+div`, etc

Johann-S: You're totally right @glebm I'll fix that asap thanks

PR #27056, opened 2018-08-12
add babel plugin proposal object rest spread to our dev deps

Johann-S:
/CC @XhmikosR

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
   "devDependencies": {
     "@babel/cli": "7.0.0-rc.1",
     "@babel/core": "7.0.0-rc.1",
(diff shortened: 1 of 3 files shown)


XhmikosR (approved):

package.json
```diff
@@ -99,6 +99,7 @@
   "devDependencies": {
     "@babel/cli": "7.0.0-rc.1",
     "@babel/core": "7.0.0-rc.1",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-rc.1",
```
XhmikosR: Hmm, maybe we should not have used `^` here like the rest of the non-stable deps.

PR #27062, opened 2018-08-13
update alert documentation without jquery

Johann-S:
Rewrite our alert page documentation without the use of jQuery

### DIFF
--- a/site/docs/4.1/components/alerts.md
+++ b/site/docs/4.1/components/alerts.md
@@ -76,7 +76,10 @@ You can see this in action with a live demo:
 Enable dismissal of an alert via JavaScript:
 
 {% highlight js %}
-$('.alert').alert()
+var alertList = document.querySelectorAll('.alert')
+alertList.forEach(function (alert) {
+  new bootstrap.Alert(alert)
+})
 {% endhighlight %}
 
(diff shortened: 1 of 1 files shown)


site/docs/4.1/components/alerts.md
```diff
@@ -76,7 +76,10 @@ You can see this in action with a live demo:
 Enable dismissal of an alert via JavaScript:
 
 {% highlight js %}
-$('.alert').alert()
+var alertList = document.querySelectorAll('.alert')
+alertList.forEach(function (alert) {
```
vsn4ik: Maybe
```js
var alertList = document.querySelectorAll('.alert')
Array.from(alertList).forEach(...);
```
?

Johann-S: `Array.from` isn't widely supported and we can use `forEach` on NodeList so I won't change this line but thank you for your feedback @vsn4ik it's always appreciated 😉

XhmikosR (approved):

Johann-S:
@XhmikosR test won't pas unfortunately 😢 because for an unknown reason they fail on Sauce Labs but they works very well on my computer, do you have the time to check that on your computer ?

PR #27064, opened 2018-08-13
Custom checkboxes and radios retheming

### LINKED ISSUE #27063: Custom checkboxes and radios retheming
The custom checkboxes and radios are cool, but I think they might need some retheming. I don't think it's a good idea to make them gray, it looks like they are disabled and there isn't much of a difference between the disabled and not disabled versions.

Imo they should have the same background and border color as the `.form-control` elements and also take over the theming of disabled state.

I made a demo of the rethemed custom checkboxes here: https://codepen.io/MartijnCuppens/pen/KBYKKz

MartijnCuppens:
Closes #27063

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -26,18 +26,28 @@
 
   &:checked ~ .custom-control-label::before {
     color: $custom-control-indicator-checked-color;
+    border-color: $custom-control-indicator-checked-border-color;
     @include gradient-bg($custom-control-indicator-checked-bg);
     @include box-shadow($custom-control-indicator-checked-box-shadow);
   }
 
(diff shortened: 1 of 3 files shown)


XhmikosR:
LGTM but I'd even go with even more darker border.

Either way, the final review is for @mdo @andresgalante to make (for now :P)

MartijnCuppens:
Made a third version with a darker border (`$gray-600`) https://codepen.io/MartijnCuppens/pen/PywvVJ. I think this is a bit too dark and the border color differs too much from `.form-control`

XhmikosR:
Yup, agreed, this one is a little too much. But let's move this on the PR itself. :)

mdo (approved):
Dig the new look, thanks!

PR #27066, opened 2018-08-14
More flexible blockquote footer font size

acjbizar:
I think it would make sense to move the font size of `blockquote-footer` to the variables. See commit message for more details.

The second commit is a tiny cleanup.


andresgalante (approved):
Thanks a lot @acjbizar

PR #27082, opened 2018-08-16
Add img-cover and img-contain helpers

kevinruscoe:
Add `.img-cover` and `.img-contain` helpers.


mdo:
Docs and examples?

MartijnCuppens:
I'm afraid we can't use this for now, `object-fit` is not supported in IE (https://caniuse.com/#feat=object-fit)

mdo:
Closing per last comment. Thanks though!

PR #27094, opened 2018-08-19
Fix size of modal dialogs at different widths

namiltd:


namiltd:
Correctly displays .modal-xl for @media min-width: between 992px and 1199px
and .modal-sm for @media min-width: less than 576px

scss/_modal.scss
```diff
@@ -150,6 +150,8 @@
   overflow: scroll;
 }
 
+.modal-sm { max-width: $modal-sm; }
```
andresgalante: Why are you adding this line? it brakes the small modal on xs sizes.

andresgalante (changes requested):
@namiltd thanks, I think this makes sense, other than the sm modal is broken, but just by removing line 153 works.

@mdo this PR makes the xl modal goes from xl, to lg, to sm, instead of just jumping from xl to sm. I think it makes sense.

scss/_modal.scss
namiltd: My tests
modal-sm on a wide screen:
normal

modal-sm on a narrow screen, original:
original

modal-sm on a narrow screen, patched:
patched

andresgalante: Yes, I'd say that the result is accurate, we want it full screen on small viewports.

I'd remove that line.

namiltd: ok

namiltd: Ok, line removed.

namiltd (commented):
Fix for .modal-sm removed.

andresgalante (approved):

andresgalante:
Thanks @namiltd!

XhmikosR:
@MartijnCuppens: LGTY?

scss/_modal.scss
```diff
@@ -175,7 +175,8 @@
 }
 
 @include media-breakpoint-up(lg) {
-  .modal-lg { max-width: $modal-lg; }
+  .modal-lg,
+  .modal-xl { max-width: $modal-lg; }
```
MartijnCuppens: @XhmikosR, I would add some line breaks here:
```suggestion
  .modal-xl { 
    max-width: $modal-lg;
  }
```

Apart from that, this PR looks fine.

MartijnCuppens (changes requested):

scss/_modal.scss
XhmikosR: Agreed. I personally never use one line expressions but here on Bootstrap I didn't enforce it. We might revisit later :)

PR #27115, opened 2018-08-21
Add breadbrumb in docs search results

### LINKED ISSUE #23389: Simplifying the search results
I know that an external tool is used for the search and I don't know if this is possible or not, but it would be neat for the results to have a breadcrumb-like look (for the headers only) because I was looking to the `description list` markup and I got the following result

capture

In the Content area, the first result is in the `Reboot` page, and the second in the `Typography` page. Instead of having 

- Lists

- Lists

having something like 

- Reboot \ Lists

- Typography \ Lists

would be a lot more straight-forward.

MartijnCuppens:
Make the subcategory visable and theme it like a breadcrumb. Made some screenshots which makes it easier to compare.

### List
See original issue (#23389), it was not clear from where the list pages came from.

**List before:**

**List after:**

### Left
This is just an example how it is displayed if the highlighted text is in the text and not in the title

**Left before:**

**Left after:**

### Alert
Without breadcrumb it was not clear why some results showed up. 

**Alert before:**

**Alert after:**

Closes #23389

### DIFF
--- a/site/docs/4.1/assets/scss/_algolia.scss
+++ b/site/docs/4.1/assets/scss/_algolia.scss
@@ -66,10 +66,23 @@
     text-align: left !important;
   }
 
+  .algolia-docsearch-suggestion--subcategory-inline {
+    display: block !important;
+    font-size: .875rem;
+    color: $gray-700;
+
+    &::after {
+      padding: 0 .25rem;
+      content: "/";
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):
Love it, thanks!

XhmikosR:
@mdo: please have a quick look before we merge.

mdo (approved):

PR #27118, opened 2018-08-22
Improve spacing class inheritance

svetoslavatanasov:
More specific spacing classes should be applied instead of those with broader range.

E.g. if you have:

```
<div class="m-md-1 mb-0"></div>
```

the bottom margin class will be overwrited by the all margins one in md viewport. The changed code fixes this.

### DIFF
--- a/scss/utilities/_spacing.scss
+++ b/scss/utilities/_spacing.scss
@@ -9,65 +9,106 @@
     @each $prop, $abbrev in (margin: m, padding: p) {
       @each $size, $length in $spacers {
         .#{$abbrev}#{$infix}-#{$size} { #{$prop}: $length !important; }
-        .#{$abbrev}t#{$infix}-#{$size},
+      }
+    }
+    @each $size, $length in $spacers {
+      @if $size != 0 {
+        .m#{$infix}-n#{$size} { margin: -$length !important; }
+      }
+    }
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
This is quite a breaking change and what you're trying to achieve is also possible with the use of other utility classes:

```html
<div class="mx-md-1 mt-md-1"></div>
```

mdo:
> the bottom margin class will be overwrited by the all margins one in md viewport. The changed code fixes this.

This is intended behavior.

svetoslavatanasov:
@MartijnCuppens

Sorry but my example was not complete. Using the current class order, the bottom margin this will be applied only to extra small devices:

```
<div class="m-1 m-sm-2 m-md-3 m-lg-4 m-xl-5 mb-0"></div>
```

In order to have bottom margin set to 0 for all viewports you will need:

```
<div class="m-1 m-sm-2 m-md-3 m-lg-4 m-xl-5 mb-0 mb-sm-0 mb-md-0 mb-lg-0 mb-xl-0"></div>
```

Even if you want to set a 0 bottom margin only to the smallest viewport, you will need less classes:

```
<div class="m-1 m-sm-2 m-md-3 m-lg-4 m-xl-5 mb-0 mb-sm-2"></div>
```

@mdo 

Wasn't it intended to have less utility classes to achieve a styling goal?

PR #27120, opened 2018-08-23
Change .figure display

svetoslavatanasov:
Setting the display of the figure to 'table' will allow auto margin to be applied:

```
<figure class="figure mx-auto p-3 bg-light">
...
</figure>
```

image


mdo:
Leaning towards no on this change—you can enable this with `.d-table` instead of changing this for everyone else and breaking the intended inline-block behavior of images we're matching here.

PR #27126, opened 2018-08-24
Fix double border on list-group

### LINKED ISSUE #27125: Double border on .list-group-item when followed by .card-footer
When using a `.list-group .list-group-flush` in a card followed by `.card-footer`, the list last list group item does have a border, so basically this is not being applied: https://github.com/twbs/bootstrap/blob/da717b03e6e72d7a61c007acb9223b9626ae5ee5/scss/_list-group.scss#L100-L104

It is however correctly applied when followed by a `.card-body`, as supposed:

image

CodePen to reproduce:
https://codepen.io/anon/pen/PdPxgG?editors=1000

royklutman:
This fixes #27125.

See comment for CodePen: https://github.com/twbs/bootstrap/issues/27125#issuecomment-415784076


MartijnCuppens:
This breaks the list group if it's the last child of the card, see https://codepen.io/MartijnCuppens/pen/mGVydV?editors=1100

MartijnCuppens:
I think you'll need to scope more to the cards, now you're also targeting the stand alone `.list-group` elements.

This solves it in every case: https://codepen.io/MartijnCuppens/pen/zJrxmO?editors=1100

royklutman:
Isn't is better to just target `.list-group-flush` then? I was a little early posting my comment, I just updated the CodePen. For the future, `.list-group-flush` might be used in other elements then just cards too.

MartijnCuppens:
Yup, I've just tested your latest changes, looks great!

MartijnCuppens (approved):

XhmikosR (approved):

PR #27133, opened 2018-08-26
Fix for double border on cards in an accordion

### LINKED ISSUE #27108: Double bottom border on collapsed cards
When collapsing cards (e.g. in an accordion), the header has a bottom border that is rendered on top of the bottom border of the card itself. It results in two borders to be rendered, which (especially for the last card with rounded edges) looks wrong:
double_bottom_border
Depending on the used browser and zoom settings there can even be a small gap between the two border lines.

https://github.com/twbs/bootstrap/pull/23839 improved the layout of cards as collapsible elements but didn't change the double bottom border.

I would expect that the border of the header is not shown for a collapsed card.

- Operating system: Windows 10
- Browser: Firefox 61.0.2
- Test case (taken from [bootstrap documentation](http://getbootstrap.com/docs/4.1/components/collapse/#accordion-example)): http://jsbin.com/zomipatuci/edit?html,output

### LINKED ISSUE #27124: No Border Bottom in Accordion First Item
There is no Border Bottom if Accordion has only single Item

https://www.bootply.com/vB8GoWkFz3

royklutman:
This PR would fix #27108 and would fix #27124, discussion about this and an example is found in the comment section: https://github.com/twbs/bootstrap/issues/27108#issuecomment-416029816


XhmikosR:
@MartijnCuppens: LGTY?

mdo (approved):

PR #27134, opened 2018-08-26
Fix custom file input focus border color

### LINKED ISSUE #27112: [Custom file input]Different focused border color between 'is-valid/invalid' and 'was-validated'
image

Demo: https://codepen.io/anon/pen/bxGmdd

The former should be red(green) border?

MartijnCuppens:
Fixes #27112.

Also did a tiny optimalization for the border of the browse button. The left border from the `::after` pseudo element is now inherited from the element so you don't have to define the border color for every state and the transition of the border color is also applied to the browse button. 

Demo: https://codepen.io/MartijnCuppens/pen/bxEyrQ

image


ysds:
LGTM 👍

PR #27137, opened 2018-08-26
add information about valid selectors

Johann-S:
Finally I revert this code, because there are a lot of use case to detect a valid uniq `id`, so it's better to add information about that in documentation


glebm (approved):
Thanks!

XhmikosR:
I'm personally thinking we might need to revert this change. I mean, technically we are right, but it is a breaking change...

That being said, I'd change this one to a warning or something to make it stand out maybe.

Johann-S:
That's not a breaking change @XhmikosR because that's what we shipped in our last release see: https://github.com/twbs/bootstrap/blob/v4.1.3/js/src/util.js#L78-L89 That's exactly the same code, we just add some information about valid selectors

XhmikosR:
Yes, I meant when this was added, not the revert.

Johann-S:
Oh ok 😄 
Why not make that into a `warning` but I thought it was to huge to be a warning 🤔

XhmikosR:
@Johann-S: is it hard to revert the related changes? If it's too hard let's leave it as is and make this a warning. But ideally we shouldn't introduce breaking changes.

XhmikosR:
I don't know at this point. When was this introduced? 4.1.3 or older?

Johann-S:
It was on 4.1.2

PR #27144, opened 2018-08-28
Remove `-webkit-text-decoration-skip`

### LINKED ISSUE #27143: Remove -webkit-text-decoration-skip?
About this line:
https://github.com/twbs/bootstrap/blob/5266669d34e80785a6a053031283cecf2516b7fc/scss/_reboot.scss#L202

This line renders links like this:  
 (&larr; This is an image)

... instead of this:
(&larr; This is an image, the difference is in the underlining of the _g_)

... in safari.

But:
- Other elements in Safari still use `-webkit-text-decoration-skip: ink` if they are underlined
- In chrome, `text-decoration-skip: ink` is used by default

IMO we should remove this line and let the browser/OS decide how the underline is rendered. What do you think?

MartijnCuppens:
Fixes #27143

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -199,7 +199,6 @@ a {
   color: $link-color;
   text-decoration: $link-decoration;
   background-color: transparent; // Remove the gray background on active links in IE 10.
-  -webkit-text-decoration-skip: objects; // Remove gaps in links underline in iOS 8+ and Safari 8+.
 
   @include hover {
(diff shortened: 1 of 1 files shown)


mdo (approved):

mdo:
Thanks for this.

Separately, if it makes your work easier at all, in the future you can skip the issue and have the discussion here in these smaller PRs :). Your call though!

PR #27147, opened 2018-08-29
Add browser-specific appearance styles to .custom-select

Flinsch:
scss/_custom-forms.scss is lacking the browser-specific versions "webkit" and "moz" of the appearance style for .custom-select. When I do an import of "~bootstrap/scss/bootstrap", the <select> menu shows up with the native dropdown button (e.g., using Mozilla Firefox). When I do an import of "~bootstrap/dist/css/bootstrap.min.css" instead, all looks fine (without the native dropdown button).

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -178,6 +178,8 @@
     border-radius: 0;
   }
   @include box-shadow($custom-select-box-shadow);
+  -webkit-appearance: none;
+  -moz-appearance: none;
   appearance: none;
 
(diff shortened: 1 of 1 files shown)


Johann-S:
We use autoprefixer when we build Bootstrap, that's the reason all is fine when you import our minified css

Flinsch:
I was not aware of that, I'm sorry. I have just installed autoprefixer and it works quite nicely with the .scss file now. Many Thanks.

PR #27153, opened 2018-08-30
Update _sizing.scss

fedorovvldmr:
Width and height with media queries


mdo:
We've turned this down a few times over now—no plans to add this to v4. Come v5, I'm hoping we can change things up a bit with the grid to have unified column/width classes, but until then I don't see a benefit in generating all these classes. Thanks though!

PR #27154, opened 2018-08-30
Update _spacing.scss

fedorovvldmr:
Reset margins to gutter size


fedorovvldmr:
updated

mdo:
This is a ton of new classes. Sadly, I don't think we can go this route without adding significant bloat to the project.

PR #27155, opened 2018-08-30
Boom toasted

Lausselloic:
Hello, @Johann-S 
I've improve some part of the toaster

- improve accessibility by adding role="alert" aria-live="assertive" aria-atomic="true" on samples and write some more details onto how make it more accessible (maybe need an english review @patrickhlauke )
- also for accessibility, I think it's really better to let user control the toaster and close it by itself (otherwise AT will vocalise the content but the user can't find it into the page) maybe need a way to inform the user with AT that's an autoclosed component? Haven't find a good idea with aria :-(

- update the hide method to close toaster immediatly on click, implement data-dismiss="toast" like for modal
- add a unit test for that purpose

### DIFF
--- a/site/docs/4.1/components/toasts.md
+++ b/site/docs/4.1/components/toasts.md
@@ -29,6 +29,9 @@ A basic toast can include a header (though it doesn't strictly need one) with wh
     <img class="rounded mr-2" data-src="holder.js/20x20?size=1&text=.&bg=#007aff" alt="">
     <strong class="mr-auto">Bootstrap</strong>
     <small>11 mins ago</small>
+    <button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
   </div>
   <div class="toast-body">
     Hello, world! This is a toast message.
@@ -47,6 +50,9 @@ They're slightly translucent, too, so they blend over whatever they might appear
     <img class="rounded mr-2" data-src="holder.js/20x20?size=1&text=.&bg=#007aff" alt="">
     <strong class="mr-auto">Bootstrap</strong>
     <small class="text-muted">11 mins ago</small>
+    <button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
   </div>
   <div class="toast-body">
     Hello, world! This is a toast message.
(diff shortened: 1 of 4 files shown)


Johann-S (approved):
JS part looks good to me 😉

patrickhlauke:
just on the aspect of having buttons to explicitly close a toast - this kind of mixes up the difference between toasts and proper alerts i think.

toasts are generally small announcements that then disappear - they're given `aria-live` so that they get announced without shifting the user's accessibility/keyboard focus. making them stay and expecting user to close them makes them...not a toast, but an alert of sorts. normally, alerts get focused, but i gather in this changed version they don't, so IF a user wanted to close them, they'd have to navigate the whole page before reaching the close button? if so, it seems a bit of a mixed-up approach (not necessarily wrong, but not a classic toast)

xref https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/alertdialog.html

site/docs/4.1/components/toasts.md
```diff
@@ -88,10 +100,12 @@ Plus, they'll easily stack.
 ## Accessibility
 
 Toasts are intended to be small interruptions to your visitors or users, so to help those on screen readers, you should wrap your toasts in an [`aria-live` region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions). This allows screen readers the ability to see suggested interruptions without any visual cues.
+To improve accessibility level, we strongly recomend to use `autohide: false` and add a `close` button into the header to let user dismiss that element.
+You also need to adapt the `role` and `aria-live` level depending on the content. If it's an important message like error, use an `alert` role `assertive` otherwise use a role `status` with a `polite` level.
```
patrickhlauke: may be more terse and unambiguous at the end to say:

...use `role="alert" aria-live="assertive"`, otherwise use `role="status" aria-live="polite"`...

site/docs/4.1/components/toasts.md
```diff
@@ -88,10 +100,12 @@ Plus, they'll easily stack.
 ## Accessibility
 
 Toasts are intended to be small interruptions to your visitors or users, so to help those on screen readers, you should wrap your toasts in an [`aria-live` region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions). This allows screen readers the ability to see suggested interruptions without any visual cues.
+To improve accessibility level, we strongly recomend to use `autohide: false` and add a `close` button into the header to let user dismiss that element.
```
patrickhlauke: "to let users" (plural i'd say)"

Also, I wouldn't say it's necessarily an accessibility improvement. toasts/notifications that appear, are announced, and autohide are common and well supported by AT. there ARE some aspects, like users with cognitive disabilities or users that employ screen magnification without AT capabilities that may have issues with toasts, but that's a much larger discussion than just saying "don't hide it". those users would generally benefit more from explicit alert dialogs rather than toasts (dialogs that are modal, and shift focus to the dialog, etc)

patrickhlauke (commented):
A few tweaks to the prose, but I'm not 100% sure about the fundamental "don't autohide them" take

patrickhlauke:
@Lausselloic "maybe we could "allow" but not strongly recommend" yup sounds like a fair compromise
(hoping my quickfire comments didn't come across as too brusque ... haven't had enough coffee yet ;) )

Lausselloic:
I've not been hurt by your comment, by the opposite it's a pleasure to have structural debate arround accessibility :-)

PR #27161, opened 2018-08-31
Update _grid.scss

zchgdn:
Added vendor prefixes due to issues with flex not rendering on iOS 6 on Chrome or Safari. Tested on physical iOS on Chrome and Safari, and on Browserstack through multiple versions and can confirm my additions to this file rectify this. Examples below:

With prefixes - https://i.imgur.com/lfuQZ2Q.png
Without prefixes - https://i.imgur.com/VBDzwVn.png

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -21,6 +21,9 @@
 }
 
 @mixin make-row() {
+  display: -webkit-box;
+  display: -webkit-flex;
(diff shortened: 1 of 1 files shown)


wolfy1339:
Do you have autoprefixer installed when you compile Bootstrap or your custom build?
Because those aren't needed in the SCSS files. Autoprefixer adds them automatically to the generated CSS

PR #27171, opened 2018-09-03
Update _grid.scss

fedorovvldmr:
Gutters with media queries


mdo:
Sorry, but we've already decided to not go this route. Instead, we're using negative margin and padding utilities for responsive grid gutters. Slightly more verbose, but saves on a lot of generated CSS>

PR #27174, opened 2018-09-03
Added type="button" to <button>s without any type again

Herst:
Basically the same as #23866.


patrickhlauke:
the lack of `type="button"` is only problematic when it's inside a `<form>` so that's often overlooked (unless you're an ASP developer that makes the entire page a form all the time). but yes, good call, we should probably err on the side of caution with these.

patrickhlauke (approved):

patrickhlauke:
(as @mdo approved the previous one, i assume it's fine to merge right away)

PR #27175, opened 2018-09-03
Custom select validation tweaks

MartijnCuppens:
The padding of custom selects doesn't cover the background, if the text in the select box is to long, the text and background image will overlap.

 This PR also removes unneeded `calc()` functions.

Before:
image

After:
image


scss/mixins/_forms.scss
```diff
@@ -98,12 +98,12 @@
       border-color: $color;
 
       @if $enable-validation-icons {
-        padding-right: $input-height-inner;
+        padding-right: $input-height-inner * 3/4 + $custom-select-padding-x + $custom-select-indicator-padding;
 
         @if $state == "valid" {
-          background: $custom-select-background, $form-feedback-icon-valid no-repeat center right ($input-height-inner * .9) / calc(#{$input-height-inner} / 2) calc(#{$input-height-inner} / 2);
+          background: $custom-select-background, $form-feedback-icon-valid no-repeat center right ($custom-select-padding-x + $custom-select-indicator-padding) unquote("/") ($input-height-inner / 2) ($input-height-inner / 2);
```
mdo: I wonder if we take the opportunity to clean this up a bit here and turn these major values into variables? Separately, was the `/` working as intended before? Curious if Sass was trying to compute this vs keeping it as a property separator.

MartijnCuppens: A `calc()` function was used, that's why the `/` worked before.

Do you want to use the variables in the `_variables.scss` file or declare them here in the `_forms.scss`?

mdo: Ah, yeah, got it. Most likely in `_variables.scss`.

mdo (approved):

XhmikosR:
@MartijnCuppens: can you move the variable to _variables.scss?

MartijnCuppens:
I've moved some values to the variables file as mdo suggested and simplified the `if/else` construction. Also fixed the linting issue and squashed everything into one commit.

mdo (approved):

PR #27188, opened 2018-09-06
Update _buttons.scss

719media:
I'm trying to deal with the iOS Safari issue with leaving sticky hover on various elements like buttons. I was overriding the hover (and hover-focus) mixins to utilize my own methods, but ran into an issue because the hover mixin is not used on the button-outline-variant mixin.

This PR corrects this issue.


mdo:
Closing per comment in #27195.

Issue #25632, opened 2018-02-15
incompatible units 'rem' and 'px ; bootstrap\scss\_variables.scss line 420


ringodotnl:
OS: Win10, Browser Chrome latest, Bootstrap 4 (installed with npm)

Trying to include the bootstrap styling into an Angular 5+ project by adding it at styles.scss with:
@import "~bootstrap/scss/bootstrap";

When compiling  gives the following error:
[1] ERROR in ./node_modules/css-loader?{"sourceMap":false,"import":false}!./node_modules/postcss-loader/lib?{"ident":"postcss","sourceMap":false}!./node_modules/sass-loader/lib/loader.js?{"sourceMap":false,"precision":8,"includePaths":[]}!./src/styles.scss
[1] Module build failed:
[1] undefined
[1]                                        ^
[1]       Incompatible units: 'rem' and 'px'.
[1]       in D..\node_modules\bootstrap\scss\_variables.scss (line 420, column 41)

Also line 423 and line 599 do give the same issue when manual fix it by line.

wolfy1339:
The issue is not with Bootstrap and is quite clear.
px and rem are incompatible units, you can't use px values in Bootstrap 4 anymore.
Also, this issue has already been posted, please search before opening an issue

andresgalante:
Hi @ringodotnl if you are trying to customize the grid breakpoints please use `px` as it's described here:

> Make sure to set grid values in px (not rem, em, or %).
http://getbootstrap.com/docs/4.0/layout/grid/

If you want to use px's for font size, you can probably read this issue:
https://github.com/twbs/bootstrap/issues/23982

ringodotnl:
@wolfy1339  Thank you, I saw that issue but this is a different story I think. 
So I only included bootstrap scss in the project, haven't use / modify anything in the project so didn't expect to see an error at this point.

@andresgalante thank you, I haven't used bootstrap at all other then including it in the project.

wolfy1339:
That's abnormal, but Bootstrap doesn't have this problem though. :shrug:
I'm not sure what it could be other than that
________________________________

ringodotnl:
@wolfy1339  when I change the 3 lines listed before in bootstrap to 1 rem then it's solved. 
Also when I use like bootstrap/dist/css/bootstrap.min.css instead of the scss there's no problem either.

Cheers,
Ringo.

ringodotnl:
It turned out that there was loaded some other scss before bootstrap which caused the issue. 
[...]

Issue #25166, opened 2018-01-02
4.0.0-beta.3 - breakpoints are broken in Safari with new support for fractional viewport widths


liquidvisual:
Beta 3 has broken some of my sites that were using the breakpoint-down mixin. This seems to only be affecting Safari (MacOS and iOS).

I realise this has to do with the support for fractional viewport widths as [discussed here](https://github.com/twbs/bootstrap/pull/24299).

I've put together a Codepen to demonstrate this:

https://codepen.io/liquidvisual/pen/VyWMGM?editors=1100 (open in Safari)

patrickhlauke:
Due to a bug in Safari. Will see if there's some way to work around this bug.

patrickhlauke:
xref https://bugs.webkit.org/show_bug.cgi?id=178261

MartijnCuppens:
This case can easily be solved with `not`: https://codepen.io/MartijnCuppens/pen/LeLgRN/?editors=1100

However, for `media-breakpoint-between` and `media-breakpoint-only` this will not solve a thing because you can only use `not` to invert a media-query-rule and not a part of a media-query rule.
For examle `@media (min-width: $min) and not (min-width: $max)` doesn't work.

Issue #22006, opened 2017-02-16
em vs px media queries


febLey:
I looked at the bootstrap v4 source and was wondering why you are using px-based media queries, so I dug trough some issues and pull request. So I know that there has been a lot of discussion about this delicate topic (https://github.com/twbs/bootstrap/pull/17403, https://github.com/twbs/bootstrap/issues/17070), but I'd like to revisit the decision of the team.

Like some people mentioned in one the referenced issues the big benefit of em based media queries is that the layout adapts if the user has changed the base font-size, so that the content is presented in a fitting way.

I think there should be at least an variable to switch between em and px-based media queries.

mdo:
We'll be sticking with pixels given the previous discussions.

> I think there should be at least an variable to switch between em and px-based media queries.

Perhaps in the future, but for now I think we should stick with what we have and get something stable in place first :grin:

jasongilmour:
I'd go as far as suggesting that we should default to EM based media queries based on this article: https://zellwk.com/blog/media-query-units/

I'd be happy to put together a pull request with this change, or adding a variable that sets the unit for breakpoints if it was wanted.

new-mikha:
To add to @jasongilmour 's point right above, here is how the stock Bootstrap 4 example from the Bootstrap docs is looking like when the Font Size is set to Very Large in Chrome:
![image](https://user-images.githubusercontent.com/5769929/40707554-5b092670-6434-11e8-98eb-28b484bf78a1.png)

I don't think it's right :-/

PR #27190, opened 2018-09-07
Unitless breakpoints

johanlef:
If I want to customise the breakpoints using `em`, I get compatibility errors.

```SCSS
$grid-breakpoints: (
    default: 0em,
    watch: 20em,
    phone: 30em
    // etc.
}
```

It is good practice to set breakpoints in `em` instead of `px` when users change font size in their browser. See https://zellwk.com/blog/media-query-units/#concluding-the-experiments for more information why someone would like to do this. Only Safari users could get annoyed: https://adamwathan.me/dont-use-em-for-media-queries/

In any case, using a unitless number at line 42 would be very convenient.


browner12:
While it seems like this may be a bad option according to Adam's article, it makes sense to me to make this change, and allow the user to choose whether or not the Safari bug is an issue for them.

I was not familiar with SCSS math, so I did a little digging into whether this is valid, and it turns out, it works!  If you use a unitless value with a value with a unit, the unit is preserved. However, mixing units will result in an error.

```scss
.style {
    height: 500em - 100;
    width: 500px - 100;
}
```

compiles to

```css
.style {
    height: 400em;
    width: 400px;
}
```

andresgalante:
Hi @johanlef thanks a lot for your contribution.

There has been a lot of discussion around this issue, among them #22006, #25632, #19619, #25166, and others related to why we are using `px`s instead of `rem`s or `em`s

Personally, I think we should look into moving to relative units, but at this point, it'd be a breaking change for v4.

Having said that I think this is a good discussion point for v5.

andresgalante (commented):
Hi @johanlef thanks a lot for your contribution.

There has been a lot of discussion around this issue, among them #22006, #25632, #19619, #25166, and others related to why we are using `px`s instead of `rem`s or `em`s

Personally, I think we should look into moving to relative units, but at this point, it'd be a breaking change for v4.

Having said that I think this is a good discussion point for v5.

johanlef:
Okay, I understand moving to other units will be a breaking change.

@andresgalante Going from `px` to unitless on this particular line is not a breaking change: as @browner12 pointed out, the units are preserved.

browner12:
Correct me if I'm wrong, but I don't believe this would be a breaking change in v4.

As I showed in my first comment, when doing math in SCSS, if one of the values is unitless, the other units will be preserved. 

By merging this PR, all existing code would work as is, and it would allow people who choose to use `em` breakpoints to not run into a compile error.

johanlef:
@andresgalante My proposal is about removing the unit, not changing to another one. When the unit is removed, the original unit is preserved, so this change would not cause any issues.

mdo (approved):
I think this is fine? Not sold on em-based media queries yet as we've gone back and forth on that a lot, but this seems fine.

XhmikosR:
Just so that we are safe, is this backwards compatible?

mdo:
@XhmikosR I believe so—this is Sass math, and previously we required folks to use pixels only for media queries. This opens it up so others can use `em`s, and doesn't prevent folks from using pixels.

PR #27206, opened 2018-09-10
Move media from layout to components

### LINKED ISSUE #27165: Location of document page of media object
I often forget the location of the [media object page](https://getbootstrap.com/docs/4.1/layout/media-object/). I don't know why it belongs to the layout category. From [the position of the `@import`](https://github.com/twbs/bootstrap/blob/v4.1.3/scss/bootstrap.scss#L34) in the SCSS, it looks like it is treated like a component. How about moving the page into the components category?

image

MartijnCuppens:
As suggested by @ysds, move the media page from layout to components  

- Moved page in menu structure
- Moved `media-object.md` file to `components` folder
- Also searched for occurrences of `media-object` page in links in the docs (but there is also a test for 404 links which covers this, I just realised this while I'm writing this 🙂) 

Closes #27165


XhmikosR:
Just a note, if we are going with this, we need to add a redirect too.

MartijnCuppens:
@XhmikosR, I was also thinking about this, but if this is going to be changed in 4.2 and not in 4.1, it won't be necessary since we only add urls and not change them?

XhmikosR:
True, but this is a fairly new scheme. And I'd rather err on the non-broken pages around the web :)

So, if this is going to be merged in a new minor version, I guess we don't need the redirect.

XhmikosR (approved):

XhmikosR:
@mdo: are you OK with this?

mdo:
Eh, I'm not strongly attached to one direction over the other. For context, I don't see `.media` sticking around in v5 anyway, so it's dead to me either way lol. Y'all decide :).

PR #27209, opened 2018-09-11
Remove background-colors example and show only the code.

### LINKED ISSUE #27068: Background gradients not working on docs
Just noticed that the Background gradients aren't working on the **Bootstrap 4.1.3** (https://getbootstrap.com/docs/4.1/utilities/colors/#background-gradient). See below image:

screen shot 2018-08-14 at 20 02 14

if you highlight over the page, the text is clearly there for the gradients, but they're not showing:

screen shot 2018-08-14 at 20 03 01

Viewing in latest Google Chrome, Macbook Pro 15 inch.

XhmikosR:
This is to avoid confusion.

Fixes #27068


andresgalante (approved):

mdo:
Should we be more explicit in the docs and make a note that says these don't have a live example because they're purposefully disabled by default?

XhmikosR:
We could but as it seems no one reads the bold text :)

Feel free to push here any tweaks.

mdo (approved):
Made some changes to the copy, should be good now.

PR #27213, opened 2018-09-12
throw error when folks try to use a bad selector

### LINKED ISSUE #27208: Tab not chaning tab-pane when element id attribute is a UUID.
Issue occurs on Windows 10 x64 Desktop, testing with Firefox 62 (64bit) using Bootstrap 4.1.3 (Tests with Bootstrap 4.0.0 does not show this problem).

The problem is that when using UUID or GUID for id attributes for the tab-pane of the tab functionality the change in the tab itself occurs but the tab pane is not changed. No messages are logged in the browser console. The classes "show" and "active" do not appear for the tab-pane with the provided id. 

Use of shorter id names does seems to work. 

When tested against the Bootstrap 4.0.0 version the problem did not occur, and UUID/GUID could be used for element id. 

I have attached an index.html file (zipped) that shows the failing functionality with version 4.1.3 using the libraries from the CDN. Problem seems to also appear using version 4.1.2. However, version 4.1.1 works as expected.

[index.zip](https://github.com/twbs/bootstrap/files/2368284/index.zip)

Johann-S:
Fixes #27208


js/tests/unit/modal.js
```diff
@@ -607,36 +607,40 @@ $(function () {
...
-    var $toggleBtn = $('<button data-toggle="modal" data-target="&lt;div id=&quot;modal-test&quot;&gt;&lt;div class=&quot;contents&quot;&lt;div&lt;div id=&quot;close&quot; data-dismiss=&quot;modal&quot;/&gt;&lt;/div&gt;&lt;/div&gt;"/>')
-      .appendTo('#qunit-fixture')
+    try {
+      var $toggleBtn = $('<button data-toggle="modal" data-target="&lt;div id=&quot;modal-test&quot;&gt;&lt;div class=&quot;contents&quot;&lt;div&lt;div id=&quot;close&quot; data-dismiss=&quot;modal&quot;/&gt;&lt;/div&gt;&lt;/div&gt;"/>')
+        .appendTo('#qunit-fixture')
 
-    $toggleBtn.trigger('click')
-    setTimeout(function () {
+      $toggleBtn.trigger('click')
+    } catch (e) {
```
XhmikosR: How about checking for TypeError or whatever the error is? With a `match` probably.

Johann-S: It's a DOMException, but not sure if it's supported in every browser, I'll check that 😄

Johann-S: Finally it works fine by check if it's a DOMException 👍

XhmikosR (approved):

PR #27226, opened 2018-09-14
docs: switch to font-weight 600.

### LINKED ISSUE #24419: `font-weight: 500` with Segoe doesn't exist
So, at least for docs, we might want to try something else.

screen shot 2017-10-17 at 21 38 12-fullpage

XhmikosR:
Segoe UI doesn't have font-weight: 500 making things not looking as expected.

I only tested this on my Windows 7 VM, so please try it in other OS'es too.

Fixes #24419


mdo:
Should we be changing anything in the core for this as well?

mdo (approved):

mdo:
Ah gotcha, in that case, we're good to leave this in the docs. We may even be able to change from `700` to `600` there, but that's a separate effort.

XhmikosR:
Should I open an issue about the core headers weight?

mdo:
Nope, no reason for the heading change just yet.

PR #27228, opened 2018-09-14
docs: use the bundle dist file.

XhmikosR:
This is WIP because we probably need to go through the docs and make sure we list/suggest the bundle everywhere


Johann-S:
IMO we did it to talk about our bundle file everywhere in our docs 🤔

Johann-S (approved):

XhmikosR:
Let's wait for one more approval then. :)

site/_includes/scripts.html
```diff
@@ -1,12 +1,10 @@
...
 <script>window.jQuery || document.write('<script src="{{ site.baseurl }}/assets/js/vendor/jquery-slim.min.js"><\/script>')</script>
 
-<script src="{{ site.baseurl }}/docs/{{ site.docs_version }}/assets/js/vendor/popper.min.js"{% if site.github %} integrity="{{ site.cdn.popper_hash }}" crossorigin="anonymous"{% endif %}></script>
-
 {%- if site.github -%}
-  <script src="{{ site.baseurl }}/docs/{{ site.docs_version }}/dist/js/bootstrap.min.js" integrity="{{ site.cdn.js_hash }}" crossorigin="anonymous"></script>
+  <script src="{{ site.baseurl }}/docs/{{ site.docs_version }}/dist/js/bootstrap.bundle.min.js" integrity="{{ site.cdn.js_hash }}" crossorigin="anonymous"></script>
 {%- else -%}
-  <script src="{{ site.baseurl }}/docs/{{ site.docs_version }}/dist/js/bootstrap.js"></script>
+  <script src="{{ site.baseurl }}/docs/{{ site.docs_version }}/dist/js/bootstrap.bundle.js"></script>
```
mdo: This seems like a less helpful if/else block now. Should we use the minified bundle here to match other views?

mdo (approved):
I'm fine with this. No strong feelings one way or another.

site/_includes/scripts.html
mdo: Right, but we're building the minified bundle locally anyway, so why not use minified here? Is it for seeing the code easier or something while developing? In fact, why include non-minified files in our dist at all for that matter? :D

Johann-S: Because it allows folks to minify our files the way they want 😉

XhmikosR: In theory, with maps in place debugging isn't an issue.

mdo: I defer to you then @XhmikosR. I think we can use the minified one here, but I'm fine either way for now.

XhmikosR: I think it'd be better if we left this as is for now and we can revisit later. I mean, it certainly doesn't hurt and is basically only for people developing locally :)

PR #27243, opened 2018-09-15
Travis: Don't install npm@5 when the current npm is newer.

### LINKED ISSUE #27242: Travis: don't install npm 5 if it's greater than 5
https://github.com/twbs/bootstrap/blob/v4-dev/.travis.yml#L12

XhmikosR:
Fixes #27242 

My Shell knowledge isn't the best for sure, so if there's any simpler/better way to do this, let me know. It does seem to do the job, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ node_js:
   - "6"
   - "8"
 before_install:
-  - if [[ `npm -v` != 5* ]]; then npm install -g npm@5; fi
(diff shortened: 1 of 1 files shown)


XhmikosR:
So, it seems to me we can just drop the `npm@5` part altogether.

Johann-S (approved):
I thought about doing that, thank you! 👍

PR #27246, opened 2018-09-15
Update svgo.yml.

XhmikosR:


Johann-S (approved):
I'll refer to your expertise here 👍

PR #27249, opened 2018-09-16
Consistently re-use input variables

### LINKED ISSUE #26953: Form variable definitions appear inconsistent
Both forms and buttons share variable definitions form the `$input-btn-` line of variables. However some of the computed default values use the `$input-btn-` variables rather than the `$input-` ones. For instance inner height is defined as `($font-size-base * $input-btn-line-height) + ($input-btn-padding-y * 2)`. Each of these btn variables is also defined so it means that changing the input line height doesn't change the actual input height, but changing the input btn line height does, which comes off as a little surprising.

In a similar manner, the `$custom-select` variables mix both `$input-btn` and `$input` variables when all of the variables are defined by `$input`.

Overall this leads to situations where overriding all the `$input` variables only partially changes the values that are used (so for instance the custom select's height does not match the other input heights).

mdo:
Fixes #26953.

We use more `$input-` variables than `$input-btn-`, but we had a few loose ends that made it a bit inconsistent in how we re-used those. In general now, we have `$input-btn-` variables as an abstraction layer to connect our inputs and buttons together. When the styles are for inputs only, they should be using the `$input-` variables instead of `$input-btn-` ones.

/cc @twbs/css-review

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -523,7 +523,7 @@ $custom-select-padding-y:           .375rem !default;
 $custom-select-padding-x:           .75rem !default;
 $custom-select-height:              $input-height !default;
 $custom-select-indicator-padding:   1rem !default; // Extra padding to account for the presence of the background-image based indicator
-$custom-select-line-height:         $input-btn-line-height !default;
+$custom-select-line-height:         $input-line-height !default;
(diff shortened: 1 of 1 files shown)


andresgalante (approved):
👍 nice cleanup

PR #27251, opened 2018-09-17
Switch to find-unused-sass-variables for finding unused Sass variables.

XhmikosR:
This is an ongoing project with @Johann-S but I think it's better compared to using custom scripts. This way everyone can benefit :)


Johann-S (approved):
What an amazing plugins ! We should definitely add it ! 😇 👍 😆

mdo (approved):
Awesome!

PR #27255, opened 2018-09-17
Switch to `.btn-secondary` for the search button so that it's accessible

kadimi:
With the `btn-outline-success` class, the color margin is not enough as opposed to `btn-secondary`.

See Contrast Checker results for both 

- https://contrastchecker.com/?c=28a745&b=212529 (fail)
- https://contrastchecker.com/?c=ffffff&b=6c757d (pass)


andresgalante (approved):
Hi @kadimi, even though we mention this issue on [Bootstrap's documentation](http://getbootstrap.com/docs/4.1/getting-started/accessibility/#color-contrast) I think it's a good idea to change it.

mdo (approved):

XhmikosR (approved):

PR #27265, opened 2018-09-18
Change erroneous documentation for .flex-fill

### LINKED ISSUE #26803: Docs about flex-fill
https://getbootstrap.com/docs/4.1/utilities/flex/#fill

> Use the .flex-fill class on a series of sibling elements to force them into equal widths while taking up all available horizontal space. Especially useful for equal-width, or justified, navigation.

This document might cause misunderstanding.
They are not the same width, they actually depend on the width of their content.

demo: https://codepen.io/anon/pen/jKoOad

I would like to make a PR, but I am not a native speaker, so I ask for someone's help.

nicohvi:
Fixes #26803

The current documentation for `.flex-fill` indicated that the flex items would have equal widths, regardless of content. This update ensures that the documentation reflects the fact that the width of the flex items depends on their actual content.

As per https://github.com/twbs/bootstrap/issues/26803#issuecomment-411993254 a change to the actual behaviour of `.flex-fill` might be preferable, but this fix at least addresses the problems in the documentation.


XhmikosR:
@mdo: another PR was merged and didn't notice we had this one too. Not sure which one is better, so I'll let you tackle this one.

mdo (approved):

mariusa:
This change was made 1 year ago and still doesn't reflect on latest version at https://getbootstrap.com/docs/4.1/utilities/flex/#fill
...

XhmikosR:
@mariusa That's because 4.1 isn't the latest version.

mariusa:
Site should be updated:
Screenshot from 2019-11-08 09-48-02

https://getbootstrap.com/docs/4.1/utilities/flex/#fill

PR #27268, opened 2018-09-19
Add overflow-auto and overflow-hidden utilities

### LINKED ISSUE #26479: New overflow-hidden utility class?

mdo:
New PR that replaces #26509 because the fork was deleted. This doesn't carbon-copy that PR's changes though. Instead, I've opted for two utilities to start—`.overflow-auto` and `.overflow-hidden`. No responsiveness, no `x`/`y` variants, and not every option. This PR also adds documentation which was missing from previous efforts.

I've also moved the `$positions` Sass list to the `_variables.scss` file alongside the new `$overflows` list. Folks can then customize both easily enough and gain more options.

Fixes #26479.

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -6,6 +6,7 @@
 @import "utilities/embed";
 @import "utilities/flex";
 @import "utilities/float";
+@import "utilities/overflow";
 @import "utilities/position";
 @import "utilities/screenreaders";
 @import "utilities/shadows";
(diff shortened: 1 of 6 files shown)


Hermanya (approved):
Looks good to me

PR #27271, opened 2018-09-19
Add a CODEOWNERS file

mdo:
This stubs out a `CODEOWNERS` file so we can automate review pings from folks. So far I'm just including coverage for our source Sass and JS files, but this might need to get blown out to cover our tests, docs, image assets, etc.

Thoughts @twbs/team?

---

See also the [`CODEOWNERS` docs from GitHub](https://help.github.com/articles/about-codeowners/)


Johann-S:
JS Unit tests are covered too by this file 👍

XhmikosR (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+*.js    @twbs/js-review
+*.scss  @twbs/css-review
```
XhmikosR: Maybe add .css too?

PR #27275, opened 2018-09-19
Minor duplicate id in forms.md

719media:
fix duplicate id attribute


XhmikosR:
Vnu jar, yes definitely. If not either it's not a valid issue or a bug. I
will have a look tomorrow,currently afk

XhmikosR (approved):

XhmikosR:
The generated HTML was fine BTW, probably kramdown takes care of this.

XhmikosR (approved):

PR #27276, opened 2018-09-19
user-select has no effect on before/after pseudo elements

719media:
computed value of user-select on before/after pseudo elements is none, this rule does nothing

https://developer.mozilla.org/en-US/docs/Web/CSS/user-select


MartijnCuppens:
I've just tested this and it looks like Edge and IE don't respect these specs. Text of pseudo elements is selectable (see https://codepen.io/MartijnCuppens/pen/EeGprm and https://s.codepen.io/MartijnCuppens/debug/EeGprm/ZoABaKQWDYyr for IE)

But since there is no text in this pseudo element, there is nothing that can be selected so this can indeed be removed. I need some confirmation on this though.

mdo (approved):

PR #27277, opened 2018-09-19
Re-add carousel control transition

MartijnCuppens:
The carousel control transition was removed on _January 29, 2012_ because of a webkit bug, but I think we can now re-add this?


midzer (changes requested):
Hi :) Just testing in Firefox 64 nightly and Chromium 69. Transition working fine.

Maybe we could stick to `$transition-fade` instead of introducing `$carousel-control-transition`:
Difference between timing-function `linear` and `ease` is marginal for me in this scenario. More info via https://developer.mozilla.org/en-US/docs/Web/CSS/transition-timing-function#Examples

mdo (approved):

PR #27278, opened 2018-09-19
Carousel fade cleanup

MartijnCuppens:
`.carousel-fade .carousel-item` already has a specificity of `0,0,2,0` so we can safely override the transform in this selector. The `transition-duration` can also be removed, since this is set in `.carousel-item{}`.


mdo:
@XhmikosR How can we get a Netlify preview here?

XhmikosR:
Just click the Update branch button :P

mdo (approved):

PR #22392, opened 2017-04-08
Carousel fixes

mdo:
This PR fixes a few carousel issues.

- Ensures carousel works in IE10 and IE11 by removing the `transform3d` detection mixin entirely as all our supported browsers support it. There was no fallback being provided, so IE10/11 were left in the dark as we were keying off `-webkit` detection. Fixes #22198.

- Fixes the image aspect ratio issues. Also moved from `.img-fluid` to `.w-100` in our examples. Pulls in @vanduynslagerp's fix from #21799. Fixes #21658, fixes #21966, closes #21799.

- Changes the timing function from `ease-in-out` to `ease` to fix the "bouncing" into place when carousel items move. This kinda sucks, but there's no other CSS change I can see us making to fix this issue. Fixes #21685, fixes #21751, fixes #22291.

PR #27279, opened 2018-09-19
Fix Edge bounce and restore original transition easing

### LINKED ISSUE #27156: IE Edge Carousel slide animation bounces slightly on completion
Similar Issues:
#22291
#21751
#22392

Issue seems **not be be solved** with current version using Microsoft Edge 42.17134.1.0 and BS v4.1.3:
https://getbootstrap.com/docs/4.1/components/carousel/

microsoft edge 30 08 2018 19_20_52_1

MartijnCuppens:
https://github.com/twbs/bootstrap/pull/22392 didn't solve the bounce effect and changed the transition from `ease-in-out` to `ease`. This PR restores the original transition.

The bounce in #27156 is caused by the switch from `position: absolute` to `position: relative` which causes a repaint. This PR uses a float based technique which will only change the transform. 

I also removed `align-items: center;`, `.carousel-item` doesn't use flexbox.

Fixes #27156.

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -23,55 +23,32 @@
   position: relative;
   width: 100%;
   overflow: hidden;
+  @include clearfix();
 }
 
 .carousel-item {
   position: relative;
   display: none;
-  align-items: center;
+  float: left;
   width: 100%;
-  backface-visibility: hidden;
-  perspective: 1000px;
+  margin-right: -100%;
+  @include transition($carousel-transition);
 }
 
 .carousel-item.active,
 .carousel-item-next,
 .carousel-item-prev {
   display: block;
-  @include transition($carousel-transition);
 }
 
-.carousel-item-next,
-.carousel-item-prev {
-  position: absolute;
-  top: 0;
-}
-
-.carousel-item-next.carousel-item-left,
-.carousel-item-prev.carousel-item-right {
-  transform: translateX(0);
-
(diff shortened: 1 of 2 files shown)


mdo (dismissed):
Wonderful debugging <3.

XhmikosR:
I see a little jump when the transition ends on Edge.

MartijnCuppens:
I removed a lot of browser hacks to see how the browser handles the transitions now and it looks like we don't need these anymore. Removing the hacks also fixes the little jump @XhmikosR spotted in Edge. Sometimes there's still a small line between the images (probably caused by the fact that 2 transitions are going on at the same time), but I don't think that's something we can easily fix without breaking changes to the carousel.

I've tested the latest changes on native devices in:
- Chrome (macOS, Windows & Android)
- Firefox (macOS, Windows & Android)
- Edge (Windows)
- IE11 (Windows)
- IE10 (Emulated in IE11 & "native" via BrowserStack)

XhmikosR:
Thanks, @MartijnCuppens for the follow-up!

I can't see any jumping with the latest patch on 42.17134.1.0/ EdgeHTML 17.17134.

I'm gonna test it a little more and if all is good merge it.

XhmikosR:
@MartijnCuppens: do you have a patch for the Edge line this PR is introducing? I wonder how breaking it'll be.

XhmikosR (approved):

PR #27281, opened 2018-09-19
feat: .navbar-{dark|light} .btn-link

zalog:
Makes `.navbar-light/dark .btn-link` to take the same color as `.nav-link`.

Please take a look at this pen: https://codepen.io/zalog/pen/mGaaVv.
Now, in case our `btn-link` is black, it won't be visible on `.navbar-dark`.

### DIFF
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -212,7 +212,8 @@
       color: $navbar-light-hover-color;
     }
 
-    &.disabled {
+    &.disabled,
+    &:disabled {
       color: $navbar-light-disabled-color;
     }
   }
@@ -265,7 +266,8 @@
       color: $navbar-dark-hover-color;
     }
 
-    &.disabled {
+    &.disabled,
+    &:disabled {
       color: $navbar-dark-disabled-color;
     }
   }
--- a/scss/_navbar.scss
+++ b/scss/_navbar.scss
@@ -204,19 +204,20 @@
     }
   }
 
-  .navbar-nav {
-    .nav-link {
-      color: $navbar-light-color;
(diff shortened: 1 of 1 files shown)


scss/_navbar.scss
```diff
@@ -204,19 +204,20 @@
...
-        color: $navbar-light-hover-color;
-      }
+    @include hover-focus {
+      color: $navbar-light-hover-color;
+    }
 
-      &.disabled {
-        color: $navbar-light-disabled-color;
-      }
+    &.disabled {
```
mdo: If we're adding a `.btn-` class, we'll likely need `:disabled` in these selectors as well.

XhmikosR (dismissed):
Blocking as per mdo's comment

zalog:
Hmm... I forgot about this, I'll come back soon.

MartijnCuppens:
I don't really see why this is needed. You can just use `.nav-link`?

mdo (approved):
I think I'm good with this! Going to slate for v4.2.1 and ask for more thoughts from @twbs/css-review though.

MartijnCuppens:
@andresgalante, what do you think about this?

ysds:
I agree with @MartijnCuppens for v4 because:
- This can cause unexpected changes to existing users
- The [bootstrap team's theme](https://themes.getbootstrap.com/product/application/) is using `.nav-link` to similar context

XhmikosR:
@twbs/css-review: can you make the final call about this?

MartijnCuppens:
We're not going to continue with this for the arguments mentioned above. Thanks @ysds for confirmation.

PR #27284, opened 2018-09-20
optimize data URIs

midzer:
Hi all,

according to https://codepen.io/tigt/post/optimizing-svgs-in-data-uris we can squeeze some more byte out of our `data:image/svg+xml` by
* removing superflous `charset=utf8`
* lowercasing URL-encoded hex pairs, via https://github.com/tigt/mini-svg-data-uri#why-lowercase-the-url-encoded-hex-pairs

Have a nice week
midzer


XhmikosR:
Seems good, but we need to test it with IE, Safari etc.

XhmikosR:
@MartijnCuppens: any thoughts?

midzer:
Those changes are in production at https://feuerwehr-eisolzried.de/
Site is not compatible with IE11- though and I did not test on Edge lately

MartijnCuppens:
I've just tested removing the `charset=utf8` and everything works as expected on IE10, IE11 and latest Chrome, FF, Edge & Safari.

All Bootstrap hex colors are lowercased by default, I don't think we can optimise something here, @midzer?

@XhmikosR, do we have the original svgs available? We might also get a small improvement by rewriting them manually.

XhmikosR:
@MartijnCuppens: no idea, these was here a long time ago. Maybe @mdo has them.

midzer:
> All Bootstrap hex colors are lowercased by default, I don't think we can optimise something here, @midzer?

With "hex pairs" i mean `%3C` to `%3c` for "<" sign for example. According to https://github.com/tigt/mini-svg-data-uri#why-lowercase-the-url-encoded-hex-pairs this compress better for gzip

XhmikosR:
Yeah, that's escape sequences.

midzer:
> @XhmikosR, do we have the original svgs available? We might also get a small improvement by rewriting them manually.

Just did a quick run over https://jakearchibald.github.io/svgomg/ for `custom-checkbox-indicator-icon-checked`: no improvement possible, except if we scrap `xmlns` attribute which we should not do according to https://developer.mozilla.org/en-US/docs/Web/SVG/Namespaces_Crash_Course#Conclusion

MartijnCuppens (approved):
This LGTM, I've tested this on the supported browsers.

I think we can further optimise the custom radio input and the indeterminate checkbox state (these behaviours can be achieved without data uri's), but this is something outside of the scope of this PR.

mdo (approved):

XhmikosR (approved):

PR #27290, opened 2018-09-21
Hover element-state utility clasess

DougPuchalski:
Make consistent use of element-state emulating classes in hover mixin.

### DIFF
--- a/scss/mixins/_hover.scss
+++ b/scss/mixins/_hover.scss
@@ -11,8 +11,8 @@
 
 @mixin hover {
   &:hover,
-  &.hover { 
-    @content; 
+  &.hover {
+    @content;
   }
 }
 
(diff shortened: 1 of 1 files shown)


alecpl:
Imo, adding`.hover` does not make much sense to me. I also don't really know why we have `.focus`.

aceofspades:
@alecpl Honestly it's really just for troubleshooting and to be able to produce a design palette showing all of the styles. IMO there needs to be a static way to activate all styling so as not to rely on actual interactions in the browser.

mdo:
I like that last comment—having a CSS debug mode would be helpful.

I think this can be skipped for now, but I'd like to see something explore more for debugging styles. The 1px red outline is a classic one, triggering focus/hover/active states could be another, etc.

aceofspades:
@mdo What's the proper way to move forward if not with an open issue?

PR #27308, opened 2018-09-25
Add license headers in js/dist files

### LINKED ISSUE #27301: Add license headers in js/dist files

helmutgranda:
By implementing the same approach of rollup.config.js it can be replicated in build-plugins.js, with this approach individual plugins display license headers as expected.

Fixes #27301

### DIFF
--- a/build/build-plugins.js
+++ b/build/build-plugins.js
@@ -8,8 +8,8 @@
 const rollup  = require('rollup')
 const path    = require('path')
 const babel   = require('rollup-plugin-babel')
-const TEST    = process.env.NODE_ENV === 'test'
 const banner  = require(path.resolve(__dirname, './banner.js'))
+const TEST    = process.env.NODE_ENV === 'test'
 
 const plugins = [
   babel({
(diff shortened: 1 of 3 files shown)


XhmikosR:
Thanks for the PR @helmutgranda!

@Johann-S do you think we could have the license header in a common place and use it in both scripts?

Johann-S (approved):
With the banner in a different file, I'll be good 👍 

Thanks @helmutgranda 👍

build/build-plugins.js
```diff
@@ -67,6 +68,7 @@ Object.keys(bsPlugins)
       external
     }).then((bundle) => {
       bundle.write({
+        banner: banner,
```
Johann-S: `banner,` should be enough 😉

build/rollup.config.js
```diff
@@ -36,11 +35,7 @@ if (BUNDLE) {
 module.exports = {
   input: path.resolve(__dirname, '../js/src/index.js'),
   output: {
-    banner: `/*!
-  * Bootstrap v${pkg.version} (${pkg.homepage})
-  * Copyright 2011-${year} ${pkg.author}
-  * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
-  */`,
+    banner: banner,
```
Johann-S: Same here

Johann-S (dismissed):
With that changes it'll be good 👍

XhmikosR:
I'm gonna test it tomorrow but I quickly made the changes.

XhmikosR:
Works fine, thanks @helmutgranda!

@Johann-S: maybe we should add the component's name like we did in v3?

build/build-plugins.js
```diff
@@ -8,6 +8,7 @@
 const rollup  = require('rollup')
 const path    = require('path')
 const babel   = require('rollup-plugin-babel')
+const banner  = require(path.resolve(__dirname, './banner.js'))
```
Johann-S: Not a big fan of using `path.resolve` here because `require('./banner.js')` will works fine 🤔

Johann-S (approved):

build/build-plugins.js
XhmikosR: True, let's change it

PR #27310, opened 2018-09-25
js without jquery: remove IE support

alekitto:
Related to #23586
Removed IE 10 and 11 support, hacks and polyfills.

Button `bsChecked` custom attribute is not needed anymore since Edge correctly set checked attribute on `input` (tested in Edge 17 on Win10)

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -1,6 +1,5 @@
 import Data from './dom/data'
 import EventHandler from './dom/eventHandler'
-import Manipulator from './dom/manipulator'
 import SelectorEngine from './dom/selectorEngine'
 import Util from './util'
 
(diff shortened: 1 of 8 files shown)


Johann-S (approved):
LGTM 👍 
But can you fix the build error ?

alekitto:
Ops, missed one reference. Fixed now

rbonomo:
What's my manager going to do when they find out Bootstrap 5 doesn't support IE? 🐙

Johann-S:
@rbonomo it was a long time ago, we added back IE 11 support for our v5

myorangeca:
> What's my manager going to do when they find out Bootstrap 5 doesn't support IE?

You can't stop progress...

rbonomo:
@myorangeca yeah take the easy way out

Issue #27293, opened 2018-09-23
Card Columns - hover fails with scrollable card content


sakthivelsekar33:
hover css effects not applied consistently on cards columns with scrollable contents.

In the code pen - https://codepen.io/sakthivelsekar33/pen/aaxLbV

- Resize the pane so that you should see 3 cards per row
- I have applied hover css effect( background-color: red) for the list items
- go to card with title **Card 2** and hover it, you will see a red background
- now on the same card **scroll to the end** of list and hover it - **the hover effects(red background) is not applied** ( if i inspect i could see the css applied in dom but not getting applied until i resize the browser itself)
- this works fine with Card 1

**Bug Background**

- macOS high seira
- Browser chrome Version 69.0.3497.81
- Bootstrap version - 4.0

RudyHui:
I think the problem is because CSS class list-group-item property position that have default value relative.. If you set to static the problem is fixed..

In the code pen - [https://codepen.io/rudyhui/pen/YOmGqy](https://codepen.io/rudyhui/pen/YOmGqy)

helmutgranda:
I recreated your layout with BS layout classes and things seem to be working as expected. It seems to be more of an edge case:

https://codepen.io/helmutgranda/pen/gdVdYr

sakthivelsekar33:
> I think the problem is because CSS class list-group-item property position that have default value relative.. If you set to static the problem is fixed..
> 
> In the code pen - https://codepen.io/rudyhui/pen/YOmGqy

This works good and fix my issue 👍 , can you explain how this change fix the issue?

sakthivelsekar33:
> I recreated your layout with BS layout classes and things seem to be working as expected. It seems to be more of an edge case:
> 
> https://codepen.io/helmutgranda/pen/gdVdYr

I don't see you using card-column class in your code pen and the problem occurs when we use that class.

RudyHui:
Change property position in class list-group-item..
See my codepen at css part..

sakthivelsekar33:
@RudyHui As already said (https://github.com/twbs/bootstrap/issues/27293#issuecomment-425122228) it's working great. Thought of asking why changing position property affects this one?

sakthivelsekar33:
Working after changing list-group-item position css property. 👍

PR #27313, opened 2018-09-26
 Fixed issue #27293

RudyHui:
Change .list-group-item property position value from relative to static


XhmikosR (changes requested):
Please make sure CI is green.

mdo:
This cannot change, sorry. It's used for positioning the list items themselves.

PR #27323, opened 2018-09-27
add to utilities vertical aligment text-sub and text-super

joaovinicius:


XhmikosR (changes requested):
Dist files shouldn't be modified.

mdo:
Closing for bad diff, but I also don't think we need these utilities.

PR #27324, opened 2018-09-27
add to utilities text untransformed text

joaovinicius:


XhmikosR (changes requested):
Dist files shouldn't be modified.

mdo:
Since virtually every element doesn't have an initial text transform, this seems unhelpful to add right now. Going to skip on it, thanks!

PR #27325, opened 2018-09-27
Add ARM support for Chrome Launcher

ossdev07:
Hi,

Bootstrap is supported by chrome browser, to launch chrome in ARM we need --no-sandbox flag.
Added custom launcher with ---no-sandbox flag in the karma.conf.js file.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>

### DIFF
--- a/js/tests/karma.conf.js
+++ b/js/tests/karma.conf.js
@@ -28,7 +28,7 @@ const detectBrowsers = {
   usePhantomJS: false,
   postDetection(availableBrowser) {
     if (typeof process.env.TRAVIS_JOB_ID !== 'undefined' || availableBrowser.includes('Chrome')) {
-      return ['ChromeHeadless']
+      return ['HeadlessChrome']
     }
 
     if (availableBrowser.includes('Firefox')) {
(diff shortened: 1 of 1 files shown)


Johann-S:
Hi @ossdev07,

I'm sorry but I don't understand what's the interest for us to test our code with Chrome in ARM ?

XhmikosR:
Yeah, I don't think this is a good idea because this switch seems to disable some security features.

ssps12:
Hi, @Johann-S 
 I am working on arm architecture and installed bootstrap package but unable to use chrome launcher, so updated karma file.
@XhmikosR 
Maybe we can add platform based conditions and add for arm architecture

ossdev07:
Hi, @Johann-S 
 I am working on arm architecture and installed bootstrap package but unable to use chrome launcher, so updated karma file.
@XhmikosR 
Maybe we can add platform based conditions and add for arm architecture

Johann-S:
I think @ossdev07 want to update how we detect browsers before the launch of our unit tests here: https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/karma.conf.js#L29 to detect ARM architecture

XhmikosR:
Yes, but why should we do it? We don't need it.

Johann-S:
Yep I think it's the personal use of @ossdev07 not ours

ossdev07:
Hi,

I was using the bootstrap package inside the docker container.  The docker works as a root user, so was facing issues with the chrome launcher.

When tried to run the package without the docker container, the chrome launcher worked on the AARCH64.

Thanks, for the help.

PR #27336, opened 2018-09-28
Tighten stylelint config.

XhmikosR:
Maybe we should clean this up and publish the config on npm? Anyone up for it?

/CC @andresgalante @MartijnCuppens


XhmikosR:
Maybe we should clean this up and publish the config on npm? Anyone up for it?

MartijnCuppens:
Wow, the list of options is huge: https://stylelint.io/user-guide/rules/, this is going to take some time.

Are we still going to extend `stylelint-config-standard` and `stylelint-config-recommended-scss`?
I think it's more clear if we don't extend lists because then we a a complete overview in one file instead of multiple lists spreaded over multiple repo's?

XhmikosR:
Actually these rules are just from https://github.com/kristerkari/stylelint-scss#list-of-rules

I don't mind extending existent configs, because it's a good thing to my eyes.

XhmikosR:
So, does anybody have any further comments? Otherwise, I think we can merge it for now.

MartijnCuppens:
I'll have a quick look, sec

.stylelintrc
```diff
@@ -6,6 +6,7 @@
   "rules": {
     "at-rule-empty-line-before": null,
     "at-rule-name-space-after": "always",
+    "at-rule-no-unknown": null,
```
MartijnCuppens: @XhmikosR, shoudn't this be set `true`?

Johann-S (approved):
Approved because @MartijnCuppens approved it 😄 I'm not the best to review PR about our css style 😆

PR #27339, opened 2018-09-29
Add new `.rounded-pill` utility

### LINKED ISSUE #26442: New btn-rounded utility class?
Hi! I'm suggesting a simple, yet effective utility class here for buttons to be added. Whilst the Bootstrap docs does indeed have a `rounded-circle` utility class found in the border section (https://getbootstrap.com/docs/4.1/utilities/borders/#border-radius) this doesn't quite work too well when applied to buttons to essentially create a pill shaped button.

The `rounded-circle` utility class looks like this and doesn't quite work as expected when added to buttons.

```
.rounded-circle {
    border-radius: 50%!important;
}
```

I suggest creating an additional utility class with a **px** value for buttons to create a nice effect, possibly something like this...

```
.btn-rounded {
    border-radius: 50px!important;
}
```

I use rounded buttons a lot in my work and find myself having to create these classes, I don't think it would take too much to add this in the next release? 

Cheers!

sts-ryan-holton:
This PR adds a new `.rounded-pill` class as per MDO's comment on https://github.com/twbs/bootstrap/issues/26442 and also adds a new `.btn-rounded` class to the **buttons** component (https://getbootstrap.com/docs/4.1/components/buttons/)

Essentially, I've made a reusable mixin which is then applied to both the `.rounded-pill` and `btn-rounded` classes. I felt it wouldn't make sense to simply apply a `.rounded-pill` class directly to a `.btn` as MDO has mentioned in many conferences about creating modifier classes, it only makes logical sense to then create a `.btn-rounded` class to follow this structure.

This PR also updates the docs with these new classes 👍 

**Hoping we can add this into Bootstrap 4.2**, and hope to see it as a useful feature as I received positive community upvotes with this ❤️ 

/CC @XhmikosR @mdo 

Fixes #26442


scss/utilities/_borders.scss
```diff
@@ -54,6 +54,10 @@
   border-radius: 50% !important;
 }
 
+.rounded-pill {
+  border-radius: 50px !important;
```
XhmikosR: I get the same result with 20px. Can't we use a relative unit here?

sts-ryan-holton: I've always used `50px`, the problem you'll absolutely find with using `20px` is it works fine with small buttons, but larger buttons wouldn't work.

I'd recommend going with `50px` as it will guarantee to work regardless of button size. If we're to use `em` rather than `px` then it would need to be: `3.125em`

sts-ryan-holton: If you could make that adjustment that would be great, it's also a mixin too.

XhmikosR:
Personally I don't if such a big value is needed. From my quick testing `1.25rem` was enough.

/CC @MartijnCuppens for thoughts :)

scss/utilities/_borders.scss
```diff
@@ -54,6 +54,10 @@
   border-radius: 50% !important;
 }
 
+.rounded-pill {
+  border-radius: 1.5rem !important;
```
XhmikosR: Any reason why you don't use the mixin?

sts-ryan-holton: @XhmikosR  I tried using the mixin that I made, I wanted to specify `!important` as it's a utility class as well and wanted to follow a similar format to the above utilities, however, I couldn't use `!important` in the mixin file as it failed the linter and wouldn't let me build:

```
Unexpected !important (declaration-no-important) 
```
So I had to use it without the mixin here.

sts-ryan-holton: Using `!important` is bad, and I wouldn't normally use it in projects. However, from my understanding Bootstrap 4.x uses this to ensure they work.

XhmikosR: Yeah, definitely use the mixin. Just disable the check for this line only.

scss/mixins/_border-radius.scss
```diff
@@ -33,3 +33,11 @@
     border-bottom-left-radius: $radius;
   }
 }
+
+// Rounded pill used for utility class
+
+@mixin rounded-pill($radius) {
+  @if $enable-rounded {
```
MartijnCuppens: I don't think the mixin is necessary, if someone adds the utility or modifying class, I'm pretty sure they want the rounded effect. `.rounded-circle` doesn't have this either.

site/docs/4.1/utilities/borders.md
```diff
@@ -61,6 +61,7 @@ Add classes to an element to easily round its corners.
   <img data-src="holder.js/75x75" class="rounded-bottom" alt="Example bottom rounded image">
   <img data-src="holder.js/75x75" class="rounded-left" alt="Example left rounded image">
   <img data-src="holder.js/75x75" class="rounded-circle" alt="Completely round image">
+  <img data-src="holder.js/75x75" class="rounded-pill" alt="Example rounded pill image">
```
MartijnCuppens: This example is not clear, use a landscape or portrait image to indicate what the effect is.

scss/utilities/_borders.scss
```diff
@@ -54,6 +54,10 @@
   border-radius: 50% !important;
 }
 
+.rounded-pill {
+  @include rounded-pill(1.5rem);
```
MartijnCuppens: The value is not enough for the documentation page:
image

MartijnCuppens (changes requested):

scss/utilities/_borders.scss
sts-ryan-holton: This is why it was originally `3.125rem` which is more reliable.

site/docs/4.1/utilities/borders.md
sts-ryan-holton: Could probably remove it completely from here as it's only the `.btn-rounded` that is needed. I simply added a `.rounded-pill` as per MDO's suggestion of a utility class and wasn't sure where to put it.

scss/mixins/_border-radius.scss
sts-ryan-holton: We could simply have: `.btn-rounded` without a mixin? Especially if removing the one from the docs page for utilities.

sts-ryan-holton:
Revised edits:

- [x] Remove unnecessary **rounded-pill** mixin.
- [x] Update docs page.
- [x] Revise the `.btn-rounded` utility class

/cc @MartijnCuppens

site/docs/4.1/utilities/borders.md
MartijnCuppens: No, not really, the `.rounded-pill` can still be used to add the pill effect on landscape or portrait images: https://codepen.io/MartijnCuppens/pen/QZwaJv

MartijnCuppens: No, not really, the .rounded-pill can still be used to add the pill effect on landscape or portrait images: https://codepen.io/MartijnCuppens/pen/QZwaJv

MartijnCuppens (changes requested):

sts-ryan-holton:
/cc @MartijnCuppens How do you suggest implementing the changes for the `.rounded-pill` as I noticed you used `50rem` in your example.

Are you suggesting to keep the `.btn-rounded` class at `1.5rem` and have a `.rounded-pill` at `50rem`

mdo:
My comment in that PR wasn't about component-specific classes like these, but instead a new `.rounded-*` utility to that a single class can be used on any element/component.

sts-ryan-holton:
/cc @mdo  This was my original thinking, but a lot of people are likely to want to see the ability to create a rounded button using a `.btn-rounded` class and I thought it would look a little weird creating an example on the buttons component page with a `.rounded-pill` class on there?

Unless we create a section on the docs that points to the new `.rounded-pill` class?

MartijnCuppens:
@mdo, wouldn't that be a little inconsistent with the other modifying classes of `.btn`? And badges also have a modifying class (https://getbootstrap.com/docs/4.1/components/badge/#pill-badges). 

<br>

@sts-ryan-holton, 
> Are you suggesting to keep the `.btn-rounded` class at `1.5rem` and have a `.rounded-pill` at `50rem`

Yes. The `50rem` I used in the demo was just a random large number, but I think this might a good value? I would definitely make them variable, so they can be overridden easily.

sts-ryan-holton:
**I've made some changes:**

- [x] Add back `.rounded-pill` utility class.
- [x] Update `.rounded-pill` utility class to use a new variable called `$rounded-pill` set at `50rem` by default.
- [x] Provide a new custom variable for the `.btn-rounded` class so that it too can be overwritten if needed: `$btn-rounded` set at a default value of `1.5rem` (we discussed this and found that this was a good value).
- [x] Update the **borders** utility docs to include the new `.rounded-pill` class.

Note: not quite sure what to do with the docs section mentioning the new `.rounded-pill` class as it looks exactly the same as `.rounded-circle`, I did try updating the SVG image and make the width larger to greater show this, but the text on the image became stretched which looked awful. Any suggestions?

/cc @MartijnCuppens

MartijnCuppens:
@sts-ryan-holton, you can change this line  
`Rounded pill image`
to
`Rounded pill image`


which looks like this:
image

Apart from that, this PR is looking good!

MartijnCuppens (approved):
LGTM, it's up to the core members to decide whether this gets merged 😉

XhmikosR:
No need to CC us, it'll happen when it'll happen. Thanks.

whitekiw (approved):

mdo (changes requested):
Let's get right of `.btn-rounded` and it's new variable for now. We only need the utility for this.

sts-ryan-holton:
@mdo  I personally think we should keep both the `.btn-rounded` and `.rounded-pill` class since they're both different use cases and having a `.btn-rounded` modifier class makes more sense than simply attaching a `.rounded-pill` class to it. Furthermore, I'm not sure adding `.rounded-pill` to the **button** component would make much sense here.

MartijnCuppens:
I'm going to change my mind on this one and follow mdo. It does make sense to only use `.rounded-pill`. The `.btn-rounded` class only uses one property and we should use utility classes for this.

If we would add the `.btn-rounded` class we should also add the `.btn-uppercase`, `.btn-bold`, etc. classes and if we'll need to do this for every component, we'll end up with a lot of selectors that'll do exactly the same as the utility classes do.

MartijnCuppens (dismissed):
I'm going to change my mind on this one and follow mdo. It does make sense to only use `.rounded-pill`. The `.btn-rounded` class only uses one property and we should use utility classes for this.

If we would add the `.btn-rounded` class we should also add the `.btn-uppercase`, `.btn-bold`, etc. classes and if we'll need to do this for every component, we'll end up with a lot of selectors that'll do exactly the same as the utility classes do.

mdo:
Thanks @MartijnCuppens—appreciate the extra eyes and feedback. Same to @sts-ryan-holton for constantly dealing with our feedback :D.

sts-ryan-holton:
I've made the relevant changes:

- [x] Remove `.btn-rounded` class & variable.
- [x] Remove `.btn-rounded` class mentioned in docs.

mdo (approved):

PR #27343, opened 2018-10-01
Adds font weight options for form controls and custom select

### LINKED ISSUE #27327: Form input does not inherit font-weight
If the variable $font-weight-normal is set to a non-default value, e.g. "300", this value is not inherited by text inputs. The computed value of the font-weight property is found to be 'reset' to 400 in the developer tools (Firefox 62.0). When you add 'font-weight: inherit;' to the text input, this is solved. Another solution could be to add a $input-font-weight variable. 

font-weight

andresgalante:
This PR fixes #27327 by adding variables to change the font weight on form controls.

@MartijnCuppens am I missing something on custom forms?


MartijnCuppens:
Maybe we can also add something for `.custom-file-label`? If someone changes the font weight for inputs, I assume they also want this change to happen to custom file inputs?

scss/_variables.scss
```diff
@@ -423,6 +423,7 @@ $label-margin-bottom:                   .5rem !default;
 $input-padding-y:                       $input-btn-padding-y !default;
 $input-padding-x:                       $input-btn-padding-x !default;
 $input-font-size:                       $input-btn-font-size !default;
+$input-font-weight:                     $font-weight-normal !default;
```
MartijnCuppens: Maybe we can use `$font-weight-base` for this? This font weight is also used by `body`?

andresgalante:
Thanks a lot @MartijnCuppens!

I don't see the need to add a variable to change the font weight for the `custom-file-label` since that label will inherit the font weight of the body, while the input wouldn't. Am I wrong?

andresgalante:
Ok done, thanks a lot @MartijnCuppens , as always you are awesome.

XhmikosR (approved):

PR #27344, opened 2018-10-01
Updated README.md

jaydeepborkar:
The latest version as of now is 4.1.3, so updated the README download link from 4.0 to 4.1.3, have provided the download link for the same. Also, made some indentation changes at the start.


XhmikosR:
The gh-pages branch isn't supposed to be human readable at all.

@mdo: I'd say just remove README.md from that branch completely.

PR #27349, opened 2018-10-02
updated README

tanmaygupta-dev:


andresgalante (approved):

andresgalante:
@tanmay21 thanks a lot! I won't merge it since e I am not a native English speaker.

cc/ @patrickhlauke or @mdo

patrickhlauke:
Seems fairly non-controversial (as in we adhere now, rather than only in future). That whole paragraph is a horrible mouthful though...

PR #27350, opened 2018-10-02
[Refactor] Map function to replace for loop

sumleo:
Use map function to make it more functional programming
Some map function reduced the code lines


Johann-S:
Hi @sumleo,

Thank you for this PR, but on your changes, there is no need to use `map` because we use map to create a new array with just the values we want. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

Here you do not handle the returned array, a `forEach` in those cases would be better, but we choose the `for` loop because it's the best in terms of perfs

PR #27356, opened 2018-10-04
fix - unnecessary comma

spencer1573:
harassing, or otherwise -> harassing or otherwise


spencer1573:
worth a try. haha. thanks! :)

Issue #22835, opened 2017-06-15
Feature Request: Close dropdown with JavaScript $().dropdown('close')


sealocal:
**Feature Request:** *Add support for closing dropdowns through the JavaScript API*

There is one JS method for dropdowns:

```javascript
$().dropdown('toggle')  // Toggles the dropdown menu of a given navbar or tabbed navigation.
```
docs:  https://v4-alpha.getbootstrap.com/components/dropdowns/#methods

Though the alert component is not exactly a parallel, I think this would provide some consistency to the API, which provides a way to close alerts:

```javascript
$().alert('close')  // Closes an alert by removing it from the DOM. If the .fade and .show classes are present on the element, the alert will fade out before it is removed.
```
docs: https://v4-alpha.getbootstrap.com/components/alerts/#methods

Further, modals provide a similar close ('hide') feature.

docs: http://v4-alpha.getbootstrap.com/components/modal/#methods

jspizziri:
IMHO it would be good to have explicit `open` & `close` methods in addition to `toggle`.

719media:
100% agree with the above, having open/close in addition to "toggle" (which should toggle instead of just open). I spent 30 minutes debugging what i thought was my bad code only to find out that "toggle" doesn't really mean "toggle" in this case.

PR #27370, opened 2018-10-07
Added 'show' and 'hide' methods to dropdown

galczo5:
Fixes: #22835


Johann-S (changes requested):
Hi @galczo5 ,

Thank you for your PR 👍 

But can you add a bit of context, why should we need those methods ? What's your use case ?

And can you add unit test for those two methods ?

galczo5:
Hi,
In company we had an issue that we wanted to close all opened dropdowns when user select an option from one of them. We couldn't use `toggle` for it because it would open closed dropdowns. We had to keep state of each dropdown and conditionally call `toggle` method. 
So I've implemented missing `hide` method, and `show` method for consistency.

Later today I'll try to add missing unit tests.

galczo5:
Unit tests pushed. Added missing trigger of `shown` and `hidden` events.

js/src/dropdown.js
```diff
@@ -194,6 +194,61 @@ class Dropdown {
...
+    if (isActive) {
+      return
+    }
+
+    const relatedTarget = {
+      relatedTarget: this._element
+    }
+    const showEvent = $.Event(Event.SHOW, relatedTarget)
+
+    const parent   = Dropdown._getParentFromElement(this._element)
```
Johann-S: trailing spaces

js/src/dropdown.js
```diff
@@ -194,6 +194,61 @@ class Dropdown {
...
+    if (!isActive) {
+      return
+    }
+
+    const relatedTarget = {
+      relatedTarget: this._element
+    }
+    const hideEvent = $.Event(Event.HIDE, relatedTarget)
+
+    const parent   = Dropdown._getParentFromElement(this._element)
```
Johann-S: trailing spaces

Johann-S (changes requested):
Can you add more tests please, there are some lines which need tests:

image

- When it's disabled
- When it's active

And there are some line on the `hide` method too

galczo5:
I've just added requested tests. I'm not sure if `preventDefault` tests are written correctly because I'm not very experienced with karma tests. 

It tests are not good, can you give me short description how should it works?
Thanks for your help!

Johann-S:
Thank you that's better, there are just two use cases missings:

- You should check that we do nothing if the dropdown is disabled on show and hide

and it'll be good 👍 

/CC @XhmikosR what do you think about that feature ? Seems interesting to me

galczo5:
There are four tests for disabled dropdowns. One when dropdown is disabled via attribute, second when dropdown is disabled via class. For both methods.

galczo5:
Are we ready for merge?

Johann-S:
I'm waiting for other team members feedbacks

js/src/dropdown.js
```diff
@@ -194,6 +194,61 @@ class Dropdown {
       .trigger($.Event(Event.SHOWN, relatedTarget))
   }
 
+  show() {
+    if (this._element.disabled || $(this._element).hasClass(ClassName.DISABLED)) {
+      return
+    }
+    const isActive = $(this._menu).hasClass(ClassName.SHOW)
```
XhmikosR: Maybe skip `isActive` here and later one since they are used only once.

galczo5:
@XhmikosR pushed

Johann-S (changes requested):
sorry @galczo5 but I forgot one thing ! 😄 

Can you add those methods in our docs ? (`dropdown.md`)

galczo5:
@Johann-S No problem :) I will edit it today

Johann-S (approved):

galczo5:
Done :)

PR #27288, opened 2018-09-21
Release v3.4.0

XhmikosR:
@mdo @Johann-S

TODO:

- [x] Fix any 404s and useless redirects
- [x] Remove customizer's gist creation
- [x] Fix Browser tests
- [x] Go back to single colon pseudo elements for backwards compatibility with the browsers we support in v3
- [x] Fix docs search
- [x] Replace ZeroClipboard with clipboards.js since Flash is dead
- [x] Tweak clipboard.js styles
- [x] Update team
- [x] Nuke jobs from docs
- [x] Add the unprefixed `appearance` property
- [x] Add BrowserStack thanks in README.md
- [x] Exclude docs less files from Jekyll
- [x] Test builds with real apps just in case
- [x] Verify sourcemaps work
- [x] Customizer: remove drag and drop support and references, or fix this
- [x] **Make sure the SRI hashes are updated and correct!**
- [x] **Make a dist commit on non-Windows and make sure there are no modifications; if there are, commit them, then rebase the branch**
- [x] Detailed references to the XSS fixes for each component in release notes/post
- [x] **Merge https://github.com/algolia/docsearch-configs/pull/577 after the release is live**

Ship list:

- 29f9237, 13bf8ae fixed XSS issue in Alert, Carousel, Collapse, Dropdown, Modal and Tab components, see https://snyk.io/vuln/npm:bootstrap:20160627.
- 48f1482 added a `.row-no-gutters` class.
- cf6f867 Add padding to `navbar-fixed-*` on modal open
- 4f4592c and subsequent commits reorganized the v3 docs CSS to use Less.
- 1d83280 added a dropdown to the docs nav for newer and previous versions.
- 059969b update the docs to use a new baseurl, `/docs/3.4`, for that versioning life when we deploy to GitHub Pages.
- 4ebe2b8 switch to BrowserStack for tests
- e105d15 use https/fix broken links in docs
- 25d017a replace ZeroClipboard with clipboard.js
- d7ab2fb remove gist creation in customizer since anonymous gists are disabled by GitHub
- b304c61 remove drag and drop support from customizer since it didn't work anymore
- d24d47c Add docs searching via Algolia
- Remove the double border on `<abbr>` elements.

Closes #25679.

**Docs Preview: <https://twbs-340.netlify.com/>**

PR #27372, opened 2018-10-08
3.4.0 fixes

XhmikosR:
I decided to split #27288 so that it's easier to review.


docs/_includes/getting-started/download.html
```diff
@@ -42,11 +42,11 @@ <h2 id="download-cdn">Bootstrap CDN</h2>
 
   <h2 id="download-bower">Install with Bower</h2>
   <p>You can also install and manage Bootstrap's Less, CSS, JavaScript, and fonts using <a href="http://bower.io">Bower</a>:</p>
-  {% highlight bash %}$ bower install bootstrap{% endhighlight %}
+  {% highlight bash %}bower install bootstrap{% endhighlight %}
```
mdo: Will this actually get published to Bower or do we need to pull/modify this?

mdo (approved):

docs/_includes/getting-started/download.html
XhmikosR: Ah, good point. No idea about this, I just replaced the `$` symbol. Your call. If so, I will remove the references in another PR when I'll backport more stuff.

Issue #26284, opened 2018-04-12
1px black line with .embed-responsive-16by9 and YouTube video in Firefox


apiening:
Since the original issue #18261 was closed without a solution, I created a new one.

Please take a look at this very simple markup using the full screen preview in Firefox: https://www.bootply.com/IBqa3KimQL

At the bottom of the YouTube video is a black line with 1 pixel.
This is especially annoying on a white background page and a white video.
The black line does not show up with Google Chrome or Apple Safari. I have this issue only with Firefox (version 59.0.2 here).

Chaosxmk:
Tested at various screen sizes, seems more like an aliasing issue on the video itself and not an actual border from no where.

XhmikosR:
I think it's something unrelated to Bootstrap.

Just zoom in or zoom out; the black line goes away.

XhmikosR:
And I can see the same line after playing around with zoom in / out on Chrome too.

XhmikosR:
On the other hand, changing `.embed-responsive-16by9::before` padding-top to 56.20% - 56.22% fixes the problem for me.

Something weird is going on with the math the browser is doing. I also see the 1px line to the right under some resolutions.

XhmikosR:
That was my idea too, @patrickhlauke. But then I noticed that with 21by9 there's a 1px black line to the right this time, after zooming in/out.

Not sure we can make this right for all cases...

apiening:
@XhmikosR Yes you're right: When I resize the window, every now and then the black bar shows up even in Chrome.
I understand that the fractional part of the percentage value might cause different results in different browsers.
One of my customers complained that the black line is on the videos. It might get tricky to explain why it is probably not possible to remove it.
One solution would be to change the background-color of the youtube iframe to the color of the page. A white 1 px line at the bottom wouldn't be noticeable on a white page.
Unfortunately it looks like there is no option to add custom styling to the youtube iframe embed url.

XhmikosR:
I don't think the border fixes the issue. What really fixes it is changing the padding. But I can't find a number that works in all cases from a quick look...

PRs welcome as always.

patrickgrey:
In my responsive container div, I added clip-path: inset(1px 1px); Works for newer browsers except Edge. Only solution I have come up with to remove the black lines from the thumbnail.

```
.embed-container iframe, .embed-container object, .embed-container embed
{ 
[...]

PR #27374, opened 2018-10-08
Using calc() for columns width

IvanKalinin:
In Firefox the gird with widths of `*.333333%` have a "half-pixel" issue on certain window widths..
To resolve the problem, use css `calc()` to let browsers calculate actual width by themselves instead of hardcoded values.

firefox bs grid


MartijnCuppens:
Which browserversion, OS and screenwidth are you testing on? I can't really reproduce this at the moment.

IvanKalinin:
@MartijnCuppens 

bug if ff


how to reproduce:
1. The bug reproduces only in Firefox
2. The container should have dynamic width (e.g. `.container-fluid`)
3. Columns should have flex-basis. `*.3333333%` (e.g. `col-4`)
4. Add borders to be able to see the bug
5. Open the template in responsive mode and change width pixel by pixel. 
6. See the bug in some certain window width

[What a video](https://www.useloom.com/share/3f3ace8be42042448b08638f1293b8de)

XhmikosR:
Can you try with a different DPI setting too?

I've definitely hit similar inconsistencies with rounding on Firefox.

IvanKalinin:
It is interesting. 
Seems like it reproduces on certain DPIs.
It is easily reproduces on my external LG 34'' screen but when I move browser window in MacBook  screen there it no rounding problems.

XhmikosR:
For me 125% on a Windows 7 VM triggered it IIRC.

MartijnCuppens:
I could just reproduce this behaviour with browser zoom. This PR fixes the issue and I think this is a better approach than the rounding we have now.

Maybe it's worth reconsidering adding `postcss-calc` to optimise the `calc()` functions if we continue with this? (#23582)

XhmikosR:
We need to check if we have other places we need to do such changes.

And we can look into postcss-calc if this lands.

MartijnCuppens:
I did some tests with `postcss-calc`, turns out it transforms `calc()` functions like `calc(100% / 3)` into `33.33333%`, so we'll just end up with the same problem.

XhmikosR:
We can look into that later, @MartijnCuppens https://github.com/postcss/postcss-calc#preserve-default-false

MartijnCuppens (dismissed):

XhmikosR:
How about other places? I know for sure we use this in embed and we have a ticket for sure.

XhmikosR:
Found it https://github.com/twbs/bootstrap/issues/26284

Can someone tackle that in a separate PR too?

XhmikosR:
Thanks but that adds another feature too :/

It's always better to split this stuff to be able to merge faster.

Anyway, let's not get more sidetracked here :) Thanks for all the help @MartijnCuppens!

mdo:
I'm inclined to skip this in v4. It only happens on zooming or non-standard DPIs it seems. I'm also wary of adding calc expressions to all our grid classes (it's a lot of them) for performance reasons.

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

patrickhlauke:
i'll just note that increasingly there's no point talking about "non-standard DPIs", as more devices come along with different DPIs and there's no clear "standard DPIs" anymore...

XhmikosR:
Yeah, I agree with Patrick on this one. DPIs != 100% will be more and more frequent.

MartijnCuppens:
I made a PR for `postcss-calc` in https://github.com/postcss/postcss-calc/pull/61. With this option we could simplify some calc expressions like `calc(100% / 4)` to `25%`.

mdo:
Noted on the DPI conversation. I'm still inclined to leave this as-is, but let me think on it a bit more.

mdo:
Let this one sit and coming back to it, going to punt on it. Thanks, though!

ysds:
I believe this is worth revisiting for v5.

ysds:
oh I mistakenly re-opened this instead of issue. Close this and create a new issue for clarity.

Issue #27377, opened 2018-10-08
Deprecate theme-color, gray, color functions


MartijnCuppens:
As suggested here https://github.com/twbs/bootstrap/issues/27338#issuecomment-425655377, we can deprecate the `theme-color`, `gray`, `color` functions. The `color` and `gray` functions are not used by Bootstrap and it's easy to remove the `theme-color` dependency. These functions don't really add anything to Bootstrap.

With a small changes we can make Bootstrap independent of these functions and throw a warning like we did here, so we can deprecate them in a next major release:
https://github.com/twbs/bootstrap/blob/f59c2d946c276ccc247a34ccfe6f5063c847306b/scss/mixins/_text-hide.scss#L10-L12

XhmikosR:
Personally I'd say go for this, but perhaps wait for @mdo and @andresgalante to chime in.

We just need to make sure we don't introduce breaking changes in minor/patch releases.

mdo:
I think the expectation I had here was that functions like `theme-color` would generate the different shades of each color, allowing us to easily create a family of one color's shades (e.g., our `$gray-100` through `$gray-900`). I imagine we can nuke these in v5 if we're not using them though.

mdo:
I believe this was resolved.

PR #27378, opened 2018-10-08
Simplify theme color usage

### LINKED ISSUE #27338: It is not possible to properly set $link-color variable in user-land config.
In your `_variables.scss`
```scss
$link-color:                theme-color("primary") !default;
```
Your docs (https://getbootstrap.com/docs/4.1/getting-started/theming/#variable-defaults) says

> However, when overriding across Sass files, your overrides must come before you import Bootstrap’s Sass files.

But it is impossible to just change `$key` for `theme-color()` function, before import `_variables.scss`, because theme-color expects `$theme-colors` that declared in `_variables.scss`.
It turns out a vicious circle.
I know i can just use a variable instead of function call, but it is a bit confusing. Why you use `theme-color("primary")` instead of just `$primary`?

MartijnCuppens:
Use variables instead of function. 

Fixes #27338. (I've made a new ticket about the deprecation of the function (#27377), because that wasn't really what the ticket was about originally)


XhmikosR:
So this is backward compatible, right, @MartijnCuppens ?

mdo:
Yeah per @XhmikosR I wonder if this counts as backward compatible. It's not a huge change, and wondering if we just save it for what comes next @XhmikosR? But also this diff is pretty small and the functions use the variable anyway, so probably safe?

XhmikosR:
I think this is OK because it's an internal-only change in _variables.scss. The removal of the function, if it happens, it'll be a separate branch.

Just waiting for @MartijnCuppens's confirmation to be safe :)

mdo (approved):

MartijnCuppens:
We're on the same level here, this merge would be pretty safe. The removal of the functions is not in the scope of this PR.

PR #27379, opened 2018-10-08
v3.4.0 fixes - batch 2

XhmikosR:
@mdo: here's the second batch of fixes for v3

I went with the full new URL in the JS files etc, so that we save the redirect.


XhmikosR:
Ping @mdo

Johann-S (approved):

Issue #27380, opened 2018-10-09
"disabled" class doesn't disable the link


matthewspace:
One feature that I'd like to request is regarding the `.nav-link.disabled` class.

![code](https://user-images.githubusercontent.com/38481138/46644977-8ee51480-cb50-11e8-899e-ca78197cf8c3.PNG)

Adding the `disabled` class to the link only changes the color of the link; however it would be convenient if the class kept the default mouse pointer state and when clicked it wouldn't redirect you to the specified URL. This is what it currently looks like when you hover your mouse over the disabled link:

![screenshot](https://user-images.githubusercontent.com/38481138/46645159-9a850b00-cb51-11e8-955b-88d945337e82.png)

I'd say changing the `.nav-link.disabled` class up a bit would be nice.

```css
.nav-link.disabled {
  color: #6c757d;
  pointer-events: none;
  cursor: default;
}
```

rafalp:
Disabling links is little more tricky than that, as documentation for disabled `btn-link` mentions:

> The .disabled class uses pointer-events: none to try to disable the link functionality of <a>s, but that CSS property is not yet standardized. In addition, **even in browsers that do support pointer-events: none, keyboard navigation remains unaffected**, meaning that sighted keyboard users and users of assistive technologies will still be able to activate these links. So to be safe, add a tabindex="-1" attribute on these links (to prevent them from receiving keyboard focus) and use custom JavaScript to disable their functionality.

mdo:
Closing per last comment.

PR #27382, opened 2018-10-09
Made the disabled state for nav more obvious.

arvind0598:
Disabled pointer events and changed cursor to default as per #27380


XhmikosR:
https://github.com/twbs/bootstrap/blob/v4-dev/.stylelintrc#L40-L250

scss/_nav.scss
```diff
@@ -46,6 +48,8 @@
 
     &.disabled {
       color: $nav-link-disabled-color;
+      pointer-events: none;
+      cursor: default;
```
mdo: These two lines are unnecessary if the base `.nav-link` already gets it from L25-26.

mdo (dismissed):

arvind0598:
@XhmikosR @mdo I think I've done the requested changes. Sorry for the delay.

scss/_nav.scss
arvind0598: Done!

MartijnCuppens (approved):
LGTM 👍

Issue #27006, opened 2018-07-31
tabs.js _activate is too restrictive


mnlipp:
I have a slightly more complicated tab layout than usual. The tab has an additional icon.
```
<li class="nav-item text-nowrap" data-portlet-tab="portlet-tab-2">
  <span class="nav-link">
    <a id="portlet-tab-2-tab" data-toggle="tab" href="#portlet-tab-2-pane" role="tab" class="active show" aria-selected="true">Hello World</a>
    <a href="#" class="fa fa-times ml-1 portlet-tab-close"></a>
  </span>
</li>
```
This *would* work if the `active` class was properly removed from the tab. But the selector for `active` is too restrictive.
```
        if (container.nodeName === 'UL') {
          activeElements = $$$1(container).find(Selector.ACTIVE_UL);
        } else {
          activeElements = $$$1(container).children(Selector.ACTIVE);
        }
```
The element with the `active` class (the one with the `data-toggle`) must be an immediate child of the UL, but in my case, it isn't. I don't think that this very restrictive rule is necessary, the active element can also be a grand-child (as in my case) or even further down in the hierarchy without causing problems.

PR #23087, opened 2017-07-14
tabs: fix nested tabs.

XhmikosR:
When one uses say a carousel inside a tab, the `.active` selector previously matches the carousel ones too leading to broken tabs.

It's not the perfect solution but should the job for now.

/CC @Johann-S 

PS. The `data-target` solution you suggested seems definitely better for the long term. I just don't have the time to implement it myself so I resorted to a quick fix for now :)

Fixes #22654.

PR #27388, opened 2018-10-09
Unnecessarily restrictive children method

sergchernata:
Children method severely limits the markup customization of this plugin. We should be using find instead. As previously highlighted by #27006.

### DIFF
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -145,7 +145,7 @@ class Tab {
     if (container.nodeName === 'UL') {
       activeElements = $(container).find(Selector.ACTIVE_UL)
     } else {
-      activeElements = $(container).children(Selector.ACTIVE)
+      activeElements = $(container).find(Selector.ACTIVE)
     }
 
     const active = activeElements[0]


Johann-S (changes requested):
Can you add a bit of context here ?

What's your use case ?

sergchernata:
Besides the issue I linked to, [here is another example](https://stackoverflow.com/questions/48814154/bootstrap-4-nested-navigation-link-opens-tab-content-only-the-first-time/52726523#52726523).

When navigation is customized and anchor elements are not direct children the plugin breaks. It's an unnecessary limitation. I see no harm, only upside.

Was there a specific reason for use of `children`?

Johann-S:
Yep we did that because of if you use a carousel in side a tabs you'll have a conflict between active classes 😟 see: https://github.com/twbs/bootstrap/pull/23087

XhmikosR:
@Johann-S: I thought we had a test case for this which I'd expect to see it fail?

Johann-S:
I was surprised too @XhmikosR 🤔 We'll have to fix that unit test.

@sergchernata this method is used to activate the tabs and the tab-content so not sure if it'll work but you should try 👍

I'm more than open if we can find a flexible solution

sergchernata:
[Here's a working pen with my proposition](https://codepen.io/anon/pen/LgyLvz), based on one of yours from a related issue.

Is there a different example I should be testing in?

Johann-S:
It seems it works well, can you push your changes here ?

XhmikosR:
We definitely need to add a test that covers the aforementioned case though...

sergchernata:
@XhmikosR I can help test if you link me to that case.

XhmikosR:
Well, I'd swear we had a test case for this, but it's been a long time, so I don't even remember. I'll need to dig into git log and PR/issues. Maybe @Johann-S remembers about it.

sergchernata:
@Johann-S my solution had an error, I'll do more testing.

XhmikosR:
So, any news @sergchernata?

XhmikosR:
Closing this due to lack of response.

PR #27396, opened 2018-10-11
fix image dimensions for holder.js

m5o:
* fix image dimensions for`holder.js`
* replace `px` with `x`

:octocat: https://github.com/imsky/holder#usage


XhmikosR:
This is definitely broken for me, please check the preview.

XhmikosR:
Actually, this is wrong. `p` stands for percentage.

Issue #26779, opened 2018-06-29
Add vh 25,50,75 and 100


guledali:
https://getbootstrap.com/docs/4.1/utilities/sizing/

desvu:
It's a good idea. It will be very useful.

guledali:
```
/* Margin */

.mt-25 {
   margin-top: 25% !important;
}
.mt-50{
    margin-top: 50% !important;
}
.mt-75{
    margin-top: 75% !important;
}
.mt-100{
    margin-top: 100% !important;
}

.mr-25 {
   margin-right: 25% !important;
}
.mr-50{
    margin-right: 50% !important;
}
.mr-75{
    margin-right: 75% !important;
}
.mr-100{
    margin-right: 100% !important;
}

.mb-25 {
    margin-bottom: 25% !important;
}
.mb-50{
    margin-bottom: 50% !important;
}
.mb-75{
    margin-bottom: 75% !important;
}
.mb-100{
    margin-bottom: 100% !important;
}

.ml-25 {
   margin-left: 25% !important;
}
.ml-50{
    margin-left: 50% !important;
}
.ml-75{
    margin-left: 75% !important;
}
.ml-100{
    margin-left: 100% !important;
}

/* Margin ends */


/* Padding */

.pt-25 {
   padding-top: 25% !important;
}
.pt-50{
    padding-top: 50% !important;
}
.pt-75{
    padding-top: 75% !important;
}
.pt-100{
    padding-top: 100% !important;
}

.pr-25 {
   padding-right: 25% !important;
}
.pr-50{
    padding-right: 50% !important;
}
.pr-75{
    padding-right: 75% !important;
}
.pr-100{
    padding-right: 100% !important;
}

.pb-25 {
    padding-bottom: 25% !important;
}
.pb-50{
    padding-bottom: 50% !important;
}
.pb-75{
    padding-bottom: 75% !important;
}
.pb-100{
    padding-bottom: 100% !important;
}

.pl-25 {
   padding-left: 25% !important;
}
.pl-50{
    padding-left: 50% !important;
}
.pl-75{
    padding-left: 75% !important;
}
.pl-100{
    padding-left: 100% !important;
}

/* Padding ends */
```

galczo5:
I can try to do it.

guledali:
@galczo5 Please also add margin-padding utilities that I have posted above. I'm kinda tired of including tacyons css along side bootstrap.

galczo5:
[...]

PR #27397, opened 2018-10-11
Added sizing and spacing based on viewport

galczo5:
Solution for https://github.com/twbs/bootstrap/issues/26779


XhmikosR:
Please make sure all CI is green.

galczo5:
I'm not sure what I should do. I've added some scss and output size check failed. Should I change value of check value or try to minimalize my code?

XhmikosR:
Yeah, then maybe this shouldn't be added at all. Or if we do really want this feature we will bump the limits.

mdo:
Thanks, but I've turned this down before and will do so again. We don't need responsive variations for everything and this just adds too much CSS.

galczo5:
No problem :) I've been thinking about this for few hours today and you guys have right. In bootstrap there is too many utilities.

PR #27400, opened 2018-10-11
refine deprecated holder.js color syntax

m5o:
* refine deprecated holder.js color syntax
  * remove `#55595c:#373a3c` which is only used once
* harmonize with all other holder.js examples

### DIFF
--- a/site/docs/4.1/components/card.md
+++ b/site/docs/4.1/components/card.md
@@ -382,7 +382,7 @@ Turn an image into a card background and overlay your card's text. Depending on 
 
 {% capture example %}
 <div class="card bg-dark text-white">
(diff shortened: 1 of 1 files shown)


XhmikosR:
Is this the only case?

XhmikosR:
Please check the preview, it seems broken here https://deploy-preview-27400--twbs-bootstrap4.netlify.com/docs/4.1/components/card/#image-overlays

XhmikosR (changes requested):
Please revert the `p` removal.

m5o:
> Is this the only case?

yes

XhmikosR (approved):

PR #27407, opened 2018-10-12
Fixing #26372: disabled btn hover issue

### LINKED ISSUE #26372: When buttons is disabled and gradients are enabled, hover state changes background
When a buttons is disabled and gradients are enabled. The hover state of a disabled button change to button hover gradient background. I think this should stay in the disabled state background.

Now it looks weird when disabled and hover changes background of button.

Can't reproduce a codepen example, because I can't enabled gradients without compiling the scss. I think the solution is to change the `background-color` to `background` in the `button-variant` mixin here: https://github.com/twbs/bootstrap/blob/4bc5cc149a104edda9da7d439c2b45810d180d89/scss/mixins/_buttons.scss#L32

- Operating system and version: Mac OS 10.13.4
- Browser and version: Chrome 66.0.3359.117

ianwalter:
When gradients are enabled there is still a hover state on disabled buttons since the hover rules apply to background-image and disabled rules apply to background-color. This applies the logic already present in dropdowns to buttons. This fix was originally proposed by @ysds.

Fixes #26372


mdo:
I think this makes sense. Sounds a lot like what we did in v3.

mdo (approved):

PR #27408, opened 2018-10-13
Assorted fixes

XhmikosR:
The popper.js situation is something we need to find a solution for, because we will definitely hit it again... We need a way to be in sync.


XhmikosR:
@Johann-S: shouldn't we have a test which we expect to fail with invalid selectors?

Johann-S:
Yep this one https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/util.js#L23

Johann-S (approved):

XhmikosR:
@Johann-S: yes, but shouldn't [this](https://github.com/twbs/bootstrap/pull/27408/files#diff-d449587e83fca68f7cfb33ea5468344eL53) throw and thus that test fail?

Johann-S:
Yep but we don't test our documentation in our unit test

XhmikosR:
That's not in the docs though, that's in tests.

PR #27410, opened 2018-10-13
Rename docs production script to production.

XhmikosR:
~~I guess eventually we could change this to `JEKYLL_ENV=production`.~~

EDIT:

I added a patch which uses `JEKYLL_ENV` since I find that cleaner. Let me know what you think.


XhmikosR:
I added a patch which uses `JEKYLL_ENV` since I find that cleaner. Let me know what you think.

Johann-S (approved):

PR #27443, opened 2018-10-15
Clean up docs Sass code.

XhmikosR:
* remove unused rules
* use Sass nesting in more places
* use the core mixins in more places
* use the color variables more


site/docs/4.1/assets/scss/_brand.scss
```diff
@@ -10,62 +8,44 @@
...
-//   color: #fff;
-//   border-radius: 50%;
-// }
-// .bd-brand-item .glyphicon-ok {
-//   background-color: #5cb85c;
-// }
-// .bd-brand-item .glyphicon-remove {
-//   background-color: #d9534f;
-// }
+  // Heading content within
```
MartijnCuppens: There are no headings in `.bd-brand-item`, this can be removed.

XhmikosR: True, but note that I didn't remove all unused CSS. Just whatever I found while digging at it.

MartijnCuppens:
This is a nice cleanup. I think there is room for more improvements (eg. more usage of utility classes where needed), but that's something we can look into later.

mdo (approved):
Diff looks good!

XhmikosR:
@MartijnCuppens: feel free to clean the docs Sass more later; I'm pretty sure we have more unused stuff there :)

PR #27449, opened 2018-10-15
Gemfile: add WDM gem.

XhmikosR:
It's recommended on Windows so that watch isn't (so) slow.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,10 @@
-source 'https://rubygems.org'
+source "https://rubygems.org"
 
-group :development, :test do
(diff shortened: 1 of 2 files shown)


Johann-S (approved):

XhmikosR:
OK this doesn't work with `--deployment`...

XhmikosR:
So the options are:

1. we remove the `deployment` flag in Travis
2. we always install WDM regardless of the platform; not sure about this since I haven't tried other OS'es
3. we drop this patch completely

PR #27453, opened 2018-10-16
let's add a contributing file, so we can work better!

Planet-Ocean:
let's add a contributing file, so we can work better!


Johann-S:
Please can you create a correct PR ? BTW our contributing file is here: https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md

PR #27455, opened 2018-10-16
Refactor `Modal.dispose` and add test

### LINKED ISSUE #27456: Modal is missing default value for `_isTransitioning`
`_isTransitioning` member is used in `Modal` without being declared

unrevised6419:
Having variables initialised from start `_isTransitioning` is better.  
Would be better to add an ~~[no-undef eslint rule](https://eslint.org/docs/rules/no-undef)~~ to check for undefined variables use.  
Reordered variables by priority at `show` and `hide` function entrance.

* [x] fix `Modal.dispose` method
* [x] add test for `Modal.dispose`
* [x] add `_isTransitioning` default value

`no-undef` - this rule is enabled by default, but does not what we want
`no-use-before-define` - kind of does what we want, but does not work for class members

Resolves #27456

### DIFF
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -81,6 +81,7 @@ class Modal {
     this._isShown             = false
     this._isBodyOverflowing   = false
     this._ignoreBackdropClick = false
+    this._isTransitioning     = false
     this._scrollbarWidth      = 0
   }
 
@@ -101,7 +102,7 @@ class Modal {
   }
 
   show(relatedTarget) {
-    if (this._isTransitioning || this._isShown) {
+    if (this._isShown || this._isTransitioning) {
       return
     }
 
@@ -153,7 +154,7 @@ class Modal {
       event.preventDefault()
     }
 
-    if (this._isTransitioning || !this._isShown) {
+    if (!this._isShown || this._isTransitioning) {
       return
     }
 
@@ -206,6 +207,7 @@ class Modal {
     this._isShown             = null
     this._isBodyOverflowing   = null
     this._ignoreBackdropClick = null
+    this._isTransitioning     = null
     this._scrollbarWidth      = null
   }
 
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -196,9 +196,17 @@ class Modal {
   }
 
   dispose() {
-    $.removeData(this._element, DATA_KEY)
+    [window, this._element, this._dialog]
+      .forEach((htmlElement) => $(htmlElement).off(EVENT_KEY))
 
-    $(window, document, this._element, this._backdrop).off(EVENT_KEY)
+    /**
+     * `document` has 2 events `Event.FOCUSIN` and `Event.CLICK_DATA_API`
+     * Do not move `document` in `htmlElements` array
+     * It will remove `Event.CLICK_DATA_API` event that should remain
+     */
+    $(document).off(Event.FOCUSIN)
+
+    $.removeData(this._element, DATA_KEY)
 
     this._config              = null
     this._element             = null
--- a/js/tests/unit/modal.js
+++ b/js/tests/unit/modal.js
@@ -697,4 +697,37 @@ $(function () {
 
     beginTimestamp = Date.now()
   })
+
+  QUnit.test('should dispose modal', function (assert) {
+    assert.expect(3)
+    var done = assert.async()
+
+    var $modal = $([
+      '<div id="modal-test">',
+      '  <div class="modal-dialog">',
+      '    <div class="modal-content">',
+      '      <div class="modal-body" />',
+      '    </div>',
+      '  </div>',
+      '</div>'
+    ].join('')).appendTo('#qunit-fixture')
+
+    $modal.on('shown.bs.modal', function () {
+      var spy = sinon.spy($.fn, 'off')
+
+      $(this).bootstrapModal('dispose')
+
+      const modalDataApiEvent = $._data(document, 'events').click
+        .find((e) => e.namespace === 'bs.data-api.modal')
+
+      assert.ok(typeof $(this).data('bs.modal') === 'undefined', 'modal data object was disposed')
+
+      assert.ok(spy.callCount === 4, '`jQuery.off` was called')
+
+      assert.ok(typeof modalDataApiEvent !== 'undefined', '`Event.CLICK_DATA_API` on `document` was not removed')
+
+      $.fn.off.restore()
+      done()
+    }).bootstrapModal('show')
+  })
 })


iamandrewluca:
I don't have so much experience with coveralls, what does this means?  
I think a test for dispose should be added.  

image

Johann-S:
Yep we have a lot of unit test we should add 😄 
Coveralls show you where you have increased/decreased the coverage.

We don't want a 100% of coverage but at least 90% would be great 👍

iamandrewluca:
@Johann-S what does this selector match?

https://github.com/twbs/bootstrap/blob/5d834e05107edebbdac98e0b2d51940f13190e8f/js/src/modal.js#L200

I want to test that line, but don't understand on what element event is turned off.  
Is that a nested selector, or parallel selector?

Johann-S:
@iamandrewluca this line remove all the event listeners we attached during the use of our modal

iamandrewluca:
@Johann-S yes, I understand this.  
But it removes listeners on each one? `window` `document` `element` and `backdrop`? Or that will search for backdrop inside element inside document inside window? jQuery does not have such docs for multiple arguments constructor. It only has that one with context, but it works in reverse order of arguments.

```js
jQuery(element1, element2)
```
This will try to find element1 inside element2.

Johann-S:
Yep for me it should remove the event listeners on each elements and that's what we should check

js/tests/unit/modal.js
```diff
@@ -658,4 +658,17 @@ $(function () {
...
       .bootstrapModal('hide')
   })
+
+  QUnit.test('should dispose modal', function (assert) {
+    assert.expect(1)
+    var done = assert.async()
+
+    var $modal = $('<div id="modal-test"/>')
+      .bootstrapModal('show')
+      .bootstrapModal('hide')
```
Johann-S: I don't understand the need to use `show/hide` methods here 🤔 

```js
var $modal = $('<div id="modal-test"/>').bootstrapModal()

assert.ok(typeof $modal.data('bs.modal') !== 'undefined', 'modal created')

$modal.bootstrapModal('dispose')

assert.ok(typeof $modal.data('bs.modal') === 'undefined', 'modal data object was disposed')
```

If you want to go further you can use a SinonJS spy to be sure `dispose` is called

iamandrewluca: From start `bs.modal` is undefined, then is set on `show`, then is removed again on `hide`

iamandrewluca: Disposing modal in `show` state is ambiguous. All listeners are removed, and you're stuck with an open modal on page.

iamandrewluca: image

Johann-S: The modal instance is created here: https://github.com/twbs/bootstrap/blob/94a0144b2cfa5d50b2976295da9d07465acb22ad/js/src/modal.js#L509
And me remove the data on `dispose` here: https://github.com/twbs/bootstrap/blob/94a0144b2cfa5d50b2976295da9d07465acb22ad/js/src/modal.js#L199

No need to show or hide

Johann-S: See my CodePen: https://codepen.io/Johann-S/pen/jepKdY

iamandrewluca: let's wait till GitHub will work again :smile:

iamandrewluca: To `dispose` a modal you have to `show` it first.  
Otherwise `$('#myModal').data('bs.modal')` will be `undefined`, and `dispose` won't have any effect.  
If you `dispose` it in `show` state you're blocked with an open modal on page.  

image

Also if you dispose a modal in transition state (with fade `show -> shown`), you get a TypeError  

image

https://codepen.io/iamandrewluca/pen/aRRbXb

This is why I `show/hide` modal, then `dispose`

Johann-S: My CodePen show the opposite please take time to Open it with your devtools open

iamandrewluca: Your CodePen works as it should be. There you only `show` the modal.  
Try to dispose it after `show`

Johann-S: Yep but what I'm showing, it's, you don't have to call `show` to create an instance of our modal.
Because my CodePen JS is just:

```js
$(function () {
  $('#myModal').modal()
  
  console.warn($('#myModal').data('bs.modal'))
})

Comment (iamandrewluca) [reply js/tests/unit/modal.js]: You mean I have just to call `$('#myModal').modal()` ?
This 2 are the same
```js
$('#myModal').modal() // will call `show` method
$('#myModal').modal('show') // will call `show` method
```
On tests page I get an error if I don't hide modal before dispose, when trying to click on something

```js
QUnit.test('should dispose modal', function (assert) {
  assert.expect(1)
  var done = assert.async()

  var $modal = $('<div id="modal-test"/>')
    .bootstrapModal()
    .bootstrapModal('dispose')

  assert.ok(typeof $modal.data('bs.modal') === 'undefined', 'modal data object was disposed')
  done()
})
```
image

Also if I don't hide, body will have the padding added when there is overflow.

Johann-S: Tested that : 
```js
  QUnit.test('should dispose modal', function (assert) {
    assert.expect(2)
    var $el = $('<div id="modal-test"/>')
    $el.bootstrapModal()

    assert.ok($el.data('bs.modal') !== undefined)

    $el.bootstrapModal('dispose')

    assert.ok($el.data('bs.modal') === undefined)
  })
```

and it works fine on our unit tests

iamandrewluca: Tried to click any input on the tests page? you will get an error, and also body padding is added.

Johann-S: That's not the purpose here, we test if our `dispose` methods do the job that's all, so we just have to check if the data is removed and if the listeners too.

I did the first check (check if the data is removed).
We did that on other tests for our Modal and it's fine.

Johann-S: If you want we can discuss about that on the Bootstrap Slack 😉

iamandrewluca:
@Johann-S Done.  
some notes:  
this is not vlaid jQuery
```js
$(window, document, this._element, this._backdrop).off(EVENT_KEY)
```

after `dispose` only `Event.CLICK_DATA_API` should remain on page

js/src/modal.js
```diff
@@ -197,7 +198,21 @@ class Modal {
...
 
-    $(window, document, this._element, this._backdrop).off(EVENT_KEY)
+    /**
+     * Bellow statement is not a valid
+     * $(window, document, this._element, ...).off(EVENT_KEY)
+     * jQuery does not support this type of constructor
+     * eg: https://codepen.io/iamandrewluca/pen/yRWpoV
+     */
+    const htmlElements = [window, this._element, this._dialog]
+    htmlElements.forEach((htmlElement) => $(htmlElement).off(EVENT_KEY))
```
Johann-S:
```js
[window, this._element, this._dialog].forEach((el) => $(el).off(EVENT_KEY))

File: js/tests/unit/modal.js [hunk 3]
```diff
@@ -697,4 +697,50 @@ $(function () {
...
+      '  <div class="modal-dialog">',
+      '    <div class="modal-content">',
+      '      <div class="modal-body" />',
+      '    </div>',
+      '  </div>',
+      '</div>'
+    ].join('')).appendTo('#qunit-fixture')
+
+    $modal.on('shown.bs.modal', function () {
+      var data = $(this).data()
```
Johann-S:
```js
var modal = $(this).data('bs.modal')

File: js/src/modal.js [hunk 4]
```diff
@@ -197,7 +198,21 @@ class Modal {
   dispose() {
     $.removeData(this._element, DATA_KEY)
 
-    $(window, document, this._element, this._backdrop).off(EVENT_KEY)
+    /**
```
Johann-S: Can you remove this comment please ? We don't need to keep the history in comment, we have git which handle that for us 😉

js/tests/unit/modal.js
```diff
@@ -697,4 +697,50 @@ $(function () {
...
+      var element = data['bs.modal']._element
+      var dialog = data['bs.modal']._dialog
+
+      $(this).bootstrapModal('dispose')
+
+      assert.ok(typeof data['bs.modal'] === 'undefined', 'modal data object was disposed')
+
+      const modalEvents = [window, document, element, dialog]
```
Johann-S: Instead of doing that, you should use SinonJS spy to be sure we call `.off` on those elements

Johann-S (changes requested):

js/src/modal.js
iamandrewluca: this will not work, example
image

Johann-S (approved):

PR #27457, opened 2018-10-17
Make jquery and popper.js dependencies

perrin4869:
Reading through issues, https://github.com/twbs/bootstrap/issues/23204 where it was decided to make `jQuery` and `popper.js` peer dependencies, I think that the main reason for the decision was to minimize the number of packages installed by `npm`. However, `peerDependencies` are required too, and emit a warning if they are not found... they are not really optional, so it does not solve the problem.
Also, peer dependencies are mostly designed for cases like when using `react-router`, you will inevitably be using `react` too... but in this case, chances are that even if you use `bootstrap`, you will not be using those other 2 dependencies on their own.
Moreover, both `jQuery` and `popper.js` are marked as `external` in https://github.com/twbs/bootstrap/blob/v4-dev/build/rollup.config.js, so they are not really bundled in. In a way they could even be considered `devDependencies`, because neither package is used in practice... but in principle they are imported in the source files, so they are technically normal deps.
This is not a breaking change, since it does not change the usage of the package in any way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5912,6 +5912,11 @@
         "topo": "2.x.x"
       }
     },
+    "jquery": {
+      "version": "3.3.1",
+      "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.3.1.tgz",
+      "integrity": "sha512-Ubldcmxp5np52/ENotGxlLe6aGMvmF4R8S6tZjsP6Knsaxd/xp3Zrh50cG93lR6nPXyUFwzN3ZSOQI0wRJNdGg=="
+    },
     "js-base64": {
       "version": "2.4.9",
       "resolved": "https://registry.npmjs.org/js-base64/-/js-base64-2.4.9.tgz",
@@ -7893,8 +7898,7 @@
     "popper.js": {
       "version": "1.14.4",
       "resolved": "https://registry.npmjs.org/popper.js/-/popper.js-1.14.4.tgz",
-      "integrity": "sha1-juwdj/AqWjoVLdQ0FKFce3n9abY=",
-      "dev": true
+      "integrity": "sha1-juwdj/AqWjoVLdQ0FKFce3n9abY="
     },
     "portfinder": {
       "version": "1.0.17",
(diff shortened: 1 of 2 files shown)


wolfy1339:
What about users that only want the CSS? That was one of the main concerns raised

perrin4869:
@wolfy1339 I am actually one of them, I use `react-bootstrap`. But users that only want the CSS will still be warned about missing peer dependencies every time they run `npm install`. Also, how is this a problem to users who only want the CSS?

ocarreterom:
I think it is a peer dependency because in `/dist/js/bootstrap.bundle.js` you have a bundle with all dependencies.

XhmikosR:
^^yup, minus jQuery.

perrin4869:
@ocarreterom how does that make it a peer dependency rather than a dependency though?
Actually, since jQuery is never even bundled into the builds, I think that in theory you could completely remove that dependency and the repository would work the same lol

Johann-S:
We chose to make them as peer deps, because we don't want to bother folks who don't need the JS part of Bootstrap.
I don't think we'll revisit that change, maybe `optionalDependencies` would fit our JS deps better 🤔

perrin4869:
@Johann-S, the thing is, you do actually bother the folks who don't need the JS part, like me, with warnings about missing dependencies... that's the main motivation behind this PR.

Johann-S:
I can understand the concern about having a warning, but that's not an error and you don't have physically in your `node_modules` useless libs (jQuery, Popper.js)

XhmikosR:
Yeah, but this is just a warning. You can ignore it and move on without installing stuff you won't use.

XhmikosR:
v5 will not use jQuery anyway. For the rest we can revisit later.

perrin4869:
I see. I still think that a mistake is being made here, using peer dependencies as though they are optional... Peer dependencies are not supposed to be optional, as far as I know, that's what optional dependencies are for. And honestly, I'd go all the way and make those normal dependencies rather than optional, because for better or worse, the JS is part of the repo...
I think that you are overestimating the problem with having useless libs in `node_modules`...
And `v5` is a long way off
But I'll leave it here

XhmikosR:
v5 is not a long way off. People don't get that we are normal people ourselves with work, real lives etc.

Either way, this was a choice in the past and we'd rather leave it as for now.

PR #27462, opened 2018-10-17
Remove redundant width and height properties from caret

MartijnCuppens:
The pseudo elements don't need these properties.


XhmikosR:
Must be some reason these were added, maybe @mdo recalls why.

MartijnCuppens:
The caret used to be an element in v3 and the the `width:0;` and `height: 0;` were probably added to prevent issues with elements that weren't `:empty`
https://getbootstrap.com/docs/3.3/css/#helper-classes-carets

XhmikosR:
Yup, that must be the reason indeed. I haven't used v3 for years thus I don't recall all details :)

mdo (approved):

PR #27476, opened 2018-10-19
Fix input placeholder color on IE Edge

ghost:
Changing `$input-placeholder-color` has no effect on IE Edge—this commit fixes that.


XhmikosR:
We use autoprefixer so this is not needed.

rolandasb:
Doh! My bad.

PR #27478, opened 2018-10-19
docs search: take into account the current URL.

### LINKED ISSUE #27463: Change docs search to take into account the current domain
So that Netlify previews' search doesn't send us to the main docs pages.

XhmikosR:
Fixes #27463


Johann-S (approved):
Works fine ! 👍

PR #27491, opened 2018-10-21
Revert "Remove custom example plugin."

### LINKED ISSUE #27395: example.html doesn't replace the whole holder.js line
This

```liquid
{{- include.content | replace: 'data-src="holder.js', 'src="...' -}}
```

Matches only the `holder.js` part and not the whole string. As far as I can tell there's no regex support here, so I'm not sure how to proceed.

/CC @m5o

XhmikosR:
This reverts commit d01b4eb02560ba67b61b1b2c0496e9f8281c154d (PR #25784).

This implementation has many limitations/issues.

Closes #27395


mdo:
Is this just the holder.js stuff or something else as well? I imagine we could replace holder.js with some SVGs, 1x1 GIFs, etc?

XhmikosR:
Yup, mostly holder.js. I also suggested SVGs in #27395 but I didn't see any traction. Plus Liquid does not do regular expressions so this cannot be extended any more than exact matches...

All this will be useless with Hugo anyway, but since I don't know when I'll get to finish that, you know.

mdo (approved):

m5o:
I don't see these _many limitations/issues_, to be honest. There is [one issue](https://github.com/twbs/bootstrap/issues/27395), to be precise.

Isn't the real question to ask, what's the main focus of the html markup example? It's primarly showing and teaching the element structure and bootstrap class naming. The `src` of the image is secondary.

There are some [image placeholder](https://www.johanbostrom.se/blog/the-best-image-placeholder-services-on-the-web) services available with a similar url structure as `holder.js`. Easiest would, replace it by e.g. `//imgplaceholder.com` instead of `...` ✌️

m5o:
I don't see these _many limitations/issues_, to be honest. There is [one issue](https://github.com/twbs/bootstrap/issues/27395), to be precise.

Isn't the real question to ask, what's the main focus of the html markup example? It's primarly showing and teaching the element structure and bootstrap class naming. The `src` of the image is secondary.

There are some [image placeholder](https://www.johanbostrom.se/blog/the-best-image-placeholder-services-on-the-web) services available with a similar url structure as `holder.js`. Easiest would, replace it by e.g. `//imgplaceholder.com` instead of `...` ✌️

m5o:
I don't see these _many limitations/issues_, to be honest. There is [one issue](https://github.com/twbs/bootstrap/issues/27395), to be precise.

Isn't the real question to ask, what's the main focus of the html markup example? It's primarly showing and teaching the element structure and bootstrap class naming. The `src` of the image is secondary.

There are some [image placeholder](https://www.johanbostrom.se/blog/the-best-image-placeholder-services-on-the-web) services available with a similar url structure as `holder.js`. Easiest would, replace it by e.g. `//imgplaceholder.com` instead of `...` ✌️

XhmikosR:
Well we are not tied to any solution. I just need something that works properly and the current solution does **not**. It's so simple.

XhmikosR:
Well we are not tied to any solution. I just need something that works properly and the current solution does **not**. It's so simple.

XhmikosR:
Well we are not tied to any solution. I just need something that works properly and the current solution does **not**. It's so simple.

XhmikosR:
Putting this on hold for now. I still don't care about using the plugin or not as long as it works for our needs. That being said, it's a good chance to replace holder.js with pure SVGs.

mdo:
Removed from v4.2 given the hold.

Next time, let's put it on hold and not lock the conversation here. No need to do that as it sends the wrong message after receiving feedback—dissenting opinions are fine, even if they trigger notifications. :D

XhmikosR:
The lock was when there was that github issue and I was getting repeated
comments about the same thing. I unlocked it after it was resolved as you
can see.

Issue #27486, opened 2018-10-20
Should treat *.png as binary in .gitattributes


shanyu:
On Ubuntu 16.04, Apache tez 0.9.1 build failed with bower install:

[...]

PR #27492, opened 2018-10-21
Revert "Simplify .gitattributes."

XhmikosR:
This reverts commit 1c78f703e08f1288208c60655967ff590cbe69cc.

Unfortunately some Linux distros use an ancient git version and this change requires git >= 2.10.

We should revisit later.

Refs #27486


Johann-S:
It's not possible for them to update their git version ?

Johann-S (approved):

PR #27500, opened 2018-10-22
Cleanup stylelint comments

MartijnCuppens:
`stylelint-disable` and `stylelint-enable` is used a lot while this can also be achived with `stylelint-disable-next-line` 

I've also added some comments to clarify why I made some changes.

### DIFF
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -25,11 +25,10 @@
 // If you want the anchor version, it requires `href="#"`.
 // See https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile
 
-// stylelint-disable property-no-vendor-prefix, selector-no-qualifying-type
+// stylelint-disable-next-line selector-no-qualifying-type
 button.close {
   padding: 0;
   background-color: transparent;
   border: 0;
-  -webkit-appearance: none;
+  appearance: none;
 }
-// stylelint-enable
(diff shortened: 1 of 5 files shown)


scss/_close.scss
```diff
@@ -25,11 +25,10 @@
 // If you want the anchor version, it requires `href="#"`.
 // See https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile
 
-// stylelint-disable property-no-vendor-prefix, selector-no-qualifying-type
+// stylelint-disable-next-line selector-no-qualifying-type
 button.close {
   padding: 0;
   background-color: transparent;
   border: 0;
-  -webkit-appearance: none;
```
MartijnCuppens: `autoprefixer` takes care of this

scss/_reboot.scss
```diff
@@ -163,12 +161,10 @@ dfn {
   font-style: italic; // Add the correct font style in Android 4.3-
 }
 
-// stylelint-disable font-weight-notation
```
MartijnCuppens: There was no need for this (see https://github.com/stylelint/stylelint/tree/master/lib/rules/font-weight-notation#font-weight-notation)

scss/_dropdown.scss
```diff
@@ -105,7 +105,7 @@
 }
 
 // When enabled Popper.js, reset basic dropdown position
-// stylelint-disable no-duplicate-selectors
+// stylelint-disable-next-line no-duplicate-selectors
```
XhmikosR: I was hoping we could reduce the amount of duplicate selectors when possible later :)

XhmikosR (approved):
LGTM, thanks!

PR #27503, opened 2018-10-23
Update Travis CI config.

XhmikosR:
Only run `coveralls` and `check-broken-links` tasks when we run the Test phase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ install:
   - npm install
 script:
   - npm test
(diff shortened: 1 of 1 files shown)


XhmikosR:
Hmm, I'd need to test this more, I don't see the npm cache being restored.

XhmikosR:
I believe it's because the `NPM_CACHE_FOLDER` variable isn't available when extracting the cache files.

Johann-S (approved):
Nice work 👍

XhmikosR:
it doesn't work right yet :/

PR #27529, opened 2018-10-28
Carousel crossfade > Prevent the background to be shown when transitioning

MartijnCuppens:
When the carousel is transitioning, the background is shown through the images. This PR prevents this from happening by disabling the transition of the carousel item that is fading out and raising the `z-index` of the carousel item that is fading in. I also needed to raise the `z-index` of the carousel controls, the indicators had a `z-index` which was high enough.

I had to add the `transition-delay` to the `getTransitionDurationFromElement` function to prevent the classes to be removed too soon. Not sure if we should rename this function to make this clear or not.

Issue:
https://getbootstrap.com/docs/4.1/components/carousel/#crossfade

Demo new carousel:
https://deploy-preview-27529--twbs-bootstrap4.netlify.com/docs/4.1/components/carousel/#crossfade


midzer (commented):
I can clearly experience an improved crossfade on your demo. Awesome, will definetly use it in some of my projects soonish. JS code looks flawlessly on first glance and CSS appears tidy, too.

MartijnCuppens:
Thanks @midzer!

Johann-S (changes requested):
Works fine to me 👍 

Can you add a unit test, which show we handle `transition-delay` too please ?

js/src/util.js
```diff
@@ -93,17 +93,21 @@ const Util = {
...
+    if (!floatTransitionDuration && !floatTransitionDelay) {
       return 0
     }
 
     // If multiple durations are defined, take the first
     transitionDuration = transitionDuration.split(',')[0]
+    transitionDelay = transitionDelay.split(',')[0]
 
-    return parseFloat(transitionDuration) * MILLISECONDS_MULTIPLIER
+    return (parseFloat(transitionDuration) + parseFloat(transitionDelay)) * MILLISECONDS_MULTIPLIER
```
XhmikosR: I wonder, can't we combine this with the same effect? Like `parseFloat(transitionDuration + transitionDelay)`

scss/_variables.scss
```diff
@@ -1006,7 +1006,8 @@ $carousel-control-icon-width:        20px !default;
 $carousel-control-prev-icon-bg:      str-replace(url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='#{$carousel-control-color}' viewBox='0 0 8 8'%3e%3cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3e%3c/svg%3e"), "#", "%23") !default;
 $carousel-control-next-icon-bg:      str-replace(url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='#{$carousel-control-color}' viewBox='0 0 8 8'%3e%3cpath d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3e%3c/svg%3e"), "#", "%23") !default;
 
-$carousel-transition:                transform .6s ease !default; // Define transform transition first if using multiple transitions (e.g., `transform 2s ease, opacity .5s ease-out`)
+$carousel-transition-duration:       .6s !default;
```
XhmikosR: So this is only added so that we get the value in JS?

js/src/util.js
MartijnCuppens: Don't think so. `transitionDuration` and `transitionDelay` are stings like `0.6s`.

scss/_variables.scss
MartijnCuppens: And to prevent the transition of the carousel item that is "fading out". We need this because otherwise the classes are removed immediately.

XhmikosR: Please try it because it should automatically be coerced.

js/src/util.js
XhmikosR: Please try it because it should automatically be coerced.

scss/_variables.scss
XhmikosR: Just wondering if we could use the transition shorthand, that's why I'm asking.

js/src/util.js
Johann-S: I don't think so 😟 
image

XhmikosR: Ok, thanks for confirming :)

scss/_variables.scss
MartijnCuppens: I can use the shorthand to combine these lines indeed:
https://github.com/twbs/bootstrap/blob/9b3c66258997c10f8bdb1ea521ed64e3775da39d/scss/_carousel.scss#L100-L101
I'll look into that

But I'll still need the `$carousel-transition-duration` option as a separate variable.

Johann-S (approved):

XhmikosR:
@MartijnCuppens so this seems to break on IE 10.

See for example https://github.com/twbs/bootstrap/pull/27529

PR #27530, opened 2018-10-28
Delete sache.json

MartijnCuppens:
Looks like we're not listed anyway (http://www.sache.in/search?query=bootstrap) and I don't think it's worth keeping this file in the root. The project also looks a bit abandoned (https://github.com/jhardy/sache).


XhmikosR:
@glebm: this shouldn't affect bootstrap-sass in any way, right?

glebm:
@XhmikosR That's right

XhmikosR (approved):

PR #27533, opened 2018-10-28
Fix tab fade out

### LINKED ISSUE #26525: Unnecessary delay before fading-in new tab pane - tab.js
Using version 4.1.0, I found out a delay before new tab pane is shown with JS behavior and fade effect.
This delay is the transition-duration value of the fade utility class.

Issue comes from the `_activate` method, [line 144](https://github.com/twbs/bootstrap/blob/8f7bd419935adfcd53c471a0202083464800619e/js/src/tab.js#L144) in tab.js file.

There is a listener added to the previously activated tab pane for the `transitionend` event, but no classes are added to this element after that, so only the emulator can triggers the callback which activates the new one.

Have a look at the [demo](https://getbootstrap.com/docs/4.1/components/navs/#javascript-behavior) in the official doc.

MartijnCuppens:
The tab fadeout doesn't work because the `.show` class isn't removed. This causes some lag before the new tab is faded in.

**Current behaviour:**
https://getbootstrap.com/docs/4.1/components/navs/#javascript-behavior

**Behaviour with fix:**
https://deploy-preview-27533--twbs-bootstrap4.netlify.com/docs/4.1/components/navs/#javascript-behavior
Beautiful, isn't it?

Closes #26525

### DIFF
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -162,6 +162,7 @@ class Tab {
       const transitionDuration = Util.getTransitionDurationFromElement(active)
 
       $(active)
+        .removeClass(`${ClassName.SHOW}`)
         .one(Util.TRANSITION_END, complete)
         .emulateTransitionEnd(transitionDuration)
     } else {
@@ -171,7 +172,7 @@ class Tab {
 
   _transitionComplete(element, active, callback) {
     if (active) {
-      $(active).removeClass(`${ClassName.SHOW} ${ClassName.ACTIVE}`)
+      $(active).removeClass(`${ClassName.ACTIVE}`)
 
       const dropdownChild = $(active.parentNode).find(
         Selector.DROPDOWN_ACTIVE_CHILD
--- a/js/src/tab.js
+++ b/js/src/tab.js
@@ -162,7 +162,7 @@ class Tab {
       const transitionDuration = Util.getTransitionDurationFromElement(active)
 
       $(active)
-        .removeClass(`${ClassName.SHOW}`)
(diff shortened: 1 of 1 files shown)


Johann-S (changes requested):
It's done on `_transitionComplete` here: https://github.com/twbs/bootstrap/blob/2781ccbf1afcccfd5fc31651988fea363514e175/js/src/tab.js#L175

Maybe it's not in the good order, so you should remove it when it's not the good place.

Johann-S:
Is it related to : https://github.com/twbs/bootstrap/issues/26525 @MartijnCuppens ? 🤔

MartijnCuppens:
Yes it is related to that ticket, thanks for mentioning that, @Johann-S.

We can indeed also remove the `.show` class from L175, at that point the class is already removed now.

Johann-S:
Not sure if there is a way to test that 🤔

MartijnCuppens:
Don't think so

Johann-S (changes requested):

js/src/tab.js
```diff
@@ -162,6 +162,7 @@ class Tab {
       const transitionDuration = Util.getTransitionDurationFromElement(active)
 
       $(active)
+        .removeClass(ClassName.SHOW)
         .one(Util.TRANSITION_END, complete)
         .emulateTransitionEnd(transitionDuration)
     } else {
```
MichelleKwa126: ACTIVE

Johann-S (approved):

PR #27544, opened 2018-10-29
download page: add Yarn

matbrgz:


XhmikosR:
The thing with this is I don't think it should be a separate section personally.

Johann-S:
But this sections is called `Package managers` to it should list all the package managers we support 🤔 

https://getbootstrap.com/docs/4.1/getting-started/download/#package-managers

XhmikosR:
That's true. It just doesn't make a lot of sense to me but anyway.

XhmikosR (approved):

MatheusRV:
> The thing with this is I don't think it should be a separate section personally.

I fully understand your consideration, NPM and Yarn are pretty much the same thing, but we have to recognize that many developers want ease by using ctrl + c and ctrl + v to add Bootstrap to their projects.

PR #27552, opened 2018-10-30
Jekyll: switch to localhost

XhmikosR:
`0.0.0.0` has issues on Windows.


patrickhlauke:
with that change, i can't run `npm run docs-serve" on the machine and access the docs from another machine on my local network (using the IP of the machine, e.g. `http://192.168.1.163:9001` - the request from the other machine/phone/device times out). with the current `0.0.0.0` i can still use `http://localhost:9001` on my local machine, AND i can access the docs from any other machine in my network using the `http://192.168.1.163:9001` (depending on IP of machine) address...

XhmikosR:
Well, yeah but doesn't work for me then.

patrickhlauke:
adding some context from the slack chat: what isn't currently working with using `0.0.0.0` is that redirects end up using that address, and then don't work (because the magic `0.0.0.0` isn't actually valid, and gets changed behind the scenes...but not, it seems, by the redirect plugin)

patrickhlauke (changes requested):
as discussed, i'd like to keep the ability to access the docs running on one machine from another machine (like a phone/tablet) on the same network. using `localhost` breaks this. xref https://zarino.co.uk/post/jekyll-local-network/

a different approach - perhaps using some sort of "if running locally for dev purposes, use this config" which could set the baseurl or url to localhost explicitly, making redirects work then - would be better perhaps

XhmikosR:
Like I said on Slack, you can use a custom `--host` flag. The default should work for most people.

patrickhlauke (approved):

PR #27553, opened 2018-10-30
New variable $component-active-bg-light

stfnschrmr:
Adds separate variable `$component-active-bg-light` to avoid multiple use of `lighten($component-active-bg, 35%)` or `rgba($component-active-bg, .25)`. Also ensures that using native css custom properties as a value for `$component-active-bg` don't lead to an error when transpilling.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -249,7 +249,7 @@ $box-shadow-lg:               0 1rem 3rem rgba($black, .175) !default;
 
 $component-active-color:      $white !default;
 $component-active-bg:         $primary !default;
-$component-active-bg-light:   lighten($primary, 35%) !default;
+$component-active-bg-light:   lighten($component-active-bg, 35%) !default;
 
 $caret-width:                 .3em !default;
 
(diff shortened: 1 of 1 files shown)


wolfy1339 (changes requested):
Missing a few spots here to change the variable names

MartijnCuppens:
Hi @createdwithlove, 
Nice to see you again. 

Caching some functioncalls in a variable is a good idea, but you can't just change the alpha version of the `$input-focus-border-color`, this will cause the border to disappear when the inputfield is focused.

This PR
image

Bootstrap:
image

It's a bit hard to see in the screenshots, but you can compare your version with the bootstrap docs:
- https://getbootstrap.com/docs/4.1/components/forms/#overview
- https://deploy-preview-27553--twbs-bootstrap4.netlify.com/docs/4.1/components/forms/#form-controls

createdwithlove:
Hi @MartijnCuppens.
You're right, that's a little detail that should not be lost. The calculation of the alpha value then could be done in the `form-control-focus()` mixin? But I don't know if that's worth it.

MartijnCuppens:
I don't think so, at the moment you can control the `border-color` and the `box-shadow` of the focus state and these colors can be changed independently. Calculating this in the mixin would make this impossible.

I'm going to close this because it's not worth adding another variable for this case. But thanks for the PR, if you spot something else that can be improved, let us know!

PR #27566, opened 2018-10-31
Fix empty custom-control-label alignment issue

### LINKED ISSUE #26453: PR #25944 broke custom-control-label when empty
Empty labels are needed when using checkboxes in horizontal forms. If you let it empty in this use case on 4.1.1, the checkbox appears way off in the bottom: https://codepen.io/anon/pen/ZoJqyd

As you can see, it was working perfectly on 4.1.0: https://codepen.io/anon/pen/VxzEMp

I tried to look into every commit between 4.1.0 and 4.1.1 and found that PR #25944 was to blame. When I reverted its changes, the behavior was as before. Because this fix looked more than useful, I found a way to keep it and get my use case working as intended.

My proposed fix is to add `vertical-align: top` to the `custom-control-label` class:
```css
.custom-control-label {
  vertical-align: top;
}
```
Here is a demo: https://codepen.io/anon/pen/NMvLEw

Since I don't know the bootstrap codebase & good practices, I hope someone else can come up with a PR though.

ysds:
Should we go ahead to find a solution with flexbox for #26453?  But that will require more time. This PR is a simple quick fix.

Before: https://codepen.io/anon/pen/yRdegO
After: https://codepen.io/anon/pen/bmPEgX

Fixes #26453

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -69,6 +69,7 @@
 .custom-control-label {
   position: relative;
   margin-bottom: 0;
+  vertical-align: top;
 
   // Background-color and (when enabled) gradient
   &::before {


MartijnCuppens (approved):
LGTM 👍This fix doesn't influence the alignment in other case, so it looks pretty safe to implement.

PR #27570, opened 2018-10-31
Avoid null value

VincentLanglet:
I have a stylesheet I used for both amp pages and classic pages.
But for amp, i removed some breakpoint in order to generate less css.
So I can't compile if I use media-breakpoint-down(md).
This fix prevents this error


VincentLanglet:
HI @MartijnCuppens 

Let's say i have style
```
// _button.scss
.btn {
  margin: 5px;

  @include media-breakpoint-down(lg) {
    margin: 0;
  }
}
```

And i have to index files, one for the website
```
// index.scss
@import '~bootrap'
@import 'button'
```

And another for the amp version of the website
```
// index_amp.scss
$grid-breakpoint: (sm: 0) // No need to generate md, lg and xl classes.
@import 'bootstrap'
@import 'button'
```

The index_amp generation will fail.
The null check i add in the mixin can avoid this.

Another case can be a typo, if someone use @include media-breakpoint-down(m).
There is no problem with the media-breakpoint-up mixin. (media-breakpoint-up(m) compile correctly)

MartijnCuppens (approved):
Thanks VincentLanglet,

In that case this fix could indeed solve the issue. Looks good to me!

mdo:
Do we need to re-evaluate any other mixins for similar conditions?

MartijnCuppens:
@mdo, as @VincentLanglet mentioned:
> Another case can be a typo, if someone use @ include media-breakpoint-down(m).
There is no problem with the media-breakpoint-up mixin. (media-breakpoint-up(m) compile correctly)

It's `$n < length($breakpoint-names)` that is causing the issue here and this check is only used in the `breakpoint-next` mixin.

mdo (approved):

PR #27573, opened 2018-11-01
Make meaning of tooltip's 'selector' option more clear

shaedrich:
Based on #4215


Johann-S (approved):

shaedrich:
@XhmikosR @twbs Should this be changed for v3 as well?

PR #25871, opened 2018-03-16
Fix readonly-plain-text with button addon

caiquecastro:
Fixes #25870

Issue #25870, opened 2018-03-16
readonly-plain-text with button addon


caiquecastro:
The readonly plain text doesnt work with button addon.

Live example: https://codepen.io/caiquecastro/pen/eMdmYo

![image](https://user-images.githubusercontent.com/3316550/37526649-b245d4ee-290e-11e8-87bc-27c0189d9e8e.png)

```
<div class="input-group mb-3">
  <div class="input-group-prepend">
    <button class="btn btn-outline-secondary" type="button">Button</button>
  </div>
  <input type="text" readonly class="form-control-plaintext" id="staticEmail" value="email@example.com">
</div>
```

patrickhlauke:
why would you use it this way? simply dropping the `.input-group-prepend` solves this of course. it's not really possible, to my knowledge, to rewrite styles so that the button takes account of an element that comes *after* it.

patrickhlauke:
ah, i thought you meant the right-hand border of the button not being rounded, and missed the email showing on a new line...doh

mdo:
I understand the desired affect here, but why would you do this? An input group is intended to visually connect button/text and form control.

patrickhlauke:
i'd imagine for consistency if the form is read/write at one point/for certain users, but some of the fields are read-only after submission/for other users, so you can keep the scaffolding the same

caiquecastro:
I have it readonly, but if I click it. I should remove its readonly state, and enable change the value.

mdo:
Why not use only the [`readonly` attribute](https://getbootstrap.com/docs/4.0/components/forms/#readonly) then? If the idea is that at some point a user could interact with that field, I would think it should be a field from the beginning.

patrickhlauke:
regardless of the specific use case here, though, i'd say the fix (or something similar to it) seems legit to me.

Defatul:
Line 3316
append -> .input-group > .form-control-plaintext,

Line 3353
append -> .input-group > .form-control-plaintext:not(:last-child),

Line 3361
.input-group > .form-control-plaintext:not(:first-child),

Line 3441
.input-group-lg > .form-control-plaintext,

Line 3460
.input-group-sm > .form-control-plaintext,

PR #27579, opened 2018-11-02
fix(input-group): Support form-control-plaintext as a sibling

tmorehouse:
PR #25871 added support for `.form-control-plaintext` inside an input-group (addressing issue #25870), but it did not include handling the situation of multiple inputs in the same input group (where `.form-control-plaintext` is not the first input).

This PR adds in the missing line to handle this situation.

### DIFF
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -23,6 +23,7 @@
     margin-bottom: 0;
 
     + .form-control,
+    + .form-control-plaintext,
     + .custom-select,
     + .custom-file {
       margin-left: -$input-border-width;


tmorehouse:
Hmmm now that I look at this, I just realized that `.form-control-plaintext` inside an input group doesn't have a border. And it is missing the left and right padding (to line it's contents up with other controls that may be in the same form) when inside an input-group.

MartijnCuppens:
No need for this PR, you can use the `readonly` attribute for this as mdo mentioned https://github.com/twbs/bootstrap/issues/25870#issuecomment-374037198

tmorehouse:
@MartijnCuppens that is a good point re `readonly`. I wonder if that suggestion should be mentioned in the input group documentation?

Issue #27531, opened 2018-10-28
Not able to scroll with touch support for carousels


MartijnCuppens:
Branch: `v4-dev`

It's not possible to scroll the page on touch devices if you start scrolling from a carousel. Tested on Chrome & FF on android and in mobile emulation mode in chrome on macOS (it's probably on every touch device). Caused by #25776.

cc @Johann-S

Johann-S:
It seems that's because we add : `pointer-events: none` on the carousel 🤔 

/CC @patrickhlauke

patrickhlauke:
nothing to do with `pointer-events: none;`. it's `touch-action: none;`. ideally, we'd have it set to `touch-action: pan-y` to allow the browser to handle vertical panning/scrolling, but Firefox in particular is far too fickle about what constitutes a horizontal-only movement and fires `pointercancel` as soon as you stray even just a few pixels up/down while swiping left/right, making it almost impossible to swipe horizontally there.

note that the scrolling is also impossible on devices that just support touch events (e.g. Safari/iOS), for different reasons (as the touch events are swallowed as they happen, and default browser action is prevented).

there's no clean way around this at the moment.

XhmikosR:
Just a reminder, we can always go back to the hammer approach if we have issues we can't solve with the current implementation.

Johann-S:
Unfortunately, using HammerJS won't fix that, because they do what we do 😟

patrickhlauke:
the choice we have is basically: accept that this is what happens (as there's no good way around it, at least in touch-events scenario), or use the `touch-action: pan-y` approach but accept that currently Firefox is fickle in this regard and more often than not it won't allow you to swipe because if's too eager to `pointercancel` (but other, more popular browsers - well, Chrome... - are ok).

if we go for the former, it may be worth noting in the docs. then again, for the latter, a little note about Firefox's current behavior may also be worth it (i've flagged this with Mozilla folks at last week's TPAC, FWIW...and could file a bugzilla bug about it)

XhmikosR:
I'm not sure how we should proceed with this after all. I guess we should aim for the best compatibility.

PR #27582, opened 2018-11-02
Carousel fixes

patrickhlauke:
Replace touch-action: none with pan-y, remove preventDefault from touch event handling

This makes scrolling and carousels coexist. Some minor drawbacks. that I guess we can live with:

- it doesn't prevent browser gestures in touch-events scenarios (primarily iOS, where swiping from the side will trigger the browser's back/forward functionality, but this is common across these scenarios)
- Firefox on touchscreen-enabled laptops/desktops with high-dpi will throw `pointercancel` and abort a swipe gesture as soon as there's even a minimal amount of vertical movement to the swipe - this seems to be a bug/overeager behavior on the part of FX; testing on a medium-dpi Android, Firefox there seems to behave sanely

Closes https://github.com/twbs/bootstrap/issues/27531

/cc @MartijnCuppens @Johann-S

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -283,14 +283,11 @@ class Carousel {
       if (this._pointerEvent && (event.originalEvent.pointerType === PointerType.TOUCH || event.originalEvent.pointerType === PointerType.PEN)) {
         this.touchStartX = event.originalEvent.clientX
       } else if (!this._pointerEvent) {
-        event.preventDefault()
         this.touchStartX = event.originalEvent.touches[0].clientX
       }
     }
 
     const move = (event) => {
-      event.preventDefault()
-
       // ensure swiping with one touch and not pinching
       if (event.originalEvent.touches && event.originalEvent.touches.length > 1) {
         this.touchDeltaX = 0
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -16,7 +16,7 @@
 }
 
 .carousel.pointer-event {
-  touch-action: none;
+  touch-action: pan-y;
 }
 
 .carousel-inner {


MartijnCuppens:
Can you remove the generated css/js?

patrickhlauke:
oh heaven's sake, i thought i had ... yup

patrickhlauke:
the dangers of getting confused while testing locally. removed all the dist rubbish and force-pushed

js/src/carousel.js
```diff
@@ -283,14 +283,11 @@ class Carousel {
       if (this._pointerEvent && (event.originalEvent.pointerType === PointerType.TOUCH || event.originalEvent.pointerType === PointerType.PEN)) {
         this.touchStartX = event.originalEvent.clientX
       } else if (!this._pointerEvent) {
-        event.preventDefault()
         this.touchStartX = event.originalEvent.touches[0].clientX
       }
     }
 
     const move = (event) => {
-      event.preventDefault()
```
Johann-S: I don't understand why do you remove this line ? 🤔
This function `move` is used only for device which do not handle PointerEvents and prevent the "go back" behaviour

patrickhlauke: `preventDefault` on `touchstart` and `touchmove` prevents scrolling (one of the browser's "default actions") in touch-events-only browsers (e.g. iOS Safari). leaving this in would still result in carousels not allowing scrolling on iPhone etc.

MartijnCuppens (approved):
Tested on several devices, everything seem to work fine. Couldn't test this on a touch desktop device with firefox (this isn't available on BrowserStack either I think).

patrickhlauke:
i tested on my Surface Pro (2017) Win 10/Firefox 63 - this is the only browser that currently fails to react to the swipe, due to the Surface's high dpi/Firefox's overeager `pointercancel`. Planning to file a bug against FX sometime this weekend, after having mentioned it already at last week's TPAC to @smaug----

patrickhlauke:
can't seem to merge this until a code owner reviews...

XhmikosR:
Then wait for it :P

Johann-S (approved):
LGTM 👍 sorry for the delay

patrickhlauke:
Closing the loop here: Firefox bug https://bugzilla.mozilla.org/show_bug.cgi?id=1502010

PR #27596, opened 2018-11-04
Add Noto Sans to the font stack

### LINKED ISSUE #27595: Noto Sans missing in font-family
# Problem
The currently specified font-family in dist/css/bootstrap-reboot.css, scss/_variables.scss and dist/css/bootstrap.css results in unreadable pages on some Linux systems. All normal font is exchanged with the according Emojies.

# What could be triaged so far
As far as I could see, the problem occurs on Linux systems that have ``Noto Sans`` as standard font and neither ``Arial``, ``Roboto``, ``Segoe UI`` or ``Helvetica Neue`` installed. 

Thus the browser will use the ``Noto Color Emoji`` font (as specified by bootstrap), but has no fallback font for normal letters. The fallback ``Noto Sans`` is not specified in the font-family. This leads to unicode characters that are not found by the browser and then replaced by the next best fitting unicode char (the according emoji).

# How the bug can be fixed
Add ``Noto Sans`` to the font family in the files dist/css/bootstrap-reboot.css, scss/_variables.scss and dist/css/bootstrap.css in front of ``sans-serif``.

The new font-family would look as follows:
```-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica Neue,Arial,Noto Sans,sans-serif,Apple Color Emoji,Segoe UI,Segoe UI Emoji,Segoe UI Symbol,Noto Color Emoji```

# Additional data
- Operating system and version: Arch Linux
- Browser and version: Chrome 70
- Reduced test case: Go to any page with bootstrap that does not override font-family ([https://artemis.ase.in.tum.de/#/](https://artemis.ase.in.tum.de/#/) for example)

# Pull request
I'll shortly provide a pull request with the necessary changes to the three files listed above.

LittleHuba:
Fixes #27595

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -263,7 +263,7 @@ $transition-collapse:         height .35s ease !default;
 // Font, line-height, and color for body text, headings, and more.
 
 // stylelint-disable value-keyword-case
-$font-family-sans-serif:      -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;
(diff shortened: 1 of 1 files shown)


MartijnCuppens (approved):
LGTM 👍

PR #27605, opened 2018-11-05
Add an examples layout.

XhmikosR:
I know in the past we wanted people to get the files from the repo and use them, but this situation results in too much duplication and also cases where things can break.

This should make everything safer and cleaner.

Also, I removed unused JS assets and only include the ones each example needs. We might be able to drop more JS files if they are not used in some examples like in the Cover example where no JS is needed.

**[Check the non-whitespace diff](https://github.com/twbs/bootstrap/pull/27605/files?utf8=%E2%9C%93&diff=unified&w=1)**

TODO:

- [x] Load our assets with the Jekyll variables and not relative to the files
- [ ] See if we can make indentation proper

### DIFF
--- a/site/_includes/analytics.html
+++ b/site/_includes/analytics.html
@@ -1,0 +1,6 @@
+<script>
+  window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
+  ga('create', 'UA-146052-10', 'getbootstrap.com');
+  ga('send', 'pageview');
+</script>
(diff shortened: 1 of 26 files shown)


Johann-S (approved):

mdo (approved):
Didn't review line-by-line, but I'm in favor! Curious at what point we offer a separate download or built HTML files in our dist directory, but hopefully that can wait :).

XhmikosR:
I'm gonna finish this soon-ish, confirm all is good and then merge :)

PR #27610, opened 2018-11-05
Travis: Add back chrome addon

XhmikosR:
Without this we weren't testing the latest stable Chromium.

BTW @Johann-S, do you know an easy way to print the version of the browser we use in karma instead of just its name?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 sudo: required
+addons:
+  chrome: stable
 language: node_js
 git:
   depth: 3


Johann-S:
I don't know 😟 we have to find the location of Chrome and run `chrome --product-version`

Johann-S (approved):

XhmikosR:
karma detect browsers already finds the executable, maybe we can do something with `postDetection`?

PR #27632, opened 2018-11-08
Replace holder.js with inline SVG data URI

### LINKED ISSUE #27395: example.html doesn't replace the whole holder.js line

m5o:
Hey 👋 

After starting discussions in #27395 and #27491 this is a POC to test the possibilities to replace [holder.js](http://holderjs.com) library with a pure SVG solution which doesn't require a 3rd party library.

I've tried to create a minimal XML markup, with a background cover and centered text. After studying [Optimizing SVGs in data URIs](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) the output is the `plalceholder.svg` jekyll include in this PR. What's your opinion?

#### Proposal optimized minimum XML markup
```xml
<svg xmlns='http://www.w3.org/2000/svg' width='200' height='150' viewBox='0 0 200 150'>
  <rect fill='#777' width='100%' height='100%' />
  <text fill='#ddd' x='50%' y='50%' font-family='Helvetica,monospace' font-size='13' dominant-baseline='central' text-anchor='middle'>Placeholder Text</text>
</svg>
```
💡 The `SVG` markup was improved after inspiration and tipps from @MartijnCuppens 

<details>
  <summary>previous XML markup generated by `holder.js`</summary>

```xml
<svg width="640" height="250" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 640 250" preserveAspectRatio="none">
  <defs>
    <style type="text/css">#holder_1670738e3a9 text { fill:rgba(255,255,255,.75);font-weight:normal;font-family:Helvetica, monospace;font-size:32pt } </style>
  </defs>
  <g id="holder_1670738e3a9">
    <rect width="640" height="250" fill="#777"></rect>
    <g>
      <text x="238.15625" y="139.85">640x250</text>
    </g>
  </g>
</svg>
```

</details>

#### ⌨️ Tasks

* use jekyll includes to create inline SVG with data URI
* replace `data-src="holder.js/"` with jekyll `holder.html` include
* remove holder.js workaround in example.html

#### 📖 References 
* https://codepen.io/tigt/post/optimizing-svgs-in-data-uris
* https://codepen.io/jakob-e/pen/doMoML
* https://css-tricks.com/probably-dont-base64-svg/

#### 🔗 Discussions
* https://github.com/twbs/bootstrap/pull/27491 – `Revert "Remove custom example plugin."`
* https://github.com/twbs/bootstrap/issues/27395 – `example.html doesn't replace the whole holder.js line`

----------

Fixes #27395, closes #27491


XhmikosR:
@MartijnCuppens has already started working on it, but I haven't had the chance to review his branch yet since it's not final.

XhmikosR:
I'm gonna get back to this later. Still though the issue remains, the example does include the img and we don't want this.

XhmikosR:
So, so far here's what I see

1. I can't seem to find a Liquid filter we can use instead of us escaping characters manually plus reinventing the wheel sort of
2. The example code still shows the src. The idea is to not show the src at all. Maybe we could try some Liquid magic with split to get the src part, and then truncate that so that `...` is shown there? Just thinking out loud, not sure if it's doable

site/_includes/holder.html
```diff
@@ -0,0 +1,41 @@
...
+{%- assign text_color = include.fg | default: "#ddd" -%}
+{%- assign text_size = include.text_size | default: "13" -%}
+
+{%- capture svg -%}
+<svg xmlns='http://www.w3.org/2000/svg' width='{{ width }}px' height='{{ height }}px' viewBox='0 0 {{ width }} {{ height }}'>
+  <path fill='{{ background }}' d='M0 0h{{ width }}v{{ height }}H0z'/>
+  <text fill='{{ text_color }}' font-family='sans-serif' font-size='{{ text_size }}' x='50%' y='50%' alignment-baseline='middle' text-anchor='middle'>{{ text }}</text>
+</svg>
+{%- endcapture -%}
+{%- assign svg_encoded = svg
```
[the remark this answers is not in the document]
m5o: Thx, for update the naming. Naming can be always improved. I've intentionally didn't escape spaces. See the findings of [this](https://css-tricks.com/probably-dont-base64-svg/) and [this](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris) to keep the characters as low as possible.

Liquid has some filters, like [`escape`](
https://shopify.github.io/liquid/filters/escape/), [`url_encode`](https://shopify.github.io/liquid/filters/url_encode/) but they don't fit perfectly because I wanted to just replace the necessary chars to keep the svg_normalize string at a minimum.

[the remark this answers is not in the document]
XhmikosR: The space escape is fixed with my last patch.

site/_includes/holder.html
```diff
@@ -0,0 +1,43 @@
...
+    {{ include.text }}
+  {%- endif -%}
+{%- endcapture -%}
+
+{%- assign background = include.bg | default: "#777" -%}
+{%- assign text_color = include.fg | default: "#ddd" -%}
+{%- assign text_size = include.text_size | default: "13" -%}
+
+{%- capture svg -%}
+<svg xmlns='http://www.w3.org/2000/svg' width='{{ width }}px' height='{{ height }}px' viewBox='0 0 {{ width }} {{ height }}'>
```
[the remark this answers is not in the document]
MartijnCuppens: Also, the SVG doesn't work in IE, can you check what's wrong there?

MartijnCuppens (dismissed):

MartijnCuppens:
@XhmikosR, this PR keeps the ``, that makes more sense, I've closed my PR.

XhmikosR:
Note that apart from checking everything in all browsers including IE, examples need to be adapted too.

So, we need #27605 merged first and then this one adapted. So, stay tuned until I'm done with that.

MartijnCuppens:
Go @XhmikosR go! 

I checked the latest change and this version does works in IE! 🎉

site/docs/4.1/content/images.md
```diff
@@ -11,7 +11,7 @@ toc: true
 Images in Bootstrap are made responsive with `.img-fluid`. `max-width: 100%;` and `height: auto;` are applied to the image so that it scales with the parent element.
 
 <div class="bd-example">
-  <img data-src="holder.js/100px250" class="img-fluid" alt="Generic responsive image">
+  <img src="{% include holder.svg size="640x250" %}" class="img-fluid" alt="Generic responsive image">
```
XhmikosR: Can't we still keep the 100% functionality? Because it's quite handy not needing to specify the size in pixels.

MartijnCuppens: Wouldn't do that, in most cases images with dimensions are used. This makes it clear that the image is rescaled instead of being cropped.

XhmikosR:
I don't agree. 100% is quite handy and remember this is not supposed to be
printed in the highlight code so it's only for our internal use.

site/docs/4.1/content/images.md
m5o: It seem's to work already with `{% include holder.svg size="100%x180" %}` I could change this, but I'm with Martijn, in the real world cases images with dimensions are used. What is the decision here?

site/_includes/holder.svg
```diff
@@ -11,13 +11,11 @@ optional: text_size - "13" (default)
...
 {%- assign height = size[1] | default: "400" -%}
 
-{%- capture text -%}
-  {%- if include.text == null -%}
-    {{ width }}×{{ height }}
-  {%- else -%}
-    {{ include.text }}
-  {%- endif -%}
-{%- endcapture -%}
+{%- if include.text == null -%}
```
XhmikosR: `null` or `nil`?

XhmikosR: Also, you could probably use `default` here

site/docs/4.1/content/images.md
MartijnCuppens: @XhmikosR, see my previous PR (example with `.mw-100`): https://deploy-preview-27633--twbs-bootstrap4.netlify.com/docs/4.1/utilities/sizing/

That's the result you 'll have when we allow `100%`, the image gets "cropped" from both sides if the viewport is resized and that is not the result we'll have with `jpeg, png,..` because we have a ratio with rasterised images.

site/_includes/example.html
```diff
@@ -18,6 +18,21 @@
 
 {%- if include.hide_markup == null -%}
   {%- highlight html -%}
-    {{- include.content -}}
+    {%- if include.content contains 'src="data:image/svg+xml' -%}
+      {%- assign modified_content = include.content
+        | replace: 'src="data:image/svg', 'src="✂️data:image/svg'
```
MartijnCuppens: Wow, genius solution!

site/_includes/holder.svg
m5o: I've kept the condition consistent with the ones from [`example.html`](https://github.com/twbs/bootstrap/blob/v4-dev/site/_includes/example.html)

XhmikosR: I'll have a look myself later because `default` is cleaner IMO.

XhmikosR: Actually we can't do that here, so your solution is fine 👍

XhmikosR:
@m5o: squashed everything, so only thing left is for this to be adapted for the examples too.

Then one final review and we are good to go.

XhmikosR:
Maybe we should rename it to `placeholder` while at it?

m5o:
> Maybe we should rename it to `placeholder` while at it?

I don't care, `placeholder` is more descriptive but also more to type on each include 😉

XhmikosR:
Let's go with placeholder. It's always better to be more descriptive.

XhmikosR:
Can you also try using the system font stack (minus the Emoji fonts)? Not sure if it'll work at all, but worth the try.

site/_includes/holder.svg
```diff
@@ -0,0 +1,41 @@
...
+{%- endif -%}
+
+{%- assign background = include.bg | default: "#777" -%}
+{%- assign text_color = include.fg | default: "#ddd" -%}
+{%- assign text_size = include.text_size | default: "13" -%}
+
+{%- capture svg -%}
+<svg xmlns='http://www.w3.org/2000/svg' width='{{ width }}' height='{{ height }}' viewBox='0 0 {{ width }} {{ height }}'>
+  <rect fill='{{ background }}' width='100%' height='100%' />
+  <text fill='{{ text_color }}' x='50%' y='50%' dy='5' font-family='sans-serif' font-size='{{ text_size }}' text-anchor='middle'>{{ text }}</text>
```
MartijnCuppens: I think you'll need to calculate your `dy` based on the font size. For example the carousel text isn't aligned in the middle.

MartijnCuppens (dismissed):

site/_includes/holder.svg
m5o: I'll convert it back to `alignment-baseline='middle'`

XhmikosR:
One issue I'm seeing is that the images are not exactly the same wdith/height as before. We need to change the include to an html img tag and accept class, alt etc so that we set the width/height there.

XhmikosR:
Example:

https://twbs-bootstrap4.netlify.com/docs/4.1/examples/blog/
https://deploy-preview-27632--twbs-bootstrap4.netlify.com/docs/4.1/examples/blog/

m5o:
> Can you also try using the system font stack (minus the Emoji fonts)? Not sure if it'll work at all, but worth the try.

I've tried if css vars (`var(--font-family-sans-serif)`) are possible inside the SVG, but it's not working. Alternative is add the static font-family string, but it's long…

XhmikosR:
Let's try with `font-family:-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,Noto Sans,sans-serif` and see. I too find it long. FYI, Holder.js was set up to use `Helvetica` in main docs.

m5o:
> Let's try with `font-family:-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,Noto Sans,sans-serif` and see. I too find it long. FYI, Holder.js was set up to use `Helvetica` in main docs.

Problem are font-families wrapped in quotes. 

🚫 `font-family='-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,Noto Sans,sans-serif'`
🆗 `font-family='-apple-system,BlinkMacSystemFont,Roboto,Arial,Noto Sans,sans-serif'`

XhmikosR:
Obviously you adapt it.

m5o:
> One issue I'm seeing is that the images are not exactly the same wdith/height as before.

For the blog example: the border style of the `.card` wrapper affect the img width/height dimensions. It's a side effect of `box-sizing: border-box` vs. `box-sizing: content-box` 🤔

XhmikosR:
Not only for blog, it happens in other places too. If you check the cards images, it's not the height we pass.

The best solution would be for us to set the width and height attributes, I believe this would fix it.

site/_includes/example.html
```diff
@@ -18,6 +18,21 @@
 
 {%- if include.hide_markup == null -%}
   {%- highlight html -%}
-    {{- include.content | replace: 'data-src="holder.js', 'src="...' -}}
+    {%- if include.content contains 'src="data:image/svg+xml' -%}
+      {%- assign modified_content = include.content
+        | replace: 'src="data:image/svg', 'src="✂️data:image/svg'
```
XhmikosR: I'm gonna backport this to a new PR to fix the holder.js replacement in v4-dev, because I'm not sure when this PR will be ready.

m5o:
against your thesis speaks that a fixed width and height didn't affect the automatic shrinking. I've tested this on `blog/index.html` without success.

```html
<img class="card-img-right flex-auto d-none d-lg-block" width="200" height="250" src="{% include placeholder.svg size="200x250" bg="#55595c" fg="#eceeef" text="Thumbnail" text_size="20" %}" alt="Card image cap">
```

XhmikosR:
The cause for that is that you are limiting the rect with specific sizes. Drop those and use percentages or something.

XhmikosR:
OK, I pushed one PoC patch which takes care of the 100% images. If we all agree with this we should adapt the patch to make use of percentages again.

XhmikosR:
**Note that the current implementation does not work on IE.**

m5o:
Thank you @XhmikosR 💪

XhmikosR:
So, here's what's left IMO:

1. the 100% works fine but the text is being stretched
2. there are still some cases the text is not vertically aligned properly
3. the image does not have the exact dimensions we pass. I think this can be solved if we switched to an `img` tag and passed there the width and height attributes. I will experiment with this later or tomorrow

PR #27636, opened 2018-11-09
Clarify length unit for SWIPE_THRESHOLD

xinbenlv:
Add clarification for unit of `SWIPE_THRESHOLD`

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -23,7 +23,7 @@ const JQUERY_NO_CONFLICT     = $.fn[NAME]
 const ARROW_LEFT_KEYCODE     = 37 // KeyboardEvent.which value for left arrow key
 const ARROW_RIGHT_KEYCODE    = 39 // KeyboardEvent.which value for right arrow key
 const TOUCHEVENT_COMPAT_WAIT = 500 // Time for mouse compat events to fire after touch
-const SWIPE_THRESHOLD        = 40
+const SWIPE_THRESHOLD        = 40 // Minimum swipe displacement in unit of DOM coordinate, usually in "px"
 
 const Default = {
   interval : 5000,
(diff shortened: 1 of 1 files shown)


js/src/carousel.js
```diff
@@ -23,7 +23,7 @@ const JQUERY_NO_CONFLICT     = $.fn[NAME]
 const ARROW_LEFT_KEYCODE     = 37 // KeyboardEvent.which value for left arrow key
 const ARROW_RIGHT_KEYCODE    = 39 // KeyboardEvent.which value for right arrow key
 const TOUCHEVENT_COMPAT_WAIT = 500 // Time for mouse compat events to fire after touch
-const SWIPE_THRESHOLD        = 40
+const SWIPE_THRESHOLD        = 40 // Minimum swipe displacement in unit of DOM coordinate, usually in "px"
```
XhmikosR: I'm not sure this is better, personally.

xinbenlv: Would you mind suggest how to improve it?

XhmikosR: I'm not a native English speaker either, so bear with me :P

But maybe, just `Minimum swipe distance`. Although I'm still not sure if it adds any value then.

xinbenlv: Well, the reason I added the doc is when I was looking into the code and was not sure what unit the number is, really I was trying to simply say that this number is in the unit of pixel rather than anything else. Do you think it is clearer by just shorten it to `In the unit of \`px\``?

xinbenlv (commented):
Updated

XhmikosR:
Yeah, not sure it's needed either. https://github.com/twbs/bootstrap/pull/27636#discussion_r232159245

Johann-S:
I think we'll pass for that but thank you @xinbenlv 👍

xinbenlv:
I see. So the reason I add this pull request was due to previously I didn't know what unit was used for the threshold. As Bootstrap 4 is moving to `rem` from `px`, it was not entirely impossible people assume it would be `rem` in many places. I was wondering, it really doesn't need a unit clarification?

Johann-S:
@xinbenlv that's because we use JavaScript properties which return coordinates in pixels.

See: https://www.w3schools.com/jsref/event_clientx.asp

PR #27646, opened 2018-11-10
Make focus cycle by `Shift+Tab` key on modals.

### LINKED ISSUE #21661: fix(modal): shift+tab without focused element within modal
Issue can be validated here: 
1) Navigate to http://getbootstrap.com/javascript/ 
2) Search for "Launch demo modal" and click on that. Now use shift+tab, nothing happens.

There was a similar bug in Angular-ui's bootstrap which got fixed: https://github.com/angular-ui/bootstrap/pull/5294

This should be done even in this library too.

I have tried on:
Operating system and version Windows, 
Browser and version Chrome: 55.0.2883.87

### LINKED ISSUE #27241: Keyboard Accessibility in Modals - SHIFT + TAB should cycle backwards
- Operating system and version : All versions
- Browser and version : All versions
- Test case: Launch demo modal and Press Shift+Tab
https://getbootstrap.com/docs/4.0/components/modal/

- Expected Result: When focus leaves first interactive element, it should go to the last interactive element.
- Actual Result: Focus stops on the first interactive element and no further action can be taken with SHIFT +TAB

boiledfroginthewell:
Fixes #21661 and fixes #27241


js/src/modal.js
```diff
@@ -63,7 +65,13 @@ const Selector = {
   DATA_TOGGLE        : '[data-toggle="modal"]',
   DATA_DISMISS       : '[data-dismiss="modal"]',
   FIXED_CONTENT      : '.fixed-top, .fixed-bottom, .is-fixed, .sticky-top',
-  STICKY_CONTENT     : '.sticky-top'
+  STICKY_CONTENT     : '.sticky-top',
+  TABBABLE           : 'a[href]:not([tabxndex^="-"]), area[href]:not([tabxndex^="-"]), ' +
```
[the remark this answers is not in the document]
XhmikosR: For example why should we even care about `tabndex`?

js/src/modal.js
```diff
@@ -184,7 +180,7 @@ class Modal {
     this._setEscapeEvent()
     this._setResizeEvent()
 
-    $(document).off(Event.FOCUSIN, Event.KEYDOWN)
+    $(document).off(`${Event.FOCUSIN} ${Event.KEYDOWN}`)
```
boiledfroginthewell: Thank you for your fix, but I found that all the events must be specified in the first argument.

And I also added the code in dispose() method.

js/src/modal.js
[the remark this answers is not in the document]
XhmikosR: I was talking about `tabndex` not `tabindex`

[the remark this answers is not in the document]
boiledfroginthewell: Oh, sorry. They are typos. `tabindex` is correct. The new commit fixes them too.

XhmikosR:
Seems to work well. @Johann-S, @patrickhlauke: thoughts?

js/src/modal.js
```diff
@@ -290,6 +294,31 @@ class Modal {
           this._element.focus()
         }
       })
+      .on(Event.KEYDOWN, (event) => {
+        if (event.isDefaultPrevented()) {
+          return
+        }
+        if (event.shiftKey && event.which === TAB_KEYCODE) {
+          const currentFocus = $(document.activeElement)
+          const tabbables = $(this._element).find(Selector.FOCUSABLE)
```
Johann-S: Please use vanilla JS to query your elements, it's a pretty big selector here, so it'll be faster

js/src/modal.js
```diff
@@ -290,6 +294,31 @@ class Modal {
           this._element.focus()
         }
       })
+      .on(Event.KEYDOWN, (event) => {
+        if (event.isDefaultPrevented()) {
+          return
+        }
+        if (event.shiftKey && event.which === TAB_KEYCODE) {
+          const currentFocus = $(document.activeElement)
```
Johann-S: No need to store it in a jQuery element.

```js
const currentFocus = document.activeElement
```

Johann-S (changes requested):

js/src/modal.js
```diff
@@ -63,7 +65,9 @@ const Selector = {
   DATA_TOGGLE    : '[data-toggle="modal"]',
   DATA_DISMISS   : '[data-dismiss="modal"]',
   FIXED_CONTENT  : '.fixed-top, .fixed-bottom, .is-fixed, .sticky-top',
-  STICKY_CONTENT : '.sticky-top'
+  STICKY_CONTENT : '.sticky-top',
+  FOCUSABLE      : 'a[href], area[href], input, button, select, textarea, ' +
+    'audio[controls], video[controls], iframe, object, embed, [tabindex], [contenteditable]'
```
patrickhlauke: would be good, if possible, to exclude things with `tabindex` lower than `0` (e.g. `tabindex="-1"`)

patrickhlauke (changes requested):
Apart from removing things with negative `tabindex`, this looks good otherwise

js/src/modal.js
tmorehouse: And should probably filter the elements to make sure they are visible (i.e. not `display: none`, etc)

tmorehouse: JQuery has a visible selector, and in vanillaJs you can just check that the element's `boundingClientRect` height and width are greater than `0`

XhmikosR:
Can you take care of the suggestions @boiledfroginthewell ?

js/src/modal.js
boiledfroginthewell: Fixed.

I made a simple benchmark: https://jsperf.com/finding-tabbable-elements/24
The fixed code runs twice faster!

js/src/modal.js
boiledfroginthewell: Fixed.

js/src/modal.js
boiledfroginthewell: Fixed.

Simply replacing the `:visible` selecter with `boundingClientRect` achieved poor performance with Chrome (https://jsperf.com/finding-tabbable-elements/24). So, `offsetWidth` is also added as jQuery does.

boiledfroginthewell:
Addressed all comments. Sorry for the late response.

mdo:
Closing for #28481 and punting to v5 given the jQuery drop. Hoping @Johann-S can help us with a pure-JS implementation of this if the feature is still warranted.

PR #25838, opened 2018-03-13
generate TOC without custom jekyll plugin

m5o:
* use [allejo/jekyll-toc](https://github.com/allejo/jekyll-toc) for toc generation
* remove custom jekyll plugin `toshimaru/jekyll-toc`
* use h5 and h6 headlines inside callouts
  * h5 and h6 by default are excluded by toc generation

I've checked generation of every site, it works like a charm 👌 

This is the final 🏁 PR wich removes custom plugins from 
bootstrap docs and enable first-class support for [github-pages](https://pages.github.com) 🙌

🔗 https://github.com/twbs/bootstrap/pull/25784
🔗 https://github.com/twbs/bootstrap/pull/25319
🔗 https://github.com/twbs/bootstrap/pull/25320

Hope you like it. 🖖

/cc @allejo

PR #27648, opened 2018-11-11
Use an `h5` for callouts

XhmikosR:
This is so that they are not included in ToC.


XhmikosR:
Now that I think about it... It doesn't make sense the header levels for some of the includes

XhmikosR:
The thing is we can't know the level of the previous sibling item and thus use the same level in the callouts.

XhmikosR:
So, perhaps we should just hide the callouts' headers consistently?

patrickhlauke:
nobody dies if a heading level is skipped. unless we have a way to embed callouts at any level, and the callout auto-adapts the level to be subordinate of the last heading level it's at, it's safest to keep the heading level to something guaranteed to be subordinate in every situation...so let's leave it to `h5`, or heck could even go to `h6`. whether it's included in the TOC or not is up for debate.

XhmikosR:
The reason for this PR wasn't the skipped level. Was to show the callout in the ToC. But after thinking about it, I believe it's better if we just hide all callouts instead from ToC.

patrickhlauke (commented):
if we can guarantee that the callout is only ever included with a preceding `h3`, and not an `h4` or `h5`, then sure. but i think it's not a major problem even as is.

patrickhlauke:
ah gotcha,hadn't seen your comment until now. yeah, i think it's fine even if the callouts AREN'T in the TOC

XhmikosR:
All right, I'm gonna hide them by using an `h5`.

XhmikosR:
Actually, it might be a bug in jekyll-toc. Waiting for confirmation/fix https://github.com/toshimaru/jekyll-toc/issues/70

m5o:
@XhmikosR Consider review/cherry-pick https://github.com/twbs/bootstrap/pull/25838/commits/08b096d18a926efea484a3d469bbebaa6191278a from #25838 this should cover the whole list.

XhmikosR:
Thanks but I want this fixed upstream.

mdo:
FWIW, I don’t want callouts to be listed in the table of contents. Any of them that are there currently are there unintentionally.

XhmikosR:
Yeah, I think I will just change them so that they don't show up for now. No idea when this will be fixed upstream.

MartijnCuppens (approved):

PR #27651, opened 2018-11-11
Customize browse text of the custom file input with HTML

MartijnCuppens:
This PR makes it possible to customize the text of the browse button in HTML. This comes in handy if a developer has no control over the Bootstrap css.

Demo: https://deploy-preview-27651--twbs-bootstrap4.netlify.com/docs/4.1/components/forms/#translating-or-customizing-the-strings-with-html


XhmikosR (approved):

sts-ryan-holton:
@MartijnCuppens  I noticed when uploading images in your file browser it creates a path to your image which is great, but for me on Windows here it's saying: "fakepath"? How do we change that to your real path of an image / disable it. Is that what the plugin is for as mentioned in your preview?

XhmikosR:
That's unrelated to this.

PR #27655, opened 2018-11-11
refactor spinners SCSS

midzer:
Hi all,

latest SCSS code from https://github.com/twbs/bootstrap/pull/22960 did not work for me on spinner `-sm` classes.

So I went along and refactored a little bit:
* introduce base class `%spinner` which `spinner-border` and `spinner-grow` both extend
* introduce `spinner-sm` modifier class which is used in combination with `spinner-border` and `spinner-grow` to make small spinners, this is the way bootstrap treats other classes as well
* rename both `@keyframes` to save more byte

Any feedback is heavily welcome.
Have a nice weekend (or start in the week)
midzer

### DIFF
--- a/scss/_spinners.scss
+++ b/scss/_spinners.scss
@@ -1,12 +1,8 @@
 //
-// Rotating border
+// Spinners
 //
 
-@keyframes spinner-border {
-  to { transform: rotate(360deg); }
-}
-
-.spinner-border {
+%spinner {
   position: relative;
   display: inline-block;
   width: $spinner-width;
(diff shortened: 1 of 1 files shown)


scss/_spinners.scss
```diff
@@ -1,36 +1,45 @@
...
   border-radius: 50%;
-  animation: spinner-border .75s linear infinite;
-}
+  animation: rotate .75s linear infinite;
 
-.spinner-border-sm {
-  width: $spinner-width-sm;
-  height: $spinner-height-sm;
-  border-width: $spinner-border-width-sm;
+  &.spinner-sm {
```
MartijnCuppens: I would not use the `&` here, we don't do that for other modifiers, eg `.btn-lg`: https://github.com/twbs/bootstrap/blob/d96c147d6715b4772608f5d4b3ddd9cb0c7af7a3/scss/_buttons.scss#L110

scss/_spinners.scss
```diff
@@ -1,36 +1,45 @@
...
-// Rotating border
+// Spinners
 //
 
-@keyframes spinner-border {
-  to { transform: rotate(360deg); }
-}
-
-.spinner-border {
+%spinner {
```
MartijnCuppens: Watch out wit extends, you idea was probably to generate less code, but with this technique, you've added 2 extra selectors. I would prefer to see a mixin instead, but don't know if @mdo also agrees?

scss/_spinners.scss
MartijnCuppens: Also, the class is not correct

MartijnCuppens (changes requested):
You can't change the modifier classes like that without changing the

mdo:
I left a reply to your comment in #22960 about this—we won't be changing to an extend based approach as it adds more selectors. Plus, you've changed the selectors here, which are fine as-is with the base class and modifier classes approach. Thanks though :).

PR #27660, opened 2018-11-13
Add white-space reset to accordion header button

moso:
This will fix the overflow that occurs if the content of `.card-header` is too long, because we're using `white-space: nowrap` on normal a `.btn`. This is fixed by resetting `white-space` to `normal`, which allows for multiple lines in a button.

Example of overflow from docs:

image

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -305,6 +305,10 @@
 
     .card-header {
       margin-bottom: -$card-border-width;
+
+      .btn {
+        white-space: normal;
+      }
     }
   }
 }


MartijnCuppens:
Hi @moso,

`white-space: nowrap;` already is removed from `.btn`, this will be fixed in the next release.


See:
https://github.com/twbs/bootstrap/commit/24bdadf9d3246b8b7668bd442dccf1f2aa5fae37#diff-8a4fc4262c9197c73b80fb80d8863ae5

PR #27665, opened 2018-11-14
Responsive images with .embed-responsive

codecalm:
Hi!
I've added `object-fit: cover` to `.embed-responsive-item`. It's useful, when you want to responsive images with exact ratio. Without this fix images are crushed. 

Demo:
http://jsfiddle.net/jhybufos/

Preview:

screenshot 2018-11-14 at 10 16 31

Code:
```
<div class="embed-responsive embed-responsive-1by1">
   <img src="https://placehold.it/1280x600" alt="" class="embed-responsive-item">
</div>
```

### DIFF
--- a/scss/utilities/_embed.scss
+++ b/scss/utilities/_embed.scss
@@ -24,6 +24,7 @@
     width: 100%;
     height: 100%;
     border: 0;
+    object-fit: cover;
   }
 }
 


MartijnCuppens:
Hi @codecalm,
This is a nice feature, but `object-fit: cover` is not supported by IE.

Also see https://github.com/twbs/bootstrap/pull/27082

PR #27668, opened 2018-11-14
enforce lowercase class name in .stylelintrc

coliff:
`"selector-class-pattern": "^[a-z]+([a-z0-9]?|[a-z0-9\\-]*[a-z0-9])$"`

I think this could be a useful addition, to prevent the use of a class name against the projects guidelines - mixed case or using underscores etc.
REF: http://codeguide.co/#css-classes


MartijnCuppens:
I think `^[a-z][a-z0-9\\-]*[a-z0-9]$` should do the job?

MartijnCuppens:
@coliff, can you add an exception to `_syntax.scss`?

XhmikosR:
I believe this regex needs more work, so don't merge this yet.

XhmikosR:
OK, it seems the regex was fine after all: https://regex101.com/r/m7ynnv/3

XhmikosR (approved):

PR #27672, opened 2018-11-15
docs: image updates

XhmikosR:
TODO:

- [x] See if we can find the Stack image as an SVG
- [ ] Try a JPEG with white background for the themes image


site/index.html
```diff
@@ -80,7 +82,9 @@ <h3>Official Themes</h3>
     <p>
       Take Bootstrap 4 to the next level with official premium themes—toolkits built on Bootstrap with new components and plugins, docs, and build tools.
     </p>
-    <img class="img-fluid mt-3 mx-auto" src="{{ site.baseurl }}/docs/{{ site.docs_version }}/assets/img/bootstrap-themes.png" alt="Bootstrap Themes" width="1024" height="388">
```
MartijnCuppens: I think we can save some bandwidth if we use a `.jpg` here, the file size of complex images tend to be higher with `.png`

site/index.html
```diff
@@ -6,7 +6,9 @@
   <div class="container">
     <div class="row align-items-center">
       <div class="col-6 mx-auto col-md-6 order-md-2">
-        <img class="img-fluid mb-3 mb-md-0" src="{{ site.baseurl }}/docs/{{ site.docs_version }}/assets/img/bootstrap-stack.png" alt="" width="1024" height="860">
```
MartijnCuppens: Do we have the original image? Maybe an svg is smaller here?

site/index.html
XhmikosR: Last time I tried, we didn't save a lot with JPG 80%.

site/index.html
XhmikosR: I do not have the original image, otherwise it'd definitely be smaller.

XhmikosR: @mdo: do you happen to have this SVG?

XhmikosR:
I'll give jpeg a go again later. I still think we should have the stack image as SVG though.

Johann-S (approved):

site/index.html
XhmikosR: @MartijnCuppens: I'm gonna merge this as is for now, but feel free to experiment with JPG for the themes image and example images later.

PR #27673, opened 2018-11-15
Prevent text decoration skip ink and reorder comments

MartijnCuppens:
The underlining of `<abbr>` tags was a bit strange in chrome with some capital letters (first line):

This PR prevents the text decoration skip ink (second line) and reorders some comments.


XhmikosR:
LGTM but being it's a change in reboot, I'd like @mdo's confirmation before merging.

mdo (approved):
:shipit:

PR #27678, opened 2018-11-16
Add documentation about .font-weight-bolder/lighter

ysds:
`.font-weight-bolder` and `.font-weight-lighter` have been added by #26580,  but no documentation.

### DIFF
--- a/site/docs/4.1/utilities/text.md
+++ b/site/docs/4.1/utilities/text.md
@@ -85,8 +85,10 @@ Quickly change the weight (boldness) of text or italicize text.
 
 {% capture example %}
 <p class="font-weight-bold">Bold text.</p>
+<p class="font-weight-bolder">Relative bolder than the parent element.</p>
 <p class="font-weight-normal">Normal weight text.</p>
 <p class="font-weight-light">Light weight text.</p>
+<p class="font-weight-lighter">Relative lighter than the parent element.</p>
 <p class="font-italic">Italic text.</p>
 {% endcapture %}
 {% include example.html content=example %}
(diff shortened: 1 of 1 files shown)


site/docs/4.1/utilities/text.md
```diff
@@ -85,8 +85,10 @@ Quickly change the weight (boldness) of text or italicize text.
 
 {% capture example %}
 <p class="font-weight-bold">Bold text.</p>
+<p class="font-weight-bolder">Relative bolder than the parent element.</p>
```
XhmikosR: You mean **compared to** the parent element?

site/docs/4.1/utilities/text.md
```diff
@@ -85,8 +85,10 @@ Quickly change the weight (boldness) of text or italicize text.
...
 <p class="font-weight-normal">Normal weight text.</p>
 <p class="font-weight-light">Light weight text.</p>
+<p class="font-weight-lighter">Relative lighter than the parent element.</p>
```
XhmikosR: Maybe we should add a note that this depends on the system font and the effect might not be visible on some systems.

For example on Windows with Firefox I get:

ysds: That screenshot is the expected result. That parent is `.bd-example` (the font weight are inherited from `body`)

XhmikosR: Yeah, that might need a note or, not.

XhmikosR:
I'm not a native English speaker, so my edit might not be 100% correct.

XhmikosR:
For reference: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Values

site/docs/4.1/utilities/text.md
MartijnCuppens: Maybe we should just use the same descriptions as in the mozilla docs?
image

XhmikosR: In that case, we should probably link to the MDN doc as a reference.

mdo (approved):

blackc03r (commented):
CMS-1691-P

Issue #27505, opened 2018-10-23
Mismatched height of input-group append/prepend elements in horizontal forms


b-nik:
Hello!

There seems to be a problem with input-groups in horizontal forms. Namely, If the form-group gets bigger (due to the label being long for example) the append/prepend elements follow the form-group size, while the input itself stays the same. This results in the append/prepend elements not matching their input size.

__Reduced Test case__: https://jsbin.com/pidudozuku/edit?html,output
__Suggested Fix:__ 
The simplest solution to me seems to be:
```css
 .input-group {
   height: $input-height;
}
```
Here's the above in jsbin (with input-height hardcoded): https://jsbin.com/moqojozega/edit?html,output

Also adding either input-group-sm/input-group-lg prevents the problem for happening, but of course it changes the size of the input so it's not an ideal solution.

__Operating System__: Ubuntu 16.04
__Tested On__: Firefox, Opera, Chromium (latest versions)

alecpl:
Confirmed. I think setting height will not be an acceptable solution, but this also works:
```
.input-group {
  align-self: flex-start;
}
```

alecpl:
@mdo, I don't think it's related, and `height: auto` does not fix it.

alecpl:
@b-nik, just note that input-group is not only for inputs, it can contain e.g. a textarea or multi-line button. So, the fact that input-group-lg/sm sets the height of the form control inside is another bug, imo. It should use `height: auto` for textareas.

b-nik:
The problem doesn't exist in 4.1.2.

However in 4.1.2 the input grows which I'm not sure about visually:
https://jsbin.com/nahabanosi/edit?html,output

florianlacreuse:
Is it right to have `col-sm-9` and `input-group` at the same level, side by side in the same class attribut?

See [horizonral form doc](https://getbootstrap.com/docs/4.1/components/forms/#horizontal-form): `div` blocks wrapping input, checkbox, etc. have only one css class, `.col*`.

Reduced test case fixed: https://jsbin.com/vuletujofo/1/edit?html,output

b-nik:
Thanks, @florianlacreuse that actually worked :)

florianlacreuse:
Good to hear.

@mdo @b-nik @alecpl Should this issue be closed or do we consider that there is still a bug somehow here?

MartijnCuppens:
[...]

PR #27677, opened 2018-11-16
Add support for custom-select sizing in input-group

ysds:
Currently, the input group size options don't support custom forms.

image

https://codepen.io/anon/pen/Zmyyoe

This PR add support them.

After: https://codepen.io/anon/pen/mQwwNN

PR #27687, opened 2018-11-19
Fixed a fixed height issue of input-group size option

ysds:
If set a `rows` attribute to a textarea in an input group, it has no effect:

image

This issue has occurred from 4.1.3 (#26820).

* 4.1.3: https://codepen.io/anon/pen/qQPprY
* 4.1.2: https://codepen.io/anon/pen/jQGZMa

This PR remove the incorrect settings of height. 

* After: https://codepen.io/anon/pen/eQGyVG

This also relates to #27505.


scss/_input-group.scss
```diff
@@ -127,26 +127,34 @@
 // Remix the default form control sizing classes into new ones for easier
 // manipulation.
 
+.input-group-lg > .form-control:not(textarea),
+.input-group-lg > .custom-select {
```
MartijnCuppens: I don't think this would be needed for `.custom-select`, or is there a specific reason why you also added this here, @ysds?

ysds: See #27677

MartijnCuppens: Hmm, it looks like my memory doesn't last for four days 😄

MartijnCuppens (approved):
Another job well done, @ysds!

PR #27698, opened 2018-11-21
Fix body scrolling issue when modal open

### LINKED ISSUE #27176: Modal scrolls down the page
The occurrence condition is not clear, but when open a modal, the page may scroll down. It appears to be affected by the window size or length of the page. Often it happens when the window size is small, it may occur even when the window is wide.

Windows 10 / macOS (Show scroll bars - always)
modal

https://getbootstrap.com/docs/4.1/components/modal/

ysds:
Fixes #27176

Before: https://getbootstrap.com/docs/4.1/components/modal/#live-demo
After: https://deploy-preview-27698--twbs-bootstrap4.netlify.com/docs/4.1/components/modal/#live-demo


js/src/modal.js
```diff
@@ -127,8 +127,6 @@ class Modal {
 
     this._adjustDialog()
 
-    $(document.body).addClass(ClassName.OPEN)
```
Johann-S: does it works to move this line abode `this._adjustDialog()` ?

ysds: I tried that but didn't work. We need to reduce the number of renderings.

Johann-S:
I can still reproduce it on our visual tests.

Steps:
 - Click on "Toggle tall body content"
 - Reduce your window
 - Open your modal and on close the scroll goes up

ysds:
It works fine on mac (Chrome).

modal

I'll check on Windows tomorrow 😉

ysds:
If you talk about scrolling up to the top edge of the button, that is correct behavior when focus an element.

Johann-S:
To see that behavior you have to scroll a bit your window before opening a modal

ysds:
I can see that, but that has nothing to do with this problem. That behavior is triggered by the following code. That behavior is caused by the following code:

https://github.com/twbs/bootstrap/blob/b7d2c3c4978a36f5d9c0c15db7ac79ab9bd9caf5/js/src/modal.js#L562-L566

And I believe that is correct behavior.

Johann-S (approved):

PR #27708, opened 2018-11-22
[Sizing] Add max-width and max-height utilities

### LINKED ISSUE #27628: Consider adding max-width & max-height variants
Bootstrap has classes for controlling the width and height, e.g: `.w-75` but it would be useful to also have a max width variant of this https://getbootstrap.com/docs/4.1/utilities/sizing/, similar to the max height class currently listed there.

royklutman:
Closes #27628.

This PR would add the max-width and max-height utilities:
```
mw-25 // max-width: 25% !important;
mw-50 // max-width: 50% !important;
mw-75 // max-width: 75% !important;
mw-100 // max-width: 100% !important;

mh-25 // max-height: 25% !important;
mh-50 // max-height: 50% !important;
mh-75 // max-height: 75% !important;
mh-100 // max-height: 100% !important;
```
Notice this doesn't include `m*-auto`, as `auto` not a valid value.

What do you guys think?

### DIFF
--- a/site/docs/4.1/utilities/sizing.md
+++ b/site/docs/4.1/utilities/sizing.md
@@ -30,16 +30,22 @@ Width and height utilities are generated from the `$sizes` Sass map in `_variabl
 {% endcapture %}
 {% include example.html content=example %}
 
-You can also use `max-width: 100%;` and `max-height: 100%;` utilities as needed.
+You can also use `max-width` and `max-height` utilities as needed.
 
 {% capture example %}
-{% include icons/placeholder.svg width="100%" height="100" class="mw-100" text="Max-width 100%" title="Max-width 100%" %}
+<div class="mw-25 p-3" style="width: 100%; background-color: #eee;">Max-width 25%</div>
+<div class="mw-50 p-3" style="width: 100%; background-color: #eee;">Max-width 50%</div>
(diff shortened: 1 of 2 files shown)


site/docs/4.1/utilities/sizing.md
```diff
@@ -30,16 +30,22 @@ Width and height utilities are generated from the `$sizes` Sass map in `_variabl
 {% endcapture %}
 {% include example.html content=example %}
 
-You can also use `max-width: 100%;` and `max-height: 100%;` utilities as needed.
+You can also use `max-width` and `max-height` utilities as needed.
 
 {% capture example %}
-{% include icons/placeholder.svg width="100%" height="100" class="mw-100" text="Max-width 100%" title="Max-width 100%" %}
+<div class="mw-25 p-3" style="width: 100%; background-color: #eee;">Max-width 25%</div>
```
XhmikosR: You should probably use a loop here. Which reminds me, we should use a loop in the examples above these too.

MartijnCuppens:
I'm a bit worried the `max-height` utilities are an addition that will only be used in rare cases.

mdo:
In agreement with @MartijnCuppens—all those variations get added for little use, and then folks will want responsive variations. Gotta call it somewhere unfortunately—thanks though!

royklutman:
Totally agree, I understand. Cheers!

PR #27713, opened 2018-11-23
Fixed missing closing tag in dashboard example

mike-odom:


ThatOdieGuy:
What happened here?

XhmikosR:
https://www.w3.org/TR/html52/sec-forms.html#the-input-element

ThatOdieGuy:
oh, sorry. I've been living in React too long.

rafalp:
Thats not correct JSX neither, even for `textarea` which is self-closing tag in react ;)

ThatOdieGuy:
JSX needs a closing tag, although I cannot find any documentation that explicitly says this, I cannot write code without closing tags in JSX. It will produce an error. `<input>` alone is invalid. `<input/>` or `<input></input>` is necessary.
The same is true for textarea, JSX appears to always needs a closing tag or a self-closing tag.

See
https://jsfiddle.net/jLdkwbte/

ThatOdieGuy:
Great, I'm still correct. Good to hear. :)

So, would you accept a fix of making that input tag a self-closing tag?

rafalp:
> So, would you accept a fix of making that input tag a self-closing tag?

Bootstrap uses HTML5, you don't need to make input self-closing to be valid.

PR #27719, opened 2018-11-23
prevent text selection for floating labels example

m5o:
* prevent text selection for floating labels
  * expand the click target by not selecting the label text

🤓 feedback welcome


MartijnCuppens:
Maybe we can use `pointer-events: none;` here? This generates less code because it doesn't need to be prefixed and has the same result?

MartijnCuppens (approved):
LGTM! Thanks @m5o!

PR #27731, opened 2018-11-26
FEATURE: Provide also `breakpoint-prev`

jonnitto:
When you use Bootstrap as a layout tool, with all it mixins, I come often in the situation where I want to select the previous breakpoint. For example, I set a variable for the breakpoint of a navigation and want to use the `media-breakpoint-down` mixin. Without this function, I have to set a second variable with the lower breakpoint. With this function, I can set it dynamically:

```scss
$variable: lg;

@include media-breakpoint-down(breakpoint-prev($variable)) {
   // Content
}
```

### DIFF
--- a/scss/mixins/_breakpoints.scss
+++ b/scss/mixins/_breakpoints.scss
@@ -29,6 +29,7 @@
 //    sm
 
 @function breakpoint-prev($name, $breakpoints: $grid-breakpoints, $breakpoint-names: map-keys($breakpoints)) {
+  $n: index($breakpoint-names, $name);
   @if ($n == 1) {
     @return null;
   }
--- a/scss/mixins/_breakpoints.scss
+++ b/scss/mixins/_breakpoints.scss
@@ -30,9 +30,9 @@
 
 @function breakpoint-prev($name, $breakpoints: $grid-breakpoints, $breakpoint-names: map-keys($breakpoints)) {
   @if ($n == 1) {
-      @return null;
-    }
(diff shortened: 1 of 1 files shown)


scss/mixins/_breakpoints.scss
```diff
@@ -19,6 +19,22 @@
...
+//
+//    >> breakpoint-prev(md)
+//    sm
+//    >> breakpoint-prev(sm, (xs: 0, sm: 576px, md: 768px, lg: 992px, xl: 1200px))
+//    xs
+//    >> breakpoint-prev(md, $breakpoint-names: (xs sm md lg xl))
+//    sm
+
+@function breakpoint-prev($name, $breakpoints: $grid-breakpoints, $breakpoint-names: map-keys($breakpoints)) {
+  @if ($n == 1) {
```
MartijnCuppens: What's `$n`?

jonnitto: Ups. Copy'n paste error. Thank you for the review. I fixed it.

jonnitto: Thanks for the review. I add the missing line.

MartijnCuppens:
Hi @jonnitto.

Thanks for this PR, but I'm not really convinced of the added value of this. The function will not be used by Bootstrap by default and imo this is not something a lot of developers will use frequently. We'll need to draw the line somewhere to prevent a proliferation of functions.

PR #27739, opened 2018-11-28
Make custom control label padding override easier

jacobmllr95:
Currently it's quite difficult to override or remove the space between the custom control indicator and the actual label. You have to change the `padding-left` of the `.custom-control` and the negative `left` positioning of the indicator background and foreground.

With this PR you only have to change the `padding-left` of the `.custom-control` to the indicator size + `x` to control the space between indicator and label. The indicator will always remain all the way left relative to the `.custom-control` itself.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -67,7 +67,6 @@
 // Build the custom controls out of pseudo-elements.
 
 .custom-control-label {
-  position: relative;
   margin-bottom: 0;
   vertical-align: top;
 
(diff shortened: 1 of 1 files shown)


ysds:
That is necessary for #25944.

jackmu95:
@ysds Can you confirm that the issue still exists with the [Netlify deloy](https://deploy-preview-27739--twbs-bootstrap4.netlify.com/docs/4.1/components/forms/#checkboxes-and-radios-1) of this PR. I have tried it on IE 11 and it seems to work for me.

ysds:
https://deploy-preview-27739--twbs-bootstrap4.netlify.com/docs/4.1/examples/checkout/

jackmu95:
@ysds Your right, the issue is still present :/

I wasn't able to come up with a better solution. I will close this for now.

PR #27747, opened 2018-11-30
Check modal close time exist open modal

muradsofiyev:
You remove `modal-open` class from the `body ` every modal close time but I think you must be check exist open modal.


Johann-S:
Hi @muradsofiyev,

We don't need that check because Bootstrap do not allow multi modal see in our docs: 
https://getbootstrap.com/docs/4.1/components/modal/#how-it-works

>Bootstrap only supports one modal window at a time. Nested modals aren’t supported as we believe them to be poor user experiences.

muradsofiyev:
Please look snippet. 
https://jsfiddle.net/92q0b43n/

muradsofiyev:
Please look this snippet 
https://jsfiddle.net/92q0b43n/

Johann-S:
yep that's what we called multi modals or nested modals

PR #27750, opened 2018-11-30
Apply arrow styles to direct descendants of bootstrap popover classes

sctape:
With the current styles, it is not possible to nest a popover inside of
another popover if they have different placements because the arrow
styles of the parent popover will conflict with the arrow styles of the
child popover.

Here are some screenshots of the described behavior from my app.
screen shot 2018-11-30 at 2 33 20 pm

screen shot 2018-11-30 at 2 32 44 pm

Please let me know if you have any questions.
Thanks,
Sam

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -38,21 +38,21 @@
 .bs-popover-top {
   margin-bottom: $popover-arrow-height;
 
-  .arrow {
+  > .arrow {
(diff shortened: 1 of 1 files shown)


mdo (approved):
Seems fine to me I think, but would like feedback from @twbs/css-review, too. Only issue here could be for folks who've customized the popover template.

MartijnCuppens:
> Only issue here could be for folks who've customized the popover template.

Indeed, but I don't think we should worry about that too much.

PR #27754, opened 2018-12-02
Update Scrollspy doc to include missing method options

### LINKED ISSUE #25799: Missing options in Scrollspy documentation
The only [scrollspy option](http://getbootstrap.com/docs/4.0/components/scrollspy/#options) listed in the documentation is `offset`. However, the same page has an example with option [`target`](http://getbootstrap.com/docs/4.0/components/scrollspy/#via-javascript). Also the code mention (and use) an option [_offset method_](https://github.com/twbs/bootstrap/blob/fb60a4a9867eb1fc18cf310de2bbb8f21e262e6b/js/src/scrollspy.js#L33-L34). Those two options should be present in the options table.

unaccomplished:
Update missing `method` options in Scrollspy documentation.

Fixes #25799


XhmikosR:
@Johann-S: LGTY?

Johann-S:
I think @unaccomplished will add some information about `target` option too (see: https://github.com/twbs/bootstrap/issues/25799#issuecomment-443554618)

Johann-S (approved):

PR #27766, opened 2018-12-04
Make form grid gutter width configurable

jacobmllr95:
This PR makes the form grid gutter width configurable by the `$form-grid-gutter-width` variable.

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -190,13 +190,13 @@ textarea.form-control {
 .form-row {
   display: flex;
   flex-wrap: wrap;
-  margin-right: -5px;
-  margin-left: -5px;
+  margin-right: -($form-grid-gutter-width / 2);
+  margin-left: -($form-grid-gutter-width / 2);
 
   > .col,
   > [class*="col-"] {
-    padding-right: 5px;
-    padding-left: 5px;
+    padding-right: ($form-grid-gutter-width / 2);
+    padding-left: ($form-grid-gutter-width / 2);
   }
 }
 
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -499,6 +499,8 @@ $input-transition:                      border-color .15s ease-in-out, box-shado
 
 $form-text-margin-top:                  .25rem !default;
 
+$form-grid-gutter-width:                10px !default;
```
MartijnCuppens: I would move this line to where `$form-group-margin-bottom` is set, those variables are a bit related.

scss/_forms.scss
```diff
@@ -190,13 +190,13 @@ textarea.form-control {
...
-  margin-right: -5px;
-  margin-left: -5px;
+  margin-right: -($form-grid-gutter-width / 2);
+  margin-left: -($form-grid-gutter-width / 2);
 
   > .col,
   > [class*="col-"] {
-    padding-right: 5px;
-    padding-left: 5px;
+    padding-right: ($form-grid-gutter-width / 2);
```
MartijnCuppens: I don't think the brackets are needed here, can you check if you can remove them?

jackmu95: I don't think so too, but they are also present in [here](https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_grid-framework.scss#L11).

MartijnCuppens: Good point. I just added #27768 to remove those redundant brackets. You can also remove it here.

MartijnCuppens (approved):
LGTM!

PR #27768, opened 2018-12-04
Remove redundant brackets

MartijnCuppens:
There are a lot of redundant brackets in the scss, this PR cleans this up a bit. I still left the brackets on some places if it increased the readability of the code like here:
https://github.com/twbs/bootstrap/blob/e9ec5e73929a46852abf9c7d4e5fc03d50f76d3c/scss/_popover.scss#L134

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -300,7 +300,7 @@ $h4-font-size:                $font-size-base * 1.5 !default;
 $h5-font-size:                $font-size-base * 1.25 !default;
 $h6-font-size:                $font-size-base !default;
 
-$headings-margin-bottom:      ($spacer / 2) !default;
+$headings-margin-bottom:      $spacer / 2 !default;
 $headings-font-family:        inherit !default;
 $headings-font-weight:        500 !default;
 $headings-line-height:        1.2 !default;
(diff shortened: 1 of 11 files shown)


XhmikosR:
I don't think there's a rule for this, right? In stylelint-scss.

scss/_functions.scss
```diff
@@ -54,7 +54,7 @@
   $g: green($color);
   $b: blue($color);
 
-  $yiq: (($r * 299) + ($g * 587) + ($b * 114)) / 1000;
+  $yiq: ($r * 299 + $g * 587 + $b * 114) / 1000;
```
XhmikosR: I'd keep the brackets here.

MartijnCuppens:
@XhmikosR, no there isn't at the moment. My gut feeling tells me that'll also be a hard one to implement.

scss/_functions.scss
MartijnCuppens: Yeah, I wasn't completely sure with this one. Just restored it.

XhmikosR:
LGTM AFAIK, but let's wait to hear from @mdo too

mdo:
I'm fine to do either way here—having the parentheses helped me with readability and consistency across the project. Obviously a stylistic thing, so if folks don't want them, I'm fine with it. Would be also nice to have a linter for it though as @XhmikosR was suggesting.

MartijnCuppens:
We're actually quite inconsistent with this, for example these calculations don't use brackets:
https://github.com/twbs/bootstrap/blob/0a55c7a8f62ff0b7d95d42e3d09db9927f85c439/scss/_variables.scss#L296-L301
https://github.com/twbs/bootstrap/blob/0a55c7a8f62ff0b7d95d42e3d09db9927f85c439/scss/_variables.scss#L487
https://github.com/twbs/bootstrap/blob/0a55c7a8f62ff0b7d95d42e3d09db9927f85c439/scss/_button-group.scss#L70-L71

> My main concern here is that if we can't automate the check, it's moot.

There are other things we always need to check ourself like the leading and trailing whitelines of titles in markdown files. I don't really like the idea that you can use whatever style you like here. Linting it would definitely be the best option, but I'm afraid this would be a plugin that is difficult to create, because you can have calc functions, scss variables, multiple value per property,...

scss/_variables.scss
```diff
@@ -486,13 +486,13 @@ $input-plaintext-color:                 $body-color !default;
 
 $input-height-border:                   $input-border-width * 2 !default;
 
-$input-height-inner:                    ($input-btn-font-size * $input-btn-line-height) + ($input-btn-padding-y * 2) !default;
+$input-height-inner:                    $input-btn-font-size * $input-btn-line-height + $input-btn-padding-y * 2 !default;
```
XhmikosR: I'd keep these and the other 2 below.

XhmikosR:
Let's go with it, just a 2-3 cases I think the brackets make it clearer.

XhmikosR (approved):

PR #27769, opened 2018-12-05
Scrollable modal

ysds:
This feature is inspired by [MDC Dialog](https://material-components.github.io/material-components-web-catalog/#/component/dialog).

The scrollable modal allows scrolling the modal body. The standard model height fits to own content height, but the scrollable modal fits within the viewport height if the content is too long.

image

demo: https://deploy-preview-27769--twbs-bootstrap4.netlify.com/docs/4.2/components/modal/#scrolling-long-content

Cheers!

### DIFF
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -51,6 +51,7 @@ const Event = {
 }
 
 const ClassName = {
+  SCROLLABLE         : 'modal-dialog-scrollable',
   SCROLLBAR_MEASURER : 'modal-scrollbar-measure',
   BACKDROP           : 'modal-backdrop',
   OPEN               : 'modal-open',
@@ -60,6 +61,7 @@ const ClassName = {
 
 const Selector = {
   DIALOG         : '.modal-dialog',
+  MODAL_BODY     : '.modal-body',
   DATA_TOGGLE    : '[data-toggle="modal"]',
   DATA_DISMISS   : '[data-dismiss="modal"]',
   FIXED_CONTENT  : '.fixed-top, .fixed-bottom, .is-fixed, .sticky-top',
@@ -244,7 +246,12 @@ class Modal {
     this._element.style.display = 'block'
     this._element.removeAttribute('aria-hidden')
     this._element.setAttribute('aria-modal', true)
-    this._element.scrollTop = 0
+
+    if ($(this._dialog).hasClass(ClassName.SCROLLABLE)) {
+      this._dialog.querySelector(Selector.MODAL_BODY).scrollTop = 0
+    } else {
+      this._element.scrollTop = 0
+    }
 
     if (transition) {
       Util.reflow(this._element)
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -50,6 +50,24 @@
   }
 }
 
+.modal-dialog-scrollable {
+  display: flex; // IE10/11
+
+  .modal-content {
+    max-height: calc(100vh - #{$modal-dialog-margin * 2});
+    overflow: hidden;
+  }
+
+  .modal-header,
+  .modal-footer {
+    flex-shrink: 0;
+  }
+
+  .modal-body {
+    overflow-y: auto;
+  }
+}
+
 .modal-dialog-centered {
   display: flex;
   align-items: center;
@@ -60,6 +78,22 @@
     display: block; // IE10
     height: calc(100vh - #{$modal-dialog-margin * 2});
     content: "";
+  }
+
+  // Ensure `.modal-body` shows scrollbar (IE10/11)
+  &.modal-dialog-scrollable {
+    flex-direction: column;
+    justify-content: center;
+    height: calc(100vh - #{$modal-dialog-margin * 2});
+    overflow: hidden;
+
+    .modal-content {
+      max-height: none;
+    }
+
+    &::before {
+      content: none;
+    }
   }
 }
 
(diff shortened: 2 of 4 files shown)


js/src/modal.js
```diff
@@ -245,6 +247,10 @@ class Modal {
     this._element.removeAttribute('aria-hidden')
     this._element.scrollTop = 0
 
+    if ($(this._dialog).hasClass(ClassName.SCROLLABLE)) {
+      this._dialog.querySelector(Selector.MODAL_BODY).scrollTop = 0
```
MartijnCuppens: Nice addition!

scss/_modal.scss
```diff
@@ -50,6 +50,28 @@
...
+.modal-dialog-scrollable {
+  display: flex;
+  flex-direction: column;
+  height: calc(100% - (#{$modal-dialog-margin} * 2));
+  overflow: hidden;
+
+  > .modal-content {
+    max-height: 100%;
+    overflow: hidden;
+    -ms-overflow-y: visible;
```
MartijnCuppens: What does this property do?

scss/_modal.scss
```diff
@@ -61,6 +83,14 @@
     height: calc(100vh - (#{$modal-dialog-margin} * 2));
     content: "";
   }
+
+  &.modal-dialog-scrollable {
+    justify-content: center;
+
+    &::before {
+      display: none;
```
MartijnCuppens: Wow, you managed to get this working in IE10 without the `::before` hack. Is it possible completely remove the `::before` hack? (maybe we can cover this in another PR if this is the case).

scss/_modal.scss
```diff
@@ -159,6 +189,10 @@
     margin: $modal-dialog-margin-y-sm-up auto;
   }
 
+  .modal-dialog-scrollable {
+    height: calc(100% - (#{$modal-dialog-margin-y-sm-up} * 2));
```
MartijnCuppens: This should also work and generates less css:
```suggestion
    height: calc(100% - #{$modal-dialog-margin-y-sm-up * 2});
```

Johann-S (changes requested):
Otherwise LGTM 👍

scss/_modal.scss
ysds: Unfortunately, that would not be possible, because scrollable modal does not require flex with `min-height`, but it is necessary for a vertically centered (not scrollable body) modal.

scss/_modal.scss
ysds: For support vertically resize on IE. Resizable modal does not work well without `overflow: visible`;or `overflow: auto`. Both are OK but the behavior of the scroll bar is different.

Johann-S:
@ysds you have to check when you a modal is shown, if this modal has `.modal-dialog-scrollable` class, the modal body `scrollTop` property should be equal to `0`

scss/_modal.scss
MartijnCuppens: That was what I was afraid of. Can you change this to `content: none;` instead of `display: none;`? It would make more sense to have no pseudo element than a hidden pseudo element.

MartijnCuppens:
@ysds, you'll need to add a test to this file: https://github.com/twbs/bootstrap/blob/v4-dev/js/tests/unit/modal.js
The scenario Johann describes should be a good test. I think you'll find out how the tests work if you look at other tests, but if encounter problems with it, let us know.

tmorehouse:
I wonder if the scrollable content should be default behaviour on vertically centered modals?

ysds:
@Johann-S I think that test result will always be OK, because the initial scrollTop of a modal body is 0. After creating a modal that overflows the height of the viewport, then it is necessary to change the scroll potion. How can I define a modal that overflows the height of the viewport? What is the height of the viewport in the test?

(I have never written a Qunit test script)

Johann-S:
@ysds it seems not see: https://coveralls.io/builds/20490418/source?filename=js/src/modal.js#L251 
If it was the case this line would be covered. 

You have to create a modal with a height greater than `window.outerHeight` to overflow the height of the viewport.
A QUnit test or a unit test in a DOM, is the same as writting JS, you have to create the HTML a be sure your logic works

MartijnCuppens (dismissed):
The css LGTM

Johann-S (approved):
LGTM 👍

XhmikosR:
Rebased.

site/docs/4.2/components/modal.md
```diff
@@ -234,9 +307,36 @@ Add `.modal-dialog-centered` to `.modal-dialog` to vertically center the modal.
   </div>
 </div>
 
+<div id="exampleModalCenteredScrollable" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="exampleModalCenteredScrollableTitle" aria-hidden="true">
```
XhmikosR: It seems this one doesn't have enough content to show as scrollable.

ysds: It is intentional for reproduce both centering and scrollable. If too much content, nobody can see centered modal.

XhmikosR: I'm not sure how useful this is then. But if everyone else thinks it's OK, that's fine I guess.

scss/_modal.scss
```diff
@@ -50,6 +50,28 @@
   }
 }
 
+.modal-dialog-scrollable {
+  display: flex;
+  flex-direction: column;
+  height: calc(100% - (#{$modal-dialog-margin} * 2));
+  overflow: hidden;
+
+  > .modal-content {
```
mdo: Only nitpick here is we likely don't need to use the `>` children selectors here—the HTML hierarchy is a requirement and we don't support nested modals. Thoughts?

XhmikosR:
@mdo: Is there something else we need here or can we merge?

ysds:
Please wait to marge! I found a better workaround for the strange scrollbar behavior on IE.

LuukLeijtens (approved):

MartijnCuppens (approved):

mdo (approved):
Excited for this! Thank you <3.

PR #27771, opened 2018-12-05
Add quotes according to the stylelint rule

ysds:
`Noto Sans` is expected to be considered a violation by the `"font-family-name-quotes": "always-where-recommended"` rule, but it does not seem to work for SCSS variables.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -275,7 +275,7 @@ $embed-responsive-aspect-ratios: join(
 // Font, line-height, and color for body text, headings, and more.
 
 // stylelint-disable value-keyword-case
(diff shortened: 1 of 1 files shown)


XhmikosR:
But it seems we got no failures.

Just FYI I remember this quotes issue in the past and IIRC the spec does allow a whitespace without quotes.

That being said I'm all for it for consistency, I'd just feel safer if Stylelint (SCSS) complained.

ysds:
It never fails. If a font family is defined as an **SCSS variable**, lint does not work (regardless of whether correct or not). In other words, it is ignored.

The following CSS gets linted:
```css
a { font-family: Noto Sans; }
```

But the following SCSS doesn't get linted.
```scss
$font: Noto Sans;
a { font-family: $font; }
```

XhmikosR (approved):

XhmikosR:
@MartijnCuppens push the button if you agree.

PR #27782, opened 2018-12-05
Replaced fixed 'bolder' font weight value in reboot with variable. 

### LINKED ISSUE #27774: Replace fixed "bolder" value in reboot with available variable "font-weight-bolder"
In [_reboot.scss#L164](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L164)
```scss
b,
strong {
  font-weight: bolder; // Add the correct font weight in Chrome, Edge, and Safari
}
```

we have a fixed value `bolder`. In my opinion this should be replaced with the available variable `$font-weight-bolder` in [_variables.scss#L291](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L291) 

Useful if you want to use a fixed bold value for `<strong>` & `<b>` tags.

Thoughts? I can submit a PR if needed.

gijsroge:
The default font weight value for `<strong>` `<br>` in reboot was set by a fixed value we now use the variable `$font-weight-bolder` to let users override this value.

Fixes #27774

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -161,7 +161,7 @@ blockquote {
 
 b,
 strong {
-  font-weight: bolder; // Add the correct font weight in Chrome, Edge, and Safari
+  font-weight: $font-weight-bolder; // Add the correct font weight in Chrome, Edge, and Safari
 }
 
 small {


MartijnCuppens (approved):
LGTM, thanks Gijs!

gijsroge:
@MartijnCuppens 🤘

PR #27785, opened 2018-12-05
Make `.close` hover focus styles less specific

### LINKED ISSUE #27784: Bootstrap 4.1.2 `.close` hover focus styles are too specific
https://github.com/twbs/bootstrap/commit/4f1bb98587411d6db16783ea116ff6d5d8f8fa49 from 4.1.2 moved hover focus styles on the `.close` component into `&:not(:disabled):not(.disabled)` making it too specific. This breaks hover focus states for those that extended the close component from previous versions.

https://github.com/twbs/bootstrap/commit/eb5d0b49dc7439fa1cc3d0c41406eef00de149c1 makes it a little better, but the problem is still there.

pat270:
This pull removes the `&:not(:disabled):not(.disabled)` selector on the `.close` component and styles `:disabled`/`.disabled` directly because the `:not` pseudo classes are too hard to overwrite (breaks for people that extended `.close` based on the old version).

Also added variables:
```
$close-opacity
$close-hover-color
$close-hover-opacity
$close-disabled-opacity
```

Fixes #27784

### DIFF
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -5,19 +5,20 @@
   line-height: 1;
   color: $close-color;
   text-shadow: $close-text-shadow;
-  opacity: .5;
+  opacity: $close-opacity;
 
-  // Override <a>'s hover style
-  @include hover {
-    color: $close-color;
+  @include hover-focus {
+    color: $close-hover-color;
     text-decoration: none;
+    opacity: $close-hover-opacity;
+  }
+
+  &.disabled,
+  &:disabled {
+    opacity: $close-disabled-opacity;
   }
 
   &:not(:disabled):not(.disabled) {
-    @include hover-focus {
-      opacity: .75;
-    }
-
     // Opinionated: add "hand" cursor to non-disabled .close elements
     cursor: pointer;
   }
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1058,7 +1058,12 @@ $close-font-size:                   $font-size-base * 1.5 !default;
 $close-font-weight:                 $font-weight-bold !default;
 $close-color:                       $black !default;
 $close-text-shadow:                 0 1px 0 $white !default;
+$close-opacity:                     .5 !default;
 
+$close-hover-color:                 $close-color !default;
+$close-hover-opacity:               .75 !default;
+
+$close-disabled-opacity:            $close-opacity !default;
 
 // Code
 


scss/_close.scss
```diff
@@ -5,21 +5,21 @@
...
   text-shadow: $close-text-shadow;
-  opacity: .5;
+  // Opinionated: add "hand" cursor to non-disabled .close elements
+  cursor: pointer;
+  opacity: $close-opacity;
 
-  // Override <a>'s hover style
-  @include hover {
-    color: $close-color;
+  &:hover,
```
MartijnCuppens: We got a `hover-focus` mixin for this. Can you change this?

MartijnCuppens (dismissed):

scss/_close.scss
pat270: @MartijnCuppens It's been deprecated at https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_hover.scss#L1. Let me know if you still want me to resend it.

MartijnCuppens: The `hover` mixin is deprecated, but I'm not sure about the state of the `hover-focus` mixin. @mdo, what shall we do here?

mdo: `hover-focus` is fine, `hover` is gone.

mdo:
We moved to this approach awhile ago across most of the project to simplify our the amount of code to properly disable things. If we change it here, I fear we need to change it everywhere. Are there other places we need to revisit as well?

pat270:
Changing it across the board is my preference, but might cause some unintended consequences for any Bootstrap themes in the wild.

`&:not(:disabled):not(.disabled)` is used in a handful of places `buttons`, `close`, `navbar` and `pagination`, generally for the cursor property. These existed in `4.0.0` and we were able to account for these. The changes to `.close` for:
```
  &:not(:disabled):not(.disabled) {
	    @include hover-focus {
	      opacity: .75;
	    }
```
came too late unfortunately.  What if we kept:
```
  &:not(:disabled):not(.disabled) {
    // Opinionated: add "hand" cursor to non-disabled .close elements
    cursor: pointer;
  }
```
and moved the hover-focus opacity outside of that selector?

mdo:
That solution could work perhaps, but then we're still adding extra selectors. Going to have to review this more in-depth.

pat270:
@mdo if you prefer to change it everywhere, I can resend with changes everywhere.

ysds:
I do not know why `.close` supports `<a>`.  `<button>` is enough for accessibility. 

> In general you should only use an anchor for navigation using a proper URL.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Accessibility_concerns

If we drop support of `<a>`, its CSS will be more simple.

patrickhlauke:
> I do not know why `.close` supports `<a>`

Because sadly, many authors (either directly, or indirectly because their tooling does it) like to misuse `<a>` for button-style controls...

ysds:
@patrickhlauke OK I understand.

@pat270 can fix the conflict ?

Issue #27786, opened 2018-12-05
dropdown.js line 471


emilmuller:
Incidentally, while creating my own dropdown class that has no jQuery dependencies, I came across this peculiar line:

`[...] (event.which !== ESCAPE_KEYCODE || event.which !== SPACE_KEYCODE) [...]`

which, by all logical means, will always be true. Should it be && ?

Line 471, dropdown.js :)

patrickhlauke:
indeed, well spotted. https://github.com/twbs/bootstrap/pull/27787

emilmuller:
@patrickhlauke Also, hide() does not set aria-expanded = false :)

Johann-S:
Maybe it can help you @emilmuller we already wrote our Dropdown plugin without jQuery here: https://github.com/twbs/bootstrap/blob/v4-without-jquery/js/src/dropdown.js

emilmuller:
@Johann-S Thanks! I'm replacing my code now :)

PR #27787, opened 2018-12-06
Remove nonsensical boolean check

patrickhlauke:
...as it always evaluates to true

closes https://github.com/twbs/bootstrap/issues/27786


Johann-S:
It seems more logical to me this way @patrickhlauke 👍

Johann-S (approved):

PR #27790, opened 2018-12-06
avoid console call in our unit tests

Johann-S:


js/tests/unit/tooltip.js
```diff
@@ -563,7 +563,6 @@ $(function () {
       $tooltip.bootstrapTooltip('show')
     } catch (err) {
       passed = false
-      console.log(err)
```
XhmikosR: Maybe this is useful to keep?

Johann-S: don't think so because, we set `passed` to false, so our unit test will fail

XhmikosR (approved):

PR #27797, opened 2018-12-07
Custom control gutter fix

gijsbotje:
The current `$custom-control-gutter` defines the gutter + the indicator size. This works great if the indicator is the same size for each custom control that uses this variable. While working on the switch element in #23004 i realized that i had to do weird calculations to get the same gutter as for the radio and checkbox. The actual gutter is .5rem, but is now 1.5rem. This PR changes that so the gutter variable is actually the space between the indicator and the label.


MartijnCuppens (approved):
This LGTM because the variable name now represents what is does.

gijsbotje:
this might introduce a breaking change for people that use the sass source code. though i doubt they would use this variable.

MartijnCuppens:
Indeed. But maybe we should document this in the release notes? @mdo, what do you think?

PR #27804, opened 2018-12-08
Fix alert showing text in one line

iiegor:
IMHO breaking words should be the default


Ruffio (commented):
What exactly is the change here?

image

iiegor:
@Ruffio stylelint threw `Expected "word-break" to come before "border"` so I needed to reorder rules.

iiegor:
> If we need something like this, I'd rather go the route of #27863. Closing on this for that PR, and waiting on some other improvements that hopefully come in v5. Thanks though!

Makes all the sense! 👍

PR #27812, opened 2018-12-11
WIP Docs: add toggle spinner

XhmikosR:
TODO:

- [x] Inject the buttons via JS
- [ ] See if we can make the initial state show something since currently when the page loads the spinner is paused


site/docs/4.1/assets/js/src/application.js
```diff
@@ -58,6 +58,16 @@
       $(this).siblings('.progress').find('.progress-bar-striped').toggleClass('progress-bar-animated')
     })
 
+    // Add Toggle Spinner buttons
+    $('.bd-example-spinner').each(function (i, el) {
+      var btnHtml = '<button type="button" class="btn btn-secondary btn-spinner" data-toggle="button" aria-pressed="false" autocomplete="off">Toggle Spinner</button>'
```
Johann-S: this line should be before the call of `.each`

MartijnCuppens: I'm not that familiar with the js best practices, but wouldn't it be better to use something like `document.createElement` here?

XhmikosR:
@mdo @Johann-S : what should we do with this?

site/docs/4.1/assets/js/src/application.js
Johann-S: @MartijnCuppens it depends if you work on the created element or not.
If you work on the created element (for example by adding attributes) you should use `document.createElement` if not you can use `innerHTML` or `appendChild`

The issue here, it's `btnHtml` is created on each iteration instead of one

MartijnCuppens:
Why are we actually doing this? Imo the docs look fine the way they are?

XhmikosR:
Well, it works fine on my machine and the devices I own, but I'm not sure if it's the same with low-end devices, with all the animations firing up at once. I do see one core being used 100%, but I don't have any low-end device handy to test.

BTW this was a thought of @mdo IIRC in the initial spinners PR.

As long as we can make the grow spinners work, it shouldn't hurt us to have this. Alternatively, we could try being fancy and maybe use Intersection  Observer to only enable the spinners that are in the viewport. That being said, I don't have time to implement this, it's just an idea.

XhmikosR:
@mdo: should we just drop this? We haven't seen any comments.

XhmikosR:
I'm gonna drop this for now since we haven't seen any reports.

PR #27813, opened 2018-12-11
Replace uglify-js with terser.

XhmikosR:
Opening this to check the diff and investigate the switch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1550,6 +1550,12 @@
       "integrity": "sha1-+PeLdniYiO858gXNY39o5wISKyw=",
       "dev": true
     },
+    "buffer-from": {
+      "version": "1.1.1",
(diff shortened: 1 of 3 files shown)


XhmikosR:
@Johann-S: your thoughts? Not sure about this personally. I mean, to me, the most stable is closure, but it requires more work to configure it and it's a lot slower, although the latter let's say we ignore it.

On the other hand, I don't see any fixes in uglify for months and the latest version has known issues.

Johann-S (approved):
The diff LGTM 👍

XhmikosR:
I'm still a little scared to make the switch yet...

Johann-S:
personnally I want this in v5.

BTW Angular switch to Terser in their v7 and Terser did a lot of things so I think now it should works better

XhmikosR:
Ok, let's merge it and see how it goes. Compared to an unmaintained package with known bugs, this *should* be better.

PR #27820, opened 2018-12-12
Fix margin above toast when previous toast is hidden

### LINKED ISSUE #27807: Remove toasts from DOM after hiding them
From https://github.com/twbs/bootstrap/pull/27792#issuecomment-445064288:
We also do this with the alerts. The problem with not removing them is that there's a margin above the second toast if we dismiss the first: https://deploy-preview-27792--twbs-bootstrap4.netlify.com/docs/4.1/components/toasts/#stacking

MartijnCuppens:
There's a margin above the second toast if we dismiss the first: https://deploy-preview-27792--twbs-bootstrap4.netlify.com/docs/4.1/components/toasts/#stacking

Demo solution: https://deploy-preview-27820--twbs-bootstrap4.netlify.com/docs/4.1/components/toasts/#stacking

Closes #27807

### DIFF
--- a/scss/_toasts.scss
+++ b/scss/_toasts.scss
@@ -10,8 +10,8 @@
   backdrop-filter: blur(10px);
   opacity: 0;
 
-  + .toast {
-    margin-top: $toast-padding-x;
+  &:not(:last-child) {
+    margin-bottom: $toast-padding-x;
   }
 
   &.showing {


MartijnCuppens:
Not sure if this fixes https://github.com/twbs/bootstrap/issues/27807, we still have an inconsistent behaviour between the alerts and toasts (alerts are removed from the DOM, toasts aren't)

Johann-S:
I think it's not the same behavior we want for Alerts and Toasts. Alerts have a state (warning, error, info) and should be displayed once one the other hand Toasts don't have states and can be reusable

Johann-S (approved):
LGTM 👍

PR #27826, opened 2018-12-12
SVG accessibility fixes

patrickhlauke:
- add `focusable="false"` to all SVGs, to stop IE from including the `<svg>` in its default focus cycle
- completely hide decorative SVGs using `aria-hidden="true"`
- add `role="img"` to meaningful/"content" SVGs, ensure they have an appropriate `<title>` and/or `aria-label`


patrickhlauke:
@XhmikosR force-pushed a change, and expanded this PR quite a bit to remove some redundant/overly descriptive text (now that this is constructed from both title and text). also includes changes to SVGO configuration to not remove `role` and to add/force `focusable="false"`

build/svgo.yml
```diff
@@ -15,6 +15,10 @@ js2svg:
   indent: 2
 
 plugins:
+  - addAttributesToSVGElement:
+      attributes:
+        - focusable: false
+        - data-image: icon
```
XhmikosR: Why do we need `data-image`?

Also, note that we'd need to run svgo with the updated config before merging this.

patrickhlauke: sorry, cargo-culted that `data-image` part from the plugin's documentation, but you're quite right it's not needed. force-pushed an updated version without it.

when you say "we'd need to run svgo" ... where do you mean? against the files in docs? when i initially tried, it choked on the `{{ ... }}` etc liquid template bits.

XhmikosR: Yeah, it won't work for those where we have liquid code, but it'll work for the rest. I will take care of it and push here later.

XhmikosR: Actually I looked at it and it's OK as it is since svgo adds whitespace with this config.

XhmikosR (approved):

PR #27837, opened 2018-12-14
Fix custom control gutter on switch

gijsbotje:
after #27797 was merged the switch component was broken.
this fixes that issue


MartijnCuppens (approved):
Thanks for the quick fix, @gijsbotje!

PR #27850, opened 2018-12-15
Add table-hover-color variables

### LINKED ISSUE #26465: v4 $table-hover-color option needed
hiya, busy theming bootstrap for my app. previously i did something like add my own class to a table and theme that class but this time im using the variables to get bootstrap itself to be themed. 

as is now:
````
$table-cell-padding:          .75rem !default;
$table-cell-padding-sm:       .3rem !default;

$table-bg:                    transparent !default;
$table-accent-bg:             rgba($black, .05) !default;
$table-hover-bg:              rgba($black, .075) !default;
$table-active-bg:             $table-hover-bg !default;

$table-border-width:          $border-width !default;
$table-border-color:          $gray-300 !default;

$table-head-bg:               $gray-200 !default;
$table-head-color:            $gray-700 !default;

$table-dark-bg:               $gray-900 !default;
$table-dark-accent-bg:        rgba($white, .05) !default;
$table-dark-hover-bg:         rgba($white, .075) !default;
$table-dark-border-color:     lighten($gray-900, 7.5%) !default;
$table-dark-color:            $body-bg !default;

$table-striped-order:         odd !default;

$table-caption-color:         $text-muted !default;
````

i propose the addition of 

````
$table-hover-color: $body-color !default;
````

as soon as you make the hover a darker colour for instance  being able to change the fonts colour is quite important. 

its easy enough to add into my custom css but it might break stuff regarding the different primary, secondary, info etc context classes

changes needed:
add `$table-hover-color: $body-color !default;` to _variables.scss
add `color:$table-hover-color` to _tables.scss line 93

bootstrap 4.1

### PR DESCRIPTION
**Pull request by mdo**
Fixes #26465.

Strictly speaking, this might constitute new features with the new variables, so we may need to do this in v4.2 or a v4.3. Lemme know your thoughts, @twbs/css-review.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -352,8 +352,10 @@ $hr-margin-y:                 $spacer !default;
 $table-cell-padding:          .75rem !default;
 $table-cell-padding-sm:       .3rem !default;
 
+$table-color:                 $body-color !default;
 $table-bg:                    transparent !default;
 $table-accent-bg:             rgba($black, .05) !default;
+$table-hover-color:           $table-color !default;
 $table-hover-bg:              rgba($black, .075) !default;
 $table-active-bg:             $table-hover-bg !default;
 
@@ -363,11 +365,12 @@ $table-border-color:          $gray-300 !default;
 $table-head-bg:               $gray-200 !default;
 $table-head-color:            $gray-700 !default;
 
+$table-dark-color:            $white !default;
 $table-dark-bg:               $gray-900 !default;
 $table-dark-accent-bg:        rgba($white, .05) !default;
+$table-dark-hover-color:      $table-dark-color !default;
 $table-dark-hover-bg:         rgba($white, .075) !default;
 $table-dark-border-color:     lighten($gray-900, 7.5%) !default;
-$table-dark-color:            $white !default;
 
 $table-striped-order:         odd !default;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: scss/_variables.scss [hunk 1]
```diff
@@ -352,8 +352,10 @@ $hr-margin-y:                 $spacer !default;
 $table-cell-padding:          .75rem !default;
 $table-cell-padding-sm:       .3rem !default;
 
+$table-color:                 $body-color !default;
```
MartijnCuppens: Where do you use this variable?

XhmikosR: And if it's unused why doesn't fusv catch it :P

mdo: This was moved up from farther down the file.

MartijnCuppens: This looks like a new variable, `$table-color` is not used in `v4-dev`. And indeed, very strange `fusv` doesn't catch this. No idea what's going on here.

XhmikosR: I think it's used in the sense it's used once, to set the `$table-hover-color` hence why fusv doesn't catch this.

mdo: Ah I see what you mean now, I assumed it was already in use within `_tables.scss`. We can add it there though.

MartijnCuppens (approved):

PR #27851, opened 2018-12-15
Add option to remove grid gutters via Sass mixins

### LINKED ISSUE #24273: 'no-gutters' implementation in Sass for grids
We can use `<div class="row no-gutters">` in html, but in case of implementing this in Sass we need to manually add zero-margin/padding:

For example:

```scss
// Grid with 'no-gutters' NOW

.example-row {  
  @include make-row(); // or just @extend .row;
  margin-left: 0;
  margin-right: 0;
}

.example-col-2 {
  @include make-col-ready(); // or just @extend .col-2;
  @include make-col(2);
  padding-left: 0;
  padding-right: 0;
}

.example-col-10 {
  @include make-col-ready();
  @include make-col(10);
  padding-left: 0;
  padding-right: 0;
}
```

Maybe it will be usefull to add an option for Sass mixins like: 
```scss
@mixin make-row($gutters: 'gutters') {
  display: flex;
  flex-wrap: wrap;
  @if $gutters == 'gutters' {
    margin-right: ($grid-gutter-width / -2);
    margin-left:  ($grid-gutter-width / -2);
  }
}

@mixin make-col-ready($gutters: 'gutters') {
  position: relative;
  // Prevent columns from becoming too narrow when at smaller grid tiers by
  // always setting `width: 100%;`. This works because we use `flex` values
  // later on to override this initial width.
  width: 100%;
  min-height: 1px; // Prevent collapsing
  @if $gutters == 'gutters' {
    padding-right: ($grid-gutter-width / 2);
    padding-left:  ($grid-gutter-width / 2);
  }
}
```

So we could write:
```scss
// Grid with 'no-gutters' AFTER

.example-row {  
  @include make-row('no-gutters');
}

.example-col-2 {
  @include make-col-ready('no-gutters');
  @include make-col(2);
}

.example-col-10 {
  @include make-col-ready('no-gutters');
  @include make-col(10);
}
```

mdo:
Fixes #24273.

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -23,8 +23,10 @@
 @mixin make-row($gutter: $grid-gutter-width) {
   display: flex;
   flex-wrap: wrap;
-  margin-right: -$gutter / 2;
-  margin-left: -$gutter / 2;
+  @if not $gutter == "no-gutters" {
+    margin-right: -$gutter / 2;
+    margin-left: -$gutter / 2;
+  }
 }
 
 @mixin make-col-ready($gutter: $grid-gutter-width) {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Since #27841 we can pass the gutter to the mixins, if `0` is passed, you'll get the same effect.

mdo:
@MartijnCuppens Ah, forgot about that! So to be clear, we don't need this PR and we can document how to accomplish a Sass-only "no gutter" solution?

MartijnCuppens:
Yes, indeed!

PR #27818, opened 2018-12-12
Remove now unnecessary shrink-to-fit viewport directive

patrickhlauke:
`shrink-to-fit=no` is not needed anymore - Apple removed the need for it from iOS9.3 onwards

See https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html and https://github.com/h5bp/html5-boilerplate/issues/2102

/cc @scottaohara

PR #27855, opened 2018-12-16
Revert "Remove now unnecessary shrink-to-fit viewport directive"

XhmikosR:
Reverts twbs/bootstrap#27818

Reasoning: In 4.x we support iOS 9 not 9.3.

### DIFF
--- a/js/tests/index.html
+++ b/js/tests/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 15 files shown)


mdo (approved):

XhmikosR:
It's a breaking change given our browsers support list, either we like it or not. This shouldn't have landed fore 4.x.

When we change the browsers we support, this can be reapplied.

PR #27863, opened 2018-12-16
Add new .text-break utility for applying word-break: break-word

### LINKED ISSUE #19315: Add utility class for word-wrap/word-break

mdo:
If we need to be changing the breaking of content, we shouldn't be doing it piece meal on our components. A utility helps in the interim, hence this PR, but v5 will likely see this added to the `<body>` element in Reboot to solve the problem across the entire project. (At that point we'll likely need the opposite utility, too).

Fixes #19315.

### DIFF
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -62,6 +62,8 @@
 
 .text-decoration-none { text-decoration: none !important; }
 
+.text-break    { word-break: break-word !important; }
+
 // Reset
 
 .text-reset { color: inherit !important; }
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -62,7 +62,7 @@
 
 .text-decoration-none { text-decoration: none !important; }
 
-.text-break    { word-break: break-word !important; }
+.text-break { word-break: break-word !important; }
 
 // Reset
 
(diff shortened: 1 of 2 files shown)


scss/utilities/_text.scss
```diff
@@ -62,6 +62,8 @@
 
 .text-decoration-none { text-decoration: none !important; }
 
+.text-break    { word-break: break-word !important; }
```
MartijnCuppens: One space should do the job.
```suggestion
.text-break { word-break: break-word !important; }
```

MartijnCuppens:
Looks good, maybe add some documentation?

mdo:
@MartijnCuppens Added!

MartijnCuppens (approved):
Wonderful!

PR #27866, opened 2018-12-17
Update robots.txt

HYWZ36:
just test.
Thank you.


mdo:
Don't test on public repos—you're literally notifying 7,000+ people.

PR #27870, opened 2018-12-17
Prevent checks from shrinking in inline forms

### LINKED ISSUE #25336: form-inline disappearing checkbox in Chrome Linux
Using dev-v4 with Chrome on Linux Ubuntu. Note: this works fine when I test with Chrome on Windows. 

When using checkbox with form-inline and it the label is more than one line, the checkbox keeps getting smaller and smaller. Thus, two lines of makes the checkbox smaller and four lines makes is super small. Effectively the label text is "squeezing out" the space available for the checkbox to appear. This only happens with form-inline. I tested without form-inline and the checkbox stays the same size regardless of how many lines of text appear next to it.

```
<form class="form-inline">

<div class="form-check col-sm-2">
  <input class="form-check-input" type="checkbox" id="inlineCheckbox1" value="option1">
  <label class="form-check-label" for="inlineCheckbox1">1 asdf sdfasdf asdf asdfasdf asdf asdf asdf asdf asdf sdaf asfd </label>
</div>

</form>
```

checkboxsmall

mdo:
Fixes #25336


MartijnCuppens:
Not sure how to test this, is this a linux only issue?

mdo:
@MartijnCuppens Not Linux only. Take any `.form-check` within `.form-inline` and give it a fixed width of like `100px`.

mdo:
To expand on that, we have styles that change the `.form-check`'s `display` to `flex`, and the default behavior of flex children is to shrink. The only way to prevent this, other than rewriting inline forms, is to prevent the input from behaving like that.

PR #27871, opened 2018-12-17
Add .list-group-striped

### LINKED ISSUE #24628: feature request: .list-striped
Looking to add a **`list-striped`** class that is similar to **`table-striped`** to create a little more visual distinction between list-group-items.  This is especially helpful on wider lists.  _Switching to a single column table to use table-striped isn't really semantically correct._

Sample codepen here: https://codepen.io/anon/pen/JOdBGG

Unfortunately this starts to fall apart when combined with other styles such as .active or the contextual classes..

If it's acceptable that a list-group-striped class doesn't play well with other list-group styles, I could submit a PR for this, otherwise perhaps someone with more expertise can provide a more elegant solution.

mdo:
Fixes #24628


ysds:
@mdo Please reconsider the need for a striped list. There is no striped list in  ios and android. If bootstrap is mobile first, that will not be necessary, IMHO. Users wanting stripes can do that with simple css.

mdo:
Yeah, after opening this and not getting much attention (only been under a week I realize), I think we'll pass on it for now. Thanks for calling it out @ysds.

frederikhors:
@mdo I think it is useful like with tables. In this case also on tables is easy to do with simple css, right?

PR #27873, opened 2018-12-18
Fix incorrect responsive alignment of dropdown menu

ysds:
It is incorrect to define left and right within separate `@media`.

The following demo shows incorrect alignment. The dropdown menu should be right aligned when large screen.

demo: https://codepen.io/anon/pen/WLGjGz

To fix the priority problem, it must be defined at the same breakpoint like a text utility.

https://github.com/twbs/bootstrap/blob/12c4d3c248b49f1730c75fcf8fbcb4a97274ccf6/scss/utilities/_text.scss#L18-L26


MartijnCuppens (approved):
This not only fixes the problem, but also combines multiple classes in a media query which saves some css. Thanks a lot, @ysds!

mdo (approved):

PR #27878, opened 2018-12-18
Add dark carousel color scheme

### LINKED ISSUE #25360: Carousel has no light theme (with dark arrows/indicators)
Carousel has no theme where the control arrows aren't white. In this case, using a bg-light (or if someone used a bg-white) it should have black or dark grey left/right controls. Right now it's white-on-light or white-on-white and so the arrows are useless.

```html
<div id="carouselReviews" class="carousel slide bg-light " data-ride="carousel">
  <div class="carousel-inner bg-light">
    <div class="carousel-item active">
        <div class="d-block w-100 bg-light">CONTENT</div>
    </div>
    <div class="carousel-item">
        <div class="d-block w-100 bg-light">CONTENT</div>
    </div>
    <div class="carousel-item">
        <div class="d-block w-100 bg-light">CONTENT</div>
    </div>
  <a class="carousel-control-prev" href="#carouselReviewsControls" role="button" data-slide="prev">
    <span class="carousel-control-prev-icon" aria-hidden="true"></span>
    <span class="sr-only">Previous</span>
  </a>
  <a class="carousel-control-next" href="#carouselReviewsControls" role="button" data-slide="next">
    <span class="carousel-control-next-icon" aria-hidden="true"></span>
    <span class="sr-only">Next</span>
  </a>
</div>
```

mdo:
Fixes #25360.

Not 100% if I want this added yet, but wanted to see how it looked and what the reception was for folks. Docs and example can be seen at <https://deploy-preview-27878--twbs-bootstrap4.netlify.com/docs/4.1/components/carousel/#dark-carousel>.

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -196,3 +196,18 @@
   color: $carousel-caption-color;
   text-align: center;
 }
+
+
+// Dark carousel
+//
+// Useful for when your carousel images are very light and white text doesn't
+// have enough contrast.
+
+.carousel-dark {
+  .carousel-control-prev,
+  .carousel-control-next {
+    color: $carousel-dark-control-color;
+  }
+  .carousel-indicators li { background-color: $carousel-dark-indicator-active-bg; }
+  .carousel-caption { color: $carousel-dark-caption-color; }
+}
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -208,6 +208,16 @@
   .carousel-control-next {
     color: $carousel-dark-control-color;
   }
-  .carousel-indicators li { background-color: $carousel-dark-indicator-active-bg; }
-  .carousel-caption { color: $carousel-dark-caption-color; }
+  .carousel-control-prev-icon {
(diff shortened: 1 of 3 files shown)


MartijnCuppens:
I would prefer a solution where you can use the `.text-*` classes to control the text-color and use `currentColor` to fill the indicators.

I also don't like adding more base64 SVGs in our css, but I'm afraid that 'll be something unavoidable if we don't want to introduce breaking changes.

mdo:
Agreed on all points. We don't have to add this to v4, though. Perhaps we can it and wait for v5 when we can revamp for `currentColor` and SVGs that don't violate CSP by being base64-ed?

XhmikosR:
They are not Base64 encoded BTW. Just SVGs inlined in CSS.

XhmikosR:
But yeah so a solution with `currentColor` sounds the best, although wouldn't that still require inlining?

mdo:
Closing for a v5 variation of this (which would put the SVGs back into the HTML and allow use of inheritance and `currentColor`).

PR #27894, opened 2018-12-21
Update nuspec and build script

supergibbs:
Update nuspec and build script to support `<license />` instead of deprecated `<licenseUrl />`


XhmikosR (commented):
There's no LICENSE.txt in the repo.

nuget/MyGet.ps1
```diff
@@ -4,6 +4,8 @@
 
 $nuget = $env:NuGet
 
+Copy-Item $env:SourcesPath\LICENSE $env:SourcesPath\LICENSE.txt #has to be .txt extension, don't check in
```
supergibbs: Which is why I added this to the package script. NuGet requires the extension.

XhmikosR:
Duh, missed that.

XhmikosR (approved):

PR #27898, opened 2018-12-22
versions: sort from newer to older.

XhmikosR:
Not sure which solution I like better, so please chime in.


mdo (approved):

XhmikosR:
@MartijnCuppens @Johann-S: what do you guys think?

Johann-S:
better with your changes 👍

MartijnCuppens:
Both versions are fine by me

PR #27899, opened 2018-12-22
Use https on gh-pages

XhmikosR:
Maybe I should revert the snippet changes? I also need to see if there are any more cases.


1.0.0/index.html
```diff
@@ -9,7 +9,7 @@
     <link href="assets/css/docs.css" rel="stylesheet" type="text/css">
 
     <!-- Le javascript -->
-    <script src="http://code.jquery.com/jquery-1.5.2.min.js"></script>
+    <script src="https://code.jquery.com/jquery-1.5.2.min.js"></script>
     <script src="http://autobahn.tablesorter.com/jquery.tablesorter.min.js"></script>
```
XhmikosR: Need a new URL for this @mdo

mdo: Unsure what version we were using here, but here's a new URL:

```
https://cdnjs.cloudflare.com/ajax/libs/jquery.tablesorter/2.31.1/js/jquery.tablesorter.min.js
```

XhmikosR: That I could find too myself :P

The problem is that I don't know which version we were using back then and if this one works at all. It's only used in the 1.x docs though.

PR #27883, opened 2018-12-19
Use border-radius() mixin for toast

ysds:

PR #27900, opened 2018-12-22
Blacklist border-radius property (use mixin instead)

MartijnCuppens:
Continuing on https://github.com/twbs/bootstrap/pull/27883#issuecomment-449481523. Blacklist the border radius property in stylelint, so we don't accidentally add this like happend with the toasts.


scss/_card.scss
```diff
@@ -200,10 +200,12 @@
 
           .card-img-top,
           .card-header {
+            // stylelint-disable-next-line property-blacklist
             border-top-right-radius: 0;
```
MartijnCuppens: Not sure if we should add a mixin for `border-top-right-radius` and the other specific border radii?

scss/_custom-forms.scss
```diff
@@ -144,6 +144,7 @@
 
 .custom-radio {
   .custom-control-label::before {
+    // stylelint-disable-next-line property-blacklist
```
MartijnCuppens: I guess the custom radios should always be round, independent of `$enable-rounded`? That's why I didn't use the mixin here.

scss/_custom-forms.scss
```diff
@@ -220,11 +223,7 @@
   background: $custom-select-background;
   background-color: $custom-select-bg;
   border: $custom-select-border-width solid $custom-select-border-color;
-  @if $enable-rounded {
-    border-radius: $custom-select-border-radius;
-  } @else {
-    border-radius: 0;
-  }
+  @include border-radius($custom-select-border-radius, 0);
```
MartijnCuppens: I didn't really like the `if/else` statement here, so I tweaked the `border-radius` mixin a bit to simplify this code here.

scss/_card.scss
mdo: Eh, probably not necessary? It'd save on a bit of CSS when folks compile without rounded corners enabled, but probably not worth it right now.

mdo (approved):

PR #27901, opened 2018-12-22
Simplify card group css

MartijnCuppens:
The `.card-group` css can be simplified a lot.


mdo (approved):
Looks good on Netlify!

PR #27902, opened 2018-12-22
Add two new meta tags for Algolia

XhmikosR:
@s-pace: here it is. Can you tell me what we need to change in our JS file with these in place?

Note that `docs_version` isn't our full version since usually we only need to change paths for major/minor versions, see https://github.com/twbs/bootstrap/blob/v4-dev/_config.yml#L38

### DIFF
--- a/site/_includes/header.html
+++ b/site/_includes/header.html
@@ -4,6 +4,9 @@
 <meta name="author" content="{{ site.authors }}">
 <meta name="generator" content="Jekyll v{{ jekyll.version }}">
 
+<meta name="docsearch:language" content="en">
+<meta name="docsearch:version" content="{{ site.docs_version }}">
+
 <title>
   {%- if page.title -%}
     {{ page.title | smartify }} · {{ site.title | smartify }}


s-pace:
This is perfect, we will not crawl the v3 since they have another dedicated config and that the layout is not the same.

PR pending ^

XhmikosR:
So @s-pace, do we still need to use the `facetFilters` and pass the version there too?

s-pace:
Yes you do @XhmikosR

XhmikosR:
So, how is this gonna work, @s-pace? Say we release 4.3 and we change paths again or not. Will your tool grab this automatically?

XhmikosR:
Thanks for the help @s-pace!

We'll test this after you merge your PR in https://twb-bootstrap4.netlify.com/

mdo (approved):

s-pace:
Nice, please ping https://github.com/algolia/docsearch-configs/pull/684 once you are live. 🙏

XhmikosR:
Will do, thanks!

PR #27917, opened 2018-12-23
Configurable display utility classes

MartijnCuppens:
With this PR you'll be able to control which responsive display classes are generated.


mdo:
Want to take a stab at a quick note in our docs? Can probably happen in the display utilities docs, but unsure if we need/want anything in our theming page.

MartijnCuppens:
Sure, I've just added this after the values in the docs: https://deploy-preview-27917--twbs-bootstrap4.netlify.com/docs/4.2/utilities/display/#notation

mdo (approved):

PR #27932, opened 2018-12-27
Prevent select word wrap in Safari

### LINKED ISSUE #24990: Select wrap bug

MartijnCuppens:
Fixes #24990.
Closes #25063

#25063 only fixes this for `<select>`s in `.card`s, but this behaviour should never occur, so it's better to deal with this in reboot.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -331,6 +331,14 @@ select {
   text-transform: none; // Remove the inheritance of text transform in Firefox
 }
 
+// Remove the inheritance of word-wrap in Safari.
+//
+// Details at https://github.com/twbs/bootstrap/issues/24990
+select {
+  word-wrap: normal;
+}
+
+
 // 1. Prevent a WebKit bug where (2) destroys native `audio` and `video`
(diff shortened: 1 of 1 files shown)


XhmikosR:
Does this supersede #25063?

MartijnCuppens:
Yeah, #25063 will cause issues when words are too long in cards.

XhmikosR:
Let's close it then when this lands.

mdo (approved):

Issue #27928, opened 2018-12-27
gh-pages: find a way to include sub-sitemaps


XhmikosR:
Probably a manual step for the existent old docs, and semi-manual for each new version.

XhmikosR:
Actually, I don't think we need this. Quite the opposite. What we do now, is provide a sitemap.xml with only the latest doc pages.

PR #27940, opened 2018-12-28
More link fixes.

XhmikosR:


XhmikosR:
@mdo: I added a couple more patches to help us with #27928. For the older versions, I will try to find a solution later.

mdo (approved):

PR #27941, opened 2018-12-28
Match input font size for .input-group-text

MartijnCuppens:
If `$input-font-size` is changed, the font size of input groups with `.input-group-text` won't change.


ysds:
LGTM

XhmikosR (approved):

mdo (approved):

PR #27963, opened 2019-01-02
Sync custom select focus with form control focus

### LINKED ISSUE #27944: Focus color mismatched between .custom-select and .form-control
form-control on focus is
    box-shadow: 0 0 0 0.2rem rgba(0,123,255,.25);

why custom-select on focus?
    box-shadow: 0 0 0 0.2rem rgba(128,189,255,.5);

MartijnCuppens:
Fixes #27944.


mdo (approved):

mdo:
Tests fail because `Variable $custom-select-focus-width is not being used!`

MartijnCuppens:
I deleted the variable. Don't think that would break anything since this variable was used to construct `$custom-select-focus-box-shadow`.

mdo:
Hmm, don't think we can delete variables in v4 (backward compatibility). Maybe first deprecate them by reassigning their values? Unsure about the unused vars linter and how to whitelist something, if we can at all.

MartijnCuppens:
I've tweaked the variable so that it makes use of `$custom-select-focus-width`. @mdo, what do you think of this?

mdo (approved):

PR #27970, opened 2019-01-04
Add horizontal list group modifier

### LINKED ISSUE #27291: Horizontal list group feature suggestion
I've been playing around the list group component recently in some projects. One thing I'd like to see in **Bootstrap 4.2** is possibly to have a **horizontal list group** or the ability to make any of the list groups provided (https://getbootstrap.com/docs/4.1/components/list-group/#basic-example) horizontal by adding a class. Technically, I don't know if it would be too difficuilt to add. I was playing around in the inspector and changed two things:

```
.list-group {
    // originally flex column
    flex-direction: row;
}

.list-group-item {
    // originally margin-bottom
    margin-left: -1px;
}
```

By making these relevant changes, you'd get the following:

list-group-horizontal

And if we remove the `max-width` from the documentation:

list-group-horizontal-100

I was thinking we could have a modifier class to the `list-group-*` something like:

`-horizontal`
`-vertical`

Thoughts?

mdo:
Alternate take to #27683 that builds the modifier class based on min-width, such that list groups are stacked until the class's breakpoint is reached. This is less verbose, matches our primary responsive approach, and prevents an extra class. Unfortunately, I believe supporting flush list groups is too much code here, so I've skipped that and made a note in the docs.

Also added examples generate from our breakpoints data file with a protip for how to do equal width items.

Closes #27683, fixes #27291.

/cc @sts-ryan-holton


sts-ryan-holton:
@mdo  Great news! This looks good. I think that horizontal list groups will certainly get some use, looking forward to this new change in 4.3 :+1:

site/docs/4.2/assets/scss/_component-examples.scss
```diff
@@ -243,7 +243,7 @@
 }
 
 // List groups
-.bd-example > .list-group {
+.bd-example > .list-group:not([class*="horizontal"]) {
```
MartijnCuppens: This is a no-go for me, attribute selectors like these are way less performant than the universal selector and are not strict enough because it could potentially target other classes that contain horizontal

MartijnCuppens (dismissed):

site/docs/4.2/assets/scss/_component-examples.scss
mdo: Usually I'm in the same boat, but this being a docs-only thing, I was fine with it. I can easily rewrite it though to override the previous selector.

MartijnCuppens: Ow, didn't notice that. In that case: nevermind

XhmikosR (approved):

scss/_list-group.scss
```diff
@@ -80,6 +80,39 @@
...
 
+// Horizontal
+//
+// Change the layout of list group items from vertical (default) to horizontal.
+
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    @each $value in $displays {
```
okazou: Remove this `each` loop is better.

XhmikosR: #28005

PR #27972, opened 2019-01-04
take account 'none' for box-shadow

### LINKED ISSUE #26478: button transition with $enable-shadows:true not working properly
see: http://jsfiddle.net/aze0jzj5/6/

steps to reproduce: 
click and hold button for 1sec and then release mouse button

result:
focus outline flickers

This is because BS tries transition from box-shadow: inset to box-shadow: outer

According to https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_buttons.scss#L89
you can disable `inset` by setting `$btn-active-box-shadow:none` but that will not work because in other places passing 'none' to box-shadow causes incorrect css rule (for example` box-shadow: none, 1px 1px 1px black`)

Solution that can be worth consider:
http://tobiasahlin.com/blog/how-to-animate-box-shadow/

wojtask9:
Fixes #26478


MartijnCuppens (approved):
I've just tested this and it LGTM. I was a bit worried the mixin couldn't handle multiple values because sass lists can be separated by as well spaces as commas (https://hugogiraudel.com/2013/07/15/understanding-sass-lists/#sass-list-fun-facts), but that doesn't seemed to be a problem

danijelGombac:
@MartijnCuppens in scss/_buttons.scss

```
&.disabled,
  &:disabled {
    @include box-shadow(none); // this is not take into account
}
```

wojtask9:
indeed. I missed that in my PR.

there is another place where it breaks
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_button-group.scss#L102

Possible solutnions:
1. change this to box-shadow(unset)
2. change mixin `box-shadow`  in that way if there is only one `none` shadow then apply to box-shadow property

PR #27976, opened 2019-01-04
Running documentation in Docker

vochicong:
Use Docker to easily run the documentation locally.


_config.yml
```diff
@@ -11,7 +11,7 @@ permalink:      pretty
 # Server
 source:         "site"
 destination:    ./_gh_pages
-host:           "localhost"
+host:           "0.0.0.0"
```
XhmikosR: This was intentional.

XhmikosR:
I see no value for this, especially since everyone can do it themselves.

vochicong:
Everyone can do it, but it takes each about 10 minutes or more.

_config.yml
vochicong: `localhost` won't work for someone using Docker.

mdo:
When none of the development of Bootstrap happens with Docker currently, why would we add this?

Issue #27927, opened 2018-12-27
4.2 breaking change: custom $grid-breakpoints that are smaller than default ones give errors


phazei:
I'm unable to compile sass with the new $grid-breakpoint changes.
This worked just fine with 4.1.  It now doesn't seem possible with 4.2

In my custom variable file that's included before everything BS4 I have the following:

```sass
$grid-breakpoints: (
        xxs: 0,
        xs: 480px,
        sm: 576px,
        md: 768px,
        lg: 992px,
        xl: 1200px
);
```

As you can see, an additional custom breakpoint exists at 480px.

Previously 4.1 `_variable.sass` contained:
```sass
$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;

@include _assert-ascending($grid-breakpoints, "$grid-breakpoints");
@include _assert-starts-at-zero($grid-breakpoints);
```

And my code overrode all the breakpoints and worked fine.

Now in 4.2 _variable.sass:
```sass
$grid-breakpoints: () !default;
// stylelint-disable-next-line scss/dollar-variable-default
$grid-breakpoints: map-merge(
                (
                        xs: 0,
                        sm: 576px,
                        md: 768px,
                        lg: 992px,
                        xl: 1200px
                ),
                $grid-breakpoints
);

@include _assert-ascending($grid-breakpoints, "$grid-breakpoints");
@include _assert-starts-at-zero($grid-breakpoints);
```

And it gives the following errors:

```
WARNING: Invalid value for $grid-breakpoints: This map must be in ascending order, but key 'xxs' has value 0 which isn't greater than 1200px, the value of the previous key 'xl' !

WARNING: First breakpoint in `$grid-breakpoints` must start at 0, but starts at 480px.

WARNING: Invalid value for $container-max-widths: This map must be in ascending order, but key 'xs' has value 500px which isn't greater than 1140px, the value of the previous key 'xl' !
```

phazei:
```javascript
/**
 * Merges two maps together into a new map. Keys in `$map2` will take
 * precedence over keys in `$map1`.
 * 
 * This is the best way to add new values to a map.
 * 
 * All keys in the returned map that also appear in `$map1` will have the
 * same order as in `$map1`. New keys from `$map2` will be placed at the end
 * of the map.
 * 
 * Like all map functions, `map-merge()` returns a new map rather than
 * modifying its arguments in place.
 * 
 * @example
[...]

PR #27977, opened 2019-01-04
Added ability override breakpoints/container-width

ichabanivan:
If developer need to remove a breakpoint (for example: to make grid without xl breakpoint), he cannot overide it from another file, if we add !default in variables file everyting will be ok!

Next code will get xl from _variables.scss
(
    sm: 540px,
    md: 720px,
    lg: 960px,
)


mdo:
This isn't the right fix for the problem reported in #27927. Thanks though!

PR #27982, opened 2019-01-05
WIP: Use husky to execute linters

### LINKED ISSUE #27973: Add Husky
I've been using [husky](https://github.com/typicode/husky) lately to run stylelint (with `--fix` option) before things are committed. Adding this to Bootstrap would prevent a lot of failing tests for our PRs (for example the properties that aren't in correct order or trailing spaces). Maybe it could be interesting to add this?

MartijnCuppens:
Closes #27973

The linting taks are now ran before anything is pushed which will help us with builds that fail because of trivial lint errors.

I preferred the `pre-push` option above the `pre-commit` option because all linting tasks take about 9 seconds to complete which could be annoying.

This PR also contains a commit which parallelizes the linting tasks. This decreases the time from around 9 seconds to 4-5 seconds.

Not sure if we should add the `--fix` command which automatically fixes fixable linting errors (like fixing the property order). Maybe we could add an additional command for it?

TODO:

- [ ]  Add documentation once we've decided how we're going to implement it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4861,6 +4861,95 @@
       "integrity": "sha512-9em7CXFa0my1DF3aIQg0sTRyAX2znEOMHolUvu9nSTUjS+bRD32y0MH+Hnm3Xu0cSWrxpYb2isXSfH9pF2LP8g==",
       "dev": true
     },
+    "husky": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/husky/-/husky-1.3.1.tgz",
+      "integrity": "sha512-86U6sVVVf4b5NYSZ0yvv88dRgBSSXXmHaiq5pP4KDj5JVzdwKgBjEtUPOm8hcoytezFwbU+7gotXNhpHdystlg==",
+      "dev": true,
+      "requires": {
+        "cosmiconfig": "^5.0.7",
+        "execa": "^1.0.0",
+        "find-up": "^3.0.0",
+        "get-stdin": "^6.0.0",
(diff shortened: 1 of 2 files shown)


Johann-S:
It's a bit slow 😟 do you know if there is a way to skip that on push ?

MartijnCuppens:
https://github.com/twbs/bootstrap/pull/27989 will increase speed to around `1.8s` (for me) which is acceptable for me as `pre-push`, don't know what you guys think? Imo it's worth it because figuring out what's wrong with some one else's PR takes a lot more time.

`--no-verify` can be used to skip the tests.

XhmikosR:
Or we could just leave things as is, and you can add the hook manually.

XhmikosR:
The thing is, many people don't even check out the repo locally and even if they do they don't run the test scripts.

Secondly, we do have CI for this anyway.

I'm not against it as long as it doesn't disrupt my current workflow, and as it seems it does since it slows us down. The dep I don't mind, it's a small one anyway.

Also, IMO this should be in precommit if we ever go with it. You could try calling the individual css and js lint scripts if that's faster instead of the wrapper `lint` script.

XhmikosR:
That's not a rebase.

MartijnCuppens:
Fixed it

XhmikosR:
You missed the update package-lock.json.

XhmikosR:
OK, I tried this again, and it's just too slow for me...

MartijnCuppens:
Ok fine, I'll close it than

PR #27983, opened 2019-01-05
Run linters in parallel and tweak stylelint scripts

MartijnCuppens:
This should speed up things a bit


XhmikosR:
Seems to be OK so far.

XhmikosR (approved):

PR #27985, opened 2019-01-05
Remove screen from media query

MartijnCuppens:
We don't use `screen` anywhere else in our media queries and there's no reason why we should disable it if the device is not a screen.


andresgalante (approved):
True, and there is no motion on a screen reader or a paper :)

mdo (approved):

PR #27989, opened 2019-01-06
Cache ESLint and stylelint

XhmikosR:
Generally it speed things up a lot. There are some cases like when changing rules that one needs to remove the cache, but other than that it should work fine.

Please test it out with your normal workflow.


MartijnCuppens (approved):
Looks ok for stylelint (didn't test `eslint` too much)

Johann-S:
it seems faster to me 👍

XhmikosR:
We'll see how this goes in the long term.

PR #27995, opened 2019-01-07
Make dropdown font size configurable

MartijnCuppens:
Dropdowns are mostly used on buttons and it would make sense if their font sizes sync.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -738,6 +738,7 @@ $navbar-dark-brand-hover-color:           $navbar-dark-active-color !default;
 $dropdown-min-width:                10rem !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
+$dropdown-font-size:                $btn-font-size !default;
 $dropdown-bg:                       $white !default;
 $dropdown-border-color:             rgba($black, .15) !default;
 $dropdown-border-radius:            $border-radius !default;
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -738,7 +738,7 @@ $navbar-dark-brand-hover-color:           $navbar-dark-active-color !default;
 $dropdown-min-width:                10rem !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
-$dropdown-font-size:                $btn-font-size !default;
+$dropdown-font-size:                $font-size-base !default;
 $dropdown-bg:                       $white !default;
 $dropdown-border-color:             rgba($black, .15) !default;
 $dropdown-border-radius:            $border-radius !default;
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -22,7 +22,7 @@
   min-width: $dropdown-min-width;
   padding: $dropdown-padding-y 0;
   margin: $dropdown-spacer 0 0; // override default ul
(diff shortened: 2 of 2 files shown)


scss/_variables.scss
```diff
@@ -738,6 +738,7 @@ $navbar-dark-brand-hover-color:           $navbar-dark-active-color !default;
 $dropdown-min-width:                10rem !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
+$dropdown-font-size:                $btn-font-size !default;
```
ysds: Should use `$font-size-base` because this font size is used not only for dropdown item button(link) but also text and forms: https://getbootstrap.com/docs/4.2/components/dropdowns/#text

ysds:
As I commented above, the font size on `.dropdown-menu` is confusing. For better understanding, should add a font size to `.dropdown-item`?

mdo:
Dropdowns are most used with buttons and nav links. Do we really need to connect the variables here, or can we rely on inheritance?

ysds:
I agree that no need a variable for the font size of `.dropdown-menu`.

MartijnCuppens:
The problem I have with a project I'm currently working on is that I've decreased my `$btn-font-size` to `.9375rem` and my `$base-font-size` is still `1rem`.

The links in the dropdown menu have a font size of `1rem`, which looks a bit weird, because they are bigger than the dropdown button.

This change would make the dropdown font size with the button font sizes.

As you both pointed out, the dropdown is also used in other contexts, so I've updated the PR to make the font size configurable. This wouldn't change anything by default (and will not introduce any breaking changes) and will give people the ability to change the value to something else like `inherit` if they prefer inheritance of the font sizes. What do you think?

ysds:
IMO:

```diff
  .dropdown-menu {
    font-size: $font-size-base;  // Use $font-size-base as it is
  }
  
  .dropdown-item {
+   font-size: $dropdown-item-font-size;  // Add $dropdown-item-font-size: inherit !default;
  }
```

ysds:
And also:
```diff
  .dropdown-header {
-   font-size: $font-size-sm;
+   font-size: $dropdown-header-font-size;
  }
```

MartijnCuppens:
I wouldn't go that far with setting the font sizes. If we go that way we open the door to setting the font size of every component element.

In `v5` we'll probably have a look if we can use utility classes to resize text.

mdo:
Stepping out of this particular PR's diff right now, I've always hated us prematurely adding `font-size`s to components that should likely inherit from the `<body>`/`<html>`. Out of the box, Bootstrap doesn't need the majority of these defaults, so it actually adds bloat to our codebase. But, it's the best way for folks to customize the source code.

I've started wondering if there's a way we can prevent some default CSS from being added until someone changes a default value. Would probably be too gnarly, but could reduce some file size for us.

jackmu95:
@mdo What about a `font-size` mixin that checks wether the given value is `inherit` or any other value and only adds the `font-size` property when the value is not `inherit`?

MartijnCuppens:
@jackmu95, we could do that, but `font-size` won't be the only property we could make optional. We'll need to find a general simple solution. I've created #28024 to drop any ideas.

@mdo, does the latest changes look good to you? (They 'll just make it possible to tweak the default font size of the dropdown)

mdo (approved):

Issue #25206, opened 2018-01-05
[v4] Modal flickering when within .list-group-item


gizburdt:
Hi,

In some cases the modal filckers (see video).

![bootstrap](https://user-images.githubusercontent.com/1470623/34611050-36f855bc-f224-11e7-96ba-a8315a4a601a.gif)

This is the html:
```html
<div class="row align-items-center">
    <div class="col-md-6">
        <strong>Test</strong> <br>
        Street 1 31 <span><span class="badge badge-info">Primary</span></span> <br>
        1234AA Amsterdam <br>
    </div> 

    <div class="col-md-6 text-right"><!----> 
        <div class="d-inline-block text-left">
            <a href="#" class="btn btn-sm btn-warning">Edit</a> 
            
            <div tabindex="-1" role="dialog" id="editModalLHzoKhSu" class="modal fade">
                <div role="document" class="modal-dialog">
                    <form class="modal-content">
                        <div class="modal-header"><p id="editAddressModalLabel" class="modal-title h5">Edit address</p> <button type="button" data-dismiss="modal" aria-label="Close" class="close"><span aria-hidden="true">×</span></button>
                        </div> 

                        <div class="modal-body">
                            // Form
                        </div>
                    </form>
                </div>
            </div>
        </div>
    </div>
</div>
```

If you need more info, please let me know :)

- OS: MacOS,
- Browser: Chrome 63.0.3239.84

Ruffio:
@gizburdt do you have some custom css/js that can cause this? Can you make a jsbin or similar that demonstrates this behaviour?

Johann-S:
Bug reports must include a live demo of the problem. Per our [contributing guidelines](https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md), please create a reduced test case via [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/) and report back with your link, Bootstrap version, and specific browser and OS details.

gizburdt:
Hmm, will have a look at it. The code is nested in my project (and Vue), so I will have a try moving it to JSbin.

Extra info: The error is not present in 4.0.0-beta.2, but it is in 4.0.0-beta.3

gizburdt:
Here is the JS bin: https://jsbin.com/sitemohuno/2/edit?html,js,output

The problem happens when the modal is within a list-group-item. Normally I place my modals at the bottom of the body, but I can't in my current project, because i'm using vue components.

[...]

PR #27996, opened 2019-01-07
Move z-index from .list-group-item to .list-group-item-action

### LINKED ISSUE #25446: list-group with dropdown and input-group with button -> button displayed over opened dropdown
bootstrap-bug

I have some code like the following example. when i open the dropdown, the button of the input-group in the next list-group-item ist showed over the dropdown, while the dropdown-menu is hovered. When the mouse is outside the dropdown-menu, all is as normal... 

    <div class="col-2">
        <ul class="list-group">
            <li class="list-group-item">
                <div class="dropdown">
                    <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                        Dropdown button
                    </button>
                    <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
                        <a class="dropdown-item" href="#">Action</a>
                        <a class="dropdown-item" href="#">Another action</a>
                        <a class="dropdown-item" href="#">Something else here</a>
                    </div>
                </div>
            </li>
            <li class="list-group-item">
                <div class="input-group">
                    <input type="text" class="form-control" placeholder="Recipient's username" aria-label="Recipient's username" aria-describedby="basic-addon2">
                        <button class="btn btn-cubeBlue" type="button">Button</button>
                    </div>
                </div>
            </li>
        </ul>
    </div>

ysds:
It is more make sense if `z-index` is defined to `.list-group-item-action` not `.list-group-item`.

And remove redundant `text-decoration: none;`

This fixes #25446, hopefully also #25206.


mdo:
`z-index: 1` is there to ensure that each subsequent list group item lays above the previous—that way the borders overlap as intended. Probably fine like this, too, but we should confirm no adverse affects to the standard items.

ysds:
Before: https://codepen.io/anon/pen/pZdMNM
After: https://codepen.io/anon/pen/jXxjOY

Based on my experience working on z-index of bootstrap, seems fine to me I think, but would like feedback and more test case.

MartijnCuppens (approved):
Couldn't find anything that could go wrong.

mdo (approved):

PR #28004, opened 2019-01-08
box-shadow mixin fix

wojtask9:
My previous fix was to optimistic. Please look at https://github.com/twbs/bootstrap/pull/27972#issuecomment-452106061

Sometimes we want pass `none` to switch off `box-shadow`.
We don't want only properties like `box-shadow: none, 1px 1px black;`  because this is invalid property for browsers

### DIFF
--- a/scss/mixins/_box-shadow.scss
+++ b/scss/mixins/_box-shadow.scss
@@ -2,9 +2,15 @@
   @if $enable-shadows {
     $result: ();
 
-    @for $i from 1 through length($shadow) {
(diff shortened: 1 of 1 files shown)


XhmikosR:
This has unrelated patches. `git rebase -i upstream/v4-dev` is your friend.

wojtask9:
@XhmikosR done. Sorry for that

MartijnCuppens (approved):
The comments are also a great addition, thanks for fixing this so fast, @wojtask9!

PR #28005, opened 2019-01-08
remove redundant loop from list-group-horizontal

wojtask9:
This loop is unnecessary. It generates always the same output.


MartijnCuppens (approved):
LGTM. Keep up the good work, @wojtask9! Really appreciated!

wojtask9:
@MartijnCuppens thank you very much

mdo (approved):

PR #28008, opened 2019-01-08
Cleanup popover arrows

MartijnCuppens:
I moved and copied the `.arrow`'s pseudos shared `border-width:` to the pseudo elements themselves. This will generate 4 more properties (oh no 😱), but will remove 16 selectors of which 8 attribute selectors (woohoo! 🎉). As you can see, this made my day.

I've also nested the pseudo's in the `scss`.


mdo:
Would we want to wait on this for v5 since this breaks anything for folks who've customized Bootstrap's default popover template? I don't know how worth it is to do now for v4.3 knowing it could be funky for folks. At the very least, it changes where the arrows are generated and folks who remove them via CSS would have a breaking change for their upgrade path.

mdo:
Also, whenever we do want this to land, we can probably modify the JS to update that template to remove the extraneous `<div>`.

ysds:
@mdo Don't worry:wink: The CSS diffs are very simple. I don't think this will cause a breaking change.

```diff
- .bs-popover-top > .arrow::before, .bs-popover-auto[x-placement^="top"] > .arrow::before,
- .bs-popover-top > .arrow::after,
- .bs-popover-auto[x-placement^="top"] > .arrow::after {
-   border-width: 0.5rem 0.5rem 0;
- }

  .bs-popover-top > .arrow::before, .bs-popover-auto[x-placement^="top"] > .arrow::before {
    bottom: 0;
+   border-width: 0.5rem 0.5rem 0;
    border-top-color: rgba(0, 0, 0, 0.25);
  }

  .bs-popover-top > .arrow::after,
  .bs-popover-auto[x-placement^="top"] > .arrow::after {
    bottom: 1px;
+   border-width: 0.5rem 0.5rem 0;
    border-top-color: #fff;
  }
```

x 4 directions

Johann-S:
it was my CSS sorry 😟
@MartijnCuppens know very well my current level of CSS 😆

mdo (approved):

PR #28011, opened 2019-01-09
Add new rounded sizes classes

### LINKED ISSUE #27934: New utility classes: `rounded-lg` and `rounded-sm`
Would be nice to have utility classes like `rounded-lg` and `rounded-sm` as you already have SCSS variables for `border-radius-lg` and `border-radius-sm`.

jt143:
- Add `.rounded-sm` and `.rounded-lg`

Fixes #27934 

Demo preview: https://deploy-preview-28011--twbs-bootstrap4.netlify.com/docs/4.2/utilities/borders/#sizes


XhmikosR (commented):
Please do not touch any dist files.

XhmikosR:
Also, there's a good reason for using `!important` in utilities.

jt143:
updated!

MartijnCuppens:
Thanks for the PR, @jt143!

These classes have `!important` because they are utility classes. The classes just do one thing and shouldn't depend on other classes.

Therefor I would go for `.rounded-sm` classes instead of `.rounded.rounded-sm` classes (same for `.rounded-lg`). And also `.rounded-top-sm` classes instead of `.rounded-top.rounded-sm` (and same behaviour for other variants).

mdo (changes requested):
We wouldn't chain these new classes to the side classes like this. Utilities should never ben extended, chained, or modified. For now, I'd suggest we do `.rounded-sm` and `.rounded-lg` only.

jt143:
Hi @mdo and @MartijnCuppens 
Thank you for your reviews! I have updated as requested.

MartijnCuppens (approved):
LGTM :+1:

mdo (approved):
Looks good!

PR #28012, opened 2019-01-09
Tighten bundlesize limits.

### LINKED ISSUE #28007: Make bundlesize limits tighter

XhmikosR:
Fixes #28007

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,11 +174,11 @@
     },
     {
       "path": "./dist/css/bootstrap.css",
-      "maxSize": "25.5 kB"
+      "maxSize": "25 kB"
     },
     {
       "path": "./dist/css/bootstrap.min.css",
-      "maxSize": "23.5 kB"
+      "maxSize": "23 kB"
     },
     {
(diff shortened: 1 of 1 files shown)


XhmikosR:
We could go even tighter, not sure though.

Johann-S (approved):

MartijnCuppens (approved):
These limits should be ok, decreasing them further would probably cause failing tests when any changes are done.

XhmikosR:
See, the thing is that I'd rather have build failures and us bump the limit then than get a change that's causing a .5KB change like it recently happened. Remember, this is gzip size.

I'll think about it a bit more.

mdo:
We can probably knock down Reboot's limits, too, right? We seem way under for that.

XhmikosR:
Yeah, I was thinking about that also @mdo. We could go to 2KB there.

XhmikosR:
OK, limits updated.

mdo (approved):

PR #28014, opened 2019-01-09
Switch to Hugo

### LINKED ISSUE #24475: Switch to Hugo
OK... this is probably something we might never get to do it due to the amount of changes.

That being said, [Hugo](https://gohugo.io/) is blazingly fast, has no dependency like Jekyll has to Ruby, and is very extensible too.

If anyone starts working on it, please let me know.

XhmikosR:
TODO:

- [ ] **make sure our folder structure is the proper one for what we want to achieve. Also see if we can make folders nesting level smaller, maybe by using permalinks or something. For content and static.**
- [ ] Review the new Chroma styles
- [ ] audit the changes by bep or someone familiar with Hugo

---

TODO (can be sorted out maybe after the merge)

- can we generate SRI hashes for existent files on the fly? For bootstrap.min.css, bootstrap.min.js, bootstrap.bundle.min.js, popper.js/1.14.7/umd/popper.min.js. This would simplify things as we could remove the release-sri script
- get image width/height with Hugo when possible (examples images, homepage)
- maybe use Hugo Extended version for docs' CSS?
- bep's suggestion:
  > Re.
  > ```
  > {{< partial "callout-warning-color-assistive-technologies.md" >}}
  > ```
  > The partial above uses the `partial` func to load, parse and execute templates. This has its limitations and is possibly ineffective.
  > I would recommend that you store your callouts (and other content files you want to include) in a so-called headless bundle -- and load them via `.GetPage`. This way you can use markdown + shortcodes etc.
- bep's suggestion:
  > Ref. `{{ block "header" . }}` in your `single.html` template.
  > Blocks only make sense inside base template(s) (i.e. `baseof.html`), which I cannot see.  I would highly recommend that you restructure your templates to use them. It removes some repetition.

---

<details>
<summary>Fixed issues</summary>

- [x] fix Netlify previews; fixed with the relative URLs
- [x] make sure all layouts are fine
    - [x] homepage
      - [x] Fix variables in code snippets
    - [x] examples
      - [x] index page
      - [x] individual examples are not being generated at all
    - [x] versions
      - [x] sort versions in reverse
      - [x] fix `4.0` showing up as `4`; fixed by using strings in data files
- [x] fix partial include/icons
- [x] fix docs sidebar menu; see if we can still use a data file for this
  - [x] fix single menu item (Migrate)
- [x] placeholder shortcode: backport the latest v4-dev changes
- [x] Fix `bugify`
- [x] Example shortcode
  - [x] remove the placeholder SVG etc
  - [x] Fix nested placeholder like in media object second example; the regex is too overzealous. We might need to split the string like we currently do.
  - [x] remove the extra newline when there's a placeholder image
  - [x] add default `lang` `html` for `example` shortcode; it seems we only pass HTML
  - [x] support `hide_markup` or `nohighlight` in `example` shortcode and adapt content
  - [x] Use `false` for text and title in the example shortcode instead of `" "`
- [x] fix ToC styling
- [x] whitespace control
- [x] Streamline the inline shortcodes whitespace
- [x] make sure the HTML comments we really want to keep are kept
- [x] Fix the remaining inline HTML content cases
- [x] Regenerate Chroma/Pygments styles and adapt it
- [x] Report/Work around the stray `p` tags issue
  - Reported in https://github.com/russross/blackfriday/issues/524
  - Reported in https://github.com/russross/blackfriday/issues/525
- [x] Figure out/report why https://github.com/twbs/bootstrap/pull/28014/files#diff-cadd93f34018d2674cc31bd8da855e07R352 isn't rendered as HTML
  - Reported in https://github.com/russross/blackfriday/issues/523
- [x] compare the number of generated pages so that we don't have extra/missing ones like `/docs/index.html`
- [x] **Before the merge:**
  - [x] Remove `--renderToDisk` from `docs-serve` script
  - [x] Remove any temp commits

</details>

---

Fixes #24475

**Preview:** <https://deploy-preview-28014--twbs-bootstrap.netlify.com/>


XhmikosR:
If anyone is familiar with Hugo and has some time, I could use some help here.

I really want us to make the switch, we just need some help setting things up.

BhaveshSGupta:
@XhmikosR, I have bit of experience in hugo. I can take some time out. How I can help?

XhmikosR:
@BhaveshSGupta: please see the TODO in the first post :)

I think I'm gonna drop the whole absolute URL. Currently with Jekyll, we use `site.baseurl` but this variable is an empty string so all the links are basically relative to the server root. So, dropping this seems to make sense and will make things simpler.

jhabdas:
@XhmikosR I'm working on a commit to move you over to block layouts as suggested by @bep. I expect to have a commit ready in an hour or two if that's alright by you.

XhmikosR:
@jhabdas: thanks! Any improvement is welcome :)

jhabdas:
I did a small refactoring to give you a feel for how block layouts work in the contents of your WIP. Basically, what I did here was to create a `baseof.html` file (your base layout) and then use it for the examples landing page and the pricing example.

https://github.com/jhabdas/bootstrap/commit/b6f9da3a45a82cfacf0db8a54e10fa8e1a04d575

Please look over the pattern as once it clicks you'll understand exactly how to refactor the rest of the examples to use them. If you have any questions just LMK. I'm +8:00 but happy to drop some more patterns on you to help you get this nailed down.

To see base layouts in use outside this project—if it helps conceptually—please feel free to look over my [Custom Layouts](https://after-dark.habd.as/feature/custom-layouts/) feature explaining how block templates are supposed to work then take a peek at [the code](https://git.habd.as/comfusion/after-dark/src/branch/master/layouts/index.html).

Finally, just in case you're using Atom, if your go templates code highlighting isn't working <kbd>Ctrl+L</kbd> then typeahead for _HTML (Go)_. (requires `language-go` syntax)

XhmikosR:
Huge thanks! I have some comments, I'll take it to your commit.

site/layouts/_default/examples.html
```diff
@@ -10,6 +10,10 @@
 
     <link rel="canonical" href="{{ .Permalink }}">
 
+    {{ with .Params.robots -}}
+    <meta name="robots" content="{{ . }}">
+    {{- end }}
+
```
jhabdas: Please see [my comment here](https://github.com/jekyll/jekyll-seo-tag/issues/100#issuecomment-293764634) before adding this. I never did hear back from @benbalter or @pathawks on their opinions before the issue was auto-closed. Hope you guys don't mind the ping. :)

XhmikosR: I don't mind personally using this because it will be used rarely anyway.

jhabdas:
https://github.com/jhabdas/bootstrap/commit/b87163422b73e5162321968741d8b2ef8f1820b8

What changed since last commit:

- Uses explicit path to files so hugo knows which _versioned_ files to build against (I believe you're going to want a [multi-site workflow](https://discourse.gohugo.io/t/hugo-multisite-workflow/103) to continue versioning your documentation)
- Removed all `layout: foo` from all content files (those aren't necessary with block layouts in hugo)
- Removed the `docsref` shortcode and used the inbuilt `relref` (not necessary with proper structure)
- Created base layouts for entire site (`_default/baseof.html`) as well as "examples" section (both single and the list layouts) so you can see how those inherit from base while remaining DRY.
- Did some high-level testing — this is not intended to be fully functional but to illustrate the block layout pattern from my last commit only this time in the context of a multi-site build.

Please pull the commit down and study it as it's doing the same thing as the last commit only it's a little more complex now given the versioning stuff added—which I don't recommend until the base layout is complete as mentioned in my earlier commit.

I'm out of time here so please feel free to comment on the pull after you try it out locally. To see the actual diff pull down the work rename `/sites` to `/site`. I did this because I had to delete examples to get the build passing with the extra complexity here.

Here's what you can expect after running `npm run docs-serve` on this commit:

https://jhabdas.keybase.pub/bootstrap-block-layout.mp4

Hope this helps unblock you.

XhmikosR:
@jhabdas: thanks for this, I really appreciate it! I'm having trouble reviewing this but I will try again tomorrow.

Remember, that we don't need examples in `getbootstrap.com/examples/`. Also, I still fail to get the partials for the individual examples because we need to examples as one file without partials.

jhabdas:
Thanks for taking taking a look @XhmikosR. For the diff do an `rm site && mv sites site` then amend or replace my last commit to compare to head~1. To get the examples markup into content I believe you could create a base layout overriding baseof for just that one section. And for the paths the `baseURL` could be set—which is where multi-site builds would come in.

While I was taking this last pass at layouts I was wondering...will there be a copy of all files once per release of bootstrap to allow for proper versioning of the docs in a dogfooded way, or are you thinking to abstract that into a bootstrap theme for hugo which could be used to build each version of the docs using release tags?

jhabdas:
Thanks for taking taking a look @XhmikosR. For the diff do an `rm site && mv sites site` then amend or replace my last commit to compare to head~1. To get the examples markup into content I believe you could create a base layout overriding baseof for just that one section. And for the paths the `baseURL` could be set—which is where multi-site builds would come in.

jhabdas:
Believe that was my last contribution, hoss. Sorry!
github-contributions-blocked

XhmikosR:
Not sure what happened there but it seems @jhabdas deleted his GitHub account or something...

config.yml
```diff
@@ -14,6 +14,9 @@ enableRobotsTXT:        true
 metaDataFormat:         "yaml"
 disableKinds:           ["404", "taxonomy", "taxonomyTerm", "RSS"]
 
+blackfriday:
+  fractions:            false
+
```
XhmikosR: Why is this needed?

bardiharborow: From https://deploy-preview-28014--twbs-bootstrap.netlify.com/docs/4.3/getting-started/browsers-devices/, this:
> If you require Internet Explorer 10/11 support, use Bootstrap 4.

Was rendering as:
> If you require Internet Explorer <sup>10</sup>&#8260;<sub>11</sub> support, use Bootstrap 4.

XhmikosR: Ah, nice good catch. Makes sense then 👍

XhmikosR:
@twbs/team: I think I'm gonna merge this as is, and hopefully we'll get some help from the community improving this later.

The only thing is that, the moment we merge this, backporting doc changes for v4 will be manual work, I think.

Johann-S:
It's time to merge this PR 👍 

I don't think it'll be hard to backport docs things, it's just some copy/paste here and their

Johann-S (approved):

bardiharborow (approved):
I can't vouch for everything here, but seems 95% so let's merge and get on with it. 😅

mdo (approved):

PR #28035, opened 2019-01-13
Make component colors configurable

MartijnCuppens:
I recently discovered properties aren't rendered if their value is set to `null`. That's quite interesting because this allows us to make things configurable without rendering useless code.

I used this technique to make it possible to change the default colors of the Bootstrap components.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -746,6 +746,7 @@ $dropdown-min-width:                10rem !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
 $dropdown-font-size:                $font-size-base !default;
+$dropdown-color:                    $body-color !default;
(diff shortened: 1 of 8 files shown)


scss/_variables.scss
```diff
@@ -755,6 +755,7 @@ $navbar-dark-brand-hover-color:           $navbar-dark-active-color !default;
 $dropdown-min-width:                10rem !default;
 $dropdown-padding-y:                .5rem !default;
 $dropdown-spacer:                   .125rem !default;
+$dropdown-color:                    $body-color !default;
```
MartijnCuppens: `$body-color` was already set on the dropdown component, so I moved to the config.

mdo:
@MartijnCuppens Can you double check conflicts? <3

mdo (approved):
Thank you!

PR #28036, opened 2019-01-13
Prevent useless property rendering

MartijnCuppens:
`font-family` and `color` of headings shouldn't be rendered by default.


mdo (approved):

mdo:
This was an awesome discovery. Hoping we can clear out more over time :).

PR #28043, opened 2019-01-13
Combine background properties into one

MartijnCuppens:


mdo (approved):
Generally I don't like unnecessary shorthand properties as it's a clearer indication of base and modifier classes working together vs overriding, but this doesn't immediately imply overriding anything else I think.

PR #28052, opened 2019-01-14
custom switches - add a right aligned option?

Lausselloic:
I propose to add an option to new custom switches component `right` class to position label before the switch for rtl website or any other needs.

Here's what it looks like : 
image

Don't know if it's better to align also the label on the right and keep the gutter to align?
Let me know or edit that PR if needed

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -188,6 +188,22 @@
     }
   }
 
+  &.right {
+    padding-left: 0;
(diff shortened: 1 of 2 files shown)


mdo:
I'm a no on this exact solution to the problem. We can do inline switches perhaps, but a chained class is against our CSS approach and making it a specific direction limits its possible usage. For now, let's skip out on this and save for an update in v5.

PR #28053, opened 2019-01-14
Build - add a cleanup step to js-compile

Lausselloic:
add a cleanup step to js-compile to prevent unbuilded files in js/dist as today there's an inde.js wich is not generated by the build tasks
Let me know if i'm wrong


Lausselloic:
I've look to put it in another task but don't find a good way :-( In which task would it be?

Okay for the dist file I know I shouldn't touch them. I will remove that part.

Johann-S:
I think we should remove those present in `coverage/dist` too

mdo:
@Johann-S For this PR, it's up to you whether we keep it open and apply it to v5. If you want it for v5, please add the `v5` label and change the base branch to `master`. Otherwise, let's close.

Johann-S:
Closed after a quick chat with @Lausselloic

PR #28060, opened 2019-01-15
Update _card.scss - border on one card accordions

dangowans:
Add back the border for accordions that only have a single card inside.

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -280,6 +280,11 @@
     &:last-of-type {
       @include border-top-radius(0);
     }
+    
+    &:first-of-type:last-of-type {
+      border-bottom: $card-border-width solid $card-border-color;
+      @include border-radius($card-border-radius);
+    }
 
     .card-header {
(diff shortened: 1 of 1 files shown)


mdo:
Do we have a demo in the docs to show this one? Want to make sure we're able to see these things better when merging PRs :).

dangowans:
First timer here, so not entirely sure what you mean.  (Sorry.)

But here's the behaviour I'm seeing.
https://jsfiddle.net/8t5oev9r/1/

And here's the card accordion with my pull request.
https://jsfiddle.net/8t5oev9r/2/

MartijnCuppens (changes requested):
Hi @dangowans, 

Thanks for the PR, but I think we can fix this in one line. Could you just change line 275 to:
`    &:first-of-type:not(:last-of-type) {`

XhmikosR:
@dangowans: ping

XhmikosR:
@MartijnCuppens can you tackle this yourself?

PR #28061, opened 2019-01-15
Refer to reason why a card has `min-width`

MartijnCuppens:
`min-width: 0` is a browser default. There's no reason to add this. (it was added here https://github.com/twbs/bootstrap/commit/891bca435fb4a13ef60e1bc65b522c6be1e31bb3)

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -6,7 +6,7 @@
   position: relative;
   display: flex;
   flex-direction: column;
-  min-width: 0;
+  min-width: 0; // See https://github.com/twbs/bootstrap/pull/22740#issuecomment-305868106
   word-wrap: break-word;
   background-color: $card-bg;
   background-clip: border-box;


ysds:
The default `min-width` is `auto` for flex items. https://developer.mozilla.org/en-US/docs/Web/CSS/min-width

ysds:
Also see https://developer.mozilla.org/en-US/docs/Web/CSS/flex

> By default flex items don't shrink below their minimum content size. To change this, set the item's `min-width` or `min-height`.

Should be added a comment for clarify?

MartijnCuppens:
And `auto == 0` (you can check this in devtools in the computed css tab)

MartijnCuppens:
Oh, now I see. Didn't remember that. Thanks @ysds! I've refered to the reason why this was added instead.

XhmikosR (approved):

PR #28062, opened 2019-01-15
Remove redundant properties from `.embed-responsive`

MartijnCuppens:
We copied the `.embed-responsive` code from somewhere else, but there are a lot of redundant properties in it:

- No need to set `display: block;` on a `<div>`
- No need to remove `padding` from a `<div>`
- No need to set `overflow: hidden;` because the inner content is spread in `.embed-responsive`
- No need for `bottom: 0;`, because we already have `height: 100%;`

### DIFF
--- a/scss/helpers/_embed.scss
+++ b/scss/helpers/_embed.scss
@@ -2,10 +2,7 @@
 
 .embed-responsive {
   position: relative;
-  display: block;
   width: 100%;
-  padding: 0;
-  overflow: hidden;
 
   &::before {
     display: block;
@@ -19,7 +16,6 @@
   video {
     position: absolute;
     top: 0;
-    bottom: 0;
     left: 0;
     width: 100%;
     height: 100%;


ysds:
I want to hear @necolas 's thoughts.

MartijnCuppens:
I also wonder if this css:
```css
.embed-responsive .embed-responsive-item,
.embed-responsive iframe,
.embed-responsive embed,
.embed-responsive object {
  ...
}
```
should be rewritten as:
```css
.embed-responsive > * {
  ...
}
```

I know we better avoid the universal selector but I think for this case with the child combinator and the  decrement of the number of selectors, it might be acceptable.

mdo:
The biggest problems we had with selector performance issue we ever had was using them too much with attribute selectors for column classes. In these instances, I think it's acceptable. However, I'm curious if folks have other content within these elements? Does anything else accompany an embed other than the source?

ysds:
I confirmed the CSS are correct in theory, and tested in IE11, Edge, Firefox and Chrome. Let’s marge this. 

@MartijnCuppens Would you fix the confrict?

MartijnCuppens:
Done

ysds (approved):

MartijnCuppens:
This is a minor optimization, not worth backporting this to `v4`, it might break some edge cases.

PR #28063, opened 2019-01-15
Fix text break in Firefox

MartijnCuppens:
`word-break: break-word` doesn't work in Firefox, but I think `hyphens: auto;`  is an acceptable workaround. I guess `!important` won't be nessesary for `hyphens: auto;`, but not 100% sure.

### DIFF
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -63,9 +63,8 @@
 .text-decoration-none { text-decoration: none !important; }
 
 .text-break {
-  // Add `hyphens: auto;` as workaround to force Firefox to break:
-  hyphens: auto;
-  word-break: break-word !important;
+  word-break: break-word !important; // IE & < Edge 18
+  overflow-wrap: break-word !important;
 }
 
 // Reset
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -63,6 +63,7 @@
 .text-decoration-none { text-decoration: none !important; }
 
 .text-break {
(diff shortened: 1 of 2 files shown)


XhmikosR:
LGTM as a workaround. Maybe add a comment to revisit this.

mdo:
Hmm, didn't know about this! [Could `overflow-wrap` work?](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)

scss/utilities/_text.scss
```diff
@@ -62,7 +62,10 @@
 
 .text-decoration-none { text-decoration: none !important; }
 
-.text-break { word-break: break-word !important; }
+.text-break {
+  word-break: break-word !important; // IE & < Edge 18
+  overflow-wrap: break-word !important;
```
XhmikosR: Should we update the docs too?

MartijnCuppens: Done!

MartijnCuppens:
Forgot to answer on this ticket, but it looked like `overflow-wrap` was indeed the best solution. I've kept `word-break: break-word` for backwards compatibility.

XhmikosR (approved):

adamwathan:
Too late but I think what you actually want is `word-wrap: break-word` for old IE support, `word-break` is a different property with subtly different behavior and `overflow-wrap` is the newer name for `word-wrap`.

image

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

PR #28066, opened 2019-01-15
Deprecate invisible() mixin

MartijnCuppens:
I wonder if someone even ever used `@include invisible(visible);`, makes no sense at all.


sts-ryan-holton:
@MartijnCuppens  believe it or not, I used this a few weeks back :)

sts-ryan-holton:
@MartijnCuppens I can't remember my exact usage, but it can easily be removed. Besides this, I can't remember the last time I even used `visibility `in a project.

mdo (approved):

PR #28068, opened 2019-01-15
Optional btn & input font family

MartijnCuppens:
Optional font family for `.btn`s and `.form-control`s. Doesn't generate any css by default.


mdo:
We'll want to check the default font stacks for some older browsers. I can't remember for certain anymore, but I feel like some weren't `inherit`ing their fonts in inputs and buttons.

MartijnCuppens:
@mdo, I guess this is already fixed in reboot:
https://github.com/twbs/bootstrap/blob/e65dccfc95c04aba3de66d9880ad1e6852b659f2/scss/_reboot.scss#L313-L322

mdo (approved):
Well then, awesome! :shipit: 😆

PR #28075, opened 2019-01-17
use absolute path to output lcov file for coveralls

Johann-S:
Without that we have this error: https://travis-ci.org/twbs/bootstrap/jobs/480784013#L1389

because there is no `lcov.info` file 

/CC @XhmikosR


XhmikosR:
Oops sorry for that, the build wasn't marked as failed so I didn't notice when I removed that.

Johann-S:
Don't worry that's happen 👍

XhmikosR (approved):

PR #28078, opened 2019-01-17
Add responsive borders

3j14:
Changed syntax for the border utilities to add or remove borders
depending on the screen size.
Using the notation from the spacing utilities like `margin` or
`padding`, you can now provide a breakpoint to the additive and
substractive border classes.
This might be useful for dividers that should only be visible on large
screens.

Examples:

 - `border-left` applies to all screen sizes
 - `border-md-left` for a border at the medium breakpoint and above
 - `border-sm-0` to remove a border at the breakpoints small and above

### DIFF
--- a/scss/utilities/_borders.scss
+++ b/scss/utilities/_borders.scss
@@ -4,17 +4,25 @@
 // Border
 //
 
-.border         { border: $border-width solid $border-color !important; }
-.border-top     { border-top: $border-width solid $border-color !important; }
-.border-right   { border-right: $border-width solid $border-color !important; }
-.border-bottom  { border-bottom: $border-width solid $border-color !important; }
-.border-left    { border-left: $border-width solid $border-color !important; }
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
I would love to have this feature, but we've reached the point where we're just adding responsive classes for everything.

I'm more in favor of a solution where you can easily enable/disable the generation of responsive classes, this way you 'll be able to get control over the classes that are generated. Not only for borders, but also for `.text-*` utilities. This will be something to look into in `v5`.

jnsdrtlf:
I understand your concerns, adding responsive classes to most utilities will noticeably increase the size.

So what you are thinking about is a configurator where you specify whether you need responsiveness for certain classes, did I get that right?

MartijnCuppens:
Yup, something that'll be uniform for all the utility classes we generate, probably something with sass maps.

PR #28081, opened 2019-01-18
Change select padding

astrahov:
The select component refers to a form. It is better to take the padding from the input padding so that the form elements look the same.


MartijnCuppens:
This was already implemented:
https://github.com/twbs/bootstrap/blob/842da1ddc3873dc72a725266c8e59a81d31d1f20/scss/_variables.scss#L559-L560

Please update your local repo to the latest version before submitting a new PR.

PR #28086, opened 2019-01-18
Create Gray Background Classes

emiliogarza:
I had a need for a Background Color utility class for a gray color. So I noticed that the `$theme-colors` map was being looped into the `bg-variant` mixin and but not the `$grays` map. I think it could be useful to others to have the bg-variant mixin classes created for the `$grays` like the `$theme-colors` to give more utilities for the all colors they define in variables.

This is my first Pull Request so please let me know if I need to do anything differently and your thoughts!

### DIFF
--- a/scss/utilities/_background.scss
+++ b/scss/utilities/_background.scss
@@ -4,6 +4,10 @@
   @include bg-variant(".bg-#{$color}", $value);
 }
 
+@each $gray, $value in $grays {
+  @include bg-variant(".bg-gray#{$gray}", $value);
+}
+
 @if $enable-gradients {
   @each $color, $value in $theme-colors {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Hi @emiliogarza,

Welcome to the wonderful world of the Bootstrap contributers! 🎉 

Your PR makes sense, but we're currently at a point we don't want to add more utility classes to our default CSS to prevent the filesize to become too large. In `v5` we'll probably be focusing more on getting more control over which utility classes we generate. (I'm currently working on a prove of concept here: https://github.com/twbs/bootstrap/compare/v4-dev-mc-poc-property-utilities?expand=1)

So I'm going to close this for now, but I'll keep these utilities in mind for `v5`.

PR #28067, opened 2019-01-15
Deprecate size mixin

MartijnCuppens:
I guess someone who uses mixins will also know the `width` and `height` properties 😃

PR #28092, opened 2019-01-19
Add `deprecate()` mixin

MartijnCuppens:
We're probably going to drop a lot of mixins, and maybe some functions in `v5`. That's why I would like to introduce a `deprecate()` mixin. This mixin will allow us to deprecate mixins and functions now. Its goal is that all functions/mixins that will be droped in v5 include this mixin so that it'll be easy to check if people are relying on deprecated stuff in their codebase if they want to upgrade from Bootstrap 4 to 5. (They can update to the latest `v4` first where all the `deprecate()` mixins are included.)

I've put #28072, #28067 and #28066 hold for this (I acctually tought `$ignore-warning` was a global variable, which wasn't the case). Probably going to submit a lot more deprecation PRs (these won't be blocking the `4.3` release)

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -117,13 +117,7 @@ $enable-grid-classes:                         true !default;
 $enable-pointer-cursor-for-buttons:           true !default;
 $enable-print-styles:                         true !default;
 $enable-validation-icons:                     true !default;
-
-// Show deprecation messages
-//
-// Some mixins and functions will be removed in v5, the `deprecate()` mixin will warn you if you're using one of these.
-// These warnings can be suppressed by setting this variable to `false`:
-
-$show-deprecation-messages:                   true !default;
+$enable-deprecation-messages:                 true !default;
 
 
 // Spacing
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -118,6 +118,13 @@ $enable-pointer-cursor-for-buttons:           true !default;
 $enable-print-styles:                         true !default;
 $enable-validation-icons:                     true !default;
 
+// Show deprecation messages
+//
+// Some mixins and functions will be removed in v5, the `deprecate()` mixin will warn you if you're using one of these.
+// These warnings can be suppressed by setting this variable to `false`:
+
+$show-deprecation-messages:                   true !default;
+
 
 // Spacing
 //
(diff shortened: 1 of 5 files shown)


scss/_variables.scss
```diff
@@ -118,6 +118,13 @@ $enable-pointer-cursor-for-buttons:           true !default;
 $enable-print-styles:                         true !default;
 $enable-validation-icons:                     true !default;
 
+// Show deprecation messages
+//
+// Some mixins and functions will be removed in v5, the `deprecate()` mixin will warn you if you're using one of these.
+// These warnings can be suppressed by setting this variable to `false`:
+
+$show-deprecation-messages:                   true !default;
```
mdo: Should we stick with `$enable-` here?

Also, this is set to true, but shouldn't it be set fo `false` to start? This replaces `$ignore-warning != true`, so it's disabled to start I believe.

scss/mixins/_deprecate.scss
```diff
@@ -0,0 +1,10 @@
+// Deprecate mixin
+//
+// This mixin can be used to deprecate mixins or functions.
+// `$show-deprecation-messages` is a global variable, `$ignore-warning` is a variable that can be passed to
```
mdo:
```suggestion
// `$enable-deprecation-messages` is a global variable, `$ignore-warning` is a variable that can be passed to
```

site/docs/4.2/getting-started/theming.md
```diff
@@ -243,6 +243,7 @@ You can find and customize these variables for key global options in Bootstrap's
 | `$enable-pointer-cursor-for-buttons`         | `true` (default) or `false`        | Add "hand" cursor to non-disabled button elements. |
 | `$enable-print-styles`                       | `true` (default) or `false`        | Enables styles for optimizing printing. |
 | `$enable-validation-icons`                   | `true` (default) or `false`        | Enables `background-image` icons within textual inputs and some custom forms for validation states. |
+| `$enable-deprecation-messages`               | `true` or `false` (default)        | Some mixins and functions will be removed in `v5`, if this variable is set to `true` the `deprecate()` mixin will warn you if you're using one of these. |
```
mdo:
```suggestion
| `$enable-deprecation-messages`               | `true` or `false` (default)        | Set to `true` to show warnings when using any of the deprecated mixins and functions that are planned to be removed in `v5`. |
```

mdo (commented):
Couple copy changes here, but otherwise looking good. Appreciate the mixin approach!

mdo (approved):

PR #28096, opened 2019-01-21
Update _variables.scss

astrahov:
New $font-weight-* (like in a Google Fonts)


mdo:
Thanks, but we don’t need this many variables for `font-weight`.

PR #28098, opened 2019-01-21
Just pass `$ignore-warning` instead of also defining it

MartijnCuppens:
`$ignore-warning` shouldn't be set here, it's an include.

### DIFF
--- a/scss/utilities/_text.scss
+++ b/scss/utilities/_text.scss
@@ -57,7 +57,7 @@
 // Misc
 
 .text-hide {
-  @include text-hide($ignore-warning: true);
+  @include text-hide(true);
 }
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Shouldn't we just go with the mixin instead?

ysds:
Both are correct. The current is a keyword argument - for easier to read IMO.

https://sass-lang.com/documentation/file.SASS_REFERENCE.html#mixin-arguments

MartijnCuppens:
@XhmikosR, we need the variable to suppress the warning because the mixin is used by bootstrap here.

I changed my mind about this after reading the sass documentation, @ysds is right here, no one will know what the true stands for here: `text-hide(true)`

Closing this.

PR #28100, opened 2019-01-22
Create `border-top-left-radius` and `border-bottom-right-radius` mixins

hunkjazz:
I couldn't find the reason why there weren't mixins for the **top-left** and **bottom-right** corners. So I did it, anyway.


MartijnCuppens:
Thanks @JLammeer!

Could you switch the positions of the mixins, this way the mixins will be defined in clockwise order?

JLammeer:
Done.

ysds:
Should be add to blacklist in [.stylelintrc](https://github.com/twbs/bootstrap/blob/v4-dev/.stylelintrc) 😉

MartijnCuppens:
@ysds, it looks like they are were already added

MartijnCuppens (approved):

PR #28105, opened 2019-01-22
Use consistent arrow class

### LINKED ISSUE #28028: Popover/tooltip arrow class names are not consistent
For the arrow inside of the popover, the `.arrow` class is used, while `.arrow` isn't a standalone component. We don't do that anywhere else in our css. A better name would be something like `.popover-arrow`. This will be something we can look in to for `v5`, since this is a breaking change.

Edit: We should do the same with the tooltips.

peterblazejewicz:
This changes the name of the `.arrow` class to:
- `.popover-arrow`
- `.tooltip-arrow`
for arrow and tooltip respectively.

Fixes #28028

Thanks!

### DIFF
--- a/js/src/popover.js
+++ b/js/src/popover.js
@@ -28,7 +28,7 @@ const Default = {
   trigger   : 'click',
   content   : '',
   template  : '<div class="popover" role="tooltip">' +
-              '<div class="arrow"></div>' +
+              '<div class="popover-arrow"></div>' +
               '<h3 class="popover-header"></h3>' +
               '<div class="popover-body"></div></div>'
 }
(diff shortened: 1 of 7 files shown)


XhmikosR:
This is a breaking change, shouldn't go in 4.3.

Johann-S:
I agree with @XhmikosR

XhmikosR:
@Johann-S @MartijnCuppens actually this PR targets the no jQuery branch which will be a major version bump.

XhmikosR:
But being it has changes from v4-dev, we need the jQuery branch to be rebased against the latest v4-dev first and then this one.

MartijnCuppens:
Yup, that's why I added the `v5` and `on-hold` labels

XhmikosR:
We can merge this if the no jQuery branch is rebased.

XhmikosR (approved):

PR #28112, opened 2019-01-24
Fix custom select caret issue on IE

ysds:
If you add a width to a custom select, the display differs between IE and other browsers.

image

Demo: https://jsbin.com/xihojulawa/1/edit?output

This PR fix that.


MartijnCuppens (approved):
LGTM. Thanks!

PR #28130, opened 2019-01-28
Add support for tooltip offset option to be a function

### LINKED ISSUE #24223: Allow function in offset option for Tooltips and Popovers
Popper.js allow to pass function to this parameter it'll allow folks more flexibility on how they handle their offsets.
See : https://popper.js.org/popper-documentation.html#modifiers..offset.fn

Related to #24208
/CC @romanlex

jgolieb:
Fixes #24223

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -273,16 +273,6 @@ class Tooltip {
       const attachment = this._getAttachment(placement)
       this.addAttachmentClass(attachment)
 
-      const offsetConf = {}
-      if (typeof this.config.offset === 'function') {
-        offsetConf.fn = (data) => {
-          data.offsets = $.extend({}, data.offsets, this.config.offset(data.offsets, this.element) || {})
-          return data
-        }
-      } else {
-        offsetConf.offset = this.config.offset
-      }
-
       const container = this._getContainer()
       $(tip).data(this.constructor.DATA_KEY, this)
 
@@ -295,7 +285,7 @@ class Tooltip {
       this._popper = new Popper(this.element, tip, {
         placement: attachment,
         modifiers: {
-          offset: offsetConf,
+          offset: this._getOffset(),
           flip: {
             behavior: this.config.fallbackPlacement
           },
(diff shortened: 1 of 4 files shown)


js/src/tooltip.js
```diff
@@ -273,6 +273,16 @@ class Tooltip {
       const attachment = this._getAttachment(placement)
       this.addAttachmentClass(attachment)
 
+      const offsetConf = {}
+      if (typeof this.config.offset === 'function') {
+        offsetConf.fn = (data) => {
+          data.offsets = $.extend({}, data.offsets, this.config.offset(data.offsets, this.element) || {})
```
Johann-S: please use the spread operator here

```js
data.offsets = {
  ...data.offsets,
  ...this.config.offset(data, this.element) || {}
}

**Review by Johann-S — CHANGES_REQUESTED** [review]
Can you move the logic in a private method please and can you add unit tests please

File: js/src/tooltip.js (hunk 1 above)
Comment (jgolieb) [reply js/src/tooltip.js]: Done.

**Comment by Johann-S** [comment]
@XhmikosR it seems we have some issues with Bundlesize, the github token is not found https://travis-ci.org/twbs/bootstrap/jobs/486082928#L833

File: js/tests/unit/tooltip.js [hunk 2]
```diff
@@ -1069,4 +1069,38 @@ $(function () {
...
+    assert.ok(typeof offset.offset === 'undefined')
+    assert.ok(typeof offset.fn === 'function')
+  })
+
+  QUnit.test('should create offset modifier correctly when offset option is not a function', function (assert) {
+    assert.expect(2)
+
+    var $trigger = $('<a href="#" rel="tooltip" data-trigger="click" title="Another tooltip"/>')
+      .appendTo('#qunit-fixture')
+      .bootstrapTooltip()
```
Johann-S:
```js
var myOffset = 5
var $trigger = $('<a href="#" rel="tooltip" data-trigger="click" title="Another tooltip"/>')
  .appendTo('#qunit-fixture')
  .bootstrapTooltip({
    offset: myOffset,
  })
```

and after you'll be able to do:
```js
assert.strictEqual(offset.offset, myOffset)
```

I think it's more explicit like that

Johann-S:
Just one comment and it'll be alright 👍

XhmikosR:
@Johann-S the GH token isn't (and wasn't) available for PRs made by people without push rights, same as Browser Stack.

The build fails here due to the bundlesize limit. We'll need to increase the two bundle JS limits a bit.

Johann-S (approved):

PR #28140, opened 2019-01-30
Fix negative margin example

garhbod:
Negative margin example wasn't displaying correctly due to border and background styling being applied directly to the `.col` element. Made a child element and applied the styling to it

Original: https://getbootstrap.com/docs/4.2/utilities/spacing/#negative-margin
Demo: https://deploy-preview-28140--twbs-bootstrap4.netlify.com/docs/4.2/utilities/spacing/#negative-margin


MartijnCuppens (approved):
Looks indeed better

Issue #28150, opened 2019-01-31
Dropdown Input Group Issue


seanbotha123:
Hi guys.

Looks like there is an issue when adding a dropdown button to an input group, and not having the dropdown as the last element, the border-radius' are not correct:

![screenshot from 2019-01-31 07-21-54](https://user-images.githubusercontent.com/5496698/52032299-06d4b180-2529-11e9-8beb-ada5f1714bb2.png)

Please see fiddle for reproduction: http://jsfiddle.net/nxqdgkas/6/

- Operating system and version: **Ubuntu 18.04.1 LTS**
- Browser and version: **Chrome 70.0.3538.77**

Thanks!

ysds:
If you want to use a dropdown on the button that isn't at the last element of the input group, you can do it by using multiple `.input-group-append`s. I mentioned about this usage in #25075, but it isn't described in the documents.

Demo: https://codepen.io/anon/pen/wNJWeL

Why need multiple `.input-group-append`? A `.dropdown-menu` must be the last child of the wrapper (`.dropdown`, `.btn-group` or an element has `position: relative`). This mean the following code is incorrect:

```html
<div class="input-group">
  <input type="text" class="form-control">
  <div class="input-group-append">
    <button class="btn btn-outline-secondary dropdown-toggle" type="button" data-toggle="dropdown">Dropdown</button>
    <!-- dropdown-menu must be at the last of the container. -->
    <div class="dropdown-menu">
      <a class="dropdown-item" href="#">Action</a>
      <a class="dropdown-item" href="#">Another action</a>
      <a class="dropdown-item" href="#">Something else here</a>
    </div>
    <button class="btn btn-outline-secondary" type="button">Button</button>
  </div>
</div>
```

So, you should use multiple `.input-group-append`s:

```html
<div class="input-group">
  <input type="text" class="form-control">
  <div class="input-group-append">
    <button class="btn btn-outline-secondary dropdown-toggle" type="button" data-toggle="dropdown">Dropdown</button>
    <div class="dropdown-menu">
      <a class="dropdown-item" href="#">Action</a>
      <a class="dropdown-item" href="#">Another action</a>
      <a class="dropdown-item" href="#">Something else here</a>
    </div>
  </div>
  <div class="input-group-append">
    <button class="btn btn-outline-secondary" type="button">Button</button>
  </div>
</div>
```

MartijnCuppens:
@seanbotha123,

You could indeed go for the approach suggested by @ysds.

[...]

PR #25076, opened 2017-12-25
Fix input group border radius issue (#25075)

ysds:
Fixes #25075

PR #28152, opened 2019-01-31
Fix multiple dropdowns in input group

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/issues/28150

Demo: https://deploy-preview-28152--twbs-bootstrap4.netlify.com/docs/4.2/components/input-group/#buttons-with-dropdowns


scss/_input-group.scss
```diff
@@ -82,11 +82,11 @@
...
-  .btn + .btn,
-  .btn + .input-group-text,
-  .input-group-text + .input-group-text,
-  .input-group-text + .btn {
-    margin-left: -$input-border-width;
+  .input-group-text,
+  .btn {
+    &:not(:first-child) {
+      margin-left: -$input-border-width;
+    }
```
MartijnCuppens: I could also have added `.dropdown-menu` here, but figured out this would save us a lot of selectors.

scss/_input-group.scss
```diff
@@ -179,6 +179,7 @@
 .input-group > .input-group-append:not(:last-child) > .btn,
 .input-group > .input-group-append:not(:last-child) > .input-group-text,
 .input-group > .input-group-append:last-child > .btn:not(:last-child):not(.dropdown-toggle),
+.input-group > .input-group-append:last-child > .dropdown-toggle:nth-last-child(n + 3),
```
MartijnCuppens: `(n + 3)` stands for third last or more. This way the radius is removed only if the `.dropdown-toggle` is not the second last element ( and `.dropdown-menu` is the last). Couldn't combine this with L181, because complex selectors for `:not()` are not well supported.

site/docs/4.2/components/input-group.md
```diff
@@ -211,17 +211,32 @@ Multiple add-ons are supported and can be mixed with checkbox and radio input ve
...
+    </div>
+  </div>
+</div>
+
 <div class="input-group">
   <input type="text" class="form-control" aria-label="Text input with dropdown button">
   <div class="input-group-append">
     <button class="btn btn-outline-secondary dropdown-toggle" type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Dropdown</button>
-    <div class="dropdown-menu">
+    <div class="dropdown-menu dropdown-menu-right">
```
MartijnCuppens: I've added this because it just looks nicer if the dropdowns are aligned with the end of the button

ysds:
Hi @MartijnCuppens #28150 would be not a CSS issue, because it has already been fixed in #25076. Please see https://github.com/twbs/bootstrap/issues/28150#issuecomment-459557777.

scss/_input-group.scss
mdo: This feels like it'd break very easily—folks introduce custom markup in all the places we wouldn't expect. 😬

MartijnCuppens:
Closing this after feedback from @ysds and @mdo, thanks guys!

PR #28158, opened 2019-02-01
minor jQuery.hasclass removal

jwilson8767:


XhmikosR:
Are there any other instances?

XhmikosR:
Hmm tests fail though.

jwilson8767:
No other occurrences of this particular issue, not sure why tests are failing, could `this._dialog` sometimes be a jquery instance?

XhmikosR:
Yeah, maybe tests need to be adapted too.

Johann-S:
That's because a lot of modals in our tests do not have a `.modal-dialog` child.

It wasn't an issue before because jQuery fail in silence but not Vanilla JS

XhmikosR:
@jwilson8767 can you fix the issue?

jwilson8767:
@XhmikosR Done, and tests are now passing.

XhmikosR:
Thanks! @Johann-S I think it's fine now, if so squash and merge please.

ghiscoding:
Hey guys, I just happened to have a peak at the PR and, I just noticed that the code is not exactly 100% bullet proof. I am using TypeScript a lot lately and in few occasion with the TSlint `no null` option and this PR would fail it. 

Technically the following is 100% safe

```diff
- if (this._dialog && this._dialog.classList.contains(ClassName.SCROLLABLE)) {
+ if (this._dialog && this._dialog.classList && this._dialog.classList.contains && this._dialog.classList.contains(ClassName.SCROLLABLE)) {
```

But I think that if you have a `classList`, you most probably have a `contains`, so the previous might be overkill and it might be more than enough to go with the simpler
```diff
- if (this._dialog && this._dialog.classList.contains(ClassName.SCROLLABLE)) {
+ if (this._dialog && this._dialog.classList && this._dialog.classList.contains(ClassName.SCROLLABLE)) {
```

Anywho, if you think my suggestion is still overkill then please disregard, but I think it's always better to avoid any possibility of errors thrown in the console... Bootstrap is awesome, thank you :)

js/src/modal.js
```diff
@@ -245,7 +245,7 @@ class Modal {
     this._element.removeAttribute('aria-hidden')
     this._element.setAttribute('aria-modal', true)
 
-    if ($(this._dialog).hasClass(ClassName.SCROLLABLE)) {
+    if (this._dialog && this._dialog.classList.contains(ClassName.SCROLLABLE)) {
```
Johann-S:
```js
if (this._dialog.classList.contains(ClassName.SCROLLABLE))
```

is enough, because a Modal must have a `dialog`

Johann-S (changes requested):

Johann-S:
Hi @ghiscoding,

Defensive code isn't something we should do here, because a modal must have a dialog, and when you have an `HTMLElement` you have a `classList`.

This PR won't fail in TS, because `_dialog` is an `HTMLElement`

Johann-S (approved):

Johann-S:
Thanks for your work @jwilson8767 👍

jwilson8767:
@Johann-S Thanks for giving this your attention! Looking forward to being off of jQuery :)

PR #28169, opened 2019-02-04
Add text-decoration to .btn-link.

### LINKED ISSUE #28157: Should .btn-link use $link-decoration
Currently `.btn-link` doesn't use the same text-decoration as anchors.

Anchors use `text-decoration: $link-decoration;`

I think that `.btn-link` should have a similar rule unless there was an explicit reason for not doing so.

bardiharborow:
Mirrors [scss/_reboot.scss#L184](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_reboot.scss#L184). Fixes #28157.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -76,6 +76,7 @@ fieldset:disabled a.btn {
 .btn-link {
   font-weight: $font-weight-normal;
   color: $link-color;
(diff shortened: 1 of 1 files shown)


MartijnCuppens (approved):
Welcome to the CSS side, Bardi 😛

PR #28172, opened 2019-02-04
Update link to _variables.scss

descovi:
Improve documentation with link to variables.
I dont know if link to branch is ok or not.


XhmikosR (changes requested):
This isn't the right way to do it. Use the variables from _config.yml and link to the release's file.

mdo:
This is also the wrong thing to link to—we shouldn't link to our variables file on GitHub, this is meant to point to your local copy of it.

PR #28178, opened 2019-02-05
Add dark toast variant

vanillajonathan:
Suitable for white backgrounds.


XhmikosR (dismissed):
I'm not sure it's useful, but regardless, your liquid is wrong.

vanillajonathan:
Cards, alerts, badges, etc are presented in many variants. So it could be useful to illustrate the toast in alternate variants too.

A dark one could fit well together with a white background. It also resembles the [snackbar](https://material.io/design/components/snackbars.html) from Material Design.

It shows a different use of the toast as an action conformation as opposed to a message coming from a source (where the source in the example is "Bootstrap").

I did not understand the liquid is wrong statement.

XhmikosR:
Did you even try your changes? Or even just open the PR preview.

vanillajonathan:
No, I only saw that it rendered properly in my React application where it is conditionally rendered according to the state of my React application. I actually have not tested it that it does work with the JavaScript provided by Bootstrap. My bad.

XhmikosR:
Yeah, what you have now doesn't output anything in the example. Please check another example to see how to do it.

vanillajonathan:
As far as I can tell, it works.
https://jsfiddle.net/9y0uk35z/

XhmikosR:
Like I said, it doesn't.

XhmikosR:
https://deploy-preview-28178--twbs-bootstrap4.netlify.com/docs/4.2/components/toasts/#dark

vanillajonathan:
Oh my bad.
It looks like I got it working.
https://deploy-preview-28178--twbs-bootstrap4.netlify.com/docs/4.2/components/toasts/#dark

XhmikosR:
Still it needs more content IMO.

vanillajonathan:
I assume you mean an introductory text.

Perhaps?
> Toasts can be customized. Here is one suited for bright backgrounds.

I could need some help in coming up with a good paragraph to describe the dark variant. Feel free to commit suitable content.

XhmikosR:
Just copy an existent toast and adapt it.

vanillajonathan:
I am not sure what you mean. Do you mean that you want the dark variant of the toast to include the header and close button?

Like this: https://jsfiddle.net/5tjs10bm/

Maybe there should be an example that illustrates that a toast does not necessarily need to have a toast-header and a close button.

XhmikosR:
Yeah like your example. The other case shouldn't be mixed with the dark example, so needs a separate section + example.

site/docs/4.2/components/toasts.md
```diff
@@ -72,6 +72,14 @@ Toasts are slightly translucent, too, so they blend over whatever they might app
 <div class="bg-light">
 {% capture example %}
 <div class="toast bg-dark text-white" role="alert" aria-live="assertive" aria-atomic="true">
+  <div class="toast-header bg-dark text-white">
+    <img src="..." class="rounded mr-2" alt="...">
```
XhmikosR: This isn't right. **Please** check an example above

site/docs/4.2/components/toasts.md
```diff
@@ -73,11 +73,11 @@ Toasts are slightly translucent, too, so they blend over whatever they might app
 {% capture example %}
 <div class="toast bg-dark text-white" role="alert" aria-live="assertive" aria-atomic="true">
   <div class="toast-header bg-dark text-white">
-    <img src="..." class="rounded mr-2" alt="...">
+    {% include icons/placeholder.svg width="20" height="20" background="#007aff" class="rounded mr-2" text=" " title=" " %}
     <strong class="mr-auto">Bootstrap</strong>
     <small>11 mins ago</small>
     <button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
-      <span class="text-white" aria-hidden="true">&times;</span>
```
XhmikosR: You still need this here I think.

XhmikosR: Err, I mean in `button`

site/docs/4.2/components/toasts.md
```diff
@@ -32,7 +32,7 @@ Toasts are as flexible as you need and have very little required markup. At a mi
     {% include icons/placeholder.svg width="20" height="20" background="#007aff" class="rounded mr-2" text=" " title=" " %}
     <strong class="mr-auto">Bootstrap</strong>
     <small>11 mins ago</small>
-    <button type="button" class="ml-2 mb-1 close text-white" data-dismiss="toast" aria-label="Close">
+    <button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
```
XhmikosR: This needs to be reverted.

vanillajonathan:
Okay, I think we got it now.
https://deploy-preview-28178--twbs-bootstrap4.netlify.com/docs/4.2/components/toasts/#dark

XhmikosR:
Thanks, now it's OK.

/CC @mdo @MartijnCuppens

MartijnCuppens:
The `background-color`, `color` and heading `color` can be configured in sass, don't really think there's a need for as well dark and light toasts.

We also lose the slight difference in text color for headings.

Also, 4 utility classes are added in the example to get to the dark version, I guess a lot of people will easily forget to add one of them.

And if we would continue with this I would go for a more global solution so that for example modals would have the same variants. But I guess it would be better if we discussed that in an issue instead of on this PR.

vanillajonathan:
Many people, I would argue most people do not compile their own Bootstrap with SASS, they just download the precompiled Bootstrap from a package manager or link to a CDN.

It would be nice to have a dark variant of the toast ship with Bootstrap.

PR #28194, opened 2019-02-06
WIP - make our methods private

Johann-S:
I try to transform our (fake) private methods to real one

but ESLint continue to throw me errors 😟


XhmikosR:
@hzoo: do you have any idea why this fails?

hzoo:
Yea, we either need to modify the rule or add it to scope https://github.com/babel/babel-eslint/issues/746, shouldn't be too bad to fix but no one has yet

Johann-S:
I tought about that PR, and we should close it, because currently we ship Bootstrap with fake private methods which are in reality public methods, so it'll introduce a sort of breaking change.

So we should tackle that in our v5

(@XhmikosR don't laugh on me 😄)

Thanks @hzoo to taking a look at that PR 👍

PR #25524, opened 2018-01-31
Examples: Prevent focus into closed offcanvas

ysds:
Fixes #25441

PR #28203, opened 2019-02-07
Offcanvas example: transition the transform

XhmikosR:
This is more efficient.

Opening this so that we don't forget about it.

### DIFF
--- a/site/docs/4.2/examples/offcanvas/offcanvas.css
+++ b/site/docs/4.2/examples/offcanvas/offcanvas.css
@@ -19,13 +19,14 @@ body {
     overflow-y: auto;
     visibility: hidden;
     background-color: #343a40;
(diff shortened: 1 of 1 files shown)


MartijnCuppens (approved):

XhmikosR:
Not sure about removing `visibility`. Let's put it on hold for now.

ysds:
Please see #25524 for why `visibility` is necessary.

XhmikosR:
OK, I added back visibility.

ysds:
These changes increase the file size, but it will be more efficient in the future.

XhmikosR:
We need to promote the best practices.

ysds:
If I am not misunderstanding your comment, I agree with these changes.

XhmikosR:
Yeah, we say the same thing :)

PR #28210, opened 2019-02-08
[WIP] Experimental carousel example.

XhmikosR:
Show three items, slide by one.

Needs some love, but opening this so that we don't forget about it.

Ideally our JS plugin should handle this.

TODO:

- [ ] Style the whole example
- [ ] Add screenshots

Preview: <https://deploy-preview-28210--twbs-bootstrap.netlify.app/docs/5.0/examples/carousel-experiment/>


XhmikosR:
This needs to be rewritten in vanilla JS.

XhmikosR:
@mdo @MartijnCuppens: this is functional thanks to @Johann-S (it was broken afte rthe jQuery removal), so you can style it and then we can ship it

MartijnCuppens:
Does it really need theming, I don't know what you had in mind?

XhmikosR:
It just feel too plain right now, you know. So maybe you guys have some suggestions how to make the whole example more appealing maybe like the normal carousel example?

mdo:
FWIW, I wouldn't mind something like this. Unsure how feasible it is to add to the existing plugin though for v5.

XhmikosR:
I'm going to close this since I think this feature should be in core. PRs welcome :)

PR #28236, opened 2019-02-11
sanitize template option for tooltip/popover plugins

Johann-S:
XSS was possible in the tooltip or popover data-template, data-content and data-title attributes.

Fixes CVE-2019-8331.

### DIFF
--- a/js/src/tools/sanitizer.js
+++ b/js/src/tools/sanitizer.js
@@ -1,0 +1,127 @@
+/**
+ * --------------------------------------------------------------------------
+ * Bootstrap (v4.3.0): tools/sanitizer.js
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ * --------------------------------------------------------------------------
+ */
+
+const uriAttrs = [
+  'background',
+  'cite',
+  'href',
+  'itemtype',
+  'longdesc',
+  'poster',
+  'src',
+  'xlink:href'
+]
+
+const ARIA_ATTRIBUTE_PATTERN = /^aria-[\w-]*$/i
+
+export const DefaultWhitelist = {
+  // Global attributes allowed on any supplied element below.
+  '*': ['class', 'dir', 'id', 'lang', 'role', ARIA_ATTRIBUTE_PATTERN],
+  a: ['target', 'href', 'title', 'rel'],
+  area: [],
+  b: [],
+  br: [],
+  col: [],
+  code: [],
+  div: [],
+  em: [],
+  hr: [],
+  h1: [],
+  h2: [],
+  h3: [],
+  h4: [],
+  h5: [],
+  h6: [],
+  i: [],
+  img: ['src', 'alt', 'title', 'width', 'height'],
+  li: [],
+  ol: [],
+  p: [],
+  pre: [],
+  s: [],
+  small: [],
+  span: [],
+  sub: [],
+  sup: [],
+  strong: [],
+  u: [],
+  ul: []
+}
+
+/**
+ * A pattern that recognizes a commonly useful subset of URLs that are safe.
+ *
+ * Shoutout to Angular 7 https://github.com/angular/angular/blob/7.2.4/packages/core/src/sanitization/url_sanitizer.ts
+ */
+const SAFE_URL_PATTERN = /^(?:(?:https?|mailto|ftp|tel|file):|[^&:/?#]*(?:[/?#]|$))/gi
+
+/**
+ * A pattern that matches safe data URLs. Only matches image, video and audio types.
+ *
+ * Shoutout to Angular 7 https://github.com/angular/angular/blob/7.2.4/packages/core/src/sanitization/url_sanitizer.ts
+ */
+const DATA_URL_PATTERN = /^data:(?:image\/(?:bmp|gif|jpeg|jpg|png|tiff|webp)|video\/(?:mpeg|mp4|ogg|webm)|audio\/(?:mp3|oga|ogg|opus));base64,[a-z0-9+/]+=*$/i
+
+function allowedAttribute(attr, allowedAttributeList) {
+  const attrName = attr.nodeName.toLowerCase()
+
+  if (allowedAttributeList.indexOf(attrName) !== -1) {
+    if (uriAttrs.indexOf(attrName) !== -1) {
+      return Boolean(attr.nodeValue.match(SAFE_URL_PATTERN) || attr.nodeValue.match(DATA_URL_PATTERN))
+    }
+
+    return true
+  }
+
+  const regExp = allowedAttributeList.filter((attrRegex) => attrRegex instanceof RegExp)
+
+  // Check if a regular expression validates the attribute.
+  for (let i = 0, l = regExp.length; i < l; i++) {
+    if (attrName.match(regExp[i])) {
+      return true
+    }
+  }
+
+  return false
+}
+
+export function sanitizeHtml(unsafeHtml, whiteList, sanitizeFn) {
+  if (unsafeHtml.length === 0) {
+    return unsafeHtml
+  }
+
+  if (sanitizeFn && typeof sanitizeFn === 'function') {
+    return sanitizeFn(unsafeHtml)
+  }
+
+  const domParser = new window.DOMParser()
+  const createdDocument = domParser.parseFromString(unsafeHtml, 'text/html')
+  const whitelistKeys = Object.keys(whiteList)
+  const elements = [].slice.call(createdDocument.body.querySelectorAll('*'))
+
+  for (let i = 0, len = elements.length; i < len; i++) {
+    const el = elements[i]
+    const elName = el.nodeName.toLowerCase()
+
+    if (whitelistKeys.indexOf(el.nodeName.toLowerCase()) === -1) {
+      el.parentNode.removeChild(el)
+
+      continue
+    }
+
+    const attributeList = [].slice.call(el.attributes)
+    const whitelistedAttributes = [].concat(whiteList['*'] || [], whiteList[elName] || [])
+
+    attributeList.forEach((attr) => {
+      if (!allowedAttribute(attr, whitelistedAttributes)) {
+        el.removeAttribute(attr.nodeName)
+      }
+    })
+  }
+
+  return createdDocument.body.innerHTML
+}
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -5,6 +5,10 @@
  * --------------------------------------------------------------------------
  */
 
+import {
+  DefaultWhitelist,
+  sanitizeHtml
+} from './tools/sanitizer'
 import $ from 'jquery'
 import Popper from 'popper.js'
 import Util from './util'
@@ -15,13 +19,14 @@ import Util from './util'
  * ------------------------------------------------------------------------
  */
 
-const NAME               = 'tooltip'
-const VERSION            = '4.3.0'
-const DATA_KEY           = 'bs.tooltip'
-const EVENT_KEY          = `.${DATA_KEY}`
-const JQUERY_NO_CONFLICT = $.fn[NAME]
-const CLASS_PREFIX       = 'bs-tooltip'
-const BSCLS_PREFIX_REGEX = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g')
+const NAME                  = 'tooltip'
+const VERSION               = '4.3.0'
+const DATA_KEY              = 'bs.tooltip'
+const EVENT_KEY             = `.${DATA_KEY}`
+const JQUERY_NO_CONFLICT    = $.fn[NAME]
+const CLASS_PREFIX          = 'bs-tooltip'
+const BSCLS_PREFIX_REGEX    = new RegExp(`(^|\\s)${CLASS_PREFIX}\\S+`, 'g')
+const DISALLOWED_ATTRIBUTES = ['sanitize', 'whiteList', 'sanitizeFn']
 
 const DefaultType = {
   animation         : 'boolean',
@@ -35,7 +40,10 @@ const DefaultType = {
   offset            : '(number|string|function)',
   container         : '(string|element|boolean)',
   fallbackPlacement : '(string|array)',
-  boundary          : '(string|element)'
+  boundary          : '(string|element)',
+  sanitize          : 'boolean',
+  sanitizeFn        : '(null|function)',
+  whiteList         : 'object'
 }
 
 const AttachmentMap = {
@@ -60,7 +68,10 @@ const Default = {
   offset            : 0,
   container         : false,
   fallbackPlacement : 'flip',
-  boundary          : 'scrollParent'
+  boundary          : 'scrollParent',
+  sanitize          : true,
+  sanitizeFn        : null,
+  whiteList         : DefaultWhitelist
 }
 
 const HoverState = {
@@ -419,18 +430,27 @@ class Tooltip {
   }
 
   setElementContent($element, content) {
-    const html = this.config.html
     if (typeof content === 'object' && (content.nodeType || content.jquery)) {
       // Content is a DOM node or a jQuery
-      if (html) {
+      if (this.config.html) {
         if (!$(content).parent().is($element)) {
           $element.empty().append(content)
         }
       } else {
         $element.text($(content).text())
       }
+
+      return
+    }
+
+    if (this.config.html) {
+      if (this.config.sanitize) {
+        content = sanitizeHtml(content, this.config.whiteList, this.config.sanitizeFn)
+      }
+
+      $element.html(content)
     } else {
-      $element[html ? 'html' : 'text'](content)
+      $element.text(content)
     }
   }
 
@@ -636,9 +656,18 @@ class Tooltip {
   }
 
   _getConfig(config) {
+    const dataAttributes = $(this.element).data()
+
+    Object.keys(dataAttributes)
+      .forEach((dataAttr) => {
+        if (DISALLOWED_ATTRIBUTES.indexOf(dataAttr) !== -1) {
+          delete dataAttributes[dataAttr]
+        }
+      })
+
     config = {
       ...this.constructor.Default,
-      ...$(this.element).data(),
+      ...dataAttributes,
       ...typeof config === 'object' && config ? config : {}
     }
 
@@ -662,6 +691,10 @@ class Tooltip {
       config,
       this.constructor.DefaultType
     )
+
+    if (config.sanitize) {
+      config.template = sanitizeHtml(config.template, config.whiteList, config.sanitizeFn)
+    }
 
     return config
   }
--- a/site/docs/4.3/components/popovers.md
+++ b/site/docs/4.3/components/popovers.md
@@ -140,6 +140,11 @@ Enable popovers via JavaScript:
 
 Options can be passed via data attributes or JavaScript. For data attributes, append the option name to `data-`, as in `data-animation=""`.
 
+{% capture callout %}
+Note that for security reasons the `sanitize`, `sanitizeFn` and `whiteList` options cannot be supplied using data attributes.
+{% endcapture %}
+{% include callout.html content=callout type="warning" %}
+
 <table class="table table-bordered table-striped">
   <thead>
     <tr>
@@ -249,6 +254,24 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>string | element</td>
       <td>'scrollParent'</td>
       <td>Overflow constraint boundary of the popover. Accepts the values of <code>'viewport'</code>, <code>'window'</code>, <code>'scrollParent'</code>, or an HTMLElement reference (JavaScript only). For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.boundariesElement">preventOverflow docs</a>.</td>
+    </tr>
+    <tr>
+      <td>sanitize</td>
+      <td>boolean</td>
+      <td>true</td>
+      <td>Enable or disable the sanitization. If activated <code>'template'</code>, <code>'content'</code> and <code>'title'</code> options will be sanitized.</td>
+    </tr>
+    <tr>
+      <td>whiteList</td>
+      <td>object</td>
+      <td><a href="{{ site.baseurl }}/docs/{{ site.docs_version }}/getting-started/javascript/#sanitizer">Default value</a></td>
+      <td>Object which contains allowed attributes and tags</td>
+    </tr>
+    <tr>
+      <td>sanitizeFn</td>
+      <td>null | function</td>
+      <td>null</td>
+      <td>Here you can supply your own sanitize function. This can be useful if you prefer to use a dedicated library to perform sanitization.</td>
     </tr>
   </tbody>
 </table>
--- a/site/docs/4.3/getting-started/javascript.md
+++ b/site/docs/4.3/getting-started/javascript.md
@@ -139,3 +139,73 @@ Bootstrap's plugins don't fall back particularly gracefully when JavaScript is d
 All Bootstrap's JavaScript files depend on `util.js` and it has to be included alongside the other JavaScript files. If you're using the compiled (or minified) `bootstrap.js`, there is no need to include this—it's already there.
 
 `util.js` includes utility functions and a basic helper for `transitionEnd` events as well as a CSS transition emulator. It's used by the other plugins to check for CSS transition support and to catch hanging transitions.
+
+## Sanitizer
+
+Tooltips and Popovers use our built-in sanitizer to sanitize options which accept HTML.
+
+The default `whiteList` value is the following:
+
+{% highlight js %}
+var ARIA_ATTRIBUTE_PATTERN = /^aria-[\w-]*$/i
+var DefaultWhitelist = {
+  // Global attributes allowed on any supplied element below.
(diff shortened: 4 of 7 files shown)


site/docs/4.3/components/popovers.md
```diff
@@ -250,6 +250,25 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>'scrollParent'</td>
       <td>Overflow constraint boundary of the popover. Accepts the values of <code>'viewport'</code>, <code>'window'</code>, <code>'scrollParent'</code>, or an HTMLElement reference (JavaScript only). For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.boundariesElement">preventOverflow docs</a>.</td>
     </tr>
+    <tr>
+      <td>sanitizeTemplate</td>
+      <td>boolean</td>
+      <td>true</td>
+      <td>Enable or disable the built-in sanitizer. If activated the <code>'template'</code> option will be sanitized.
+      You can disable this option if you want to sanitize your template with a dedicated library.</td>
```
markcarver: Just disabling this option doesn't actually allow anyone to use a 3rd party library.

Users cannot intercept the code where this value is being used to create DOM content; that is why an option is needed in the first place.

This needs to allow a function to be passed that can be used instead of the built-in method.

Also, I would refrain from naming this `sanitizeTemplate`... perhaps just `sanitize`; this isn't unique to just templates.

It should be required for any plugin that could potentially create/inject insecure markup (e.g. the `content` option in popovers, for example: https://github.com/twbs/bootstrap/blob/v4-dev/js/src/popover.js#L29).

js/src/tools/sanitizer.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * --------------------------------------------------------------------------
+ * Bootstrap (v4.3.0): tools/sanitizer.js
```
markcarver: Why not just add this to the existing https://github.com/twbs/bootstrap/blob/v4-dev/js/src/util.js?

Is it to prevent the potential bloat of individual plugin files that don't require this?

js/src/tools/sanitizer.js
```diff
@@ -0,0 +1,71 @@
...
+      return attr.nodeValue.match(SAFE_URL_PATTERN) || attr.nodeValue.match(DATA_URL_PATTERN)
+    }
+
+    return true
+  }
+
+  return false
+}
+
+export function sanitizeHtml(unsafeHtml, forbiddenTagList, allowedAttributeList) {
```
markcarver: This should use a default value if one isn't supplied (using the `DefaultWhitelist` example from below):

```js
export function sanitizeHtml(unsafeHtml, whitelist = DefaultWhitelist) {
```

js/src/tools/sanitizer.js
```diff
@@ -0,0 +1,71 @@
...
+ * Shoutout to Angular 7 https://github.com/angular/angular/blob/7.2.4/packages/core/src/sanitization/url_sanitizer.ts
+ */
+const DATA_URL_PATTERN = /^data:(?:image\/(?:bmp|gif|jpeg|jpg|png|tiff|webp)|video\/(?:mpeg|mp4|ogg|webm)|audio\/(?:mp3|oga|ogg|opus));base64,[a-z0-9+/]+=*$/i
+
+function allowedAttribute(attr, allowedAttributeList) {
+  const attrName = attr.nodeName.toLowerCase()
+
+  if (allowedAttributeList.indexOf(attrName) !== -1) {
+    if (uriAttrs.indexOf(attrName) !== -1) {
+      return attr.nodeValue.match(SAFE_URL_PATTERN) || attr.nodeValue.match(DATA_URL_PATTERN)
```
markcarver: I think it may make more sense and performant to use the `test` method here, yes?

```js
return SAFE_URL_PATTERN.test(attr.nodeValue) || DATA_URL_PATTERN.test(attr.nodeValue)
```

Otherwise, it should be wrapped so it returns a bool (just so it matches the other returns booleans):
```js
return !!(attr.nodeValue.match(SAFE_URL_PATTERN) || attr.nodeValue.match(DATA_URL_PATTERN))
```

js/src/tooltip.js
```diff
@@ -35,7 +38,10 @@ const DefaultType = {
   offset            : '(number|string|function)',
   container         : '(string|element|boolean)',
   fallbackPlacement : '(string|array)',
-  boundary          : '(string|element)'
+  boundary          : '(string|element)',
+  sanitizeTemplate  : 'boolean',
```
markcarver: There's no line yet in the PR for the code I'm talking about, so I'll just comment it here.

Any option that has to deal with sanitization must be excluded from being able to be supplied by data attributes.

Otherwise, it defeats the whole point of a sanitization method because they could simply disable it entirely or supply their own whitelist (e.g. `data-sanitize="false"`).

Currently, it just merges them in (https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L641):
```js
  _getConfig(config) {
    config = {
      ...this.constructor.Default,
      ...$(this.element).data(),
      ...typeof config === 'object' && config ? config : {}
    }
    // ...
  }
``` 

These needs to be a way to exclude specific data attributes so they can only be supplied via default values or the JavaScript initiator:

```js
  _getConfig(config, excludeData = []) {
    const data = $(this.element).data()
    for (let i = 0, l = excludeData.length; i < l; i++) {
      delete data[excludeData[i]]
    }
    config = {
      ...this.constructor.Default,
      ...data,
      ...typeof config === 'object' && config ? config : {}
    }
    // ...
  }
```

markhalliwell (changes requested):

js/src/tools/sanitizer.js
Johann-S: Yep, it adds a lot of code and it's just for two plugins

js/src/tools/sanitizer.js
Johann-S: It won't be the case

js/src/tooltip.js
Johann-S: Unfortunately we cannot do that, Bootstrap works with data API.

People which use our plugins can force `sanitizeTemplate` if they need it

js/src/tools/sanitizer.js
markcarver: Fair enough 😄

js/src/tooltip.js
markcarver: Then this entire PR is pointless...

site/docs/4.3/components/popovers.md
Johann-S: It's just for the `template` option so we'll keep that name.

If folks choose to disable that option they have to use a third library.

js/src/tooltip.js
Johann-S: Sanitize string on a front library is pointless too, but it can prevent things 👍

As you can see here: https://getbootstrap.com/docs/4.3/getting-started/javascript/#data-attributes

You can disable the data API for specific plugins or for all of our plugins.

Or you can force this value to always be `true` because options passed in JavaScript override those passed by HTML.

```js
$('#myTooltip').tooltip({
  sanitize template: true,
})
```
It will be always true even if it's at false on the HTML with data attribute.

Personally I think it would be a bit strange to disable that 😣

/CC @XhmikosR what do you think about that ?

site/docs/4.3/components/popovers.md
markcarver: Except that isn't __not__ just for the `template` option....

markcarver: Are you even reading my comments?

Johann-S: You're right we should use that for the `content` option if the `html` option is at `true` 👍

js/src/tooltip.js
markcarver: > Sanitize string on a front library is pointless too

No, it's not... it's a requirement if it creates content (like this is doing). That's why is called a vulnerability. Are you even taking this seriously?

> It will be always true even if it's at false on the HTML with data attribute.

Except that is not how secure code __should__ work. It should be __secure by default__. A user shouldn't have to "opt in" for protection.

The reality is that the way this framework handles data attributes are backward, to begin with. Almost any other JS plugin allows data attributes to override global and supplied options. That is because data attributes are a local element configuration.

Because of this backward way of handling data attributes, other code (like http://cgit.drupalcode.org/bootstrap/tree/js/bootstrap.js?h=7.x-3.x#n175) has to supply data attributes on top of their dynamic CMS settings just so they can work at all.

So yes, data attributes could still make their way through; especially if there existing code (e.g. a simple "update" wouldn't actually "secure" anything at all until that manually supplied said forced override... which is absolutely ridiculous).

> Personally I think it would be a bit strange to disable that 😣

Why? There is absolutely no reason sanitization should be configurable at an element level! This is about security and this kind of option should only ever be supplied by trusted sources. Data attributes are not "trusted sources".

js/src/tooltip.js
```diff
@@ -60,7 +66,10 @@ const Default = {
   offset            : 0,
   container         : false,
   fallbackPlacement : 'flip',
-  boundary          : 'scrollParent'
+  boundary          : 'scrollParent',
+  sanitizeTemplate  : true,
+  forbiddenTags     : ['script', 'iframe'],
```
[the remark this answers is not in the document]
markcarver: Edit: updated global ("*") whitelist to include the `dir` and `lang` attributes as these can be applied to any element as well.

js/src/tooltip.js
Johann-S: You cut my sentence but anyway...

We are taking that seriously, we built a solution in one day, that's not a complete solution but we work on it... So stay kind in your words and manners.

I think it's a bit strange because we allow that everywhere in Bootstrap and doing that for just one option feel a bit strange to me, but we can make exceptions.

js/src/tooltip.js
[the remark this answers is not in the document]
markcarver: To clarify why `forbiddenTags` is a bad idea: it's a known anti-pattern for XSS prevention. It's essentially a blacklist which only supports "known vectors of attacks"; it doesn't protect against the _unknown_.

This probably sums this up the best (https://stackoverflow.com/a/46469262):
> We tried blacklists but we kept finding new tags or attributes to use to bypass the blacklist, or malformed html and other encodings were used by bypass filters, making blacklists impractical and ineffective. So now the **default assumption is that if a tag, attribute, or style isn't explicitly specified as safe, then it's unsafe**. This protects not just against the xss attacks we already know about, but many new tyes as well.

js/src/tooltip.js
markcarver: > We are taking that seriously, we built a solution in one day, that's not a complete solution but we work on it

I am certainly impressed, no doubt. But there is a certain level of responsibility once one decides to take on a topic such as this. I know this framework doesn't like to bloat the codebase.

It's partially why I originally suggested simply disabling the data attribute entirely unless a 3rd party library was present. That way this framework wouldn't be responsible for sanitization at all and could offload it to something else; while also keeping security in mind.

I really do appreciate the effort and responsiveness here.

> So stay kind in your words and manners.

I'm sorry. It just felt like you weren't reading my comments and quickly dismissing them. I spent a lot of time reviewing this and writing up suggestions, which have logical explanations and reasoning behind them. I'm very familiar with the FE and XSS related issues; this isn't my first rodeo either 😉 

> I think it's a bit strange because we allow that everywhere in Bootstrap and doing that for just one option feel a bit strange to me, but we can make exceptions.

Security reasons are always an exception.

Johann-S: Even if I dismiss your comments I'm always open to speak about it. BTW some of them I reopened your comments 😉

If you present to me why you suggested that I'll think about it that's sure.
Sometimes when you suggest something to someone you have to add a bit of context why it should be like this.

I don't want to add a third library because folks which use Bootstrap with NPM will still needs to install the third library even if they don't need it. So I made the decision to made a simple sanitizer which can be disabled if folks now how to sanitize themselves their HTML

XhmikosR:
@markcarver: you need to tone down the discussion.

js/src/tooltip.js
markcarver: > BTW some of them I reopened your comments 😉

I appreciate that



> Sometimes when you suggest something to someone you have to add a bit of context why it should be like this.

I have, that's why I wrote a lot.



> I don't want to add a third library

I'm not saying that it should be a requirement. I'm saying that it **could** just be supported. This isn't a real example, but it may be better (language wise) for me to simply give you a simplistic if/else block:

```js
// Support https://jsxss.com (or something else, or multiple, etc.).
if (config.html && typeof window.filterXSS === 'function') {
  // The `data-template`, `data-title` and `data-content`  attributes would
  // be allowed to use `$.html()` or `innerHTML`  by supplying sanitized
  // content using filterXSS().
}
else {
  // The `data-template`, `data-title` and `data-content`  attributes would
  // only be allowed to use `$.text()` or `innerText`  by supplying  unsanitized
  // content based on what was passed. This would essentually "break"
  // `data-template` until they supplied a supported XSS filter library.
}
```

Users could enable it, but it would only truely become enabled whether a supported 3rd party sanitization library is actually present.

Regardless, the LOC bump here is actually quite small. If y'all are happy with supporting a simplistic whitelisted built-in approach, that would most certainly be better. We can always replace it with something a little more hefty if we need.

Johann-S:
Ok things I have to do:

- [ ] Update the docs
- [ ] Sanitize `title` and `content` if `html` is at `true`
- [ ] Not allowing `sanitize` to be set by data attributes in HTML

And I think it'll be good, do not hesite @markcarver if you have any feedbacks, I think I heard you 👍

js/src/tools/sanitizer.js
```diff
@@ -0,0 +1,118 @@
...
+
+  if (allowedAttributeList.indexOf(attrName) !== -1) {
+    if (uriAttrs.indexOf(attrName) !== -1) {
+      return Boolean(attr.nodeValue.match(SAFE_URL_PATTERN) || attr.nodeValue.match(DATA_URL_PATTERN))
+    }
+
+    return true
+  }
+
+  return false
```
markcarver: Just prior to this, it should check for regular expressions. This would actually allow the default global attribute `ARIA_ATTRIBUTE_PATTERN` to work:

```js
// Check if a regular expression validates the attribute.
const regExp = allowedAttributeList.filter(v => v instanceof RegExp);
for (let i = 0, l = regExp.length; i < l; i++) {
  if (attrName.match(regExp[i])) {
    return true
  }
}
```

js/src/tools/sanitizer.js
```diff
@@ -0,0 +1,118 @@
...
+  'poster',
+  'src',
+  'xlink:href'
+]
+
+// The allowedAttribute method should check to see if the supplied value is
+// a regular expression, this would allow this to support any number of aria
+// attributes by default and give users the ability to specify their own regular
+// expressions should the need arise (e.g. /^data-my-app-[\w-]*$/).
+const ARIA_ATTRIBUTE_PATTERN = /^aria-[\w-]*$/i
```
markcarver: This comment was more for the review, it should probably just say:

`Supply a regular expression pattern to detect aria-* attributes.`

Although the documentation could also mention that regular expressions can be passed as values for more custom needs.

js/src/tooltip.js
```diff
@@ -419,18 +427,24 @@ class Tooltip {
...
       } else {
         $element.text($(content).text())
       }
+
+      return
+    }
+
+    if (this.config.html) {
+      content = sanitizeHtml(content, this.config.whiteList)
+      $element.html(content)
```
markcarver: This needs to be wrapped with the sanitize option.

```js
    if (this.config.html) {
      if (this.config.sanitize) {
        content = sanitizeHtml(content, this.config.whiteList)
      }
      $element.html(content)
```

site/docs/4.3/getting-started/javascript.md
```diff
@@ -139,3 +139,68 @@ Bootstrap's plugins don't fall back particularly gracefully when JavaScript is d
...
+## Sanitizer
+
+Tooltips and Popovers use our built-in sanitizer to sanitize options which receive HTML.
+
+The default `whiteList` value is the following:
+
+{% highlight js %}
+var DefaultWhitelist = {
+  // Global attributes allowed on any supplied element below.
+  '*': ['class', 'dir', 'id', 'lang', 'role', ARIA_ATTRIBUTE_PATTERN],
```
markcarver: Might want to replace `ARIA_ATTRIBUTE_PATTERN` with the actual regular expressions value here.

site/docs/4.3/getting-started/javascript.md
```diff
@@ -139,3 +139,68 @@ Bootstrap's plugins don't fall back particularly gracefully when JavaScript is d
...
+myDefaultWhiteList.td = ['data-option']
+{% endhighlight %}
+
+Maybe you want to disable our sanitizer because you prefer to use a dedicated library, for example <a href="https://www.npmjs.com/package/dompurify">DOMPurify</a>
+then you should do the following.
+
+{% highlight js %}
+$('#yourTooltip').tooltip({
+  sanitize: false,
+  template: DOMPurify.sanitize(yourDirtyTemplate),
```
markcarver: While I see why you're doing it this way, it makes it more difficult to actually implement custom (per tooltip) templates.

Before, one could simply do a global Data API implementation (e.g. `$('[data-toggle="tooltip"]').tooltip(...)`) and the template would be imported automatically as a data attribute. It's a rare case when tooltips are initialized individually.

With this current implementation, one would have to literally iterate over each individual tooltip just to extract the template and then sanitize it before initializing the tooltip.

This is highly inefficient, especially considering since the jQuery plugin already implements its own each loop:
```js
$('[data-toggle="tooltip"]').each(() => {
  var $el = $(this);
  $el.tooltip({
    sanitize: false,
    template: DOMPurify.sanitize($el.data('template'))
  })
```

An easier and more efficient approach is to simply support a function being passed to the `sanitize` option:
```js
$('[data-toggle="tooltip"]').tooltip({
  sanitize: DOMPurify.sanitize,
})
```

That way, no matter where the template is supplied, either passed as a JavaScript option upon initialization or via a data attribute, it will always go through the correct sanitization function.

Alternatively, if a sanitizer requires additional options or doesn't support the content being the first parameter, an anonymous function could be used:

```js
$('[data-toggle="tooltip"]').tooltip({
  sanitize: unsafeMarkup => someOtherXssFilter(sanitizeOptions, unsafeMarkup),
})
```

markcarver:
I think the `title` option still needs to be sanitized, didn't see that in the latest code changes yet.

markcarver:
> * [ ]  Not allowing `sanitize` to be set by data attributes in HTML

Not allowing `sanitize` or `whiteList` to be set by data attributes in HTML

Johann-S:
The title is sanitized by the same methods which sanitized `content`

site/docs/4.3/getting-started/javascript.md
Johann-S: sorry I don't understand what do you mean here ? 🤔

markcarver:
Ah, you're right. Forgot that BS4 added `setElementContent`.

site/docs/4.3/getting-started/javascript.md
Johann-S: I prefer to allow passing a function something like:

```js
$('[data-toggle="tooltip"]').tooltip({
  sanitizeFn: unsafeMarkup => someOtherXssFilter(sanitizeOptions, unsafeMarkup),
})
```
it's more flexible for our users

site/docs/4.3/getting-started/javascript.md
markcarver: Because this is documentation and constants don't really many anything here (at least it's not shown what `ARIA_ATTRIBUTE_PATTERN ` actually means).

I just think it would be beneficial to show that this constant (which is really just an internal performance variable) is actually a regular expression.

The verbiage below it should probably explain that users can extend it with their own regular expressions too:
`^data-my-app-[\w-]+`.

Should also probably add a warning that using regular expressions can potientially lead a security vulnerability if the regular expression is too permissive.

For example: while a user could supply ^data-[\w-]+` to allow any data attribute, this may ultimately lead to a potential security risk if another library present on the page has a vulnerability that can be exploited through data attributes.

Essentially, an attacker could use the permissive regular expression defined here to exploit another vulnerability in a different library.

site/docs/4.3/getting-started/javascript.md
markcarver: So you're saying there would be three new options: `sanitize`, `sanitizeFn` and `whiteList`?

That's fine with me. I was just trying to keep it as simple as possible and follow the established option pattern (other options allow functions to be passed to them, no "Fn" suffix).

Johann-S: yep exactly 👍 I pushed what I did

site/docs/4.3/getting-started/javascript.md
```diff
@@ -139,3 +139,74 @@ Bootstrap's plugins don't fall back particularly gracefully when JavaScript is d
...
+
+// To allow table elements
+myDefaultWhiteList.table = []
+
+// To allow td elements and data-option attributes on td elements
+myDefaultWhiteList.td = ['data-option']
+
+// You can push your custom regex to validate your attributes
+// Be carefull about your regular expressions if they are too permissives
+var myCustomRegex = ^data-my-app-[\w-]+
```
markcarver: Sorry. I didn't add this in my original comment, but this should be `/^data-my-app-[\w-]+/`.

markcarver:
Just that tiny doc nit, but other than that I think this looks good!

@Johann-S++

Ty and amazing work!

site/docs/4.3/components/popovers.md
```diff
@@ -250,6 +250,24 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>'scrollParent'</td>
       <td>Overflow constraint boundary of the popover. Accepts the values of <code>'viewport'</code>, <code>'window'</code>, <code>'scrollParent'</code>, or an HTMLElement reference (JavaScript only). For more information refer to Popper.js's <a href="https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.boundariesElement">preventOverflow docs</a>.</td>
     </tr>
+    <tr>
+      <td>sanitize</td>
+      <td>boolean</td>
+      <td>true</td>
+      <td>Enable or disable the sanitization. If activated <code>'template'</code>, <code>'content'</code> and <code>'title'</code> options will be sanitized.</td>
```
markcarver: Oh for both tooltips and popovers, maybe add a note either here stating something to the effect of:

`Note: for security reasons this option cannot be supplied using a data attribute.`

Or if that becomes too verbose in the table, maybe adding something below the table:

```
Note: for security reasons the `sanitize`, `sanitizeFn` and `whiteList` options cannot be supplied using a data attribute.
```

mdo:
Should `js/src/tools/sanitizer.js` live at `js/src/sanitizer.js` instead? Unsure why we need the subdirectory here.

XhmikosR (approved):

XhmikosR:
@mdo: we could do that, but I guess @Johann-S did this to separate the plugins from the other files or something.

laserlemon:
@Johann-S @XhmikosR @mdo 👋 I'm on the team responsible for delivering GitHub vulnerability alerts and could use just a couple of clarifications on this vulnerability:

[CVE-2019-8331](https://nvd.nist.gov/vuln/detail/CVE-2019-8331) reports that _all_ versions `< 4.3.1` are vulnerable. Can you confirm this? I want to make sure that not just the 4.x versions are affected.

It appears that multiple packages are hosted in this repository. Are all three (NPM's `bootstrap`, NuGet's `bootstrap`, and NuGet's `bootstrap.sass`) affected?

Thank you! 💛

XhmikosR:
@laserlemon: yeah it affects all 3.x and all 4.x. For even older releases I have no idea.

The other packages you refer to are only wrappers around this one so they are all affected too.

See also https://github.com/twbs/bootstrap/issues/28330

laserlemon:
Would the `twbs` team like to issue a recommendation to all users of `bootstrap-sass`? At the moment, I've appended the following to the vulnerability's description for Ruby users:

> Users of the `bootstrap-sass` gem might consider upgrading to version 4.3.1 (or later) of the `bootstrap` gem.

XhmikosR:
bootstrap-sass is for v3 so there's only v3.4.1 for them. I guess it makes sense to notify them too.

responsivedevelopment:
Hello are all vulnerabilities fixed for Bootstrap version 4.3.1 and 3.4.1?
Retire JS is still listing both of these versions as being vulnerable: http://retirejs.github.io/retire.js/ 
Also "CVE-2019-8331 Medium Severity Vulnerability detected by WhiteSource pecherskiy-v/bossandhall#18" is still "open" above.
@XhmikosR @markcarver @eoftedal

marcelo321:
hello, is it possible for you to show how can we test the XSS when a page has this vulnerable library?
it would be much appreaciated

MaPePeR:
Is there any special reason why the `style` attribute is not present in the default attribute whitelist? Is there some security issue with it or can it be safely added to the whitelist?

NicoleG25:
Hi @markcarver could you please confirm for me if version 4.2.1 of bootstrap is vulnerable under CVE-2019-8331 ? 
From what I've gathered, I understand that the vulnerable versions are : 
from 3.0.0 (including) up to 3.4.1 (excluding)
from 4.3.x up to 4.3.1 (excluding)

Thanks in advance !

yusup-supriyanto:
a

akhalil-xiq:
Now it is safe to use v4.1.3? @markcarver @Johann-S

PR #28243, opened 2019-02-12
fix: remove 100% width on `.form-control` when is `.flex-grow-1`

dahfazz:
Inside a flex container, one may set a `.flex-grow-1` class on element to give it the available space.

When element is a `.form-control`, that rule is broken because of `width: 100%` property.

I wrote a demo on codepen -> https://codepen.io/dahfazz/pen/PVajJx

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -6,7 +6,6 @@
 
 .form-control {
   display: block;
-  width: 100%;
   height: $input-height;
   padding: $input-padding-y $input-padding-x;
   font-family: $input-font-family;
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
You should use input groups for this:
https://getbootstrap.com/docs/4.3/components/input-group/#button-addons

dahfazz:
True.
I just had a look at `input-group.scss` and think maybe better to look at my proposal in order to remove that hacky `width: 1%` -> https://github.com/twbs/bootstrap/blob/v4-dev/scss/_input-group.scss#L22

PR #28259, opened 2019-02-14
Integration extra small sizing for components

alexandreDavid:
Adding extra small class for each component which already has small and large class.


MartijnCuppens:
Hi @alexandreDavid,

Adding these extra sizes would come in handy, but due to the amount of code and selectors, we'll pass on it right now. We'll first need to look to avoid repetition of similar code like this:
https://github.com/twbs/bootstrap/blob/666e2b2262ebe21f58c468a07c540cdc51bb9a4d/scss/_input-group.scss#L135-L140

Please open an issue next time you want include a feature like this, this would have saved you a lot of work.

PR #28264, opened 2019-02-14
v5: Drop IE10 & IE11

mdo:
**WIP, do not merge yet.**

- Removes IE10/11 from our supported browsers list.
- Removes an IE-specific hack for breadcrumbs.
- Removes the `max-width` from grid classes, though this may need to be revisited because comments indicate it applies to Firefox, too.

### DIFF
--- a/scss/_breadcrumb.scss
+++ b/scss/_breadcrumb.scss
@@ -21,16 +21,6 @@
     }
   }
 
-  // IE9-11 hack to properly handle hyperlink underlines for breadcrumbs built
-  // without `<ul>`s. The `::before` pseudo-element generates an element
-  // *within* the .breadcrumb-item and thereby inherits the `text-decoration`.
-  //
-  // To trick IE into suppressing the underline, we give the pseudo-element an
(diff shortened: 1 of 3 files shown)


.browserslistrc
```diff
@@ -6,7 +6,6 @@ not dead
 Chrome >= 45
 Firefox >= 38
 Edge >= 12
-Explorer >= 10
```
Johann-S: Already did that here: https://github.com/twbs/bootstrap/pull/23586/commits/25a348d4037bdf95ba799b9f43ec10ec969d5666 😄

XhmikosR: Yeah we need to decide on the order of PRs merging.

XhmikosR (approved):

XhmikosR:
I'm gonna merge this in our no jQuery branch to have everything together.

XhmikosR:
OK, I actually had to revert the max-width removal. It seems it's causing issues for me even with Chrome when viewing the docs preview.

PR #28265, opened 2019-02-14
v5: Drop .media

mdo:
The media component is basically a series of utilities that have been assigned to a more unique name. This can all be done via our other utility classes.

- [x] Delete the `_media.scss` file
- [x] Delete the docs file
- [ ] ~Deprecate the component in v4.3.2 /cc @MartijnCuppens (see https://github.com/twbs/bootstrap/pull/28265#issuecomment-463731482)~
- [x] Create a v4 to v5 migration doc, then add this to it with examples.


MartijnCuppens:
> Deprecate the component in v4.3.2 /cc @MartijnCuppens

We can't use the `deprecate()` mixin here since we can only use it in mixins of functions, not in for selectors. I guess we'll just need to document this in the migration doc.

MartijnCuppens:
About the deprecation notice: we could just add a message to the `v4` docs this will be removed.

mdo:
Should be ready for review now, @twbs/css-review.

mdo:
@ysds Done!

XhmikosR:
@MartijnCuppens: LGTY?

MartijnCuppens (approved):

suhailvs:
is there any example of how to create a **media object** using utility class?

martinbean:
@suhailvs A “media” component with a left media and right body:

```html
<div class="d-flex">
  <div class="flex-shrink-0">
    <img src="https://placekitten.com/100" alt="Placekitten">
  </div>
  <div class="flex-grow-1 pl-3">
    <p>An image of a cat.</p>
  </div>
</div>
```

obask:
Could you please add this example into docs, so people won't have to wonder why you dropped media-object and media-list?

PR #28267, opened 2019-02-14
v5: Drop hover query

mdo:
We deprecated it a long time ago, so should be good to clear this out.

- [x] Delete deprecated Sass code
- [x] Add to migration docs page


MartijnCuppens:
@mdo, why did you keep the `hover` and `hover-focus` mixins here?

mdo:
For simplicity's sake perhaps? Can't remember exactly, other than the fact that I believe we use them in a handful of places right now. We can likely get rid of them though if you think it's an improvement.

MartijnCuppens:
A hover mixin which just prints a `:hover` block feels a bit useless. I think we can also drop the `hover-focus` mixin, this might be a little shorter but I guess we'll end up pretty fast with a mixed usage of the mixin and just css.

XhmikosR:
Note to self, this needs a proper rebase.

mdo:
Just pushed another commit that should have tests passing shortly. Should be ready for review after that, @twbs/css-review.

MartijnCuppens (approved):
I just rebased & removed some additional documentation about why we used to have the hover mixin: https://getbootstrap.com/docs/4.3/getting-started/browsers-devices/#sticky-hoverfocus-on-ios

LGTM now

ysds (changes requested):

ysds (approved):

ysds:
LGTM 👍

PR #28276, opened 2019-02-15
Fix white screen when use radio with overflow-scrolling

duyank:
this issue is produced when we use `-webkit-overflow-scrolling: touch` within `z-index: -1` 

Example:
https://jsfiddle.net/fvck7rq3/3/
https://drive.google.com/file/d/15HNgzmi8fX7_CPte9pS7ZA_MYX6fPpAp/view

**Device**
iPhone XS / IOS 12.0
iPhone 7 / 12.1.1
iPhone 8 / 12.1.4 (16D57)

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -21,7 +21,7 @@
 
 .custom-control-input {
   position: absolute;
-  z-index: -1; // Put the input behind the label so it doesn't overlay text
+  pointer-events: none;
   opacity: 0;
 
   &:checked ~ .custom-control-label::before {


mdo:
Your fiddle is missing an `id` on the `<input>` and a `for` attribute on the `<label>`. Add those to connect the label to the radio and you should be set.

duyank:
@mdo  `id` and `for` are not related to this issue. So if I add them or not it still error. Did you check it on the device that I noted?

duyank:
@patrickhlauke @XhmikosR could you take a look?

Issue #28227, opened 2019-02-11
Crossfade carousel with transparent background


Daegaladh:
- Windows 10 x64
- Chrome 72.0.3626.96 (64 bits)
- Firefox 65.0 (64-bit)

In v4.2.1 crossfade carousel performs a sudden transition when changing between images.
This is not visible with opaque backgrounds, but looks really ugly when working with transparent backgrounds or images.

This bug was introduced in 5aed968750177ec35f99dc03e71eec3632beb844

- Example: https://jsbin.com/xomelak/2

jahanzaibsuleman07:
Hi @XhmikosR, I have created the [PR](https://github.com/twbs/bootstrap/pull/28278) for this fix. This is my first PR for bootstrap. Please review. Thanks!

mdo:
Closing per closed PR.

PR #28278, opened 2019-02-15
Fixed transition delay for carousel fade

jahanzaibsuleman07:
Closes https://github.com/twbs/bootstrap/issues/28227.

Transition duration was getting `0` value. Fixed the issue by setting transition delay value to `0` and setting transtion duration to `$carousel-transition-duration`.


MartijnCuppens:
This will fade fade out the old carousel item before the new item is faded in, which isn't the behaviour we want. Compare these carousels to check out what I mean:

Your implementation:
https://deploy-preview-28278--twbs-bootstrap.netlify.com/docs/4.3/components/carousel/#crossfade

Current implementation:
https://getbootstrap.com/docs/4.3/components/carousel/#crossfade

jahanzaibsuleman07:
Hi @MartijnCuppens : Actually the [Issue](https://github.com/twbs/bootstrap/issues/28227) states that the images with transparent background in that case disappears immediately and looks ugly. If we persist old slide to stay there before the new one fades in, it will eventually be the same. Can you please suggest me the desired behavior for this case so that I can update my PR? Thanks!

This will be my first contributing PR for bootstrap and want to contribute more :)

mdo:
@jahanzaibsuleman07 I believe you need to set a `background-color` or something else instead of modifying our source files for all images. I'm in agreement with @MartijnCuppens—the current version in our docs is the desired implementation here.

PR #28285, opened 2019-02-15
Updated broken contributing links in v3 README

seanfahey:


XhmikosR:
See https://github.com/twbs/bootstrap/pull/28286#issuecomment-464124266

seanfahey:
Agree with the comment as it applies to the other PR, but I was thinking v3's contributing link in the readme should still point to `master` at its new location, and not be mass-replaced with `v3-dev`.

XhmikosR:
v3-dev has diverted so much already. Plus I don't think we'll bother a lot more with 3.x.

PR #28297, opened 2019-02-18
Update link to CONTRIBUTING.md in README.md

patrickhlauke:
since the former is now contained in the `.github` subdirectory

(we probably want this in master, v4 and v5 branches...do we cherry-pick it into those?)


XhmikosR:
Not a cherry pick because then the hashes will change. It needs to be a manual merge commit.

XhmikosR (approved):

PR #28299, opened 2019-02-18
Remove incorrect aria-haspopup from dropdown toggles

### LINKED ISSUE #28295: Dropdown does not conform to ARIA authoring practices
The dropdown widget (as of 4.3.1) does not conform to the [ARIA 1.1 specification](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup). That should present some accessibility concerns.

Per the ARIA 1.1 spec, when a button with aria-haspopup is set to true, the popup itself **MUST** have a role=menu, and should conform to the [ARIA specification for the menu role](https://www.w3.org/TR/wai-aria-1.1/#menu). In general, the descendant options should have such roles as role=menuitem.

One solution would be simply adding the role=menu and role=menuitem to the dropdown and dropdown items. I'm hesitant to recommend that, however, as there is dropdown content such as headers, dividers, text, forms, etc. The ARIA spec isn't explicit, but I don't read it as allowing for the inclusion of that kind of content within an ARIA menu. None of the [menu button examples in the ARIA 1.1  authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/#menubutton) include such content, for example. 

Further, if you add role=menu and role=menuitem to the dropdown and dropdown items, you should take the menu items out of the tab focus order. The convention with ARIA widgets like this is that [composite UI components should have only one focusable element, and arrow keys move focus within the component](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_general_within).

The other option may be to remove the aria-haspopup=true attribute altogether from the button altogether. In doing so, the dropdown stops following a menu button widget pattern and starts following a [disclosure widget pattern](https://www.w3.org/TR/wai-aria-practices-1.1/#disclosure). 

In contrast to the menu, if you go this route, you may want to remove support for arrows to navigate among items, as there's no convention to moving focus through the focusable items of a disclosure using arrow keys.

A third option is instead of implementing half of each ARIA widget into the dropdown component is to create more discreet widgets, each of which adheres to the ARIA patterns more closely.

I'm not sure what to recommend, since it might mean rethinking how the component would work. But whatever direction you choose, you should become more in line with the ARIA spec than you are now.

patrickhlauke:
`aria-haspopup` use requires the use of an ARIA `menu`, `listbox`, `tree`, `grid` or `dialog` (see https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup)

in our use (as generic disclosure widgets), it's incorrect and sets up the wrong assumption/behavior from assistive technologies.

in future, we likely want to change our dropdowns to essentially be like a `dialog` - move focus to it when opened, possibly make it modal so focus is maintained inside it and the rest of the page
is hidden/inert.

but for now, removing the incorrect attribute is probably the quickest, most immediate fix.

Closes #28295

### DIFF
--- a/js/tests/unit/dropdown.js
+++ b/js/tests/unit/dropdown.js
@@ -727,7 +727,7 @@ $(function () {
     var done = assert.async()
     var html = '<nav class="navbar navbar-expand-md navbar-light bg-light">' +
         '<div class="dropdown">' +
-        '  <a class="nav-link dropdown-toggle" href="#" id="dropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">Dropdown</a>' +
+        '  <a class="nav-link dropdown-toggle" href="#" id="dropdown" data-toggle="dropdown" aria-expanded="false">Dropdown</a>' +
         '  <div class="dropdown-menu" aria-labelledby="dropdown">' +
         '    <a class="dropdown-item" href="#">Action</a>' +
         '    <a class="dropdown-item" href="#">Another action</a>' +
--- a/js/tests/visual/dropdown.html
+++ b/js/tests/visual/dropdown.html
@@ -28,7 +28,7 @@
               <a class="nav-link" href="#">Link</a>
             </li>
             <li class="nav-item dropdown">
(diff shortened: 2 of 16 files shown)


Johann-S (approved):

XhmikosR (dismissed):
blocking until we sort the branches

PR #28302, opened 2019-02-18
Remove --save option as it isn't required anymore

stevemoser:
"As of npm 5.0.0, installed modules are added as a dependency by default, so the --save option is no longer needed. The other save options still exist and are listed in the documentation for npm install."

https://stackoverflow.com/a/19578808/142358


XhmikosR:
Unfortunately, node.js 6.x is still supported and that ships with node 3.x.

XhmikosR:
Reopening as we are going to require Node >= 8 on master.

XhmikosR (approved):

Johann-S:
@XhmikosR shouldn't we remove `jquery` here ?

XhmikosR:
Not in master, in your no jquery branch these changes should be made.

PR #28317, opened 2019-02-19
[WIP] Bump supported browsers for v5

### LINKED ISSUE #28310: Revisit our .browserslistrc for v5

mdo:
WIP while we discuss, but wanted to use the PR to get the conversation going.

- Chrome history: https://en.wikipedia.org/wiki/Google_Chrome_version_history
- Firefox latest ESR (v60): https://www.mozilla.org/en-US/firefox/organizations/
- Edge history: https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history
- Desktop Safari history: https://en.wikipedia.org/wiki/Safari_version_history
- Opera history: https://help.opera.com/en/opera-version-history/

Fixes #28310


.browserslistrc
```diff
@@ -3,11 +3,10 @@
...
 not dead
-Chrome >= 45
-Firefox >= 38
-Edge >= 12
-Explorer >= 10
-iOS >= 9
-Safari >= 9
-Android >= 4.4
-Opera >= 30
+Chrome >= 70
```
XhmikosR: I wouldn't go there. IIRC the Google search bot is based on Chrome 47/48 or something like that. I need to find the source for this though.

.browserslistrc
```diff
@@ -3,11 +3,10 @@
...
+Firefox >= 60
```
XhmikosR: 38 was the LTS at the time. Just reminding you.

.browserslistrc
```diff
@@ -3,11 +3,10 @@
...
+Edge >= 15.15063
+iOS >= 10
+Safari >= 10
+Android >= 5
+Opera >= 50
```
XhmikosR: I don't know if we should have Opera here at all. Opera is a Chromium based browser now.

XhmikosR:
Also, perhaps it'd be wise to exclude some browsers explicitly.

Try playing with https://browserl.ist/?q=%3E%3D+1%25%2C+last+1+major+version%2C+not+dead%2C+Chrome+%3E%3D+45%2C+Firefox+%3E%3D+38%2C+Edge+%3E%3D+12%2C+iOS+%3E%3D+9%2C+Safari+%3E%3D+9%2C+Android+%3E%3D+4.4%2C+Opera+%3E%3D+30

.browserslistrc
```diff
@@ -3,11 +3,10 @@
...
-iOS >= 9
-Safari >= 9
-Android >= 4.4
-Opera >= 30
+Chrome >= 70
+Firefox >= 60
+Edge >= 15.15063
+iOS >= 10
+Safari >= 10
+Android >= 5
```
XhmikosR: I'd go with 6 here. It was released in 2015.

.browserslistrc
mdo: I think it's fine to be explicit though? Each browser has different functionality that comes with major versions, so we might want to review some of the milestones there to ensure we're not shooting ourselves in the foot.

XhmikosR: Yes, but why make an exception for Opera? I mean, we don't do the same for the other Chromium-based browsers.

BTW just thinking out loud here.

.browserslistrc
XhmikosR: Found it: https://developers.google.com/search/docs/guides/rendering

>Googlebot uses a web rendering service (WRS) that is based on Chrome 41 (M41)

Maybe by the time v5 stable is out this has changed, but until then, we shouldn't do this.

thecodingdude:
Is IE11 being kept around? Issue #28349 removes some IE leftovers but there's queries as to whether Bootstrap should work at all in IE11.

Personally, it'd be easier to keep V4 for IE11 and use V5 (and beyond) to target more modern browsers. Now jQuery is being dropped it's possible to look into ESM (and maybe the JS can be shipped without being transpiled) but with IE11 that won't be possible.

voltaek:
The issue with keeping v4 for IE11 is it also supports IE10. I'm hoping to see v5 work decent enough on IE11 with the recommended polyfills that I'll be able to justify moving to it instead of having to stick with v4 for the foreseeable future.

.browserslistrc
MisatoTremor: ESR is 60 since May 2018, isn't it?

derdiggn: The current ESR version is 60

.browserslistrc
tkrotoff: FYI *"The new evergreen Googlebot [...] now runs the latest Chromium rendering engine (74 at the time of this post) [...] Moving forward, Googlebot will regularly update its rendering engine"*

Source: https://webmasters.googleblog.com/2019/05/the-new-evergreen-googlebot.html

And https://developers.google.com/search/docs/guides/fix-search-javascript does not specify anymore the Chrome version used by Googlebot

sebaherrera:
Hi. Since Android 5, Webview is available at its own apk (based in Chromium 75.0.3770.101 as of today), so it will be useful to support that OS version as well.

PR #28322, opened 2019-02-20
doc - Fix some accessibility issues

Lausselloic:
This PR goal is to fix some accessibility issues in actual documentation : 

- bad hierarchy title in dropdowns component doc
- svg are missing accessible title on blog and product
- define secondary navigation as `nav` landmark in documentation pages

let @patrickhlauke check it


site/_layouts/docs.html
```diff
@@ -15,9 +15,9 @@
         </div>
 
         {% if page.toc %}
-          <div class="d-none d-xl-block col-xl-2 bd-toc">
+          <nav class="d-none d-xl-block col-xl-2 bd-toc">
```
patrickhlauke: Bit conflicted about this. Not every set of links necessarily needs to be wrapped in a `<nav>`. If we were to make this change, we should probably also want to wrap the links in the `<header>` in a `<nav>` ? And, when there are multiple `<nav>`s, you really should start to name them (e.g. using `aria-label="..."`>.

site/docs/4.3/examples/blog/index.html
```diff
@@ -18,7 +18,7 @@
       </div>
       <div class="col-4 d-flex justify-content-end align-items-center">
         <a class="text-muted" href="#">
-          <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="mx-3" role="img" viewBox="0 0 24 24" focusable="false"><title>Search</title><circle cx="10.5" cy="10.5" r="7.5"/><path d="M21 21l-5.2-5.2"/></svg>
+          <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="mx-3" role="img" viewBox="0 0 24 24" focusable="false" aria-labelledby="svg-search-title"><title id="svg-search-title">Search</title><circle cx="10.5" cy="10.5" r="7.5"/><path d="M21 21l-5.2-5.2"/></svg>
```
patrickhlauke: advantage of this over just adding an `aria-label="..."` on the link?

site/docs/4.3/examples/product/index.html
```diff
@@ -7,7 +7,7 @@
 <nav class="site-header sticky-top py-1">
   <div class="container d-flex flex-column flex-md-row justify-content-between">
     <a class="py-2" href="#">
-      <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="d-block mx-auto" role="img" viewBox="0 0 24 24" focusable="false"><title>Product</title><circle cx="12" cy="12" r="10"/><path d="M14.31 8l5.74 9.94M9.69 8h11.48M7.38 12l5.74-9.94M9.69 16L3.95 6.06M14.31 16H2.83m13.79-4l-5.74 9.94"/></svg>
+      <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="d-block mx-auto" role="img" viewBox="0 0 24 24" focusable="false" aria-labelledby="svg-product-title"><title id="svg-product-title">Product</title><circle cx="12" cy="12" r="10"/><path d="M14.31 8l5.74 9.94M9.69 8h11.48M7.38 12l5.74-9.94M9.69 16L3.95 6.06M14.31 16H2.83m13.79-4l-5.74 9.94"/></svg>
```
patrickhlauke: advantage of this over just adding an `aria-label="..."` on the link?

site/docs/4.3/components/dropdowns.md
```diff
@@ -821,7 +821,7 @@ $('.dropdown-toggle').dropdown()
 {% endhighlight %}
 
 {% capture callout %}
-##### `data-toggle="dropdown"` still required
+#### `data-toggle="dropdown"` still required
```
patrickhlauke: Yup, this change is quite non-controversial and correct

site/docs/4.3/components/dropdowns.md
```diff
@@ -662,7 +662,7 @@ Add a header to label sections of actions in any dropdown menu.
 
 {% capture example %}
 <div class="dropdown-menu">
-  <h6 class="dropdown-header">Dropdown header</h6>
+  <h4 class="dropdown-header">Dropdown header</h4>
```
patrickhlauke: i seem to remember a recent discussion about having these as `<h6>` to avoid having them show up in the TOC, but can't seem to find it just now.

these headings are, weirdly, not really part of the document structure per se, but in an example...but as HTML has no way of defining this difference (I guess if it were an iframe, perhaps it would be more logical in a way if the heading there in the separate example has, say, `<h1>` or something...). As it's not a huge problem (nobody dies) that this is `<h6>` rather than `<h4>`, I think we can live with the way it currently is (to avoid it showing in the TOC)

patrickhlauke (changes requested):
* conflicted about the use of `<nav>`
* change of the `<h6>` is (from memory) not a good idea as it will make it show up in the TOC (currently, until we switch to a different site generator solution)
* that one heading level change in the markdown - yup, that's right
* shorter/simpler to add `aria-label` to links around those `<svg>`s ?

XhmikosR:
Please do not change any headers. I'll tackle this later myself.

site/docs/4.3/components/dropdowns.md
Lausselloic: I agree nobody will die :-)
There's a config in _config.yml that prevent example heading to be displayed in toc
```
toc:
  min_level:        2
  max_level:        4
  no_toc_section_class:
    - "bd-callout"
    - "bd-example"
```

site/docs/4.3/examples/product/index.html
Lausselloic: I've not tested with aria-label. I will made a codepen.
I've propose that solution based on https://developer.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/ but that doesn't discuss about svg used in a link

XhmikosR:
https://github.com/twbs/bootstrap/commit/a280f2a8bb1f1f7d54aafcd47a6e105f5dbd8e4f

But this will need to be adapted after the Hugo switch, and I'll probably need to do it from scratch then.

site/_layouts/docs.html
Lausselloic: Yes I need to name the two navs with aria-label.
Don't need to wrap links in header because they're already in a landmark. The purpose of this commit was to put secondary navigation link in a landmark.

Lausselloic:
I could remove the Hx update from my PR to prevent conflict during hugo migration. And I will update the PR with aria-label on nav elements and make some tests about aria-label on link versus aria-labelledby on svg element

site/docs/4.3/examples/product/index.html
Lausselloic: First tests give the advantage to aria-label on the link that prevent to vocalize the type of the content, but impose to duplicate the title (no impact in our case)
test NVDA 2016.2.1 / chrome 72 : test NVDA 2016.2.1 / firefox ESR 52.7.2 : test Jaws 17 / IE11 : Same result
labelledby on the SVG say : Search graphique lien (in french)
aria-label on the link say : Search lien (in french)

If you've more feedback with voiceover/talkback/narrator let me know

site/docs/4.3/components/dropdowns.md
XhmikosR: Please revert this too.

Lausselloic (commented):
I've update my PR wonder it's better now?

XhmikosR:
Nope, there are conflicts. You need to rebase properly.

site/docs/4.3/examples/product/index.html
Lausselloic: anyone had made some more test?
What solution have your favour @patrickhlauke ? labelledby or aria-label? Let me know if I need to update the PR

XhmikosR:
A merge is not a rebase FYI.

Lausselloic:
Yes I know that a merge isn't a rebase, but it works like a charm no?

Lausselloic:
I've split that PR into 2 smallest one, with clean branch

XhmikosR:
No it doesn't work, because it doesn't allow us to rebase and merge. We can do squash and merge or a merge commit. The later we don't do anymore.

PR #28328, opened 2019-02-20
[css] Fix breaking layout breaking error in code block

wackerm:
### Problem: Code block does not break long lines
[see miminal example](https://jsbin.com/dabopijehi/edit?html,output)

Was before:
`word-break: break-word;`
Should be (in my Opinion):
`word-wrap: break-word;`


wackerm:
This change could/should also be done with current master...

The break-word option for word-break ist not supported by all browsers (for e.g. Firefox), [see ](https://developer.mozilla.org/de/docs/Web/CSS/word-break#Browserkompatibilit%C3%A4t), the break-word option with word-wrap is supported by all major browsers [see](https://developer.mozilla.org/de/docs/Web/CSS/word-wrap#Webbrowserkompatibilit%C3%A4t) and does fix the issue.

XhmikosR:
Please retarget against master.

wackerm:
Done:  retarget to master branch

XhmikosR:
Friendly ping @twbs/css-review

MartijnCuppens (approved):

XhmikosR:
Should this be backported to v4?

wackerm:
I think that would be great since not everyone will upgrade existing v4 projects...
Shall i submit this patch as new PR against v4?

XhmikosR:
Nah, if we are going to backport it, I'll cherrypick it.

MartijnCuppens:
> Should this be backported to v4?

Yup, it's a bug

PR #28332, opened 2019-02-21
Add 404 page (for Hugo)

### LINKED ISSUE #28319: Custom 404 Page for Docs
Currently the 404 page is the default GitHub Pages one. It'd be nice to have a custom Bootstrap branded one with links in header and footer so any folks reaching the page can quickly get to where they want to go on the site. Jekyll has support for this: https://jekyllrb.com/tutorials/custom-404-page/

coliff:
Fixes #28319

### DIFF
--- a/site/content/404.md
+++ b/site/content/404.md
@@ -1,11 +1,12 @@
 ---
 title: "404 - File not found"
-layout: "default"
 description: ""
+url: /404.html
+robots: noindex,follow
 sitemap_exclude: true
 ---
 
 <div class="text-center py-5">
-<h1 class="display-1">404</h1>
-<h2>File not found</h2>
+  <h1 class="display-1">404</h1>
+  <h2>File not found</h2>
 </div>
--- a/site/content/404.md
+++ b/site/content/404.md
@@ -1,0 +1,11 @@
+---
+title: "404 - File not found"
+layout: "default"
+description: ""
+sitemap_exclude: true
+---
+
+<div class="text-center py-5">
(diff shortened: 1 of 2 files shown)


XhmikosR:
Yeah this needs more work.

1. the output should be 404.html
2. the layout that's used now makes no sense to me
3. Just showing that a page isn't found, isn't very useful IMO. There should be a back to home button and ideally the search.

For the third thing, don't worry about it, we can tweak it later.

XhmikosR:
@bardiharborow @MartijnCuppens thoughts? https://deploy-preview-28332--twbs-bootstrap.netlify.com/404.html

I will try fixing the 404 layout later/tomorrow but I think we need to improve the page, maybe add search or at least a button to the homepage.

MartijnCuppens:
I pushed the footer to the bottom of the page, it was weird having some white space below the footer.

XhmikosR:
How about content :)

BTW see https://github.com/opencollective/opencollective/issues/1764

bardiharborow (approved):
LGTM, there's probably some room to be more creative but no need to delay merging this.

coliff:
Thanks for finishing the work on this PR @XhmikosR  - looks good - and a big improvement over the default GitHub Pages 404. The footer looks much better - thanks @MartijnCuppens !

I was thinking it might not be a good idea to include a search as a user might have been looking for a Bootstrap v3 class in which case the search (if it was search for v4 may not show results)... and if it were to combine search results for all versions it could get confusing.

PR #28337, opened 2019-02-21
Remove DOMContentLoaded.

XhmikosR:
There's no reason for us to wait here.

### DIFF
--- a/js/tests/visual/modal.html
+++ b/js/tests/visual/modal.html
@@ -217,86 +217,83 @@
     <script src="../../dist/tooltip.js"></script>
     <script src="../../dist/popover.js"></script>
     <script>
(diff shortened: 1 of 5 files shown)


js/tests/visual/modal.html
```diff
@@ -217,8 +217,9 @@ <h4 class="modal-title" id="slowModalLabel">Lorem slowly</h4>
     <script src="../../dist/tooltip.js"></script>
     <script src="../../dist/popover.js"></script>
     <script>
-      var firefoxTestDone = false
-      document.addEventListener('DOMContentLoaded', function () {
+      var firefoxTestDone = false;
+
+      (function () {
```
XhmikosR: Need to check if this is needed to be in an IIFE

Johann-S (approved):

PR #28339, opened 2019-02-21
v5: Drop all print styles and classes

mdo:
Removes the print styles, print utilities, and more. Far too many bugs around this stuff for us to keep maintaining as so much of it depends on the exact page design and layout.


MartijnCuppens:
I wouldn't drop `.d-print-none`, that's a pretty handy utility class.

mdo:
Should be ready for review now, @twbs/css-review.

site/content/docs/4.3/utilities/display.md
```diff
@@ -75,24 +75,24 @@ To show an element only on a given interval of screen sizes you can combine one
...
-- `.d-print-flex`
-- `.d-print-inline-flex`
-
-The print and display classes can be combined.
-
-{{< example >}}
-<div class="d-print-none">Screen Only (Hide on print only)</div>
-<div class="d-none d-print-block">Print Only (Hide on screen only)</div>
-<div class="d-none d-lg-block d-print-block">Hide up to large on screen, but always show on print</div>
+## Display in print
```
MartijnCuppens: @mdo, could you remove the trailing spaces here?

MartijnCuppens (approved):

site/content/docs/4.3/components/navbar.md
```diff
@@ -14,7 +14,6 @@ Here's what you need to know before getting started with the navbar:
 - Navbars and their contents are fluid by default. Use [optional containers](#containers) to limit their horizontal width.
 - Use our [spacing]({{< docsref "/utilities/spacing" >}}) and [flex]({{< docsref "/utilities/flex" >}}) utility classes for controlling spacing and alignment within navbars.
 - Navbars are responsive by default, but you can easily modify them to change that. Responsive behavior depends on our Collapse JavaScript plugin.
-- Navbars are hidden by default when printing. Force them to be printed by adding `.d-print` to the `.navbar`. See the [display]({{< docsref "/utilities/display" >}}) utility class.
```
XhmikosR: Isn't this still relevant, though?

XhmikosR: OK probably not.

XhmikosR:
So, just a last comment. Are we 100% sure about this? I mean, if you just do a print preview of our docs and compare you will see how bad it is now.

MartijnCuppens:
It's actually better here (Chrome, mac) and on Firefox it's equally bad. Our print styles fix some things while introducing other issues. We might have a look at reintroducing print style sheets or optimizing our docs if we know for sure we don't break anything else.

PR #28345, opened 2019-02-22
move util in a util folder with the sanitizer

Johann-S:
I hope @XhmikosR won't be angry against me but... I like to break everything and it's just the beginning 😈


Johann-S (commented):
it generate a strange issue when you hide a collapse 🤔 
https://deploy-preview-28345--twbs-bootstrap.netlify.com/docs/4.3/components/collapse/#multiple-targets

js/src/collapse.js
```diff
@@ -224,8 +233,6 @@ class Collapse {
 
     this._element.style[dimension] = `${this._element.getBoundingClientRect()[dimension]}px`
 
-    Util.reflow(this._element)
```
XhmikosR: Maybe this is the cause for the collapse issue?

Johann-S: No it was a little trick to force IE to refresh the UI

Johann-S:
I fixed the collapse plugin 👍

XhmikosR (approved):

PR #28348, opened 2019-02-23
Remove `role="main"` from `main` element.

XhmikosR:
This was needed for Internet Explorer 11.


bardiharborow (dismissed):
About time...

XhmikosR:
Putting this on hold until we have a consensus about IE support.

patrickhlauke (commented):
I'd say keep it, as it's a very small concession that can make a big difference for users that still have to use IE11 with AT. but not overly fussed tbh

XhmikosR:
We need to come to a consensus about IE support. When this happens, we'll decide how we proceed.

MartijnCuppens:
I agree with @XhmikosR, we must be strict which will make it way easier for us to make decisions about things.

XhmikosR:
Closing since we are going to keep IE11 support in v5.

Issue #25100, opened 2017-12-28
What would be the advantages of dropping IE10?


Herst:
Thinking aloud, probably just relevant for post-4.0 (right, @mdo?):

* #23770 could be reverted, possibly improving JS minimization and be a micro-optimization for `undefined` checks
* Notes about missing `pointer-events` (as added in e.g. #23826) could be removed.
* The `pointer-events` workaround for #23222 might be worth checking out again.
* SVG images fix not needed any more (#23476)
* Some of the notes about disabled `<a>` buttons could be possibly removed
* #22493?
* Doing `flex-basis: 100%` instead of `flex-grow: 1` might not be needed any more (#23652).
* No need for `-ms-` prefix for flexbox properties (as noted by @alecpl, thanks!), dropping `-ms-` prefix for `touch-action`

_Can I use_ comparison: <http://caniuse.com/#compare=ie+10,ie+11>
IMHO, most noteworthy from there:
* `pointer-events`
* `hidden` attribute
* `const` and `let` (not whole ES6 though, no idea whether Babel can make use of this)

Some other possibly relevant things:
* Various touch improvements, sadly most of them have to do with the pointer events API, but also various touch related (but mostly IE/Edge specific) CSS properties: 
   * https://msdn.microsoft.com/library/bg182636.aspx
   * https://blogs.msdn.microsoft.com/ie/2013/07/31/ie11-touch-browsing-for-todays-web-and-beyond/
   * https://blogs.msdn.microsoft.com/ie/2013/10/22/enabling-new-interoperable-panning-experiences-through-the-css-scrolling-snap-points-specification/
---
(Concerning desktop browser usage, it is below the .45% of Edge16 and .59% of IE8 according to <http://gs.statcounter.com/browser-version-partially-combined-market-share/desktop/worldwide/#monthly-201711-201711-bar>, with no significant numbers in any continent. It's also considered unsupported by Microsoft on any desktop end-user operating system, see <https://support.microsoft.com/help/17454>, under _"What is the Lifecycle Policy for Internet Explorer?"_; Other UI libs which don't support IE10: [[1]](https://jqueryui.com/browser-support/), [[2]](https://github.com/Dogfalo/materialize#supported-browsers), [[3]](https://github.com/uikit/uikit#browser-support), [[4]](https://material-ui-next.com/getting-started/supported-platforms/))

alecpl:
No need for `-ms-` prefix for flexbox properties - smaller css output.

[...]

PR #28349, opened 2019-02-23
Remove Internet Explorer leftovers.

XhmikosR:
**Untested!**

There are more cases in _modal.scss but I thought someone else should tackle those.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -258,11 +258,6 @@
     color: $custom-select-disabled-color;
     background-color: $custom-select-disabled-bg;
   }
-
-  // Hides the default caret in IE11
-  &::-ms-expand {
-    display: none;
-  }
 }
 
 .custom-select-sm {
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -24,12 +24,6 @@
   @include box-shadow($input-box-shadow);
   @include transition($input-transition);
 
-  // Unstyle the caret on `<select>`s in IE10+.
-  &::-ms-expand {
-    background-color: transparent;
-    border: 0;
-  }
-
   // Customize the `:focus` state to imitate native WebKit styles.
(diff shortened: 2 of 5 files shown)


XhmikosR:
BTW maybe we should leave these? On the other hand we removed a few IE specific CSS stuff already.

The idea was if one wanted to support IE, they could add the prefixes and the polyfills. So I'm not sure which way to go. But we should be consistent.

patrickhlauke:
It would be nice if a site didn't completely break/explode in IE11. so walking the fine line of not trying to patch things to work in IE11, but have it at least present content in a semi-sensible way...

XhmikosR:
Yeah, I don't know myself either. Because if we are going the way you suggest, which is fine by me, we should actually revert a couple of changed that have landed on master.

Herst:
But at least nobody needs to care about IE10: #25100

XhmikosR:
Yeah, the discussion is for IE 11.

ray73864:
Would love to have used Bootstrap 5 when it comes out, but the lack of IE11 support would be a deal-breaker for me unfortunately.

In our business, all our stores log onto a series of terminal servers (our admin people log onto another one), a store might have 5 thin clients and all 5 are logged onto the same user, the unfortunates of Chrome, Firefox, etc... is that they don't support session-profiles, only user-profiles, so with 5 logged into the same user, the first one to open Chrome/FF sets the lock on the profile and both require kludges to get them to be semi-useful.

We have in the past attempted Chrome, but it's never been 100% successful. IE on the other hand supports user-session profiles, so each logged in session has a slightly separate way of using the browser (especially from the task-manager POV).

If there was a stylesheet + polyfills for V5 that allow it to continue to work on IE11 that would be super-awesome, then I could get rid of jQuery completely.

XhmikosR:
Closing since we are going to keep IE 11 support for v5.

PR #28352, opened 2019-02-24
Responsive sticky top

### LINKED ISSUE #28311: Sticky based on viewport
Would be lovely with the possibility to add the "sticky-top" class based on viewport.
Like for an example: sticky-top, sticky-sm-top, sticky-md-top, sticky-lg-top sticky-xl-top, and same for sticky-bottom.

Couldn't find this as a feature request, so here it is.

The good thing is, it wouldn't break any existing sites :)

jahanzaibsuleman07:
This is solution for creating responsive sticky top class which closes #28311.

cc: @mdo

### DIFF
--- a/site/content/docs/4.3/components/scrollspy.md
+++ b/site/content/docs/4.3/components/scrollspy.md
@@ -307,7 +307,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>method</td>
       <td>string</td>
       <td>auto</td>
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Hi @jahanzaibsuleman07 

- Could you move `@supports (position: sticky)` around the loop, this prevents the `supports` query from being generated multiple times
- This needs to be documented in our docs
- We also need documentation for if you want to use the mixin below a specific size. I guess we'll need to work with the `.position-static` utility class then?

jahanzaibsuleman07:
> Hi @jahanzaibsuleman07
> 
> * Could you move `@supports (position: sticky)` around the loop, this prevents the `supports` query from being generated multiple times
> * This needs to be documented in our docs
> * We also need documentation for if you want to use the mixin below a specific size. I guess we'll need to work with the `.position-static` utility class then?

Hi @MartijnCuppens:

I have addressed the first two points. Please review.
For third point, do you want something in the documentation like below?

```
**Mixins**

Sass mixin can be used to create a class for a specific viewport and below:

 @include media-breakpoint-down(sm) {
      $infix: breakpoint-infix(sm, $grid-breakpoints);

      .position#{$infix}-static {
        position: static !important;
      }
    } 

```

jahanzaibsuleman07:
Hi @MartijnCuppens and @mdo, do we need any further changes here?
Also please see my previous comment. Thanks!

PR #28360, opened 2019-02-26
docs- add an explicit aria-label on link with svg on sample pages

Lausselloic:
splitting first PR (https://github.com/twbs/bootstrap/pull/28322) into small PR.

After some tests, according to @patrickhlauke feedback looks like aria-label is a better solution than labelledby for the doc purpose


patrickhlauke (approved):

XhmikosR:
@patrickhlauke : you want this and #28361 backported to v4?

patrickhlauke:
@XhmikosR yes please

PR #28362, opened 2019-02-26
Fix position: sticky

ilcomp:
.sticky-top is deleted after compilation.

### DIFF
--- a/scss/utilities/_position.scss
+++ b/scss/utilities/_position.scss
@@ -23,8 +23,8 @@
   z-index: $zindex-fixed;
 }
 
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
No, it's not:
https://github.com/twbs/bootstrap/blob/f7b55da4509e62b559f1e8181cf4b03e964775a4/dist/css/bootstrap.css#L7796-L7803

ilcomp:
.sticky-top is removed, leaving the wrong css code, if you use only the scss compiler

PR #28365, opened 2019-02-26
Fix border if only one card

MartijnCuppens:
Closes #28060.

I'm going to do some checks if Netlify has finished, so currently in draft

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -273,7 +273,7 @@
       }
     }
 
-    &:first-of-type {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Ok, looks good. 
Before:
https://codepen.io/MartijnCuppens/pen/aMzqoB
After:
https://codepen.io/MartijnCuppens/pen/LaEQPL

XhmikosR (approved):

PR #28386, opened 2019-03-01
build bootstrap in esm

Johann-S:
I continue my todo list for v5.

Our main dist file will be available as a module ~and next step we'll have to build our plugins in ESM too~.
Not sure about building our modules in ESM too because they are available separately.

What does it add for our users ?

use Bootstrap as a module in their browser
```html
<script type="module">
  import { Toast } from 'bootstrap.esm.min.js'

  Array.from(document.querySelectorAll('.toast'))
    .forEach((toastNode) => new Toast(toastNode))
</script>
```

And bundlers will load the ESM version because we added `module` in our package.json

TODO:

- [x] See if we need the babel object spread plugin for our target browsers
- [x] Explain the known issue with the plugins that require Popper and link to https://developers.google.com/web/fundamentals/primers/modules#package-name-maps. Maybe make this an info/warning callout

Demo: https://codepen.io/Johann-S/pen/XGdLVK


XhmikosR:
Made a few tweaks and rebased this.

How about the docs? We need to list this and have a recipe for its use.

XhmikosR (approved):

Issue #26933, opened 2018-07-20
Custom file LABEL text overlapping neighbouring HTML elements on Firefox


maggyero:
On Firefox 63 (but not on Chrome 70, Edge 44 nor Safari 12), the text in the "custom-file-label" LABEL element seems to invisibly overlap the neighbouring button INPUT element and prevents from clicking on it:

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.2/css/bootstrap.min.css" integrity="sha384-Smlep5jCw/wG7hdkwQ/Z5nLIefveQRIY9nfy6xoR1uRYBtpZgI6339F5dgvm/e9B" crossorigin="anonymous">
  </head>
  <body>
    <div class="form-row" style="width: 400px">
      <div class="form-group col">
        <div class="custom-file">
          <input type="file" class="custom-file-input" id="browser">
          <label class="custom-file-label" for="browser">long_filename.txt</label>
        </div>
      </div>
      <div class="form-group col">
        <input type="button" value="Load" class="btn btn-primary">
      </div>
    </div>
  </body>
</html>
```

Joyrex:
@maggyero - does it still overlap if you remove the `style="width: 400px"` from the `div`?

maggyero:
@Joyrex If you choose a file with a very long name, yes. While other browsers are fine (it is a Firefox only issue).

smartygus:
I just encountered this problem in Firefox as well, and for me at least it's specifically when the file name is extra long as mentioned by @maggyero.

The elements in the custom file input have z-indexes and I tried unsetting these, which fixed the problem of not being able to click parts of the button to the right of my custom file input, but unfortunately broke the ability to drag and drop files onto the custom input (I'm guessing that's the reason for the z-indexes!!).

But then I figured because it only happened when the filename is long, it must be because the text of that filename is overflowing, and then "covering" part of the neighbouring element, in this case a button, and absorbing click events.

So I looked at the overflow property of the file input and voila! It was set to `visible`.

I've overridden the overflow property for the `<input>` tag in question and set it to `hidden` as a tentative solution. It seems to work for me, in that my neighbouring button is no longer obscured when a file with a long filename is selected in the input and normal behaviour (click to bring up file dialog, drag and drop) remains.

[...]

PR #28388, opened 2019-03-01
Make sure we can tweak the custom-file-border-radius-after

### LINKED ISSUE #28389: custom-file-border-radius-after is not customizable
When the border width for input fields is more than 1px, the radius of the `custom-file-border` label after is not correct.
This PR introduces a way to customize that radius with a new variable: `$custom-file-border-radius-after`

screenshot 2019-03-01 at 09 20 58
screenshot 2019-03-01 at 09 26 33

samdemaeyer:
Fixes #28389.
When the border width for input fields is more than 1px, the radius of the `custom-file-border` label after is not correct.
This PR introduces a way to customize that radius with a new variable: `$custom-file-border-radius-after`

**Before:**
screenshot 2019-03-01 at 09 20 58
screenshot 2019-03-01 at 09 26 33

**After:**
screenshot 2019-03-01 at 09 21 12
screenshot 2019-03-01 at 09 26 22


XhmikosR:
@twbs/css-review: should this be backported in v4?

alecpl:
Maybe it would be better to remove border-radius on ::after element and add overflow:hidden to the label? Another ticket is also asking for overflow:hidden.

alecpl:
The ticket is #26933

PR #28400, opened 2019-03-04
Disable ascending check if the previous unit was a percentual value

### LINKED ISSUE #28399: Mixing percentual and pixel values in $container-max-widths
Given this variable override value:

```scss
$container-max-widths: map-merge(
  (
    sm: 100%,
    md: 90%,
    lg: 90%,
    xl: 1440px,
  ),
  $container-max-widths
);
```

The `_assert-ascending` function fails because px and percent values are not comparable.

Maps which only contain percentages work, because of this fix https://github.com/twbs/bootstrap/pull/26689 – However this does not work if the map contains both types.

I believe this is a valid (and common) use case: There should be a maximum size for the container from a certain viewport width onward, before then the container should be flexible.

Conversely it could be possible that the container should turn into a percentual width from a certain viewport width onward.

mspae:
Fixes #28399 

This prevents the `comparable` check in the `_assert-ascending` if the previous value was a percentual value.

This removes compilation warnings when providing mixed percentual/pixel values to `$container-max-widths`.

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -8,7 +8,7 @@
   $prev-key: null;
   $prev-num: null;
   @each $key, $num in $map {
-    @if $prev-num == null or unit($num) == "%" {
+    @if $prev-num == null or unit($num) == "%" or unit($prev-num) == "%" {
       // Do nothing
     } @else if not comparable($prev-num, $num) {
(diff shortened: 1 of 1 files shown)


XhmikosR:
@MartijnCuppens: if this lands in master, should we backport it to v4?

MartijnCuppens:
Support for mixed usage of `px` and `%` feels more like a feature, so no need to backport this.

MartijnCuppens (approved):

PR #28401, opened 2019-03-04
Hyphenate server-side

vanillajonathan:


XhmikosR (changes requested):
There are more cases like the header in that file and also a comment in _forms.scss

vanillajonathan:
@XhmikosR I appended the PR with a commit for the comment in _forms.css.
I did not hyphenate the header, because it is two words only, not three.

XhmikosR:
Ah, right it's not an adjective there.

XhmikosR (approved):

PR #28404, opened 2019-03-04
Expand border utilities to include {sm,md-lg-xl}-{top,bottom,left,right}

### LINKED ISSUE #28403: Expand border utilities to include {sm,md-lg-xl}-{top,bottom,left,right}
As we already have utilities like: `text-sm-left` and `float-sm-left`, it would make sense to  alsohave `border-sm-left` and similar for the rest of sizes. (#25024)

I find it pretty often using borders that I don't want to show in smaller sizes when the grid may change to a stacked one. For example: a `border-left` makes sense in a column layout to separate columns, but not when they are stacked.

marcanuy:
Fix #28403

### DIFF
--- a/scss/utilities/_borders.scss
+++ b/scss/utilities/_borders.scss
@@ -73,3 +73,17 @@
 .rounded-0 {
   border-radius: 0 !important;
 }
+
+
+// Responsive alignment
+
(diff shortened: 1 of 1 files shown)


mdo:
Closing per https://github.com/twbs/bootstrap/issues/28403#issuecomment-469378466.

PR #28409, opened 2019-03-05
Added safe HTML tags abbr, kbd, q to sanitization whitelist

SilverFire:
Hello, team!

This pull request adds bulletproof-safe HTML tags such as `abbr`, `kbd`, and `q` to sanitization whitelist. If nobody minds, I will submit the same PR for v4.


Johann-S:
Hi @SilverFire,

Thanks for your PR but the current state is to stop adding things in our v3. We're working on our v5 and what you add can be done on people's side. 👍

SilverFire:
Well, you've added a new feature month ago (as a security fix, nevertheless), that introduced a BC break on people's side. I will submit the same PR for v4 and v5, but I think that this regression should be fixed in v3 as well.

SilverFire:
Btw, do you want me to submit a PR like this in v4 and v5?

Johann-S:
We added a way for our users to whitelist what they need so if you need to whitelist those elements you're free to do it using: https://getbootstrap.com/docs/3.4/javascript/#js-sanitizer

Yep we would accept those changes for v4 and v5

PR #28417, opened 2019-03-06
Scrollbar fix

### LINKED ISSUE #25686: Overflow issue on .sr-only
As identified by @Orhleil on boosted : https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/84

Reduced test case https://jsfiddle.net/etr3yjv6/10/

On chrome (v64 tested) there's an oveflow onto the main container. That one could be fixed by adding a `margin: -1px` onto the sr-only class

MartijnCuppens:
Fixes #25686.


ysds:
I don't know why, but it seems to work well.

ysds (approved):

mdo (approved):

Issue #28425, opened 2019-03-07
Remove tabindex ruleset


danielnixon:
The following isn't specific to any browser or OS.

Bootstrap contains the following SCSS in [_reboot.scss](https://github.com/twbs/bootstrap/blob/ff29c1224c20b8fcf2d1e7c28426470f1dc3e40d/scss/_reboot.scss#L57):
```
// Suppress the focus outline on elements that cannot be accessed via keyboard.
// This prevents an unwanted focus outline from appearing around elements that
// might still respond to pointer events.
//
// Credit: https://github.com/suitcss/base
[tabindex="-1"]:focus {
  outline: 0 !important;
}
```

This was actually [removed from suitcss/base five months ago](https://github.com/suitcss/base/commit/bf1d0bedecbc87fb3cf623e49066b4b54fb0a3ed):

> `tabindex=-1` CSS ruleset has been removed in order to support visual feedback on programmatically focused elements.

Please consider also removing it from Bootstrap.

The visual focus indicator is required for accessibility. See e.g.:
* https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html
* https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F78
* http://www.outlinenone.com/

If an author programmatically sets focus to a div (e.g. when a search form returns results you might move focus to the results) then they will need to undo Bootstrap's `outline: 0`.

patrickhlauke:
i agree in principle - though would caution that because of the naive way many browsers still apply the focus outline as a result of mouse interaction as well in these cases will have some odd repercussions. if we remove that rule from our reboot styles, we should reintroduce it specifically for some things such as modal dialogs - otherwise, even after a mouse activation, our dialogs will have a very prominent focus outline (since they're `tabindex="-1"` and programmatically focused).

in future, we can probably generalise this with `:focus-visible`, but for now we need to weigh up both the aesthetics and visual presentation for non-keyboard users and clarity for keyboard users.

patrickhlauke:
and for reference, once `:focus-visible` is supported, the rule could be modified to

```
[tabindex="-1"]:focus:not(:focus-visible) { outline: 0 !important; }
```

(can be verified in Chrome > 67 with experimental web features flag enabled)

patrickhlauke:
[...]

PR #28437, opened 2019-03-09
Remove indiscriminate outline suppression for tabindex="-1" elements

patrickhlauke:
instead, only apply outline suppression if the browser wouldn't normally apply the focus outline, using the (currently experimental) `:focus-visible` pseudo-class

Closes https://github.com/twbs/bootstrap/issues/28425


MartijnCuppens:
This will add the focus ring back everywhere since this property is not yet supported (https://caniuse.com/#search=focus-visible). Don't know if that's an issue?

patrickhlauke:
> This will add the focus ring back everywhere

everywhere there's a `tabindex="-1"` and the element is then focused programmatically - either because, say, it's an in-page anchor, or because it was invoked via `focus()` JS method. unless i missed anything, only our modal dialogs (which have an explicit `outline: 0` already) fall into the latter category.

patrickhlauke:
but yes, the `:focus-visible` part is speculative/future-compatible (as per the comment in the scss)

MartijnCuppens (approved):
Aight, fine by me

PR #28445, opened 2019-03-10
Utilities/helpers split & utilities API

### LINKED ISSUE #28367: Split utilities from property utilties and get more control over generation of property utilities
## Split utility groups
In `v4` we have 2 different kind of utilities that are a bit mixed in the documentation and in the code.

The first group is what I call the normal utilities. These utilities have more than one property and don't have the `!important` suffix. These utilities are:
- [Clearfix](https://getbootstrap.com/docs/4.3/utilities/clearfix/)
- [Embeds](https://getbootstrap.com/docs/4.3/utilities/embed/)
- [Text hide](https://getbootstrap.com/docs/4.3/utilities/image-replacement/)
- [Fixed top](https://getbootstrap.com/docs/4.3/utilities/position/#fixed-top), [fixed bottom](https://getbootstrap.com/docs/4.3/utilities/position/#fixed-bottom) & [Sticky top](https://getbootstrap.com/docs/4.3/utilities/position/#sticky-top)
- [Screen reader utilities](https://getbootstrap.com/docs/4.3/utilities/screen-readers/)
- [Stretched link](https://getbootstrap.com/docs/4.3/utilities/stretched-link/)

The second group is what I call the property utilities. These utilities typically have one property (except for `.mx-*`, `.my-*`, `.px-*` and `.py-*` utilities) and always have the `!important` suffix.

I think we should split these utility groups. Not sure how we should call them, but for now I'll call them the utilities and property utilities. 

## More control over property utilities
We got a lot of requests and PRs about the generation of more responsive properties (for example responsive border utilities https://github.com/twbs/bootstrap/issues/28191) but we can't just add utility classes for all the properties because this will increase the size of the css too much. 

On the other hand there's a need for more control over the generated code (which is also mentioned in https://github.com/twbs/bootstrap/issues/25941).

On yet another hand (let's just assume there's a third hand somewhere) I sometimes have the need to extend the utilities for more properties like `opacity` or other properties Bootstrap doesn't cover with utilities.

To tackle this problem I've started developing something to gain more control over property utilities, with some kind of API where you can choose whether you also want to have responsive utilities, but before I'm going to continue with it, I'm going to check for some other opinions here. It 'll probably also need some solid documentation to explain it all.

I'm working under `v4-dev-mc-poc-property-utilities`, the branch will need to be renamed, because this will not land in `v4`, but maybe something for `v5` if I can get to something ready in time. You can see the diff here: https://github.com/twbs/bootstrap/compare/v4-dev-mc-poc-property-utilities?expand=1#diff-1500afbbcf5fcafdd95fddb9b709de83 `scss/_property-utilities.scss` and `scss/utilities/_property-utilities.scss` will probably be the most interesting files. 

What do you think?

@ysds @gijsroge @twbs/css-review

MartijnCuppens:
Fixes #28367

Split up helpers/utilities and utilities API for extending or getting more control over the generated utilities. 

Features
- Split between helpers & utilities
- Utility API which allows changing & extending utility classes
- Documentation about the API
- Documentation split between helpers & utilities.
- New `map-get-multiple()` function to get multiple keys from a map

Notes
- I left some utilities because their naming required to create a utility group per class, which beats the purpose of the API. Not sure what to do with them, but for me, it fine leaving them in the utility folder for now.
- I had to increase the bundle sizes because the vertical & horizontal spacing utilities are generated differently

New documentation about the utility API:
https://deploy-preview-28445--twbs-bootstrap.netlify.com/docs/4.3/utilities/api/


scss/utilities/_utilities.scss
```diff
@@ -0,0 +1,79 @@
...
+  @include media-breakpoint-up($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+    // Loop over each utility property
+    @each $key, $utility in $utilities {
+
+      // Only render if responsive media queries are enabled or if it's the base media query
+      @if type_of($utility) == "map" and (map_get($utility, responsive) or $infix == "") {
+        $values: map_get($utility, values);
+
+        // If the values are a list or string, convert it into a map
```
mdo: Could we remove this check if we only accepted maps for the values?

scss/utilities/_utilities.scss
```diff
@@ -0,0 +1,79 @@
...
+          $values: zip($values, $values);
+        } @else if type_of(nth($values, 1)) != "list" {
+          $values: zip($values, $values);
+        }
+
+        @each $value in $values {
+          $properties: map_get($utility, property);
+
+
+          @if type_of($properties) == "string" {
```
mdo: What's this string check for?

scss/bootstrap-grid.scss
```diff
@@ -24,6 +24,43 @@ html {
...
 
 @import "grid";
-@import "utilities/display";
-@import "utilities/flex";
-@import "utilities/spacing";
+
+@import "utilities";
+// Only use the utilities we need
+// stylelint-disable-next-line scss/dollar-variable-default
+$utilities: (
```
mdo: I'd like to keep pushing on simplifying things—how do we get rid of this entire Sass map? I'm hoping we can essentially have the new `_utilities.scss` file and that's it.

mdo:
For viewport utilities, I'm thinking `.vw-100` and `.vh-100` for their classes as the names are implicit to their unit of measurement and their intended use (`100vw` and _viewport width 100%_).

scss/utilities/_utilities.scss
MartijnCuppens: Some utility groups have a single property: (string)
https://github.com/twbs/bootstrap/blob/b1a7b1a335e813bc030b4fa4918f9b826445f9e2/scss/_utilities.scss#L323
others have multiple properties: (list)
https://github.com/twbs/bootstrap/blob/b1a7b1a335e813bc030b4fa4918f9b826445f9e2/scss/_utilities.scss#L433

scss/utilities/_utilities.scss
MartijnCuppens: I did the type check because of:
https://github.com/twbs/bootstrap/blob/b1a7b1a335e813bc030b4fa4918f9b826445f9e2/scss/_utilities.scss#L50-L53

This was the simplest way to disable it

scss/bootstrap-grid.scss
MartijnCuppens: Not sure how to approach that, I could also disable the utilities we don't need, but it will still be a sass map

mdo:
So here's a baseline version of what I'm thinking about—one giant Sass map for all `$utilities` and nested `@each` loops for generating them. No additional map of `map-get`s. I've punted on all the more complex stuff though (any spacing utils, anything with a `map-merge`, etc) though for demo purposes. Will need help with that if we go this route.

https://www.sassmeister.com/gist/efa2593de91ddcc60808c44fe1ea9301

Thoughts?

MartijnCuppens:
I guess you're talking about the `map-get`s in the `scss/bootstrap-grid.scss`? I did that to prevent the repetition of properties here, but maybe I should tackle this in a function where you can send the keys of the utilities you need to?

Or am I missing something here?

mdo:
> I guess you're talking about the `map-get`s in the `scss/bootstrap-grid.scss`?

Oh! Damn, yes I am! I was completely off base, apologies!

XhmikosR:
@MartijnCuppens: something isn't right with this branch. You are removing files you are not supposed to touch here.

scss/utilities/_utilities.scss
MartijnCuppens: I clarified this now: https://github.com/twbs/bootstrap/blob/a41ab9db631e187bc01b4654bbf620cd60eeb286/scss/utilities/_utilities.scss#L45

MartijnCuppens:
@XhmikosR, fixed it. @mdo, I simplified the `bootstrap-grid.scss` with a new `map-get-multiple` function which prevents repetition of the `map-get` function.

This is still a work in progress, I'm personally going to give this PR more priority to avoid the situations like th Hugo branch where we needed to rebase it every so often.

site/content/docs/4.3/helpers/clearfix.md
```diff
@@ -2,7 +2,8 @@
 layout: docs
 title: Clearfix
 description: Quickly and easily clear floated content within a container by adding a clearfix utility.
-group: utilities
+group: helpers
+aliases: "/docs/4.3/helpers/"
```
XhmikosR: Do we really need the redirects since this will be a major version bump?

MartijnCuppens: I wasn't sure about this either, but since every css group redirects to it's first page, this would make sense.

XhmikosR: Ah, yeah, if it's for the first page, you should add it.

mdo:
@MartijnCuppens Thoughts on ditching the `map-merge` aspect of this? Had some thoughts on this in another thread where if memory serves, we can perhaps ditch it? https://github.com/twbs/bootstrap/issues/27927#issuecomment-450583968

MartijnCuppens:
I've opened #28506 to track this, that's a decision we can make independent from this, but I think it's a good idea. This will make it possible to ditsh this part here and simplify the map-gets you mentioned before:
https://github.com/twbs/bootstrap/blob/a41ab9db631e187bc01b4654bbf620cd60eeb286/scss/_utilities.scss#L50-L53

MartijnCuppens:
Ping @twbs/css-review for review

MartijnCuppens:
FYI: the diff between classes: https://www.diffchecker.com/noUio3fe. Only some responsive `.text-justify` classes were added.

mdo (approved):
Let's do this!

PR #28450, opened 2019-03-11
v5: Forms update

mdo:
WIP on some relatively major form changes:

- [x] Move forms documentation to it's own top level docs nav group
  - [x] Move `/components/forms` to `/forms`
  - [x] Move `/components/input-group` to `/forms/input-group`
  - [x] Break apart new `/forms` into subpages
- [x] Combine native and custom checks, radios, and switches (going all custom!) into `.form-check`, etc
- [x] Combine native and custom selects into `.form-select`
- [x] Combine custom file and custom range styes into the browser native `.form-control-file` and `.form-control-range`
- [x] Drop `.form-group` for a margin utility?
- [ ] Revisit inline forms to redo their flex styles to fix some bugs
- [ ] Any changes to input groups for rounding and form validation?

Probably want to wait to get Hugo stuff merged before continuing on. I imagine when I start to move the docs around there will be hella conflicts.

Demo: https://deploy-preview-28450--twbs-bootstrap.netlify.com/docs/4.3/forms/overview/


site/content/docs/4.3/components/forms.md
```diff
@@ -163,6 +163,93 @@ Set horizontally scrollable range inputs using `.form-control-range`.
 
 ## Checkboxes and radios
 
+<script>
```
XhmikosR: Note that this should go in application.js and without the `DOMContentLoaded`.

mdo:
Just marked this ready for review solely could I see the diff and conflicts here in the PR. Will try to resolve those soon.

mdo:
I'm going to split this PR apart soon given the scope of changes included is quite massive now. I'm on a roll though with a ton of things here, so want this PR left for me to tinker :).

mdo:
@patrickhlauke Damn, okay. One idea that comes to mind then is exploring `data` attributes for validation message content and style. Something like this maybe?

```html
<div class="form-check">
  <input type="checkbox" id="formCheckEx" class="form-check-input" 
         data-validation="true" 
         data-validation-style="tooltip" 
         data-validation-body="This box must be checked.">
  <label for="formCheckEx">Example</label>
</div>
```

Still want to dig in more to understand the problem—for example, which browser/AT combos?

patrickhlauke:
> Still want to dig in more to understand the problem—for example, which browser/AT combos?

Using https://codepen.io/patrickhlauke/pen/qzNqpO as a reduced non-BS test case:

- NVDA + all browsers (IE, Edge, Firefox, Chrome): both for `aria-labelledby` and `aria-describedby` it makes no difference if using `display:none` ... they're seen and announced
- JAWS + all browsers: same, with one exception - in JAWS+Edge, the `aria-describedby` ignored the exta content when using `display:none`
- iOS VO+Safari: seems to be bugged/ignore `aria-labelledby` in all circumstances (even when NOT hidden); does announce the `aria-describedby` content (after a 2 second delay) even when `display:none`d

In short...pretty much safe to say "in practically all cases".

Using `data-` and making a custom plugin seems to be the best path forward. Happy to collaborate, if you show me the ropes on how to go about carving out a whole new plugin, as that's beyond my knowledge just now...

Johann-S:
Maybe I can help you by pushing a new plugin structure

site/content/docs/4.3/forms/validation.md
```diff
@@ -0,0 +1,404 @@
...
+
+<script>
+// Example starter JavaScript for disabling form submissions if there are invalid fields
+(function() {
+  'use strict';
+  window.addEventListener('load', function() {
+    // Fetch all the forms we want to apply custom Bootstrap validation styles to
+    var forms = document.getElementsByClassName('needs-validation');
+    // Loop over them and prevent submission
+    var validation = Array.prototype.filter.call(forms, function(form) {
```
Johann-S: pretty weird to use `filter` because you do not filter here, you should use `forEach`

mdo:
> Maybe I can help you by pushing a new plugin structure

No need to jump on it just yet, but maybe as a separate PR if we definitely need it :).

site/content/docs/4.3/forms/validation.md
mdo: This is in the current docs, so going to leave as-is and if y'all wanna customize, I'm down to see improvements in a dedicated PR :).

patrickhlauke: https://github.com/twbs/bootstrap/pull/28995 ?

MartijnCuppens:
- Why was `padding-right` added here?
https://github.com/twbs/bootstrap/blob/7a2c2b754399c74b2c54496bb90889a8b77942c5/scss/_reboot.scss#L160-L164
- The layout is broken in IE
- Custom checkbox theming is broken in IE
- This change to the tables don't make sense: https://github.com/twbs/bootstrap/blob/7a2c2b754399c74b2c54496bb90889a8b77942c5/scss/_tables.scss#L18-L20
- Why was the table cell padding changed?
- The checkboxes look a bit too large imo. A width and height of 1em would be better I think.
- Same for radios
- Some markdown tables were replaced by html tables, why?
- The icons page is quite empty (https://twbs-bootstrap.netlify.com/docs/4.3/components/icons/), what was the purpose of this page?
- Is there any overview of all the changes in this PR or is there a todo list?

Justsnoopy30:
@mdo Why do the new form components not have transitions/motion? Is there a way to re-enable them?

MartijnCuppens:
@Justsnoopy30: https://deploy-preview-30502--twbs-bootstrap.netlify.com/docs/4.3/getting-started/accessibility/#reduced-motion

PR #28453, opened 2019-03-11
Rewrote displays type variables to use them as a map

### LINKED ISSUE #26515: More display-* headings - .display-5

afanjul:
Rewrited displays type variables to use them as a map so it would be easier to extend and modify.

Fixes #26515.


XhmikosR (dismissed):
please fix the test failures.

afanjul:
> please fix the test failures.

Could you check now? I tried to just push the fixes to my own branch but the pull request would never get updated (I don't know if its a github bug or what)... so I pushed to my own fork master... according to github message "_Add more commits by pushing to the master branch on afanjul/bootstrap_." 

It's a bit complicated to just update a pull request... :(

XhmikosR:
GitHub could have a slowness issue right now, but usually you just push to your branch.

MartijnCuppens:
Hi @afanjul,

I'm a bit afraid it's not clear what `300` means here for people who don't have a clue the font weight is set:
https://github.com/twbs/bootstrap/blob/3182990fee5aa32b3e4146685e257821c37d9666/scss/_variables.scss#L306-L309

I did a similar PR once, but closed it since I was a bit worried we're mixing up too many different "styles" of configuration: https://github.com/twbs/bootstrap/pull/26533.

afanjul:
> Hi @afanjul,
> 
> I'm a bit afraid it's not clear what `300` means here for people who don't have a clue the font weight is set:
> [bootstrap/scss/_variables.scss](https://github.com/twbs/bootstrap/blob/3182990fee5aa32b3e4146685e257821c37d9666/scss/_variables.scss#L306-L309)
> 
> Lines 306 to 309 in [3182990](/twbs/bootstrap/commit/3182990fee5aa32b3e4146685e257821c37d9666)
> 
>    1:(6rem, 300), 
>    2:(5.5rem, 300), 
>    3:(4.5rem, 300), 
>    4:(3.5rem, 300) 
> I did a similar PR once, but closed it since I was a bit worried we're mixing up too many different "styles" of configuration: #26533.

Another alternative (a worse one from my point of view) is simply split it in two maps:

**In _variables.scss**

```
$display-sizes: (
  1:6rem,
  2:5.5rem,
  3:4.5rem,
  4:3.5rem
);
$display-weights: (
  1:300,
  2:300,
  3:300,
  4:300
);
```

The only "issue" is that you have to use the $index in for loop (and maintain a coherence in the maps):

```
@for $i from 1 through length($display-sizes) {
  $size: nth($display-size, $i);
  $weight: nth($display-weight, $i);
  .display-#{$index} {
    @include font-size($size);
    font-weight: $weight;
    line-height: $display-line-height;
  }
}
```

But the worst method, for me, is the actual one, with separate values/variables for size and weights. It's ugly and not extendible.

afanjul:
Anyway, if you are using Sass as a "web developer", I'm assuming that you can read and interpret docs, comments, functions, etc.  Either way, you will be using compiling bootstrap.css

afanjul:
For example, what if I want to add another "display-5" in the actual bootstrap version? 
I can add a new $spacers or $sizes element, but you don't let me do the same with "$displays"? A bit odd..

scss/_variables.scss
```diff
@@ -301,15 +301,13 @@ $headings-font-weight:        500 !default;
...
-$display3-size:               4.5rem !default;
-$display4-size:               3.5rem !default;
-
-$display1-weight:             300 !default;
-$display2-weight:             300 !default;
-$display3-weight:             300 !default;
-$display4-weight:             300 !default;
+//display types as font-size and font-weight
+$displays: (
+  1:(6rem, 300),
```
mdo: Let's use nested maps with explicit key-value pairs instead of this setup that requires `nth` value navigation.

mdo (changes requested):

mdo:
> I did a similar PR once, but closed it since I was a bit worried we're mixing up too many different "styles" of configuration: #26533.

Reading that thread again @MartijnCuppens, it looks like I punted on it for redoing bigger things in v5. I can't recall exactly what though, but perhaps it was related to your RFS work?

scss/_type.scss
```diff
@@ -26,10 +26,10 @@ h6, .h6 { @include font-size($h6-font-size); }
 }
 
 // Type display classes
-@each $index, $values in $displays {
+@each $index, $display-type in $display-types {
   .display-#{$index} {
-    @include font-size(#{nth($values, 1)});
-    font-weight: #{nth($values, 2)};
+    @include font-size(map-get($display-type, font-size));
+    font-weight: map-get($display-type, font-weight);
```
afanjul: I'm getting an "Expected numeric font-weight notation" css-lint error in the map-get function, but all seems right and the compiling output is also right... so I don't know...

MartijnCuppens: You can disable this with `// stylelint-disable-next-line font-weight-notation`. Stylelint doesn't know what the `map-get` result will be, so you need to explicitly disable the rule on that line.

afanjul:
Now I think that the source code is complete and prettier... Using a map for $display-types, with key value explicit names. So now is pretty easy to modify or extend by the people... Just be aware of the new variable called $display-types for docs, and so...

scss/_variables.scss
afanjul: Now using nested maps with explicit key-value pairs

XhmikosR:
This needs a rebase + squash

mdo:
Replaced by #30568.

Issue #28444, opened 2019-03-10
Remove focus & hover states from `bg-variant` mixin


MartijnCuppens:
About this code:
https://github.com/twbs/bootstrap/blob/8896827bd072a120705a0576a5bcde69c8ce2cda/scss/mixins/_background-variant.scss#L9-L14

I think we should remove it. IMO focus and hover background states don't belong in these utilities and we are also not defining an active state here. We have the buttons with backgrounds, if this doesn't cover the needs I think the developer will need to write custom code, because it's quite specific.

I guess we can just remove the mixin, just use the `background-color` property in `v5` and use the `deprecate` mixin to deprecate this in `v4`.

Thoughts?

mdo:
Big fan of this. Nix the mixin and the hover/focus states. We'll inevitably get the question of how to set hover styles via utilities, but I don't think that's a path we should go down. Maybe provide some guidance in our docs though for how to customize that kind of stuff?

MartijnCuppens:
> Maybe provide some guidance in our docs though for how to customize that kind of stuff?

If we ditch the mixin, we can use the utility API https://github.com/twbs/bootstrap/issues/28367 (or whatever it will be called) and  clarify this in its documentation section.

> We'll inevitably get the question of how to set hover styles via utilities, but I don't think that's a path we should go down

I agree, I experimented with hover utilities before, but you'll end up with elements tons of utility classes. It's just better to create a custom component if an element needs hover states to get a better overview of all the properties.

PR #28457, opened 2019-03-11
Remove hover states on background utilities

MartijnCuppens:
Ref https://github.com/twbs/bootstrap/issues/28444. (going to trigger the fixes thing in the deprecation PR for `v4`)


MartijnCuppens:
@XhmikosR, I was thinking about doing another PR for `v4-dev` to add the deprecation warning, but this will be a commit that will not be present in the `v5` branch, is that a problem?

XhmikosR:
No, but since we haven't merged `v4-dev-new`, you can push it there. BTW, are you sure we won't be able to cherry pick it?

XhmikosR:
BTW, we still need to add here backport or v4 labels since we need to track the changes somehow.

MartijnCuppens:
In `v5` we remove the mixin, in `v4` we throw a deprecation warning, it's different.

XhmikosR:
Ok then just reference the issue somehow in the v4 project so that we don't
miss it in the release notes.

mdo (approved):

PR #28461, opened 2019-03-12
v5: Drop remaining deprecated mixins

mdo:
Didn't realize we had this many deprecated mixins to drop. See also #28459 and #28460. The `.form-control-focus` mixin is being removed in #28450.


XhmikosR:
It seems the mixin is currently in use in some core parts.

MartijnCuppens:
We should also update the docs. I think we better drop these mixins after we've merged the Hugo branch to avoid too many conflicts.

XhmikosR:
There shouldn't be too many doc changes here, only the text-hide page removal page and the nav update which I can rebase.

But yeah, we need to move forward because I have rebased it like 150 times already.

mdo:
Agreed, I'm waiting on any meaningful docs changes until after Hugo merges. I have plenty coming per our v5 plans issue in the team repo.

mdo:
Pushed updated docs, so this is ready for review.

MartijnCuppens (approved):

PR #28463, opened 2019-03-12
v5: Move from JS input button group toggling to all CSS

mdo:
Takes #25185 and starts work on the docs and JS to drop code that was previously handling active states.

/cc @ysds 

- [x] make sure the CSS changes are OK
- [x] **make sure the JS changes are OK, tests pass, visual tests are also OK**
- [ ] Check if we need to revisit bundlesize limits
- [x] Finally, **make sure all functionality we had before is still there, especially accessibility-wise** /CC @patrickhlauke


mdo:
Superseded by #28463.

ysds:
The CSS LGTM👍  IMO, it is better to move the documents to the button group page.

XhmikosR:
Do not merge yet, @mdo, until 1) we make sure we don't lose any functionality, and 2) JS and its tests/including visual are fine.

site/docs/4.3/components/buttons.md
```diff
@@ -125,32 +125,36 @@ Add `data-toggle="button"` to toggle a button's `active` state. If you're pre-to
...
-  </label>
-  <label class="btn btn-secondary">
-    <input type="radio" name="options" id="option2" autocomplete="off"> Radio
-  </label>
-  <label class="btn btn-secondary">
-    <input type="radio" name="options" id="option3" autocomplete="off"> Radio
-  </label>
+<div class="btn-group">
+  <div class="btn-group">
+    <input type="checkbox" class="btn-group-input" name="btnGroupRadio" id="btnGroupRadio1" checked>
```
ysds: This and the remain two lines should be `type="radio"`.

scss/_button-group.scss
```diff
@@ -148,16 +150,8 @@
...
-    input[type="checkbox"] {
-      position: absolute;
-      clip: rect(0, 0, 0, 0);
-      pointer-events: none;
-    }
-  }
+.btn-group-input {
+  position: absolute;
+  clip: rect(0, 0, 0, 0);
+  pointer-events: none;
```
ysds: Need to remove label's bottom margin.
```suggestion
  pointer-events: none;

  ~ .btn {
    margin-bottom: 0; // Override default `<label>` value
  }
}
```

js/src/button.js
```diff
@@ -65,56 +62,10 @@ class Button {
...
-      this._element.setAttribute('aria-pressed',
-        !this._element.classList.contains(ClassName.ACTIVE))
-    }
+    this._element.setAttribute('aria-pressed',
+      !this._element.classList.contains(ClassName.ACTIVE))
 
-    if (triggerChangeEvent) {
-      this._element.classList.toggle(ClassName.ACTIVE)
-    }
+    this._element.classList.toggle(ClassName.ACTIVE)
```
XhmikosR: So, I guess we don't need the active class any longer, right?

I think we'll need to find another way to add the `aria-pressed` attribute.

mdo: Yeah, we might be able to nix all the JS—can anyone confirm that? Only reason I could see us leaving some JS for a non-input implementation is quickly toggling active on/off, but I feel like most folks could do that themselves with very little code and their own syntax.

XhmikosR:
And who's going to handle aria pressed?

scss/_button-group.scss
```diff
@@ -148,16 +150,13 @@
...
-    input[type="checkbox"] {
-      position: absolute;
-      clip: rect(0, 0, 0, 0);
-      pointer-events: none;
-    }
+  ~ .btn {
+    margin-bottom: 0; // Override default `<label>` value
   }
 }
+}
```
ysds: Sorry, my previous suggestion added a duplicate '}' 🙏 

```suggestion
```

ysds:
It will be difficult to solve the #25122 and #27600 issues as long as we use `<input type="checkbox">` and `<input type="radio">` to handle the `aria-pressed`.

According to @scottaohara's [explanation](https://scottaohara.github.io/a11y_styled_form_controls/src/radio-button--pill/), there is a radio button group implementation pattern without `aria-pressed`.

Also, his comment at the end of the article suggests a grouping of `<button>`s. It is probably something like the [WAI-ARIA practice](https://www.w3.org/TR/wai-aria-practices-1.1/examples/toolbar/toolbar.html).

> Alternatively, a grouping of button elements that are aware of each other's state, and utilize aria-pressed="true/false" to indicate their own state, may be more appropriate and understandable to users.

So in terms of Pure CSS and accessibility, we'll need decide to which one:

- A. Implement by Pure CSS (`<input>`) without `aria-pressed`
- B. Implement with `<button>` using JavaScript to control the aria attributes

scottaohara:
> Also, his comment at the end of the article suggests a grouping of `<button>`s. It is probably something like the [WAI-ARIA practice](https://www.w3.org/TR/wai-aria-practices-1.1/examples/toolbar/toolbar.html).

that could be one method but not necessarily the only method, in that a grouping could be just a `group` of buttons, rather than a `toolbar`. More importantly though, each would provide different user expectations to the manner in which they were keyboard accessible, as would checkboxes and radios.

As I'm not deeply familiar with what it is you're trying to solve for here, and honestly i'm a bit confused reading through just this thread with the number of different elements being referenced, I'd definitely defer to @patrickhlauke's guidance.

patrickhlauke:
coming in late on this...so: skimming over the changes, it looks like we're going for clean CSS version here? just layering over texisting `<input>` elements? in that case, if those inputs are there and focusable (and not suppressed), there won't be need for any JS or aria or anything, unless I'm missing something.

XhmikosR:
Yeah, the idea is to use only CSS for buttons. If we don't need to handle aria, I can update the branch dropping more stuff.

patrickhlauke:
ah, hang on...i think the confusion here is because there's two separate things going on...there's the "checkboxes and radio buttons that look like a toolbar/button strip" https://getbootstrap.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons (which is what wouldn't need any JS nor `aria-pressed` or anything, and can be solved purely with CSS), and then there's the "set of buttons that acts as mutually exclusive toggles" https://getbootstrap.com/docs/4.3/components/button-group/ (which DO need `aria-pressed` to denote their current toggle state)

I'll be honest, i find these two things related, but confusing. Most confusing that the checkbox/radio ones are under the "Buttons" documentation - should they perhaps be under their own component topic for checkbox/radio button group?

XhmikosR:
Hmm then I'm also lost. Not sure how to proceed. Maybe we should rename `button` to something else and only keep what we need in JS?

patrickhlauke:
incidentally, am i missing something, or are the groups in https://getbootstrap.com/docs/4.3/components/button-group/ not actually doing anything fancy JS wise at the moment? are they supposed to have aria-pressed and act like sets of radio buttons or not? the "button plugin" mention on there simply points back to the buttons page...so seems we're making some kind of circular reference here.

worth taking a step back and deciding overall what we want. certainly the easiest part is:

- we can style checkboxes and radio buttons with CSS to appear button-like; behind the scenes, it's still just an `<input>` and a `<label>`, and uses CSS - no need for JS/ARIA. this is probably a special case of form element styling, not necessarily/logically something that should fall under "button" component documentation?
- for `<button>` elements to act as on/off type toggles, they need JS and `aria-pressed` handling

Now, button groups are like an extended version of the single button that acts as a toggle. A group can be mutually exclusive (acting like a set of radio buttons) or not (like a toolbar in a wysiwyg tool, where each on/off toggle is independent - set bold, or italic, or underline, independently. Those, again, need JS. and for the mutually exclusive capability, they need JS beyond just the pure toggle JS functionality, as it needs to not only toggle `aria-pressed` on what the user selected, but also unset it on any others in the same group.

Conceptually then, the single toggle button plugin thing, and button groups, probably should go into their own component documentation page together, as they're thematically related (groups build on the single toggle). Whereas the implementation using radio/checkbox is just a differently styled set of form controls?

mdo:
Answering some questions here...

- Buttons in button groups are not intended to include JS functionality. Button groups are a component for controlling layout of buttons. None of these buttons _need_ JS from us—anyone can write their own JS to toggle the `.active` class.
- My proposal earlier is to drop our button JS plugin and use only `<input>`s and `<label class="btn">` combos.
- Given that, and per @ysds's feedback, we can then move the input/label examples to the button group docs page.

If all that's true and agreeable, it sounds like we don't need any JS here or any `aria` attributes. If that doesn't sound right, let's reframe the conversation around "how do we remove all this JS?" and answer any questions that come from that.

patrickhlauke:
@mdo thanks for clarifying. so yes, basically removing the "Add on optional JavaScript radio and checkbox style behavior with our buttons plugin." sentence, and just making button groups a visual style? And just having actual radio/checkbox inputs to act as such, without the button plugin? In that case yes, no need anymore for the JS nor ARIA. and it would remove the confusion.

Would be good to provide some clearer examples for button groups, as currently some of them make it still look like they're supposed to be some mutually exclusive radios. But that can be revisitied in a separate PR.

patrickhlauke:
related, if we want to at least fix current keyboard issue in v4 https://github.com/twbs/bootstrap/pull/28565

patrickhlauke:
one thing i did notice: when using real `<button>` (for toggles and elsewhere), IE11 does the "content of the button moves down and right a few pixels when pressed" dance. surprised this wasn't normalised away...was it not before?

patrickhlauke:
https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons work and look as they should
https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/buttons/#button-plugin works great too
and both automated and visual tests are all ok
so, at this stage - barring any fundamental changes or concerns that maybe i've gone too far in slashing things apart - this seems functionally good from my point of view...

patrickhlauke:
I'll see if I can fix up the old-school button in IE11 issue with the text dropping down when pressed/clicked...after that, this may be ready to properly review

patrickhlauke:
> I'll see if I can fix up the old-school button in IE11 issue with the text dropping down when pressed/clicked...after that, this may be ready to properly review

doesn't seem to be any clean fix (some stuff i've seen involves wrapping all the content of the button in a `<span>` and applying some CSS translate stuff, or setting `position:relative` ... but in any case, stuff that would need extra markup that I wouldn't want to require from authors just so that the IE11 fix works). as we've not normalized/reset this in the past, i think then that that can stay as is, and this PR is ready for review. /cc @twbs/css-review @twbs/js-review

patrickhlauke:
just skimming over the previous early conversation here, there was the idea of moving the checkbox/radio stuff to button group rather than leaving them in the button docs page. i see the point to an extent, but i think more logically the now amended checkbox/radio stuff goes well in button docs page, as it shows how you can make checkboxes/radio buttons that look like buttons. the fact that they're in a button group is incidental

patrickhlauke:
actually, few more tiny tweaks that make this even more explicit...

patrickhlauke:
"tiny" tweaks ended up simplifying the required structure for the new button-like radio/checkbox buttons (now just needs the input and label adjacent, no wrapper around it), more clearly showing the difference between having a series of these, and having them in a seamless-looking button group. docs now show this nicely/clearly as well

site/content/docs/4.3/components/progress.md
```diff
@@ -127,7 +127,7 @@ The striped gradient can also be animated. Add `.progress-bar-animated` to `.pro
   <div class="progress">
     <div class="progress-bar progress-bar-striped" role="progressbar" aria-valuenow="75" aria-valuemin="0" aria-valuemax="100" style="width: 75%"></div>
   </div>
-  <button type="button" class="btn btn-secondary mt-3" data-toggle="button" id="btnToggleAnimatedProgress" aria-pressed="false" autocomplete="off">
```
XhmikosR: Maybe this change shouldn't be in this PR?

XhmikosR: Or am I wrong... I thought you had another PR for removing autocomplete.

patrickhlauke: think it makes sense to keep it here as it's related to toggle buttons and how they're then presented/used throughout the docs, rather than leave it in and remove it later (the other branch is still in limbo as it's not quite clear if Firefox still has that bug that prompted it in the first place or not)

ysds:
> now just needs the input and label adjacent, no wrapper around it

Oh, I intentionally wrapped those in `.btn-group` for margin control. https://codepen.io/anon/pen/QRXYgm Please see the left edge (-1px margin).

And it also available of labeling without `id/for`.

patrickhlauke:
> Oh, I intentionally wrapped those in `.btn-group` for margin control. https://codepen.io/anon/pen/QRXYgm Please see the left edge (-1px margin).

Can that be fixed in the current approach?

> And it also available of labeling without `id/for`.

Example?

ysds:
e.g.
```html
<label class="btn-group">
  <input type="checkbox" class="btn-group-input">
  <span class="btn btn-primary">Single toggle</span>
</label>
```

> Can that be fixed in the current approach?

The current negative margins and border radius control are very complex:

image

Current negative margins and border radius control are very complex.

It is not easy to check that the current approach has no problem in various markup by various users. The `.btn-group` wrappers will be a safe approach for that.

patrickhlauke:
then i suggest the classnames get changed. it's not a button group, it's a custom button wrapper. i'll have a think

patrickhlauke:
> Current negative margins and border radius control are very complex.

would need some testing, but I'm sure something like

```
.btn-group .btn-input + .btn:first-of-type { margin-left: $btn-border-width }
```

of similar could do it.

Perhaps it would be good offering this way (using just the input adjacent label) as an option, and also providing a new class `.btn-input-wrapper` ? I just feel `.btn-group` is misleading, as you're not wrapping it around a group, but around a single checkbox/radio `<input>` and its related label text element (either a `<label>` or a `<span>` or whatever, like in your example. Thoughts?

patrickhlauke:
@ysds @mdo see what you think of this approach...a bit more involved than i originally anticipated, but essentially adds a `.btn-input-wrapper` (instead of piggy-backing inappropriately on `.btn-group`). expanded the docs to show various examples, both with explicit (`for`/`id`) and implicit `<label>` elements. also tweaks the toggle example slightly to show both pre-toggled and regular toggle.

see https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/buttons/#checkboxes-and-radio-buttons / https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/buttons/#toggle-states / https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/button-group/#checkbox-and-radio-button-groups and the added radio button example for vertical variant https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/button-group/#vertical-variation

patrickhlauke:
as mentioned, would have loved to use the outline variants for these examples (particularly for the button group examples), but it appears currently we lack visual differentiation between hover and active on those (by design it seems), which makes them horribly confusing. will look at addressing that in a separate PR, after which we can revisit the examples here, but that shouldn't hold up review of this one.

patrickhlauke:
late night realisation ... these radio/checkbox style buttons lack styles for disabled inputs. putting this back in the oven for a bit longer, should find some time tomorrow to add those as well.

patrickhlauke:
ok...the checkbox/radio button buttons now have proper handling for disabled input inside, and i fixed up a few stragglers like the box shadow not being correct. lastly, added extensive examples to both the buttons and button group docs page to show variations/uses. @mdo @ysds thoughts?

mdo:
Trying to figure out the HTML and CSS here... I'd like to avoid adding tons of new selectors to account for this. Might be impossible, but I'm not giving up just yet. 😬 Starting a Codepen to capture some things with a modified version of our compiled CSS: https://codepen.io/emdeoh/pen/agZGqa?editors=1100.

patrickhlauke:
> Trying to figure out the HTML and CSS here... I'd like to avoid adding tons of new selectors to account for this. Might be impossible, but I'm not giving up just yet. 😬 Starting a Codepen to capture some things with a modified version of our compiled CSS: https://codepen.io/emdeoh/pen/agZGqa?editors=1100.

fundamentally, it's confusing to just piggy-back onto `btn-group` for what is not an actual button group, but rather a button-like input. i'd take a few extra selectors for clarity over a bastardisation of btn-group, particularly since these constructs aren't just button groups, but live conceptually under the button component itself (but then CAN be turned into an actual group).

patrickhlauke:
i.e. look here https://deploy-preview-28463--twbs-bootstrap.netlify.com/docs/4.3/components/buttons/#checkboxes-and-radio-buttons - these aren't "button groups" and reusing that class just to encapsulate the button-like "input followed by label" or "label wrapped around an input and some text acting as the visible label text" would be confusing and misleading, i'd say

ysds:
> fundamentally, it's confusing to just piggy-back onto btn-group for what is not an actual button group, but rather a button-like input. 

The current inline dropdown's markup is:

```html
<div class="btn-group">
  <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" ...>
    Action
  </button>
  <div class="dropdown-menu">
     ...
  </div>
</div>
```
(Color variant snippets in https://getbootstrap.com/docs/4.3/components/dropdowns/#single-button)

What do you think about this? That is not an actual button group.

(I'm still supporting using `.btn-group` for dropdowns and toggle buttons)

patrickhlauke:
not a fan of the use for dropdowns either, but can more easily rationalise that in that it groups a button and a dropdown together. less so when it's grouping only one thing (and a hidden radio/checkbox)

patrickhlauke:
incidentally, the `.btn-input-wrapper` class i proposed here could be renamed to something more generic like `.btn-wrapper`, and applying that to those dropdowns would have the same effect as well.

or perhaps taking a step back, renaming `.btn-group` to `.btn-wrapper`, and patching in any extra styles/selectors created here for `.btn-input-wrapper` and applying them to that, could also work. gives the class itself a more generic name, and it's then applied in different contexts where it "wraps" various things relating to a button-like look.

ysds:
I don't stick to the name of the `.btn-group`. I just don't want to increase the complexity of selectors and CSS. The changes to the `_button-group.scss` are very complex for now.

Looking for a wrapper class name (alt `.btn-group`) that everyone is convinced might be a great solution.

patrickhlauke:
btw note that even if we just reused `.btn-group` it wouldn't remove the need for *all* of the new selectors (like `.btn-input`), as they cover scenarios that weren't covered before just through `.btn-group`

patrickhlauke:
@mdo let me know if you want me to have a crack at renaming `btn-group` (and what your preference would be, if any) and re-simplifying some of the css selectors?

ysds:
> btw note that even if we just reused .btn-group it wouldn't remove the need for all of the new selectors (like .btn-input)

Yep, I showed that (just adding one new class) in my original PR #25185.

Johann-S:
Hey @twbs/css-review and @twbs/js-review  I removed QUnit and switch to Jasmine, so now your unit tests should go there: https://github.com/twbs/bootstrap/blob/master/js/src/button/button.spec.js

js/tests/unit/button.js have been removed, so if you need any help do not hesitate 😉

patrickhlauke:
@mdo just bumping this to see if it's likely to be used/picked up/worked on further...

patrickhlauke:
so...coming back to this, as i worked my socks off at the time to reorganise stuff, create some more explanations and examples, etc ... what's happening? should this all be closed? is it too far gone to salvage the bits that aren't controversial (since the creation/renaming of the class wasn't to everybody's liking)? /cc @mdo

patrickhlauke:
so... 5 months later, and still tumbleweed on this? makes me regret having put time and effort into trying to rationalise/improve stuff here to be honest...

XhmikosR:
Who's up to do a proper rebase for this branch? 😇

patrickhlauke:
cursed branch :)

ffoodd:
Just discovering this one! Looks interesting but pretty confused with all the things going on here, so if anyone's able to summarize the need, I'd be glad to help.

And if not, this topic might not be clear enough!

patrickhlauke:
in very broad strokes, from what i remember, this change does the following:

- kills off the clunky JS-based button groups for checkbox/radio button groups, instead relying purely on using the underlying `<input type="radio">` / `<input type="checkbox">` ... avoiding a LOT of the extra messing around we have had with these components in the past (all to make them behave like real radios/checkboxes)
- renames (currently, though that was then a big point of contention) the button group class
- reorders and expands examples to clarify and more clearly distinguish between the *styles* to make button groups and then the actual dynamic component to handle ARIA-based toggles

ffoodd:
If I understood correctly, we have three features that kind of interact with each others:
1. styling a button group — without any specific behaviour, just to style multiple appended buttons of any kind;
2. first behaviour: using `radio`s or `checkbox`es;
2. second behaviour: using JS and `aria-pressed`.

And after reading this dicussion carefully, I have a question: the first need for this PR was to remove JS. I can't see anyone in this discussion arguing to keep this JS part (except the fact that it already exists…). Am I wrong?

@patrickhlauke IMHO both behaviours are to be accomplished with `input`s.

Is there any reason to keep mentionning the ARIA pattern here? Am I missing something?

patrickhlauke:
ARIA is still needed for toggle buttons (`<button aria-pressed="true">`). faking them using, say, an `<input type="checkbox">` is not ideal/recommended.

patrickhlauke:
not to keep flogging a dead horse, but... @mdo et al, any idea on this whether we want to proceed or not?

MartijnCuppens:
@patrickhlauke, I'll have a look in another PR with another approach. So you're ok with dropping the js think with active classes and just rely on the CSS only label-buttons for v5?

patrickhlauke:
yes, the move to CSS only (except for controls/toggles that need `aria-pressed`) is non-controversial. but this PR also contains reordering/reorganisation of some docs aspects (moving the radio button group stuff, making up new classes for those situations, and lots more specific usage examples). it's that stuff as well that i'm quite keen to get a decision on

mdo:
Closing for #30650.

PR #28466, opened 2019-03-12
V4 dev rtl support

mohamedragaey:


XhmikosR:
First of all, you need to target the master branch.

Secondly try to clean up your commits for this to be reviewable.

Third and most important, do not commit any dist files.

PR #28472, opened 2019-03-13
[WIP] v5: Prep migration docs page

mdo:


site/content/docs/4.3/migration.md
```diff
@@ -40,7 +46,7 @@ Changes to any layout tools and our grid system.
 - **Todo:** Remove `position: relative` from grid columns
 - **Todo:** Integrate CSS grid into our grid system
 
-## Content changes
+## Content
```
XhmikosR: This was done on purpose otherwise we get a duplicate `#content`.

MartijnCuppens:
Can we turn this PR into a draft? The browsers versions need to be determined first.

XhmikosR:
I don't think there's a way to do that. You can set a PR as draft when opening it.

mdo:
> Can we turn this PR into a draft? The browsers versions need to be determined first.

The changes aren't set in stone here and we're under active development, so I'm not too worried about the order of operations there. I've kept the versions as `NN` and have them as placeholders.

PR #28484, opened 2019-03-13
Mock GA object when in Hugo server mode

peterblazejewicz:
The project started using Hugo after #28014
Now when in development mode the JavaScript errors occurs when using some links on the website:

```text
Uncaught ReferenceError: ga is not defined
    at HTMLAnchorElement.onclick ((index):75)
onclick | @ | (index):75
```

This minor change removes these errors during development.

Thanks!

### DIFF
--- a/site/layouts/partials/analytics.html
+++ b/site/layouts/partials/analytics.html
@@ -5,4 +5,8 @@
   ga('send', 'pageview');
 </script>
 <script async src="https://www.google-analytics.com/analytics.js"></script>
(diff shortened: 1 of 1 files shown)


XhmikosR:
I think I will just revert the local change. I find no use for it.

Thanks for the PR though :)

peterblazejewicz:
for ref:
https://github.com/twbs/bootstrap/pull/28489#issuecomment-472701600

PR #28486, opened 2019-03-14
browsers-devices.md: escape what we read.

XhmikosR:


mdo (changes requested):
Why are we reverting the GA exclusion here? We just approved and merged that change.

XhmikosR:
Because it introduces an JS error locally, and TBH I find this change useless myself.

mdo:
> Because it introduces a JS error locally, and TBH I find this change useless myself.

I don't see an error on my local dev for `master`. And if you find it useless, open an issue or PR to discuss that again—that change is unrelated to this PR.

XhmikosR:
It does introduce a JS error. Try clicking on the themes navbar link for example.

>ReferenceError: ga is not defined[Learn More]

mdo:
Split the commits myself—see cross-referenced PRs at #28488 and #28489.

PR #28487, opened 2019-03-14
Tweak our docs headers for proper document outline.

XhmikosR:
~~I need to go through the whole docs again after the Hugo switch to make sure I haven't missed anything.~~

Apart from this, I'm not sure if we should use the header classes (.h5, .h4) or not.


XhmikosR:
Ok, I finished the first pass. Please review carefully :)

patrickhlauke:
those header classes look confusing to me. i take it without them the examples/demos will look too big/odd? there's a danger that people will cargo-cult this code in their own projects and drag them along unnecessarily, or worse change the class (but leave the heading the way they copied it) to make it look appropriate to their needs...

patrickhlauke (commented):
apart from the use of the heading classes, as commented already, this seems fine. headings in code samples are always going to be rather odd/special. unfortunately, there's no way in html to say "yes, it's a heading, but it's just an example and doesn't really count as being part of the document itself..."

then again, even if those headings were left as they are, nobody dies if a few heading levels are skipped for those...

patrickhlauke (approved):
tentatively approving regardless...

XhmikosR:
I don't know myself if I like all of the changes. I just made a first pass to make sure the headers hierarchy is fine.

So, any tweaks are welcome. Even dropping some of the changes it's fine to me. We don't need to be 100% correct everywhere and since we don't automatically test for this, it's bound to break again at some point, anyway.

XhmikosR:
Any comment about this @mdo @MartijnCuppens ? See my comment above, I'm not sure all of these changes make sense. Some are good, others seems a little bit like an overkill.

XhmikosR:
I'm gonna drop this. If anyone wants to tackle this go ahead.

PR #28384, opened 2019-03-01
Don't run Google Analytics locally.

coliff:
This is for the Hugo docs.
Refs #28351

Issue #28351, opened 2019-02-24
Stop Allowing Google to use its "Backdoor" script loader


ghost:
I use [Dan Pollock's hosts file](https://someonewhocares.org/hosts/) to block Google Analytics but they found a way around it. I've seen this on the Telegram website and now I'm seeing it on Bootstrap while helping with the Hugo dev:

https://www.reddit.com/r/privacy/comments/8clrzn/google_your_backdoor_to_the_internet/

The way this works is Google uses a backdoor script loader as explained above—it's base64 encoded—the specific script Bootstrap tried to load on my machine while working doing development was this one here: https://www.google.com/js/bg/4mrRnEPnWo81qWPG8_xcGP85HCscbg1O2YUrIewxQYY.js

This script loader relies on the fact many may have explicitly blocked google analytics and it uses the technique to get around the block in a most uncouth way. If you'd like a HAR file showing the requests I will reproduce this for you—if you can commit to an investigation and fix for this cross-domain JS script loader.

Johann-S:
Hi @jhabdas,

I read what you linked to us, but I can't figure how it's related to Bootstrap source code 🤔

patrickhlauke:
i'll take a stab in the dark and say this relates to getbootstrap.com's analytics snippet? (which you'd also get when viewing offline/locally built site/documentation when developing)

jhabdas:
@patrickhlauke yep. it defeated my hosts file. i have more info on it which I can dig up but I haven't seen it in about 6 months (not a GA user personally—don't have a google account). again, if you need more research I can dig it up but the basics should be in the reddit link above.

jhabdas:
Overall I feel this can be mitigated with a tracking opt-out, which probably should've been added last may along with a so called "cookie disclaimer".

Ref: https://github.com/freeCodeCamp/devdocs/issues/971 https://github.com/badges/shields/issues/3026

bardiharborow:
We should disable GA for local docs, consider GDPR implications, and see what can be done to make GA well-behaved. Ideally we'd remove it in favour of server/CDN-side analytics but that's probably more mdo's call.

Lausselloic:
or put a cookie-consent manager like https://github.com/AmauriC/tarteaucitron.js ?

XhmikosR:
A cookie consent is doable although TBH they annoy me a lot.

@mdo: what do you prefer: opt-in or opt-out tracking?

bardiharborow:
[...]

PR #28489, opened 2019-03-14
Revert "Don't run Google Analytics locally. (#28384)

mdo:
Reverts #28384.

Second of two PRs to supersede #28486.

### DIFF
--- a/site/layouts/partials/analytics.html
+++ b/site/layouts/partials/analytics.html
@@ -1,8 +1,6 @@
-{{- if not .Site.IsServer -}}
 <script>
   window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
   ga('create', 'UA-146052-10', 'getbootstrap.com');
   ga('send', 'pageview');
 </script>
 <script async src="https://www.google-analytics.com/analytics.js"></script>
-{{- end -}}


mdo:
Currently doing some light reading into GDPR and GA. #28351 was opened by a deleted account and while the motivation for that issue is in question, there do seem to be some things we should audit and consider. Hence, this PR.

XhmikosR:
I don't this really solves anything hence the revert.

If we want to care about GDPR we should do it properly in another PR.

mdo:
Having reviewed our GA settings, the data retention period looks good and no data is being shared with Google's advertising or any other partner (thus, no need for consent to sharing data as I understand it). Whatever GA collects stays in GA.

Will revisit and confirm Carbon separately.

XhmikosR:
Just so we are on the same side, GDPR isn't about that. We need to inform every visitor their data is being collected and get their consent.

That being said, we can at least anonymize the IPs in GA. I will make a PR for this tomorrow.

mdo:
I understand GDPR is about both—collection and sharing. My last comment was from reading multiple articles that has that guidance for GA as it relates to sharing though. Still reading more about it though and will see what we need to do for the pure collection aspect.

I'm no expert, so keep me honest. Just sharing what I'm learning along the way.

> That being said, we can at least anonymize the IPs in GA. I will make a PR for this tomorrow.

Awesome, thanks for that!

mdo:
Actually while reviewing our GA settings, realized we're on an old script and we can anonymize from that new one. See first stab at #28491 before getting going @XhmikosR.

bardiharborow:
This wasn't about GDPR, but rather that local users should not have their information beamed to GA (because local files don't match the domain, I think GA rejects their pings anyway). The references to `ga` in onclick handlers need to be made conditional on `.Site.IsServer` as well and then this won't throw errors.

PR #28491, opened 2019-03-14
GA: anonymize IPs

mdo:
Follow-up to #28489.

/cc @XhmikosR


XhmikosR:
TBH, I personally haven't used the so called new snippet yet, and I think I remember why.. I'm still using the plain old inline script snippet.

With your snippet, it loads GTM, and then it loads GA. I personally don't see any gain from this, so I'd say no for this change.

The anonymize IPs change LGTM.

XhmikosR:
So, I'd personally go with the old way, not even what we have in master here.

```
(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

ga('create', 'UA-146052-10', 'auto');
ga('set', 'anonymizeIp', true);
ga('send', 'pageview');
```

And also move it in the footer. But these changes should happen in separate clear patches, not in one, so that we can track them in the future.

Just my 2 cents.

mdo:
Sounds good, I might need to confirm the anonymize IP setting there for that snippet but I can dig.

XhmikosR:
I pushed this here in a separate patch.

The rest, we can tackle in a separate one.

XhmikosR:
@mdo: should I apply this to all current docs in the gh-pages branch?

XhmikosR:
I went ahead and did it with a regexp, see #28501

coliff:
Yes, it's good to not use the newer gtag.js as it's not really an analytics.js replacement, it's a wrapper library on top of analytics.js (and others), meaning the gtag library still needs to load analytics.js, and it needs to convert all relevant gtag commands to analytics.js commands and run them.

As of right now, gtag is 54.5K and analytics.js is 33.7K, and since gtag will also load analytics.js, it's 262% more code for less functionality. There's also the problem that gtag is site-specific (because you add your tracking ID as a query string), so it's unlikely your users will have it cached; analytics.js on the other hand is used all over the web, so it's very likely to already be in your visitors cache when they load your site.

We recommend to analytics.js on html5boilerplate for this reason. The script here is best for performance; https://github.com/h5bp/html5-boilerplate/blob/master/dist/index.html

XhmikosR (approved):

PR #28500, opened 2019-03-15
v5: Simplify docs CSS build and tasks

mdo:
Currently the docs include Bootstrap's built CSS via ` css-copy` task, duplicating the `dist/css/` directory into `site/static/docs/4.3/dist/css/`. This task is currently busted in `master` for some reason, but it reminded me that there's likely a simpler way.

We already import the functions, variables, and mixins into `docs.scss`—so why not simply include all of Bootstrap? That's what this PR suggests. We're still generating the dist files, but this reduces the number of stylesheets in our docs and saves some npm tasks.

Any concerns about doing it this way? I'm not 100% sure yet, but doing this might mean I'm currently missing some tasks that get run on the `dist` copy but not the `docs` copy? And what about `dist/js/`—can we make the same improvements somehow?


XhmikosR:
I'm not in favor of this.

1. people still copy the files from the website
2. not sure what you mean by busted. If I had more details like which command is broken, I could try fixing it
3. the real solution to this is to use Hugo to build the docs CSS/prefix/minify. It's in the Hugo TODO issue.

mdo:
The thing I like most about this direction is it essentially treats the docs like how you'd build with Bootstrap in an ideal world—include, customize, and extend. It keeps me in the mindset of building for those others, and helps us see what they run into as well.

Regarding your points:

1. This doesn't mean folks can't copy-paste from our docs. It also doesn't mean we're not generating the `dist `files.
2. Running `npm run dist` and `npm start` didn't pick up any new CSS for me. Rerunning those tasks and trying them manually didn't bring in the new changes.
3.  If I understand you right, the same approach would apply there for Hugo, just with a different build tool. The ideal is to include Bootstrap into the source Sass files (and not via another `<link>` in the HTML), extend with new variables, and add our docs-specific styles. We'd just swap from npm tasks to Hugo's built in build tooling.

Does that sound right? Happy to hash it out more or to even take a look at Hugo's build tooling to see if I can make sense of it and try it out.

XhmikosR:
I tried this with just `npm run start` and made a change in the `/scss/` folder, and the docs reloaded just fine.

BTW, I don['t think you should even run `dist` before `start`. Because `start` already runs the nodemon scripts, which builds the core CSS and JS.

MartijnCuppens (dismissed):
I like this approach

XhmikosR:
The branch isn't ready, even if we decide to go with it.

XhmikosR:
Actually this is ready from my side.

I'm still not sure about the change, though. I do see it might simplify the dev process, but on the other hand, why not do the same for JS and so on.

Also, this might be worse for performance, I haven't measured it.

If our main issue is some of the scripts failing for dev, I could have a look as long as I can reproduce the issue.

MartijnCuppens:
We just got to make sure the bootstrap file is still available standalone (which [is the case](https://deploy-preview-28500--twbs-bootstrap.netlify.com/docs/4.3/dist/css/bootstrap.min.css)), might be handy to keep it for codepen demos ect.

> Also, this might be worse for performance, I haven't measured it.

I don't think that would matter much, it could be faster if one of the css files loads slower. Did one test, seemed to be a tiny bit faster: https://gtmetrix.com/compare/hQcD8HPN/VtmFPF7n

mdo:
@XhmikosR You brought back the script for copying the built `dist` CSS, which is something I was hoping to avoid doing. Is this because of the Examples we have, or another reason?

XhmikosR:
It's for the examples plus for generally keeping the folder structure we
had all these years with the dist folder having all of our dist files.

mdo:
> It's for the examples plus for generally keeping the folder structure we had all these years with the dist folder having all of our dist files.

So the root `dist` should stay, but other than the examples right now, we don't really need another `dist` folder elsewhere, which is what the `css-copy` task was built for due to Jekyll's limitations. Curious if we can get the examples away from this duplicated dist folder...

XhmikosR:
I thought we wanted to keep the dist folder of the website will all of our files? It's what we are doing since the beginning.

XhmikosR:
I'm going to close this since we no longer have the copy tasks.

PR #28501, opened 2019-03-15
GA: Anonymize IPs and use https.

XhmikosR:


coliff (changes requested):
The script in this PR isn't loading async and is slower than the one used in production now. I would suggest change move to:
```<script>
  window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
  ga('create', 'UA-146052-10', 'getbootstrap.com');
  ga('set', 'anonymizeIp', true);
  ga('send', 'pageview');
</script>
<script async src="https://www.google-analytics.com/analytics.js"></script>```

XhmikosR:
I don't see where you see it's not loaded async.

The only change I did was to add the anonymize IP setting. So, it's whatever it was before.

coliff:
ah, sorry- I had just quickly checked the first few diffs in this PR which were on the old Bootstrap 1.x docs which were/are using a much older analytics snippet so had wrongly assumed the same was used for all the docs. I think it's all good now 👍

XhmikosR:
Yeah, those are left as is for now, since, you know, it's just ancient docs.

mdo (approved):

PR #28503, opened 2019-03-15
Use favicon/apple-touch-icon from each individual folder instead of the root one.

### LINKED ISSUE #28502: gh-pages: fix meta paths (favicons mostly)
Currently, in the gh-pages branch we do `link rel=favicon here=/favicon.ico` etc.

While this works, the idea is to have each docs version completely independent.

I'm gonna take a stab later and change this to reference each version's assets only.

XhmikosR:
**Should be merged after #28501**

Fixes #28502


XhmikosR:
Ping @mdo

XhmikosR:
Ping @mdo

mdo (approved):

PR #28504, opened 2019-03-15
Examples: include favicons.

XhmikosR:
Refs #28502


XhmikosR:
@MartijnCuppens @mdo @Johann-S: thoughts? I did this in the old docs too, so that each version is completely independent. Before it worked because we also keep a favicon.ico in root for the live docs.

MartijnCuppens (approved):

mdo (approved):

PR #28508, opened 2019-03-15
Remove map merges

### LINKED ISSUE #28506: Ditch the sass maps
Continuing on https://github.com/twbs/bootstrap/pull/28445#issuecomment-473151602 and https://github.com/twbs/bootstrap/issues/27927#issuecomment-450583968. Copy/paste from @mdo s comment which describes the situation:

---

I think we need to ditch these built-in `map-merge`s for regular Sass maps. It now feels over-engineered, and perhaps poorly documented for how folks can customize things.

With the usage of `map-merge` everywhere, we have the ability to easily add key-value pairs, but not to remove them. There's no option in Sass to do this, but you can replace a value or add new ones (with `map-merge`). This is an issue because if you want fewer key-value pairs in a Sass map, you'll see errors like those of https://github.com/twbs/bootstrap/issues/27927#issuecomment-450060013.

So, what if we remove all `map-merge` in our variables?

Let's consider an example where we look at our `$grid-breakpoints`.

```scss
$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;
```

To replace all of the values, we redefine the keys' values:

```scss
$grid-breakpoints: (
  xs: 0,
  sm: 400px,
  md: 800px,
  lg: 1200px,
  xl: 1600px
);
```

To replace individual key-value pairs, we use `map-merge`:

```scss
$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px
) !default;

$grid-breakpoints: map-merge($grid-breakpoints, (xl: 1600px));
```

To add additional values, we also use `map-merge`:

```scss
$grid-breakpoints: map-merge(
  $grid-breakpoints,
  (
    xxl: 1600px,
    xxxl: 1800px
  )
);
```

To do a bit of everything, we can use replace the default values and use multiple `map-merge`s:

```scss
$grid-breakpoints: (
  xs: 300px,
  sm: 500px,
  md: 800px,
  lg: 1100px,
  xl: 1400px
);
$grid-breakpoints: map-merge(
  (
    xxs: 0
  ),
  $grid-breakpoints
);
$grid-breakpoints: map-merge(
  $grid-breakpoints,
  (
    xxl: 1600px,
    xxxl: 1800px
  )
);
```

**You can [see all of this in this Sassmeister gist](https://www.sassmeister.com/gist/245cd80c5efd1ed77ade9e2a6b322a08).**

MartijnCuppens:
Remove the complex map merges we currently have.

Fixes #28506

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -17,22 +17,19 @@ $gray-800: #343a40 !default;
 $gray-900: #212529 !default;
 $black:    #000 !default;
 
-$grays: () !default;
-// stylelint-disable-next-line scss/dollar-variable-default
-$grays: map-merge(
-  (
-    "100": $gray-100,
-    "200": $gray-200,
-    "300": $gray-300,
-    "400": $gray-400,
-    "500": $gray-500,
-    "600": $gray-600,
-    "700": $gray-700,
-    "800": $gray-800,
-    "900": $gray-900
-  ),
-  $grays
-);
+// fusv-disable
+$grays: (
+  "100": $gray-100,
+  "200": $gray-200,
+  "300": $gray-300,
+  "400": $gray-400,
+  "500": $gray-500,
+  "600": $gray-600,
+  "700": $gray-700,
+  "800": $gray-800,
+  "900": $gray-900
+) !default;
+// fusv-enable
 
 $blue:    #0d6efd !default;
 $indigo:  #6610f2 !default;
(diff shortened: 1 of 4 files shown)


browner12:
so what you're proposing is moving the `map_merge` calls outside of the Bootstrap code and into the userland code?

Don't we run into an order of operations issue here? Currently we define our overrides before importing the BS code.

```scss
$theme-colors: (
    "primary": #ff000,
    "secondary": #00ff00,
);

@import '~bootstrap/scss/bootstrap';
```

Now if we try to use the new code this will no longer work, because our variables won't have been defined yet.

```scss
$theme-colors: map-merge($theme-colors, ("mycolor": #0000ff));

@import '~bootstrap/scss/bootstrap';
```

You also can't move it after, because then none of your custom changes would be used.

This will force us to switch back to using all of the separate imports, and splicing our custom code in between.

```scss
@import "functions";
@import "variables";
@import "mixins";

$theme-colors: map-merge($theme-colors, ("mycolor": #0000ff));

@import "root";
@import "reboot";
````

Which isn't the end of the world, but I think this is the problem we were trying to solve when the current code was added.

MartijnCuppens:
> so what you're proposing is moving the `map_merge` calls outside of the Bootstrap code and into the userland code?

The idea is to simplify the maps in general and getting more control over the maps without having to know where the maps should be included.

MartijnCuppens:
The `$grays` variable is not used by Bootstrap. Ditch it?

MartijnCuppens:
I've updated the docs. Ready to review, @twbs/css-review

ysds (approved):
👍

mdo:
Yay! Thanks for getting this merged!

Issue #25497, opened 2018-01-28
Feature/Enhancement: The CSS of custom file input is unnecessarily dumbed-down and limiting.


kogli:
There are multiple things to this that I will discuss below. (And none of my suggestions break browser compatibility).

## CSS `attr()` function for the "Browse" text
Why do we need to keep translations for the "Browse" text in CSS? Such limiting workaround is unnecessary, since we can use the **well-supported** `attr()` function (it is **not** experimental technology if you are using it with a `content` property - [proof](https://developer.mozilla.org/en-US/docs/Web/CSS/attr)) for example like this:

```SCSS
.custom-file-label {
 // ...

  &::after {
    // ...
    content: attr(data-button);
  }
```

And then in HTML:
```HTML
<div class="custom-file">
  <input type="file" class="custom-file-input" id="id">
  <label class="custom-file-label" data-button="My browse text" for="id">My file text</label>
</div>
```

## HTML5's `<label>` _is allowed_ to contain other tags such as `<span>`
Proof [here](https://www.w3.org/TR/html5/sec-forms.html#the-label-element) and [here (see the "permitted content" section on top)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).

With this knowledge we are able to do tons of things, such as:

### Text truncation
What I was able to do is this:
![text truncation in practice](https://user-images.githubusercontent.com/7883957/35481941-bb9ec7ae-042d-11e8-9559-938c526742f0.png)

I achieved this in my custom project with this code: (note the added `<span>` tag)
```HTML
<div class="custom-file">
  <input type="file" class="custom-file-input" id="id">
  <label class="custom-file-label" data-button="My browse text" for="id"><span>Some looooooooooooooooooooooooooooong text</span></label>
</div>
```

(the SCSS code below extends Bootstrap's code)
```SCSS
.custom-file-label {
        display: flex;
        flex-direction: row;

        span {
            text-overflow: ellipsis;
            white-space: nowrap;
            overflow: hidden;
            padding-right: 20px;
        }

        &:after {
            content: attr(data-button);
            position: relative;
            margin: #{-$input-padding-y} #{-$input-padding-x} 0 auto;
         }
    }
```

### "Browse" button as a regular element
We can **ditch the `::after` element completely** and instead use a `<span>` tag to display the button. This allows us, for example, to use an SVG icon in place of the "Browse" text. (`<svg>` is permitted content of both `<label>` and `<span>` - see proof above.)

[...]

PR #28511, opened 2019-03-15
Fix custom file truncation

MartijnCuppens:
Closes https://github.com/twbs/bootstrap/pull/28497

Alternative approach which doesn't depend on the width of the browse button. This solution is also compatible with the non-`.text-truncate` custom input, the text just gets cut off in that case.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -328,26 +328,28 @@
   right: 0;
   left: 0;
   z-index: 1;
+  display: flex;
+  justify-content: space-between; // Push the button to the right side
   height: $custom-file-height;
   padding: $custom-file-padding-y $custom-file-padding-x;
   font-family: $custom-file-font-family;
   font-weight: $custom-file-font-weight;
   line-height: $custom-file-line-height;
   color: $custom-file-color;
+  white-space: nowrap;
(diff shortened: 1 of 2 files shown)


alecpl:
I'm not sure this is supported scenario, but that margin trick will not work with .form-control-lg/.form-control-sm without additional code.

MartijnCuppens:
Good point, I'll have a look at that

ysds:
I’m not a big fan of the current(4.3.1) implementation using CSS `content`. It seems a bit tricky.

If DOM changes are allowed in v5, I'd like to propose like a following change:

```html
<div class="custom-file">
  <input type="file" class="custom-file-input" id="customFile" lang="es">
  <label class="custom-file-label" for="customFile">
    <span class="custom-file-text">Choose file</span>
    <span class="custom-file-button">Browse</span>
  </label>
</div>
```

This not need negative margins.

MartijnCuppens:
I'm going to close this after feedback from @ysds & @alecpl. I think we should go the way @ysds is suggesting. I think we can also ditch the `attr()` hack mentioned here: https://github.com/twbs/bootstrap/issues/25497.

@ysds, could you have a look at this one of these days? (no pressure on this)

PR #28512, opened 2019-03-15
Move close icon to components in docs

MartijnCuppens:
The close icon is a component, not a utility.

### DIFF
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -1,5 +1,4 @@
 .close {
-  float: right;
   @include font-size($close-font-size);
   font-weight: $close-font-weight;
   line-height: 1;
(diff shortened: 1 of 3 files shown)


XhmikosR:
Should we backport this?

MartijnCuppens:
Nah, it's just the place where it's documented, not that important but if we're going to use the https://github.com/twbs/bootstrap/pull/28445 for utilities, it would be better if we move this.

XhmikosR:
@mdo: your thoughts?

ysds:
> The close icon is a component

My only concern is:

https://github.com/twbs/bootstrap/blob/06635be7bd4e1d469a25e48a0a36c4427f7d7fe9/scss/_close.scss#L2

If remove this, the `close` button is going to be a truly component.

MartijnCuppens:
`.close` is only used by `.modal` (positioned with flexbox) and `.alert` (positioned with `position: absolute`) so we can easily drop `float: right` (which I just did)

ysds:
LGTM👍,  but would be perfect by changing its name to "Close button" IMO 😄

ysds (approved):

MartijnCuppens:
It could also be a link in another context, so let's leave it for now.

PR #28514, opened 2019-03-16
More compact color-yiq function

MartijnCuppens:
Minor change. Use `if()` function instead of the less compact if test.


ysds (approved):
I don't think there are any advantage to change, but there is no problem.

mdo (approved):
I always forget there's a convention like this for writing style. Like @ysds said, no difference so I'm in favor.

Issue #21943, opened 2017-02-08
Expand the Sizing Utility to allow for responsive breakpoints like "w-xl-50" (css, feature, v4)


matthiku:
Currently, the **_Sizing_ utility** (http://v4-alpha.getbootstrap.com/utilities/sizing/) only allows for the sizing of the height or width of an element **without regard to the viewport size**.

So for instance the "w-50" class makes a lot of sense on a large device in order to limit the width of, for instance, a help text, but not so much on a small device. 

This could be resolved by adding **breakpoints** to the utility in the same way as it is done for the _Spacing_ utility:
- w-lg-25 (_width: 25%_ only on large+ devices)
- w-md-50 (_width: 50%_ only on large+ devices)
- w-sm-75 (...)
- w-100 (would be like _w-xs-100_)

This enhancement would not change existing usage, but would allow more flexibility for various device sizes.

### Summary: Add support for the following classes:
w-sm-100, w-sm-75, w-sm-50 and w-sm-25
w-md-100, w-md-75, w-md-50 and w-md-25
w-lg-100, w-lg-75, w-lg-50 and w-lg-25
w-xl-100, w-xl-75, w-xl-50 and w-xl-25
h-sm-100, h-sm-75, h-sm-50 and h-sm-25
h-md-100, h-md-75, h-md-50 and h-md-25
h-lg-100, h-lg-75, h-lg-50 and h-lg-25
h-xl-100, h-xl-75, h-xl-50 and h-xl-25

matthiku:
Why can't I see and add labels to this issue?

dospuntocero:
this pull request fixes this issue https://github.com/twbs/bootstrap/pull/21975

mdo:
We'll pass on this for now, thanks though.

mdo:
> Why can't I see and add labels to this issue?

Only project members of the repository's organization have access to set these.

sunyang713:
I was also looking for this when I first came across the sizing utilities and was disappointed that it didn't exist. Is there a good reason why there are breakpoints for things like [spacing](https://v4-alpha.getbootstrap.com/utilities/spacing/), but not this?

madmax:
+1 I have lots of this class in bootstrap3. For example on Media object I have 60px image on xs viewport and 80px on larger screens. 

Now with bootstrap i can just add

```
$sizes: (
  25: 25%,
  50: 50%,
  75: 75%,
  100: 100%,
  60px: 60px,
  80px: 80px,
[...]

PR #28517, opened 2019-03-16
Bootstrap 5 grid

### LINKED ISSUE #25254: [v4] Why are all col classes 'position: relative'?

### LINKED ISSUE #26512: Dropdown `data-boundary` option not working in scrollable column
When a dropdown is placed in a scrollable column (`.col`) the `data-boundary` option has no effect.
As @rinick pointed out [here](https://github.com/bootstrap-vue/bootstrap-vue/issues/1819) this is due the columns `position: relative;` property.

Is this something Bootstrap or popper.js can/should fix or has the user to deal with it?

See: https://codepen.io/jackmu95/pen/XqBWwx

### LINKED ISSUE #27072: No way to restore the default 30px gutter when using negative margin class
The new negative margin classes (#26825) allow to change gutters width in responsive. But there is no way to restore the default 30px, since spacing classes are based on rem (16px). I think that this is not a problem, but what do you think?

### LINKED ISSUE #28312: v5: Revisit `max-width` on grid classes
From #28264:

> Removes the `max-width` from grid classes, though this may need to be revisited because comments indicate it applies to Firefox, too.

This change was reverted by @XhmikosR before he merged the PR into the no-jQuery branch, so I need to revisit to see if it's still true.

### LINKED ISSUE #29505: Size utility screen dependent
Would be nice to have 

```
w-100
w-sm-100
w-lg-100
```

and so on...

I am using Flex utility and I want to make a sort of grid but not by rows. Automatically adjustable. And I need to control width of every element so that it looks as grid and can change width depending on screen size. It would be analog of `col-md-3` in grid but for flex.

### LINKED ISSUE #29866: Should row-cols-* support auto-sizing?
Bootstrap 4.4.0 added the row-cols-* classes (awesome job btw, so nice to have).

I did notice that it does not add classes for autosizing columns based on breakpoints and that might be something to add.

Consider the following case:
- For small devices we set `.row-cols-2`
- For medium devices `.row-cols-md-4`
- But for some reason we might want large devices to auto-size the columns, thus arises the need for `.row-cols-lg-auto` (not currently existing)

This is currently solvable by setting `.row-cols-2.row-cols-md-4` on the row and then setting `.col-lg-auto` on the individual cols.

I think that this addition would give even greater flexibility in using the new .row-cols without the need to add classes to the nested columns.

I can create a PR for it, if we decide that it is a worthwhile addition. I read through parts of the discussion with row-cols but didn't manage to find anything regarding this (sorry if I have missed any prior discussions about it).

MartijnCuppens:
### Set gutter width in `rem` instead of `px`

The gutter width is now `1.5rem` instead of `30px`. This way we 'll be able to use the spacing utilities to align things with the grid.

The container paddings are now set to `1rem`.

Fixes #27072

### Change approach on how column paddings are set

The col classes were extended lead to this list of [60 selectors](https://github.com/twbs/bootstrap/blob/a88d58605a2aa864eb28da683c187824a580b7ec/dist/css/bootstrap.css#L606-L616) in `v4`. By using `.row > *`, we can simplify this a lot. This change will apply paddings to all children of `.row`. Specificity won't be influenced by this change.

### Columns: switch to `width` instead of `flex-basis` and `max-width`

Just using width has the benefit the `.col-*` classes can be used without the need of a `.row` class. This is also how the [primer](https://styleguide.github.com/primer/objects/grid/) works.

This also provides a solution for sizing utilities https://github.com/twbs/bootstrap/issues/21943. Because we only set paddings to direct children of `.rows`, the cols won't have any padding whenever they're not in a `.row`.

Closes #28312
Closes https://github.com/twbs/bootstrap/issues/29505

### More control over gutter widths & vertical gutters

This PR introduces new responsive gutter classes. There are 3 types of gutter classes available:
- `gx-*` classes control the horizontal/column gutter width
- `gy-*` classes control the vertical/row gutter width
- `g-*` classes control the horizontal & vertical gutter width

These gutter classes can be added to the `.row` and influence  the negative margins on the row and the padding on the columns.

Responsive variants are also available to get control per breakpoint. With this change we might consider ditching (or disable by default) the negative margins which increase our file size quite a lot. 

#### How do the gutters work?

The way the gutters are set in horizontal direction is kept the same as in v4 (negative margins on the row and paddings on the columns). The vertical gutters work a little different. Margin is added to the top of each column and to counteract the top margin, a negative margin is added to the row. We use margins instead of paddings to prevent overlapping issues (like we have with the horizontal paddings). 

### New `.row-cols-auto`

Fixes https://github.com/twbs/bootstrap/issues/29866

### Removal of `.form-row`

`.form-row`s had a smaller gutter width, but since we now have the gutter classes, we can use them for even more control over the gutter widths.

### Removal of `.form-inline`

`.form-inline` is removed in favor of the more flexible grid. The children can now be wrapped in `.col-md-auto` divs when needed. With the gutter classes, we can have easier control over the vertical spacing instead of using the responsive margin utilities.

### Remove `position: relative` from cols

Closes #25254
Closes #26512

### Removal of card decks

We currently have as well card decks as the grid system, but our grid offers more responsive control, so there's not really a reason to keep the decks.

### Remove global `box-sizing` reset from `bootstrap-grid.css`

In `bootstrap-grid.css`, `box-sizing` was inherited which introduces this issue: https://github.com/twbs/bootstrap/issues/22872. On the other hand, setting the global `box-sizing` behaviour can introduce unexpected behaviour for custom CSS. By only adding `box-sizing` to the columns, we only apply the `box-sizing` to the elements where it's needed.

## Examples

- [Grid documentation](https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/layout/grid/)
- [Form layout documentation](https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/forms/layout/)
- [Migration guide](https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/migration/#grid-and-layout)

### To do:

- [x] Change default gutter width to `1.5rem`
- [x] Change gutter classes
- [x] Document breaking changes
- [x] Update grid documentation
- [x] Document stand alone `.col-*` utility classes 
- [x] Update examples
- [x] `.form-inline` & `.form-row` are removed, document it
- [x] Update _stretched link_ documentation (assumes col has `position: relative`)
- [x] Cleanup git


ysds:
Umm... using universal selector is one of the most complex things of CSS, so it is very difficult to comment on this.

It looks like there is a dilemma, judging from the difference of the selectors between the 60 classes and the `.no-gutters`. At least, I agree with the performance problem of the `.no-gutters` selector. If we take consistency seriously, the changes will be right. There is no problem with the breaking change.

MartijnCuppens:
> Umm... using universal selector is one of the most complex things of CSS, so it is very difficult to comment on this.

What exactly do you mean with _complex_?

Also just added an edit for `.form-row`, the same technique from `.no-gutters` was used there.

ysds:
Sorry for late reply.

> What exactly do you mean with complex?

It means that it's difficult to understand and use them effect. Universal selectors have the power to easily break other styles. 

For example, expects the following case? (Aside from whether the such usage is good or bad):
https://codepen.io/anon/pen/oOwYQO

MartijnCuppens:
> For example, expects the following case? (Aside from whether the such usage is good or bad): codepen.io/anon/pen/oOwYQO

Aside from the fact that's not a good idea, the theming can easily be overridden with a class (even without `!important` because `*` has no effect on specificity: https://codepen.io/MartijnCuppens/pen/yroKYz

mdo:
Haven't spent nearly enough time in Bootstrap work, but what about something akin to Primer? For example: https://styleguide.github.com/primer/objects/grid/#flexbox-grids

MartijnCuppens:
@mdo, this PR only covers the way paddings are set. We could also change the grid so that the `.col-*` classes can be used without the need of a `.row` parent like Primer. This would also provide a solution for https://github.com/twbs/bootstrap/issues/21943.

Shall I extend the scope of this PR and also add that change here?

mdo:
Perhaps! What are your thoughts on going that direction? It'd be one of the bigger changes to arguably one of our biggest and most well known features.

MartijnCuppens:
I've experimented with widths instead of flex-basis and it looks like we can drop the max-widths, gonna check some browsers and test some edge cases, but it looks promising. I'll tweak this PR further later this week!

jackmu95:
Would also close #26512.

MartijnCuppens:
Updated the PR and introduced a second grid. The second grid mimics the behaviour of the block gutters of `display: grid`. Documentation is still lacking, we'll probably need to have a look at how we combine this PR with #28450.

mdo:
@MartijnCuppens Looking at this new secondary grid, maybe we just use `display: grid`? Support appears close enough https://caniuse.com/#feat=css-grid.

MartijnCuppens:
IE doesn't support auto placement or gutters. Can't do this yet.

Will update this PR soon. Not sure if we should go for the second grid (`.grid`) or if we should introduce gutter classes in x and y direction. What do you think, @mdo?

mdo:
Ah, wasn't aware of those shortcomings. Vertical margins for gutters have been asked in the past, so if that saves on a few selectors (adding `.grid` to `.row`), I'm in favor.

robrez:
might anything be in the works for the negative margins on `.row`? I think those exist to offset grids being placed in a container, but the rows/cols otherwise work fine w/o container

MartijnCuppens:
I've updated the PR.

I went for a gutter approach with `.g-*` classes, I think this 'll be the easiest to understand. The gutter classes now work alike the spacing utilities.

I've updated all the docs, and this PR is now ready for review.

If anything is unclear, let me know

midzer:
I want to through into consideration that using `*` may be somewhat less performant according to https://csswizardry.com/2011/09/writing-efficient-css-selectors/#the-key-selector

Performance tests might help to show differences between old and new grid.

MartijnCuppens:
@midzer, that article is quite outdated (2011). The universal selector in combination with the combinator shouldn't be a bottleneck.

> Performance tests might help to show differences between old and new grid.

Any suggestions on how to benchmark this?

midzer:
I've built something: https://github.com/midzer/bootstrap-grid-performance

Conclusion: at least on my machine and with given data, v5 performs slightly better, but differences are marginal

MartijnCuppens:
Thanks for looking into this @midzer!

scss/_grid.scss
```diff
@@ -44,18 +44,52 @@
...
+          margin-left: -$value / 2;
+
+          > * {
+            padding-right: $value / 2;
+            padding-left: $value / 2;
+          }
+        }
+
+        .g#{$infix}-#{$key},
+        .gy#{$infix}-#{$key} {
```
mdo: Should this be `.gb-` since it's only setting `margin-bottom`? And, since it's only `margin-bottom`, does this double up on margin utilities or is it adding meaningful changes/shortcuts?

MartijnCuppens: Nope, it adds a bottom margin to the cols, and adds a negative margin to the row, therefore there are only margins between the cols in vertical direction.

You can check out the forms docs to see how that works: https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/forms/layout/

mdo: Rephrasing, `.my-*` is currently used as a shortcut to `margin-top` and `margin-bottom`. In the `.gy-` proposed here, we're only setting `margin-bottom`. It sets the wrong expectation that `.gy-` is affecting only a single vertical `margin` property as opposed to two.

Does that make sense?

MartijnCuppens: The `.gy-` classes don't only add margin to the bottom but also counteract the margins of the last cols. So we don't have margins below the cols, but between the cols. The `gb-` class might not be ideal to cover this?

Alternative namings could be:
- `.gap`, `.gapx` & `.gapy`
- `.g`, `.g-inline` & `.g-block` (using the [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties))
- `.gap`, `.row-gap` & `.column-gap` (more like the grid [gaps](https://developer.mozilla.org/en-US/docs/Web/CSS/gap))

florianlacreuse:
@MartijnCuppens Quick questions:

1. Why does the vertical gutter use margin on `col`  elements instead of padding just like the horizontal gutter? May be in some cases we would like to be able to use margin top/bottom auto **and** keep a vertical gutter.
2. Why does the vertical gutter only apply margin (padding?) bottom on the `col` elements and not split in two (top/bottom) just like the horizontal gutter (left/right)? This could resolve your discussion with @mdo about the meaning of `.gy-`?

MartijnCuppens:
> Why does the vertical gutter use margin on col elements instead of padding just like the horizontal gutter? 

If we're going to use padding, elements below the row won't be clickable anymore.

> May be in some cases we would like to be able to use margin top/bottom auto and keep a vertical gutter.

These margin's should be set on a wrapper class

> Why does the vertical gutter only apply margin (padding?) bottom on the col elements and not split in two (top/bottom) just like the horizontal gutter (left/right)? This could resolve your discussion with @mdo about the meaning of .gy-?

That's a possible, It would add a bit more css though.

florianlacreuse:
> > Why does the vertical gutter use margin on col elements instead of padding just like the horizontal gutter?
> 
> If we're going to use padding, elements below the row won't be clickable anymore.

Good catch! I no longer had this css issue in mind, thanks for the reminder.

But this issue can already occured because of the horizontal gutter (using padding) if the clickable elements are on the left / right of the grid. Example: https://codepen.io/florianlacreuse/pen/GRRjeeP

> > May be in some cases we would like to be able to use margin top/bottom auto and keep a vertical gutter.
> 
> These margin's should be set on a wrapper class

Currently, margin utilies are directly applied on `col`s: https://getbootstrap.com/docs/4.3/layout/grid/#margin-utilities

But may be BS grid is not made to be used like this for instance: https://codepen.io/florianlacreuse/pen/KKKgYVM ?

> > Why does the vertical gutter only apply margin (padding?) bottom on the col elements and not split in two (top/bottom) just like the horizontal gutter (left/right)? This could resolve your discussion with @mdo about the meaning of .gy-?
> 
> That's a possible, It would add a bit more css though.

May be the right question is, if I may ask: why does horizontal gutter split gutter width and not only apply it on the left or on the right of a `col`? I raise the point just because I like consistency, but may be you have some technical limitations?

MartijnCuppens:
I like to keep the horizontal padding gutters for these reasons:
- It'll be easier to upgrade from `v4` to `v5`
- If we're going to switch to margin's we'll override the gutter widths when using `mr-*` or `ml-*` utilities
- We can now (in this PR) use the cols as as stand alone "utilities", if we're going to switch to margins, we'll need a lot more CSS.

Also keep in mind the vertical spacing is only added when the grid classes are used. You'll still be able to not use the grid classes and use the margin utilities on the columns if you need more control over spacings.

florianlacreuse:
@MartijnCuppens Ok, thanks for the explanations.

Having horizontal gutters between cols and in the same time being able to use `mr-*` / `ml-*` (especially `mr-auto` / `ml-auto`) on `col` is very useful, we use it a lot in our projects. I know this only works because of the use of both padding and margin properties.

I was hoping we could also have it with the upcoming vertical gutters. I undestand it will not be possible since the "padding issue" (close up and bottom elements not clickable, as you said) will badly affect the vertical flow (also happen for the horizontal flow as shown in my previous codepen but seems less critical, you guys confirm?), and there is not other way around.

Thanks again for your time!

MartijnCuppens:
> also happen for the horizontal flow as shown in my previous codepen but seems less critical, you guys confirm?

Yup, that's why I went for the margin approach 😉

MartijnCuppens:
Pushed one (last) change, the vertical gutters now use top margins. The benefits:
- We encourage to set bottom margins (https://twbs-bootstrap.netlify.com/docs/4.3/content/reboot/#approach). With these top margins, these margins won't conflict with utilities
- `top` is shorter than `bottom`, this change will make load Bootstrap incredibly fast since we 'll need to load less bytes 🚀
- The top margins won't cause any page overflow (Check https://codepen.io/MartijnCuppens/pen/QWWaaQp?editors=1100 to see what I mean)

michaelrobertson-fico-com:
Apologies if this is not the right place for the question. Are there plans for a true CSS Grid implementation in Bootstrap? It doesn't appear this is in the works for V5. We want to incorporate CSS Grid into our design system AND have the same compositional power we do with Bootstrap 4's flexbox utility classes. The two layout mechanisms are supposed to be complimentary. Will Bootstrap eventually support both?

MartijnCuppens:
> Are there plans for a true CSS Grid implementation in Bootstrap?

We can't due to IE11 support, but we might switch to CSS grid once we drop it in Bootstrap 6 (but this can take a while).

voltaek:
It might be beneficial to combine this with #29146.

MartijnCuppens:
> It might be beneficial to combine this with #29146.

This PR already has a lot of changes. I'm going to put #29146 on hold until we've merged this (and we might do some other folder/file restructuring we can combine with that PR).

auvipy (approved):

site/content/docs/4.3/components/navbar.md
```diff
@@ -66,9 +66,13 @@ Here's an example of all the sub-components included in a responsive light-theme
           <a class="nav-link disabled" href="#" tabindex="-1" aria-disabled="true">Disabled</a>
         </li>
       </ul>
-      <form class="form-inline my-2 my-lg-0">
-        <input class="form-control mr-sm-2" type="search" placeholder="Search" aria-label="Search">
-        <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+      <form class="row g-2 mt-3 mt-lg-0 pl-lg-2">
```
mdo: This example is busted in the docs.

Screen Shot 2020-01-09 at 1 08 39 PM

site/content/docs/4.3/components/navbar.md
```diff
@@ -66,9 +66,13 @@ Here's an example of all the sub-components included in a responsive light-theme
...
+        <div class="col-auto">
```
mdo: Know we're saving on CSS with using the grid for inline forms, but I imagine folks will complain about this extra HTML. Just calling out for awareness, don't make any changes for this just yet.

site/content/docs/4.3/components/navbar.md
```diff
@@ -245,29 +259,25 @@ Input groups work, too:
...
       </div>
-    </form>
-  </div>
+      <input type="text" class="form-control" placeholder="Username" aria-label="Username" aria-describedby="basic-addon1">
+    </div>
+  </form>
 </nav>
 {{< /example >}}
 
 Various buttons are supported as part of these navbar forms, too. This is also a great reminder that vertical alignment utilities can be used to align different sized elements.
```
mdo: We should call out that if your navbar is an entire form, or mostly form, you can use the `<form>` element as the container and save some HTML. Applies to the option above and below this copy. Otherwise this catches me off guard.

site/content/docs/4.3/content/typography.md
```diff
@@ -239,15 +239,15 @@ Remove the default `list-style` and left margin on list items (immediate childre
 </ul>
 {{< /example >}}
 
-### Inline
+### Inline lists
```
mdo: By code, this works, but for the mental model of it, I wouldn't think to use grid classes to make something appear what was previously `display: inline-block`. Feels like this exchanges minor file size for a relationship we don't need (between quick text lists and gridded content).

scss/_variables.scss
```diff
@@ -318,9 +318,22 @@ $container-max-widths: (
 // Set the number of columns and specify the width of the gutters.
 
 $grid-columns:                12 !default;
-$grid-gutter-width:           30px !default;
+$grid-gutter-width:           2rem !default;
```
mdo: Thoughts on moving to `1.5rem` here? Feels a little too spaced out at certain sizes right now.

site/content/docs/4.3/examples/carousel/index.html
```diff
@@ -24,9 +24,13 @@
             <a class="nav-link disabled" href="#" tabindex="-1" aria-disabled="true">Disabled</a>
           </li>
         </ul>
-        <form class="form-inline mt-2 mt-md-0">
-          <input class="form-control mr-sm-2" type="text" placeholder="Search" aria-label="Search">
-          <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+        <form class="row gx-2 mt-2 mt-md-0">
```
mdo: Similar to the form in the navbar docs, this is busted at narrow viewports.

site/content/docs/4.3/examples/navbar-fixed/index.html
```diff
@@ -23,9 +23,13 @@
           <a class="nav-link disabled" href="#" tabindex="-1" aria-disabled="true">Disabled</a>
         </li>
       </ul>
-      <form class="form-inline mt-2 mt-md-0">
-        <input class="form-control mr-sm-2" type="text" placeholder="Search" aria-label="Search">
-        <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+      <form class="row gx-2 mt-2 mt-lg-0">
```
mdo: Another form to fix.

site/content/docs/4.3/examples/navbar-static/index.html
```diff
@@ -23,9 +23,13 @@
           <a class="nav-link disabled" href="#" tabindex="-1" aria-disabled="true">Disabled</a>
         </li>
       </ul>
-      <form class="form-inline mt-2 mt-md-0">
-        <input class="form-control mr-sm-2" type="text" placeholder="Search" aria-label="Search">
-        <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+      <form class="row gx-2 mt-2 mt-lg-0">
```
mdo: Another form to fix.

site/content/docs/4.3/examples/navbars/index.html
```diff
@@ -32,7 +32,7 @@
           </div>
         </li>
       </ul>
-      <form class="form-inline my-2 my-md-0">
+      <form class="my-2 my-md-0">
```
mdo: Wondering if we still need the `my-2` on these forms? Feels like we're jumping the height on them too much on xs/sm viewports.

site/content/docs/4.3/examples/offcanvas/index.html
```diff
@@ -38,9 +38,13 @@
           </div>
         </li>
       </ul>
-      <form class="form-inline my-2 my-lg-0">
-        <input class="form-control mr-sm-2" type="text" placeholder="Search" aria-label="Search">
-        <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+      <form class="row gx-2">
```
mdo: Maybe related to this exact line: the navbar container is broken (content touches the edge of the viewport). Happens on the navbar-bottom example, too.

site/content/docs/4.3/examples/starter-template/index.html
```diff
@@ -32,9 +32,13 @@
           </div>
         </li>
       </ul>
-      <form class="form-inline my-2 my-lg-0">
-        <input class="form-control mr-sm-2" type="text" placeholder="Search" aria-label="Search">
-        <button class="btn btn-secondary my-2 my-sm-0" type="submit">Search</button>
+      <form class="row gx-2 mt-2 mt-lg-0">
```
mdo: Same issue as offcanvas navbar—content touches edges of viewport.

site/content/docs/4.3/forms/layout.md
```diff
@@ -189,44 +175,36 @@ At times, you maybe need to use margin or padding utilities to create that perfe
 Be sure to use `.col-form-label-sm` or `.col-form-label-lg` to your `<label>`s or `<legend>`s to correctly follow the size of `.form-control-lg` and `.form-control-sm`.
 
 {{< example >}}
-<form>
-  <div class="mb-3 row">
-    <label for="colFormLabelSm" class="col-sm-2 col-form-label col-form-label-sm">Email</label>
-    <div class="col-sm-10">
-      <input type="email" class="form-control form-control-sm" id="colFormLabelSm" placeholder="col-form-label-sm">
-    </div>
+<form class="row g-2">
```
mdo: These are spaced out much too far on narrow viewport:

Screen Shot 2020-01-09 at 1 30 38 PM

mdo (commented):
Something is off on the navbar color scheme page, perhaps an incorrectly closed element? Didn't an obvious problem in first look here.

Screen Shot 2020-01-09 at 1 13 17 PM

mdo:
Had some false starts reviewing this, sorry for the delay.

Aside from the comments above, I'm having difficulty wrapping my head around the `margin-top`, and when I see it break the form layouts in one of the review comments above, it makes me nervous. It's a new property to apply to our grids and seems to not come into play most of the time, other than forms? Feels like we simply shouldn't have any vertical margin of any kind by default, and you can add that with the `.gy-` classes.

This gets back to the conversation I just marked as resolved that we had a long time ago here about those classes—if we do remove this new `margin-top`, I could see us adding a `gb-` and `gt-` class, but that's obviously something to ignore until we can agree on the need for `margin-top`.

scss/_grid.scss
```diff
@@ -44,18 +44,54 @@
 @if $enable-grid-classes {
   .row {
     @include make-row();
-  }
+    @include list-unstyled(); // Remove padding and list styles for `ul`s
```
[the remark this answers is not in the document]
MartijnCuppens: Done:
```scss
  ul,
  ol {
    // stylelint-disable-next-line selector-no-qualifying-type
    &.row {
      @include list-unstyled();
    }
  }
```

site/content/docs/4.3/content/typography.md
MartijnCuppens: Ok, I just restored the inline lists.

site/content/docs/4.3/forms/layout.md
MartijnCuppens: Nice catch. I've added a wrapping `.row` per form element now.

MartijnCuppens:
> Something is off on the navbar color scheme page, perhaps an incorrectly closed element? Didn't an obvious problem in first look here.

These were caused by the latest Hugo update (https://github.com/twbs/bootstrap/pull/29750). Fixed it by removing newlines.

site/content/docs/4.3/components/navbar.md
MartijnCuppens: Indeed, the grid is a bit overkill to align these. I've used `.d-flex` here.

site/content/docs/4.3/examples/navbars/index.html
MartijnCuppens: I've now added a margin between the navigation and the form element and removed the additional bottom margin which was set here.

micka-fdz:
Hi @MartijnCuppens, according to your description:

> `gy-*` classes control the vertical/row gutter width

But in your PR, you are using `gt-*` classes.
I think we should use `gy-*` classes like margin and padding.

MartijnCuppens:
@micka-fdz,

I had been experimenting with these `.gt-` classes, but they felt a bit weird indeed. Switched back to `gy-` now.

micka-fdz:
Thanks @MartijnCuppens, it seems better to me indeed.

MartijnCuppens:
This PR is now ready to review, @twbs/css-review could you have a look at this?

site/content/docs/4.3/components/navbar.md
```diff
@@ -214,14 +213,14 @@ You may also utilize dropdowns in your navbar nav. Dropdown menus require a wrap
...
-Place various form controls and components within a navbar with `.form-inline`.
+Place various form controls and components within a navbar:
 
 {{< example >}}
 <nav class="navbar navbar-light bg-light">
   <div class="container-fluid">
-    <form class="form-inline">
-      <input class="form-control mr-sm-2" type="search" placeholder="Search" aria-label="Search">
-      <button class="btn btn-outline-success my-2 my-sm-0" type="submit">Search</button>
+    <form>
```
ysds:
```suggestion
    <form class="d-flex">
```

site/content/docs/4.3/examples/carousel/carousel.css
```diff
@@ -38,11 +38,6 @@ body {
 /* MARKETING CONTENT
 -------------------------------------------------- */
 
-/* Center align the text within the three columns below the carousel */
-.marketing .col-lg-4 {
-  margin-bottom: 1.5rem;
-  text-align: center;
```
ysds: If remove this `text-align: center;`, please remove the below too:
```scss
.marketing .col-lg-4 p {
  margin-right: .75rem;
  margin-left: .75rem;
}
```

scss/_variables.scss
```diff
@@ -318,9 +318,10 @@ $container-max-widths: (
 // Set the number of columns and specify the width of the gutters.
 
 $grid-columns:                12 !default;
-$grid-gutter-width:           30px !default;
+$grid-gutter-width:           1.5rem !default;
```
ysds: @mdo  I prefer 2rem close to 30px. Please double check that you intend that this 1.5rem affects the container padding. You can see the examples here:

* Before: https://twbs-bootstrap.netlify.com/docs/4.3/examples/navbars/
* After: https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/examples/navbars/

site/content/docs/4.3/examples/checkout/index.html
```diff
@@ -2,7 +2,7 @@
 layout: examples
 title: Checkout example
 extra_css:
-  - "form-validation.css"
+- "form-validation.css"
```
ysds:
```suggestion
  - "form-validation.css"
```

ysds:
My review list:

- [x] bundlesize.config.json
- [x] scss/_card.scss
- [x] scss/_forms.scss
- [ ] scss/_grid.scss
- [ ] scss/_variables.scss
- [x] scss/bootstrap-grid.scss
- [x] scss/forms/_layout.scss
- [ ] scss/mixins/_grid.scss
- [x] docs/4.3/components/card.md
- [x] docs/4.3/components/navbar.md
- [x] docs/4.3/examples/album/index.html
- [x] docs/4.3/examples/blog/index.html
- [x] docs/4.3/examples/carousel/carousel.css
- [x] docs/4.3/examples/carousel/index.html
- [x] docs/4.3/examples/checkout/index.html
- [x] docs/4.3/examples/grid/grid.css
- [ ] docs/4.3/examples/grid/index.html
- [ ] docs/4.3/examples/navbar-bottom/index.html
- [ ] docs/4.3/examples/navbar-fixed/index.html
- [ ] docs/4.3/examples/navbar-static/index.html
- [ ] docs/4.3/examples/navbars/index.html
- [ ] docs/4.3/examples/offcanvas/index.html
- [ ] docs/4.3/examples/starter-template/index.html
- [ ] docs/4.3/examples/sticky-footer-navbar/index.html
- [ ] docs/4.3/forms/form-control.md
- [ ] docs/4.3/forms/layout.md
- [ ] docs/4.3/forms/overview.md
- [ ] docs/4.3/forms/validation.md
- [ ] docs/4.3/helpers/stretched-link.md
- [ ] docs/4.3/layout/grid.md
- [ ] docs/4.3/migration.md

scss/_variables.scss
MartijnCuppens: The container padding can work independent from the gutter width. I'm ok with changing the container padding to `1.5rem`.

site/content/docs/4.3/forms/validation.md
```diff
@@ -98,31 +94,10 @@ Custom feedback styles apply custom colors, borders, focus styles, and backgroun
...
-        if (!form.checkValidity()) {
-          event.preventDefault();
-          event.stopPropagation();
-        }
-
-        form.classList.add('was-validated');
-      }, false);
-    });
-})();
-</script>
```
ysds: I don't think this script should be hidden. If you want it, please create another PR.

site/content/docs/4.3/forms/layout.md
```diff
@@ -236,126 +224,128 @@ As shown in the previous examples, our grid system allows you to place any numbe
...
 {{< /example >}}
 
 ## Inline forms
 
-Use the `.form-inline` class to display a series of labels, form controls, and buttons on a single horizontal row. Form controls within inline forms vary slightly from their default states.
-
-- Controls are `display: flex`, collapsing any HTML white space and allowing you to provide alignment control with [spacing]({{< docsref "/utilities/spacing" >}}) and [flexbox]({{< docsref "/utilities/flex" >}}) utilities.
-- Controls and input groups receive `width: auto` to override the Bootstrap default `width: 100%`.
-- Controls **only appear inline in viewports that are at least 576px wide** to account for narrow viewports on mobile devices.
+Use the `.col-auto` class to create horizontal layouts. By adding the `.g-2` class, we'll have gutters in horizontal and vertical directions. The `.align-items-center` aligns the form elements to the middle, making the `.form-checkbox` align properly.
```
ysds: > By adding the `.g-2` class,

But the following example uses `g-3`.  My suggestion is

> By adding [gutter modifier classes]({{< docsref "/layout/grid#gutters" >}}) class,

scss/_grid.scss
```diff
@@ -5,18 +5,52 @@
...
-  // Remove the negative margin from default .row, then the horizontal padding
-  // from all immediate children columns (to prevent runaway style inheritance).
-  .no-gutters {
-    margin-right: 0;
-    margin-left: 0;
+  ul,
+  ol {
+    // stylelint-disable-next-line selector-no-qualifying-type
+    &.row {
+      @include list-unstyled();
```
ysds: I'm worried whether this reset should be included. This might be unexpected behavior and I don't know how often this will be used. It just be used only one location in the new document.

If it's not the core of this PR, I prefer discuss it in a separate issue.

site/content/docs/4.3/forms/validation.md
MartijnCuppens: I moved this to the end of the file to fix form validation for https://twbs-bootstrap.netlify.com/docs/4.3/forms/validation/?#custom-styles. Shall make another PR for this

site/content/docs/4.3/examples/carousel/carousel.css
MartijnCuppens: Reverted this change

scss/_grid.scss
mdo: I had suggested the scoping of these at least to `ul, ol` earlier in the PR. It makes sense if we want folks to be using this kind of markup as a shortcut, but we could just as easily ask folks to use `.list-unstyled` in addition to `.row`.

site/content/docs/4.3/forms/validation.md
ysds: I missed [the comment](https://github.com/twbs/bootstrap/pull/28517#discussion_r364956887). `ul.row` raises the CSS priority, so for example overwriting its `padding-left` may cause confusion.

```html
<ul class="row custom-class">...</ul>
```
```css
.custom-class {
   padding: 1rem;
}
```

It is a safe way for users to add `.list-unstyled` deliberately. Or, I think as in the first @MartijnCuppens 's suggestion, applying rest list style to `.row` without limiting to `ul` and `li` may not cause confusion of CSS priority. e.g.:

```css
.row {
  ...
  padding: 0;
  list-style: none;
}
```

scss/_grid.scss
ysds: I missed [the comment](https://github.com/twbs/bootstrap/pull/28517#discussion_r364956887). The `ul.row` raises the CSS priority, so for example overwriting its `padding-left` may cause confusion.

```html
<ul class="row custom-class">...</ul>
```
```scss
.custom-class {
   padding: 1rem; // left padding is not apply
}
```

It is a safe way for users to add `.list-unstyled` deliberately. Or, I think as in the first @MartijnCuppens 's suggestion, applying rest list style to `.row` without limiting to `ul` and `li` may not cause confusion of CSS priority. e.g.:

```css
.row {
  ...
  padding: 0;
  list-style: none;
}
```

ysds:
@MartijnCuppens #30223 may be resolved by this new grid system. But, please do not include any additional changes in this PR except for fixes of review comments. This PR is already very huge, my hands are full😇

site/content/docs/4.3/layout/grid.md
```diff
@@ -365,6 +364,160 @@ You can also use the accompanying Sass mixin, `row-cols()`:
 }
 {{< /highlight >}}
 
+## Gutters
+
+Gutters can be responsively adjusted by breakpoint-specific gutter classes in as well horizontal, vertical and both directions. By default, `.row`s have a horizontal gutter of `1.5rem`. Removing this default gutter is possible by adding the `.g-0` class.
```
ysds: > Removing this default gutter is possible by adding the `.g-0` class.

Very minor thing, but this description is explained in detail below, so it feels redundant here.

site/content/docs/4.3/layout/grid.md
```diff
@@ -365,6 +364,160 @@ You can also use the accompanying Sass mixin, `row-cols()`:
...
+);
+```
+
+### Horizontal gutters
+
+`.gx-*` classes can be used to control the horizontal gutter widths. The `.container` or `.container-fluid` parent may need to be adjusted if larger gutters are used too to avoid unwanted overflow, using a matching padding utility. For example, in the following example we've increased the padding with `.px-lg-4` at the large (`lg`) breakpoint and above:
+
+{{< example >}}
+<div class="container px-lg-4">
+  <div class="row gx-lg-5">
```
ysds: Not limited to this line, some examples in the Gutters section use `lg` breakpoints, making it difficult to understand on mobile devices. Is it possible to rewrite for mobile friendly examples?

site/content/docs/4.3/layout/grid.md
```diff
@@ -365,6 +364,160 @@ You can also use the accompanying Sass mixin, `row-cols()`:
...
+The gutters between columns in our predefined grid classes can be removed with `.g-0`. This removes the negative `margin`s from `.row` and the horizontal `padding` from all immediate children columns.
+
+**Need an edge-to-edge design?** Drop the parent `.container` or `.container-fluid`.
+
+In practice, here's how it looks. Note you can continue to use this with all other predefined grid classes (including column widths, responsive tiers, reorders, and more).
+
+{{< example class="bd-example-row" >}}
+<div class="row g-0">
+  <div class="col-sm-6 col-md-8">.col-sm-6 .col-md-8</div>
+  <div class="col-6 col-md-4">.col-6 .col-md-4</div>
```
ysds: It would be good to not use `.bd-example-row` here for consistent colors.

ysds:
@twbs/css-review I have read through all the changes! Please see https://github.com/twbs/bootstrap/pull/28517#issuecomment-586609234

scss/_variables.scss
MartijnCuppens: Ok, I've now changed it to `1rem`, which looks more or less the same as it used to look like.

scss/_grid.scss
mdo: Good call, I think it'd be better to ask folks to add the `.list-unstyled` to make this work.

mdo: Good call, I think it'd be better to ask folks to add the `.list-unstyled` to make this work.

site/content/docs/4.3/layout/grid.md
MartijnCuppens: Synced the `.bd-example-row` with utility colors, to make them look alike.

site/content/docs/4.3/layout/grid.md
MartijnCuppens: Good point, removed the breakpoint specific gutters

site/content/docs/4.3/layout/grid.md
MartijnCuppens: Changed the copy below

scss/_grid.scss
MartijnCuppens: Indeed, removed it.

site/content/docs/4.3/forms/validation.md
MartijnCuppens: Reverted this

ysds (approved):
@MartijnCuppens I revisited only the parts that I commented on in the previous review, but I haven't see other lines. Rebase does not leave the differences between before and after the review as independent commits, so need to recheck the all differences in the files. This is not a problem for files with few changes, but it is a big task to see all the files that have changed a lot. Keeping the commit log clean is important, but it is helpful to choose rebase or not depending on the situation.

ysds:
@MartijnCuppens, @XhmikosR Can we merge this? Are there any remaining tasks?

XhmikosR:
This is quite a change and personally I haven't reviewed it. I'd say wait until we get an approval from @mdo before moving forward with this.

MartijnCuppens:
We might consider a switch to CSS grid now that we've descided to drop IE11 support. Not sure how to provide a solution for `col-auto` and `col` though.

Anyway, merging this would be a good base to build upon later on.

mdo (commented):
Couple quick suggestions here after a final review.

MartijnCuppens:
:shipit:

mdo:
🙌🏻👏🏻🎉

michaelrobertson-fico-com:
We're all looking forward to seeing IE11 disappear over the horizon. I'm repeating myself but I love the flexbox utility classes, and wish we had the same utilities available for CSS Grid. We're open to an asbstraction of CSS Grid but not sure we'd implement another approach to grids, even as an interim step. Our new design system is heavily invested in web standards, semantic markup and accessibility, and we have application developers already creating PoCs with CSS Grid. Appreciate everyone's work on making BS better.

scss/mixins/_grid.scss
```diff
@@ -10,27 +10,26 @@
...
 
 @mixin make-col-ready($gutter: $grid-gutter-width) {
-  position: relative;
+  // Add box sizing if only the grid is loaded
+  box-sizing: if(variable-exists(include-column-box-sizing) and $include-column-box-sizing, border-box, null);
   // Prevent columns from becoming too narrow when at smaller grid tiers by
-  // always setting `width: 100%;`. This works because we use `flex` values
+  // always setting `width: 100%;`. This works because we set the width
   // later on to override this initial width.
+  flex-shrink: 0;
```
chalin, line 18: Hi all: while I understand the rational for dropping `position: relative`, can you comment on why you added `flex-shrink: 0`? If would be great if you could respond to the following discussion thread. Thanks!

- https://github.com/twbs/bootstrap/discussions/37951

PR #28519, opened 2019-03-17
Update .gitignore

XhmikosR:
Add back ruby/bundler folders so that they don't show up when we switch branches.

This can be removed when v4 hits EOL, along with the old Jekyll docs dist folder.


Johann-S (approved):

ahtee:
Is the bundled ignore going to remove the Jekyll site?

PR #28521, opened 2019-03-17
added break-all

Ken-vdE:
The helper class to set word-break: break-all was not yet included. There are cases where break-all is a better match than break-word.


XhmikosR:
I don't know if this will be added, but please always make sure tests pass.

XhmikosR:
@twbs/css-review what do you want to do with this?

XhmikosR:
Closing in favor of #28758

PR #28535, opened 2019-03-21
WIP: Add variable to documentation pages

jhechtf:
Document the variables that each component uses or otherwise declares in order to make theme-ing a bit more streamlined.

Similar to how Bulma's documentation lists the applicable variables [Example of Bulma's Navbar component](https://bulma.io/documentation/components/navbar/#variables)


XhmikosR:
Hmm, so many conflicts.

jhechtf:
Yeah, that part I don't understand. I merged in the V4 master from the main
repo before pushing the code up. So it should all match.

XhmikosR:
You are not supposed to do that. v4-dev is never merged on master, only the other way around.

You need to do `git rebase -i upstream/master` and drop all unrelated patches.

jhechtf:
My mistake. I had a vage memory of V4-Dev being the base for V4 related
development. I'll fix it in a second.

PR #28543, opened 2019-03-22
Move Manipulator test cases from QUnit to Jasmine

bershanskiy:
@Johann-S Here is a small commit that rewrites `Manipulator` test cases for Jasmine. This is not for merging but instead for getting feedback on [my commit](https://github.com/twbs/bootstrap/commit/532c2935372a5a95aaeb49adab5c8635e70a2512).

A random question: `toggleClass` uses `classList.remove`, `classList.add`, `classList.contains`, which according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) have worse support than `classList.toggle`. Is MDN wrong or does compatibility table not tell the whole story?

Edit: I resolved merge conflicts for simplicity of review.


js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+    it('should get all data attributes', () => {
+      fixtureEl.innerHTML = '<div data-test="js" data-test2="js2" ></div>'
+      const div = fixtureEl.querySelector('div')
+      expect(Manipulator.getDataAttributes(div)).toEqual({
+        test: 'js',
+        test2: 'js2'
+      })
+    })
+
+/* TODO: this test appears in the original test suite. What is it doing?
```
Johann-S: this part can be removed it was at the beginning when IE 10 was supported

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+
+    it('should throw error if element is undefined', () => {
+      const incorrect = function() {
+        removeDataAttribute(undefined, 'key')
+      }
+      expect(incorrect).toThrow();
+    })
+
+    it('should remove data attribute', () => {
+      fixtureEl.innerHTML = '<div data-key="value" ></div>'
```
Johann-S: trailing space after data-key="value"

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+      expect(position.left).toEqual(jasmine.any(Number))
+    })
+  })
+
+  describe('toggleClass', () => {
+    it('should not error out if element is null', () => {
+      Manipulator.toggleClass(null, 'test')
+    })
+
+    it('should not error out if element is undefined', () => {
```
Johann-S: you can merge `should not error out if element is null` with `should not error out if element is undefined`

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+      Manipulator.toggleClass(undefined, 'test')
+    })
+
+    it('should not error out if class is null', () => {
```
Johann-S: not needed

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+      fixtureEl.innerHTML = '<div></div>'
+      const div = fixtureEl.querySelector('div')
+      Manipulator.toggleClass(div, null)
+    })
+
+    it('should not error out nothing if class is undefined', () => {
```
Johann-S: not needed

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+    })
+
+    it('should remove class if it is set', () => {
+      fixtureEl.innerHTML = '<div class="test"></div>'
+      const div = fixtureEl.querySelector('div')
+      Manipulator.toggleClass(div, 'test')
+      expect(div.classList.contains('test')).toEqual(false)
+    })
+
+    it('should remove class if it is set without disturbing other classes', () => {
```
Johann-S: not needed, browser behavior

js/src/dom/manipulator.spec.js
```diff
@@ -0,0 +1,219 @@
...
+      fixtureEl.innerHTML = '<div class="test stay"></div>'
+      const div = fixtureEl.querySelector('div')
+      Manipulator.toggleClass(div, 'test')
+      expect(div.classList.contains('test')).toEqual(false)
+      expect(div.classList.contains('stay')).toEqual(true)
+    })
+
+    it('should add class if it is set without disturbing other classes', () => {
```
Johann-S: not needed, browser behavior

Johann-S (changes requested):
Thanks for your PR 👍 

It's pretty good, there is some little things to fix and it'll be fine

Johann-S:
About the support of `classList.add`, `classList.contains` and `classList.remove` it's in the basic support to it's supported everywhere

js/src/dom/manipulator.spec.js
bershanskiy: I'll remove it. (I added it in the first place out of extra caution because there I saw weird behavior somewhere that didn't like `<div [something]/>` but worked with `<div [something] />`.)

Johann-S (approved):

Johann-S:
Thank you very much @bershanskiy for your work 👍 much appreciate

PR #28549, opened 2019-03-23
[WIP] Adding in variables used in each component.

jhechtf:
The goal of this merge request is to add in the variable names used by each component to give people an easier to access way to theme bootstrap. This is a work in progress, as each component involves me reading through the code to get an appropriate determination of what they do.

| Status | Component | Comments | 
| --- | --- | ---|
| ✅ | alerts | |
| ✅ | badge | |
| ✅ | breadcrumb | |
| ✅ | button-group | |
| ✅ | buttons | |
| ✅ | card | |
| ✅ | carousel | |
| ✖ | collapse | |
| ✖ | dropdowns | |
| ✖ | forms | |
| ✖ | input-group | |
| ✖ | jumbotron | |
| ✅ | list-group | |
| ✖ | media-object | |
| ✖ | modal | |
| ✖ | navbar | |
| ✖ | navs | |
| ✖ | pagination | |
| ✖ | popovers | |
| ✅ | progress | |
| ✖ | scrollspy | |
| ✅ | spinners | |
| ✖ | toasts | |
| ✖ | tooltips | |
| ✅ | code | |
| ✅ | figures | |
| ✅ | close-icon | |

### DIFF
--- a/site/docs/4.3/components/alerts.md
+++ b/site/docs/4.3/components/alerts.md
@@ -133,3 +133,19 @@ myAlert.addEventListener('closed.bs.alert', function () {
   // do something…
 })
 {% endhighlight %}
+
+## Variables
+
+The alert component uses the following variables to help define its look and feel.
+
+| Variable | Default | Purpose |
+| --- | --- | --- |
+| $alert-padding-y | 0.75rem | Determines padding across the y-axis |
+| $alert-padding-x | 1.25rem | Determines padding across the x-axis |
+| $alert-margin-bottom | 1rem | Determines the size of the margin
+| $alert-border-radius | $border-radius(`0.25rem !default`) | Determines the size of the border radius for the alert element. |
+| $alert-link-font-weight: | $font-weight-bold (`700 !default`) | Determines the font weight for links in the alert
+| $alert-border-width: | $border-width(`1px !default`) | Determines the thickness of the borders around the alerts
+| $alert-bg-level: | -10 | Determines amount of mixing with the color variants for the background. <br> *Note* Negative numbers will increase baby the lightness of the color, positive will increase the darkness of the color |
+| $alert-border-level: | -9 | Determines amount of mixing with color variants for the border.  <br> *Note* Negative numbers will increase baby the lightness of the color, positive will increase the darkness of the color |
+| $alert-color-level: | 6 | Determines amount of mixing for the font color. <br> *Note* Negative numbers will increase baby the lightness of the color, positive will increase the darkness of the color.  <br> *Note* Negative numbers will increase baby the lightness of the color, positive will increase the darkness of the color |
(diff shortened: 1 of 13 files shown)


XhmikosR:
@jhechtf: this is wrong again... You need to rebase against upstream/master.

XhmikosR:
Also, to be honest, I'm not sure I like the maintenance weight this adds.

jhechtf:
@XhmikosR 
In response to the maintenance weight, it really wouldn't add much so long as people follow the same pattern that I am currently seeing in the `_variables.scss` file, which is where I'm pulling the values from. It would just have to be an expectation for anyone in the future who introduces a new variable to make sure that they document and describe its use, which I don't think is unreasonable, especially considering that Bootstrap is as much a stand-alone CSS/Javascript framework as it is a themeable base.

As for the rebase -- can you please do me a favor and clarify what you mean by rebasing against upstream/master? in the last ticket you told me to remove anything that was unrelated. I took that to mean "remove anything that isn't from my changes." Assuming I rebase from upstream master, what are types of changes are ones that I should keep, and what are ones I should remove? I'm asking for clarity as I'd really rather like to only do this once more before going back to documenting the other components.

XhmikosR:
But we'd need to update those manually, which is an extra things for us.

Sure, I can automate it probably with Hugo's `readFile` and some filtering, but I'm still not sure this is worth it.

/CC @twbs/css-review

jhechtf:
I really don't think "if you add / edit / remove a variable from the `_variables.scss` file, you need to make sure to add/edit/remove the entry in the corresponding Variables table to reflect the changes" is asking too much, especially since I am going to be giving the brunt of most of that work, unless the bootstrap variables change very suddenly in 4.4 or beyond. Even then I'd be willing to go through this all again to make sure that the tables are updated for future use, because documenting how these values interconnect is important.

XhmikosR:
It is a lot of work unless it's automated, and I speak from experience, when I say that anything not automated will be a problem.

Leaving this to @twbs/css-review . I'd personally wouldn't add this, unless it's properly automated.

MartijnCuppens:
I like the idea of additional variables documentation, but this approach will be hard to maintain. We'll probably get out of sync at some point, because it 'll be easy to miss a variable because they can be listed anywhere in the docs. I'm more in favour in something that's automated (and simple). That 'll be quite a challenge I guess.

jhechtf:
I'm all ears if anyone has another way to build the documentation.

MartijnCuppens:
@JLammeer once did a pretty nice job here: https://jlammeer.github.io/bootstrap-sassdoc/

But that lead to a pretty huge variables file which was mad it a bit more unclear on the `scss` side because of the amount of comments:
https://github.com/JLammeer/bootstrap-sassdoc/blob/master/documented-items/_variables.scss

We don't have an out of the box solution yet, it looks like a lot of people are looking for this or want to create a solution for Bootstrap. A simple solution we can plug into Hugo would be ideal I guess.

jhechtf:
Let me look into Hugo (never used before) and see if I can find something.

XhmikosR:
It's doable with `readFile` like I mentioned. Then we need to find a way to filter only the related variables, which can be done with a specific comment in _variables.scss, like `// ** Tooltips` or similar.

MartijnCuppens:
Thanks @jhechtf, you can ping us when you got something. I'm going to close this PR for the maintenance issues we mentioned above.

PR #28561, opened 2019-03-27
added more potential sizes for css

GiantRobato:
Right now people can use `class="w-25"` or `class="w-50"`. Adding all 1-100% increments is overkill but I think that a lot of people would at least expect every 10% increment be defined (I sure would find it handy!).

# Adds additional classes to be generated

- `w-10`, `h-10`
- `w-20`, `h-20`
- `w-30`, `h-30`
- `w-40`, `h-40`
- `w-60`, `h-60`
- `w-70`, `h-70`
- `w-80`, `h-80`
- `w-90`, `h-90`


XhmikosR:
Deferring this to @twbs/css-review, but if it's gonna be merged the docs need to be updated to reflect the changes.

penguindustin:
Any ETA or decision on this? I can update the docs to reflect this as well if that's what's blocking.

MartijnCuppens:
The height utilities will probably barely be used. We are also considering using the `.col-*` classes as width utilities.

PR #28565, opened 2019-03-27
[WIP] Refactor and simplify button plugin

### LINKED ISSUE #26855: Radio buttons: .btn-group-toggle: Keyboard selection only works visually.
In a toggle button group, if you select a button using the keyboard, the underlying selection of the radio inputs is not changed. Thus, if you have these buttons in a form and submit the form, the user might submit an unintended selection.

I guess this is related to #18874. I added a comment there, but there was no reaction.
#25122 and #26778 are possibly related but no duplicates imo.

### Steps to reproduce:
- In a toggle button group, click on one button with the mouse
- Then, change your selection with the arrow keys (left / right)

### Expected Result:
The radio input that represents the now selected button is checked

### Actual Result:
The radio input selection is not changed although the visuals indicate that it was changed. (tested in firefox 60.0.2, firefox 61.0.1 and chrome 67.0.3396.99)

### Example Code
[You can reproduce the issue here](https://jsbin.com/xatikicere/edit?html,css,output). The html code is copied from [the docs](https://getbootstrap.com/docs/4.1/components/buttons/#checkbox-and-radio-buttons). I only added css to remove the clip property on the inputs so you can see which input is selected.

### LINKED ISSUE #28564: JS - button plugin stop event propagation on checkboxes/radio when browsing with keyboard
After looking at the previous issue https://github.com/twbs/bootstrap/issues/26904

And play with the associated codepen : https://codepen.io/Johann-S/pen/GBqoZG?editors=1111

I could still reproduce the use-case in current documentation on 3 way radio toggle buttons : 
https://getbootstrap.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons

1 - browse the input element with keyboard
2 - change the selected item using keyboard arrow
3 - look in the console document.getElementById('option2').checked or other optionid the current value isn't the visual one

Works well when using mouse

My first tests : 
- $(input).trigger('change') is well executed with keyboard or mouse but maybe there's sub-jascent events to propagate?
- by removing the `event.preventDefault()` from plugin initialisation state change well with keyboard

```
$(document)
  .on(Event.CLICK_DATA_API, Selector.DATA_TOGGLE_CARROT, (event) => {
    // event.preventDefault()
```

First tests doesn't show anything is broken by removing this

image

patrickhlauke:
Fixes persistent keyboard issues with checkbox and radio button "buttons" https://getbootstrap.com/docs/4.3/components/buttons/#button-plugin

Instead of setting the "checked" state on click (which doesn't happen when navigating with the keyboard), this relies on checkboxes and radio
buttons just doing their normal thing, and then simply sets the relevant `.active` class.

Similarly, it simplifies the handling/differentiation for toggles using `aria-pressed`

Closes https://github.com/twbs/bootstrap/issues/28564
Closes https://github.com/twbs/bootstrap/issues/26855

(pro-tip to test if keyboard is being handled correctly: for the checkbox and radio button buttons, use devtools, find the nested `input`, and disabled the `position:absolute` rule in the relevant CSS ... now you can see exactly if a radio or checkbox gets toggled properly or not)


patrickhlauke:
with this change, things *seem* to work ok based purely on testing the "Toggle states" and "Checkboxes and radio buttons" examples from https://getbootstrap.com/docs/4.3/components/buttons/

are there other ways to use the button plugin? if so, not tested those...

patrickhlauke:
worth adding here that this refactor assumes that checkbox/radio button buttons will use the correct `<label><input></label>` construct. it won't work if authors use some other generic container, like `<div><input></div>` or similar (as in those cases, a click on the outer button won't result in the input being checked/unchecked, and THAT would require some of the code that was stripped out that programmatically sets the checked value).

is the latter scenario one we support (but not encourage)?

Johann-S:
Hi @patrickhlauke,

it seems your change works fine 👍 

About your last question I cannot answer you, but we are at the time we can break things, so it's the right moment to make clear decision.

You have some lint issues, and I think you'll have to fix unit tests too

patrickhlauke:
well ideally i'd like this fixed in v4, not just for v5, as it's currently fundamentally broken and inaccessible otherwise.

and for v5, i thought we were going to move completely away from this anyway (for checkboxes/radio buttons) and just go for pure CSS implementation. it would certainly save a lot of this hassle and be much more robust compared to this brittle house of cards. (e.g. something more like http://adrianroselli.com/2019/03/under-engineered-toggles.html)

patrickhlauke:
> You have some lint issues, and I think you'll have to fix unit tests too

when i was building locally, nothing came up as a problem though...?

patrickhlauke:
xref for v5 https://github.com/twbs/bootstrap/pull/28463

XhmikosR:
Probably ancient local npm modules.

patrickhlauke:
well, if anybody can nudge me towards how to get travis not to bork here, and how to see these linting/test issues, that'd be super...

and comment on whether the approach here (stripping this out to bare minimum, and relying on checkbox/radio to use just the `<label><input></label>` pattern, where then clicking on `<label>` handles setting the checked state of the `<input>` automatically already) that'd be super too.

Johann-S:
calling `npm ci` should fix your issue, I pushed a patch to fix the linting issue

Lausselloic:
When you say "simplify" it's not a little :-D

I've give a try to your branch and doesn't fin another case with that plugin. Just maye add a `disabled` item to make show that it also works with it : 
image


I also play without the plugin : https://codepen.io/lausselloic/pen/bZXpJZ a full css solution based on input state and sibling label but maybe keep it for V5

patrickhlauke:
for info @Lausselloic the pure CSS approach for v5 is already being tackled here https://github.com/twbs/bootstrap/pull/28463

Lausselloic:
I haven't Seen that before sorry

Lausselloic:
I've try to help on fixing unit test on the V4 branch but I've create a new PR :-(

https://github.com/twbs/bootstrap/pull/28587

If it can help feel free to pick-up the code and drop the PR

patrickhlauke:
@mdo so the question is: do we want to keep supporting the case where `<button>` elements behave like checkboxes/radio buttons (we seem to have no examples per se in the docs, just a mention in https://getbootstrap.com/docs/4.3/components/button-group/#basic-example and a related unit test which this PR fails (`button should toggle aria-pressed on buttons with container`). i guess if we stopped supporting this, it would be a breaking change, so no chance for a 4.x release?

if we do want to keep supporting this case, the functionality will need to be put back into this PR's javascript...

fixed the other `button should check for closest matching toggle` unit test failure ... turns out the unit test was flaky

dvmuccillo:
I believe the examples for radios and checkbox are here: https://getbootstrap.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons

patrickhlauke:
@dvmuccillo yes but those don't have an example of using regular `<button>` elements inside a toggle group and setting `aria-pressed` on them...

dvmuccillo:
I see,  sorry i always confuse those

Em qua, 3 de abr de 2019 05:19, Patrick H. Lauke <notifications@github.com>
escreveu:

patrickhlauke:
Closing this in favor of a more focused, v4-specific fix (since in v5 we'll drop the JS-reliant checkbox/radio button groups and use a pure CSS approach instead)

patrickhlauke:
Successor https://github.com/twbs/bootstrap/pull/28834

PR #28570, opened 2019-03-28
Update devDependencies.

XhmikosR:


js/src/tooltip.js
```diff
@@ -462,7 +462,7 @@ class Tooltip {
           element.appendChild(content)
         }
       } else {
-        element.innerText = content.textContent
+        element.textContent = content.textContent
```
XhmikosR: This we should backport in v4 if we agree on it.

Johann-S: In v4 we use `.text` of jQuery which is pretty much the same

XhmikosR: True that.

Johann-S:
Maybe we should update the Nuget files too ?

XhmikosR:
Yup, I forgot about those. I'll take care of it later.

Johann-S (approved):

PR #28587, opened 2019-04-01
fix(buttons): fix button unit test onclic input event

Lausselloic:
fix one unit test down due to onclic event

Don't remove the latest unit test failing : 
```
  QUnit.test('should toggle aria-pressed on buttons with container', function (assert) {
    assert.expect(1)
    var groupHTML = '<div class="btn-group" data-toggle="buttons">' +
        '<button id="btn1" class="btn btn-secondary" type="button">One</button>' +
        '<button class="btn btn-secondary" type="button">Two</button>' +
      '</div>'
    $('#qunit-fixture').append(groupHTML)
    $('#btn1').bootstrapButton('toggle')
    assert.strictEqual($('#btn1').attr('aria-pressed'), 'true')
  })
```

Not sure if that structure need to be available? If yes need to fix the plugin, if no need to remove the test

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -68,21 +68,21 @@ class Button {
       Selector.DATA_TOGGLE
     )
 
+    const input = SelectorEngine.findOne(Selector.INPUT, this._element)
+
     if (rootElement) {
       const activeElement = SelectorEngine.findOne(Selector.ACTIVE, rootElement)
 
       if (activeElement) {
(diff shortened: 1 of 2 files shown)


patrickhlauke:
sorry, just been dipping into my own PR not really checking other stuff...and ended up fixing this now already (using a slightly different approach). so, this is probably superceded by https://github.com/twbs/bootstrap/pull/28565 now

PR #28589, opened 2019-04-01
v4-dev backports and dep updates

XhmikosR:


XhmikosR:
I'm think I'm gonna merge this as is and we do the jQuery update when they release v3.4.1.

Johann-S (approved):

PR #28613, opened 2019-04-05
Added 'xxl' and 'xxxl' columns prefix breakpoints for ultrawide screens

ghost:
I works with ultrawide screen and I need two new prefix breakpoints for columns.

Example:
```
<div class="col-xxl-2">Your amazing content</div>
<div class="col-xxxl-3">Your amazing content</div>
```
The breakpoints:
**xxl** min-width 1700px
**xxxl** min-width 2180px

The bootstrap cards now fell good in ultrawide screen :)


MartijnCuppens:
We'll pass on this (see https://github.com/twbs/bootstrap/issues/20593#issuecomment-243200953)

PR #28621, opened 2019-04-08
Change toast border radius to use $border-radius

tkesgar:
Currently, $toast-border-radius uses the value of .25rem. This is inconsistent with other variables such as $btn-border-radius and $input-border-radius, which uses the "global" $border-radius variable.


MartijnCuppens (approved):
Thanks @tkesgar!

PR #28630, opened 2019-04-09
Call toast hide() after `delay` time

### LINKED ISSUE #28629: Toast hide event fired timing
When use an autohiding toast, the hide event is fired just after shown, it ignore the delay time.

demo: https://codepen.io/anon/pen/YMpRyE?editors=1111 (Please see the console view)

I expected the hide callback is called after 3000ms, but it is actually called right after the toast is shown. Is it intended? Thanks.

ysds:
The hide callback function should be called when start closing the toast.

Fixes #28629

### DIFF
--- a/js/src/toast.js
+++ b/js/src/toast.js
@@ -127,7 +127,21 @@ class Toast {
     }
 
     EventHandler.trigger(this._element, Event.HIDE)
-    this._close()
+
+    const complete = () => {
+      this._element.classList.add(ClassName.HIDE)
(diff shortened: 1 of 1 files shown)


Johann-S (changes requested):
After looking the change, seems a right move to me 👍 

Just remove the `_close` methods and move everything in `hide`

ysds:
@Johann-S Thank you for your review 😄

> Just remove the _close methods a move everything in hide

Done!

Johann-S (approved):

PR #28635, opened 2019-04-09
Fix floating labels overlay with chrome auto fill

e3duardo:
I had this error with the example of floating labels today, and looking for some solutions I found the fix I'm committing.
I tested in latest chrome for latest mac, I do not know about other browsers, but I would like to help.

Captura de Tela 2019-04-09 às 20 34 50


MartijnCuppens (changes requested):
This will invalidate the selector for non-webkit browsers, see https://deploy-preview-28635--twbs-bootstrap.netlify.com/docs/4.3/examples/floating-labels/ in for example Firefox to reproduce this issue.

e3duardo:
> This will invalidate the selector for non-webkit browsers, see https://deploy-preview-28635--twbs-bootstrap.netlify.com/docs/4.3/examples/floating-labels/ in for example Firefox to reproduce this issue.

I saw it here, that bad! I really wanted this functionality in my project, even with fields filled in automatically, do you think this is only possible with javascript?

XhmikosR:
@MartijnCuppens if you think this is a no go let's close it.

MartijnCuppens:
Should be possible by splitting the selectors & duplicating the properties. @e3duardo, could you have a look at that?

MartijnCuppens:
Just tested this, this doesn't seem to be a problem in Chrome (not on mac or Windows). If this still is a problem please reopen and provide additional testing info.

PR #28637, opened 2019-04-10
Fix broken link to css-discuss.incutio.com/wiki/Printing_Tables

coliff:
Instead now pointing to the most recent Wayback Machine archive copy.


MartijnCuppens:
We're probably going to remove a lot of the print styles in `v5`, so we might better target the `v4-dev` branch here (ping @XhmikosR)?

Other than that, linking to Wayback Machine is creative solution, but we might just better remove it or link it to another article instead of linking to something that once existed on the internet 😄

XhmikosR:
Yeah, I remember this one. I'd swear I had removed it at some point, but apparently not :P

I think this is fine targeting master, since any removal will happen later.

About removing or replacing the link I think it's better to link to the original page for historical reasons.

XhmikosR (approved):

PR #28642, opened 2019-04-11
remove custom browserlist for our esm build

Johann-S:
According to those articles:
 - https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features
 - https://webpack.js.org/guides/author-libraries/#final-steps

the `module` field in our package.json is used by bundlers too, so we shouldn't have a different browserlist for our ESM build, because our users won't be able to use it with our older browsers

### DIFF
--- a/build/rollup.config.js
+++ b/build/rollup.config.js
@@ -11,42 +11,18 @@ const ESM = process.env.ESM === 'true'
 let fileDest = `bootstrap${ESM ? '.esm' : ''}`
 const external = ['popper.js']
 const plugins = [
-  babel(ESM ?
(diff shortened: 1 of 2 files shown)


XhmikosR:
But then *our* ESM build is not what we want it to be, i.e. no `var` etc.

Johann-S:
That's not what I wanted first, I wanted a better compatibility with current bundlers and allow Bootstrap ESM build to be used in a browser

XhmikosR (approved):

PR #28649, opened 2019-04-13
Remove autocomplete from buttons and radio/checkbox inputs

### LINKED ISSUE #28623: put back autocomplete="off" warning and add it on all checkboxes and radio?
I haven't find it in v4 documentation and all input checkboxes and radio doen't have the autocomplete attribute but as mentioned in v3 :
https://getbootstrap.com/docs/3.4/javascript/#callout-buttons-ff-autocomplete firefox persist state during browsing https://bugzilla.mozilla.org/show_bug.cgi?id=654072

So maybe need to add autocomplete="off" on all radio and checkboxes in V4/V5 documentation + put back the callout?

patrickhlauke:
While the bug in Firefox is still present, apparently (see https://bugzilla.mozilla.org/show_bug.cgi?id=654072) we should remove this (as otherwise we'd need explanations everywhere on why this Firefox-specific workaround has been added everywhere).

Fixes #28623

/cc @miketaylr

### DIFF
--- a/build/vnu-jar.js
+++ b/build/vnu-jar.js
@@ -23,10 +23,6 @@ childProcess.exec('java -version', (error, stdout, stderr) => {
   // vnu-jar accepts multiple ignores joined with a `|`.
   // Also note that the ignores are regular expressions.
   const ignores = [
-    // "autocomplete" is included in <button> and checkboxes and radio <input>s due to
-    // Firefox's non-standard autocomplete behavior - see https://bugzilla.mozilla.org/show_bug.cgi?id=654072
-    'Attribute “autocomplete” is only allowed when the input type is.*',
-    'Attribute “autocomplete” not allowed on element “button” at this point.',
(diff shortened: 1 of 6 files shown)


patrickhlauke:
However, note https://github.com/twbs/bootstrap/issues/28623#issuecomment-482807772 ... I've not tested, but is there potential for the autocomplete behaviour to get our styles/behaviours out of sync? If that's the case, I'd tend to say we should keep autocomplete until the bug has been fixed/addressed, as otherwise it'll lead to weird and difficult to debug issues for Firefox users... /cc @mdo

XhmikosR:
Rebased. @mdo ping for your opinion.

patrickhlauke:
Closing this, as I'm actually going to tackle the root cause of this (visual presentation of buttons/toggles not matching their real value on page load)

patrickhlauke:
Successor (which fundamentally tackles the problem) https://github.com/twbs/bootstrap/pull/28952

PR #28660, opened 2019-04-15
docs SVGs: Allow passing a custom title.

XhmikosR:


patrickhlauke:
Noting that as they're `aria-hidden`, even the current title won't  be announced, as we generally set the label/name on whatever uses those icons

XhmikosR:
Well, the thing is that only 3 of them have `aria-hidden=true` and secondly, the title is useful for hovering.

See https://deploy-preview-28449--twbs-bootstrap.netlify.com/ after hovering on the Toggle Dark Mode navbar button

Not sure if we should have aria-hidden on all of them, I leave this stuff to you :)

XhmikosR:
Ping @MartijnCuppens @Johann-S

Johann-S (approved):

PR #28681, opened 2019-04-23
Separate btn-link colors from link

GeoSot:
I would like to be able to separate link colors from btn-link colors, so to be able to choose different values for each

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -76,11 +76,11 @@ fieldset:disabled a.btn {
 // Make a button look and behave like a link
 .btn-link {
   font-weight: $font-weight-normal;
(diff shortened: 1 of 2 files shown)


MartijnCuppens (approved):
LGTM

GeoSot:
@MartijnCuppens  can we pass this change on v4 too?

MartijnCuppens:
Hi @GeoSot,

We're blocking all new features for `v4` and will add them in `v5`.

tomsommer:
Why not `text-decoration` as well then?

MartijnCuppens:
@tomsommer, I don't mind adding that as well, feel free to open a PR.

PR #28682, opened 2019-04-23
Modal full screen

GeoSot:
Add Full-Screen size to Modal and Responsive variations for breakpoints

Demo: https://deploy-preview-28682--twbs-bootstrap.netlify.com/docs/4.3/components/modal/#optional-sizes

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -76,11 +76,11 @@ fieldset:disabled a.btn {
 // Make a button look and behave like a link
 .btn-link {
   font-weight: $font-weight-normal;
-  color: $link-color;
+  color: $btn-link-color;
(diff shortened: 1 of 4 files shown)


MartijnCuppens:
Hi @GeoSot, 
This doesn't seem to work (See link to preview). Also make sure to remove the border radius and border from fullscreen modals.

I'm also not sure if we should call this a modal.

patrickhlauke:
> I'm also not sure if we should call this a modal.

if it covers the whole screen, but is conceptually "over" the existing page, it's a-ok to call this a modal in my view.

GeoSot:
@MartijnCuppens I forgot to to build the css

Issue #28513, opened 2019-03-15
[bug] Dropdown inside scrollable body modal clipped and/or inaccessible


tmorehouse:
Version 4.3.1

When one has a scrollable body modal, with a dropdown inside, and the model content is short, the opened dropdown gets cut off/hidden, and in some cases on desktop, impossible to scroll to select the last dropdown menu item.  Changing the boundary option of dropdown does not correct the issue.

![image](https://user-images.githubusercontent.com/2781561/54464010-c4230d80-4753-11e9-9163-b65996bc308d.png)

Reproduction: https://codepen.io/anon/pen/XGEwPV?editors=1000

After playing a bit, I was able to come up with a workaround:

- add class `.position-static` to the `.modal-body`
- override the style `overflow: hidden;` on the `.modal-content` div with `overflow: initial;` (as the SCSS defines it as `overflow: hidden;` for some reason in scrollable mode).
- add `data-boundary="viewport"` (or  window) to the dropdown button

Working example: https://codepen.io/anon/pen/rRdggJ?editors=1000

Possible fix:
When the modal is in scrollable body mode, add the above `position: static` to `.modal-body` rule, and remove `overflow: hidden;` on `.modal-content` (which appears to be added only in scrollable mode)

Then it would be up to the user to apply the `boundary` to the dropdown if needed.

ysds:
The `.modal-content {overflow: hidden}` is to prevent overflow the content in a small height screen:

<img width="1280" alt="screenshot" src="https://user-images.githubusercontent.com/4065765/54474920-60b9de00-482e-11e9-8a57-3f2448979647.png">

But I agree the CSS cause that dropdown issue.

tmorehouse:
Good point...

Maybe for small (short) viewports, the height of the scrollable modal could be replaced with a non-vh absolute value (or removed), and add a max height on the modal-body?

mdo:
Is this still valid?

moosetraveller:
![image](https://user-images.githubusercontent.com/3456293/195154349-d0f14e67-4cd7-4c98-85b2-57e9a15c5894.png)

Bootstrap 5.0.2 and 5.2.2

GreathostRo:
After more than a few hours and various attempts I found the working version:

```
const dropdowns = document.querySelectorAll('.dropdown-toggle')
const dropdown = [...dropdowns].map((dropdownToggleEl) => new bootstrap.Dropdown(dropdownToggleEl, {
[...]

PR #28683, opened 2019-04-23
Add Full-Screen size to Modal and Responsive variations for breakpoints

GeoSot:
Add Full-Screen size to Modal and Responsive variations for breakpoints
Works from breakpoint and down

( @MartijnCuppens , please ignore my previous pull request)

Preview: <https://deploy-preview-28683--twbs-bootstrap.netlify.com/docs/4.3/components/modal/#fullscreen-modal>
Previous PR: #28682


MartijnCuppens:
Yu @GeoSot,

Nice to see you getting involved in Bootstrap! I've readded the netlify preview in your (you can get the link in the PR tests). Try not to reopen too many PR's for the same feature, we can miss some essential information sometimes. 

I noticed there's a spacing at the right side of the screen, could you fix that?

GeoSot:
@MartijnCuppens you are right about the two PR's. (I 've messed up the commits and wanted a clear solution)

As for the padding on the right, I saw that it is done from js. 
As a solution I can change max-width to min-width, in order to have a nice optical result, but is it acceptable?

MartijnCuppens:
Hmmm, I can't think of another better solution right now. Maybe @ysds knows a better way of handling this?

ysds:
I’ll take a look later :)

ysds:
As far as looked through that, I felt three things:

- The class name is a bit confusing. e.g. I misunderstood the `.modal-md-full` makes the modal to be full screen when the screen is wide.

- There is a display problem when the full screen modal has a lot of content.

- The modal's right padding is for centering the modal to the page content not the window (#14892). But the full screen modal will not require the padding. To remove the padding, should change the JS because it is added by JS.

GeoSot:
@MartijnCuppens  I have made some changes and fixes according to the previous suggestions. 

As for the className, i am open to  suggestions

site/content/docs/4.3/components/modal.md
```diff
@@ -659,6 +675,43 @@ Our default modal without modifier class constitutes the "medium" size modal.
...
+          <p>Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor.</p>
+          <p>Aenean lacinia bibendum nulla sed consectetur. Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Donec sed odio dui. Donec ullamcorper nulla non metus auctor fringilla.</p>
+          <p>Cras mattis consectetur purus sit amet fermentum. Cras justo odio, dapibus ac facilisis in, egestas eget quam. Morbi leo risus, porta ac consectetur ac, vestibulum at eros.</p>
+          <p>Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Vivamus sagittis lacus vel augue laoreet rutrum faucibus dolor auctor.</p>
+          <p>Aenean lacinia bibendum nulla sed consectetur. Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Donec sed odio dui. Donec ullamcorper nulla non metus auctor fringilla.</p>
+      </div>
+      <div class="modal-footer">
+              <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
```
XhmikosR: Please make sure you use 2-space indentation in all of your changes.

MartijnCuppens:
> As for the className, i am open to suggestions

`fullscreen` would make more sense I guess.

There's something else. We use a [mobile-first](https://getbootstrap.com/docs/4.3/layout/overview/#responsive-breakpoints) approach in Bootstrap, which uses `min-width` media queries instead of `max-width`. We also need extra documentation and examples to show the responsive effect.

GeoSot:
> There's something else. We use a [mobile-first](https://getbootstrap.com/docs/4.3/layout/overview/#responsive-breakpoints) approach in Bootstrap, which uses `min-width` media queries instead of `max-width`. We also need extra documentation and examples to show the responsive effect.


I have changed the class name. 

As for the mobile first approach thing, the purpose is to have a Full Screen modal on mobile , till the chosen breakpoint. So I am not sure if it needs any change

XhmikosR:
@GeoSot something seems broken. Also please squash your patches into one after rebasing against upstream.

GeoSot:
@XhmikosR  can I do something more in order to help this pull-request to proceed?

mdo:
Trying this out myself, I don't think the fullscreen modal should "drop in" like the other modals—gives it a weird looking animation. Thoughts on revisiting that part?

GeoSot:
Maybe if we omit the 'fade' class. Or try another transition???

MartijnCuppens:
I would keep the fade effect. Maybe we split the fade effect and the slide-in effect into 2 classes? Better cover this in another PR then because all modal classes.

GeoSot:
So, what you propose me to do?

MartijnCuppens:
> So, what you propose me to do?

Interesting question. I know @ysds is working on splitting the transition classes from the components in https://github.com/twbs/bootstrap/pull/29081.

@mdo, maybe we should postpone the `drop in` change until https://github.com/twbs/bootstrap/pull/29081 is finished?

mdo:
Sounds fine by me!

MartijnCuppens:
I'm still a bit concerned this implementation will be too confusing, other responsive components work the other way around:

- `.col-md-4`: 4 columns **above** `md` breakpoint
- `.navbar-expand-md`: expand **above** `md` breakpoint
- ...

This implementation:
- `.modal-md-fullscreen`: fullscreen modal **below** `md` breakpoint

Maybe we should have a look at responsive modal sizing in general first?

GeoSot:
> I'm still a bit concerned this implementation will be too confusing, other responsive components work the other way around:

I understand your concern and the logic of it.

The only thing I can say as simple user is, that a fullscreen modal seems more reasonable on small screen sizes than on wide screens. As for the naming conversion, if anyone want to suggest another approach I am open to think of it

scss/_modal.scss
```diff
@@ -236,3 +236,26 @@
...
+      margin: 0;
+
+      .modal-content {
+        height: 100%;
+        border: 0;
+        @include border-radius(0);
+      }
+
+      .modal-body {
+        overflow-y: auto;
```
MartijnCuppens: I think we can safely move this to the global `.modal-body`, would generate less code.

scss/_modal.scss
```diff
@@ -236,3 +236,26 @@
...
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    .modal#{$infix}-fullscreen {
+      width: 100vw;
+      max-width: inherit;
+      height: 100vh;
+      margin: 0;
+
+      .modal-content {
+        height: 100%;
```
MartijnCuppens: I think we can safely move the height to the global `.modal-content`, would generate less code.

scss/_modal.scss
```diff
@@ -236,3 +236,26 @@
...
   .modal-xl { max-width: $modal-xl; }
 }
+
+@each $breakpoint in map-keys($grid-breakpoints) {
+  @include media-breakpoint-down($breakpoint) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    .modal#{$infix}-fullscreen {
+      width: 100vw;
+      max-width: inherit;
```
MartijnCuppens: Why `inherit`?

GeoSot: Because .modal-dialog has a max-width:500px and has to be overwritten

MartijnCuppens: Use `none` then, that's the default value. `inherit` implies it **needs** to be inherited.

GeoSot: @MartijnCuppens  Done

Glenn-Domin:
I'd definitely be confused by `.modal-md-fullscreen` applying to `md` and down where everything else in Bootstrap is `md` and up. Not sure how practical this is but if you wanted to do a mobile-only implementation I'd suggest doing something like `.modal-fullscreen .modal-lg-boxed` to make it not full from `lg` upward (I don't like the name "boxed" but it's off the top of my head to demonstrate the point).

ysds:
The implementation with media-break-down would be an appropriate way here. But the issue is the class naming. I searched for past considerations about class naming of the similar case, found the [hidden-xx-down](https://v4-alpha.getbootstrap.com/layout/responsive-utilities/#available-classes) and the [btn-block-xx-down](https://github.com/twbs/bootstrap/pull/21720#issuecomment-285197633). I think such naming makes sense here too. So my suggestion is `.modal-fullscreen` (all viewports) and `.modal-fullscreen-md-down` ...

Regarding to the animation, implementation of #29081 still takes time. so how about leaving it as is for now for proceed this PR?

MartijnCuppens:
Brilliant plan @ysds. @GeoSot, could change the implementation to @ysds' suggestion?

site/content/docs/4.3/components/modal.md
```diff
@@ -772,6 +772,162 @@ Our default modal without modifier class constitutes the "medium" size modal.
...
+      <th>Availability</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>.modal-fullscreen</code></td>
+      <td>Always</td>
+    </tr>
+    <tr>
+      <td><code>.modal-xl-fullscreen-down</code></td>
```
nteetor: Chanage to `.modal-fullscreen-xl-down`.

site/content/docs/4.3/components/modal.md
```diff
@@ -772,6 +772,162 @@ Our default modal without modifier class constitutes the "medium" size modal.
...
+      <td><code>Below 1200px</code></td>
+    </tr>
+    <tr>
+      <td><code>.modal-lg-fullscreen-down</code></td>
```
nteetor: Change to `.modal-fullscreen-lg-down`.

site/content/docs/4.3/components/modal.md
```diff
@@ -772,6 +772,162 @@ Our default modal without modifier class constitutes the "medium" size modal.
...
+      <td><code>Below 992px</code></td>
+    </tr>
+    <tr>
+      <td><code>.modal-md-fullscreen-down</code></td>
```
nteetor: `.modal-fullscreen-md-down`

site/content/docs/4.3/components/modal.md
```diff
@@ -772,6 +772,162 @@ Our default modal without modifier class constitutes the "medium" size modal.
...
+      <td><code>Below 768px</code></td>
+    </tr>
+    <tr>
+      <td><code>.modal-sm-fullscreen-down</code></td>
```
nteetor: `.modal-fullscreen-sm-down`

nteetor:
I noticed the names were off in the list of classes under the Fullscreen section. I dropped in a few comments pointing out which classes needed to be corrected.

ysds:
Since there are still some issues, I will add the commit directly.

Adding `overflow-y: auto` to the global `.modal-body` will increase encountering #28513 (See [demo](https://codepen.io/fellows3/pen/xxGdwra)). Maybe should handle with JavaScript, but for workaround here, moving it to the each `.modal-body`.

andreivictor:
Just my humble opinion:

- looking at the generated css, we have:
`@media (max-width: 767.98px)
.modal-fullscreen-sm-down { ... }`

- but in the documentation it says:
`.modal-fullscreen-sm-down | Below 576px`

By using the same logic from https://v4-alpha.getbootstrap.com/layout/responsive-utilities/#available-classes, `modal-fullscreen-*-down` would mean that the modal is fullscreen at the given breakpoint or smaller.

ysds:
I added changes:
* Fix header and footer shrink issue
* Move `overflow-y: auto` from the global `.modal-body` to each fullscreens
* Fix wrong breakpoint postfixs and remove modalFullscreen() mixin
* Improve documentation

ysds (approved):
LGTM, but need another approval from @twbs/css-review

MartijnCuppens (approved):
I've added border radii resets to the footer and header, just in case a background color is added. Thanks all for contributing to feature! ❤️

MartijnCuppens:
@XhmikosR, how are we gonna merge this? Squash it into one commit?

XhmikosR:
This should probably me manually rebased and squash the patches per author.

XhmikosR:
Unless you guys don't mind squashing everything into one when merging (with a clean commit message, though)

MartijnCuppens:
I personally don't care. We could maybe add Co-authors in the commit message:

```
Co-authored-by: XhmikosR <xhmikosr@gmail.com>
Co-authored-by: Shohei Yoshida <fellows3@gmail.com>
Co-authored-by: Martijn Cuppens <martijn.cuppens@gmail.com>
```

XhmikosR:
@ysds can you have another look after the rebase?

@GeoSot please make sure you don't overwrite my rebase 🙂

yadoudou:
unsubscribed



---Original---

ysds (approved):
All looks good to me 👍

PR #28687, opened 2019-04-25
Add utility border-width

GeoSot:
Give the opportunity to change Border widths, using utility class "border-w-{number}"

### DIFF
--- a/scss/utilities/_borders.scss
+++ b/scss/utilities/_borders.scss
@@ -73,3 +73,12 @@
 .rounded-0 {
   border-radius: 0 !important;
 }
+
+//
+// Border-width
+//
+@for $i from 1 through 5 {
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
Hi @GeoSot,

We 're considering a different way to define our utilities for `v5` (see https://github.com/twbs/bootstrap/pull/28445) to get more control over all utilities, so I'm going to put this on hold for now. But border-width utilities is definitely something we should look in to, thanks.

mdo:
While @MartijnCuppens put this on hold, I don’t see us adding multiple border widths to the project other than `0` and our global `border-width` variable value. Beyond that, it’s custom styles and extensions (perhaps via Sass maps for two default values)?

PR #28689, opened 2019-04-25
Remove redundant properties in card components & fix #29396

### LINKED ISSUE #29396: Card images in card deck render with incorrect aspect ratio on IE11 when image type is JPG
On Windows 10 using IE 11:

https://getbootstrap.com/docs/4.3/components/card/#card-decks

The HTML snippet for card decks shows `` being used for the card images in a card deck structure. However, the example actually shown uses `<svg class="card-img-top">`, which behaves differently on IE 11.

With an `svg` element in that place, as the width scales down, the aspect ratio is preserved but the image contents are cropped appropriately. With an `img` element, however, on most browsers, the aspect ratio is preserved, but on IE 11, it is not preserved.

The specific problems I would call out are:
1. Your HTML snippet is inconsistent with your rendered example of Card Decks in your documentation.
2. Your HTML snippet for card deck does not work appropriately on IE 11 when the image source is JPG.

MartijnCuppens:
-  `display: flex;` & `flex-direction: column;` has no effect on the card decks & group for smaller viewports.
- `.card` already has `display: flex;` and `flex-direction: column;`, no reason to re add this.

Closes #29396

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -131,19 +131,20 @@
   padding: $card-img-overlay-padding;
 }
 
-.card-img {
+.card-img,
+.card-img-top,
+.card-img-bottom {
+  flex-shrink: 0; // For IE: https://github.com/twbs/bootstrap/issues/29396
   width: 100%; // Required because we use flexbox and this inherently applies align-self: stretch
-  @include border-radius($card-inner-border-radius);
 }
 
-// Card image caps
+.card-img,
 .card-img-top {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
This one should be fairly save to merge, ping @twbs/css-review for a review.

ysds (approved):

MartijnCuppens:
LGTM now

PR #28696, opened 2019-04-26
Change custom file markup

### LINKED ISSUE #25497: Feature/Enhancement: The CSS of custom file input is unnecessarily dumbed-down and limiting.

### LINKED ISSUE #26933: Custom file LABEL text overlapping neighbouring HTML elements on Firefox

### LINKED ISSUE #28389: custom-file-border-radius-after is not customizable

ysds:
Please see https://github.com/twbs/bootstrap/pull/28511#issuecomment-474381351.

By changing the markup of custom file, remove the way of translating button labels by CSS, and support a long file name. Now, you can change the button label by HTML.

And for fix #28389, move the border to the each children of the `.custom-file-label`. Another way by `overflow: hidden` is suddested (https://github.com/twbs/bootstrap/pull/28388#issuecomment-468898020), but it does not round the inside of the corner.

I also added the `transition` to` .custom-file-text` and `.custom-file-button` for focused border color animation.

Closes #25497
Closes #26933
Closes #28388
Closes #28389

Preview: https://deploy-preview-28696--twbs-bootstrap.netlify.com/docs/4.3/components/forms/#file-browser


MartijnCuppens:
Maybe we should also include an example wit a longer placeholder to demonstrate the truncation?

ysds:
I checked the SCSS again and confirmed that there are no problem, I updated the PR comment and remove the WIP.

@MartijnCuppens Could you add the example? if you are okay with that 🙏

MartijnCuppens (approved):
Done

XhmikosR:
Wouldn't make more sense to use the grid in this and the previous example? That way we won't have those long forms.

mdo:
I think I’m good with this, but does it come at the cost of a screen reader announcing the button text since it’s in the label?

patrickhlauke:
@mdo had a super-cursory check with NVDA and Jaws on Win (Chrome/Firefox/IE), and it seems to behave no different from the current implementation, unless i'm missing something?

mdo:
Hadn’t compared to current behavior, so just wanted to ask the question to see if this a regression or area for improvement for us. If you’re good with it @patrickhlauke, I’m good here.

patrickhlauke:
it seems appropriate to me (not extensively tested/compared to native clean unaltered file input, but these are notoriously icky - particularly in Firefox/NVDA - and just checking the above seemed good for me)

patrickhlauke (approved):

mdo (approved):
<3, thanks @ysds!

Btw, I sent you a note to the email on your GitHub profile. You have some fantastic contributions to Bootstrap and we’d love to have you join our team officially if you’ll accept! No pressure either way.

alecpl:
Thanks. I suggest small improvements. Set line-height/font-weight on the `.custom-file-label` element instead of both spans inside. And use `.custom-file-label > *` selector to assign some common styles like display, padding, border and transition.

MartijnCuppens:
> Set line-height/font-weight on the .custom-file-label element instead of both spans inside.

Might indeed be a nice improvement.

> And use .custom-file-label > * selector to assign some common styles like display, padding, border and transition.

I don't mind using the universal selector in case we gain a lot of improvement by it. Since we know what the classes of the children will be (`.custom-file-text` & `.custom-file-button`) I don't think it's worth adding another selector to add common styles here.

PR #28697, opened 2019-04-26
Improve progress-bar-stripes animation

MartijnCuppens:
- `background-position: 0 0;` is the default background position, so redundant
- The background position should only be animated in the x direction
- `0%` is a little shorter than `from`


mdo (approved):

XhmikosR:
I personally find it clearer the way it was with `from`/`to`...

How is this in the minified version without this patch?

MartijnCuppens:
I prefer percentages over `from/to` because we have animation which does not stop at `100%`: https://github.com/twbs/bootstrap/blob/09c99fe6e79ad397fdef262017d7b2bc49377b15/scss/_spinners.scss#L31-L38

I guess we better always use the percentages values. It seemed like there's no stylelint rule for this yet, so I've opened https://github.com/stylelint/stylelint/issues/4065 upstream.

In the minified version those values are not converted.

XhmikosR:
BTW this reminds me, maybe we should remove the button to toggle the animations? I mean, we don't do the same for spinners.

XhmikosR:
FYI this broke https://twbs-bootstrap.netlify.com/docs/4.3/components/progress/#animated-stripes

Tested FF, Chrome and Edge

XhmikosR:
Turns out this is a regression of cf821e1d4d1d67f6b4ce9651ae64c72a502c40ba, not CSS-related.

PR #28698, opened 2019-04-27
Reboot & Type cleanup

MartijnCuppens:
- Line height & font family from html were directly overridden by body, so I removed those
- The remaining properties are moved to body
- Shim for IE10 tags removed
- hr cleanup
  - Combined styles from reboot & `_type.scss`
  - Changed from borders to background
    - This allows us to keep the color inherited.
    - Opacity can be used to change the intensity
    - [Size](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_hr_size) attribute will change the height instead of making the hr look like an input field
  - Better variable names
- headings cleanup
  - Moved all element styling to reboot. This was actually [a bug](https://codepen.io/MartijnCuppens/pen/bJzQWJ) because before our docs claimed we had different [font sizes for headings](https://getbootstrap.com/docs/4.3/content/reboot/#headings-and-paragraphs)
  - `_type.scss` will now extend the element styling. This way we will save some redundant code
- `small` & `mark` are also moved to reboot. `_type.scss` does not contain any element styling anymore
- font weight removed from `small`, this did not really made sense


scss/_reboot.scss
```diff
@@ -70,13 +59,20 @@ body {
 
 // Content grouping
 //
-// 1. Add the correct box sizing in Firefox.
-// 2. Show the overflow in Edge and IE.
+// 1. Reset Firefox' gray color
```
mdo: Missing an `s`?

mdo (approved):
Dig the improvements and clearer boundaries between Reboot and custom styles.

PR #28701, opened 2019-04-28
Added !important to sr-only and sr-only-focusable SASS mixins

### LINKED ISSUE #28691: .sr-only should use !important
It seems like .sr-only should explicitly use `!important` for all attributes. If you are using `.sr-only` anywhere, you want ALL of it's attributes to take priority no matter what. There are no exceptions to this.

There are many cases where server side templating logic would add the `sr-only` class to an element depending on various conditions. The problem here is that if that element has, for example `padding` on it, and this style is defined AFTER you added your Bootstrap CSS to a page, then the `sr-only` class is never going to have priority in the cascade, so therefore the `padding: 0` that is part of the `sr-only` class is going to get overwritten by the `padding` on the element that you defined elsewhere. Does that make sense?

Jakobud:
Closes #28691


Jakobud:
This is regarding #28691

patrickhlauke (dismissed):

ysds:
@patrickhlauke I added comment in #28691. Could you re-review the necessity need to set "! Important" for all properties?

patrickhlauke:
@ysds i really don't see the harm of just having everything `!important` regardless of explicit need. to me, makes it easier than having to potentially miss out/test permutations of what does/doesn't need to be set...
so whatever, retracting my review for now...

MartijnCuppens:
Hmm, this looks like a _nothing can go wrong here_ PR, but @ysds' seems to be right here. Adding `!important` to everything can lead to some unexpected results. In fact even without `!important` we already have an issue: https://codepen.io/MartijnCuppens/pen/Pgrpwe

I think we should have a look at which properties are still necessary and then rethink about which should be `!important` before maybe introducing more problems.

I also just noticed we did a way shorter notation for the button groups, maybe that's more what we should look into:
https://github.com/twbs/bootstrap/blob/09c99fe6e79ad397fdef262017d7b2bc49377b15/scss/_button-group.scss#L156-L161

Jakobud:
I'm not sure I'm seeing the issue in that codepen. The sr-only button appears only when focused, which appears to be functioning correctly...

Jakobud:
Maybe the solution on `sr-only-focusable` would be to set values to use `inherit` so that they inherit their values (padding, margin) to what they should be according to their other classes when it's focused...

PR #28705, opened 2019-04-29
Travis CI: add Node.js 12 and use `cache: npm`.

XhmikosR:
Needs #28655 

@Johann-S: should we add Node.js 12 in v4 too? Note that I haven't switched to 12 in all places, I still use the current LTS (10).


Johann-S:
I don't think we should add Node 12 for v4, because Node 10 will be maintained for a long time compared to our v4

XhmikosR:
Yup, agreed then.

Johann-S (approved):

PR #28713, opened 2019-05-01
Remove redundant support query

MartijnCuppens:
The documentation about `.sticky-top` is incorrect & the I doubt the `@support` query is really necessary.

> **IE11 and IE10 will render `position: sticky` as `position: relative`.**

Nope, it won't. It 'll render as it's initial position value, probably `static`.

> As such, we wrap the styles in a `@supports` query, limiting the stickiness to only browsers that can render it properly.

If it's not supported, there's not really any difference in using the `@support` query or not, unless the position is explicitly set to anything else than `static`. IMO we can remove the support query, it's feels a bit inconsistent to only check for it on that place.


mdo:
I specifically remember running into multiple issues in my Windows installs and this was the only solution to those within our docs layout. Can you put together some test cases, or at least explore how this affects the docs (bare in mind, docs might have custom CSS, can’t recall)?

XhmikosR:
And use BrowseStack to test this on real browsers :)

ysds:
I tested on IE11.

> It 'll render as it's initial position value, probably `static`.

Right, but the `.navbar` sets `position: relative`. 

I'm afraid we can't rule out the possibility that the `position: relative` may cause some unintended problems by `top` and `z-index`. but, I could not find the problem in the scope of the documents and the examples.

MartijnCuppens:
Closing per @ysds' comment

PR #28715, opened 2019-05-02
Correcting parentheses in the text

lucalves:


patrickhlauke:
no, current parentheses are correct (both the js and minified min.js are "bundled JS files")

PR #25806, opened 2018-03-07
Collapse with flex elements - simple approach

MartijnCuppens:
Fixes #22600 

Simple solution instead of #22758. This solution doesn't add a display property to an element when it's visible.

Demo: https://codepen.io/MartijnCuppens/pen/jzNyyy

cc @Johann-S @mdo @XhmikosR

PR #28720, opened 2019-05-02
`.sr-only` & `.sr-only-focusable` tweaks

### LINKED ISSUE #28691: .sr-only should use !important

MartijnCuppens:
### Analysis of the `.sr-only` properties:
`position: absolute`: needed to prevent the element from taking up space in the document
`width: 1px`: prevents the element from becoming too wide and therefor creating scrollbars
`height: 1px`: prevents the element from becoming too tall and therefor creating scrollbars
`padding: 0`: padding could increase the width or height even when those are set to `1px`
`overflow: hidden`: prevent the text inside from breaking out the box
`clip: rect(0, 0, 0, 0)`: make the element invisible
`white-space: nowrap`: see https://github.com/twbs/bootstrap/pull/22154
`border: 0`: border could increase the width or height even when those are set to `1px`

So yeah, we all need them.

We had this problem: https://github.com/twbs/bootstrap/issues/28691, which could be fixed by adding `!important` to all properties, but that would just make this problem bigger: https://codepen.io/MartijnCuppens/pen/Pgrpwe.

And than I remembered a technique I used before (https://github.com/twbs/bootstrap/pull/25806), just make use of the `:not()` selector. This would solve #28691 without introducing padding issues.

This would also not require the `.sr-only` class on elements with `.sr-only-focusable`. I've changed the docs a bit to make this clear. 

I added @Jakobud as co author for the `!important` commit since he already fixed that in #28701.

Fixes https://github.com/twbs/bootstrap/issues/28691
Closes https://github.com/twbs/bootstrap/pull/28701

### DIFF
--- a/scss/mixins/_screen-reader.scss
+++ b/scss/mixins/_screen-reader.scss
@@ -1,34 +1,28 @@
+// stylelint-disable declaration-no-important
+
 // Only display content to screen readers
 //
 // See: https://a11yproject.com/posts/how-to-hide-content/
 // See: https://hugogiraudel.com/2016/10/13/css-hide-and-seek/
 
 @mixin sr-only {
-  position: absolute;
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px; // Fix for https://github.com/twbs/bootstrap/issues/25686
-  overflow: hidden;
-  clip: rect(0, 0, 0, 0);
-  white-space: nowrap;
-  border: 0;
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: -1px !important; // Fix for https://github.com/twbs/bootstrap/issues/25686
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
 }
 
-// Use in conjunction with .sr-only to only display content when it's focused.
+// Use to only display content when it's focused.
 //
 // Useful for "Skip to main content" links; see https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1
-//
-// Credit: HTML5 Boilerplate
 
 @mixin sr-only-focusable {
-  &:active,
-  &:focus {
-    position: static;
-    width: auto;
-    height: auto;
-    overflow: visible;
-    clip: auto;
-    white-space: normal;
+  &:not(:focus) {
+    @include sr-only();
   }
 }
--- a/site/content/docs/4.3/getting-started/accessibility.md
+++ b/site/content/docs/4.3/getting-started/accessibility.md
@@ -37,10 +37,10 @@ Content which should be visually hidden, but remain accessible to assistive tech
 </p>
 {{< /highlight >}}
 
-For visually hidden interactive controls, such as traditional "skip" links, `.sr-only` can be combined with the `.sr-only-focusable` class. This will ensure that the control becomes visible once focused (for sighted keyboard users).
+For visually hidden interactive controls, such as traditional "skip" links, the `.sr-only-focusable` class. This will ensure that the control becomes visible once focused (for sighted keyboard users). **Watch out, since Bootstrap 5 the `.sr-only-focusable` class may not be used in combination with the `.sr-only` class.**
 
 {{< highlight html >}}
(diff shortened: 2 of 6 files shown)


ysds:
Using `:not` is nice approach.

I think that the `:not(:active)` come from the current code, but I doubt it's needed (`:not(:focus)` would be enough).

CC: @patrickhlauke

XhmikosR:
@MartijnCuppens: rebased. Please make any further changes if needed.

scss/mixins/_screen-reader.scss
```diff
@@ -1,34 +1,28 @@
...
-  &:active,
-  &:focus {
-    position: static;
-    width: auto;
-    height: auto;
-    overflow: visible;
-    clip: auto;
-    white-space: normal;
+  &:not(:focus):not(:active){
+    @include sr-only;
```
XhmikosR: Maybe `sr-only()` like below? We should probably see if stylelint-scss has a rule for this (IIRC it does).

scss/mixins/_screen-reader.scss
```diff
@@ -1,34 +1,28 @@
...
 @mixin sr-only-focusable {
-  &:active,
-  &:focus {
-    position: static;
-    width: auto;
-    height: auto;
-    overflow: visible;
-    clip: auto;
-    white-space: normal;
+  &:not(:focus):not(:active){
```
patrickhlauke: conceptually, this would mean that as soon as the control/thing is activated (so focus is still on it, user has pressed it) it would disappear again. may result in oddness (e.g. it may be focused, then when activated it disappears, and then once it's still loading but hasn't quite gone through yet, it reappears as focused but not active).

long story short, i'd say try dropping the `:not(:active)`

patrickhlauke (changes requested):
Conceptually, i like it - and prefer the fact that you don't have to double up/combine the `sr-only` and `sr-only-focusable` classes anymore, but just pick one. Not had time to test, but trust that if it does what it says on the tin, the approach is sound (modulo that thing about `:not(:active)`

scss/mixins/_screen-reader.scss
ysds: I created a PR in the stylelint-config-twbs-bootstrap, but should separate to another issue for this point.

XhmikosR: Agreed, I just got reminded about this here :)

scss/mixins/_screen-reader.scss
MartijnCuppens: `:not(:active)` dropped.

ysds (approved):

MartijnCuppens:
I've also added a notice in the migration guide. @patrickhlauke, could you have a look if everything is ok now?

XhmikosR:
Ping @patrickhlauke

site/content/docs/4.3/getting-started/accessibility.md
```diff
@@ -37,10 +37,10 @@ Content which should be visually hidden, but remain accessible to assistive tech
 </p>
 {{< /highlight >}}
 
-For visually hidden interactive controls, such as traditional "skip" links, `.sr-only` can be combined with the `.sr-only-focusable` class. This will ensure that the control becomes visible once focused (for sighted keyboard users).
+For visually hidden interactive controls, such as traditional "skip" links, the `.sr-only-focusable` class. This will ensure that the control becomes visible once focused (for sighted keyboard users). **Watch out, since Bootstrap 5 the `.sr-only-focusable` class may not be used in combination with the `.sr-only` class.**
```
patrickhlauke: first sentence needs a verb...like "use"

"For visually hidden interactive controls, such as traditional "skip" links, use the `.sr-only-focusable` class."

patrickhlauke (dismissed):
With exception of the tiny change request, everything else looks good to me and can be merged after change is made.

patrickhlauke (approved):

PR #28733, opened 2019-05-06
Fix incorrect variable for popover border radius

ysds:
Use `$popover-border-radius variable` instead of `$border-radius-lg`.

### DIFF
--- a/scss/_popover.scss
+++ b/scss/_popover.scss
@@ -157,8 +157,7 @@
   color: $popover-header-color;
   background-color: $popover-header-bg;
   border-bottom: $popover-border-width solid darken($popover-header-bg, 5%);
-  $offset-border-width: calc(#{$popover-border-radius} - #{$popover-border-width});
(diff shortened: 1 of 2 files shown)


ysds:
I noticed that the `_popover.scss` uses the local scope `$offset-border-width` variable.

Should we add `$popover-inner-border-radius` to `_variable.scss` like the `$card-inner-border-radius` for consistency?

But I like local scope because there is no case to customize those variables.

MartijnCuppens:
> Should we add $popover-inner-border-radius to _variable.scss like the $card-inner-border-radius for consistency?

Yup

> But I like local scope because there is no case to customize those variables.

But disables the ability to set the radius to `null` for example. I also know there are some issues with `calc()` and width of `0` which can be avoided by moving this variable to `_variables.scss`

ysds:
Done :)

MartijnCuppens (approved):

PR #28736, opened 2019-05-06
Add alert glyphicon

ashfahan:
Add Alert glyphicon for

alert-success
alert-info
alert-warning
alert-danger

Untitled


patrickhlauke:
thanks, but we'll pass. we've already removed glyphicons elsewhere when we moved from version 3 to version 4.

PR #28741, opened 2019-05-07
Album example: remove misleading jumbotron-heading classname

### LINKED ISSUE #28737: Causers of confusion.
You use custom css in the examples, I was trying to find what the jumbotron-heading class is, I did not find it, after losing 20 minutes I discovered it's a custom css ... I think the examples should not use custom css.

patrickhlauke:
Closes #28737

### DIFF
--- a/site/content/docs/4.3/examples/album/album.css
+++ b/site/content/docs/4.3/examples/album/album.css
@@ -15,7 +15,7 @@
   margin-bottom: 0;
 }
 
-.jumbotron-heading {
+.jumbotron h1 {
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
The jumbotron is going to be removed in `v5`, so I removed it from the project.

XhmikosR:
I still think this should target master since as of the time of writing, jumbotron isn't removed. It's the same flow we've been following.

MartijnCuppens (approved):

Issue #27888, opened 2018-12-20
Page overlaps header/footer when printing in chrome


shoe788:
Windows 10
Chrome 71.0.3578.98

[Image 1 depicting the problem](https://i.imgur.com/cvr3wu6.png)

[Image 2 depicting the problem](https://i.imgur.com/7BQAH9p.png)

Steps to reproduce.

1. Visit one of the examples on https://getbootstrap.com/docs/4.1/examples/
2. Ctrl-P to print
3. Set Paper size to 8.5x11, margins to "Default", and Headers and Footers to "On"
4. header/footer is visibly cut off from print page

bakedbird:
The root of this issue is that inside the media print rule, the page size is explicitly set to A3, thus setting the default page margins the browser has for the A3 print preset.
Setting it to auto will solve your problem. 

I don't know why this has been set to A3 explicitly. If it has been done inadvertently, I am up to a quick PR.

shoe788:
I changed page size to auto and it cleared that up. Thank you.

SanjoyNath:
I faced the issue with Wikipedia prints. I hided the side bars in wiki and it solved the issue

Issue #26598, opened 2018-05-26
Changing page size to A4 causes overflow


mikeday:
https://github.com/twbs/bootstrap/blob/c2b13b9c7851d62d3072989552a784e6c598580b/scss/_print.scss#L93

Specifying a large min-width for the body element (992px) makes the content overflow when the page size is reduced to A4; if Safari does not respect page size, could there be another way of doing this? Or perhaps restricting the min-width rules for Safari only?

When our customers use Bootstrap with Prince the output is often suboptimal and breaks in confusing ways due to the print style sheet, and fixing the page size would be a big help.

SampsonCrowley:
yeah the min width here is breaking a lot of print styles that are relying on sizes to be accurate (eg. div set to height: 1.5in in print media) because the min-width is bigger than the set page size and that forces the browser to scale the page down. this is worse because for some reason `min-width: none` isn't a valid CSS value, and you have to remember to use `unset`

it at least needs to be brought down a bit (maybe 768px?) so people don't spend as much time as I just did trying to find why fixed unit print sizes were all being scaled down when actually printed

SampsonCrowley:
for future users that end up here, this is what I'm using to make sure printing margins and page sizes are scaled to the printed medium:

```
@media print {
  ::first-letter{
    margin: 0;
    padding: 0;
  }
  html {
    &, body {
      margin: 0;
      background: #fff;
      background: transparent;
      padding-top: 0;
      padding-bottom: 0;
      font-size: 11px;
    }
  }
  .container {
    padding: 0 !important;
    margin:0 !important;
    width: 100% !important;
    max-width: none !important;
    min-width: unset !important;
  }

  @page {
    size: 8.5in 11in;
    margin: 0.25in;
  }
}
```

mdo:
Closing as a stale won't fix I think—we've dropped the bulk of the print styles in v5 and I don't know how much we can change in v4 without it being considered a breaking change.

SampsonCrowley:
@mdo would adding the small snippet above be acceptable?

PR #28753, opened 2019-05-09
Set auto as default @page size

coorasse:
See:

* https://github.com/twbs/bootstrap/pull/25164
* https://github.com/twbs/bootstrap/issues/27888
* https://stackoverflow.com/questions/49785039/printing-bootstrap-4-html-page-in-chrome-cuts-off-chrome-header-footers/50400421

I think any option would be better than "a3"


MartijnCuppens (dismissed):
I've changed the value to `null`, since this won't generate any additional css. Need an additional approval from @twbs/css-review to be sure.

patrickhlauke:
does this cause any additional issues a la https://github.com/twbs/bootstrap/issues/26598, or does it solve them?

MartijnCuppens:
I did some research about this and it seems browsers behave pretty inconsistent and there are a lot of factors we can not control:
- We never know what the size of the paper will be, so whenever we try to fix this for one paper size with fixed widths, it will break for other cases.
- Media queries seem to behave very buggy for print. I've used [this technique](https://codepen.io/MartijnCuppens/pen/jjBJyG) to determine the width of a page, but the results I got were pretty inaccurate:
  - IE/Edge: `648px`
  - Chrome/ New Edge: `540px`
  - Safari: The window width
  - Firefox: `793px`
- Safari seems to print the page with the width of the window, while chrome has a width around ~1260px (A4 paper size), for Firefox it's pretty hard to determine, since it scales all content whenever its wider than, I don't know what happens with Edge or IE, I gave up after FF.  
- Browsers rescale the page "when needed". This "when needed" is pretty opinionated by the browsers since every browser seems to do this in a different way.

I think we should just ditch the page size and `$print-body-min-width`. Yes, browsers will generate different print layouts, but that's better than the behaviour we have now where parts of the pages are cut of.

Maybe we should also mention the problems with print layouts in our docs.

XhmikosR:
Just a reminder, we need to be extra careful with this since it's for v4. If we can't be 100% sure about any side effects, I say we leave it as is...

patrickhlauke:
historically the print styles have been broken/problematic in many tiny aspects, so if this fixes at least some of the common situations even though it may not get to 100% perfection every time, it's probably an improvement though

MartijnCuppens:
I agree with Patrick. https://github.com/twbs/bootstrap/pull/25164 broke more than it fixed I'm afraid. But this PR shouldn't be merged without removing (or `null`-ifying) the `$print-body-min-width` thing.

MartijnCuppens:
@mdo, what do you think, should we remove the size and min-width or leave it as is?

MartijnCuppens:
Ping @mdo

XhmikosR:
Ping @mdo for a decision.

mdo (dismissed):

XhmikosR:
I'm still not sure about this change @mdo. It can and will introduce change in behavior. See the above comments https://github.com/twbs/bootstrap/pull/28753#issuecomment-504671905

mdo:
Hmm, re-reading again, I could be convinced to not do this and just drop it entirely like we have in v5.

MartijnCuppens:
Fine by me, close it than?

PR #28758, opened 2019-05-10
Add new .text-break-all utility for applying word-break: break-all

CrutchTheClutch:
Builds off of #27863 by adding an additional text-break utility that forces line breaks regardless of word spacing.  I feel that this utility would be useful for situations where an element's content may dynamically generate a long string, such as displaying a username.


MartijnCuppens:
Sorry for postponing this so long. It would indeed make sense to have this utility class.

We should use `word-wrap` for this (See https://github.com/twbs/bootstrap/pull/29325#issuecomment-570803307). And we 're using the utility API nowadays, so this PR needs to be adjusted.

MartijnCuppens:
Forked repo is deleted, unable to rebase this. Since this is the only request for this utility, I'm not gonna create a new PR for this yet.

PR #28760, opened 2019-05-12
fix popover arrow by selecting the first element child

### LINKED ISSUE #28692: popover arrow position broken (v5)
'Popover' inherits from 'Tooltip'. 
Before https://github.com/twbs/bootstrap/pull/28105 there was only `.arrow` selector.
Currently in Tooltip there is `.tooltip-arrow` so `Popover` pass wrong selector (should pass `.popover-arrow` to popper.js.

Johann-S:
I use the `tip.firstChild` element as arrow, it's better than passing a selector to Popper.js

Fixes: #28692

/CC @wojtask9

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -111,8 +111,7 @@ const ClassName = {
 }
 
 const Selector = {
-  TOOLTIP_INNER: '.tooltip-inner',
-  TOOLTIP_ARROW: '.tooltip-arrow'
+  TOOLTIP_INNER: '.tooltip-inner'
 }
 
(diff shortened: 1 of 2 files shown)


XhmikosR:
But we do know what the selector will be, don't we?

Johann-S:
Yep we know but for doing that we have to create two new getters and the arrow selector isn't customisable

Johann-S:
@XhmikosR finally found a solution pretty much the same as it was before 👍

XhmikosR (approved):

PR #28783, opened 2019-05-17
Add auto-positioning for our toast component

Johann-S:
Changes:
 - added auto positioning for our toast component (top-center, top-left, top-right, bottom-center, bottom-left, bottom-right).
 - Removed the use of `.hide` because I cannot get elements widths

TODO:

- [x] Top middle
- [x] Bottom middle
- [x] Docs
- [x] Unit tests
- [x] Fix docs example situation while a toast is still showing
- [ ] Revisit the apply styles at once approach
- [ ] Add migration notes

You can test it here: 

- https://deploy-preview-28783--twbs-bootstrap.netlify.app/docs/5.0/components/toasts/#placement
- https://codepen.io/Johann-S/pen/BedGrO


ysds:
If can add a something wrapper element with `flex-direction: column;` or `flex-direction: column-reverse;` it may be able to remove the dynamic top/bottom calculation.

Johann-S:
It depends because with that feature our users don't have to create a specific wrapper for their toast, so they can add their toast at the end of their body and it'll be well positionned, so if they add `flex-direction: column` or `flex-direction: column-reverse` on the body element it can happen strange things 😄

ysds:
Dynamic top/bottom calculation will be OK if we does not expect a use case of dynamically increase/decrease a toast message and combination with collapsible (expandable) content:

スクリーンショット 2019-05-18 0 06 10

Johann-S:
That's true, in that case we can add an `updateSize` method for our toast component

MartijnCuppens:
What's the purpose of the changes in `scss/_toasts.scss`? The margins are now removed

Johann-S:
To handle margin for bottom toasts and I didn't removed any margin, please take a look at the demo

MartijnCuppens:
But there's no bottom margin applied to the last toast? This change will make it required to add the `.top-left`, ect classes for the margins to apply: https://codepen.io/MartijnCuppens/pen/Ezvebb

I think we better make use of stand alone utilities instead of setting the position in js. Something like this: https://codepen.io/MartijnCuppens/pen/joLvxr

Johann-S:
You have to build the JS to see it working 😉 because it'll add the correct CSS class according to the positioning

With your utilities on small screen it looks like that:

image

We can handle that pretty easely with JS. 
I made another live example like that you should see it working: https://codepen.io/Johann-S/pen/BedGrO

Johann-S:
@mdo any feedbacks on that proposal?

mdo:
Somehow missed the ping here, but I'm super into this!

XhmikosR:
It seems there's an issue when changing the position in the docs example while a toast is still showing.

js/src/toast.js
```diff
@@ -35,20 +36,31 @@ const EVENT_SHOW = `show${EVENT_KEY}`
 const EVENT_SHOWN = `shown${EVENT_KEY}`
 
 const CLASS_NAME_FADE = 'fade'
-const CLASS_NAME_HIDE = 'hide'
 const CLASS_NAME_SHOW = 'show'
 const CLASS_NAME_SHOWING = 'showing'
+const PositionMap = {
+  TOP_CENTER: 'top-center',
```
XhmikosR, line 42: Can we change these to strings?

js/src/toast.js
```diff
@@ -166,6 +183,122 @@ class Toast {
...
+          bottom += (parseInt(height, 10) + parseInt(marginTop, 10))
+        }
+
+        toastEl.style.bottom = `${bottom}px`
+      }
+    })
+  }
+
+  _clearPositioning() {
+    this._element.style.position = ''
```
XhmikosR, line 277: Isn't there any way to set multiple properties at once? Maybe with `style.cssText`? IIRC this can be quite costly because the element is repainted every time a property changes. Might be worth checking other places too later :)

Johann-S:
> It seems there's an issue when changing the position in the docs example while a toast is still showing.
@XhmikosR yep that's because we use just one toast, we doesn't generate new one each time, so we cannot cancel the previous transition

js/src/toast.js
Johann-S: yep sure 👍

js/src/toast.js
Johann-S: I will check that thanks 👍

XhmikosR:
I'm pretty sure other will hit the issue I mentioned too, but I'm unsure how to work around it.

Maybe we could have a setTimeout with the default Toast duration and add toggle the disabled attribute?

js/src/toast.js
```diff
@@ -166,6 +182,126 @@ class Toast {
...
+
+  _clearPositioning() {
+    Manipulator.applyCss(this._element, {
+      position: '',
+      right: '',
+      left: '',
+      bottom: '',
+      top: ''
+    })
+    this._element.classList.remove(POSITION_TOP_RIGHT)
```
XhmikosR, line 286: I think we can pass multiple classes here?

js/tests/visual/toast.html
```diff
@@ -64,19 +82,28 @@ <h1>Toast <small>Bootstrap Visual Test</small></h1>
             new Toast(toastNode)
           })
 
+          var selectToastPlacement = document.getElementById('selectToastPlacement')
+          selectToastPlacement.addEventListener('change', function () {
+            Array.from(document.querySelectorAll('.toast'))
```
XhmikosR, line 87: This is ES6, just pointing it out

Johann-S: yep, that's just for our tests

js/src/toast.js
Johann-S: I wasn't sure about it, but yep I'll do it 👍

Johann-S:
@XhmikosR finally I was able to avoid that side effect by resetting the css before showing the toast

site/content/docs/5.0/components/toasts.md
```diff
@@ -132,11 +131,24 @@ Building on the above example, you can create different toast color schemes with
...
-{{< example class="bg-dark" >}}
-<div aria-live="polite" aria-atomic="true" style="position: relative; min-height: 200px;">
-  <div class="toast" style="position: absolute; top: 0; right: 0;">
+Our toast are dynamically placed according to the position you specified and to their container. The top right is often used for notifications, as is the top middle. If you're only ever going to show one toast at a time. If you have several toasts displayed at once, they will stack automatically.
+
+{{< example >}}
+<form>
+  <div class="form-group mb-3">
+    <label for="selectToastPlacement">Toast placement</label>
+    <select class="form-control mt-2" id="selectToastPlacement">
```
XhmikosR, line 140: We need an initial placeholder option value here :) After we add it, feel free to backport it to the visual test too

site/content/docs/5.0/components/toasts.md
```diff
@@ -132,11 +131,24 @@ Building on the above example, you can create different toast color schemes with
 
 ## Placement
 
-Place toasts with custom CSS as you need them. The top right is often used for notifications, as is the top middle. If you're only ever going to show one toast at a time, put the positioning styles right on the `.toast`.
-
-{{< example class="bg-dark" >}}
-<div aria-live="polite" aria-atomic="true" style="position: relative; min-height: 200px;">
-  <div class="toast" style="position: absolute; top: 0; right: 0;">
+Our toast are dynamically placed according to the position you specified and to their container. The top right is often used for notifications, as is the top middle. If you're only ever going to show one toast at a time. If you have several toasts displayed at once, they will stack automatically.
```
XhmikosR, line 134: This needs some word proofing @mdo

js/src/toast.js
```diff
@@ -183,6 +183,7 @@ class Toast {
   // Private
 
   _positionToast() {
+    this._clearPositioning()
```
XhmikosR, line 186: Doesn't this change affect the current behavior though? I mean, we didn't clear their position previously? If it's only for our docs, maybe this shouldn't be here?

site/content/docs/5.0/components/toasts.md
MartijnCuppens: The toast seems to behave a little bugged since we autohide it. Imo, we should disable that and use a default position value like @XhmikosR suggested. 

Also, the class needs some adjustment:
```suggestion
    <select class="form-select mt-2" id="selectToastPlacement">
```

js/src/toast.js
Johann-S: It doesn't change the current behavior it improve it, because before that we cleared the positioning on hide, but with that change we can change position without hiding the toast

mdo:
The codepen has a small bug maybe? If I click the button more than once, the toasts overlap for some reason. See the doubled up box-shadow on part of the toast.

Screen Shot 2020-10-21 at 9 13 14 AM

site/content/docs/5.0/components/toasts.md
```diff
@@ -132,11 +131,25 @@ Building on the above example, you can create different toast color schemes with
...
```
mdo, line 134: If I understand what's being proposed here, this should suffice.

```suggestion
Toasts are dynamically placed relative to their container with the position you specify. The top right and top middle are often the most often used for notifications across devices and operating systems. If you have more than one toast displayed at the same time, they will automatically stack.
```

mdo (commented):
We'll need to mention this in our Migration docs.

Johann-S:
@mdo it's a little bug in our Toast component we don't check if a toast is shown (see: https://github.com/twbs/bootstrap/blob/jo-auto-pos-toast/js/src/toast.js#L102)

Not sure if I should handle that here

mdo:
If we can handle it, that'd be awesome!

Johann-S:
@mdo done 👍

XhmikosR:
Hmm, now when you select another position on the docs while the toast is still showing it does nothing.

Johann-S:
is it really a bad thing? It doesn't seems to me a bad behavior

mdo:
I think it'd be nice if we maybe didn't use a select menu, and instead had a button for each option? Or keep the select, and click a separate button to trigger a toast per click? Right now I agree with @XhmikosR that it doesn't feel right.

XhmikosR:
I'm also having second thoughts about the "apply styles at once" patch unless of course we plan to make more use of it later in the other components too.

XhmikosR:
@Johann-S how about disabling the select while a toast is showing? Just thinking out loud. If anyone has any other suggestion let us know.

Johann-S:
yep that's something we can do, but for me it's a good thing it doesn't nothing we try to change the position and a toast is showing, if someone wants to do that, he should call `preventDefault` and change the position

XhmikosR:
Yeah, but we do want that because the current behavior isn't good UE. I know for sure people won't wait for a toast to finish showing before switching positions. Hence why I suggested that we do that for the docs example :)

XhmikosR:
@Johann-S the last patch doesn't seem to work :/

Johann-S:
yes I made a typo mistake I'll fix that

Johann-S:
@XhmikosR can you explain a bit more about "Revisit the apply styles at once approach"? It's the last thing to tackle

XhmikosR:
The disable patch still doesn't work right for me: it does disable the select, but enables it again before the Toast has finished showing instead of after.

About "Revisit the apply styles at once approach". Do we have other places in the codebase that we could use the newly added function? I do know that we shouldn't change an element's style a lot because it's causing a reflow, so this should fix it, but I'm wondering if we can apply this to other components later. If not, perhaps the patch should be moved to a separate PR.

Johann-S:
> About "Revisit the apply styles at once approach". Do we have other places in the codebase that we could use the newly added function? I do know that we shouldn't change an element's style a lot because it's causing a reflow, so this should fix it, but I'm wondering if we can apply this to other components later. If not, perhaps the patch should be moved to a separate PR.

We don't have another place where we can use it, so if you prefer I can move that function in the Toast component ?

XhmikosR:
> We don't have another place where we can use it, so if you prefer I can move that function in the Toast component ?

Are you sure we can't do something similar in the other components? From a quick look I see that modal might benefit from this.

I don't mind having this in utils assuming we can reuse it later :)

MartijnCuppens:
A bit late to the party here, but I think we can fix this with CSS quite easily. Thoughts on this concept:

https://codepen.io/MartijnCuppens/pen/qBaEYNG

This way we don't need to position our elements with JS. Which could fix dynamic changes to the toast height (image loaded, font loaded, async data loaded) & responsive behaviour.

Johann-S:
@XhmikosR yep it would be great if we can use that in our Modal component 👍 

> Which could fix dynamic changes to the toast height (image loaded, font loaded, async data loaded) & responsive behaviour.

Those things can be handle by JavaScript too @MartijnCuppens but again here it's a matter of should we keep JavaScript in our codebase or not ?

mdo:
Closing for the CSS-only approach in #32280.

PR #28785, opened 2019-05-17
Clean input group flex basis

### LINKED ISSUE #28782: input-group looks compressed and broken.
(Mac OS Mojave 10.14.4, Chrome 74.0.3729.157)

Looking at the following code for an `input-group`
https://github.com/twbs/bootstrap/blob/2d6e086342a93a59ae8abeb17460e9fab50ec267/scss/_input-group.scss#L7-L30
there are two interesting lines here: line 10 (`flex-wrap: wrap`) and line 22 (`width: 1%`) which seem to be the cause for the following rendering

input-group

And the code is
```html
<div class="input-group input-group-sm">
  <select class="custom-select" name="some-name">
    …
  </select>
    <button class="btn btn-outline-secondary dropdown-toggle" type="button" data-toggle="dropdown"></button>
    <div class="dropdown-menu vw-30 p-2">
    …
    </div>
  </div>
</div>
```

I’ve fixed this by adding `flex-wrap: nowrap` to the input group itself, and `width: auto` to the select element further down in the cascade. But I am curious about why these two values were chosen in the first place? Shouldn’t the `flex-wrap` be at least `nowrap` to ensure that an input group doesn’t break visually?

ysds:
Fixes #28782

* The `width: 1%` overwrites the default `width: 100%` of `.form-control` etc.
* But the `width: 1%` can be replaced with `flex-basis`, because `flex-basis` has priority (ref: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)
* `%` is needed for IE bug https://github.com/philipwalton/flexbugs#flexbug-4

I've tested on https://codepen.io/fellows3/pen/Ezpdqj


MartijnCuppens (approved):

MartijnCuppens:
Maybe we should backport this to `v4`, @ysds?

ysds:
@MartijnCuppens Yeah, but should check that this work on IE, but I don't have an IE10 env😞 Would you test in browserstack?

XhmikosR:
We need to test on IE 11 + Edge (pre Chrome), and ideally temporarily backport this to a v4-dev based branch and test there too with the aforementioned browsers plus IE 10.

ysds:
Testing in IE11 and Edge has been done. But in IE10 on browserstack is difficult, because need a compiled CSS with autoprefixer for IE10 (not the latest browserslist). I'll create a PR from v4-dev brunch for test. (maybe tomorrow)

PR #28814, opened 2019-05-23
Fix background position for form-validation

sezeresen:
`background-position: center right $input-height-inner-quarter;` is not valid `background-position` property.

After building `create-react-app`, it converts to `100% calc(.375em + .1875rem)`.

This PR fixes this issue.


MartijnCuppens (approved):
LGTM, it would indeed also make more sense if the x position is defined before the y position

Issue #28791, opened 2019-05-19
change carousel-indicators style


forghan89:
![Capture1](https://user-images.githubusercontent.com/47449690/57978970-dd825780-7a2b-11e9-94c8-c28e463835ea.PNG)

Many users want to change the carousel indicators style, 
so please add two classes below:

- **square** :
**.carousel-indicators-square**

`.carousel-indicators-square{
width: 14px;
height: 14px;
}`

![Capture2](https://user-images.githubusercontent.com/47449690/57978976-ff7bda00-7a2b-11e9-904d-2564ea8f7bac.PNG)

- **round** :
**.carousel-indicators-round**

`.carousel-indicators-round{
width: 14px;
height: 14px;
border-radius: 100%;
}`

![Capture3](https://user-images.githubusercontent.com/47449690/57979017-9e083b00-7a2c-11e9-9606-5bc90c65ee95.PNG)

mdo:
Closing per https://github.com/twbs/bootstrap/pull/28815#issuecomment-496408360.

CJpaekil:
Dear
How can I change the style when I click third indicator?

7CodeWizard:
Hi there? your skill is very good, I wanna ask you sth.
![Screenshot 2024-09-04 100735](https://github.com/user-attachments/assets/68a8395a-e29b-4204-a477-d8ef364cac8d)
I am trying to make this slide indicators, then I can't do that. Can you help me to do this?

PR #28815, opened 2019-05-24
add .carousel-indicator-square and .carousel-indicator-circle

aditya1906:
This is a fix to the issue #28791

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -143,7 +143,9 @@
 // Add an ordered list with the following class and add a list item for each
 // slide your carousel holds.
 
-.carousel-indicators {
+.carousel-indicators,
+.carousel-indicators-square,
+.carousel-indicators-circle {
   position: absolute;
   right: 0;
   bottom: 0;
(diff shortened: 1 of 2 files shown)


morrissey-ingenious:
I don't really like this implementation, I would rather see indicators accept overrides with the `rounded` class

aditya1906:
Do you think instead of using .carousel-indicators-circle we should use .rouned?

forghan89:
> 
> 
> Do you think instead of using .carousel-indicators-circle we should use .rounded? I think indicator-square is fine.

I think `.carousel-indicators-round` is better.

aditya1906:
@MartijnCuppens, @XhmikosR is this PR okay?

XhmikosR:
It's up to @twbs/css-review, but for what is worth, I think it might be useful to others.

ysds:
My concern is to increase CSS even though there are few cases where different styles are used in one application.

I think it may be enough to just document about  customization.

mdo:
Agreed with @ysds, we don’t need these variants in the project. I also don’t think we need to do much there to customize the look and feel since the styles include `width`, `border-radius`, etc. There’s nothing clever here if I recall.

Thanks, but we’ll pass.

PR #28817, opened 2019-05-24
vnu.jar: Update ignores.

XhmikosR:
Pass the `--no-langdetect` flag instead of ignoring the warning manually.

### DIFF
--- a/build/vnu-jar.js
+++ b/build/vnu-jar.js
@@ -14,7 +14,7 @@ const vnu = require('vnu-jar')
 
 childProcess.exec('java -version', (error, stdout, stderr) => {
   if (error) {
-    console.error('Skipping vnu-jar test; Java is missing.')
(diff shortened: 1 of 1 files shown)


build/vnu-jar.js
```diff
@@ -14,7 +14,7 @@ const vnu = require('vnu-jar')
 
 childProcess.exec('java -version', (error, stdout, stderr) => {
   if (error) {
-    console.error('Skipping vnu-jar test; Java is missing.')
```
Johann-S: it seems more appropriate to use `console.error` here because the line before you have `if (error)`

Johann-S (approved):

build/vnu-jar.js
XhmikosR: Yeah, I know, I'm in the middle too. I will probably revert this.

PR #28824, opened 2019-05-25
Add initial FUNDING.yml

XhmikosR:
@mdo: can you enable the repo option?


XhmikosR:
BTW it seems we don't need to backport this as long as it's in the main branch.

MartijnCuppens (approved):

PR #28828, opened 2019-05-26
Utility API improvement: use $key/$value instead of nth

ashfahan:
instead of zipping values together convert list or strings to map and then use them


ashfahan:
@mdo @ysds @MartijnCuppens any reviews over pr . thanks 😅

MartijnCuppens:
@ashanbrown, no need to mention us here, patience is your best friend.

I don't think the new function has a lot of added value.

But I like what you did with the `$key/$value` thing, could you just tackle that in this PR?

ashfahan:
sorry about that @MartijnCuppens . i was just curious . 

about this pr yeah you are right it doesn't really add or remove much from workflow . however its main reason is to increase the readability and better porting for various other things that needs to be converted to maps and around

Thanks for your time.

MartijnCuppens:
The problem is this PR adds about 50 lines to our codebase that 'll need to be maintained by us and doesn't provide that much of a benefit (it's only used in one place).

Let's just keep the `$key/$value` thing you did, that's a small improvement which does make the code easier to understand.

ashfahan:
you got a point 😅 .

Pr updated

scss/mixins/_utilities.scss
```diff
@@ -9,6 +9,8 @@
   }
 
   @each $value in $values {
```
MartijnCuppens: I guess you can just use this here: 
```suggestion
  @each $key, $value in $values {
```

scss/mixins/_utilities.scss
```diff
@@ -21,12 +21,12 @@
...
 
     // Don't prefix if value key is null (eg. with shadow class)
-    $property-class-modifier: if(nth($value, 1), "-" + nth($value, 1), "");
+    $property-class-modifier: if($key, "-" + $key, "");
 
     .#{$property-class + $infix + $property-class-modifier} {
       @each $property in $properties {
         // stylelint-disable-next-line declaration-no-important
-        #{$property}: #{nth($value, 2)} !important;
+        #{$property}: #{$value} !important;
```
MartijnCuppens: Value should not be wrapped in `#{}` here:
```suggestion
        #{$property}: $value !important;
```

MartijnCuppens (approved):
Thanks!

ashfahan:
> Thanks!

It's my pleasure.

PR #28829, opened 2019-05-27
Correct grammar in modal.scss

aditya1906:
Instead of "where should there be a fixed height" I think it should be "where there should be a fixed height".


MartijnCuppens (approved):
Thanks

aditya1906:
@MartijnCuppens you are welcome. I'm looking forward to contribute even more.

PR #28518, opened 2019-03-17
Switch from qunit to jasmine

Johann-S:
Things to do:
- [x] Move each plugin in a separate folder
- [x] Rewrite all the plugins unit tests
	- [x] dom
		- [x] data
		- [x] eventHandler
		- [x] manipulator (#28543)
		- [x] selectorEngine
	- [x] util
		- [x] index
		- [x] sanitizeHtml
	- [x] Alert
	- [x] Button
	- [x] Carousel
	- [x] Collapse
	- [x] Dropdown
	- [x] Modal
	- [x] Popover
	- [x] Scrollspy
	- [x] Tab
	- [x] Toast (#28562)
	- [x] Tooltip

PR #28834, opened 2019-05-27
Fix keyboard handling of button-style checkbox/radio button groups

patrickhlauke:
After various past attempts, this hopefully addresses any outstanding keyboard-handling issues with the "faked" button-style checkbox/radio button groups https://getbootstrap.com/docs/4.3/components/buttons/#checkbox-and-radio-buttons

Closes https://github.com/twbs/bootstrap/issues/28564
Closes https://github.com/twbs/bootstrap/issues/26855
Closes https://github.com/twbs/bootstrap/issues/28847

(pro-tip to test if keyboard is being handled correctly: for the checkbox and radio button buttons, use devtools, find the nested `input`, and disabled the `position:absolute` rule in the relevant CSS ... now you can see exactly if a radio or checkbox gets toggled properly or not)


patrickhlauke:
This PR is specifically targeted at v4, as in v5 this whole approach will be dropped in favor of a CSS-only solution, which will obviate most of the bolted-on functionality of the button.js script

XhmikosR:
@patrickhlauke: it seems BrowserStack tests fail.

patrickhlauke:
> @patrickhlauke: it seems BrowserStack tests fail.

interesting. assume the BrowserStack tests don't run locally, right? (as doing `npm run test` didn't show any problems at the time).

worth noting that the qunit tests relating to sending a `click` to `<label>` elements are incomplete, as they won't trigger the default browser behaviour (which my code now DOESN'T suppress with `preventDefault()`) ... should really set up a whole asyn sending of a synthetic mouse event. but that's beyond my knowledge/capability (and since we'll be abandoning this whole construct for faked button-style checkboxes/radios, i didn't bother going too deep into that)

patrickhlauke:
ok, i give up...can't figure out why the test blows up in qunit in Firefox. testing the same scenario manually in Firefox shows the correct behaviour, so it seems definitely something to do with the way the test harness does something unexpected in Firefox. this could well be tied to the same problem I saw with the other test, where using `triger('click')` doesn't actually trigger the default browser behaviour (see https://github.com/qunitjs/qunit/issues/73) unless we go down some rabbit hole of using something like https://github.com/bitovi/syn for the tests to trigger real events on those `<label>` elements. but this is well beyond my capabilities/knowledge...

...alternatively, which is what i'll try next, the qunit test itself can be modified to avoid reliance on default browser behaviour that comes from a real mouse click on a real `<label>`, and just use another element instead of `<label>`. this is admittedly jury-rigged, but unless somebody else can help out with fundamentally fixing the tests, i can live with this as it does fix the actual long-standing keyboard problem in real life use.

XhmikosR:
No, the Browserstack tests run on Browserstack. Maybe @Johann-S has an idea.

Johann-S:
Instead of using `.trigger('click')` you can simply use `.click()` which trigger the default browser behavior

patrickhlauke:
> Instead of using `.trigger('click')` you can simply use `.click()` which trigger the default browser behavior

ok, going to give that a try, but i'm fairly sure in the past when i tried that it wasn't sufficient as the tests were still happening earlier than the actual default behaviour taking place (so it would need some async or a timeout to wait for the headless browser to do its thing)

patrickhlauke:
still not able to work out why the test fails in Firefox 67. testing this manually shows me the correct behaviour. no idea why the qunit test itself fails. either there's some bug in Firefox that I'm not aware of, or...don't know, to be honest. i will need help trying to work this out (which is frustrating as this appears to correctly fix the keyboard issue at last). @Johann-S any ideas?

patrickhlauke:
This sort of messing is really making me pine for the v5 approach of dumping these faked checkboxes/radio buttons...

patrickhlauke:
good lord...the very persistent problem there in Firefox appears to be a bug https://bugzilla.mozilla.org/show_bug.cgi?id=1540995

patrickhlauke:
Added lots of extra defensive coding and checks to circumvent situations where the Firefox bug would rear its head. It's fixed already in Nightly, but these changes will prevent things blowing up in current, hopefully.

For reference, all this was needed because this PR removes the `preventDefault` (which was shielding us from these bugs, but which also suppressed correct keyboard behaviour).

This will need a bug squash before any attempt to merge...

It could also be ported to v5, but probably not worth it if we're going for the streamlined CSS-only approach in v5 anyway soon...

patrickhlauke:
Uff, seems to finally be ok. So, to recap what this whole PR did (leaving aside the various false starts/attempts/etc):

- adds more defensive checks to make sure no unnecessary toggling happens on disabled buttons; this also fixes an up-to-now undiscovered bug where a toggle button with `.disabled` class would still have its `aria-pressed` toggled
- adds a set of explicit tests for the above case of disabled buttons and `aria-pressed`
- remove a now irrelevant (or at least very nonsensical) test for `<label>` containing both an actionable and a `hidden` `<input>`
- expand the test for disabled checkbox to also explicitly test starting conditions (used mainly in my debugging)
- ensure that `$btn[0].click()` is used to click checkboxes in tests, rather than the `click()` on the jquery object which is simply a shorthand for `trigger('click')` and does not actually trigger the browser default behavior
- remove the `preventDefault()` from the button handling, which was preventing correct keyboard functionality for checkboxes/radio buttons
- add extra logic to the button.js code to handle checkboxes correctly and avoid double-triggering as a result of mouse interactions (which saw the checkboxes being toggled twice, thus returning them to their original state)

patrickhlauke:
@XhmikosR do you agree/think that this should just go to v4, as it will soon be redundant for v5 (assuming the change to CSS-only will happen there)?

patrickhlauke:
> > Instead of using `.trigger('click')` you can simply use `.click()` which trigger the default browser behavior
> 
> ok, going to give that a try, but i'm fairly sure in the past when i tried that it wasn't sufficient as the tests were still happening earlier than the actual default behaviour taking place (so it would need some async or a timeout to wait for the headless browser to do its thing)

btw, just clarifying this for posterity too: the issue was that `click()` on a jquery object really just maps to `trigger('click')`. the trick was to target the actual DOM node with `[0]` and trigger the native `click()` instead.

Johann-S:
It seems some lines you added @patrickhlauke aren't covered by the unit tests, see: https://coveralls.io/builds/23676439/source?filename=js/src/button.js

patrickhlauke:
sigh, seems we're getting more and more stumbling blocks in the way of fixing issues. this coveralls thing is new? not part of any tests we run locally or as part of PRs?

XhmikosR:
Everything can be run locally too. The reason we don't run everything by default locally is that it's slow and in some cases requires authentication.

Coveralls isn't new, it's there for year(s). It shows what part of the code is covered by tests or not.

patrickhlauke:
> Everything can be run locally too

I know it CAN, but it's frustrating that even for somebody like myself who's mildly familiar with the project, who tries to get everything ok locally with the default `npm test` etc, then finds out only at PR stage that things are not fully ok. and then even if the automatic extra tests for PRs all run fine, there's the extra "oh, but coveralls" mention...
if just feels like a stop and start process, whenever you think you finally got it all ship-shape, some new spanner is thrown in the works, which then just drags it on further...

Johann-S:
If you do `npm test` you'll see the coverage too like that:

https://travis-ci.org/twbs/bootstrap/jobs/538848434#L762

image


And you can check the coverage for each files in: `js/coverage/Icov-report` but you have to run `npm test` first

XhmikosR:
CI is here to help us, one just needs to get used to this.

patrickhlauke:
> So you can check everything before opening a PR

should be documented somewhere then...

Johann-S:
You can find information here: https://github.com/twbs/bootstrap/tree/master/js/tests#code-coverage

patrickhlauke:
@Johann-S well after some further refactoring and adding more tests, i've now upped the actual test coverage from what it was originally, but still have some lines not covered. unsure why the `type === "checkbox"` one is showing up as not covered (?), and not sure how to write a unit test that would cover the basic defensive `if (button)` as that simply tries to avoid code not running when there's a `data-toggle="button"` which doesn't contain any actual buttons ... can't think of an assertion to test there...

Johann-S:
Thanks for increasing the coverage here @patrickhlauke 👍 

About the defensive part `if (button) {` I'm in favor of removing this check, because it's important to let know developers when they do something wrong, by doing that they won't know why it doesn't work and defensive code is often a bad idea.

For this check `type === "checkbox"` it seems we don't have a unit test which cover that. But that's ok I rewrote every plugins unit tests here: #28518 and the button plugin is a 100% on this branch see: https://coveralls.io/builds/23704503/source?filename=js/src/button/button.js

So we should at least cover new code added, that's all

patrickhlauke:
> About the defensive part `if (button) {` I'm in favor of removing this check, because it's important to let know developers when they do something wrong, by doing that they won't know why it doesn't work and defensive code is often a bad idea.

Turns out that without it, there were situations where the modified button.js threw up nasty errors, and I pinpointed when that's the case (click sent to outer `data-toggle` container). Can happen in real life if user manages to click in empty space inside the container but not on a button itself. Now covered in test.

> For this check `type === "checkbox"` it seems we don't have a unit test which cover that. I saw you tried to add one, but I don't see why it doesn't cover that case, I think you should debug that and see why it doesn't cover that case. If you want help on that do not hesitate 👍

I think it wasn't so much that the test didn't cover it, but that there was no test where that second `if` statement failed - I assume that those cases are also flagged in test coverage as a "why you checking this if it's always true?". Anyway, as result, I expanded the actual button plugin code with an extra `else` and added relevant tests.

Which now brings the coverage to 100% ... https://coveralls.io/builds/23801480/source?filename=js/src/button.js

patrickhlauke:
btw this plugin code is absolutely atrocious and convoluted, trying to do so many things at the same time. can't wait to ditch the whole "faked checkboxes/radio buttons" stuff for these and move to just CSS-styled radio buttons/checkboxes...

patrickhlauke:
For squashing purposes, here's the revised single big commit message I'd suggest:

- adds more defensive checks to make sure no unnecessary toggling happens on disabled buttons; this also fixes an up-to-now undiscovered bug where a toggle button with `.disabled` class would still have its `aria-pressed` toggled
- adds a set of explicit tests for the above case of disabled buttons and `aria-pressed`
- remove a now irrelevant (or at least very nonsensical) test for `<label>` containing both an actionable and a `hidden` `<input>`
- expand the test for disabled checkbox to also explicitly test starting conditions (used mainly in my debugging)
- ensure that `$btn[0].click()` is used to click checkboxes in tests, rather than the `click()` on the jquery object which is simply a shorthand for `trigger('click')` and does not actually trigger the browser default behavior
- remove the `preventDefault()` from the button handling, which was preventing correct keyboard functionality for checkboxes/radio buttons
- add extra logic to the button.js code to handle checkboxes correctly and avoid double-triggering as a result of mouse interactions (which saw the checkboxes being toggled twice, thus returning them to their original state)
- add logic that prevents the `checked` property from being added incorrectly for any inputs other than radio buttons and checkboxes
- added more tests (including the most basic test for a properly triggered fake checkbox button)
- work around Firefox bug #1540995 (which this code was hitting after removing the `preventDefault()`, due to Firefox's incorrect toggling of disabled checkboxes when programmatically (but not manually) activated with a `click()` event

js/src/button.js
```diff
@@ -140,15 +143,19 @@ class Button {
...
 
     if (!$(button).hasClass(ClassName.BUTTON)) {
-      button = $(button).closest(Selector.BUTTON)
+      button = $(button).closest(Selector.BUTTON)[0]
     }
 
-    Button._jQueryInterface.call($(button), 'toggle')
+    if (!button || button.hasAttribute('disabled') || button.classList.contains('disabled')) {
+      event.preventDefault() // work around Firefox bug #1540995
+    } else if (button.querySelector(Selector.INPUT) && (button.querySelector(Selector.INPUT).hasAttribute('disabled') || button.querySelector(Selector.INPUT).classList.contains('disabled'))) {
```
Johann-S: Here you call 3 times `querySelector` please store the returned value.

Something like:
```js
if (!button || button.hasAttribute('disabled') || button.classList.contains('disabled')) {
  event.preventDefault() // work around Firefox bug #1540995
} else {
  const inputBtn = button.querySelector(Selector.INPUT)

  if (inputBtn && (inputBtn.hasAttribute('disabled') || inputBtn.classList.contains('disabled')) {
    event.preventDefault() // work around Firefox bug #1540995
    return
  }

  Button._jQueryInterface.call($(button), 'toggle')
}

Johann-S:
Can you made the last change I suggested ?

And feel free to squash your commit as you want 👍

patrickhlauke:
sigh...fairy snuff. done. so, want me to squash and force-push? or can this be "Squash and merge"d when it comes time?

XhmikosR:
@patrickhlauke since you know exactly how you want to have the commits, feel free to clean it up locally and force push, otherwise if you just want one commit, you can do it via the web UI when merging.

Johann-S (approved):

XhmikosR:
I changed the target branch to my `v4-dev-xmr` since I have more stuff there for v4.

patrickhlauke:
so, good to squash and merge?

XhmikosR:
Yup

patrickhlauke:
Should I add this to the shiplist, or is that sorted out separately when you do stuff in `v4-dev-xmr` ?

XhmikosR:
We just track the PRs in the v4.3.2 project.

mimenz:
Hi there,
do you have an approximate schedule for the productive delivery of the bugfix?
Soon would be great! ;)
Thanks a lot, bye,
Michael

BTW: great framework!

XhmikosR:
No exact date, but we'll try to release 4.x soon.

PR #28836, opened 2019-05-28
Clean input group flex basis (v4)

ysds:
For v4 of #28785


ysds:
I've confirmed this works on IE10 👍 

スクリーンショット 2019-05-29 1 17 30

XhmikosR:
OK let's close this and move on with #28785

PR #28845, opened 2019-05-29
Formatting in term of the contributing guidelines.

dani832m:


scss/utilities/_sizing.scss
```diff
@@ -1,5 +1,9 @@
 // stylelint-disable declaration-no-important
 
+//
+// Sizing
+//
```
mdo: This is not necessary—please remove.

PR #28874, opened 2019-06-05
Decrease number of order utilities

MartijnCuppens:
The list of number of order utilities is [huge](https://getbootstrap.com/docs/4.3/utilities/flex/#order) and I guess it's ok to decrease this list a bit. I've now limited the list to only 5 (+ first/last) responsive order utilities which should be ok


mdo:
The number was dictated by the column count. Unsure how much we need to really map to that though, so I'd be curious about usage there from people using these utilities.

MartijnCuppens:
I personally never used the order classes with numbers, only the first & last utilities to shift some cols for different screen widths. These 5 (+2) order utilities still gives people to completely reorder at least 7 cols which isn't that common I guess.

mdo:
Yeah, it was a holdover from the old `push`/`pull` classes in v3. We had all 12 for each direction, and `order` saved us at least one set of utilities that could produce the same effect. I could see us cutting this in half to start in v5 and see what folks think during the alpha period.

Can you document this in the migration page too then?

MartijnCuppens:
Done: https://github.com/twbs/bootstrap/pull/28874/commits/132fa64cfe2a22914d7d21aa16c9cd1d18137629#diff-bdd3aaef666a8581309fadd06d14b895

XhmikosR:
TBH the current way makes more sense due to the number of columns.

ysds:
I like this ❤️ 

Why the `last` is not 6?

MartijnCuppens:
> Why the last is not 6?

idk, 6 felt a bit too low so I chose 9. Maybe 6 makes more sense indeed.

MartijnCuppens:
Changed `last` to 6 now

ysds (approved):

MartijnCuppens:
@mdo, just to make sure, is it ok to merge this for you?

MartijnCuppens:
Ping @mdo

mdo (approved):

mdo:
Sorry, and thank you!

PR #28882, opened 2019-06-09
fix custom-select-indicator in IE10

mort3za:
Inline svg of custom-select is broken in IE10 because there is no width and height attributes.
Tested on a real IE10 in browserstack, it works.

Current: https://getbootstrap.com/docs/4.3/components/forms/#select-menu
Fix: https://deploy-preview-28882--twbs-bootstrap.netlify.com/docs/4.3/components/forms/#select-menu


MartijnCuppens:
Hi @mort3za,

Thanks for the fix! We're dropping IE10 support in Bootstrap `v5` but will fix this for the next hotfix for `v4`. Could you target the `v4-dev` branch and rebase?

mort3za:
Yes, I'll

PR #28893, opened 2019-06-11
improves media breakpoint mixin readibility

ashfahan:
I felt that its better to use media-breakpoint-above and media-breakpoint-below rather than media-breakpoint-up and media-breakpoint-down because in plain english the mixin give better idea when reading code along with media-breakpoint-between

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -158,7 +158,7 @@
     margin-bottom: $card-deck-margin;
   }
 
-  @include media-breakpoint-up(sm) {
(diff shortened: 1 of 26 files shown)


ashfahan:
Update : am not sure about updating whole docs. Is there any place left to update in docs?

mdo:
Thanks, but we'll leave this as-is. I don't see much of an improvement to the readability myself, and I haven't seen much of any other feedback on requesting such a change.

PR #28896, opened 2019-06-12
tooltip dispose:removing only own event handler

david-lallement:
If the tooltip component is in a modal, only the listeners created by the tooltip component should to be removed in the dispose method.
For example, if we subscribe to the _**hide.bs.modal**_ event of the **_modal_** component and call the **_dispose_** method of a **_tooltip_** component inside, the _**hide.bs.modal**_ event subscription is lost.

Using of  [event namespace feature of jquery](https://api.jquery.com/event.namespace/) ensures that  only the handler created in _setListeners function is removed in dispose function.

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -237,7 +237,7 @@ class Tooltip {
     Data.removeData(this.element, this.constructor.DATA_KEY)
 
     EventHandler.off(this.element, this.constructor.EVENT_KEY)
-    EventHandler.off(SelectorEngine.closest(this.element, '.modal'), 'hide.bs.modal')
+    EventHandler.off(SelectorEngine.closest(this.element, '.modal'), 'hide.bs.modal.tooltip')
 
     if (this.tip) {
       this.tip.parentNode.removeChild(this.tip)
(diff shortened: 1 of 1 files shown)


Johann-S:
Hi @david-lallement,

Can you create a live demo of the problem via [CodePen](https://codepen.io/)/[JS Bin](https://jsbin.com/) or [Stackblitz](https://stackblitz.com/) ?

Because we already use event namespace of jQuery

david-lallement:
Hi Johann,
a codepen to reproduce the issue : [https://codepen.io/david-lallement/pen/KjwoYx ](https://codepen.io/david-lallement/pen/KjwoYx ).

Johann-S:
Nice catch @david-lallement 👍 

It seems your change broke our unit tests, please can you fix that ?

david-lallement:
Hi Johann,
I fixed the issue in new commit.
The use of jquery event namespace worked in release 4.3.1 but not in the master branch.
I replaced it with a named function.

XhmikosR:
@Johann-S: when you merge this please adapt it for v4 and push it to my v4-dev-xmr branch

Johann-S (approved):

Johann-S:
Now it's fixed thanks @david-lallement 👍 you can see it here: https://codepen.io/Johann-S/pen/zVGmzZ

and it'll be available in our next v4 release

PR #28912, opened 2019-06-17
Don't open dropdown on ESC on trigger element

### LINKED ISSUE #28751: Dropdowns should not open/trigger on "ESC"
This seems to be a regression from 3.x (see https://github.com/twbs/bootstrap/issues/14646)

Currently, dropdowns https://getbootstrap.com/docs/4.3/components/dropdowns/ can be opened by setting focus to the trigger element and pressing the "ESC" key. It's not a major problem, but non-standard and unexpected behaviour, which should really be removed.

patrickhlauke:
Closes #28751

(doesn't need porting to v5, as there the issue doesn't appear to be present anymore)

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -475,6 +475,10 @@ class Dropdown {
     const parent   = Dropdown._getParentFromElement(this)
     const isActive = $(parent).hasClass(ClassName.SHOW)
 
+    if (!isActive && event.which === ESCAPE_KEYCODE) {
+      return
+    }
+
     if (!isActive || isActive && (event.which === ESCAPE_KEYCODE || event.which === SPACE_KEYCODE)) {
       if (event.which === ESCAPE_KEYCODE) {
         const toggle = parent.querySelector(Selector.DATA_TOGGLE)
--- a/js/tests/unit/dropdown.js
+++ b/js/tests/unit/dropdown.js
@@ -67,6 +67,31 @@ $(function () {
     $dropdown.trigger($.Event('click'))
   })
 
+  QUnit.test('should not open dropdown if escape key was pressed on the toggle', function (assert) {
+    assert.expect(1)
+    var done = assert.async()
+    var dropdownHTML = '<div class="tabs">' +
+        '<div class="dropdown">' +
+        '<button disabled href="#" class="btn dropdown-toggle" data-toggle="dropdown">Dropdown</button>' +
+        '<div class="dropdown-menu">' +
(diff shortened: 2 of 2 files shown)


Johann-S:
How is it possible it's not present in v5 ? 😲

js/src/dropdown.js
```diff
@@ -475,6 +475,10 @@ class Dropdown {
     const parent   = Dropdown._getParentFromElement(this)
     const isActive = $(parent).hasClass(ClassName.SHOW)
 
+    if (!isActive && event.which === ESCAPE_KEYCODE) {
```
Johann-S: Can you add the unit test for this check ?

patrickhlauke:
> How is it possible it's not present in v5 ? 😲

no idea. all i know is that when i tried running current v5, it didn't exhibit the behavior (unless i messed up somewhere along the line)

Johann-S:
I'm very surprise the v5 code follow the v4 code, it shouldn't have such difference, is it possible for you to list what changed between v4 and v5 ?

patrickhlauke:
> I'm very surprise the v5 code follow the v4 code, it shouldn't have such difference, is it possible for you to list what changed between v4 and v5 ?

incidentally, i just added the new unit test to the clean master here locally, and ran the tests ... and they all passed. will dig deeper into why v5 is unaffected...

Johann-S (approved):

Johann-S:
Thanks @patrickhlauke 👍 Let me know if you have any questions about the v5 code

patrickhlauke:
> Thanks @patrickhlauke 👍 Let me know if you have any questions about the v5 code

without digging too deeply, one change that seems relevant is that in dropdown.js in v4 on line 484 it triggers a "click" (which happens on keydown for ESC as well) whereas in v5 on the matching line 492 is does `Dropdown._clearMenus()` which, from what I can see, does all sorts of extra sanity checking.

patrickhlauke:
and for reference, the new issue i found in v5 https://github.com/twbs/bootstrap/issues/28914

Johann-S:
Hmm true, nice catch, I don't remember why we made that change 🤔

patrickhlauke:
> Hmm true, nice catch, I don't remember why we made that change 🤔

well, a real button already reacts to all the appropriate things like SPACE and ENTER to open the dropdown (by firing the click event), so unnecessary for that. this leaves keydown reaction when menu is already open, which can really only be cursor keys and esc (plus special cases where somebody stuck an input/textarea in the menu itself) ... so i guess this makes more sense.

PR #28915, opened 2019-06-17
Add "add" and "subtract" func to return valid calc

### LINKED ISSUE #28792: CSS Bug if $card-border-width: 0;
Hey there,

if I set the `$card-border-width` to "0", 
the output will look like this and ends in an error (at least Chrome):

`.card-img { border-radius: calc(0.25rem - 0); }`

As a workaround I also had to change:

`$card-inner-border-radius: calc(#{$card-border-radius} - #{$card-border-width});`

to 

`$card-inner-border-radius: $card-border-radius;`

Maybe it's possible to fix this (small) issue. :) 

Kind regards
André

ysds:
Fixes #28792

Add the two Sass functions (`add()`, `subtract()`) to return a valid `calc`. These functions basically return a calc expression, but if one value is 0 the functions return another value.

e.g.
```scss
div {
  padding: add(1rem, 1px);  // calc(1rem + 1px)
  padding: add(1rem, 0);    // 1rem
}
```

The each functions has the third argument (boolean). If `true`, the expression is returned with wrapped by `calc`, but if `false`, the expression is not wrapped with `calc`. The default is `true`. This option is intended for use in complex nested expressions:

e.g.
```scss
div {
  padding: add(1rem, add(1em, 1px));         // calc(1rem + calc(1em + 1px)) : invalid in IE
  padding: add(1rem, add(1em, 1px, false));  // calc(1rem + 1em + 1px) : valid in IE
}
```

* [x] add calc() to stylelint function-blacklist

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -108,3 +108,40 @@
 @function shade-color($color, $level) {
   @return mix(black, $color, $level * $theme-color-interval);
 }
+
+// Return valid calc
+@function add($value1, $value2, $return-calc: true) {
+  @if $value1 == null {
+    @return $value2;
+  }
+
+  @if $value2 == null {
+    @return $value1;
+  }
+
+  @if type-of($value1) == number and type-of($value2) == number and comparable($value1, $value2) {
+    @return $value1 + $value2;
+  }
+
+  @return if($return-calc == true, calc(#{$value1} + #{$value2}), #{$value1} + #{$value2});
+}
+
+@function subtract($value1, $value2, $return-calc: true) {
+  @if $value1 == null and $value2 == null {
+    @return null;
+  }
+
+  @if $value1 == null {
+    @return -$value2;
+  }
+
+  @if $value2 == null {
+    @return $value1;
+  }
+
+  @if type-of($value1) == number and type-of($value2) == number and comparable($value1, $value2) {
+    @return $value1 - $value2;
+  }
+
+  @return if($return-calc == true, calc(#{$value1} - #{$value2}), #{$value1} - #{$value2});
+}
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -706,8 +706,7 @@ $form-select-indicator:             url("data:image/svg+xml,<svg xmlns='http://w
 
 $form-select-background:            no-repeat right $form-select-padding-x center / $form-select-bg-size !default; // Used so we can have multiple background elements (e.g., arrow and feedback icon)
 
-// stylelint-disable-next-line function-blacklist
-$form-select-feedback-icon-padding-right: calc(#{add((#{add(1em, 2 * $form-select-padding-y, false)}) * 3 / 4, $form-select-padding-x + $form-select-indicator-padding, false)}) !default;
+$form-select-feedback-icon-padding-right: add(1em * .75, (2 * $form-select-padding-y * .75) + $form-select-padding-x + $form-select-indicator-padding) !default;
 $form-select-feedback-icon-position:      center right ($form-select-padding-x + $form-select-indicator-padding) !default;
 $form-select-feedback-icon-size:          $input-height-inner-half $input-height-inner-half !default;
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -616,13 +616,13 @@ $input-plaintext-color:                 $body-color !default;
 
 $input-height-border:                   $input-border-width * 2 !default;
 
-$input-height-inner:                    calc(#{$input-line-height * 1em} + #{$input-padding-y * 2}) !default;
-$input-height-inner-half:               calc(#{$input-line-height * .5em} + #{$input-padding-y}) !default;
-$input-height-inner-quarter:            calc(#{$input-line-height * .25em} + #{$input-padding-y / 2}) !default;
+$input-height-inner:                    add($input-line-height * 1em, $input-padding-y * 2) !default;
+$input-height-inner-half:               add($input-line-height * .5em, $input-padding-y) !default;
+$input-height-inner-quarter:            add($input-line-height * .25em, $input-padding-y / 2) !default;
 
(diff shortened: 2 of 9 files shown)


MartijnCuppens:
We might need to rename this to `subtract()` & `add()` (or even `sum()`) if we continue with this.

https://github.com/postcss/postcss-calc also fixes this issue, but I'm not sure if we should another dependency since this doesn't fix anything by default.

ysds:
> We might need to rename this to subtract() & add() (or even sum()) if we continue with this.

okay 👍 

> https://github.com/postcss/postcss-calc also fixes this issue, but I'm not sure if we should another dependency since this doesn't fix anything by default.

Using postcss-calc will add burden to users to add the plugin to them projects.

MartijnCuppens:
> Using postcss-calc will add burden to users to add the plugin to them projects.

Yeah, I know. It's just the easiest way to handle this.

I'm ok with the additional functions, but maybe we should add something to our docs about this?

ysds:
@mdo @MartijnCuppens  How dow you feel about this concept?  I want to know before release v4.4 and resolve conflicts 😉

MartijnCuppens:
Not really a fan of this, but I guess this is the only way to avoid issues with `calc()` and `0` in Sass. If we continue we this we should document why we have these functions. I would also check for `null`s, not only `0`.

ysds:
I updated the PR:

* The `add`/`subtract` functions now pass `null` and unitless value not only 0
* Add documents: https://deploy-preview-28915--twbs-bootstrap.netlify.com/docs/4.3/getting-started/theming/#add-and-subtruct-function
* Add stylelint config : `function-blacklist: ["calc"]`

test cases on sassmeisater: https://www.sassmeister.com/gist/19551dac442677ecf10cca73f3f6598e

This is ready for review @twbs/css-review :)

XhmikosR:
Saves 0.02 KB from bootstrap.min.css BTW 😄

scss/_functions.scss
```diff
@@ -100,3 +100,44 @@
 
   @return mix($color-base, $color, $level * $theme-color-interval);
 }
+
+// Return valid calc
+@function add($value1, $value2, $return-calc: true) {
+  @if $value1 == null and $value2 == null {
```
MartijnCuppens: This test is redundant, it's covered by the following 2 tests.

scss/_functions.scss
```diff
@@ -100,3 +100,44 @@
...
+
+  @if $value1 == null {
+    @return $value2;
+  }
+
+  @if $value2 == null {
+    @return $value1;
+  }
+
+  @if type-of($value1) == number and type-of($value2) == number and comparable($value1, $value2) {
```
MartijnCuppens: I guess only `comparable()` is good enough here

XhmikosR:
In order to make the backport easier, if this lands, make sure it's squashed.

scss/_functions.scss
ysds: The `type-of($value1) == number and type-of($value2) == number` is necessary to pass nested cases such as:

```scss
.el {
  padding: add($rem, add($rem, $px));
  padding: add($rem, add($rem, $px, false));
}
```

If the condition is missing, a compile error will occur:

```
argument `$number-2` of `comparable($number-1, $number-2)` must be a number ...
```

XhmikosR:
Just to verify, could this be a breaking change for v4?

XhmikosR:
Hmm, this fails Stylelint, can someone have a closer look?

scss/_variables.scss
```diff
@@ -603,7 +603,8 @@ $form-select-indicator:             url("data:image/svg+xml,<svg xmlns='http://w
 
 $form-select-background:            no-repeat right $form-select-padding-x center / $form-select-bg-size !default; // Used so we can have multiple background elements (e.g., arrow and feedback icon)
 
-$form-select-feedback-icon-padding-right: calc((1em + #{2 * $form-select-padding-y}) * 3 / 4 + #{$form-select-padding-x + $form-select-indicator-padding}) !default;
+// stylelint-disable-next-line function-blacklist
```
XhmikosR: Which function is that we have blacklisted that we need to ignore?

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -215,6 +215,44 @@ You can also specify a base color with our color map functions:
 
 We use the `escape-svg` function to escape the `<`, `>` and `#` characters for SVG background images. These characters need to be escaped to properly render the background images in IE.
 
+## Add and Subtract function
```
mdo:
```suggestion
## Add and Subtract functions
```

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -215,6 +215,44 @@ You can also specify a base color with our color map functions:
...
+
+We use the `add` and `subtract` functions instead of the CSS `calc` function. The primary purpose of these functions is to avoid errors when "unitless" 0 is given to the `calc` expression.
```
mdo:
```suggestion
We use the `add` and `subtract` functions to wrap the CSS `calc` function. The primary purpose of these functions is to avoid errors when a "unitless" `0` value is passed into a `calc` expression. Expressions like `calc(10px - 0)` will return an error in all browsers, despite being mathematically correct.
```

mdo (commented):
Couple ideas for how to clarify the docs here.

XhmikosR:
@twbs/css-review so should we do with this?

MartijnCuppens (approved):

XhmikosR:
This now needs a rebase. Please be careful. And we'll need to cherry pick it in v4-dev, not sure how clean that will be.

XhmikosR:
@MartijnCuppens so only thing left is to blacklist calc?

MartijnCuppens:
Added `calc()` to blacklist & fixed a leftover `calc()` in the docs.

XhmikosR:
@MartijnCuppens this needs to be backported manually (sort of)

Issue #23307, opened 2017-08-11
input[type=date] with empty value shrunken height on mobile


rixlabs:
On iPad, an input of type date that doesn't have any initial value appear smaller until touched or a date selected.

I tested the same situation on Chrome for Android and there is working fine

Reduced test case here: [http://jsfiddle.net/ctesyxxx/1/](http://jsfiddle.net/ctesyxxx/1/)
Tested on: Safari for iPad , iOS 10.1.1

ianbrandt:
@rixlabs Same here.  Did you find a workaround by chance?

rixlabs:
@ianbrandt I did exactly the same...

ianbrandt:
Updated my workaround to also account for https://github.com/twbs/bootstrap/issues/18842:

```
input[type="text"],
input[type="email"],
input[type="date"] {
	height: 2.375rem;
}
```

So far tested on Safari and Chrome on iOS and macOS, and Edge on Windows 10.

(In case it's not obvious you'll want to add any other input types you use, e.g. "password", "url", etc.)

jckh1234:
I found it not only shorten the but also move the input slightly lower when placing inside a div.

https://stackoverflow.com/questions/49932842/html5-input-type-date-not-vertical-align-property-in-mobile-browser-when-value-i

parisiam:
I encountered exactly the same problem in ios 11 and bootstrap 4.1.2: the temporal fields shrink if no initial value is selected.
[https://jsfiddle.net/parisiam/0teuo9dy/50/](https://jsfiddle.net/parisiam/0teuo9dy/50/)
Another solution is to combine the field inside an input group with an icon (a calendar for instance). 
![capture d ecran 2018-07-14 a 15 38 51](https://user-images.githubusercontent.com/627813/42724958-3985577e-877c-11e8-8c5b-66f4a07ea4a5.png)

pekanct:
My solution was:
```
<input type="text "placeholder="Select a date" onfocus="$(this).hide().next().attr('hidden',false).focus();"/>
<input type="datetime-local"  hidden="true">
```

mdo:
This should’ve been solved in v4.1.3 when we added fixed heights to inputs.

floppyTop:
[...]

PR #28917, opened 2019-06-17
Reboot updates

MartijnCuppens:
- ~Inherit box-sizing and set `box-sizing: border-box;` on body
  This makes it possible for 3th party components to use `box-sizing: content-box` by just applying the box-sizing to the wrapper div.~ Nope: https://github.com/twbs/bootstrap/pull/28917#issuecomment-512039100
- Remove unsupported browsers fixes (IE10 & Firefox 39-)
- More consistent comment blocks
- Reorganise order of some blocks
- Add height fix for date/month/week/time inputs (#18842, this was already fixed by adding a height to form controls, with this change we don't need to set the height anymore)
- Remove redundant type selectors for some pseudo elements
- Additional resets and theming options for `legend`
- Element theming from `_code.scss` moved to reboot
- All input fields now have the same borders

**Todo:**
- [ ] Merge code page with reboot page

Result: https://deploy-preview-28917--twbs-bootstrap.netlify.com/docs/4.3/content/reboot/


scss/_reboot.scss
```diff
@@ -318,15 +316,17 @@ th {
 // Forms
 //
```
ysds: Left this comment style on purpose? or might be better to change:

```diff
- //
  // Forms
- //
```

scss/_reboot.scss
```diff
@@ -373,147 +381,159 @@ select {
...
-// 1. Correct the text wrapping in Edge and IE.
-// 2. Correct the color inheritance from `fieldset` elements in IE.
+
+// 1. Correct the color inheritance from `fieldset` elements in IE.
+// 2. Correct the text wrapping in Edge and IE.
+
 legend {
   display: block;
   width: 100%;
-  max-width: 100%; // 1
```
ysds: I found a issue: 
https://jsbin.com/wobotesige/edit?html,css,output

I think we should set `display: table` and `max-width: 100%;`.
Please see https://github.com/necolas/normalize.css/blob/8.0.1/normalize.css#L243-L244

scss/_reboot.scss
```diff
@@ -373,147 +381,159 @@ select {
...
   }
 }
 
+
+// Fix height of inputs with a type of datetime-local, date, month, week, or time
+// See https://github.com/twbs/bootstrap/issues/18842
+
+::-webkit-datetime-edit {
+  line-height: 1;
+}
```
ysds: I don't understand this point. Could you teach me more in detail about this?

MartijnCuppens: See https://codepen.io/MartijnCuppens/pen/MMywaL

I did this so we might might switch from `height` to `min-height` for form controls, so we don't have height issues anymore when the form control is within a flex container. We could also remove this:

https://github.com/twbs/bootstrap/blob/ceffc3a5fd7af7609150048f0651f025ef19672c/scss/_forms.scss#L153-L163

Was planning to investigate that when this is merged, I might also just add this in another PR

scss/_reboot.scss
MartijnCuppens: I see, `display: table` looks a bit better indeed.

Or we might just move the legend inside the fieldset? See https://jsbin.com/caqijiqiyo/1/edit?html,css,output

scss/_reboot.scss
MartijnCuppens: Woops, fixed it.

scss/_reboot.scss
ysds: If we support the aim of ​​Normalize.css:

> Preserve useful browser defaults rather than erasing them.
http://nicolasgallagher.com/about-normalize-css/

that position should be left as it is.

scss/_reboot.scss
ysds: Looks good on the desktop and Android browsers. but I found the #23307 is still reproduced (but not relative on this PR).

scss/_reboot.scss
MartijnCuppens: Nope, we don't, reboot adds some opinionated styling (https://getbootstrap.com/docs/4.3/content/reboot/): 
> Reboot builds upon Normalize, providing many HTML elements with somewhat opinionated styles using only element selectors

I think moving the legend inside the fieldset (last example https://jsbin.com/caqijiqiyo/1/edit?html,css,output) will simplify development for custom theming.

But I don't really have a strong opinion on this. @mdo, what do you think?

scss/_reboot.scss
```diff
@@ -10,37 +10,36 @@
...
-// Change from `box-sizing: content-box` so that `width` is not affected by `padding` or `border`.
+// Inherit box-sizing from parents.
+
 *,
 *::before,
 *::after {
-  box-sizing: border-box;
-}
-
-// Shim for IE11
```
coliff: Do we want to keep this for IE11?

MartijnCuppens: This was just moved to the bottom, see:
https://github.com/twbs/bootstrap/blob/d81ab18f3f6f704f8146132e99cd298a0c9327f5/scss/_reboot.scss#L523-L531

scss/_reboot.scss
mdo: @MartijnCuppens Yup, this aims to be much more opinionated than Normalize. I could be down with the `float` method here to further simplify things, but I'm wondering if that doesn't just sign us up for more resetting when folks do other things with `<legend>`s?

scss/_reboot.scss
MartijnCuppens: I removed this in favor of https://github.com/twbs/bootstrap/blob/cda30cc9950949e9e85320089551b83380f1f22a/scss/_reboot.scss#L516-L522

The line height fix didn't seem to work in all scenarios

scss/_reboot.scss
MartijnCuppens: Since Bootstrap removes the borders by default and tries to undo the typical behaviour of legends, I've added the `float: left`. Now the fieldsets and legend both behave like block elements and can easily be themed with utilities if needed.

mdo:
Regarding the box-sizing change: https://github.com/twbs/bootstrap/issues/22872

mdo:
Last thing I've noticed here is that if we're moving all those styles to Reboot, we need to move the docs content to the Reboot page as well.

MartijnCuppens:
I'm not sure what to do with the `.pre-scrollable` class. It's the only class left in `_code.scss` and I can't document it on the reboot page since it isn't part of it. Should we leave it as only class on the code page and in `_code.scss` or is it not worth it and can we drop it?

ysds (approved):

ysds:
I agree to remove the `.pre-scrollable`. I'm OK remove it in this PR, but I prefer to separate for clarity.

MartijnCuppens:
I've opened https://github.com/twbs/bootstrap/pull/29135/ to drop the class.

mdo (approved):

PR #28920, opened 2019-06-18
Typo fix

MartijnCuppens:
@XhmikosR, this page lost its git history since the Hugo switch, so probably we'll need a new PR for `v4`?


XhmikosR:
We'll see how it goes. Git is generally pretty smart.

XhmikosR (approved):

Issue #28921, opened 2019-06-18
Bootstrap Carousel Sliding outside only in CSS Columns structured


dvinodeluru:
Before opening:

Searched for related issue causing with CSS columns. This is a unique issue no one could find it.
Issue observed after updating the Chrome browser 74 to 75

Before Update Chrome version 74
![image](https://user-images.githubusercontent.com/2167566/59663014-c15d0c00-91cb-11e9-8d4b-51f91be667dc.png)

After Update Chrome version 75
![image](https://user-images.githubusercontent.com/2167566/59663075-dc2f8080-91cb-11e9-84a1-f9e34909f293.png)
![carousel-csscolumns-bootply](https://user-images.githubusercontent.com/2167566/59661089-b0120080-91c7-11e9-8309-3cf3301021f5.png)

Bug reports must include:
- Operating system and version (Windows)
- Browser and version (Chrome 75)
- Bootstrap Version 3.3.7, jQuery v2.2.4 and also above v3.3
- Added code and CSS snippents at https://www.bootply.com/LNYf6MbnOj

No future plans for updating the bootstrap version.

Production 
![image](https://user-images.githubusercontent.com/2167566/59661124-bbfdc280-91c7-11e9-924e-69a4d9d0504e.png)

Same issue with Bootstrap version 4.3.1
https://www.bootply.com/kIkBOZi7bH

MartijnCuppens:
The card columns are known to cause some rendering issues and will probably be ditched in `v5`. 

You can use masonry to tackle your problem: https://deploy-preview-28922--twbs-bootstrap.netlify.com/docs/4.3/components/card/#card-columns-masonry-layout

dvinodeluru:
I'm using the 3.3.7 and custom column structure and future update plans to 4.3 is not at confirmed. Updating the Masonry on the Dynamic cards are not working, hence used the css column standers approach to full fill the requirements. 

So even removing the CSS Column from Upcoming Bootstrap version, the issue still exists with Chrome browser. I reported the bug to chrome forum. and awaiting for the response.

mdo:
Thanks for reporting! Closing as a won't/can't fix :).

Issue #28891, opened 2019-06-11
card-columns and google maps issue


cjshaw1976:
Good day,

There is a silly bug when you insert a google map in a card div, inside a card-columns div and it seems to be something to do with the column count property. There are 4 div's in my case, 2 below and 2 above the map that prevent any other element being clicked. In my example below, the update address button cannot be clicked.
 
<img width="465" alt="Screen Shot 2019-06-11 at 09 18 36" src="https://user-images.githubusercontent.com/13312104/59251495-3bb8e980-8c2a-11e9-86e6-f1c4c3a99a01.png">

I have tried playing around with the z-index of the button, but this does not work.

florianlacreuse:
Can you provide a live demo with a reduced test case?

cjshaw1976:
Will work on it now

cjshaw1976:
I have created two fiddles. The only difference is one uses 'card-deck' and the other 'card-columns'.

With card-columns, you cannot click the button as a hidden part of the map is in the way. In fact you can even drag the map from over the button: https://jsfiddle.net/cjshaw/ws46qbv1/

With card-deck, you can click the button and everything works as you would expect: https://jsfiddle.net/cjshaw/ws46qbv1/1/

mdo:
Card columns use the `columns` CSS property. If you need columns and have stuff like Google Maps to place within, I'd suggest the grid versus a browser-based content layout mechanism. Content is distributed so oddly across CSS columns and I'm betting they have no idea how to handle these kind of objects within.

Food for thought on eliminating the `.card-columns` class outright in v5 given their confusion and problems we've seen, @twbs/css-review.

morrissey-ingenious:
I don't think its a Bootstrap issue ... it only happens in Google Chrome.

mdo:
Closing as a won’t/can’t fix.

Issue #22136, opened 2017-03-08
Introduce card-columns-count as class


punkkeks:
I'd like to specify the number of columns which a .card-columns class has by using a class, e.g.
card-columns-{2,3,4}.

This would give the user the opportunity to specify the layout even further. The css would be something like this for the correspoding classes.

    .card-columns-2 {
        -webkit-column-count: 2;
        -moz-column-count: 2;
        column-count: 2;
    }

Maybe even with the sm, md, lg, xl prefix: card-columns-{sm,md,lg,xl}-{2,3,4}, so you could resolve the issue, that card decks for example aren't responsive currently, but this would lead to responsiveness. I'm currently using the sm prefix for my project to distinct between one and two columns. Works like a charm :)

mdo:
For that level of control, I imagine the grid is what you want. I don't think we'll implement this much control over card column layout at this time. Thanks though!

Issue #20905, opened 2016-10-16
card-columns should be responsive


rocifier:
Can card-columns have responsive variants? I can't set the number of columns on different device widths without doing custom media queries.

mdo:
Maybe in the future, but not for v4.0.

mdo:
Closing as dupe of #20321.

shivampaw:
I think this needs to be done as a priority. It defeats the point of cards and it ruins the responsiveness.

rocifier:
FYI this isn't a dupe of #20321 - that feature request is about card-deck this one is about card-columns. Anyway, I just question the usefulness of releasing 3 hard-coded columns in final production v4 - I mean, why 3? and for all devices?

ayoubkhan558:
Try This 
https://codepen.io/migli/pen/OQVRMw

PR #28922, opened 2019-06-18
Remove the card columns in favor of a Masonry grid

### LINKED ISSUE #28076: Masonry-ish grid
Follow-up on https://twitter.com/getbootstrap/status/1083797119742267392

Maybe we should just add a masonry-ish css grid? We currently only have the [card columns](https://getbootstrap.com/docs/4.2/components/card/#card-columns), but we might just provide a card independant implementation with responsive classes?

### LINKED ISSUE #28439: .card-columns causes svg elements inside to not be visible upon resize in Chrome
We are seeing a problem rendering highcharts (svg) within .cards, within .card-columns, and I have isolated the exact cause. This worked fine in 4.0.0, but breaks in higher versions.

Here is a pen showing it working fine in 4.0.0, before the responsive .card-columns was in play. Look at the Capacity chart.
If you resize the html area (Drag the vertical to the right) when the cards reshuffle, the blue bar area is visible and works nicely.
https://codepen.io/anon/pen/LaLzZb

Here is a pen showing it not working in 4.3.3

https://codepen.io/vinjosh/pen/OqWRMZ

The culprit is the column-count property, that is applying to the svg element here:

```
`@media (min-width: 576px)
.card-columns {
    -webkit-column-count: 3;
    -moz-column-count: 3;
    column-count: 3;`
```

If you take out the column-count, the columns shuffle, and the blue is visible. I think the bleed effect of the column count is affecting the svg element. 

The delta between 4.0 and 4.3 is the addition of the responsive .card-columns breakpoints. I would tend to think that this is a fairly common thing, not specific to highcharts, they are just svg within cards.

MartijnCuppens:
Mentioned before in https://github.com/twbs/bootstrap/issues/28891#issuecomment-500917538 and https://github.com/twbs/bootstrap/issues/28076#issuecomment-488386892, we should ditch the card columns in favor of masonry.

Main reasons are:

- Cards are ordered per column instead of per row (https://github.com/twbs/bootstrap/issues/17709, https://github.com/twbs/bootstrap/issues/17805, https://github.com/twbs/bootstrap/issues/17882)
- Several rendering issues (https://github.com/twbs/bootstrap/issues/19060, https://github.com/twbs/bootstrap/issues/19600, https://github.com/twbs/bootstrap/issues/19725, https://github.com/twbs/bootstrap/issues/20161, https://github.com/twbs/bootstrap/issues/28921)
- No control over responsive number of columns (https://github.com/twbs/bootstrap/issues/17511, https://github.com/twbs/bootstrap/issues/22136,  https://github.com/twbs/bootstrap/issues/20321, https://github.com/twbs/bootstrap/issues/20905)

I've updated the docs and included Masonry via cdnjs, @XhmikosR is that ok for you, or do you have another proposal? (I also watched the releases of masonry github repo so I know when to upgrade) 

**Edit:**
The demo is a example page now. It's also possible to add additional parameters (`async` & `intergrity`) to the extra js.

TODO:
- [x] Link to this PR in migration section
- [x] Move documentation to examples
- [x] Allow to pass `async` attribute
- [x] Check if integrity attribute is present before printing

Closes #28076, closes #28439

Card docs: https://deploy-preview-28922--twbs-bootstrap.netlify.com/docs/4.3/components/card/#card-columns-masonry-layout
Example page: https://deploy-preview-28922--twbs-bootstrap.netlify.com/docs/4.3/examples/#integrations
Demo: https://deploy-preview-28922--twbs-bootstrap.netlify.com/docs/4.3/examples/masonry/


ysds (dismissed):
LGTM 👍, and @mdo?

mdo:
Thoughts on punting the inclusion of external JS to an example instead of a core component page? We can link to it from the Card docs, but I'd prefer to build something there with Masonry instead of adding it to the core docs.

mdo:
But also, yes, definitely down to ditch the card columns lol.

MartijnCuppens:
Examples might be a better place indeed, I'll change it

site/layouts/_default/docs.html
```diff
@@ -30,5 +30,9 @@ <h1 class="bd-title" id="content">{{ .Title | markdownify }}</h1>
     </div>
 
     {{ partial "scripts" . }}
+
+    {{ range .Page.Params.external_js }}
+      <script src="{{ .src }}" integrity="{{ .integrity }}" crossorigin="anonymous"></script>
```
XhmikosR: I'd check for `.integrity` in case we don't want a script to have it. (with `{{ with .integrity }} integrity="{{ . }}"{{ end }}` or something similar if this doesn't work due to the scope of `.`).

site/content/docs/4.3/components/card.md
```diff
@@ -4,6 +4,9 @@ title: Cards
 description: Bootstrap's cards provide a flexible and extensible content container with multiple variants and options.
 group: components
 toc: true
+external_js:
+  - src: "https://cdnjs.cloudflare.com/ajax/libs/masonry/4.2.2/masonry.pkgd.min.js"
```
XhmikosR: perhaps allow passing other attributes like `async` might be good, like in this case.

inwardmovement:
Maybe we should simply use native css subgrids to prevent having to use a "card-deck" container:
[Hello Subgrid! – Rachel Andrew at CSSconf EU 2019](https://youtu.be/vxOj7CaWiPU?t=1172)

Not supported everywhere (anywhere?) but should be since next Bootstrap version I guess.

MartijnCuppens:
@inwardmovement, we'll probably introduce grid in Bootstrap 6 (the next version(5) will still have IE support), but the masonry layout still differs from (sub)gird.

site/layouts/_default/docs.html
MartijnCuppens: @XhmikosR, could you check if this is the way to go, not sure how the scope of variables works in Hugo:
https://github.com/twbs/bootstrap/blob/6cc2142f4b69e0c1609325e189cff3919b96f43c/site/layouts/_default/examples.html#L57-L61

XhmikosR: When you are done moving this to an example, I will be able to try it and let you know.

site/content/docs/4.3/components/card.md
```diff
@@ -601,88 +604,94 @@ Just like with card groups, card footers in decks will automatically line up.
 </div>
 {{< /example >}}
 
-### Card columns
+### Card columns (Masonry layout)
 
-Cards can be organized into [Masonry](https://masonry.desandro.com/)-like columns with just CSS by wrapping them in `.card-columns`. Cards are built with CSS `column` properties instead of flexbox for easier alignment. Cards are ordered from top to bottom and left to right.
+In `v4` we used a CSS-only technique to mimic the behaviour of [Masonry](https://masonry.desandro.com/)-like columns, but this technique came with lots of unpleasant [side effects](https://github.com/twbs/bootstrap/pull/28922) so in `v5` we decided to just use the wonderful Masonry plugin. **Masonry is not included in Bootstrap**, but can easily be added by adding the following javascript file:
```
midzer: I would end the sentence after "side effects", it becomes too long:

...came with lots of unpleasant side effects. So in `v5`...

site/content/docs/4.3/components/card.md
midzer: I think we still have to add `async: true` to card.md for full effect in example layout

site/content/docs/4.3/components/card.md
```diff
@@ -601,88 +604,94 @@ Just like with card groups, card footers in decks will automatically line up.
...
+```js
+<script src="https://cdnjs.cloudflare.com/ajax/libs/masonry/4.2.2/masonry.pkgd.min.js" integrity="sha256-Nn1q/fx0H7SNLZMQ5Hw5JLaTRZp0yILA/FRexe19VdI=" crossorigin="anonymous"></script>
```
midzer: It would be good practice to add `async` attribute here, too.

Furthermore we could give a hint to host file locally as an alternative to CDN.

site/content/docs/4.3/components/card.md
MartijnCuppens: Thanks, I've edited the copy.

MartijnCuppens:
The masonry example is now moved to the examples section.

site/content/docs/4.3/examples/masonry/index.md
```diff
@@ -0,0 +1,101 @@
...
+active_menu: "example"
+extra_js:
+  - src: "https://cdnjs.cloudflare.com/ajax/libs/masonry/4.2.2/masonry.pkgd.min.js"
+    integrity: "sha256-Nn1q/fx0H7SNLZMQ5Hw5JLaTRZp0yILA/FRexe19VdI="
+    async: true
+---
+
+Masonry is not included in Bootstrap, but can easily be added by including your locally hosted file or using the following hosted javascript file:
+```js
+<script src="https://cdnjs.cloudflare.com/ajax/libs/masonry/4.2.2/masonry.pkgd.min.js" integrity="sha256-Nn1q/fx0H7SNLZMQ5Hw5JLaTRZp0yILA/FRexe19VdI=" crossorigin="anonymous" async></script>
```
midzer: I realize we're using CloudFlare CDN for every extra JS which specifies an exact version right now. Thinking about maintainance effort:
with an unspecific version naming, we don't have to update the docs everytime a new version is out, see https://masonry.desandro.com/ CDN example

site/layouts/_default/examples.html
```diff
@@ -42,16 +42,6 @@
...
-        <script src="/docs/{{ .Site.Params.docs_version }}/dist/js/bootstrap.bundle.min.js" integrity="{{ .Site.Params.cdn.js_bundle_hash }}" crossorigin="anonymous"></script>
-      {{ else }}
-        <script src="/docs/{{ .Site.Params.docs_version }}/dist/js/bootstrap.bundle.js"></script>
-      {{ end }}
-
-      {{ range .Page.Params.extra_js }}
-        <script src="{{ . }}"></script>
-      {{- end }}
-    {{ end }}
+    {{ partial "scripts" . }}
```
XhmikosR: We should probably revert any layout changes and use a separate layout for the new example like we do with everything else.

site/content/docs/4.3/examples/masonry/index.md
XhmikosR: We do the same anyway in another example, with hardcoded versions, so I'd say it's fine for now.

Ideally, maybe we should move all this in a data file or something so that it's not so deeply hidden.

site/layouts/_default/examples.html
MartijnCuppens: I did this so that it would be possible for all layouts to add `extra_js`. Are you suggesting to duplicate the `single` layout specifically for examples with a docs-look?

XhmikosR: The Masonry example should use the examples layout like the other examples I think.

mdo:
Agreed with @XhmikosR about using the existing Examples layout. That's a little tougher for some of the needs here, but this does feel slightly over-engineered right now. We may end up adding third party JS to more examples, but I don't know when that'd be. Until then, these examples should look and feel isolated from our docs.

Here's what I'm leaning towards after taking a crack at this branch (see latest commit):

Screen Shot 2019-07-17 at 2 40 45 PM

XhmikosR:
I'm gonna fix the conflict later but we should remove the copy example because it requires our docs CSS and JS.

mdo (approved):

XhmikosR:
This needs some more work and especially the part with the hardcoded script is something that I personally don't like. Maybe we should link to their docs?

TinajaLabs:
Somehow I missed all this and used .card-columns.  Now wanting to go from 4.6 to 5.3 and yuck, .card-columns is broken.  My view is x number of columns wide, starting at the upper left going down each column as many as are needed for more or less evenly distributed column counts.

There's no getting around the efficiency of having all the cards/images stacked in columns when the images are the same width but different heights.  No gaps or wasted screen space; only at the bottom.

Does this mean there is no longer a mechanism for columnar lists?

It seems in masonry, the best alternative is horizontalOrder: true  and within a few rows down is impossible to follow in any reasonable visual order.  If anyone has tips...  Thanks.

julien-deramond:
IMO you'll have more chance of getting an answer by opening a [new Discussion](https://github.com/twbs/bootstrap/discussions/new?category=q-a).

igorsantos07:
I'm migrating from 4 to 5 now, and after some research, I realized that `columns: XXrem 2` was enough for my scenario, together with two breakpoint changes in `column-count` and `.card-columns .card { break-inside: avoid }`.

Was there a reason for not mentioning the underlying CSS techniques used for the v4 element, or maybe some tutorial that describes them, instead of sending the user into plugins? I understand Masonry solves _many more_ problems, but simpler use cases can be fixed with two CSS lines, more or less :)

PR #28927, opened 2019-06-18
WIP make dynamic tabs follow ARIA 1.1 practices

### LINKED ISSUE #28918: make dynamic tabs actually ARIA 1.1 practices compliant
A breaking change, but would like to squeeze this into v5 if we can: turn dynamic tabs into actual "Tabs with automatic activation" (see https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html and https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)

- only the currently selected/active tab receives focus, all other tabs are not focusable (`tabindex="-1"` added dynamically to them)
- cursor keys cycle between tabs, setting both the `tabindex="0"` and `aria-selected="true"` to the newly active tab (and `tabindex="-1"` / `aria-selected="false"` on the now not selected one)
- mouse behavior can remain as now (though need to flip `tabindex` accordingly)
- document this in the docs, referencing specifically the "automatic activation" route (so authors aren't confused/don't wonder why using `SPACE` or whatever does nothing)

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/28918


patrickhlauke:
WIP only at this stage, as i'd need some help with two particular aspects:

- how to properly run the initialisation code on page load (which sanitizes any tablists on the current page, to ensure they have correct tabindex values, and that one item is always the active/selected one)
- how to actually call the tab activation from inside the keyboard handler - currently doing naive `click()` but there must be a better way (particularly since it seems if moving too quickly between tabs the script doesn't fire correctly and you may end up with two tab panels showing at the same time)

@Johann-S or others...any pointers? (and yes, no unit tests yet...they'll come)

patrickhlauke:
https://deploy-preview-28927--twbs-bootstrap.netlify.com/docs/4.3/components/navs/#javascript-behavior

- only one (the selected/active) tab control can be focused using keyboard
- cursor keys left/right change tabs
- tabs auto-activate (following the "Tabs with automatic activation" https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html and https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel / pattern)
- `aria-orientation="vertical"` uses cursor up/down
- removed the code for dropdowns inside tabs, as it's not a sensible/supported pattern for ARIA and has issues, so might as well remove it...this is arguably a breaking change (though from previous question to devs, nobody seemed to really use it https://github.com/twbs/bootstrap/issues/23035)

js/src/tab.js
```diff
@@ -229,6 +217,48 @@ class Tab {
     })
   }
 
+  static _dataApiKeydownHandler(event) {
+    const tablist = SelectorEngine.closest(event.target, Selector.TABLIST)
+    let tablistorientation = tablist.getAttribute('aria-orientation')
```
Johann-S: use `const` here and maybe a bit of camel case, for example: `const tabListOrientation`

js/src/tab.js
```diff
@@ -229,6 +217,48 @@ class Tab {
...
+    if (tablistorientation !== 'vertical') {
```
Johann-S: We should store tabs orientation in something like:
```
const Orientation = {
	VERTICAL: 'vertical',
	HORIZONTAL: 'horizontal'
}

File: js/src/tab.js [hunk 3]
```diff
@@ -240,6 +270,36 @@ class Tab {
  * ------------------------------------------------------------------------
  */
 
+EventHandler.on(window, Event.LOAD_DATA_API, () => {
+  const tablists = makeArray(SelectorEngine.find(Selector.TABLIST))
+  if (tablists.length === 0) {
```
Johann-S: `if (!tablists.length)`

js/src/tab.js
```diff
@@ -240,6 +270,36 @@ class Tab {
...
+    return
+  }
+
+  // iterate over all found sets of tab lists
+  for (let i = 0; i < tablists.length; i++) {
+    const tabs = makeArray(SelectorEngine.find(Selector.DATA_TOGGLE, tablists[i]))
+    let selectedTabFound = false
+
+    // iterate over each tab in the tablist, make sure they have correct tabindex/aria-selected
+    for (let j = 0; j < tabs.length; j++) {
```
Johann-S: just a suggestion but you can use `.forEach` too

js/src/tab.js
```diff
@@ -240,6 +270,36 @@ class Tab {
...
+        tabs[j].setAttribute('tabindex', '0')
+        selectedTabFound = true
+      } else {
+        tabs[j].setAttribute('tabindex', '-1')
+        tabs[j].setAttribute('aria-selected', 'false')
+      }
+    }
+
+    // if none of the tabs were explicitly marked as selected, pick first one
+    if (selectedTabFound === false) {
```
Johann-S: `if (!selectedTabFound)`

Johann-S (changes requested):

Johann-S:
Hi @patrickhlauke,

About your use of `.click()` for me the main issue is when the page is loaded you do not create any Tab instance, you just add accessibility attributes, you should initialize Tabs with: `const tab = new Tab(element)` and after that on your keyboard listener you'll be able to do: `Data.getData(element)` to get a Tab instance.
Like that you'll be able to call any public methods of our Tab component.

I hope what I said is clear 😟 if not, do not hesitate to ask me 😃

js/src/tab.js
patrickhlauke: but I'm potentially changing the value of `tablistorientation` in the next bit, so not a constant?

js/src/tab.js
patrickhlauke: still old-school me, so not quite sure how to remunge the code...particularly since at the end I jump back to the very first tab in the array if no selected tab was found during the run.

patrickhlauke:
> About your use of `.click()` for me the main issue is when the page is loaded you do not create any Tab instance, you just add accessibility attributes, you should initialize Tabs with: `const tab = new Tab(element)` and after that on your keyboard listener you'll be able to do: `Data.getData(element, DATA_KEY)` to get a Tab instance.
> Like that you'll be able to call any public methods of our Tab component.
> 
> I hope what I said is clear 😟 if not, do not hesitate to ask me 😃

I'll admit that I have no clue what all our magic instances and scripts and DATA_KEY things etc actually do at this point...I just cobble together bits from other things that seem to work. I'll have another crack at this tonight though to see if I can work it out...

As for the initialisation...does it make sense to say it should be in a separate function, and to call that rather than cramming it all into the event listener definition there? he main thing for my page initialisation is: it needs to run on page load before the user even starts interacting with any tab interfaces, to make sure the correct tabindices and `aria-selected` are set (so before any keyboard listener is even called). But not quite sure how to achieve that with all the code BS uses (I was trying some basic approaches, but kept getting various weird errors with my naive approaches). Maybe if you have a basic gist of how it could/should look that could give me a clue...

(that's partly the problem for me with these things...give me basic vanilla JS, and I can probably spaghetti code my way to the solution...but with the various abstractions like SelectorEngine etc that we use in BS, I'm mostly lost now...)

js/src/tab.js
Johann-S: in that case `let` if fine 👍

js/src/tab.js
Johann-S: something like: 
```js
tabs.forEach((tab) => {
  if (tab.getAttribute('aria-selected') === 'true' && !selectedTabFound) {
    tab.setAttribute('tabindex', '0')
    selectedTabFound = true
  } else {
    tab.setAttribute('tabindex', '-1')
    tab.setAttribute('aria-selected', 'false')
  }
})
```

btw I added some changes, it's not needed to use `.hasAttribute` because if no attribute `.getAttribute` will return `null`.

js/src/tab.js
```diff
@@ -287,15 +287,15 @@ EventHandler.on(window, Event.LOAD_DATA_API, () => {
     let selectedTabFound = false
 
     // iterate over each tab in the tablist, make sure they have correct tabindex/aria-selected
-    for (let j = 0; j < tabs.length; j++) {
-      if (tabs[j].getAttribute('aria-selected') === 'true' && !selectedTabFound) {
-        tabs[j].setAttribute('tabindex', '0')
+    tabs.forEach(tab => {
```
XhmikosR: You should use our helper function `makeArray` here.

patrickhlauke: any pointers of how to do that would be welcome. as said, i've got very little idea of the idiosyncracies of our helper functions, admittedly...

XhmikosR:
You can look at how it's already being used like here https://github.com/twbs/bootstrap/blob/master/js/src/carousel.js#L337

js/src/tab.js
XhmikosR: You can look at how it's already being used like here https://github.com/twbs/bootstrap/blob/master/js/src/carousel.js#L337

patrickhlauke: tried to tighten that into `makeArray(SelectorEngine.find(Selector.DATA_TOGGLE, tablists[i])).forEach(tab => {` but then not obvious how after the loop i can revisit the first tab to set it unless i run some other SelectorEngine thing again, which seems wasteful/pointless?

XhmikosR: There's a reason we have this utility. Some browsers don't support `NodeList.forEach`

patrickhlauke: but i am using makeArray a few lines earlier...not quite sure i follow what you're saying here?

patrickhlauke: as in `tabs` is defined a few lines earlier already as `const tabs = makeArray(SelectorEngine.find(Selector.DATA_TOGGLE, tablists[i]))` ... just that i'm not currently chaining the `forEach` directly after it but storing it in the variable

XhmikosR: Ah, nvm then, it should be fine.

patrickhlauke:
Think at this stage i got to the end of my understanding/knowledge of how to tackle this for now. @XhmikosR or @Johann-S  ... not sure if you have time or inclination to point me to the most sensible way to do the initialization (rather than cramming it all directly into the load event listener thing)? Or is that actually ok? (looks ugly to me)

additionally, one aspect that is still broken: for tab panels with a `fade`, it seems that using keyboard i can switch tabs quicker than the fade actually happening/completing, leading to situations where a new tab is being shown while the one that was in the process of being shown isn't fully shown yet...and you end up with two tab panels showing at the same time. again, any pointer/indication of how this could be tackled would be appreciated...

and yes, also need to come up with unit tests to check the new keyboard behavior, and to check the initialisation happens properly...

Johann-S:
I don't think it's ugly to initialize our tabs in the load event, we do that for our other components for example Carousel or Scrollspy, so feel free to do it that way.

I don't have an answer for your other issue, but I'll take a look into it when I'll find the time

patrickhlauke:
if you pardon the enormous GIF for a second, here's an illustration of the problem when switching tabs too quickly with keyboard...

skillsoft-learning-transcript-animated-focus-demonstration

patrickhlauke:
I've tried tackling this problem, but my understanding of how the whole async stuff is happening is preventing me from making any progress. In essence, the problem is due to the `.fade` and the assumption the code seems to make that there's only ever one previous tab that is currently still transitioning. when using a mouse, it seems you're never faster than the transition in clicking (or it somehow just ignores the click?), but when using the keyboard you can fire the tab changes very quickly. it further happens specifically when you do a quick "left, then right" with the cursor keys, i.e. you leave a tab, the previous active shown panel begins to fade out, the new panel begins to fade in, but then you're going back to the previous tab. somewhere along the line the code gets confused and then BOTH those tabs are shown. @Johann-S @XhmikosR any help on this would be appreciated, as apart from this stupid fading behavior, everything else seems to work just great (barring the absence of unit tests, which i'll tackle once we get this all working properly)...

Johann-S:
just a question @redux, is it possible for a `.tab-content` to have `role="tablist"` ? Or is it just for the list of tabs

patrickhlauke:
> just a question @Redux, is it possible for a `.tab-content` to have `role="tablist"` ? Or is it just for the list of tabs

the latter. it's only for the list of the tabs/triggers. the content panels need to be outside of the tablist. (that was an old bug in ARIA 1.0 that was amended in 1.1)

patrickhlauke:
just checked the deployed preview of your change @Johann-S ... YAY, seems to work fine. i'll polish this a bit more and add unit tests etc, but this fills me with joy :)

Johann-S:
The main reason was because when I worked on your branch, I found an error in `tabs.html` in visual tests, and it was because one tab content had `role="tablist"` 😄 

You're welcome 😉

patrickhlauke:
last remaining (non-trivial) work here is to add some sensible handling of disabled tabs. they should be skipped/ignored (e.g. if i have 3 tabs, and the second one is disabled, then when focused on the first one and pressing right cursor key, the third tab should be focused/become active)

Johann-S:
Hey @patrickhlauke, I'm sure you'll hate me 😆 but I removed QUnit and switch to Jasmine, so now your unit tests should go there: https://github.com/twbs/bootstrap/blob/master/js/src/tab/tab.spec.js

`js/tests/unit/tab.js` have been removed, so if you need any help do not hesitate 😉

patrickhlauke:
heh, the dangers of taking too long / not having time to get this over the line. no worries, i'll have to look at this jasmine thing eventually anyway, might as well do it as part of this ;)

patrickhlauke:
would love to review efforts along these lines...at this point, probably worth scrapping this PR and starting from scratch again. but my lack of JS chops will probably still raise its head once I hit the trickier parts. so preemptively asking: anybody (@Johann-S, others?) able to collaborate on a new take? I can explain what things should do, but once it gets to integrating that code and how all the event model in BS works...

patrickhlauke:
putting out a call again for anybody who'd have time/inclination to help take this forward with me and help me overcome my woefully lacking JS skills...

XhmikosR:
@patrickhlauke if you could make this from scratch against main, it should make things easier to look into.

patrickhlauke:
Superseded by https://github.com/twbs/bootstrap/pull/33079

PR #28931, opened 2019-06-19
Final v3 changes

XhmikosR:
I need someone to do the usual process to verify the dist files, because I keep seeing changes in the CSS map files in some cases (on Linux, Windows is expected to have differences due to a bug in clean-css):

```sh
git checkout v3-dev-xmr
git clean -dfx
npm i
npm t
# here check about modifications with `git status`
npm run release
# check again here about modifications
```

Changes: <https://github.com/twbs/bootstrap/compare/v3.4.1...v3-dev-xmr>


mdo:
Are these just docs updates or is there something I'm not seeing in the diff here that requires us to update v3? I don't want to signal to folks that this version of our project is actively maintained other than important changes.

XhmikosR:
Well, if you recall I didn't want to release a patch version just for these. But from the internal communications, I thought the consensus was to release it due to the broken links.

Note that the gh-pages branch already has these fixes.

mdo:
Yeah, I'm fine having `gh-pages` have the changes that an official release doesn't have.

XhmikosR:
Alright I'd still like to merge this without the dist and the version bump.

PS we should make that blog post about EOL etc

XhmikosR:
@mdo: removed the dist and version bump patches. Let's merge this and leave v3 once and for all.

Johann-S (approved):

PR #28939, opened 2019-06-21
Combine duplicate documentation examples into one to fix the confusing text.

### LINKED ISSUE #28892: Documentation: Equal-width multi-row example doesn't illustrate concept
[https://getbootstrap.com/docs/4.3/layout/grid/#equal-width-multi-row](https://getbootstrap.com/docs/4.3/layout/grid/#equal-width-multi-row)

**Layout > Grid > Equal-width multi-row**

The documentation for this topic begins with, "Create equal-width columns that span multiple rows by..." however the example given does not demonstrate any columns spanning multiple rows. In the example:

```
<div class="container">
  <div class="row">
    <div class="col">col</div>
    <div class="col">col</div>
    <div class="w-100"></div>
    <div class="col">col</div>
    <div class="col">col</div>
  </div>
</div>
```

Each column spans a single row in a 2 x 2 grid.

There's a similar topic located higher up on the same page with the same code example that has the text, "Equal-width columns can be broken into multiple lines" which makes a little more sense as the concept seems to be showing how you can split, or break, equal-width columns over multiple lines within a row.

jchawkins725:
Fixes #28892

### DIFF
--- a/site/content/docs/4.3/layout/grid.md
+++ b/site/content/docs/4.3/layout/grid.md
@@ -151,9 +151,9 @@ For example, here are two grid layouts that apply to every device and viewport, 
 
 ### Equal-width multi-line
 
-Create equal-width columns that span multiple lines by inserting a `.w-100` where you want the columns to break to a new line. Make the breaks responsive by mixing the `.w-100` with some [responsive display utilities]({{< docsref "/utilities/display" >}}).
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

MartijnCuppens (approved):

XhmikosR:
I need a confirmation about the text edits and then we merge.

MartijnCuppens:
Lgtm

PR #28944, opened 2019-06-22
Change invalid icon from cross to exclamation mark

### LINKED ISSUE #28440: Cross icon is confusing in an input element
The icon used on input.is-invalid. There are a few tickets about this issue and some more thumbs up on them, but closed. Please, reconsider.

I also just found the icon confusing. It's usually used as a reset button, and I'm pretty sure IE or Edge uses such icon to reset the input value. Maybe use an exclamation mark.

MartijnCuppens:
The cross icon was confusing since some browsers use it as a reset button.

The svg is handcrafted by myself.

Fixes #28440

Demo: https://deploy-preview-28944--twbs-bootstrap.netlify.com/docs/4.3/components/forms/#server-side


patrickhlauke (approved):

XhmikosR:
Any reason why you have width/height set to 10?

MartijnCuppens:
The viewport used to be 10 by 10, but this rendered bad in IE. I've now updated it to 12 by 12.

XhmikosR (approved):

usulix:
This looks A LOT like the fontawesome font-not-found flashing exclamation mark and is definitely a different look and feel from the solid checkmark - so much so that I thought it was an error until I walked back through blame and found this commit. And, since validation icons have been removed from the documentation - I had no way of knowing this as "as intended". I'll lock to the version before this change and hereby put in at least one vote to revert it, please?

PR #28948, opened 2019-06-24
Allow .page-links to get left margin.

GeoSot:
Give the choice to set a left-margin to pagination links.


ysds:
In general there are many paginations with separated links, so I have no strong reason to pass this. But, if we add this variable, it will be wanted to adjust the rounded border too.

image

Any thoughts? @twbs/css-review

scss/_pagination.scss
```diff
@@ -33,12 +33,19 @@
...
   }
   &:last-child {
     .page-link {
-      @include border-right-radius($border-radius);
+      @include border-right-radius($pagination-border-radius);
+    }
+  }
+
+  //Add border-radius to all pageLinks in case they have left margin
+  @if $pagination-margin-left != ( - $pagination-border-width ) {
```
MartijnCuppens: We can use a if/else statement here to prevent the generation of the `:last-child` when ` $pagination-margin-left != ( - $pagination-border-width )`

This line can also be simplified as 
```suggestion
  @if $pagination-margin-left != -$pagination-border-width {
```

scss/_pagination.scss
```diff
@@ -30,7 +30,7 @@
 }
 
 .page-item {
-  @if $pagination-margin-left == - $pagination-border-width {
```
GeoSot: I guess that, latest space fix doesn't match with linting rules

MartijnCuppens: Woops, I've updated this to brackets which looks a bit better

MartijnCuppens:
The `scss/mixins/_pagination.scss` file still needs some border-radius love (the thing with if/else).

GeoSot:
> The `scss/mixins/_pagination.scss` file still needs some border-radius love (the thing with if/else).

I  didn't find more love inside me to give. 
( Help me on this, I don't see something that needs refactoring )

GeoSot:
I hadn't  see the mixin. 

We can change


```scss 

.page-item {
  @if $pagination-margin-left == (-$pagination-border-width) {
    &:first-child {
      .page-link {
        margin-left: 0;
        @include border-left-radius($pagination-border-radius);
      }
    }

    &:last-child {
      .page-link {
        @include border-right-radius($pagination-border-radius);
      }
    }
  } @else {
    &:first-child .page-link {
      margin-left: 0;
    }

    //Add border-radius to all pageLinks in case they have left margin
    .page-link {
      @include border-radius($pagination-border-radius);
    }
  }
```
to 


```scss 
  @include pagination-size($pagination-padding-y, $pagination-padding-x, $font-size, $line-height, $pagination-border-radius);

.page-item:first-child .page-link {
      margin-left: 0;
    }

```

to avoid duplicated code

MartijnCuppens:
@GeoSot, yup we can. I've tried that in https://github.com/twbs/bootstrap/compare/master...MartijnCuppens:pagination-margin?expand=1

This commit also prevents the generation of `margin-left: 0;`

The only downside is it generates a duplicate `.page-link` selector, but our css minifier will combine that anyway.

What do you think, @GeoSot?

GeoSot:
I think it's ok.  BUT do not remove pagination.scss line 11 
```scss 
 margin-left: -$pagination-border-width;
```

MartijnCuppens:
Instead of removing the margin for the first child later, I've changed that to:
https://github.com/twbs/bootstrap/blob/6ccad20da3e402573e5445b1428e47fbdef258ce/scss/_pagination.scss#L30-L32

XhmikosR:
>The only downside is it generates a duplicate .page-link selector, but our css minifier will combine that anyway.

Please verify this because I'm not 100% sure that's the case.

MartijnCuppens:
> Please verify this because I'm not 100% sure that's the case.

Nope it doesn't combine the selectors. But even than, it's just a minor increment of css.

I'll have a look if we can increase to level 2 of css-clean.

XhmikosR:
>I'll have a look if we can increase to level 2 of css-clean. (unrelated to this PR)

I'd suggest to stay away from this since it can break many things compared to the gains.

mdo:
I'm late to this PR, but I'm not in favor of this change. The pagination component is designed to be a unified component. If you need separate links, I'd suggest creating your own component with some kind of button class.

GeoSot:
@mdo 

By adding this change, the pagination still works the same way, but will fit into some projects with special pagination needs.
We already customize the component to meet the needs of different design approaches, and while the point of variables.scss file is to give the option to customize things based on our needs, it would be very comfy to have one more available customization and keep component usage

MartijnCuppens:
I personally like this feature because in the past I always needed to make my own components whenever the pager needed to be themed like this. With this change I can always keep my markup in sync with the Bootstrap markup and theme things within sass.

ysds (approved):
There is nothing wrong with the changed CSS.

MartijnCuppens (approved):
lgtm too. Just checking with @mdo if you're ok with merging this?

PR #28952, opened 2019-06-24
Ensure button plugin sets/removes active class correctly on page load

patrickhlauke:
More fundamentally fixes a problem with Bootstrap's "faked" toggles, checkboxes, radio buttons ... the fact that the visual presentation can get out of sync with the real value/state of controls if, for instance, autocomplete happened (which would change the underlying value, but not set the correct `active` class). This does a sanity check on page load and makes sure the `active` class is only set if the underlying control is really pressed/checked, and otherwise removes it. This guarantees that they're in sync from the moment the page is loaded, regardless of whether or not a browser did an autocomplete, or even the author forgot to set BOTH the class and the relevant attribute.

Also, removes the now unnecessary `autocomplete="off"` attributes used to essentially prevent out-of-sync situations from happening (in Firefox, primarily). Now, if Firefox does an autocomplete, this new logic ensures that the `active` class will be correct, no problem.

Closes https://github.com/twbs/bootstrap/issues/28623

Supersedes https://github.com/twbs/bootstrap/pull/28649

Could potentially be ported to v5, but once https://github.com/twbs/bootstrap/pull/28463 lands the part relating to checkboxes/radio buttons will be irrelevant ... will create a separate v5-specific patch to keep just the `aria-pressed` toggles in sync on page load (probably as part of that PR itself)

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -27,17 +27,20 @@ const ClassName = {
 }
 
 const Selector = {
-  DATA_TOGGLE_CARROT : '[data-toggle^="button"]',
-  DATA_TOGGLE        : '[data-toggle="buttons"]',
-  INPUT              : 'input:not([type="hidden"])',
-  ACTIVE             : '.active',
-  BUTTON             : '.btn'
+  DATA_TOGGLE_CARROT   : '[data-toggle^="button"]',
+  DATA_TOGGLES         : '[data-toggle="buttons"]',
+  DATA_TOGGLE          : '[data-toggle="button"]',
+  DATA_TOGGLES_BUTTONS : '[data-toggle="buttons"] .btn',
+  INPUT                : 'input:not([type="hidden"])',
+  ACTIVE               : '.active',
+  BUTTON               : '.btn'
 }
 
 const Event = {
   CLICK_DATA_API      : `click${EVENT_KEY}${DATA_API_KEY}`,
   FOCUS_BLUR_DATA_API : `focus${EVENT_KEY}${DATA_API_KEY} ` +
-                          `blur${EVENT_KEY}${DATA_API_KEY}`
+                          `blur${EVENT_KEY}${DATA_API_KEY}`,
+  LOAD_DATA_API  : `load${EVENT_KEY}${DATA_API_KEY}`
 }
 
 /**
@@ -63,7 +66,7 @@ class Button {
     let triggerChangeEvent = true
     let addAriaPressed = true
     const rootElement = $(this._element).closest(
-      Selector.DATA_TOGGLE
+      Selector.DATA_TOGGLES
     )[0]
 
     if (rootElement) {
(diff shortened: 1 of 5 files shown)


js/src/button.js
```diff
@@ -167,6 +170,33 @@ $(document)
...
+$(window).on(Event.LOAD_DATA_API, () => {
+  // ensure correct active class is set to match the controls' actual values/states
+
+  // find all checkboxes/readio buttons inside data-toggle groups
+  let buttons = [].slice.call(document.querySelectorAll(Selector.DATA_TOGGLES_BUTTONS))
+  for (let i = 0, len = buttons.length; i < len; i++) {
+    const button = buttons[i]
+    const input = button.querySelector(Selector.INPUT)
+    if (input.checked || input.hasAttribute('checked')) {
+      $(button).addClass(ClassName.ACTIVE)
```
Johann-S: button.classList.add(ClassName.ACTIVE) 😉

js/src/button.js
```diff
@@ -167,6 +170,33 @@ $(document)
...
+    } else {
+      $(button).removeClass(ClassName.ACTIVE)
```
Johann-S: button.classList.remove(ClassName.ACTIVE)

js/src/button.js
patrickhlauke: sure, but i thought we're still jQuery-ing? didn't want to mix and match styles as it's used in the existing code. happy to change to pure DOM calls though

Johann-S: we try to use jQuery only when it's needed (mostly for the event system)

js/src/button.js
```diff
@@ -27,17 +27,20 @@ const ClassName = {
...
+  ACTIVE               : '.active',
+  BUTTON               : '.btn'
 }
 
 const Event = {
   CLICK_DATA_API      : `click${EVENT_KEY}${DATA_API_KEY}`,
   FOCUS_BLUR_DATA_API : `focus${EVENT_KEY}${DATA_API_KEY} ` +
-                          `blur${EVENT_KEY}${DATA_API_KEY}`
+                          `blur${EVENT_KEY}${DATA_API_KEY}`,
+  LOAD_DATA_API  : `load${EVENT_KEY}${DATA_API_KEY}`
```
Johann-S: a bit of indentation here

js/src/button.js
```diff
@@ -167,6 +170,33 @@ $(document)
     $(button).toggleClass(ClassName.FOCUS, /^focus(in)?$/.test(event.type))
   })
 
+$(window).on(Event.LOAD_DATA_API, () => {
+  // ensure correct active class is set to match the controls' actual values/states
+
```
Johann-S: trailing line

Johann-S (approved):

js/src/button.js
patrickhlauke: wonder if that could be left there, as it makes it clearer that there's then two separate blocks that follow (otherwise the comment on the next line sort of merges with this one?

js/src/button.js
patrickhlauke: oops, fixed

js/src/button.js
Johann-S: Yep you can merge with the other one seems fine to me

patrickhlauke: do you mean yep can leave that extra line, or "no, remove the line"?

it's a small thing of course, but feel that `// ensure correct active class is set to match the controls' actual values/states` is the comment/description of what happens in the whole thing, and then `// find all checkboxes/readio buttons inside data-toggle groups` and `// find all button toggles` are subsections...so it's visually clearer if there's that empty line

patrickhlauke: or was it a "you can merge the PR"?

PR #28959, opened 2019-06-25
Add focus style for checkboxes and radios

### LINKED ISSUE #28957: No focus style on checkbox and radios
### Description

There's no visual feedback when focusing a checkbox or radio.

### Ask

- Is this intended ?
- If not, will it be fixed ?

### Test case

Can be tested on [the current doc](https://getbootstrap.com/docs/4.3/components/forms/#checkboxes-and-radios)

### Additional Info

- A similar bug was opened for v.3 and has been fixed #14923
- [Custom form controls](https://getbootstrap.com/docs/4.3/components/forms/#custom-forms) checkboxes and radios have a focus style.

alienlebarge:
Add focus style for default checkboxes and radios 

#28957

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -217,6 +217,18 @@ textarea.form-control {
   &[disabled] ~ .form-check-label {
     color: $text-muted;
   }
+
+  &:focus {
+    color: $input-focus-color; // only needed for fallback to text input
+    background-color: $input-focus-bg;
+    border-color: $input-focus-border-color;
+    outline: 0;
+    @if $enable-shadows {
+      box-shadow: $input-box-shadow, $input-focus-box-shadow;
+    } @else {
+      box-shadow: $input-focus-box-shadow;
+    }
+  }
 }
 
 .form-check-label {


patrickhlauke:
I'm in favor of this, although note that it's not always 100% perfect in, say, Chrome

> with the new styles

but then, neither is Chrome's default

> default chrome outline on radio buttons

patrickhlauke:
@mdo if you're not planning to have these styles for v5 (though I'm not sure I understand why/what the rationale is), then at least this can be backported to v4 though

patrickhlauke:
urgh ok, so changing base just to v4 made this PR super-ugly...in which case (as we'd want this still in v4 i'd say) would you mind closing this one and starting a fresh PR targeting `v4-dev`? sorry about the to-ing and fro-ing

XhmikosR:
We can rebase it if needed.

MartijnCuppens:
@alienlebarge,
Could you clarify what these lines are for:
https://github.com/twbs/bootstrap/blob/cf521701c46407d8ca479b0037c21e9a62471896/scss/_forms.scss#L222-L224

patrickhlauke:
i think a more concise way may be to just add the mixin, without even needing to replicate the `:focus` block manually here
```
@include form-control-focus($ignore-warning: true);
```

alienlebarge:
@MartijnCuppens

> @alienlebarge,
> Could you clarify what these lines are for:
> 
> https://github.com/twbs/bootstrap/blob/cf521701c46407d8ca479b0037c21e9a62471896/scss/_forms.scss#L222-L224

I have simply use the code of the `@include form-control-focus($ignore-warning: true);` mixin but since it will be deprecated, I havn't use the mixin itself.

As @patrickhlauke propose, it could be better to use it and it will be refactor when removed

@patrickhlauke 

> urgh ok, so changing base just to v4 made this PR super-ugly...in which case (as we'd want this still in v4 i'd say) would you mind closing this one and starting a fresh PR targeting `v4-dev`? sorry about the to-ing and fro-ing

Do you still want a PR targeting `v4-dev`?

patrickhlauke:
@alienlebarge sorry for all the to-ing and fro-ing, and yes there's lots of moving parts now that we've got v5 but still maintaining some parts of v4 etc...

if you don't mind, simplest would probably be to close this, make a new PR just with the mixin inclusion, and target v4-dev

PR #28972, opened 2019-06-27
Drop support for .form-control-plaintext inside .input-group

### LINKED ISSUE #28205: Readonly plain text inside input group
`.form-control-plaintext` inside `.input-group`s are supported from v4.2  (#25871). But the result is:

image

Demo: https://codepen.io/anon/pen/aXEGMZ

It this an expected appearance?

ysds:
Drop support for `.form-control-plaintext` inside `.input-group`, because this cause #28205.

Closes #28205


patrickhlauke:
I'd be cautiously in favor of this.

MartijnCuppens (approved):

gcacars:
Why? This is so usefull. I lost minutes trying to solve this. Makes no sense if an input field could have append and prepend, but a readonly input not. 🤷‍♂️
We just couldn't use something like below?

```css
.input-group > .form-control-plaintext:not(:first-child) {
  padding-right: 0.75rem;
}
```

PR #28975, opened 2019-06-28
Add check input focus style

### LINKED ISSUE #28957: No focus style on checkbox and radios

alienlebarge:
### Description

Add focus style for checkboxes and radios

### Additional Info

1. Initial Issue: close #28957 
1. It was first propose tu make the change for current and next version of bootstrap. But it was decided to only target v4. (See #28959 for more info)

Demo: https://deploy-preview-28975--twbs-bootstrap.netlify.com/docs/4.3/components/forms/#without-labels


patrickhlauke:
@XhmikosR this probably needs some kind of squashing/rebasing/cherry-picking i think ... the change itself is just to that one scss file but because it started out against master it carries a lot of baggage...my git-fu is not quite there, so you ok tackling that?

alienlebarge:
If you prefer I can start from `v4-dev`

MartijnCuppens:
We need to fix the border radius of the checkbox and radio somehow. The problem is the radius of the checkbox cannot be detected since not all browsers/OS'es have the same radius.

patrickhlauke:
> We need to fix the border radius of the checkbox and radio somehow. The problem is the radius of the checkbox cannot be detected since not all browsers/OS'es have the same radius.

checkboxes and radio buttons themselves are stubborn (in Chrome in particular) against styling their border or anything, and the box shadow can't be influenced either it seems. the only way to get the rounded indication here would be to essentially redo the whole checkbox/radio input with faked elements, which is outside of the scope of this tiny change. adding a `border-radius` for radio type inputs does seem to have an effect in IE/Edge/Firefox though

i'd say we can live with the non-rounded focus indication in some browsers (Chrome for instance ... in Firefox, it's actually reasonably pleasing, particularly for the radio button), as this is a major step forward compared to some browsers (Firefox, IE, ...) not having a clear focus indication at all

scss/_forms.scss
```diff
@@ -215,6 +215,8 @@ textarea.form-control {
   &[disabled] ~ .form-check-label {
     color: $text-muted;
   }
+
+  @include form-control-focus($ignore-warning: true);
 }
```
patrickhlauke: tried adding a commit to this directly, but seems it may not be accepting commits from maintainers...

in any case, I'd suggest adding the following here to mitigate the slightly unsightly look of round radio buttons with a completely square box shadow in IE/Edge (Firefox already rounds them, and Chrome is just stubborn about any styling of checkboxes/radio buttons in general)

```
  &[type="radio"] {
    // Note: This has no effect on the appearance of the radio button, but does smooth out the box shadow on focus in most browsers.
    @include border-radius($input-border-radius, 0);
  }
```

patrickhlauke (changes requested):

scss/_forms.scss
alienlebarge: I made [the change](https://github.com/alienlebarge/bootstrap/commit/565177310f939c6d452971649c115d18edf07f18) on the [add-check-input-focus-style-v4](https://github.com/alienlebarge/bootstrap/tree/add-check-input-focus-style-v4) branch on [alienlebarge/bootstrap](https://github.com/alienlebarge/bootstrap) but it doesn't appear in the PR...

Do you have any idea why ?

alienlebarge: @patrickhlauke done

patrickhlauke:
for info @twbs/css-review not sure what that travis problem is, but seems unrelated to the CSS change here

patrickhlauke (approved):

patrickhlauke:
noting that if the change is still felt too disruptive (because yes, in Chrome it currently doesn't look super fantastic), this could be flipped to use `:focus-visible` instead (which works in Firefox right now, and will be coming to Chrome and Edge ... so a future-facing change for when those browsers support it, and in the meantime it fixes the issue in Firefox and maintains the status quo in other browsers)

XhmikosR:
@twbs/css-review: what should we do with this? It targets v4-dev BTW.

MartijnCuppens:
I personally don't really like it. It just doesn't look that good:
image & image

XhmikosR:
Your guys' call, feel free to close it. Looking at the remaining v4-dev stuff.

XhmikosR:
@twbs/css-review final call for action about this PR?

MartijnCuppens:
Not a fan of the current theming (see https://github.com/twbs/bootstrap/pull/28975#issuecomment-512275014), however I have no clue how I could improve it, so it's fine by me if it gets merged.

scss/_forms.scss
```diff
@@ -215,6 +215,14 @@ textarea.form-control {
   &[disabled] ~ .form-check-label {
     color: $text-muted;
   }
+
+  @include form-control-focus($ignore-warning: true);
```
XhmikosR: @MartijnCuppens should this be here?

ysds:
I have the same opinion as @MartijnCuppens

PR #28980, opened 2019-07-01
Remove redundant border radius on pagination

MartijnCuppens:
Pagination used to have box shadow back in the days which required this property, but nowadays it's redundant. (see https://github.com/twbs/bootstrap/blame/0cd186183c5e18f0517d6037d53f29c56adbac72/less/pagination.less#L17)

### DIFF
--- a/scss/_pagination.scss
+++ b/scss/_pagination.scss
@@ -1,7 +1,6 @@
 .pagination {
   display: flex;
   @include list-unstyled();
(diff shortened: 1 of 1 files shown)


ysds (approved):

mdo:
We had added this originally so folks didn't need to add the property when adding a `box-shadow`.

PR #28983, opened 2019-07-02
Add ARIA search landmark to documentation.

bardiharborow:
Reference: [WAI-ARIA Authoring Practices 1.1 § 4.3.8](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_search).

Addendum: `.bd-sidebar` and `bd-toc` don't appear to have enough roles, but I'm new at this whole ARIA thing so would prefer advice on this one.


patrickhlauke (approved):

patrickhlauke:
i'd say `.bd-toc` is fine as is. `.bd-sidebar` probably can be seen as a navigation, but so does the stuff in the header...when you then have more than two containers acting as navigation, they need to be disambiguated with some form of label/name (e.g. an `aria-label`) so it's clear what the difference is (like "general site navigation" and "documentation navigation" or similar). draft some PR up and we can workshop it :)

PR #28986, opened 2019-07-04
Simplify code for 'cover' example using utility classes

coliff:
This reduces the size of the CSS and (I think) makes the example source code easier to read.

Note; the `background-color` is very slightly different as I used `bg-dark`.

https://deploy-preview-28986--twbs-bootstrap.netlify.com/docs/4.3/examples/cover/


site/content/docs/4.3/examples/cover/index.html
```diff
@@ -23,13 +23,11 @@ <h3 class="masthead-brand">Cover</h3>
...
-      <a href="#" class="btn btn-lg btn-secondary">Learn more</a>
+      <a href="#" class="btn btn-lg btn-secondary border-white">Learn more</a>
     </p>
   </main>
 
   <footer class="mastfoot mt-auto">
-    <div class="inner">
-      <p>Cover template for <a href="https://getbootstrap.com/">Bootstrap</a>, by <a href="https://twitter.com/mdo">@mdo</a>.</p>
-    </div>
+    <p>Cover template for <a href="https://getbootstrap.com/">Bootstrap</a>, by <a href="https://twitter.com/mdo">@mdo</a>.</p>
```
XhmikosR: It seems there are a few instances of `inner` but I don't see anything in the CSS. Maybe remove those too?

coliff: good idea. I've updated the PR.

XhmikosR: BTW it seems the whole cover example could be simplified a lot more by using more utils.

Let's wait for the others to chime in, but it might be worth tackling all of it here.

coliff: Cool. One thing I was wondering about. It'd be good if the examples could specify in the frontmatter a CSS class to add to the `<html>` tag. For example for this one we could add:
`html_class: "h-100"` so that would add the `h-100` class to the `html` tag. Then this could be removed from the CSS. I think there's a few examples that could take advantage of specifying CSS classes in the `html` tag like this.

MartijnCuppens: I'm definitely in favor of `html_class: "h-100"`!

XhmikosR: This is already possible for the examples. https://github.com/twbs/bootstrap/blob/master/site/layouts/_default/examples.html#L2

MartijnCuppens:
We can remove this in favor of the `.text-white` utility:
https://github.com/twbs/bootstrap/blob/01af91124b738c96433ef22971abd06cad2ab2bd/site/content/docs/4.3/examples/cover/cover.css#L5-L10

We can use the `.btn-light` or maybe `.btn-outline-light` class here instead:
https://github.com/twbs/bootstrap/blob/01af91124b738c96433ef22971abd06cad2ab2bd/site/content/docs/4.3/examples/cover/cover.css#L12-L19

This css is overridden by `.mt-auto`:
https://github.com/twbs/bootstrap/blob/01af91124b738c96433ef22971abd06cad2ab2bd/site/content/docs/4.3/examples/cover/cover.css#L44-L46

We can use the responsive float classes here:
https://github.com/twbs/bootstrap/blob/01af91124b738c96433ef22971abd06cad2ab2bd/site/content/docs/4.3/examples/cover/cover.css#L70-L77

We can use `.text-white-50` class here:
https://github.com/twbs/bootstrap/blob/01af91124b738c96433ef22971abd06cad2ab2bd/site/content/docs/4.3/examples/cover/cover.css#L92-L97

XhmikosR:
Plus the `.cover` styles minus the button's vertical padding.

coliff:
thanks for the good feedback @XhmikosR and @MartijnCuppens . I've further simplified the CSS and HTML.

site/content/docs/4.3/examples/cover/index.html
```diff
@@ -3,33 +3,32 @@
...
 include_js: false
 ---
 
 <div class="cover-container d-flex w-100 h-100 p-3 mx-auto flex-column">
   <header class="masthead mb-auto">
-    <div class="inner">
-      <h3 class="masthead-brand">Cover</h3>
-      <nav class="nav nav-masthead justify-content-center">
+    <div>
+      <h3 class="masthead-brand mb-0 float-md-left">Cover</h3>
```
mdo:
```suggestion
      <h3 class="float-md-left mb-0">Cover</h3>
```

Since `.masthead-brand` is unused, we can drop it.

site/content/docs/4.3/examples/cover/index.html
```diff
@@ -3,33 +3,32 @@
 title: Cover Template
 extra_css:
   - "cover.css"
-body_class: "text-center"
+html_class: "h-100"
+body_class: "text-center text-white bg-dark d-flex h-100"
```
mdo:
```suggestion
body_class: "d-flex h-100 text-center text-white bg-dark"
```

I order utilities by our CSS property order whenever possible, so all the display/box-model stuff is at the front.

site/content/docs/4.3/examples/cover/index.html
```diff
@@ -3,33 +3,32 @@
...
 include_js: false
 ---
 
 <div class="cover-container d-flex w-100 h-100 p-3 mx-auto flex-column">
   <header class="masthead mb-auto">
```
mdo:
```suggestion
  <header class="mb-auto">
```

mdo (commented):
Few more changes to simplify just a bit more, plus some nits :D.

XhmikosR (approved):

MartijnCuppens (approved):

coliff:
Not really a big deal or anything, but can this one be backported to v4? I saw many of the other example code updates were backported so thought this might be worth doing too.

MartijnCuppens:
We switched from jekyll to Hugo, so I don't know if we can just cherrypick this.

XhmikosR:
Yeah, it didn't apply clean that's why I didn't cherry pick it.

PR #28988, opened 2019-07-04
Remove attribute selectors from buttons

MartijnCuppens:
We needed this, back in the days when this was still present in our code (see https://github.com/twbs/bootstrap/issues/4666):
https://github.com/twbs/bootstrap/blob/fd49d4a44f7fe73054b861461b9ec0f36a1bb29a/less/forms.less#L102-L108

But nowadays it's redundant.


ysds (approved):

mdo (approved):
I feel like this had something to do with Normalize back in the day, but I don’t see anything relevant there still today, so likely safe to remove.

PR #28989, opened 2019-07-04
Remove text align from body

### LINKED ISSUE #28238: reboot styles breaks rtl text direction alignment
### Current behavior

Bootstrap reboot sets `text-align:left` on the `body` HTML element which in turn negates all effects of setting `dir=rtl` attribute OR `direction` CSS property on an element.

> Test case behaviour _with_ bootstrap (current working): https://jsbin.com/jeteqisema

### Expected behavior

Setting text direction as right-to-left on elements should change the text direction and align it to the right (viz. the default browser behavior on all supported browsers).

> Test case behavior _without_ bootstrap (expected behavior): https://jsbin.com/riwafurose

### Workaround

Set the value of CSS property `text-align` to `initial` for an element fixes on most modern browsers. On [browsers which do not support `initial`](https://caniuse.com/#feat=css-initial-value), explicitly setting the value to `right` fixes the issue.

> Test case behavior _with_ bootstrap override (workaround): https://jsbin.com/zomezoyepu

### Background

Issue #24807 tracks the rtl language support within bootstrap as an enhancement feature; however the current reboot behavior breaks the default semantics for CSS. Making bootstrap components work with 
rtl languages is an enhancement feature, however breaking the syntax and flow direction of plain text content seems to be a bug.

#### Additional cases to consider:
- For user generated rich text content on a page using a right-to-left language, its **_not_** possible to predict/determine and set right alignment using CSS utility classes
- Setting the unexpected `text-align` on `body` tag leads to CSS specificity wars and IE hacks.
- The overall benefits of setting an explicit `text-align:left` on the body with CSS vs. setting it for individual components/controls is not apparent from the code comments. ref [original commit](https://github.com/twbs/bootstrap/commit/09019a8ecdccf61b52681dec3376a88509b4e6a3). This makes it hard to figure if this behavior is intended for the current bootstrap version OR a new reset for the `direction` attribute be provided as part of the reboot code.

#### Environment
- OS: all
- Browser: all

MartijnCuppens:
This was causing issues for RTL, see https://github.com/twbs/bootstrap/issues/28238.
The text align didn't needed to be set to fix (https://github.com/twbs/bootstrap/pull/22858), adding `th{ text-align: inherit; }` was enough (see https://codepen.io/MartijnCuppens/pen/zVLZbG)

Fixes https://github.com/twbs/bootstrap/issues/28238

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -25,11 +25,9 @@ main {
 // Body
 //
 // 1. Remove the margin in all browsers.
-// 2. Set an explicit initial text-align value so that we can later use
-//    the `inherit` value on things like `<th>` elements.
-// 3. As a best practice, apply a default `background-color`.
-// 4. Prevent adjustments of font size after orientation changes in IE on Windows Phone and in iOS.
-// 5. Change the default tap highlight to be completely transparent in iOS.
+// 2. As a best practice, apply a default `background-color`.
+// 3. Prevent adjustments of font size after orientation changes in IE on Windows Phone and in iOS.
+// 4. Change the default tap highlight to be completely transparent in iOS.
 body {
   margin: 0; // 1
   font-family: $font-family-base;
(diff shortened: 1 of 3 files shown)


ysds:
looks good, but I'm not sure this can be safely backported to v4.

MartijnCuppens:
> I'm not sure this can be safely backported to v4.

It could start aligning things to the right in rtl-layouts indeed, maybe we should only apply it to `v5`, just to be sure?

mdo (approved):

XhmikosR:
Maybe this should be configurable and not set in reboot?

MartijnCuppens:
> Maybe this should be configurable

Done

> and not set in reboot?

We already set other font related things in the body, so I guess that's the right spot.

XhmikosR:
Yeah, but since reboot can be used standalone, you know.

PR #28990, opened 2019-07-04
Update alignment __tables.scss

a21ns1g4ts:
This change align  the content of table row to vertical alignment
**Before**
before
**After:** 
after

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -6,13 +6,17 @@
   width: 100%;
   margin-bottom: $spacer;
   color: $table-color;
+  vertical-align: $table-cell-vertical-align;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
 
   th,
   td {
     padding: $table-cell-padding;
-    vertical-align: top;
     border-top: $table-border-width solid $table-border-color;
+  }
+
+  tbody {
+    vertical-align: inherit;
   }
 
   thead th {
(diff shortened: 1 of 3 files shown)


MartijnCuppens:
`vertical-align: top` will be the desired behaviour in many cases, so I'm not eager to merge this.

Maybe we can change the value to a variable, like `$table-cell-vertical-align`, which can be changed?

Atiladanvi:
Yes the alignment needs to be changed
I think the behavior of alignment of rows could be switchzes by a new class. [ .table-cell-bottom .table-cell-middle ... ]. What you think about this ?

MartijnCuppens:
We already have align utilities which can be used (https://getbootstrap.com/docs/4.3/utilities/vertical-align/).

Maybe we should remove the alignment on `td,th` so we can use the utility classes on the tables? (see https://codepen.io/MartijnCuppens/pen/MMqbbg?editors=1100)

Atiladanvi:
Good ! this solve the issue. You can request a Pr ?

MartijnCuppens:
PR updated, demo see: https://deploy-preview-28990--twbs-bootstrap.netlify.com/docs/4.3/content/tables/#vertical-alignment

XhmikosR:
@MartijnCuppens can you make a new PR with your patch?

MartijnCuppens:
Done https://github.com/twbs/bootstrap/pull/29039

Issue #28414, opened 2019-03-05
Validation feedback/tooltips lack association with their form control


patrickhlauke:
the feedback and tooltip containers need to be programmatically associated with their respective form controls, e.g. using `aria-describedby="..."`, e.g.

```html
<label for="validationTooltip03">City</label>
<input type="text" class="form-control" id="validationTooltip03" ... required>
<div class="invalid-tooltip">
   Please provide a valid city.
</div>
```

to

```html
<label for="validationTooltip03">City</label>
<input type="text" class="form-control" id="validationTooltip03" aria-descriebedby="validationTooltip03-tip" ... required>
<div class="invalid-tooltip" id="validationTooltip03-tip">
   Please provide a valid city.
</div>
```

danielnixon:
Interesting to note that this is actually a regression from Bootstrap 3 🤔 

https://getbootstrap.com/docs/3.4/css/#forms-help-text

>  Help text should be explicitly associated with the form control it relates to using the aria-describedby attribute. This will ensure that assistive technologies – such as screen readers – will announce this help text when the user focuses or enters the control.

patrickhlauke:
sure. lots of moving parts to bootstrap, so things often fall between the cracks on large refactors/reworkings.

patrickhlauke:
this issue is complicated by the fact that, despite `display:none`, certain browser/AT combinations will still announce the additional validation text referenced via `aria-describedby`. so these need to be added dynamically via JS, not as a pure CSS solution. thoughts @mdo ?

patrickhlauke:
@mdo i think at this stage we probably can't rely on pure CSS to have both the "just hang off the `:valid` / `:invalid`" AND avoiding that validation tooltips get announced right away if we were properly using `aria-describedby`, as the `display:none` won't prevent them from being announced. feels like we may need an actual plugin/JavaScript behavior for form validations...thoughts? and if so, any chance of a pointer on how to go about this?

patrickhlauke:
alternatively i guess we *could* cram all the extra logic into that example JS snippet we provide, but that feels very dirty, expecting users to manually copy/paste some custom JS in to provide actual core functionality that should be handled directly by BS silently?

patrickhlauke:
xref https://github.com/twbs/bootstrap/pull/28995 for a first very naive approach

patrickhlauke:
[...]

PR #28995, opened 2019-07-07
WIP Initial naive form validation fix

patrickhlauke:
A first attempt at mocking up the actual JavaScript behavior for correct custom form validation handling

Not mergeable yet, just pushing this as a proof of concept. This should really be rolled into a proper plugin/JavaScript behavior.

x-ref #28414

### DIFF
--- a/site/content/docs/4.3/components/forms.md
+++ b/site/content/docs/4.3/components/forms.md
@@ -745,17 +745,13 @@ Custom feedback styles apply custom colors, borders, focus styles, and backgroun
   <div class="form-row">
     <div class="col-md-4 mb-3">
       <label for="validationCustom01">First name</label>
-      <input type="text" class="form-control" id="validationCustom01" value="Mark" required>
-      <div class="valid-feedback">
-        Looks good!
-      </div>
+      <input type="text" class="form-control" id="validationCustom01" data-validation-container="validationCustom01-feedback" data-validation-valid="Looks good!" value="Mark" required>
+      <div class="form-validation-feedback" id="validationCustom01-feedback"></div>
     </div>
     <div class="col-md-4 mb-3">
       <label for="validationCustom02">Last name</label>
-      <input type="text" class="form-control" id="validationCustom02" value="Otto" required>
-      <div class="valid-feedback">
-        Looks good!
-      </div>
+      <input type="text" class="form-control" id="validationCustom02" data-validation-container="validationCustom02-feedback" data-validation-valid="Looks good!"  value="Otto" required>
+      <div class="form-validation-feedback" id="validationCustom02-feedback"></div>
     </div>
     <div class="col-md-4 mb-3">
       <label for="validationCustomUsername">Username</label>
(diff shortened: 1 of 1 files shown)


patrickhlauke:
@mdo i remembered you discussed potential use of data attributes for this ... here's the first naive stab at it. apart from the general ugliness of the code, one aspect that's still not ideal is that the feedback does show dynamically based on validity of the form control, meaning that the problem will still happen that if there was an initial feedback which is then tied to the control, if the validity changes dynamically (after the first submit), the feedback may be hidden, but it's still associated via `aria-describedby` and announced by AT. or it *wants* to show another feedback now, but it can't because the JS doesn't kick in. we likely need to also hook into the change event of each form control to trigger the whole processing dynamically (to keep up with the CSS that is evaluated dynamically)...

patrickhlauke:
for this proof of concept, only fiddled with that first form at https://deploy-preview-28995--twbs-bootstrap.netlify.com/docs/4.3/components/forms/#custom-styles

MartijnCuppens:
I like the idea. I would also provide an example with as well the `data-validation-valid` as `data-validation-invalid` attributes.

Johann-S:
Like I said in Slack, I think it's a good idea but for me, it lacks of features, just accessibility, and a simple validation isn't enough for a core components

patrickhlauke:
For me, the fact the validation errors aren't actually conveyed properly is a blocker. Annoyed I've not caught that sooner in v4, but...it is what it is.

mdo:
Given #31666 and #31677, thinking we can isolate the JS changes here in another PR?

XhmikosR:
Does this still apply to v5?

patrickhlauke:
> Does this still apply to v5?

yes, https://v5.getbootstrap.com/docs/5.0/forms/validation/ still has issues - custom styles and tooltips examples, the errors are not associated with the form controls and not announced by AT. also, in the support elements example, the errors are likewise not associated. for custom and tooltip, this needs javascript to dynamically inject the error, OR to dynamically add/remove the `aria-describedby` (as if it's there from the start, even when it's set to `display:none`, AT will announce the error otherwise)

mdo:
This one is pretty out of date, is it still relevant?

patrickhlauke:
the original problem still persists, but this got scuppered by my lack of JS chops / how to make the scripting etc apply automagically.

XhmikosR:
@patrickhlauke can you explain a bit more in detail what you are missing? Maybe @GeoSot could help us out here.

patrickhlauke:
> @patrickhlauke can you explain a bit more in detail what you are missing? Maybe @GeoSot could help us out here.

what's really missing is some kind of clean way to actually make this part of Bootstrap's JS. how to generalise this somehow to make it easily usable, automatically adding/removing the `aria-describedby` ... ideally, somebody who's familiar with BS' slightly over-complicated JS way of doing things. maybe @rohit2sharma95 can have a look as well?

patrickhlauke:
Superseded by https://github.com/twbs/bootstrap/pull/34043

PR #29003, opened 2019-07-09
Move the linkinator test in `npm t`.

XhmikosR:
Only thing I'm not sure about is running these in parallel. Some more testing should help decide.


MartijnCuppens:
> Only thing I'm not sure about is running these in parallel.

What could be a problem here?

XhmikosR:
The whole thing being slower compared to running the tests sequentially :P

I doubt it, but I could use more testing from others.

XhmikosR:
Can someone test this?

Johann-S:
it works fine to me 👍

MartijnCuppens (approved):

PR #29006, opened 2019-07-10
Make tabs ARIA accessible.

maedi:
Bootstrap Tabs almost conforms to the ARIA spec for tabs but not quite:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role#Description

To match the spec the following changes have been made:
- Change the `tab` from using `aria-expanded` to `aria-selected`
- Connect to `tabpanel`s via the `tab`'s `aria-controls` value
- Change the `tabpanel` from using an `.active` css class to `aria-expanded` (in addition)

The HTML will need to ensure that their `tab`'s `aria-controls` and `tabpanel`'s `id` match.

### DIFF
--- a/js/tab.js
+++ b/js/tab.js
@@ -79,8 +79,8 @@
         .find('[data-toggle="tab"]')
         .attr('aria-selected', false)
       // Deactive previous tabpanel.
-      var tabpanel = $('#' + $($active).first().find('a').attr('aria-controls'));
-      tabpanel.attr('aria-expanded', false);
+      var $active_tabpanel = $('#' + $($active).first().find('a').attr('aria-controls'));
+      $active_tabpanel.attr('aria-expanded', false);
 
       // Activate next tab.
       element
@@ -88,8 +88,8 @@
         .find('[data-toggle="tab"]')
         .attr('aria-selected', true)
       // Activate next tabpanel.
-      var tabpanel = $('#' + $(element).first().find('a').attr('aria-controls'));
-      tabpanel.attr('aria-expanded', true);
+      var element_tabpanel = $('#' + $(element).first().find('a').attr('aria-controls'));
+      element_tabpanel.attr('aria-expanded', true);
 
       // Animate transition.
       if (transition) {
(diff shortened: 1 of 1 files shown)


XhmikosR:
@mdo @Johann-S @patrickhlauke: this is for v3... We *could* merge it if we are going to release v3.4.2.

patrickhlauke:
open `tabpanel` doesn't need nor benefit from `aria-expanded`, since it's completely hidden from the accessibility tree when not open

js/tab.js
```diff
@@ -70,26 +70,36 @@
...
         .removeClass('active')
         .find('> .dropdown-menu > .active')
         .removeClass('active')
         .end()
         .find('[data-toggle="tab"]')
-        .attr('aria-expanded', false)
+        .attr('aria-selected', false)
+      // Deactive previous tabpanel.
+      var $activeTabpanel = $('#' + $($active).first().find('a').attr('aria-controls'));
+      $activeTabpanel.attr('aria-expanded', false);
```
patrickhlauke: `aria-expanded` is unnecessary on the tab panel

js/tab.js
```diff
@@ -70,26 +70,36 @@
...
 
+      // Activate next tab.
       element
         .addClass('active')
         .find('[data-toggle="tab"]')
-        .attr('aria-expanded', true)
+        .attr('aria-selected', true)
+      // Activate next tabpanel.
+      var elementTabpanel = $('#' + $(element).first().find('a').attr('aria-controls'));
+      elementTabpanel.attr('aria-expanded', true);
```
patrickhlauke: `aria-expanded` is unnecessary on the tab panel

patrickhlauke (changes requested):
In short, the only minimal change from this PR that is required is switching from using `aria-expanded` (which was wrongly used in v3) to `aria-selected`

(there's more fundamental things that make our tab implementation not quite compliant, like lack of keyboard support, but we're tackling those for v5 with potential v4 backport)

Johann-S:
I'm not in favor of merging that, because v3 is EOL

mdo:
v3 is EOL, we shouldn't be shipping anything else there.

PR #29009, opened 2019-07-10
Update devDependencies.

XhmikosR:


XhmikosR:
@Johann-S it seems we sill have issues with the newer babel versions. I'll pin them for now but we should report it upstream.

Johann-S (approved):

PR #29017, opened 2019-07-11
Offcanvas as component

### LINKED ISSUE #24718: Drawer, sidebar, offcanvas nav

### LINKED ISSUE #27033: Feature request: side Modals

GeoSot:
I did a first approach on an offcanvas component.
Αll of us, have used custom solutions, but maybe an official version,  would be more stable.

Do you find it interesting? May I continue on this?

Preview: <https://deploy-preview-29017--twbs-bootstrap.netlify.app/docs/5.0/components/offcanvas/>

TODO:

- [x] we need to see if we can stop body scroll
- [x] maybe blur the bg if it doesn't have any performance hit
- [x] tweak blur, the current blur is too much IMO (XhmikosR)
- [x] Remove height from bottom offcanvas
- [x] Provide some default theming (white background, maybe a border or shadow like the modals)
- [x] Disable body scrolling by default.
- [x] measure FPS when blurring the bg (@XhmikosR  i need guidance)
- [x] update offcanvas example to use this (also move it away from the experimental category)
- [x] reindent and fix missing end tags in the doc examples
- [x] switch to `event.key`
- [x] add more content in right, bottom examples
- [x] `<div class="h4">` should be changed to `h4` or similar?
- [x] Make sure the element is hidden and not offscreen when inactive 
- [x] enforce focus, only on open element
- [x] Enable blur by default?
- [x] Revert the scrollbar decoupling
- [x] Consistent behavior of modals and offcanvas (eg. same way of setting a backdrop) #32439
- [x] [non-text elements inside button](https://github.com/twbs/bootstrap/pull/29017#issuecomment-706302016) related issue #32372, #31646
- [x] Make sure all data attributes and any CSS vars are prefixed
- [x] Bring back the backdrop option, but keep it enabled by default
- [x] Make sure we aren't using any recently changed mr, etc classes
- [ ] Should the backdrop transition be the same as the offcanvas one? Right now, backdrop shows instantly but finishes hiding after the offcanvas is closed .(GeoSot: it will be changed when #32439 is merged)
- [x] Decide about the name: `offcanvas`, `off-canvas` etc
- [x] Maybe rename the classes to consider RTL? Not sure if it's needed
- [x] Fix BrowserStack tests
- [x] spell checking from a native English speaker (@mdo?)
- [x] check accessibility (@patrickhlauke)
- [ ] **clean up/squash patches manually**

Fixes #27033, fixes #24718.

---

mdo's TODO:

- [x] Finish and merge this PR
  - [x] Undo the Offcanvas example change for now (we'll replace that later)
- [x] Close any other PRs
- [x] Iterate on the docs a bit more
- [x] Detach and de-dupe the backdrop in #32439


js/src/offcanvas.js
```diff
@@ -0,0 +1,194 @@
+/**
+ * --------------------------------------------------------------------------
+ * Bootstrap (v4.3.1): offcanvas.js
+ * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
+ * --------------------------------------------------------------------------
+ */
+
+import { getSelectorFromElement, jQuery as $ } from './util/index'
+
```
Johann-S: trailing line

js/src/offcanvas.js
```diff
@@ -0,0 +1,194 @@
...
+  SHOW: 'show'
+}
+
+/**
+ * ------------------------------------------------------------------------
+ * Class Definition
+ * ------------------------------------------------------------------------
+ */
+
+class Offcanvas {
```
Johann-S: I would name it: `OffCanvas` 🤔 but maybe I'm wrong ^^

js/src/offcanvas.js
```diff
@@ -0,0 +1,194 @@
...
+    this._element.classList.remove(ClassName.SHOW)
+    this._element.setAttribute('aria-hidden', true)
+    this._element.removeAttribute('aria-offcanvas')
+
+    EventHandler.trigger(this._element, Event.HIDDEN)
+  }
+
+  // Static
+
+  static _getParentFromElement(element) {
```
Johann-S: No need for a static method just a private one is enough

js/src/offcanvas.js
```diff
@@ -0,0 +1,194 @@
...
+  constructor(element) {
+    this._element = element
+    this._isShown = element.classList.contains(ClassName.SHOW)
+    if (this._element) {
+      Data.setData(element, DATA_KEY, this)
+    }
+
+    EventHandler.on(this._element, Event.CLICK_DISMISS, Selector.DATA_DISMISS, event => this.hide(event))
+    EventHandler.on(document, Event.CLICK_DATA_API, event => {
+      const target = Offcanvas._getParentFromElement(event.target)
```
Johann-S: I'm more in favor of adding a css class to the body element to detect when a OffCanvas is open, it'll be faster to check if the body contains this class, and if it's the case we can try to figure out the parent of the target.

js/src/offcanvas.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  static _jQueryInterface(config) {
+    return this.each(function () {
+      let data = Data.getData(this, DATA_KEY)
+
+      if (!data) {
+        data = new Offcanvas(this)
+      }
+
```
Johann-S: it lacks of the check for undefined methods, it's better to allow jQuery folks to call other methods too
```js
    if (typeof config === 'string') {
      if (typeof data[config] === 'undefined') {
        throw new TypeError(`No method named "${config}"`)
      }

      data[config]()
    }

**Review by Johann-S — CHANGES_REQUESTED** [review]
Hi @GeoSot it's very interesting thanks! 👍 
We are totally interested in this plugin

BTW you'll have to add unit tests

**Comment by mdo** [comment]
Awesome! Haven’t dived into the code yet m, but obviously there’s a huge need for this. Can we ensure we have support for left, right, and bottom slide-in? Can we also make sure to stop the body scroll behind the sidebar?

**Comment by patrickhlauke** [comment]
a crucial aspect of a properly working off-canvas is that focus should move to the offcanvas, remain trapped inside it until it's closed, and ideally the underlying document should be hidden via `aria-hidden="true"`

**Review by patrickhlauke — CHANGES_REQUESTED** [review]
keyboard and assistive technology behavior must be addressed before this can be considered. conceptually, the off-canvas behaves pretty much like a dialog.

File: js/src/offcanvas.js (hunk 4 above)
Comment (GeoSot) [reply js/src/offcanvas.js]: Please be comfortable, to make corrections, as you have better knowledge of the code functionalities

File: js/src/offcanvas.js (hunk 5 above)
Comment (GeoSot) [reply js/src/offcanvas.js]: I think it's ok now

**Comment by GeoSot** [comment]
Hello Guys. 

I really love this framework, and I want to contribute.

As you can see,  I am a bit slower than the requested changes, and really I don't have any experience on unit testing.
So I would appreciate any help

**Comment by Johann-S** [comment]
@GeoSot you should first focus on request changes and after I'll help you for the unit tests don't worry 😉

**Comment by MartijnCuppens** [comment]
> I don't have any experience on unit testing

Just have a look at the existing tests (`js/tests/unit`) and try to figure out how it works (for me, that worked 😛)

Other concerns:
- I wouldn't nest the selectors (like `&-right`), we don't do that either for our modals. The full classname also makes searching for classes easier.
- I would provide some default theming (like a default background) and create a theming section on the page with the colored components.
- I spotted some inline styles like `style="z-index: 1072;"`, we should avoid that.
- Accessibility wise I would like to close the drawer when I press <kbd>ESC</kbd> key.

**Comment by GeoSot** [comment]
>     * I spotted some inline styles like `style="z-index: 1072;"`, we should avoid that.
Only because the doc header has 1071 z-index

**Comment by Johann-S** [comment]
Hi @GeoSot, do not hesitate to hit me up if you're stuck somewhere or if you did everything you planed to 😉

**Comment by GeoSot** [comment]
Thank u @Johann-S 

You may have a look on  Js code and leave me some feedback. (Feel free to change things)

File: js/src/offcanvas.js [hunk 6]
```diff
@@ -0,0 +1,208 @@
...
+  // Private
+  _enforceFocus() {
+    const savedTabIndex = this._element.getAttribute('tabindex')
+    this._element.setAttribute('tabindex', '0')
+    this._element.focus()
+    this._element.setAttribute('tabindex', savedTabIndex || 1)
+  }
+
+  _handleClosing() {
```
Johann-S: I would call it: `_addEventListener`

js/src/offcanvas.js
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/**
+ * ------------------------------------------------------------------------
+ * Data Api implementation
+ * ------------------------------------------------------------------------
+ */
+
+EventHandler.on(document, Event.CLICK_DATA_API, Selector.DATA_TOGGLE, function (event) {
+  if (['A', 'AREA'].includes(this.tagName)) {
```
Johann-S: `includes` isn't supported by the browsers we target 😟 you can use `.indexOf` instead

Johann-S:
@GeoSot we switched from QUnit to Jasmine, so you'll have to update your branch and use Jasmine to write your unit tests, do not hesitate to take a look at other plugins

js/src/offcanvas.js
GeoSot: I didn't got the point...

Johann-S:
Hi @GeoSot, I improved your unit tests, so the JavaScript part should be ok 👍 now we have to check the docs and maybe the SCSS ? @twbs/css-review

scss/_offcanvas.scss
```diff
@@ -0,0 +1,40 @@
...
+  .offcanvas:not(.offcanvas-bottom) {
+    width: $offcanvas-horizontal-width;
+  }
+  .offcanvas-bottom {
+    height: $offcanvas-vertical-height;
+  }
+}
+
+.offcanvas.show {
+  transform: translate(0, 0);
```
MartijnCuppens: This can be simplified:
```suggestion
  transform: none;
```

MartijnCuppens:
> I preferred to keep it simple and focus on functionality and minimalism. If it succeed to be merged we can add some styling

Our other components also work out of the box without additional utilities for backgrounds (like the modal). For consistency, I would also like to see this here.

We should also ditch the heights for the bottom variant. Also there's some visible outline when the component is focused which need to be removed, like we do with the modals.

This PR also needs an approval from @patrickhlauke, to make sure everything is accessible.

XhmikosR:
Before the final merge, I'd like to clean up/squash any patches manually.

js/src/offcanvas/offcanvas.js
```diff
@@ -0,0 +1,205 @@
...
+    this._element.removeAttribute('aria-hidden')
+    this._element.setAttribute('aria-expanded', true)
+    this._element.setAttribute('aria-offcanvas', true)
+
+    setTimeout(() => {
+      this._enforceFocus()
+
+      EventHandler.trigger(this._element, Event.SHOWN, { relatedTarget })
+      document.body.classList.add(ClassName.OPEN)
+    }, getTransitionDurationFromElement(this._element) + 9)
```
XhmikosR: What's this magic number `9`?

js/src/offcanvas/offcanvas.js
```diff
@@ -0,0 +1,205 @@
...
+      EventHandler.trigger(this._element, Event.HIDDEN, { relatedTarget })
+    }, getTransitionDurationFromElement(this._element))
+  }
+
+  // Private
+  _enforceFocus() {
+    const savedTabIndex = this._element.getAttribute('tabindex')
+    this._element.setAttribute('tabindex', '0')
+    this._element.focus()
+    this._element.setAttribute('tabindex', savedTabIndex || 1)
```
XhmikosR: Shouldn't this be `'1'`?

XhmikosR:
Also, we need to see if we can stop body scroll like @mdo said above https://github.com/twbs/bootstrap/pull/29017#issuecomment-510986442

And maybe blur the bg if it doesn't have any performance hit.

GeoSot:
> Our other components also work out of the box without additional utilities for backgrounds (like the modal). For consistency, I would also like to see this here.

Maybe like the [navBar](https://getbootstrap.com/docs/4.3/components/navbar/#color-schemes)?  Honestly, I don't mind, if anybody wants to commit some styling changes, but anyway I cannot make this choice by myself


>What's this magic number 9?

Was the ugly way to keep the 'offcanvas-open' class, on body while toggling two different elements. I 've made some changes (maybe will need some refactoring)

XhmikosR:
Add a TODO list in the first post.

GeoSot:
@XhmikosR is offcanvas example needed anymore? 
With offcanvas as a basic component, maybe the example is redundant

XhmikosR:
You have a point there @GeoSot, but it's nice to have it as an individual example like we have navs etc.

XhmikosR:
Alright, I rebased this and cleaned up the patches.

Only a few things to do, and then another round of reviews.

XhmikosR:
The examples need to be reindended BTW.

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,200 @@
+---
+layout: docs
+title: Offcanvas
+description: Toggle the visibility sidebars across your project with a few classes and our JavaScript plugins.
```
XhmikosR: This needs some rewording. Also `plugins` -> `plugin`.

scss/_offcanvas.scss
```diff
@@ -0,0 +1,69 @@
...
+  right: 0;
+  left: auto;
+  transform: translateX(100%);
+}
+
+.offcanvas-bottom {
+  top: auto;
+  right: 0;
+  width: 100%;
+  height: $offcanvas-vertical-height-mobile;
```
MartijnCuppens: I would remove the height, it makes more sense if the component scales to the content

GeoSot: I will leaf a max height for safety reasons. ok?

GeoSot:
@MartijnCuppens  I disagree with the "Provide some default theming (white background, maybe a border or shadow like the modals)" todo. 
I would like an agnostic component, and I am trying to save, some bytes, but as you insist, I added color, shadow etc

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: docs
 title: Offcanvas
-description: Toggle the visibility sidebars across your project with a few classes and our JavaScript plugins.
+description: Toggle the visibility sidebars across your project with a few classes and our JavaScript plugin.
```
XhmikosR: `visibility sidebars` doesn't seem right. Not sure how to reword this.

site/content/docs/4.3/examples/offcanvas/index.html
```diff
@@ -63,7 +60,7 @@
 
 <main role="main" class="container">
   <div class="d-flex align-items-center p-3 my-3 text-white-50 bg-purple rounded shadow-sm">
-    <img class="mr-3" src="/docs/{{< param docs_version >}}/assets/brand/bootstrap-outline.svg" alt="" width="48" height="48">
+    <img class="mr-3" src="/docs/{{< param docs_version >}}/assets/brand/bootstrap-outline.svg" alt="" width="48" height="48"/>
```
XhmikosR: This is redundant, end tags are useless in HTML 5 :P

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,200 @@
...
+  <a class="btn btn-primary" data-toggle="offcanvas" href="#offcanvasExample" role="button" aria-expanded="false" aria-controls="offcanvasExample">
+    Link with href
+  </a>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample" aria-expanded="false" aria-controls="offcanvasExample">
+    Button with data-target
+  </button>
+</p>
+<div class="offcanvas offanvas-push-body bg-dark text-white" tabindex="-1" id="offcanvasExample">
+  <div class="offcanvas-header">
+     <div class="h4 col text-center">Offcanvas</div>
```
XhmikosR: These examples need reindendation @GeoSot. All of them in this page.

It seems a `</div>` is missing somewhere in all of them.

MartijnCuppens:
> @MartijnCuppens
I disagree with the "Provide some default theming (white background, maybe a border or shadow like the modals)" todo.
I would like an agnostic component, and I am trying to save, some bytes, but as you insist, I added color, shadow etc

You definitely got a point about saving some bytes, but those variables can be set to `null` if people really want to generate as less CSS possible. We also got to be consistent with the other components (modals, cards, toasts), they have some basic styling applied to.

@twbs/css-review, what do you guys think?

GeoSot:
A new thought....... It could be responsive.  

Ex: .offcanvas-sm => hide sidebar bellow 'sm' breakpoint

XhmikosR:
@GeoSot please rebase and fix the linting errors (i.e don't use a section)

site/content/docs/4.3/examples/offcanvas/index.html
```diff
@@ -61,77 +58,91 @@
...
 
-<main role="main" class="container">
-  <div class="d-flex align-items-center p-3 my-3 text-white-50 bg-purple rounded shadow-sm">
-    <img class="mr-3" src="/docs/{{< param docs_version >}}/assets/brand/bootstrap-outline.svg" alt="" width="48" height="48">
-    <div class="lh-1">
-      <h6 class="mb-0 text-white lh-1">Bootstrap</h6>
-      <small>Since 2011</small>
-    </div>
+<main role="main" class="container pt-5 d-flex flex-column align-items-center">
+  <div class="section">
```
XhmikosR: Why is this needed? Just use our utils and drop the section divs

juberahme (approved):

XhmikosR:
Alright, I had some time to do another quick review.

First of all, I want to say thanks for sticking with the PR. We are low on resources and we have so many things to do. I definitely want this added for v5, so I'll do my best to move forward with this.

1. you shouldn't have squashed the patches. It's not very ethical IMO. But what's done is done :)
2. the offcanvas example's mobile menu doesn't work anymore
3. the "blur" implementation isn't really a blur, but a `background-color`; we should be able to use `filter: blur(foo)`, /CC @MartijnCuppens 
4. when one clicks the right offcanvas and doesn't close it, and then clicks the bottom one, the bottom one shows up and the right one is automatically closed; I'm not sure if this is right, we need more people to chime in
5. we definitely need spell checking from a native English speaker (ping @mdo)
6. accessibility-wise, does this LGTY @patrickhlauke?

MartijnCuppens:
> the "blur" implementation isn't really a blur, but a background-color; we should be able to use filter: blur(foo), /CC @MartijnCuppens

The `filter` property can perform poorly when animated and doesn't work in IE. We might better keep the implementation but rename this to `backdrop`, just like the modals.

We should also keep the backdrop color the same as the modal for more consistency. Also the demos look a bit too colorful, just one example on how to change the background color should be enough. I also don't like the outlined focus state of the modal (Chrome), we should remove this (like we do with the modals).

XhmikosR:
Doesn't seem the blur is animated, though.

thehitechpanky:
Is this planned for v4 as well?

MartijnCuppens:
> Is this planned for v4 as well?

We're not planning on adding new features for v4 (bugfixes only)

GeoSot:
Ok Let's sum up.

I have updated the 'todos' on first message and I have to work the following


-    "the offcanvas example's mobile menu doesn't work anymore" 
It's true. in order to re-create the same example, the 'offcanvas' component has to be breakpoint -activated or to override it through css. Which approach should I follow?    **(need decision to continue)**

-  "blur" implementation has to be renamed to "backdrop" and try to keep a consistency with modal's implementation

-    we have to decide, the offcanvas behavior when another one opens **(need decision to continue)**

MartijnCuppens:
@GeoSot,

Please also address these issues I mentioned:
> We should also keep the backdrop color the same as the modal for more consistency. Also the demos look a bit too colorful, just one example on how to change the background color should be enough. I also don't like the outlined focus state of the modal (Chrome), we should remove this (like we do with the modals).

> we have to decide, the offcanvas behavior when another one opens (need decision to continue)

We don't support this with our modals. If we are going to support this, imo, the most logic approach is close the current offcanvas and open the next one. Supporting two open offcanvas elements will be an accessibility hell I guess.

GeoSot:
Hello guys. May I have some feedback to keep the request rolling?

@patrickhlauke can you help with accessibility (todo on first post)?
@XhmikosR  do I have to do something with the  FPS measurement?
@MartijnCuppens any comment after the last changes?

XhmikosR:
@GeoSot sorry for the radio silence.

Please bear with us, I'm trying to get everyone together so that we can move faster. Don't let this get you down, I hope we'll manage to regroup soon.

MartijnCuppens:
@GeoSot, Sure:

- The blue outline still feels weird (chrome):
image
- The first example of the offcanvas can be opened with a `<a>`. I'm not sure that's a good idea accessibility-wise (ping @patrickhlauke)
- I don't think it's a good idea to use that many background colors in our examples. Maybe it's better to bundle colored examples under the `Color schemes
` section?
- There's still some _blur_ copy left in the docs

GeoSot:
@MartijnCuppens 

> The blue outline still feels weird (chrome):

I think the blue outline is because of  :focus styling (which is not mine). I have removed the outlines & shadows.  Do you want to override the default focus??

>The first example of the offcanvas can be opened with a <a>. I'm not sure that's a good idea accessibility-wise 

I tried to copy the other components behavior. So yes, if we want, I can remove it

>I don't think it's a good idea to use that many background colors in our examples. Maybe it's better to bundle colored examples under the Color schemes section?

I will handle it. I have changed only the example. my mistake ;)

MartijnCuppens:
> I think the blue outline is because of :focus styling (which is not mine). I have removed the outlines & shadows. Do you want to override the default focus??

Indeed, those are browser defaults. We also remove them for the modals, so let's get rid of it.

> > The first example of the offcanvas can be opened with a . I'm not sure that's a good idea accessibility-wise.

> I tried to copy the other components behavior. So yes, if we want, I can remove it

Leave it for now, unless @patrickhlauke has a strong opinion about this.

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+    </tr>
+    <tr>
+      <td>hidden.bs.offcanvas</td>
+      <td>This event is fired when a offcanvas element has been hidden from the user (will wait for CSS transitions to complete).</td>
+    </tr>
+  </tbody>
+</table>
+
+{{< highlight js >}}
+var myOffcanvas = document.getElementById('myCollapsible')
```
leomp12:
```suggestion
var myOffcanvas = document.getElementById('myOffcanvas')
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+myCollapsible.addEventListener('hidden.bs.offcanvas', function () {
```
leomp12:
```suggestion
myOffcanvas.addEventListener('hidden.bs.offcanvas', function () {
```

leomp12 (commented):
Hey @GeoSot ,
Just leaving minor fixes on last JS sample snippet (docs)

hassanbinsafdar:
if its possible to make off-canvas with body animate to right like reveal

scss/_offcanvas.scss
```diff
@@ -0,0 +1,74 @@
...
+  left: 0;
+  z-index: $zindex-offcanvas;
+  width: $offcanvas-horizontal-width-mobile;
+  max-width: 100%;
+  padding: $offcanvas-padding-y $offcanvas-padding-x;
+  overflow-y: auto;
+  color: $offcanvas-color;
+  background-color: $offcanvas-bg-color;
+  border: $offcanvas-border-width solid $offcanvas-border-color;
+  // Remove focus outline from opened offcanvas
```
mdo:
```suggestion
```

site/assets/scss/docs.scss
```diff
@@ -51,3 +51,8 @@
 @import "syntax";
 @import "anchor";
 @import "algolia";
+
+
+.offcanvas{
+  z-index: 1072 // over .bd-subnavbar
+}
```
mdo: Move these styles to the appropriate section in `_component-examples.scss`, please. Also remember to add a space before the opening curly brace.

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
+---
+layout: docs
+title: Offcanvas
+description: Toggle the visibility sidebars across your project with a few classes and our JavaScript plugin.
```
mdo:
```suggestion
description: Build hidden sidebars into your project for navigation, shopping carts, and more with a few classes and our JavaScript plugin.
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+group: components
+toc: true
+---
+
+## How it works
+
+The offcanvas JavaScript plugin is used to show and hide a sidebar (or bottom-bar). Buttons or anchors are used as triggers, to specific elements you toggle.
```
mdo:
```suggestion
The offcanvas JavaScript plugin shows and hides sidebar on the left, right, or bottom of your viewport. Buttons or anchors are used as triggers that are attached to specific elements you toggle.
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+Given how CSS handles animations, you cannot use `margin` & `translate` on a `.offcanvas` element. Use the class on an independent wrapping element.
```
mdo:
```suggestion
Given how CSS handles animations, you cannot use `margin` or `translate` on a `.offcanvas` element. Instead, use the class as an independent wrapping element.
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Example
+
+Click the buttons below to show and hide another element via class changes:
```
mdo:
```suggestion
Click the buttons below to show and hide an offcanvas element via class changes:
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  <a class="btn btn-primary" data-toggle="offcanvas" href="#offcanvasExample" role="button" aria-expanded="false" aria-controls="offcanvasExample">
+    Link with href
+  </a>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample" aria-expanded="false" aria-controls="offcanvasExample">
+    Button with data-target
+  </button>
+</p>
+<div class="offcanvas offanvas-push-body bg-dark text-white" tabindex="-1" id="offcanvasExample">
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Offcanvas</div>
```
mdo:
```suggestion
    <div class="h4">Offcanvas</div>
```

scss/_variables.scss
```diff
@@ -1224,6 +1225,25 @@ $close-color:                       $black !default;
 $close-text-shadow:                 0 1px 0 $white !default;
 
 
+//Offcanvas
+
+$offcanvas-padding-y:               $modal-inner-padding !default;
+$offcanvas-padding-x:               $modal-inner-padding * 2 !default;
```
mdo:
```suggestion
$offcanvas-padding-x:               $modal-inner-padding !default;
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+{{< example >}}
+<p>
+  <a class="btn btn-primary" data-toggle="offcanvas" href="#offcanvasExample" role="button" aria-expanded="false" aria-controls="offcanvasExample">
+    Link with href
+  </a>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample" aria-expanded="false" aria-controls="offcanvasExample">
+    Button with data-target
+  </button>
+</p>
+<div class="offcanvas offanvas-push-body bg-dark text-white" tabindex="-1" id="offcanvasExample">
```
mdo: `offanvas-push-body`?

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Offcanvas</div>
+    <button type="button" class="close text-reset" data-dismiss="offcanvas" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+  <div class="card card-body text-dark ">
```
mdo: I'd suggest nixing the bulk of this custom content and replacing it with something more functional and expected:

- Navigation elements (perhaps a stacked pill nav or a list group?)
- Simple `<p>`s of content

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+      <li><a class="dropdown-item" href="#">Action</a></li>
+      <li><a class="dropdown-item" href="#">Another action</a></li>
+      <li><a class="dropdown-item" href="#">Something else here</a></li>
+    </ul>
+  </div>
+</div>
+{{< /example >}}
+
+## Position Modifiers
+
```
mdo: Add an intro line:

> Change the placement of an offcanvas element with modifier classes:

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+    <ul class="dropdown-menu" aria-labelledby="dropdownMenuButton">
+      <li><a class="dropdown-item" href="#">Action</a></li>
+      <li><a class="dropdown-item" href="#">Another action</a></li>
+      <li><a class="dropdown-item" href="#">Something else here</a></li>
+    </ul>
+  </div>
+</div>
+{{< /example >}}
+
+## Position Modifiers
```
mdo: We use sentence case in Bootstrap, so this would be `Position modifiers`, but I think we can leave just as `Position` for brevity's sake.

```suggestion
## Position
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `.offcanvas-right` modifier, places offcanvas on the right side
```
mdo:
```suggestion
- `.offcanvas-right` places offcanvas on the right of the viewport
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+- `.offcanvas-bottom` modifier, places offcanvas on the bottom of the screen
```
mdo:
```suggestion
- `.offcanvas-bottom` places offcanvas on the bottom of the viewport
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+{{< example >}}
+<p>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample2" aria-expanded="false" aria-controls="offcanvasExample2">Toggle Right offcanvas</button>
```
mdo:
```suggestion
  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample2" aria-expanded="false" aria-controls="offcanvasExample2">Toggle right offcanvas</button>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample3" aria-expanded="false" aria-controls="offcanvasExample3">Toggle Bottom offcanvas</button>
```
mdo:
```suggestion
  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample3" aria-expanded="false" aria-controls="offcanvasExample3">Toggle bottom offcanvas</button>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+</p>
+<div class="offcanvas bg-dark text-white offcanvas-right" tabindex="-1" id="offcanvasExample2">
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Offcanvas Right Side</div>
```
mdo:
```suggestion
    <div class="h4">Offcanvas right</div>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+    <button type="button" class="close text-reset" data-dismiss="offcanvas" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+</div>
+<div class="offcanvas bg-dark text-white offcanvas-bottom" tabindex="-1" id="offcanvasExample3">
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Offcanvas Bottom</div>
```
mdo:
```suggestion
    <div class="h4">Offcanvas bottom</div>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+    <button type="button" class="close text-reset" data-dismiss="offcanvas" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+</div>
+{{< /example >}}
+
+## Color schemes
+
+Theming offcanvas is very easy by using our [color utilities]({{< docsref "/utilities/colors" >}}).
```
mdo:
```suggestion
Easily style an offcanvas element with a different `background-color` or `color` with our [color utilities]({{< docsref "/utilities/colors" >}}).
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+Any combination of `bg-*` and `text-{color}` utilities, can give a different result
```
mdo: Safe to delete this, this is self explanatory.

```suggestion
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+{{< example >}}
+<p>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample31" aria-expanded="false" aria-controls="offcanvasExample31">Colored Offcanvas</button>
```
mdo:
```suggestion
  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasColored" aria-expanded="false" aria-controls="offcanvasColored">Colored offcanvas</button>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+</p>
+<div class="offcanvas bg-primary text-white" data-body="scroll" tabindex="-1" id="offcanvasExample31">
```
mdo:
```suggestion
<div class="offcanvas bg-primary text-white" data-body="scroll" tabindex="-1" id="offcanvasColored">
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Hey I got New Colors</div>
```
mdo:
```suggestion
    <div class="h4">Colored offcanvas</div>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  </div>
+</div>
+
+{{< /example >}}
+
+## Body Options
```
mdo:
```suggestion
## Options
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+By default, when an offcanvas is visible, body cannot be scrolled. You can use the following data-options:
```
mdo:
```suggestion
By default, when an offcanvas is visible, the `<body>` of your page cannot be scrolled. You can use the following data-options to change this behavior:
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `data-body="scroll"` enables scrollBar on body when offcanvas is open
```
mdo:
```suggestion
- `data-body="scroll"` enables scrolling on the `<body>` when offcanvas is open
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+- `data-body="backdrop"` hides scrollBar and creates a backdrop on body, when offcanvas is open
```
mdo:
```suggestion
- `data-body="backdrop"` disables scrolling and creates a backdrop over the `<body>` when offcanvas is open
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+{{< example >}}
+<p>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample4" aria-expanded="false" aria-controls="offcanvasExample4">Enable body scrolling </button>
+  <button class="btn btn-primary" type="button" data-toggle="offcanvas" data-target="#offcanvasExample5" aria-expanded="false" aria-controls="offcanvasExample5">Body backdrop</button>
+</p>
+<div class="offcanvas bg-dark text-white" data-body="scroll" tabindex="-1" id="offcanvasExample4">
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Wow, body scroller is with us again</div>
```
mdo:
```suggestion
    <div class="h4">Offcanvas with scrolling</div>
    <p>Try scrolling the rest of the page to see this option in action.</p>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  </div>
+</div>
+<div class="offcanvas bg-dark text-white" data-body="backdrop" tabindex="-1" id="offcanvasExample5">
+  <div class="offcanvas-header">
+    <div class="h4 col text-center">Body is covered by a backdrop</div>
```
mdo:
```suggestion
    <div class="h4">Offcanvas with backdrop</div>
    <p>Try clicking on the page's content to toggle the offcanvas.</p>
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+  </div>
+</div>
+{{< /example >}}
+
+## Accessibility
+
+Make sure to add `aria-expanded` to the control element. This attribute explicitly conveys the current state of the collapsible element tied to the control to screen readers and similar assistive technologies. If the collapsible element is closed by default, the attribute on the control element should have a value of `aria-expanded="false"`. If you've set the collapsible element to be open by default using the `show` class, set `aria-expanded="true"` on the control instead. The plugin will automatically toggle this attribute on the control based on whether or not the collapsible element has been opened or closed (via JavaScript, or because the user triggered another control element also tied to the same collapsible element). If the control element's HTML element is not a button (e.g., an `<a>` or `<div>`), the attribute `role="button"` should be added to the element.
```
mdo: Do we need to update the language here to be changed from `collapsible` to `offcanvas`? Or, to avoid duplication, should we point folks to this section in the collapse docs?

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+If your control element is targeting a single collapsible element – i.e. the `data-target` attribute is pointing to an `id` selector – you should add the `aria-controls` attribute to the control element, containing the `id` of the collapsible element. Modern screen readers and similar assistive technologies make use of this attribute to provide users with additional shortcuts to navigate directly to the collapsible element itself.
+
+## Usage
+
+The offcanvas plugin utilizes a few classes to handle the heavy lifting:
```
mdo:
```suggestion
The offcanvas plugin utilizes a few classes and attributes to handle the heavy lifting:
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+- `.offcanvas` hides the content
+- `.offcanvas.show` shows the content
+- `.offcanvas-right` hides the offcanvas on the right
+- `.offcanvas-bottom` hides the offcanvas on the bottom
+- `data-body="scroll"` enables body scrollbar when offcanvas is open
```
mdo:
```suggestion
- `data-body="scroll"` enables `<body>` scrolling when offcanvas is open
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+- `data-body="backdrop"` hides scrollBar and add backdrop to body, when offcanvas is open
```
mdo:
```suggestion
- `data-body="backdrop"` disables `<body>` scrolling and adds backdrop when offcanvas is open
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+- Add a dismiss button with the `data-dismiss="offcanvas"` attribute, which triggers the JavaScript functionality. Be sure to use the `<button>` element with it for proper behavior across all devices.
```
mdo: Separate this out from the list of classes/attributes:

```suggestion

Add a dismiss button with the `data-dismiss="offcanvas"` attribute, which triggers the JavaScript functionality. Be sure to use the `<button>` element with it for proper behavior across all devices.
```

site/content/docs/4.3/components/offcanvas.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Via data attributes
+
+Just add `data-toggle="offcanvas"` and a `data-target` or `href` to the element to automatically assign control of one offcanvas element. The `data-target` attribute accepts a CSS selector to apply the offcanvas to. Be sure to add the class `offcanvas` to the offcanvas element. If you'd like it to default open, add the additional class `show`.
```
mdo:
```suggestion
Add `data-toggle="offcanvas"` and a `data-target` or `href` to the element to automatically assign control of one offcanvas element. The `data-target` attribute accepts a CSS selector to apply the offcanvas to. Be sure to add the class `offcanvas` to the offcanvas element. If you'd like it to default open, add the additional class `show`.
```

site/content/docs/4.3/examples/offcanvas/index.html
```diff
@@ -2,20 +2,20 @@
 layout: examples
 title: Offcanvas template
 extra_css:
-  - "offcanvas.css"
-extra_js:
-  - src: "offcanvas.js"
-body_class: "bg-light"
+- "offcanvas.css"
 ---
```
mdo: This entire file needs to be rethought. There's no reason to remove the custom page content from this. I imagine this should instead be updated to leave the page content as-is, and change the behavior of the navigation sidebar.

mdo (changes requested):

site/content/docs/4.3/components/offcanvas.md
GeoSot: I bet it is left by mistake

site/content/docs/4.3/components/offcanvas.md
GeoSot: I chose the 'link' way, to avoid extra text.

site/content/docs/4.3/components/offcanvas.md
GeoSot: @mdo 
I think now it is ok. Please take a look

site/content/docs/4.3/examples/offcanvas/index.html
GeoSot: @mdo 

Sorry, I'm not sure I understood what you propose

site/assets/scss/docs.scss
```diff
@@ -51,3 +51,9 @@
 @import "syntax";
 @import "anchor";
 @import "algolia";
+
+// Temp
+
+.offcanvas{
+  z-index: 1072 // over .bd-subnavbar
```
XhmikosR: We should probably move this in the docs variables and reuse it.

GeoSot: Or as override on `site/assets/scss/docs.scss` before  the general variables import, but css-linter is complaining.

XhmikosR: @MartijnCuppens can you please have a look when you have some time?

I also added another TODO regarding the `h4`s

MartijnCuppens: Why is this added? The z-index of the off canvas component is higher than `.bd-subnavbar` anyway?

GeoSot: it was added because  in a previous version, .bd-subnavbar had 1071 z-index and was covering offcanvas. I can make a check to find if it is changed. 

(offcanvas is on hold about 10 months, so many things have changed and sure  I have missed some of them)

GeoSot: Finally, it was redundant

XhmikosR:
@GeoSot please make sure you do **not** overwrite my rebase.

@MartijnCuppens and I will have a look at the remaining things and let you know after that.

MartijnCuppens:
Pushed some changes. What bothers me, is the inconsistency between this offcanvas component and the modals (for example the way the backdrop is set). We got to make sure to keep this consistent (maybe we should make a backdrop component?).

GeoSot:
@MartijnCuppens during the last commit you removed the choice to change size (width or height) below a breakpoint (ex:mobile devices). Are you sure you want to remove it?

MartijnCuppens:
> @MartijnCuppens during the last commit you removed the choice to change size (width or height) below a breakpoint (ex:mobile devices). Are you sure you want to remove it?

Yes, we try to avoid `mobile` variables. Smaller viewports != mobile. Widths can be changed with the Sass variables.

XhmikosR:
Rebased. We should look into wrapping this up for v5.0.0-alpha1.

719media:
One thing that's always annoyed me with body overflow: hidden is with scroll bar width (e.g. Edge on windows, or if you have Show Scroll bars: always on your mac OS settings).

Screen Shot 2020-06-19 at 1 33 57 PM


You get a jutter when applying overflow: hidden, just as the linked example for this PR. Anyway, I get around this by doing

```
const scrollBarGap = window.innerWidth - document.documentElement.clientWidth;
if (scrollBarGap > 0) {
  document.documentElement.style.setProperty('--scroll-bar-width', `${scrollBarGap}px`);
} else {
  document.documentElement.style.removeProperty('--scroll-bar-width');
}
```
In my javascript before applying the class that applies body overflow: hidden. Then, in that same class, I use the css
```
padding-right: var(--scroll-bar-width, 0);
```

Anyway, just a thought in case you find it useful. I'm not sure how the bootstrap project considers references to `window`, so there may be a different opinion there.

dotnetshadow:
Would this component also handle this style of offcanvas, where the sidebar would push the body content as well? https://www.w3schools.com/howto/howto_js_off-canvas.asp

GeoSot:
> 
> 
> Would this component also handle this style of offcanvas, where the sidebar would push the body content as well? https://www.w3schools.com/howto/howto_js_off-canvas.asp

I am afraid not with this implementation. Maybe on a future update

GeoSot:
@719media 

Please make a reminder again, when you see this 'pr' is going to be merged (or after it). It is too early to add more things and as you see it is already difficult to proceed on this process

vosidiy:
Actually it is easy to convert modal component into offcanvas. 
Here is I made https://bootstrap-menu.com/demos/modal.html 
Why not add special class like .modal-aside-right  and  .modal-aside-left  to make offcanvas from modal itself .

GeoSot:
> Actually it is easy to convert modal component into offcanvas.
> Here is I made https://bootstrap-menu.com/demos/modal.html
> Why not add special class like .modal-aside-right and .modal-aside-left to make offcanvas from modal itself .

I  had in mind a class that extends modal, but you went much further.
Yes personally i like it. 
It's a bit confusing as concept,  but it is simple with some lines of pure css gaining smaller size on final result and already tested js code

@mdo 
@XhmikosR 
@MartijnCuppens 


Please take a look on this.
It could be a good idea

ux-engineer:
Please consider supporting these kind of features: https://github.com/bootstrap-vue/bootstrap-vue/issues/5095

Just finished a custom implementation of a similar sidebar main navigation (also using collapsable submenus). Implemented it first without, and now with, BSV Sidebar component. Turned out to be much more work than anticipated. And still need to optimize for narrow mobile widths...

Should be pretty common layout for a modern web app that is optimized for mobile use.

mdo:
Capturing some of these latest comments in a new PR: see #31285. I know there have been previous efforts on this as well, and I appreciate @vosidiy for sharing their own here again. Please head over there to share any feedback if y'all have any!

vahidalvandi:
please add rtl support

happynewlife:
Thank you very much for adding rtl support.
I compared:
https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/cheatsheet/
with:
https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/cheatsheet-rtl/

I have to say it's more than very good, It's one of the best. 

Here are some thoughts:
1. Typography: every thing look fine. Thank you for that.
2. Images: Totally fine. nice.
3. Tables: Lovely.
4. Forms: Everything is nice.
5. Input group:
FYI: Email, URL and numbers in RTL languguages are LTR. Must of the time username is English (LTR).
image

6. Validation: great. very great. very very very great. Thanks. It's save a lots of work. Beautiful.
7. Alerts: Lovely.
8. Breadcrumb: seems nice to me.
9. Buttons: good
10. Dropdowns: By default it most starts from right to left:
image
and also:
image
and:
image

11. list groups: nice.
12. Modal: Lovely.
13. Navs: Good.
14. Navbar: Very good.
15. Pagination: Nice work.
16. Popovers: Default is better to be on the left.
image

17. Progress: Very nice, also ltr progress is also usable in rtl languages.
18. Scrollspy: Very good. 
19. Spinners: Thanks but why? (Clockwise in rtl countries is same as ltr countries). Really thanks for that honest work.
20. Toasts: I love you for that.
21. Tooltips: Beautiful.

Thanks again.
By the way my first language is Persian and it's RTL.

ffoodd:
@happynewlife there's a dedicated PR for RTL (#30980), this is off-topic on this one. Thanks for your feedback though, this is well aligned with what is being done.

m-peterman:
Another thing to consider is the type of effect on off canvas. e.g. Switching between overlay and push(you could even consider putting in push where is shrinks the main content or 1 where is pushes this content off the other side off the canvas if off canvas expanded.)
see below for an example of what i mean.
https://tympanus.net/codrops/2013/04/17/slide-and-push-menus/

ux-engineer:
@m-peterman also I'd suggest to consider that kind of use case, that the 'offcanvas' sidebar would not be entirely hidden but just minimized. For example showing only icons while in expanded state having also link texts. That kind of layout seems pretty common need when implementing navigations.

HansUXdev:
Hey, if anyone wants help with the documentation on this tag me.
I'm maintaining a bootstrap 5 extension that already uses the off-canvas example template so I want to add this to it the second it gets merged into the main branches.

mdo:
Thinking we'll still be going with side modals instead of a brand new component. Please see #31285 and weigh in with any feedback or contributions via PR if y'all have ideas!

HansUXdev:
@mdo yeah, I saw that and I'm keeping an eye on it.
Should this issue be closed then?

GeoSot:
@mdo may I propose something different? 

I agree that till now both components have the same functionality. So yes it' a good idea the side modal. 
The disadvantages of the 'side' implementation, are that both components are chained by the same css and same functionality, and if you choose right now to keep them as one, they may be tied for ever.

I think it could be better to make an new component(js class), named as offcancvas, which for now will **extend** the **modal** js implementation and tests, so there is not need for extra writing, but with its **own css** classes.

Adapting this idea, will give the choice for a future decoupling or extra features, for both components

xcartmods:
https://github.com/xcartmods/bootstrap-4-side-modals

Boberro:
Hi, @GeoSot I found an issue with current version of a script. If a button (or link) used as a toggle to show an OffCanvas component has non-text elements, eg. `<svg>` or `` tags, button only works if you click on the padding area of the button/link.

https://codepen.io/demicanadian/pen/LYZErQP

GeoSot:
@Boberro I confirm the case. 
For now I do not want spend more time  to make a fix. 
I do not have any new feedback to proceed on this PR, and as I see side-modal is in favor to be the new offcanvas

patrickhlauke:
checked https://deploy-preview-29017--twbs-bootstrap.netlify.app/docs/5.0/components/offcanvas/ and it has major focus / focus order issues.

navigating through the page just with `tab`/`shift`+`tab`, the offcanvas can already be navigated into even when not visible (resulting in focus getting lost into the currently invisible offcanvas).

when the offcanvas is triggered, focus is not automatically moved to the offcanvas. in fact, it seems to reset itself to the start of the document (so on next `tab` you're all of a sudden on the "skip to content" link right at the top of the page).

once the user IS in the offcanvas, they can still tab back out of it.

at a high level, offcanvas should behave like a modal. this includes the way it handles focus.

having `aria-expanded="true"` on the offcanvas container makes no sense.

`aria-offcanvas="true"` makes absolutely no sense, that attribute does not exist.

js/src/offcanvas.js
```diff
@@ -0,0 +1,222 @@
...
+      setScrollbar(this._scrollbarWidth)
+      document.body.classList.add(CLASS_NAME_STOP_OVERFLOW)
+    }
+
+    this._element.removeAttribute('aria-hidden')
+    this._element.classList.add(CLASS_NAME_SHOW)
+    this.hidden = false
+    setTimeout(() => {
+      this._element.setAttribute('aria-expanded', true)
+      this._element.setAttribute('aria-offcanvas', true)
```
patrickhlauke, line 102: this attribute does not exist...

patrickhlauke (changes requested):
see previous comment for various other quite fundamental issues, beyond this code one

js/src/offcanvas.js
GeoSot: Thank you @patrickhlauke for your feedback 

Please when you have time, take a second look. I think I have fixed indicated problems

mariusa:
Will this make it to 5.0, please?

patrickhlauke:
Testing https://deploy-preview-29017--twbs-bootstrap.netlify.app/docs/5.0/components/offcanvas/

* once the offcanvas is closed, focus / tabbing is broken in the page; can't reach anything when using `Tab`/`Shift`+`Tab`
* offcanvas falls below the search bar in the docs image
* the offcanvas is, in essence, a modal dialog that just happens to show on the side rather than centered. it should have all the same things as modal dialog, e.g. `role="dialog"`, `aria-modal="true"` etc
* once the offcanvas is closed, focus should be programmatically set back to the control that originally opened the offcanvas

GeoSot:
Thank you again @patrickhlauke 

The issue with search bar z-index it happens on modal too cause the search bar on the docs css has the biggest z-index of all.
So I didn't change something, in case you want to fix it on another scope
Capture

newyhj:
hope 5.0 release publish this features

patrickhlauke:
@GeoSot thanks for making the changes requested - can confirm that this now behaves like a modal. One last suggestion: I'd make the title of the offcanvas an actual heading (in theory it should be `<h1>`, but probably something like an `<h2>` or `<h3>` would be acceptable too), give it an `id`, and reference it from the modal container using `aria-labelledby` (same as with the regular modals) - this gives the modal a label that is then announced by assistive technologies when the modal first pops up and focus is moved to it.

patrickhlauke:
regarding the subnavbar displaying over the offcanvas, i filed https://github.com/twbs/bootstrap/pull/32839

patrickhlauke:
oh, additionally, would be good to swap out the lorem ipsum text with some actual/pseudo text (see https://github.com/twbs/bootstrap/pull/32246 and a few related recent PRs)

patrickhlauke:
when the off-canvas is hidden, but has keyboard-focusable elements, it is still in the focus order (`aria-hidden="true"` hides its content from AT, but does not prevent focusable controls from being focused). when not visible, the off-canvas needs to be fully neutralised (e.g. with `display:none`), the same way modal dialogs are.

color schemes and options examples should also have an explicit "X" close button. otherwise, a touch+AT user will not be able to close the off-canvas (nothing to swipe to and then click/activate to close it, and no ability to press `Esc`).

accessibility section points to collapse component accessibility docs, but really this should say that functionally, an off-canvas is like a modal dialog, that just happens to be positioned to the side. and then i'd say copy (rather than reference) and adapt the accessibility guidance for modal https://getbootstrap.com/docs/5.0/components/modal/#accessibility

a more high-level question (@mdo et al): do we call it "offcanvas" (one word, ugly) or "off-canvas", or even "off-canvas panel"? i'd tend towards the latter.

some of the wording in the docs part may also need a bit of tweaking, but i can get around to that once we're closer to finished.

ux-engineer:
What about 'Sidebar Navigation/Panel' or 'Side Panel'?

GeoSot:
@patrickhlauke you got me in thoughts. Should support swipe to close too?

patrickhlauke:
> @patrickhlauke you got me in thoughts. Should support swipe to close too?

i'd stay away from that to be honest, as it's...non-trivial (and swipe support would still not work for touch+AT users, as swipe gestures are handled directly by AT to control it first and foremost, unless a user explicitly does a pass-through gesture)

newyhj:
@GeoSot  wait for you,everyone needs this feature.

thor456:
Offcanvas is an essential feature, and required feature. 

In Dashboard and Admin themes, the only required thing is Offcanvas. 
The developers need to understand this and make this permanent feature of `Bootstrap 5`

mdo:
@geosot I owe you a huge apology for the delays here. Thank you for continuing to update and contribute elsewhere. After talking with @XhmikosR and re-reviewing your PR here, I'd love to make this happen in v5 if you're still up for it. No more delays on our part, we'd love to get this in the core.

Here's what @XhmikosR and I discussed over Slack:

- [ ] Finish and merge this PR
- [ ] Close any other PRs
- [ ] Iterate on the docs a bit more
- [ ] Detach and de-dupe the backdrop in #32439

Anything else we'd want to do?

GeoSot:
@mdo  I would like to super thank you. It is almost an honor for me to achieve this level of contribution. I 'll try my best

newyhj:
@mdo so,v5 beta2 can publish this feature?

thor456:
Bootstrap Offcanvas should be visible in large screen, only hide in smaller screen. 

Example: w3schools, tailwind. 
Visible in large screen, collapse with hamburger in small(540px) screen. 

For the intended behavior, take as example [codeply-offcanvas](https://www.codeply.com/p/FpImnZeaPt), 

this is how I am looking for the **Offcanvas**. I only need official support, and official touch - so that this does not get broken in say other browsers, mobile screens, other devices from other geographies. 

Uses: 
1. Tutorial Websites like w3schools
2. Dashboard theme(most used purpose)

patrickhlauke:
> a more high-level question (@mdo et al): do we call it "offcanvas" (one word, ugly) or "off-canvas", or even "off-canvas panel"? i'd tend towards the latter.

Still angling for at least the first hyphenated naming - @mdo thoughts?

vanillajonathan:
The classes are currently named `.offcanvas`, `.offcanvas-right` and `.offcanvas-bottom`. Considering the recent move to RTL, should they be named that, or should they be named `.offcanvas-start` and `.offcanvas-end` instead?

site/content/docs/5.0/examples/offcanvas/offcanvas.css
```diff
@@ -1,67 +1,3 @@
-html,
-body {
-  overflow-x: hidden; /* Prevent scroll on narrow devices */
-}
-
 body {
```
XhmikosR, line 1: Would be nice if we could replace this with a core util and get rid of the CSS file here. Maybe `pt-5` is enough @mdo?

patrickhlauke:
Thank you for tackling the things I mentioned in https://github.com/twbs/bootstrap/pull/29017#issuecomment-766848961

Last few comments:

* remove the `aria-expanded="false"` from the various trigger elements. it's incorrect to use these (it's not an open/close disclosure widget here)
* wondering if for the code samples it would make sense to shorten the actual shown markup / only show `...` for the actual content of the panel? the samples appear super long and complex, but most the complexity is just all the example markup stuff in the off-canvas itself. i'd shorten them samples to basically show what's needed for the trigger and what the most important constituent parts of the panel are, e.g.

```
<div class="offcanvas offcanvas-left" tabindex="-1" id="offcanvasExample" aria-labelledby="exampleOffCanvasLiveLabel">
  <div class="offcanvas-header">
    <h5 class="modal-title" id="exampleOffCanvasLiveLabel">offcanvas</h5>
    <button type="button" class="btn-close text-reset" data-bs-dismiss="offcanvas" aria-label="Close">
    </button>
  </div>
  <div class="offcanvas-body">
    ...
  </div>
</div>
```

* I would still love to see (at least in the documentation) a better name for this. "Offcanvas" as one word is awkward. still angling for "Off-canvas" (which seems quite popular as a term, from some googling). the class names can remain `.offcanvas`, I don't mind that, but in documentation, I'd strongly favour the hyphenated version. 
* There's some tweaks to the wording in the documentation that I'd do, but I'll go for that in a separate PR once this is merged.

site/content/docs/5.0/components/offcanvas.md
```diff
@@ -0,0 +1,259 @@
...
+layout: docs
+title: Offcanvas
+description: Build hidden sidebars into your project for navigation, shopping carts, and more with a few classes and our JavaScript plugin.
+group: components
+toc: true
+---
+
+## How it works
+
+The offcanvas JavaScript plugin shows and hides sidebar on the left, right, or bottom of your viewport. Buttons or anchors are used as triggers that are attached to specific elements you toggle.
```
patrickhlauke, line 11: "shows and hides sidebar on the left..." > "shows and hides an off-canvas panel on the left..." (or something similar)

site/content/docs/5.0/components/offcanvas.md
```diff
@@ -0,0 +1,259 @@
+---
+layout: docs
+title: Offcanvas
```
patrickhlauke, line 3: Offcanvas > Off-canvas

(and throughout the docs page)

site/content/docs/5.0/components/offcanvas.md
```diff
@@ -0,0 +1,259 @@
...
+
+Click the buttons below to show and hide an offcanvas element via class changes:
+
+- `.offcanvas` hides content
+- `.offcanvas.show` shows content
+
+You can use a link with the `href` attribute, or a button with the `data-bs-target` attribute. In both cases, the `data-bs-toggle="offcanvas"` is required.
+
+{{< example >}}
+<a class="btn btn-primary" data-bs-toggle="offcanvas" href="#offcanvasExample" role="button" aria-expanded="false" aria-controls="offcanvasExample">
```
patrickhlauke, line 25: remove the `aria-expanded="false"` here and in all other trigger elements. it's wrong.

site/content/docs/5.0/components/offcanvas.md
```diff
@@ -0,0 +1,259 @@
...
+## Example
+
+Click the buttons below to show and hide an offcanvas element via class changes:
+
+- `.offcanvas` hides content
+- `.offcanvas.show` shows content
+
+You can use a link with the `href` attribute, or a button with the `data-bs-target` attribute. In both cases, the `data-bs-toggle="offcanvas"` is required.
+
+{{< example >}}
```
patrickhlauke, line 24: for examples, which are extremely long and complex-looking mainly due to the amount of example content in the off-canvas, consider perhaps not using `{{< example >}}` and instead doing a split of actual code vs shown code, eliding the content of the `offcanvas-body` as it's not really that interesting from a sample code perspective.

see how https://github.com/twbs/bootstrap/blob/main/site/content/docs/5.0/components/modal.md for instance does it for vertically centered modals (https://getbootstrap.com/docs/5.0/components/modal/#vertically-centered)

site/content/docs/5.0/components/offcanvas.md
```diff
@@ -0,0 +1,259 @@
...
+  </div>
+  <div class="offcanvas-body">
+    <p>Try scrolling the rest of the page to see this option in action.</p>
+  </div>
+</div>
+{{< /example >}}
+
+## Accessibility
+
+Be sure to add aria-labelledby="...", referencing the modal title, to .offcanvas. Note that you don’t need to add role="dialog" since we already add it via JavaScript.
```
patrickhlauke, line 172: I'd start this with a short mention a la "The off-canvas panel is, conceptually, a modal dialog."

patrickhlauke (changes requested):

vanillajonathan:
The classes are currently named `.offcanvas`, `.offcanvas-right` and `.offcanvas-bottom`. Considering the recent move to RTL, should they be named that, or should they be named `.offcanvas-start` and `.offcanvas-end` instead?

site/content/docs/5.0/examples/offcanvas/index.html
```diff
@@ -31,27 +28,27 @@
         </li>
         <li class="nav-item dropdown">
           <a class="nav-link dropdown-toggle" href="#" id="dropdown01" data-bs-toggle="dropdown" aria-expanded="false">Settings</a>
-          <ul class="dropdown-menu" aria-labelledby="dropdown01">
-            <li><a class="dropdown-item" href="#">Action</a></li>
-            <li><a class="dropdown-item" href="#">Another action</a></li>
-            <li><a class="dropdown-item" href="#">Something else here</a></li>
-          </ul>
+          <div class="dropdown-menu" aria-labelledby="dropdown01">
```
XhmikosR, line 31: Should we revert this change?

GeoSot:
Hey guys. I will need little help or your opinion.
Js tests for scrollbar fail in combination with modal's tests. I am very confident about the implementation and almost sure that it fails cause of some nailed bodystyle values on modal test. Both codes do their job (I think correct) but with different ways. Scrollbar.js uses only js and client widths, modal uses dummy html element. 
The difference on tests is that scrollbar.js tries to find out only if there is a scrollbar and its width only, as modal tests do some tricks with body right padding and though time are over-engineered on what this code can do.

The right way for me, is to fix modal tests but that's me. Is there another advice? Can anyone help to align them with reality? 

cc/
@mdo, @XhmikosR , @rohit2sharma95

mdo:
I think I'd have to defer to @rohit2sharma95 and @XhmikosR on the best path forward there @geosot. The less clever a solution is though, the better it generally is. If we can streamline things and make it more accurate, that'd be the solution we should aim for IMO.

mdo:
Pushed another update to edit some wording in the docs. For now, I'm leaving everything as `offcanvas`, from source code to usage in the docs copy. We can potentially change this later, but I prefer the one way to spell it, even though it's grammatically incorrect. 

I've also renamed the `-left` and `-right` classes to `-start` and `-end`, to match RTL naming.

From my POV, this is basically good to merge. Then I believe we can tackle the backdrop issue.

js/src/offcanvas.js
```diff
@@ -0,0 +1,239 @@
...
+ * Constants
+ * ------------------------------------------------------------------------
+ */
+
+const NAME = 'offcanvas'
+const DATA_KEY = 'bs.offcanvas'
+const EVENT_KEY = `.${DATA_KEY}`
+const DATA_API_KEY = '.data-api'
+const ESCAPE_KEY = 'Escape'
+const DATA_BODY_ACTIONS = 'data-bs-body'
```
rohit2sharma95, line 32: Better if you can accept this option in the configuration object (Like other plugins) :thinking:

js/src/offcanvas.js
```diff
@@ -0,0 +1,239 @@
...
+ * ------------------------------------------------------------------------
+ */
+
+class OffCanvas extends BaseComponent {
+  constructor(element) {
+    super(element)
+
+    this._isShown = element.classList.contains(CLASS_NAME_SHOW)
+    this._bodyOptions = element.getAttribute(DATA_BODY_ACTIONS) || ''
+    this._handleClosing()
```
rohit2sharma95, line 63: Better if you name this like `_addEventListeners` :slightly_smiling_face: 
It keeps the consistent naming convention in the codebase

js/src/util/scrollbar.js
```diff
@@ -0,0 +1,70 @@
...
+ * --------------------------------------------------------------------------
+ */
+
+import SelectorEngine from '../dom/selector-engine'
+import Manipulator from '../dom/manipulator'
+
+const SELECTOR_FIXED_CONTENT = '.fixed-top, .fixed-bottom, .is-fixed'
+const SELECTOR_STICKY_CONTENT = '.sticky-top'
+
+const getWidth = () => {
```
rohit2sharma95, line 14: It would be good if you change its name so that it can be understood whose width you are getting :slightly_smiling_face:

js/tests/unit/util/scrollbar.spec.js
```diff
@@ -0,0 +1,182 @@
...
+      htmlOffset: document.documentElement.offsetWidth,
+      docClient: document.body.clientWidth,
+      htmlBound: document.documentElement.getBoundingClientRect().width,
+      bodyBound: document.body.getBoundingClientRect().width,
+      window: window.innerWidth,
+      width: Math.abs(window.innerWidth - document.documentElement.clientWidth)
+    }
+  }
+
+  const scrollBarIsHidden = () => { // IOS devices, Android devices and Browsers on Mac, hide scrollbar by default and appear it, only while scrolling. So the tests for scrollbar would fail
```
rohit2sharma95, line 18: It's better to prefix it with 'is' since it is like a boolean flag

js/src/offcanvas.js
```diff
@@ -0,0 +1,239 @@
...
+        element !== event.target &&
+        !element.contains(event.target)) {
+        element.focus()
+      }
+    })
+    element.focus()
+  }
+
+  _bodyOptionsHas(option) {
+    return this._bodyOptions.split('|').includes(option)
```
rohit2sharma95, line 160: Few plugins are already using comma separated values, so if you also use a comma here instead of a pipe, that would keep consistency IMO.

GeoSot: I agree and I 've changed it. IMO I would prefer 'pipe'. Seems more programmatic in my eyes

js/src/util/scrollbar.js
GeoSot: I am open to any proposal. I have started naming `getScrollbarWidth` etc but is seem too much. We are inside **scrollbar.js** and we need the width of scrollbar.

js/src/offcanvas.js
GeoSot: @Rohit. I really would prefer it to keep it simple right now, in order to proceed and to avoid more mess. 
If you find it so important, I can change it after. I am not very confident with the `Config` object, as every component uses it, in a different way

XhmikosR:
All right, this is pretty hard to rebase to clean up the patches into less commits, so I'll have to squash it into one patch.

Huge thanks to @GeoSot for sticking with us after all this time, to everyone else who contributed to the branch and to those who helped reviewing the PR!

We'll need to make some refactoring later to deduplicate some stuff, tweak doc examples, add a navbar offcanvas example, etc, but at least this should help a lot of people.

mdo (approved):

mdo:
Took us long enough 😆 ❤️

rohit2sharma95:
Thank you George for the hard work :heart::blue_heart:  and for debugging BrowserStack tests :smile:

GeoSot:
Thank you all for your support on this MR. Thanks @XhmikosR  for your patience

PR #29020, opened 2019-07-13
Drop breakpoint CSS custom properties

### LINKED ISSUE #29012: Remove breakpoint custom properties
Mentioned by me in #28936: we cannot use CSS custom properties in media queries, so we should remove the breakpoint variables we have there currently.

/cc @twbs/css-review for feedback, as well as ideas on perhaps what to add if anything

mdo:
Media queries do not support CSS custom properties (CSS variables), so we've had some useless variables lying around since shipping v4. This finally removes them and the associated documentation content that explained the situation.

Fixes #29012.


XhmikosR:
Shouldn't we backport this since it's non-functional?

mdo:
> Shouldn't we backport this since it's non-functional? It should reduce the CSS size, so it's a win-win.

No, because it'd be a breaking change if someone were using the variables elsewhere.

MartijnCuppens (approved):

PR #29021, opened 2019-07-13
Document Firefox bug for custom file focus styles

### LINKED ISSUE #26563: Custom file input has no focus indication in Firefox

mdo:
Closes #26563.


alecpl:
A possible solution was proposed at https://github.com/twbs/bootstrap/pull/26576#issuecomment-506943872, could you consider this?

PR #29022, opened 2019-07-13
v5: Fix disabled styling on .form-file

### LINKED ISSUE #28176: Use disabled styling for the button part of the custom-file input
Disabled styling was added to the custom file input recently, but that didn't cover the button part of the component, so the button looks enabled when the input is disabled.

I think this could be achieved by adding something like:

  &::after {
    @include box-shadow(none);
  }

to the   
&:disabled {
    ~ .custom-control-label {

section of the _custom_forms.scss

mdo:
While #28450 overhauled a ton of forms work, the  support for this custom file input was ineffective. This updates the selector to target the custom field and gray out the entire field.

Fixes #28176.


ponomarenko16 (approved):

MartijnCuppens (approved):
LGTM

PR #29023, opened 2019-07-13
v5: Reuse box-shadow variables

### LINKED ISSUE #23775: Create component shadow variables to keep consistency across components

mdo:
- Removed default `<kbd>` shadow
- Add new `$box-shadow-inset` variable to better match multiple components' inner shadows (when enabled)
- Reuse `$box-shadow` variables across several components for consistency

Fixes #23775.


MartijnCuppens (approved):

scss/_code.scss
```diff
@@ -17,13 +17,11 @@ kbd {
   color: $kbd-color;
   background-color: $kbd-bg;
   @include border-radius($border-radius-sm);
-  @include box-shadow($kbd-box-shadow);
 
-  kbd {
+  & {
```
XhmikosR: maybe we should revert this to `kbd` so that stylelint passes or ignore the warning?

PR #29029, opened 2019-07-13
Doc tweaks

XhmikosR:
@mdo: not sure why you used 600px, but this is the proper way to do it. If you want it bigger, we'd need a bigger set of the 2 images.


XhmikosR:
IE should be fixed.

/CC @mdo for approval

XhmikosR:
@patrickhlauke: should I add ` aria-hidden="true"` in the SVG?

XhmikosR:
Perhaps we should move the arrow in the CSS pseudo selector since it's repeated a lot.

patrickhlauke:
> @patrickhlauke: should I add ` aria-hidden="true"` in the SVG? And `focusable="false"`?

yes please

mdo:
We can skip those SVG changes if we instead make it part of the CSS. Lemme have a crack at that.

mdo:
Pushed up the changes for the sidebar, but it occurs to me that I'm also using the same icon elsewhere in the subnav, so we likely need those optimizations as well.

XhmikosR:
>Pushed up the changes for the sidebar, but it occurs to me that I'm also using the same icon elsewhere in the subnav, so we likely need those optimizations as well.

Which optimizations are you referring to?

site/layouts/partials/docs-sidebar.html
```diff
@@ -17,7 +17,7 @@
 
       <li class="bd-sidenav-group my-1{{ if $active_group }} active{{ end }} js-sidenav-group">
         <a class="d-inline-flex align-items-center bd-sidenav-group-link" href="/docs/{{ $.Site.Params.docs_version }}/{{ $group_slug }}/{{ if $group.pages }}{{ $link_slug }}/{{ end }}">
-          <svg class="bd-sidenav-group-link-icon" viewbox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path stroke="currentColor" stroke-width="2" d="M5 14l6-6-6-6" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round"/></svg>
+          <svg class="bd-sidenav-group-link-icon" width="16" height="16" viewbox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path stroke="currentColor" stroke-width="2" d="M5 14l6-6-6-6" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round"/></svg>
           <div>{{- $group.title -}}</div>
```
XhmikosR: Should this be just a `span` @mdo?

site/layouts/partials/docs-sidebar.html
```diff
@@ -38,7 +37,9 @@
 
     <li class="my-3 mx-4 border-top"></li>
     <li class="bd-sidenav-group pl-3">
-      <a class="d-inline-flex bd-sidenav-group-link" href="/migration">Migration</a>
+      <a class="d-inline-flex align-items-center bd-sidenav-group-link" href="/migration">
+        <div>Migration</div>
```
XhmikosR: Shouldn't this be in nav.yml?

XhmikosR:
Alright, I left a couple more comments, but I'd like them addressed soon so that we don't end up with a huge PR.

XhmikosR:
I'm putting this on hold because I believe it can be optimized further.

1. We shouldn't use an `a` tag for the group item since it's basically a button
2. Maybe we could just use our collapse plugin here?

MartijnCuppens (approved):

site/layouts/partials/docs-sidebar.html
mdo: In this case, it doesn't matter because the parent element is a flex container and will align everything essentially the same way.

PR #29038, opened 2019-07-14
V4 backports

XhmikosR:


XhmikosR:
@Johann-S @patrickhlauke it seems browser tests fail for v4-dev here in this branch.

patrickhlauke:
sigh...ok, i'd say don't backport that change, as i see all other tests there use the `trigger` method. wondering if in this particular case the `trigger` does bubble properly (compared to when i tried to use that method of firing key events in unit tests for the work on v5 aria tabs and the tests were failing/key events weren't propagating properly).

the original change is still a valid change for the v5 branch, as there we don't test IE11, and there the unit test uses the `dispatchEvent` method throughout.

XhmikosR:
Alright, reverted the patch. BTW we do test IE11 on master.

patrickhlauke:
> Alright, reverted the patch. BTW we do test IE11 on master.

so how can the exact same test fail in QUnit here but not there? (and it failed not in a "the test doesn't pass" way, but rather an "object doesn't support this" more fundamental problem with actual QUnit/the command used way)

XhmikosR:
That I don't know. I will try again tomorrow just in case it was a one-time thing.

XhmikosR:
OK, the test failures persist.

js/tests/unit/dropdown.js
```diff
@@ -84,10 +84,11 @@ $(function () {
     $(dropdownHTML).appendTo('#qunit-fixture')
     var $dropdown = $('#qunit-fixture').find('[data-toggle="dropdown"]').bootstrapDropdown()
     var $button = $('button[data-toggle="dropdown"]')
+    $button[0].focus()
     // Key escape
-    $button.trigger('focus').trigger($.Event('keydown', {
-      which: 27
-    }))
+    var keydown = new Event('keydown')
```
Johann-S: using `new Event` isn't supported on IE 10 and IE 11, you should use jQuery event system

Johann-S (changes requested):

js/tests/unit/dropdown.js
XhmikosR: I've already dropped the patch from this branch. If @patrickhlauke wants, we should have a separate PR.

Johann-S (approved):

PR #29044, opened 2019-07-15
Revert incorrect change

MartijnCuppens:
The `kdb` change from https://github.com/twbs/bootstrap/pull/29023 didn't nest the selectors

### DIFF
--- a/scss/_code.scss
+++ b/scss/_code.scss
@@ -18,7 +18,6 @@ kbd {
   background-color: $kbd-bg;
   @include border-radius($border-radius-sm);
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
I guess the warning was right after all :/

XhmikosR (approved):

mdo:
Huh, my bad! Thanks for fixing. Must have been thinking about Less syntax :).

MartijnCuppens:
`& #{&}` would be the appropriate syntax, but that's definitely less clear than just redefining the selector.

PR #29053, opened 2019-07-16
Made the button selectors more explicit

phistuck:
[type="button"] also matches `<i type="button">` as well,
which is not intended and will soon be invalid.
This change makes the button selector only match actual button types.

Example -
http://dixi-bg.com/

More examples may be found on the list at the bottom of this page -
https://www.chromestatus.com/metrics/feature/timeline/popularity/2774

And what do you know, even GitHub suffers from that. See -

[type="button"] also matches <i type="button"> as well, which is not intended and will soon be invalid.
This change makes the button selector only match actual button types.

Example -
http://dixi-bg.com/

More examples may be found on the list at the bottom of this page -
https://www.chromestatus.com/metrics/feature/timeline/popularity/2774

image

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -378,9 +378,12 @@ select {
 //    controls in Android 4.
 // 2. Correct the inability to style clickable types in iOS and Safari.
 button,
-[type="button"], // 1
-[type="reset"],
-[type="submit"] {
+button[type="button"], // 1
+button[type="reset"],
+button[type="submit"],
+input[type="button"],
+input[type="reset"],
+input[type="submit"] {
   -webkit-appearance: button; // 2
 }
 


MartijnCuppens:
If I'm not mistaken `type` is not a valid attribute on `<i>`. Other than that this PR increases specificity (from `0,0,1,0` to `0,0,1,1`) making it impossible to change the `appearance` with just a class (`0,0,1,0`).

phistuck:
@MartijnCuppens - I agree with everything.  `<i>` was just an example, it applies to any element with those types. Applying `-webkit-appearance: button` to any element with `type=button` (and others) is also not ideal nor expected, especially since it has side effects in WebKit based browsers (and currently, for Blink based browsers, too).

This prompted me to detect this -
https://groups.google.com/a/chromium.org/d/msgid/blink-dev/CAGH7WqHN0xDH%2Bt6%3DTC2KW80kNoH0zqizcY%2BKyMkSodQ3jdYtOQ%40mail.gmail.com?utm_medium=email&utm_source=footer

mdo:
Given it is invalid and not used, nor recommended, by Bootstrap anywhere in our codebase, I don't see what we'd do this.

MartijnCuppens:
Me neither, closing this

phistuck:
What `<object type>`, by the way?
I have not tested the effect, but it should also match here. `type` is a standard attribute of `<object>` and Bootstrap is leaking `-webkit-appearance: button` into it.

PR #29059, opened 2019-07-16
Add support and docs for color input

mdo:
As shown in Safari:

Screen Shot 2019-07-16 at 3 22 15 PM

- [x] Add new `.form-control-color` modifier class to limit the width
- [x] Add basic code example
- [ ] Document styling details, browser support, etc

/cc @twbs/css-review 

Preview: <https://deploy-preview-29059--twbs-bootstrap.netlify.com/docs/4.3/forms/form-control/#color>

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -528,6 +528,12 @@ progress {
   -webkit-appearance: none;
 }
 
+// Remove padding around color pickers in webkit browsers
+
+::-webkit-color-swatch-wrapper {
+  padding: 0;
+}
+
 // 1. Change font properties to `inherit` in Safari.
 // 2. Correct the inability to style clickable types in iOS and Safari.
 
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -110,3 +110,8 @@
   line-height: $input-line-height-lg;
   @include border-radius($input-border-radius-lg);
 }
+
+.form-control-color {
+  max-width: 3rem;
+  padding: ($input-padding-y / 2) ($input-padding-x / 2);
+}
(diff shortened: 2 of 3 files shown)


XhmikosR (approved):

XhmikosR:
Perhaps we shouldn't document browser support? I mean, we'd need to cover macOS, Linux and Windows for each of the browsers.

MartijnCuppens:
I 've added rounded corners to the swatches and fixed the paddings in Firefox. I've also added the a fix which removes padding from webkit browsers in reboot.

XhmikosR:
Should we add cursor pointer?

XhmikosR:
Or a title or something. I feel like it's not clear one can interact with it.

MartijnCuppens:
Yeah, that might be a good idea. Shall we add it with the `$enable-pointer-cursor-for-buttons` variable? 
https://github.com/twbs/bootstrap/blob/dd383ff27b9058c12c2c6d4d6fb0b40413836f2d/scss/_buttons.scss#L14

XhmikosR:
>Should we add cursor pointer?
>Or a title or something. I feel like it's not clear one can interact with it.

@patrickhlauke: what do you think about this?

MartijnCuppens:
Btw, I don't think we need to document what the color picker will look like, browsers might change this anytime.

I would add something about browser support (IE) though.

XhmikosR:
I definitely agree with the above. See https://github.com/twbs/bootstrap/pull/29059#issuecomment-521980988

MartijnCuppens:
Ok, I've changed the copy to a notice about IE not supporting color inputs.

MartijnCuppens (approved):

PR #29075, opened 2019-07-18
Optimize the embedded SVGs.

XhmikosR:
On hold until we have the SVG escape function and do this properly.

I'd like to do the same for v4-dev later.


XhmikosR:
BTW that dropdown checkmark SVG is currently unused. I will fix it in one of the other PRs that tackle the docs navbar issues.

mdo (approved):

PR #29089, opened 2019-07-21
Update svgo.yml.

XhmikosR:


Johann-S (approved):

build/svgo.yml
```diff
@@ -15,6 +15,9 @@ js2svg:
   indent: 2
 
 plugins:
+#  - addAttributesToSVGElement:
+#      attributes:
+#        - focusable: false
```
MartijnCuppens: Shouldn't we remove these lines instead?

XhmikosR: I wanted them as documentation because in v4-dev we have it enabled. I'm still unsure if we need this and for which cases (inline, external SVG)

PR #29048, opened 2019-07-15
v5 Docs followup

mdo:
With #28450 (major forms rewrite) I changed the designed of the docs to get back to how we used to build this thing in v3 and earlier. v4's docs have always felt like a web app to me instead of an documentation site. Three columns of content, most of which stayed stickied was also a tad frustrating for me. More importantly though, I felt things were lacking polish, attention to detail, etc.

This PR will tackle the todos from @MartijnCuppens's comment and from @patrickhlauke's issue at #29047.

### Changes

Here's some clarification and additional fixes (including todos) to follow-up:

- [x] Clarify and update padding changes to `<ol>` and `<ul>` in Reboot. I added `padding-right` originally by mistake thinking I needed to reset that, but really I only needed `padding-left` redone.
- [x] Add new Icons component to the migration docs (I created these myself to replace the Open Iconic ones we'd been using). More icons are planned to come, but for now it focuses on a handful for our components.
- [x] Clarify migration docs for the changes in apperance and behavior for form components (namely checks and files).
- [x] Add note to migration docs about table changes.
- [x] Fix remaining tables that were lacking `.table`.
- [x] Improve icons component docs with code snippets, design decisions, and usage guidelines.
- [ ] ~~Look into `box-shadow`s support on icons with custom form controls~~
- [x] After #29029 merges, add search form to the docs homepage
- [x] Consider moving sidebar icon to `::before` per https://github.com/twbs/bootstrap/pull/29029#issuecomment-511722549 **Done in @XhmikosR's PR**
- [x] Fix color contrast in navbar
- [x] Update icons one more time—add some new ones, rename the class from `.booticon` to `.bi` because it's shorter and comes with fewer jokes, and remove unused styles in docs CSS.
- [x] Make navbar, subnav, and sidebar work on mobile
- [x] Move the icons to includes
- [ ] Move icons outside of content folder
- [ ] Stop inlining the SVGs; we have http/2 so it should be pretty fast anyway
- [ ] Verify if we need to add `role` or `focusable="false"`
- [ ] Move the nav changes to a separate PR

---

Explaining some other decisions...

### Tables

Tables saw a redesign to simplify their CSS and overall appearance. I also took the opportunity to remove several lines of custom docs CSS that re-applied the Bootstrap table styles to Markdown tables. Doing so meant rewriting the tables to allow explicit classes, which I was fine to do given we're inconsistent in our Markdown/HTML tables usage.

[...]

PR #29092, opened 2019-07-21
v5: Icons!

mdo:
Replaces the other half of #29048.

- Drops the Extend docs and moves the icons page to the Components section so we can have our own set
- Adds docs for customizing and using Bootstrap Icons
- Renames the classes to `.bi`
- Optimizes (thanks @XhmikosR!)
- Replaces the carousel icons with new compact chevrons

Still todo:

- [ ] ~~Tabnav between grid and table view of icons?~~
- [x] Move icons outside of content folder
- [x] Stop inlining the SVGs; we have http/2 so it should be pretty fast anyway?
- [ ] Verify if we need to add `role="img"` or `focusable="false"`
- [ ] ~~Add a tooltip in each icon showing its title and/or class?~~
- [ ] Click to copy SVG code?
- [ ] Fix inline customize SVGs display

/cc @XhmikosR @twbs/css-review 

Preview: <https://deploy-preview-29092--twbs-bootstrap.netlify.com/docs/4.3/components/icons/>


TryHardDood:
`Add a tooltip in each icon showing its title and/or class?`

And maybe click to copy the icon's code?

patrickhlauke:
yes, when the icons are actually shown, they need to somehow be presented to assistive technology/screen reader users. the `<svg>` itself, or its parent container, need a `role="img"` and an appropriate name, possibly via `aria-label="..."` on the element with that role.

and yes, the raw `<svg>` elements will need `focusable="false"

patrickhlauke:
if planning to add a tooltip, the `<svg>` or container will need to be focusable though (so tooltip can be triggered via keyboard).

site/data/icons.yml
```diff
@@ -1,25 +1,97 @@
...
-    website: https://glyph.smarticons.co/
-  - name: Icons8
-    website: https://icons8.com/
-  - name: icofont
-    website: https://icofont.com/
+- name: bootstrap
+- name: bootstrap-fill
+- name: reboot
+
+- name: x
```
XhmikosR: I'd use the non-unicode `x`

site/content/docs/4.3/components/icons.md
```diff
@@ -11,21 +11,16 @@ toc: true
 Meet the Bootstrap icons, a limited set of icons designed first and foremost to work with Bootstrap's components. From form controls to navigation, these icons are the perfect complement to our toolkit. They're SVGs, so they'll scale quickly and easily to any size, they can be easily recolored with CSS, and more. They're also included in Bootstrap under the MIT license.
 
 <div class="bd-bi-grid d-flex flex-wrap mt-4 mb-5 mx-n2">
-  {{< listIcons.inline >}}
-  {{- $dirName := printf "site/static/docs/%s/dist/icons/" .Site.Params.docs_version -}}
-
-  {{- range (readDir $dirName) -}}
```
XhmikosR: @mdo: is the only reason for this patch the fact that you want the Bootstrap icons shown first? This adds an extra step which we could avoid. I don't mind either way, just thinking out loud.

I could probably hook something up to change the order, but it might complicate things.

PR #29098, opened 2019-07-22
V4 dev backports and updates

XhmikosR:


XhmikosR:
@mdo: the responsive containers patch isn't merged yet. IMO we should make the next v4 release a patch version, but if we are going to merge this, be it a feature, we'll definitely need to make it a minor version bump.

XhmikosR:
I split the container-related patches so that we move forward with merging this, because it keeps growing.

MartijnCuppens (approved):

PR #29100, opened 2019-07-22
Switch to cssnano

XhmikosR:
Let's investigate switching to cssnano which seems to be better maintained.


XhmikosR:
After looking into this closer, it seems there are some optimizations we are not very keen of with @MartijnCuppens .

1. the pseudo selectors are changed to single colon ones. While it works and it'll probably continue to do so, I'd rather be explicit
2. the hsla color change
3. the svgo plugin seems to uppercase the escape characters which hurts gzip compression so I had to disable the plugin
4. some of the selectors grouping it does are OK but it changes the CSS compared to what we had. I disabled this optimization though
5. the background-position keywords to percentages conversion is good. The same goes for calc reduction which I guess we could achieve by hooking up postcss-calc ourselves

PR #29101, opened 2019-07-22
Replace http-server with zeit/serve.

XhmikosR:
The only "downside" is that serve doesn't accept a CLI flag to specify the directory to serve; it needs a config file. So, in this patch, I opted for just doing `cd _gh_pages` before `serve`.

EDIT: apparently it does accept a folder/path, it's just not documented.


XhmikosR:
OK, so it seems that serve requires Node.js 8.x so we can't backport it. The current v4-dev solution is broken anyway, so I think I will just remove it in that branch.

Johann-S (approved):

PR #29107, opened 2019-07-22
v4/v5: Spike for inputs with spinners

### LINKED ISSUE #28600: Spinner inside input
v4.2 features spinner. They can be embedded inside a button: https://getbootstrap.com/docs/4.3/components/spinners/#buttons

Would be nice to be able to also embed a spinner inside an input.

Sometimes you have fields that perform async actions, a spinner inside the input would give feedback to the user (something is pending).

Example: https://github.com/tkrotoff/bootstrap-spinner

bootstrap-spinner demo

mdo:
Screen Shot 2019-07-22 at 2 23 04 PM

Aims to fix #28600 

This is a spike for now, and will need to have the CSS moved, but I wanted to see how feasible this could be. Unfortunately the standard `.form-control` needs the small spinner, so we could support a large input but I'm unsure how to handle smaller inputs. An even smaller spinner?

Would love thoughts from @twbs/css-review.

Preview: <https://deploy-preview-29107--twbs-bootstrap.netlify.com/docs/4.3/components/spinners/#with-inputs>


ysds:
The official theme has a similar component.

image
https://bootstrap-themes.github.io/dashboard/docs/index.html#icon-input

It would be great to be able to support various icons, not just Spinners.

MartijnCuppens:
I agree with @ysds, would be cool if we could come up with a solution for more icons.

mdo:
The frustrating part about this for icons and spinners is eventually having to support icons/spinners on the left and right of the input. This will break form validation somehow, and requires additional styles and markup for the icons when they're not `background-image`s.

Here's a snippet of the CSS I have with an updated pass at this:

```scss
.form-control-with-icon-start,
.form-control-with-icon-end {
  position: relative;

  > .form-control-icon {
    position: absolute;
    top: 0;
    bottom: 0;
    display: flex;
    align-items: center;
  }

  > .spinner-grow,
  > .spinner-border {
    position: absolute;
    top: calc(50% - #{$spinner-width-sm} / 2) ;
  }
}

.form-control-with-icon-start {
  > .form-control {
    padding-left: 2rem;
  }

  > .form-control-icon,
  > .spinner-grow,
  > .spinner-border {
    left: .5rem;
  }
}

.form-control-with-icon-end {
  > .form-control {
    padding-right: 2rem;
  }

  > .form-control-icon,
  > .spinner-grow,
  > .spinner-border {
    left: .5rem;
  }
}
```

Starting to feel like a lot more than I wanted. Still looking for ways to simplify while allowing for some additional flexibility.

tmorehouse:
What about input groups with no borders between the append/prepend and the input? (or move the border to the input group and remove the borders from the input/text addon)?

It would give the visual look of the icon within the input, and would handle validation icons (although the focus ring would not be on the input-group - unless one used `:focus-within` which isn't supported everywhere yet)

mdo:
Closing for #31667.

PR #29115, opened 2019-07-23
fix arrow popover

Johann-S:


XhmikosR:
@Johann-S: are there any other instances we might have missed?

Johann-S:
@XhmikosR  no it's just popover which inherit from Tooltip, so now it's fine 👍

XhmikosR (approved):

PR #29095, opened 2019-07-22
v4/v5: Add responsive containers

mdo:
Pulling this from @browner12's PR to simplify everything and document the new classes.

PR #29118, opened 2019-07-23
Responsive containers (follow-up to #29095)

mdo:
This PR fixes the responsive containers that were added in #29095, originally stubbed out in #25631. Apologies to @browner12 for getting that wrong.

I still need to verify if at all we need to make changes to the navbar CSS. That looked more complicated than it needed to be, so need to verify the attribute selectors.

Fixes #25631.


browner12:
Thank you Mark for addressing this.

Apologies if my tone was harsh.

mdo:
No need to apologize, I was trying to bring it over and screwed it up—and it only took me 1.5 years to agree with you. 😬 Thanks for sticking with me.

mdo:
@browner12 I'm pulling in something like the table you made in your original PR, but with the `max-width`s instead. Can you confirm this matches?

Screen Shot 2019-07-23 at 3 59 10 PM

@MartijnCuppens Sorry for removing your changes before, they were the right move. Let me know once you've had a chance to review this, please.

browner12:
yes, that looks great.  makes it very clear to the user how this feature works.

mdo:
@browner12 Thank you! <3

acshef:
Looking at the table above... both `.container` and `.container-sm` have the same behavior. Is that expected?

browner12:
yah, below 576px it's so small that everything is just full width.

ocarreterom:
`.container-sm` is a 1140px wide responsive container?
Why not `.container-fluid-sm`? A `.container` behavior but fluid at _sm_ breakpoint.

With this change, `.container-sm}` can follow the `max-width` approach.

browner12:
`.container-sm` is 1140px wide _only_ on extra large screen sizes.

Containers are either **responsive** or **fluid**.  The infix determines where that switch happens.

I personally think the max width feature should be handled in userland, but if you'd like to attempt a separate PR, give it a shot.

mdo:
Yeah, and that container `max-width` is configurable in `$container-max-widths`. We just don't specify one for `xs` right now.

voltaek:
With `.container` and `.container-sm` operating the same, perhaps the example given in the Layout docs should use `.container-md` or `.container-lg` instead of `.container-sm` since their functionality is unique to them.

MartijnCuppens:
> @MartijnCuppens Sorry for removing your changes before, they were the right move. 

No problem man, nice to see you used the `@extend` technique here!

It's pretty weird `.container` and `.container-sm` do exactly the same. What do you guys think about if we set a max width to the `.container` of `360px` below the `sm` breakpoint?

browner12:
I don't thing we could do that on v4, since that would break BC?

MartijnCuppens:
Aha, didn't notice the `backport-to-v4` tag here. Never mind than, but it might me something to consider for `v5` in the future.

mdo:
Yeah @MartijnCuppens I want to make sure v4 gets this one because of how long I dragged my feet and proposed other ideas.

If you could, give a once over and thumbs up if these new changes match what you and @browner12 had landed on <3.

MartijnCuppens:
I was more in need of more container sizes like the initial PR, but this PR make sense too and it will probably be something more people want. For the different container sizes we can still make use of our utility API and combine `max-width` utilities with the container to center the content.

I do think we should use use a `xs` breakpoint, like `360px`, but since that introduces a BR like @browner12 mentioned, I'll leave that for `v5`.

So yeah, codewise this looks good to me.

I do think the documentation needs more clarification though, the description here (https://deploy-preview-29118--twbs-bootstrap.netlify.com/docs/4.3/layout/overview/#responsive) is clear, but the demo is confusing. This is probably the same thing @voltaek mentioned.

mdo:
Will revisit docs for this tonight!

MartijnCuppens:
We could also just drop the demo, the explanation and table are clear enough imo. The thing with the demo is that might confuse people if they start resizing their screen and notice nothing happens since we don't actually use the container classes.

Daemach:
>  but this PR make sense too and it will probably be something more people want.

Yes, I'm waiting anxiously for this.  I need standard containers on a desktop where there is a ton of (too much extra) room, and fluid containers on mobile where I need every bit of real estate I can get.  Thanks to @browner12 for adding it and for you guys staying on top of it.

PR #29123, opened 2019-07-24
set min-height for form-control textarea

m5o:
* set `min-height` for `.form-control` textarea
  * prevent user from shrink textarea to minimum height

| **before**  | **after**  |
|---|---|
| Screenshot 2019-07-24 at 16 40 50 | Screenshot 2019-07-24 at 16 41 13 |

### DIFF
--- a/scss/_forms.scss
+++ b/scss/_forms.scss
@@ -160,6 +160,7 @@ select.form-control {
 
 textarea.form-control {
   height: auto;
+  min-height: $input-height;
 }
 
 // Form groups


Ruffio:
If you don't want the user to resize the textarea, you can use:

`textarea {
  resize: none;
}`

XhmikosR:
Also generally we don't target patches against the v4-dev branch if they apply in master too.

m5o:
> Also generally we don't target patches against the v4-dev branch if they apply in master too.

Sorry, I thought `v4-dev` is the target, will change ✌️

PR #29124, opened 2019-07-24
set min-height for form-control textarea

m5o:
* set `min-height` for `.form-control` textarea
  * prevent user from shrink textarea 
  * set minimum height equals default input height

| **before**  | **after**  |
|---|---|
| Screenshot 2019-07-24 at 16 40 50 | Screenshot 2019-07-24 at 16 41 13 |
| Screenshot 2019-07-24 at 17 36 03  | Screenshot 2019-07-24 at 17 35 18  |

📷 Chrome 75 / Mac OS X 10.14.6

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -505,9 +505,9 @@ progress {
 // Fix height of inputs with a type of datetime-local, date, month, week, or time
 // See https://github.com/twbs/bootstrap/issues/18842
 
-::-webkit-inner-spin-button,
-::-webkit-outer-spin-button {
-  height: 0;
+::-webkit-datetime-edit {
+  overflow: visible;
+  line-height: 0;
 }
 
 
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -7,7 +7,7 @@
 .form-control {
   display: block;
   width: 100%;
-  height: $input-height;
+  min-height: $input-height;
   padding: $input-padding-y $input-padding-x;
   font-family: $input-font-family;
   @include font-size($input-font-size);
@@ -96,7 +96,7 @@
 // Repeated in `_input_group.scss` to avoid Sass extend issues.
 
 .form-control-sm {
-  height: $input-height-sm;
+  min-height: $input-height-sm;
   padding: $input-padding-y-sm $input-padding-x-sm;
   @include font-size($input-font-size-sm);
   line-height: $input-line-height-sm;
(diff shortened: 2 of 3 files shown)


ysds:
To fix also #28901, is it better add `min-height` to `.form-control` and remove the specific `height`s?

MartijnCuppens:
@ysds, that would reintroduce https://github.com/twbs/bootstrap/issues/18842, unless https://github.com/twbs/bootstrap/pull/28917 is merged.

ysds:
@m5o Would you add the `min-height` to `.form-control` and `form-select` directory and remove the specific `height` from those? Also need to modify the sizing options and the combinations with `.input-group`s.

m5o:
@ysds done, hopefully as you requested. 

🤔 I don't know if, for consistency, it would make sense to set `min-height` also to `.form-file`

ysds (approved):
@m5o nice, LGTM 👍

ysds:
`.form-file` use `position: absolute`, so would have to use `height`.

MartijnCuppens (changes requested):
Firefox add 2px to its select because of this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1560824, so we should set `height` on selects unfortunately

ysds:
umm, confirmed

image

scss/forms/_form-select.scss
```diff
@@ -42,7 +42,6 @@
 
   &[multiple],
   &[size]:not([size="1"]) {
-    height: auto;
```
MartijnCuppens: @m5o, we should also revert this line.

m5o: ok, will do – technically it was not my change 😉

mdo:
@MartijnCuppens Can you re-review, please?

XhmikosR:
Are we sure about backporting this BTW?

MartijnCuppens:
#18842 still seemed to be an issue. I've pushed a working fix (219e1c68ce6405778bcaf865841cc89f3d80b7c6) and also committed a fix for the input groups.

MartijnCuppens (approved):

XhmikosR:
@MartijnCuppens: this doesn't apply clean on v4-dev-xmr. Can you backport it manually please?

MartijnCuppens:
I'm afraid we can't backport this easily since `.form-controls` can be `select`s too in `v4`. To make this work we'll need to rewrite quite a lot and this might introduce breaking changes.

PR #29126, opened 2019-07-24
v5: Docs Utilities API updates

joseigor:
Improvement to give a better understand of how utilities class name are generated by the (_api.scss) base in the line of code below:
[`$property-class: if($property-class, $property-class, nth($properties, 1));`](https://github.com/twbs/bootstrap/blob/fc02932946424e986a72bb7b47044eab815851cb/scss/mixins/_utilities.scss#L21)


MartijnCuppens (approved):
LGTM

PR #29132, opened 2019-07-25
Add `--mangle-props regex="/^_/"` in terser for our core dist files.

### LINKED ISSUE #23635: Look into enabling mangle-props and mangle-regex terser's options
See #21715

XhmikosR:
bootstrap.min.js: 17.03KB -> 16.35KB (gzip)

For the bundle I'm not sure we should enable it since we need to make sure Popper follows the same scheme.

The gains are pretty big, though. We should really look into enabling this.

- [ ] test the minified dist file to make sure everything works

Fixes #23635


XhmikosR:
Closing this since we can't unit test it right now.

PR #29135, opened 2019-07-25
Drop `.pre-scrollable` class

MartijnCuppens:
So that https://github.com/twbs/bootstrap/pull/28917 can be merged. This class is probably not used that much, we don't even use it in our docs ourselves. Custom css can be written to replace this class.

TODO:
- [x] Document in migration

### DIFF
--- a/scss/_code.scss
+++ b/scss/_code.scss
@@ -38,9 +38,3 @@ pre {
     word-break: normal;
   }
 }
-
-// Enable scrollable blocks of code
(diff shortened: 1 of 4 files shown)


mdo (approved):

vanillajonathan:
For anyone who relied on this class, here is the CSS code previously used.

```css
.pre-scrollable {
  max-height: 340px;
  overflow-y: scroll;
}
````

PR #29138, opened 2019-07-26
Fix dist sync during local dev

mdo:
This updates the css-copy and js-copy tasks to first remove their respective directories, create them again, and then recursively copy over subdirectories and files. This is the only way I've been able to fix the syncing issues I've faced on macOS that don't appear on Windows. Could be a bug in how we cross-env-shell/shx works here?

@MartijnCuppens Can you give this a go locally? I did some obvious changes in the source `scss/**/*.scss` files and saw them sync pretty quickly and without issue on my end. Still have the Hugo crashing at times and not wanting to restart though.


MartijnCuppens:
We'll probably need to use https://www.npmjs.com/package/rimraf to get a cross browser solution (don't think `rm -rf` works in Windows, @XhmikosR will know for sure).

This seems to fix the live reload, but my browser still seems to update multiple times (which is annoying when you're searching something in the search field).

It's an improvement for sure, but we still got to find the root of the problem.

XhmikosR:
>We'll probably need to use https://www.npmjs.com/package/rimraf to get a cross browser solution (don't think rm -rf works in Windows, @XhmikosR will know for sure).

We don't need rimraf since we use shx.

XhmikosR:
Is [this](https://github.com/gohugoio/hugo/issues/6109) something you guys are hitting?

MartijnCuppens:
> Is this something you guys are hitting?

Nope.

I've just updated `hugo-bin` to `0.44.0` and that seemed to fix the watch issue. @mdo, could you check if that works for you?

XhmikosR:
The update is already in #29141 BTW.

XhmikosR:
Let's wait for #29161. If all goes well we might be able to ditch the css/js copy steps and also use Hugo for SRI hashes.

XhmikosR:
Closing this in favor of #29161

PR #29143, opened 2019-07-26
config.yml: add OpenCollective.

XhmikosR:
This was missing since the Hugo switch (d95d300).


MartijnCuppens (approved):
Woohoo, let's get rich!

PR #29144, opened 2019-07-26
Minor doc tweaks

XhmikosR:


XhmikosR:
I'll see about backporting the trailing slashes patch after this is merged.

MartijnCuppens:
Do we even need a trailing slash?

XhmikosR:
Yeah, I'd rather include it than not because usually it's a redirect. In the root URL it doesn't seem to matter, but try this:

```
C:\Users\XhmikosR\Desktop>curl -ILls https://blog.getbootstrap.com/2019/02/13/bootstrap-4-3-1-and-3-4-1 | grep Location
Location: https://blog.getbootstrap.com/2019/02/13/bootstrap-4-3-1-and-3-4-1/
```

Note that this does **not** change the behavior. It just makes it clearer and reduces duplication.

MartijnCuppens (approved):

PR #29146, opened 2019-07-26
Split grid & containers

MartijnCuppens:
The grid and containers are 2 different components, let's split them into different scss files.


mdo:
Why wouldn't we continue to wrap all this in a the `$enable-grid-classes`?

MartijnCuppens:
I don't know, the containers and grid system are two different things IMO.

ysds:
Why not organize the mixin files?

mdo:
Containers are required for grid system though, too, because of the horizontal padding.

ysds:
I agree with @MartijnCuppens. What is required is not a `.container`, but a wrapper with left and right paddings.

MartijnCuppens:
> Containers are required for grid system though, too, because of the horizontal padding.

Valid point. However, the `$enable-grid-classes` name might be confusing. I wouldn't expect the containers from not working anymore when I disable it. If we want a single variable to enable the grid classes and container classes it might be clearer to rename it to `$enable-grid-and-container-classes`.

Or maybe we should introduce a new `$enable-container-classes` variable?

MartijnCuppens:
> Why not organize the mixin files?

Done

site/content/docs/4.3/migration.md
```diff
@@ -35,6 +35,7 @@ Changes to our source Sass files and compiled CSS.
 - Removed print styles and `$enable-print-styles` variable. Print display classes, however, have remained intact. [See #28339](https://github.com/twbs/bootstrap/pull/28339).
 - Dropped `color()`, `theme-color()` & `gray()` functions in favor of variables. [See #29083](https://github.com/twbs/bootstrap/pull/29083)
 - The `theme-color-level()` function is renamed to `color-level()` and now accepts any color you want instead of only `$theme-color` colors. [See #29083](https://github.com/twbs/bootstrap/pull/29083)
+- `$enable-grid-classes` doesn't disable the generation of container classes anymore [See #TODO]()
```
XhmikosR: Don't forget to add the PR number

mdo (approved):
Revisiting this, I'm fine to do the split here and give folks more granular control. @MartijnCuppens let me know if you want to handle in the first alpha or subsequent release.

MartijnCuppens:
:shipit:

PR #29149, opened 2019-07-27
Change property from background to background-image

itxchandan:
For Gradient, We have to use the background-image instead of only background.

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -112,13 +112,13 @@
 .carousel-control-prev {
   left: 0;
   @if $enable-gradients {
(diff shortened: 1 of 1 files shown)


MartijnCuppens (dismissed):
Correct, thanks!

XhmikosR:
Shouldn't we backport this?

XhmikosR:
Yeah, but it's always better to be explicit than using the shorthand which resets many of the other properties. We actually try to do this across the codebase.

/CC @mdo

mdo (approved):
Correct, we always strive to be explicit in our CSS properties and avoid unnecessary shorthand notations.

PR #29152, opened 2019-07-27
Add possibility to use RFS in utility API

### LINKED ISSUE #25832: Type scale map and type utility classes. (Font)

MartijnCuppens:
RFS `v9` [has launched](https://github.com/twbs/rfs/pull/144) and with the new version it will be possible to apply the same rescaling technique to any property.

Features:
- Responsive font sizes are now enabled by default
- Utility API integration
  - Fluid rescaling is also applied to the spacing utilities (option can be disabled)
  - ~New `.font-size-sm`, `.font-size-base`, `.font-size-lg` and `.font-size-xl` utilities~
  - New `.text-sm`, `.text-base`, `.text-lg` and `.text-xl` utilities
- Documentation page with basic how-to-use examples.

Docs: https://deploy-preview-29152--twbs-bootstrap.netlify.com/docs/4.3/getting-started/rfs/
Responsive displays: https://deploy-preview-29152--twbs-bootstrap.netlify.com/docs/4.3/content/typography/#display-headings

Closes #25832

### DIFF
--- a/scss/mixins/_utilities.scss
+++ b/scss/mixins/_utilities.scss
@@ -1,6 +1,6 @@
 // Utility generator
 // Used to generate utilities & print utilities
-@mixin generate-utility($utility, $infix) {
+@mixin generate-utility($utility, $infix, $is-rfs-media-query: false) {
   $values: map-get($utility, values);
 
   // If the values are a list or string, convert it into a map
@@ -25,10 +25,25 @@
     // Don't prefix if value key is null (eg. with shadow class)
     $property-class-modifier: if($key, if($property-class == "" and $infix == "", "", "-") + $key, "");
 
-    .#{$property-class + $infix + $property-class-modifier} {
-      @each $property in $properties {
-        // stylelint-disable-next-line declaration-no-important
-        #{$property}: $value !important;
+    @if map-get($utility, rfs) {
+      // Inside the media query
+      @if $is-rfs-media-query {
+        $val: rfs-value($value);
+
+        // Do not render anything if fluid and non fluid values are the same
+        $value: if($val == rfs-fluid-value($value), null, $val);
+      }
+      @else {
+        $value: rfs-fluid-value($value);
+      }
+    }
+
+    @if $value != null {
+      .#{$property-class + $infix + $property-class-modifier} {
+        @each $property in $properties {
+          // stylelint-disable-next-line declaration-no-important
+          #{$property}: $value !important;
+        }
       }
     }
   }
--- a/site/content/docs/4.3/utilities/api.md
+++ b/site/content/docs/4.3/utilities/api.md
@@ -13,6 +13,7 @@ The `$utilities` map contains all utilities and is later merged with your custom
 
 - `property`: Name of the property, this can be a string or an array of strings (needed for eg. horizontal paddings or margins).
 - `responsive` _(optional)_: Boolean indicating if responsive classes need to be generated. `false` by default.
+- `rfs` _(optional)_: Variable to enable fluid rescaling. Have a look at the [RFS]({{< docsref "/getting-started/rfs" >}}) page to find out how this works. `false` by default.
 - `class` _(optional)_: Variable to change the class name if you don't want it to be the same as the property. In case you don't provide the `class` key and `property` key is an array of strings, the class name will be the first element of the `property` array.
 - `values`: This can be a list of values or a map if you don't want the class name to be the same as the value. If null is used as map key, it isn't rendered.
 - `print` _(optional)_: Boolean indicating if print classes need to be generated. `false` by default.
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -23,6 +23,8 @@ html {
 @import "mixins/grid";
 @import "mixins/utilities";
 
+@import "vendor/rfs";
+
 @import "grid";
 
 @import "utilities";
--- a/scss/utilities/_api.scss
+++ b/scss/utilities/_api.scss
@@ -16,6 +16,24 @@
   }
 }
 
+// RFS rescaling
+@media (min-width: $rfs-mq-value) {
+  @each $breakpoint in map-keys($grid-breakpoints) {
+    $infix: breakpoint-infix($breakpoint, $grid-breakpoints);
+
+    @if (map-get($grid-breakpoints, $breakpoint) < $rfs-breakpoint) {
+      // Loop over each utility property
+      @each $key, $utility in $utilities {
+        // The utility can be disabled with `false`, thus check if the utility is a map first
+        // Only proceed if responsive media queries are enabled or if it's the base media query
+        @if type-of($utility) == "map" and map-get($utility, rfs) {
+          @include generate-utility($utility, $infix, true);
+        }
+      }
+    }
+  }
+}
+
 
 // Print utilities
 @media print {


mdo:
@MartijnCuppens One of the things I've loved about Primer at GitHub is responsive `.f*` and `.h*` classes. We get this now with the heading classes (`.h1`-`.h6`), but those come with `margin-bottom`. Any chance we could get just the `font-size`s as `.f1`-`.f6`? Unsure what needs to be done in RFS or if this is just a Bootstrap thing.

MartijnCuppens:
The Utility API is used to generate the font size utilities, and it seemed that I made a mistake in the documentation above, because I used the `text` class:

https://github.com/twbs/bootstrap/blob/fb36c60a75d43c60f61f9f125412ac984c419d8d/scss/_utilities.scss#L352-L362

So actually, the `.text-sm`, `.text-base`, `.text-lg` and `.text-xl` classes are generated.

We could change the change the class to `f`, but than the dash between the classes will still be generated (eg. `.f-3`). We could move the classes away from the utility API, but than these classes will be less extendible.

scss/vendor/_rfs.scss
```diff
@@ -2,38 +2,41 @@
 
 // SCSS RFS mixin
 //
-// Automated responsive font sizes
+// Automated responsive values for for font sizes, paddings, margins and much more
```
XhmikosR: double for

site/content/docs/4.3/content/typography.md
```diff
@@ -284,6 +284,4 @@ Align terms and descriptions horizontally by using our grid system's predefined
 
 ## Responsive font sizes
 
-Bootstrap v4.3 ships with the option to enable responsive font sizes, allowing text to scale more naturally across device and viewport sizes. <abbr title="Responsive font sizes">RFS</abbr> can be enabled by changing the `$enable-responsive-font-sizes` Sass variable to `true` and recompiling Bootstrap.
-
-To support <abbr title="Responsive font sizes">RFS</abbr>, we use a Sass mixin to replace our normal `font-size` properties. Responsive font sizes will be compiled into `calc()` functions with a mix of `rem` and viewport units to enable the responsive scaling behavior. More about <abbr title="Responsive font sizes">RFS</abbr> and its configuration can be found on its [GitHub repository](https://github.com/twbs/rfs).
+In Bootstrap `v5`, we've enabled responsive font sizes by default, allowing text to scale more naturally across device and viewport sizes. Have a look at the [RFS]({{< docsref "/layout/overview#responsive-breakpoints" >}}) page to find out how this works.
```
XhmikosR: Is the link right? Or should it be the rfs page?

scss/vendor/_rfs.scss
MartijnCuppens: Upstream PR: https://github.com/twbs/rfs/pull/161

ysds:
I prefer to separate the addition of the text utilities from this PR for clarification.

scss/mixins/_utilities.scss
```diff
@@ -1,6 +1,6 @@
 // Utility generator
 // Used to generate utilities & print utilities
-@mixin generate-utility($utility, $infix) {
+@mixin generate-utility($utility, $infix, $isRfsMediaQuery: false) {
```
ysds: Is there an intention for the variable name of uppercase? I'm concerned about the consistency.

MartijnCuppens: Good call, changed it to kebab case

MartijnCuppens:
> I prefer to separate the addition of the text utilities from this PR for clarification.

It's kinda hard to separate this now. The documentation changes and the standalone RFS page cover as well the responsive font sizes as the the utility API implementation.

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -271,6 +271,7 @@ You can find and customize these variables for key global options in Bootstrap's
 | `$enable-caret`                              | `true` (default) or `false`        | Enables pseudo element caret on `.dropdown-toggle`. |
 | `$enable-pointer-cursor-for-buttons`         | `true` (default) or `false`        | Add "hand" cursor to non-disabled button elements. |
 | `$enable-rfs`                                | `true` (default)  or `false`       | Globally enables [RFS]({{< docsref "/getting-started/rfs" >}}). |
+| `$enable-rfs-for-spacing-utilities`          | `true` or `false` (default)        | Enables RFS for [spacing utilities]({{< docsref "/utilities/spacing" >}}). |
```
midzer: As `$enable-rfs-for-spacing-utilities` default value is set to true, we have to adapt the docs

site/content/docs/4.3/utilities/api.md
```diff
@@ -13,6 +13,7 @@ The `$utilities` map contains all utilities and is later merged with your custom
 
 - `property`: Name of the property, this can be a string or an array of strings (needed for eg. horizontal paddings or margins).
 - `responsive` _(optional)_: Boolean indicating if responsive classes need to be generated. `false` by default.
+- `rfs` _(optional)_: Variable to enable fluid rescaling. Have a look at the [RFS]({{< docsref "/getting-started/rfs" >}}) page to find out how this works. `false` by default.
```
midzer: `$enable-rfs` default value is set to `true`. So we have to adapt the boolean value in the docs as well. Maybe change the sentence to something like `Variable to disable fluid rescaling [...]` and change default value accordingly.

midzer (changes requested):
Just two minor adjustments necessary for the docs.
Can't comment on the RFS implementation itself, as I did not test it on a live site yet.

PR #29157, opened 2019-07-28
remove underscore for static methods

Johann-S:


XhmikosR:
We should probably mention this in Migration.

XhmikosR (approved):

PR #29161, opened 2019-07-29
Make use of Hugo's 0.56+ module feature.

### LINKED ISSUE #29142: See if we can use Hugo v0.56's enhancements
https://gohugo.io/news/0.56.0-relnotes/

XhmikosR:
Need to make sure we don't have the dist folder copied locally before testing this.

Fixes #29142


XhmikosR:
Maybe this requires go to be available, in which case we can't proceed with this. Adding a new devDependency isn't so good...

XhmikosR:
@bep: I read the docs, but I'm probably missing something. Is this how modules are supposed to work or do we need Go installed for this too?

config.yml
```diff
@@ -26,6 +26,13 @@ layoutDir:              "site/layouts"
 staticDir:              "site/static"
 themesDir:              "site/themes"
 
+module:
+  hugoVersion:
+    min: 0.56.0
+  mounts:
+    - source:           "dist"
+      target:           "site/static/docs/4.3/dist"
```
bep: Target must start with one of `static`, `content` etc. They are "virtual folders", so to speak.

bep: Also (I'm pretty sure, but I got a little bit overloaded when implementing this): When you create a mounts section, you need to add all you need, e.g. "static => static", "content => content". I.e. is you're specific we don't add the defaults.

XhmikosR: I tried it with this too but the dist folder isn't copied to the output dir.

```yml
module:
  hugoVersion:
    min: 0.56.0
  mounts:
    - source:           dist
      target:           static/docs/4.3/dist
```

Maybe I'm missing something? The first issue is that we manually need to copy the dist folder into the static folder and generating the SRI hashes is the second one, which like you said, I need to place the files into the assets folder.

bep: static is copied, assets is not (only the assets you process and link to). You may create 2 mounts with the same source if you want a combination.

XhmikosR: Weird, whatever I have tried, the dist folder is never copied for me into the publishDir hence why CI is failing due to the missing assets.

bep: I suggest you let this PR cool for a week or so and then ping me. I'm a little bot overloaded in the notification department at the moment. Which is a bad thing considering I'm on vacation :-)

config.yml
```diff
@@ -26,6 +26,27 @@ layoutDir:              "site/layouts"
 staticDir:              "site/static"
 themesDir:              "site/themes"
```
bep: OK, I have checked the Hugo code. I have tried to be really careful about not breaking existing sites, so the logic is this:

* If there is a legacy config setting for a given target (e.g. `static`), we're ignoring any mounts.

So, in your case, you need to remove the `staticDir` setting above (also if you have others).

XhmikosR: I still can't make it to work, so I'll put it on hold until you are back and we figure this out.

If we manage to skip the copying step and also generate the SRI hashes natively in Hugo, this would be a huge win for us.

BTW, Thank you for all the hard work :)

bep: I will put this issue in the back of my head, but please ping me next week if I forget. What you want to do should not be a problem, and I don't think there should be any bug to prevent you from doing it, either, but one never knows ...

bep: OK, I found a related bug in Hugo. @XhmikosR could I get commit access to this PR so I can use this to test that fix? You will still have to wait for Hugo 0.56.3, but that should be a day or two.

XhmikosR: Unfortunately this branch is on the main repo so I can't give you rights. Can you fork the main repo and push to this branch on your fork?

bep: Sure.

bep:
This commit works with the unreleased (and unmerged) Hugo version: https://github.com/bep/bootstrap/commit/154417df22aa442b67e593aab0cf301a7f4cfc41

XhmikosR:
Nice! Can't wait to try it when the next version is out :)

BTW, just thinking out loud for the sake of flexibility. Wouldn't it be more flexible if one didn't have to add all mount points, but only the ones they want to use? I know our use case might not be the most common one, but still. Maybe I'm missing something, I'll play with it when the new version is out :)

bep:
>didn't have to add all mount points, but only the ones they want to use? I

You only have to add the mounts you use. There are some superfluous mounts there now. (archetypes? i18n?), but I didn't know the project good enough to remove them.

XhmikosR:
Perfect, thanks! I'll play around with the branch when 0.56.3 is out and let you know if we hit any more issues.

XhmikosR:
@bep: I just gave this a go and works pretty good. Thanks for the fast fixes!

The only thing I noticed when playing around with SRI generation (I dropped the patch from this branch now though), and I had set another mount to `assets/dist` in order for me to use `resources.Get`, then we had another dist folder in `publishDir/dist`.

I'll make a branch for this later and see if we have any issues there.

config.yml
```diff
@@ -18,13 +18,21 @@ blackfriday:
...
-assetDir:               "site/assets"
-contentDir:             "site/content"
-dataDir:                "site/data"
-layoutDir:              "site/layouts"
-staticDir:              "site/static"
-themesDir:              "site/themes"
+
+module:
+  hugoVersion:
+    min:                0.56.3
```
XhmikosR: Maybe we don't need to set `hugoVersion` at all?

bep: That value is mostly for theme creators -- to warn people using your module.

XhmikosR: Oh, I see, so it doesn't really make sense for us here. I will remove it.

XhmikosR:
@bep: can you checkout this branch and do `npm run docs-serve` in one cmd and `npm run watch-css-main` in another? Hugo complains it can't find the file. I'm on Windows but @MartijnCuppens is on macOS and happens there too. Maybe the changes are too many or happen too fast and Hugo doesn't catch them?

If I change an HTML or markdown file and hugo rebuilds the site, then the error goes away.

mdo:
Confirming the npm scripts run for me, but at `localhost:9001` I see `file does not exist` errors.

MartijnCuppens:
No `file does not exist` errors here. The docs css and js just don't appear in the dom, this a screenshot from the homepage where you should see the Bootstrap css:
image

Styles on the Netlify preview:
image

XhmikosR:
I definitely don't have the issue @MartijnCuppens mentions. I do have the same issue @mdo says above with watch. This doesn't happen for me on Windows without the mount changes.

XhmikosR:
In fact, I'm able to trigger the file not found error if I modify dist/css/bootstrap.css manually:

```
C:\Users\xmr\Desktop\bootstrap>npm run docs-serve

> bootstrap@4.3.1 docs-serve C:\Users\xmr\Desktop\bootstrap
> hugo server --port 9001 --disableFastRender

Building sites …
                   | EN
+------------------+-----+
  Pages            | 105
  Paginator pages  |   0
  Non-page files   |  20
  Static files     | 125
  Processed images |   0
  Aliases          |  16
  Sitemaps         |   1
  Cleaned          |   0

Total in 792 ms
Watching for changes in C:\Users\xmr\Desktop\bootstrap\{dist,site}
Watching for config changes in C:\Users\xmr\Desktop\bootstrap\config.yml
Environment: "development"
Serving pages from memory
Web Server is available at http://localhost:9001/ (bind address 127.0.0.1)
Press Ctrl+C to stop

Change of Static files detected, rebuilding site.
2019-08-02 12:55:25.701 +0300
Syncing \css\bootstrap.css to C:\Users\xmr\Desktop\bootstrap\
ERROR 2019/08/02 12:55:25 file does not exist
ERROR 2019/08/02 12:55:25 file does not exist
```

XhmikosR:
I think now we are ready to merge this. I tested it on Windows, no errors when watching.

Huge thanks to @bep for working towards fixing the bugs we found with mounts!

XhmikosR:
I need an approval @mdo @Johann-S @MartijnCuppens after you guys try this (make sure you do `npm i` and remove the `old docs/4.3/dist` folder locally)

MartijnCuppens (approved):
It works here!

PR #29172, opened 2019-07-30
Add missing role="document"

Herst:
Self-explanatory. Found using upcoming Bootlint.


patrickhlauke:
i'd need to test, but i seem to remember that `role="document"` was only needed because NVDA decided to treat dialogs with `role="dialog"` as forms, switch into forms mode by default, and prevent users from reading the actual content normally, and i seem to remember that NVDA finally changed that odd behavior of theirs in this regard...

patrickhlauke:
ok, based on my sources (well, @scottaohara), the NVDA issue was fixed, but `role="document"` is indeed still needed as otherwise VoiceOver has issues...so long story short, yes this is needed.

patrickhlauke (approved):

scottaohara:
some context to what @patrickhlauke mentioned:

If using macOS VoiceOver's (tested with version 8 and 9) Quick Navigation the contents of modal dialogs may be inaccessible if the dialog element has an accessible name set by `aria-label` or `aria-labelledby`. 

To mitigate this, either do not set an accessible name to the modal dialog, or wrap the contents of the modal dialog within a `role="document"` and then the content of the dialog will be navigable whether Quick Nav is on or off...

PR #29197, opened 2019-08-03
Updated callouts design for v5

mdo:
Screen Shot 2019-08-03 at 7 21 15 AM


XhmikosR:
Can someone confirm the contrast ratio is OK now?

XhmikosR:
Another point is this might look weird when we have a code snippet in a callout (for example yellow bg and light gray for the code block). But it's your call.

ysds:
https://codepen.io/anon/pen/dxZEXz Please click the glasses icon at the bottom left, and check the "Contrast".

mirorauhala:
Just leaving this here. I like this tool because it calculates the colors that are accessible. Makes it so much easier to work with accessible colors.

https://accessible-colors.com/

PR #29198, opened 2019-08-03
Tweak colors to be WCAG 2AA valid

XhmikosR:
Feel free to push here any further color tweaks. I check this on our docs page, but the highlighting colors mostly don't pass WCAG 2.0 AA so this makes it harder.

Also, not sure if this should be backported or not.

Preview: <https://deploy-preview-29198--twbs-bootstrap.netlify.com/>

TODO:

- [x] darken links color - `#0d6efd`
- [x] darken pink color - `#d63384`


XhmikosR:
I added a small TODO. After this we should be mostly covered.

XhmikosR:
I split the doc-related fixes to #29208. I'll rebase this after the merge and see where we are at.

XhmikosR:
@patrickhlauke: any suggestions?

XhmikosR:
Maybe we should move forward with the blue and pink changes only since those are the least intrusive ones?

MartijnCuppens:
> Maybe we should move forward with the blue and pink changes only since those are the least intrusive ones?

Yup

XhmikosR:
Alright, I cleaned up the patch. I don't think we should backport this, though, because it can be a breaking change for some people.

@mdo please confirm if you are OK with these 2 color changes. The difference isn't that big.

mdo (approved):

PR #29221, opened 2019-08-07
Update devDependencies.

XhmikosR:


XhmikosR:
Actually, I'll wait until the Hugo issue is fixed before updating the deps. The current version still has the error when a static file in mounts changes.

Johann-S (approved):

Leenie83 (approved):

PR #29222, opened 2019-08-07
Remove unneeded Stylelint suppressions.

XhmikosR:
Found with `--report-needless-disables`. Not sure if we should add this yet.

### DIFF
--- a/scss/_type.scss
+++ b/scss/_type.scss
@@ -1,5 +1,3 @@
-// stylelint-disable declaration-no-important, selector-list-comma-newline-after
-
 //
 // Headings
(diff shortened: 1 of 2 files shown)


MartijnCuppens (approved):
`--report-needless-disables` might be useful to add indeed

XhmikosR:
My main problem with this switch is

1. it doesn't say which disable rule is redundant
2. with caching locally this can cause issues when changing the inline suppressions. CI will fail properly though

MartijnCuppens:
Hmm, leave it for now then

PR #29229, opened 2019-08-08
Prep for migration to GitHub Actions CI/CD

mdo:
Woo, glad to be in on the beta launch of the new Actions! I've had a chance to play with this with @xhmikosr over the last couple weeks and it's been fun. We still have some work to do to bring it all up to speed and get this integrated with v4, but this PR adds Actions CI for Bootstrap 5.

Here's what's changed:

- Adds new `.github/workflows/test.yml`.
- Actions CI runs on three versions of Node like our current Travis build. It's triggered on `push` and `pull_request`, so this covers us maintainers and forks I believe.
- Separate from this PR, I've also created the new secrets in our repo settings.
- Adds an unofficial badge to the project readme (official ones are coming I think),
- Removed the `.travis.yml` file.

Still more to get figured out while the beta happens, but stoked to have been able to get this up and running!


.github/workflows/test.yml
```diff
@@ -0,0 +1,44 @@
...
+        fetch-depth: 3
+
+    - name: Set Node.js version
+      uses: actions/setup-node@v1.1.0
+      with:
+        version: ${{ matrix.node }}
+
+    - run: node --version
+    - run: npm --version
+    - run: java -version
```
Johann-S: Where do we need Java ? 😯

XhmikosR: 1, Java is already installed anyway
2. It's needed for vnu-jar

XhmikosR (approved):

PR #29238, opened 2019-08-10
Reindent workflows/test.yml and limit bundlesize to pushes

XhmikosR:


XhmikosR:
@mdo I added one small patch here that I makes sense to me.

mdo (approved):

PR #29240, opened 2019-08-10
Use `hugo.Environment` instead of `getenv`

XhmikosR:
This is set automatically to "development" when the local server is running, and to "production" when Hugo builds the site.

~~WIP because I like the flexibility. I might change the patch to set the environment with Hugo's `--environment` CLI flag instead.~~


mdo (approved):
Haven't tested this locally, but diff looks right to me!

PR #29262, opened 2019-08-17
Using Hugo for our docs Scss and JS.

XhmikosR:
Still investigating this, but simplifies the build process even more.

A few notes

- [ ] I can't seem to be able to use a method to get all JS files in the folder in a simple manner, but maybe I'm missing something
- [ ] The JS minify tool is very basic

@mdo @Johann-S @MartijnCuppens let me know your thoughts and I'll try to simplify this even further.


XhmikosR:
I'll experiment this further. I meant to make this a draft PR.

PR #29264, opened 2019-08-17
Don't bind validation to form controls

MartijnCuppens:
Alternative solution for https://github.com/twbs/bootstrap/pull/29228. This PR doesn't bind the `.is-valid` and `.is-invalid` classes to the form control.

Issue: https://codepen.io/browner12/pen/zgjKbJ?editors=1000
Solution: https://codepen.io/MartijnCuppens/pen/ExYyXBy?editors=1100

CC @browner12 

Closes https://github.com/twbs/bootstrap/pull/29228


browner12:
Okay, so instead of binding to `.form-control .is-valid`, we are now just binding to `is-valid`, correct?

I'm okay with that solution, but it seems like this is probably a breaking change since we're becoming less specific, so it can only go to 5?

MartijnCuppens:
We're switching from `.form-control.is-valid ~ .valid-feedback` to `.is-valid ~ .valid-feedback`. Since `~` only targets siblings, it's pretty safe to merge without introducing breaking changes IMO.

browner12:
I would agree the risk is **extremely** low. Wasn't sure what our threshold is in this project.

ysds (approved):
@mdo Do you have any concerns?

mdo (approved):
No concerns here!

PR #29268, opened 2019-08-17
Add deprecation warning

MartijnCuppens:
Add deprecation warning for `text-emphasis-variant()` mixin which will be dropped by https://github.com/twbs/bootstrap/pull/29267

### DIFF
--- a/scss/mixins/_text-emphasis.scss
+++ b/scss/mixins/_text-emphasis.scss
@@ -2,7 +2,7 @@
 
 // Typography
 
-@mixin text-emphasis-variant($parent, $color) {
+@mixin text-emphasis-variant($parent, $color, $ignore-warning: false) {
   #{$parent} {
     color: $color !important;
   }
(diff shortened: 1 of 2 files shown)


bardiharborow (approved):
This is fine, provided #29267 merges.

mdo (approved):

ELI7VH:
what do we use instead?

MartijnCuppens:
@ELI7VH,

From `v5` you will be able to use the text utilities https://twbs-bootstrap.netlify.com/docs/4.3/utilities/colors/#color to colorize text and https://twbs-bootstrap.netlify.com/docs/4.3/helpers/colored-links/ for links with hover states.

Eonasdan:
@MartijnCuppens doc links are dead. I got a theme using this and I'd like to fix it.

MartijnCuppens:
I've just updated them. You can still use `text-emphasis-variant()` in `v4`, but make sure the 3th parameter is `true`.

digitaldonkey:
It might be helpful to provide an alternative for BS 4.4+ when you add a deprecation warning. 
Links above do not provide an alternative for the **mixing**.

PR #29270, opened 2019-08-18
change-version.js: Add _gh_pages folder in ignores.

XhmikosR:


Johann-S (approved):

build/change-version.js
```diff
@@ -90,6 +90,7 @@ function main(args) {
   const newVersion = args[1]
   const EXCLUDED_DIRS = new Set([
     '.git',
+    '_gh_pages',
```
XhmikosR: BTW this could be more clever like read .gitignore and add all of those in the excluded files/dirs. But that's for another time

bardiharborow (approved):

PR #29271, opened 2019-08-18
Clean up line heights & add line height utilities

### LINKED ISSUE #28622: Line height text utility classes
It would be great to be able to set narrower line-heights via text utility classes such as `text-line-height-narrow`, `text-line-height-narrower`, etc. similar to font weight.

Would a PR be appreciated?

### LINKED ISSUE #28907: Ditch line heights
A lot of our components contain `line-height`s, and `sm/lg` versions with configurable `line-height`s. I wonder if it's really needed for all those components to have configurable `line-heights`, a mixed usage of different line heights might result in an inconsistent design, so I really doubt people really change these values per component. Also these variables look a bit strange: https://github.com/twbs/bootstrap/blob/0829decce075e190c0bb6e9992ca2a1ca56c60c7/scss/_variables.scss#L219-L220

Imo, we can just remove the different line-heights for `sm/lg` versions which would make our codebase a bit less bloated. 

We might add some utilities to increase or decrease line heights though.

Thoughts @twbs/css-review?

MartijnCuppens:
The small and large `line-height`s are the same for `sm` and `lg` components, so they can easily be dropped. I still left the ability to change the line-heights for buttons.

Fixes #28622
Fixes #28907

Preview: https://deploy-preview-29271--twbs-bootstrap.netlify.com/docs/4.3/utilities/text/#line-height

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -394,6 +394,16 @@ $utilities: map-merge(
         )
       )
     ),
+    "line-height": (
+      property: line-height,
+      class: lh,
+      values: (
+        1: 1,
+        sm: $line-height-sm,
+        base: $line-height-base,
+        lg: $line-height-lg,
+      )
+    ),
     "background-color": (
       property: background-color,
       class: bg,
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -8,6 +8,7 @@
   display: inline-block;
   font-family: $btn-font-family;
   font-weight: $btn-font-weight;
+  line-height: $btn-line-height;
   color: $body-color;
   text-align: center;
   vertical-align: middle;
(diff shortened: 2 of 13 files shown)


scss/_utilities.scss
```diff
@@ -361,6 +361,16 @@ $utilities: map-merge(
         reset: inherit,
       )
     ),
+    "line-height": (
+      property: line-height,
+      class: lh,
+      values: (
```
mdo: Maybe we use a similar numerical scale as our margin/padding utilities?

MartijnCuppens: I've been thinking about that, but I'm not sure what values to use here. `lh-1` would make sense for a line height of `1`, but what could we use for the other values?

mdo: Yeah, it'd be too weird to do anything numerical. Maybe something else using similar vocabulary from the rest of the project?

- `lh-1`
- `lh-compact`
- `lh-base`
- `lh-spacious`

But I think `sm`, `base`, and `lg` are best.

mdo (approved):

XhmikosR:
Added links in the migration page. @MartijnCuppens LGTY?

MartijnCuppens:
Excellent, @XhmikosR!

PR #29274, opened 2019-08-18
Responsive Card Decks alternative

MartijnCuppens:
Alternative implementation of https://github.com/twbs/bootstrap/pull/29073. In draft to show difference.

The `.row-cols-*` modifier now targets all children, and the `.col-` classes can still be used to override this behaviour if needed. This implementation has cards of `100%`, which makes them have equal heights by default.

TODOs:
- [ ] Make the card height variable.


MartijnCuppens:
The `height: 100%` will be a breaking change for `v4`, so maybe I should change that in a configurable `null` variable.

MartijnCuppens:
https://github.com/twbs/bootstrap/pull/29317 makes more sense, closing this

PR #29278, opened 2019-08-19
V4 dev backports

XhmikosR:


XhmikosR:
I need to fix Netlify, it seems it fails with `Could not find 'bundler' (< 2) - did find: [bundler-2.0.2]`

Johann-S (approved):

PR #29280, opened 2019-08-19
Use Hugo for our docs Sass and JS.

XhmikosR:
Pros:

1. we simplify our docs development process and makes the whole process faster; `npm run docs-serve` and everything doc-related will be handled by Hugo
2. we get rid of the versioned folder for CSS and JS
3. we get rid of docs.min.css, docs.min.css.map and docs.min.js files from the repo
4. the `assets/scss` folder is no longer output to the build folder
5. ~~we have SRI for the docs CSS, but perhaps I should remove this in case we want to edit docs.css in the future so that we don't have to edit the SRI too?~~ - **EDIT**: reverted this

Cons:

1. the JS minifier is pretty basic. That's OK, we won't have the possible best compression of docs.min.js, but being that we already use minified vendor JS files, the difference shouldn't be that big. **EDIT**: it's **~350 bytes** gzipped.
2. for CSS I opted to not minify the CSS at all, but rather just use the minified output from Sass. The CSS minifier does some stuff I don't like, so we better not add another POF. The difference is negligible, **~15 bytes gzipped** - **Beautified diff**: https://www.diffchecker.com/AWtS29IX
3. we don't have sourcemaps for production. There's nothing we can do currently about this. I have them enabled for Sass only for development
4. we don't have PostCSS for development. It's either sourcemaps or PostCSS due to a limitation. Let me know how to proceed. Upstream issue: https://github.com/gohugoio/hugo/issues/6189
5. there's a chance the [Sass lib](https://github.com/wellington/go-libsass) Hugo is using has a bug, but it's well tested. It's just that this adds another potential POF. We shouldn't worry about it a lot I guess, though.

To try it, switch to this branch and do `rm -rf ./node_modules/ && npm i`; this is because we switch to the Extended Hugo version which has Sass support.

@mdo @MartijnCuppens @Johann-S @ysds @bardiharborow please try it and let me know your thoughts.


MartijnCuppens:
Hooray for all the pro arguments!

> the JS minifier is pretty basic. That's OK, we won't have the possible best compression of docs.min.js, but being that we already use minified vendor JS files, the difference shouldn't be that big. EDIT: it's ~350 bytes gzipped.

I don't have a strong opinion on this, so I don't mind the extra bytes

> for CSS I opted to not minify the CSS at all, but rather just use the minified output from Sass. The CSS minifier does some stuff I don't like, so we better not add another POF.

What stuff are we talking about here?

> we don't have sourcemaps for production. There's nothing we can do currently about this. I have them enabled for Sass only for development

I personally don't care, so fine by me.

> we don't have PostCSS for development. It's either sourcemaps or PostCSS due to a limitation. Let me know how to proceed.

That's a bummer. I would prefer to have autoprefixer run in development instead of the sourcemaps so that we can debug local.

> there's a chance the Sass lib Hugo is using has a bug, but it's well tested. It's just that this adds another potential POF. We shouldn't worry about it a lot I guess, though.

What exactly is this about?


So PostCSS does run on production, doesn't it?

XhmikosR:
>What stuff are we talking about here?

You can try it yourself too by adding `resources.Minify`, but from a quick look it base64 encoded the SVGs which I definitely didn't like. And since the gain isn't big, I'd rather not introduce another POF.

>That's a bummer. I would prefer to have autoprefixer run in development instead of the sourcemaps so that we can debug local.

That's easy to change, and things might improve eventually in this regard, see the upstream issue.

>What exactly is this about?

Not sure what you mean. By POF I mean point of failure. Previously we used node-sass + clean-css for both core and docs. With this PR, we switch to Hugo's internal stuff for docs Sass, which in theory, could cause issues we might not see easily. I doubt it, but I thought I'd mention the change in behavior.

>So PostCSS does run on production, doesn't it?

Yeah, for production we have everything we had, minus CSS minification. But you can see the CSS diff in my OP.

Johann-S:
It works very well, I like it! Nice job @XhmikosR 👍

package.json
```diff
@@ -48,20 +42,18 @@
...
-    "js-minify-docs": "cross-env-shell terser --mangle --comments \\\"/^!/\\\" --output site/static/docs/$npm_package_version_short/assets/js/docs.min.js site/static/docs/$npm_package_version_short/assets/js/vendor/anchor.min.js site/static/docs/$npm_package_version_short/assets/js/vendor/clipboard.min.js site/static/docs/$npm_package_version_short/assets/js/vendor/bs-custom-file-input.min.js \"site/static/docs/$npm_package_version_short/assets/js/src/*.js\"",
     "js-test": "npm-run-all --parallel js-test-karma js-test-integration",
     "js-debug": "cross-env DEBUG=true karma start js/tests/karma.conf.js",
     "js-test-karma": "karma start js/tests/karma.conf.js",
     "js-test-integration": "rollup --config js/tests/integration/rollup.bundle.js",
     "js-test-cloud": "cross-env BROWSER=true npm run js-test-karma",
     "lint": "npm-run-all --parallel js-lint css-lint",
     "coveralls": "shx cat js/coverage/lcov.info | coveralls",
-    "docs": "npm-run-all css-docs js-docs docs-build docs-lint",
```
XhmikosR: Maybe I should keep this script to just run `docs-build docs-lint`? Does anyone use it?

XhmikosR:
@mdo @ysds can you guys try this and let me know? It makes things a lot simpler for docs development.

We just need to agree how to handle sourcemaps/postcss on dev environments.

ysds:
@XhmikosR Thanks for your nice job!

The sourcemap of the doc.css is not important for me. I'm sure that users would accept this inconvenience.

> We just need to agree how to handle sourcemaps/postcss on dev environments.

PostCSS should be given priority than sourcemaps for cross-browser testing on develop env.

XhmikosR:
Alright, I enabled PostCSS for dev too.

package.json
```diff
@@ -21,44 +21,37 @@
...
-    "css-minify-docs": "cross-env-shell cleancss --level 1 --format breakWith=lf --source-map --source-map-inline-sources --output site/static/docs/$npm_package_version_short/assets/css/docs.min.css site/static/docs/$npm_package_version_short/assets/css/docs.min.css",
     "css-prefix": "npm-run-all --parallel css-prefix-*",
     "css-prefix-main": "postcss --config build/postcss.config.js --replace \"dist/css/*.css\" \"!dist/css/*.min.css\"",
-    "css-prefix-docs": "cross-env-shell postcss --config build/postcss.config.js --replace \"site/static/docs/$npm_package_version_short/assets/css/*.css\"",
-    "css-prefix-docs-examples": "postcss --config build/postcss.config.js --replace \"site/content/**/*.css\"",
+    "css-prefix-examples": "postcss --config build/postcss.config.js --replace \"site/content/**/*.css\" \"site/content/**/*.css\"",
```
XhmikosR: Hmm, I need to add this back to a script that runs it.

bep:
>there's a chance the Sass lib Hugo is using has a bug, 

Just a quick note that LibSASS (which Hugo uses and we do pull in updates) is one of the major SASS implementations, and I suspect also what sits behind what you currently use. It's a C++ library, but one of those that are battle tested.

I think this is cool PR. I'm biased, but I find that the speed of the feedback when doing CSS etc. tweaks makes for a great editing experience ...

XhmikosR:
Yeah, we use Node-sass which is using libsass too. I was mostly referring to the go implementation, but like I said it's a small chance. It's just that traditionally we have been using the same tooling everywhere.

Personally I find this patch a big improvement. There is the downside of sourcemaps, but for now we should be fine. The minifiers I don't mind so much since in this case we don't lose a lot.

package.json
```diff
@@ -21,44 +21,37 @@
...
     "js-main": "npm-run-all js-lint js-compile js-minify-main",
-    "js-docs": "npm-run-all js-lint js-minify-docs",
     "js-compile": "npm-run-all --parallel js-compile-*",
     "js-compile-standalone": "rollup --environment BUNDLE:false --config build/rollup.config.js --sourcemap",
     "js-compile-standalone-esm": "rollup --environment ESM:true,BUNDLE:false --config build/rollup.config.js --sourcemap",
     "js-compile-bundle": "rollup --environment BUNDLE:true --config build/rollup.config.js --sourcemap",
     "js-compile-plugins": "node build/build-plugins.js",
     "js-lint": "eslint --cache --cache-location .cache/.eslintcache .",
-    "js-minify": "npm-run-all --parallel js-minify-main js-minify-docs",
+    "js-minify": "npm-run-all --parallel js-minify-main",
```
MartijnCuppens: This line can be combined with `js-minify-main` I think?

XhmikosR: True, there's a discrepancy in the minify js scripts. I'll see how to proceed with this

package.json
```diff
@@ -32,15 +32,14 @@
     "css-prefix-main": "postcss --config build/postcss.config.js --replace \"dist/css/*.css\" \"!dist/css/*.min.css\"",
     "css-prefix-examples": "postcss --config build/postcss.config.js --replace \"site/content/**/*.css\"",
     "js": "npm-run-all js-compile js-minify",
-    "js-main": "npm-run-all js-lint js-compile js-minify-main",
+    "js-main": "npm-run-all js-lint js-compile js-minify",
```
XhmikosR: @Johann-S: can you confirm these are OK with you?

XhmikosR:
Alright, I think I'm done with the script changes. It's messy and things can easily break but I hope everything is good. Everything should be faster now.

MartijnCuppens (approved):

XhmikosR:
@mdo: please try this, I want to merge it ASAP because it simplifies things *a lot*.

Johann-S (approved):

PR #29281, opened 2019-08-19
Update devDependencies.

XhmikosR:


XhmikosR:
Hmm, it seems the new ESLint version introduce some false errors, I'll wait for a fix.

tmorehouse:
Upstream with babel-eslint: https://github.com/babel/babel-eslint/issues/791

XhmikosR:
Locked eslint to 6.1.0 but applied the eslint-utils update for the vulnerabilities fix.

Not sure when the bug will be fixed, so we can go with this in the meantime.

Johann-S (approved):

PR #29283, opened 2019-08-21
Use Hugo mounts for our docs vendor files.

### LINKED ISSUE #27412: Find a way to consistently use deps accross the whole codebase
- [x] docs: jQuery, popper
- [ ] docs: anchor.js, docsearch, clipboard.js - See #29283
- [x] tests:  jQuery, popper
- [x] tests visual:  jQuery, popper
- [x] docs SRI hashes: currently popper.js isn't in sync with the version we use in our docs, it requires manual changes

I'd also stop using the CDN for the docs for these files since we already are behind Cloudflare. Thus simplifying things a bit.

We need to come up with a way to automate all this once and for all. Like use npm for everything, and have a script which will take care of this.

XhmikosR:
Draft since currently we can't specify directories with holes in mounts.

~~Also, it seems there isn't any error checking for file existence yet, which can be dangerous, but only for local development. In production, terser will throw.~~ But if we eventually move with #29280, we need this to throw in case a mount does not exist.

Fixes #27412.

Preview: <https://deploy-preview-29283--twbs-bootstrap.netlify.app/>


XhmikosR:
@bep: any chance we could get this improved?

XhmikosR:
@bep: this "works" now with v0.64.1 :)

But it seems `{{- $vendor := resources.Match "js/vendor/*.js" -}}` doesn't take into consideration our mounts. Maybe we are doing something wrong here?

Do note that this works `{{- $js := resources.Match "js/*.js" -}}` since the files are physically present.

site/layouts/partials/scripts.html
```diff
@@ -5,7 +5,7 @@
 {{- end }}
 
 {{ if (or (eq .Page.Layout "docs") (eq .Page.Layout "single")) -}}
-<script src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script src="/docs/{{ .Site.Params.docs_version }}/assets/js/vendor/docsearch.min.js"></script>
```
XhmikosR: For what is worth, this file is accessible just fine

MartijnCuppens (dismissed):

XhmikosR:
Note that I haven't added error checking yet but I think we should be fine for now.

MartijnCuppens:
LGTM, any reason not to merge?

XhmikosR:
There's no error checking, i.e. if a file is not found locally, the build will not throw an error.

XhmikosR:
Which I don't know how to fix without duplicating some stuff...

XhmikosR:
@regisphilibert do you have any suggestion about this PR? 🙂

It works quite well, but it doesn't do error checking if one of the mounts is missing.

regisphilibert:
Just looked at it, and it appears `fileExists` is not working with mounts... Is mounting the file under the `assets` directory and using Hugo Pipes with it out of the question?

XhmikosR:
I'm not sure I follow what you mean there 🙁

In this patch I mount the files in assets when we need to use Hugo pipes, otherwise in static. If mounting all of them in assets works and solves the error checking issue, we can do that of course 🙂

regisphilibert:
Well you can still put it under assets even if you don't perform any operation on it. 
The benefit is you'll be able to use `{{ with resources.Get "docs/5.0/assets/js/vendor/docsearch.min.js" }}` before using its `.RelPermalink` and warn or stop the build if the asset is not found which will probably mean the package has not been installed.

XhmikosR:
As long as we can have error checking for all files, let's do this :) Can you make a PR against this branch please?

We are getting there one thing at a time :)

regisphilibert:
Sure. 

It seems for now, you concat every javascript files under `vendor/js/` and `js/` into one bundle loaded on every page. 

But now we need another bundle for the pages of layout doc which only fetch the `js/vendor/docsearch.min.js` script which we'll mount under assets alongside the other.

Correct?

XhmikosR:
Correct. Well, not really a bundle since it's just one file, unless you mean it in the Hugo sense 🙂

regisphilibert:
I think we'll still use it as a bundle so you guys can keep the same code if any new script is needed. 

One question though, it seems the default Bundle is already being refered as `docsJS`. So how would you refer to this new bundle meant for the "docs" page :)

XhmikosR:
Maybe just docsearch? I don't think we'll ever concat docsearch with the other files because it's already huge. I should be able to have a better opinion after I see your patch.

Thanks again for your help!

XhmikosR:
Actually, now that I think about it we might create a second bundle indeed: docsearch + our search.js file and load that async. Should be better than now for sure.

XhmikosR:
@regisphilibert friendly ping :)

regisphilibert:
@XhmikosR I'm finally getting back to this... sorry for the dotted attention :(

Please shout if something changed since then and I should stay away. Thanks!

XhmikosR:
@regisphilibert no worries, everything is still the same where we left it :)

Let's go with the one bundle approach and we can tweak it later.

regisphilibert:
Awesome, @XhmikosR where can I see which Hugo version is used?

XhmikosR:
It's the latest one; we use `hugo-bin` which installs hugo and at the time of writing we use hugo-bin 0.61.0 which installs Hugo 0.73.0 https://github.com/fenneclab/hugo-bin/blob/f826b1ac5d9a9fae9464999abffe9a2856bc6fa1/package.json#L4

MartijnCuppens (approved):

PR #29292, opened 2019-08-22
Update z-indices

MartijnCuppens:
- `.carousel-caption` doesn't need a z-index. There are no non-static element in its containing block (`.carousel-item`)
- The `z-index` of `.carousel-indicators` is way higher than it needs to be (`15`), the highest sibling `z-index` is `1`, so it can be set to `2` to be the top layer.


sirius-ab (approved):
LGTM 👍

mdo (approved):

PR #29293, opened 2019-08-22
Switch percentage `font-size`s to `em` and change some font sizes for consistency

MartijnCuppens:
We use as well percentage as em based font sizes. I've changed them to `em` font sizes and tweaked some sizes for more consistency. I also noticed setting the font-size on the subnav docs was redundant.


scss/_variables.scss
```diff
@@ -1037,7 +1041,7 @@ $thumbnail-box-shadow:              $box-shadow-sm !default;
 
 // Figures
 
-$figure-caption-font-size:          90% !default;
+$figure-caption-font-size:          $small-font-size !default;
```
XhmikosR: Isn't this smaller than what it was?

MartijnCuppens: It is, but the change it minimal (from `14.4px` to `14px`). As I said, I tweaked some font sizes to make things more consistent.

site/static/docs/4.3/assets/scss/_subnav.scss
```diff
@@ -7,11 +7,6 @@
     opacity: .25;
   }
 
-  .dropdown-toggle {
```
mdo: On my iPad so I can’t confirm, but was the removal of all of this intentional? Can’t recall on the `text-decoration`.

MartijnCuppens: Yes it is. I noticed the `font-size` was redundant and the only property left in the block was `text-decoration`, which also is redundant (text-decoration is removed by `.btn`)

mdo (approved):

PR #29296, opened 2019-08-22
Simplify disabled styles

### LINKED ISSUE #27784: Bootstrap 4.1.2 `.close` hover focus styles are too specific

MartijnCuppens:
Since we can now officially use `pointer-events` since IE10 is dropped, we can simplify our code a bit and decrease our selector specificity of all active `.btn`s and `.close` button.

I've also removed the `cursor: auto;` from the links since `pointer-events: none` is applied. And I 've added `pointer-events: none` to the form ranges which doesn't require to reset the cursor styles.

Closes https://github.com/twbs/bootstrap/pull/27785
Fixes #27784

TODO:
- [x] Document the effect of `pointer-events`


mdo (commented):
Oh hell yeah, love this. Let’s document it in the migration guidelines, and possibly the docs for each documented example featuring a disabled component.

ysds:
I was thinking about this too 👍 

For compatibility, disabled should be written after active.

Before:
image

https://codepen.io/fellows3/pen/YzKQLZo

Current:
image

https://codepen.io/fellows3/pen/ExYXLgR

scss/_close.scss
```diff
@@ -12,11 +12,14 @@
...
 
-  &:not(:disabled):not(.disabled) {
-    &:hover,
-    &:focus {
-      opacity: .75;
-    }
+  &.disabled,
+  &:disabled {
+    pointer-events: none;
+  }
```
ysds: It is more readable that this block is written after the `: hover,: focus`.

scss/_buttons.scss
```diff
@@ -32,12 +30,13 @@
   // Disabled comes first so active can properly restyle
   &.disabled,
   &:disabled {
+    pointer-events: none;
     opacity: $btn-disabled-opacity;
     @include box-shadow(none);
   }
```
ysds: This block should also move after the `.active`.

MartijnCuppens:
I've tackled @ysds' comments. The colors have been updated in the meantime, that's why there's still a difference. The updated codepen of the master branch is https://codepen.io/MartijnCuppens/pen/povEOBB

XhmikosR:
Is there something else preventing us from merging this?

ysds (approved):
LGTM 👍

XhmikosR:
@patrickhlauke LGTY?

patrickhlauke (approved):

PR #29297, opened 2019-08-22
Remove "Wall of browser bugs" page.

### LINKED ISSUE #29265: Nuke Wall of browser bugs page?
It feels like we don't really maintain it, thus it will become (or even already is) obsolete, plus it's not in the sidebar menu or any other menu.

Should we remove it or keep it?

/CC @twbs/team

XhmikosR:
While important in the past, it has become obsolete, so removing it seems the best solution.

Note that the root page redirect will keep pointing to the v4.x page.

Closes #29265 

TODO:
- [x] Add notice to migration page


MartijnCuppens (approved):

mdo:
Let’s also add a note to the migration docs that this is no longer around.

XhmikosR:
Done.

mdo (approved):

PR #29298, opened 2019-08-22
dropdown show method should do the same as toggle

Johann-S:


XhmikosR:
Do we have an issue about this? This changes the behavior, not sure if it should go in 4.x...

Johann-S:
No we don't have an issue but the current behaviour is really bad, why toggle should use Popper and not show ? Toggle should use show and show should have the same behaviour as our old toggle method

XhmikosR:
Well, 1) tests don't pass and 2) it's a breaking change.

So for master this would be OK assuming the tests are fixed, but v4 I don't think this should land...

Johann-S:
The tests are just a question of time you know that...
And it's not a breaking change because it doesn't change the current behaviour and fix some accessibility problem too

Johann-S:
there is a way to backport that to v4 without a breaking change, I'll show you that 😉

XhmikosR (approved):
LGTM AFAICT for master

PR #29300, opened 2019-08-24
Allow modification of the value of rem

### LINKED ISSUE #21899: $font-size-base does not scale padding/margins in v4
If I set $font-size-base to .75rem it does not scale all padding and margins to that base font-size, making the padding and margins too large for the scale of the type. I could manually set html font-size but it seems like this scaling should be officially supported.

MartijnCuppens:
`$font-size-root` can be used to change the value of `rem`.

Closes https://github.com/twbs/bootstrap/issues/21899

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -17,6 +17,14 @@
   box-sizing: border-box;
 }
 
+// Root
+//
+// 1. Ability to the value of the root font sizes, affecting the value of `rem`.
+//    null by default, thus nothing is generated.
+
+:root {
+  font-size: $font-size-root; // 1
+}
 
 // Body
 //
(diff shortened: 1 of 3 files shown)


scss/_reboot.scss
```diff
@@ -17,6 +17,14 @@
   box-sizing: border-box;
 }
 
+// Root
+//
+// 1. Ability to the value of the root font sizes, effecting the value of `rem`.
```
XhmikosR: Affecting

mdo:
Oh whoa, this is cool. Does IE11 support this? I know it doesn’t handle CSS custom properties, but I don’t know if handles `:root`.

MartijnCuppens:
> Does IE11 support this?

Yup (https://caniuse.com/#feat=css-sel3). I can also use `html`, but `:root` felt more appropriate since the root `em`s are set.

mdo (approved):
Pushed a small docs update for this, thanks!

PR #29208, opened 2019-08-05
Docs accessibility fixes

XhmikosR:
IMO we should backport these.

@patrickhlauke: there are probably more color and general accessibility issues, but I try to solve the issues in batches. There's one thing that bugs me:

>This form does not contain a submit button, which creates issues for those who cannot submit the form using the keyboard. Submit buttons are INPUT elements with type attribute "submit" or "image", or BUTTON elements with type "submit" or omitted/invalid.

For the search form.

Preview: https://deploy-preview-29208--twbs-bootstrap.netlify.com/

PR #29315, opened 2019-08-28
Add pa11y-ci for automated accessibility tests

XhmikosR:
This is mostly a POC and it will need a lot more tweaking and some of my other PRs finished first, like #29208.

- [x] Look into ignoring some classes known to fail in `hideElements` (`*-success`, etc) `("hideElements": "[class*=-success], [class*=-info]"`)
- [ ] Maybe ignore examples from color contrast check?
- [ ] See if we can use `html` reporter (https://github.com/pa11y/pa11y-reporter-html)
- [ ] Find a way to use the HTML reporter but still log the results in CLI
- [x] See if we can drop delay-cli
- [x] Upload HTML report on CI
- [ ] See if we can use npx maybe, since Chromium is a huge download or see if we can cache puppeteer on CI


XhmikosR:
I've made some progress with this a few days ago. I added the HTML results to CI so one can download them and review them. The pa11-ci tests run only on Node.js 12 so you need to visit this action to get the results.

AFAICT we need to ignore some selectors that we know that will fail color contrast, and we can try fixing the remaining issues.

Any help is welcome. :)

ffoodd:
FWIW, pa11y has a `--wait` argument, than can be passed as an argument in CLI or in the json config file. Not tried in CI but will do soon, could allow to remove the `delay-cli` dep.

Also while working on this, I temporarily disabled contrast checking since there were too many failures (more than 15.000!) by using:

```json
"ignore": [
      "WCAG2AA.Principle1.Guideline1_4.1_4_3.G18"
    ]
```

Result is much more readable :)

Still working on it but I might have more to share soon!

XhmikosR:
I will try your suggestions when I get some time, thanks. About the contrast check, we shouldn't disable it. Instead we should ignore specific selectors. I tried doing this in the branch, but I haven't had the time to double check if it works.

ffoodd:
Indeed, that's why I mention "temporarily" :D It's only while getting pa11y in place, to be able to see changes easier.

XhmikosR:
I looked into this one more time and I think if we fix the following errors, only a few will be left:

```
Error: Not all th elements in this table have a scope attribute. These cells should contain a scope attribute to identify their association with td elements.

WCAG2AA.Principle1.Guideline1_3.1_3_1.H63.1

Error: This form does not contain a submit button, which creates issues for those who cannot submit the form using the keyboard. Submit buttons are INPUT elements with type attribute "submit" or "image", or BUTTON elements with type "submit" or omitted/invalid.

WCAG2AA.Principle3.Guideline3_2.3_2_2.H32.2

Error: This select element does not have a name available to an accessibility API. Valid names are: label element, title undefined, aria-label undefined, aria-labelledby undefined.

WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.Select.Name

Error: This form field should be labelled in some way. Use the label element (either with a "for" attribute or wrapped around the form field), or "title", "aria-label" or "aria-labelledby" attributes as appropriate.

WCAG2AA.Principle1.Guideline1_3.1_3_1.F68
```

patrickhlauke:
anywhere we can actually see the exact bits failing? i can never make sense of those CI things (and currently the pa11y task seems to be failing due to some configuration thing?)

XhmikosR:
@patrickhlauke please follow the link above (Or in the future the Node.js 12.x Action) and you will see the archive with the results.

Screenshot_2020-03-22 Add pa11y-ci for automated accessibility tests by XhmikosR · Pull Request #29315 · twbs bootstrap

patrickhlauke:
> Error: Not all th elements in this table have a scope attribute. These cells should contain a scope attribute to identify their association with td elements.
>
> WCAG2AA.Principle1.Guideline1_3.1_3_1.H63.1

that, i would argue, is a false positive. should be a warning at most, and in general isn't needed. but i guess adding `scope="col"` to those table headers won't do any harm either.

> Error: This select element does not have a name available to an accessibility API. Valid names are: label element, title undefined, aria-label undefined, aria-labelledby undefined.
>
>WCAG2AA.Principle4.Guideline4_1.4_1_2.H91.Select.Name
>
>Error: This form field should be labelled in some way. Use the label element (either with a "for" attribute or wrapped around the form field), or "title", "aria-label" or "aria-labelledby" attributes as appropriate.
>
>WCAG2AA.Principle1.Guideline1_3.1_3_1.F68

i think most of those are due to the still ongoing vagueness on whether or not `placeholder` counts as being a valid way to provide an accessible name. in some browsers, `placeholder` is used as a last resort accessible name (even when a value was chosen/entered in a form control). in some others, it's ignored/overridden as soon as a value is chosen/entered.

to be safe, all those cases where we just relied on `placeholder` to give an accessible name we should double-up with an `aria-label`.

patrickhlauke:
For all those cases where the error is "Error: This form does not contain a submit button..." i'd suggest simply removing the `<form>` and `</form>` from the code samples. they give the wrong impression that it's a complete form, when it's only really showing a handful of form elements as examples.

patrickhlauke:
btw happy to tackle the above things, but unsure if i should do it in this branch here, or a separate branch off of current master?

XhmikosR:
So, it seems we still get the FP about the duplicate ID constructor. Also, I don't see any changes when using axe vs htlm_cs.

ffoodd:
Based on latest results and @patrickhlauke's comments and PR:

1. Should we ignore the missing `legend` test completely? Or maybe add an ID to those `fieldset`s and add them to the `hideElements` whitelist in Pa11y config file, and documenting the opinionated stuff?
2. Should we skip the `#constructor` ID too — temporarily, linking to upstream issue maybe? I can't find a duplicate ID either, so I guess that's a bug… However we had the same in axe or HTML_CodeSniffer, so I'm confused about this: do they share the same HTML parser or something?
3. the color contrast issues in "Theming" bother me: I'll try to get on this, scss-side.
4. I still have errors due to `form`s missing `submit`s in Navbars example page, with this pattern: `<form><input class="form-control" type="text" placeholder="Search" aria-label="Search"></form>`, not sure how to proceed those. I guess they should match the "Navbar" component page examples, with a "Search" button aside. Let me try this.

patrickhlauke:
@ffoodd 

- yes i'd ignore the `<legend>` thing
- yes seems to be an error/bug. it IS odd that the error appears regardless of testing engine...is there perhaps something happening in the generated page which is then removed/modified AFTER pa11y runs before it becomes final output?
- not sure we can avoid having some contrast issues with theming other than completely dropping certain tints/combinations. we could exclude from testing but add a note/callout to the page pointing out that naturally some of the tints do NOT pass contrast requirements and authors should use their best judgement about which to choose/when to use them?
- as said in #30464 is, in my view, not a failure per se, should really be just a warning, and i COULD see an argument for this not being a problem for forms that only contain a single text/search input

XhmikosR:
I think our best way moving forward with pa11y-ci is to find a way to add a unique page class in body and ignore anything we think we should. I'll probably need some help with that, but I will try to tackle this in a separate PR. After this is done, we can ignore further things more easily.

ffoodd:
@XhmikosR What do you mean by "a unique page class in body"?

@patrickhlauke About contrasts, I'll just submit a PR to try defining a 4.5 threshold for our `color-contrast()` function. It could lead to good things :)

XhmikosR:
Something like this, adapted for Hugo: https://github.com/twbs/blog/blob/e687b66fee2a4ebffed09c91279bb4f8a8c51fcf/_layouts/default.html#L23-L26

patrickhlauke:
@ffoodd ah i now looked specifically at which color combos in theming fail, and yes...it's the edges where it flips from using black text to white text. so yes, changing the contrast algo to enforce 4.5:1 (what does it currently do? 3:1?) will solve it

ffoodd:
It's defined as 3:1 for now. But it could have side effects in component, would require a few checks :)

ffoodd:
Two things here:
1. I ignored the missing submit button test;
2. I effectively switched to aXe: after comparing to what I've done in Boosted, I realized that `runners` needed to be declared in `defaults`.

Aaand… We have (a lot) more failures :innocent:

ffoodd:
@XhmikosR I can apply it again, I just followed @patrickhlauke recommendation there.

XhmikosR:
Alright, let's leave it as is for now. I still think the warning is useful to have so we should ignore it on a per-case basis. And currently we can't target only the examples, unfortunately.

ffoodd:
FWIW we have a lot of headings hierarchy failures (easy to fix) and aXe seems a bit zealous about ARIA landmarks.

patrickhlauke:
> FWIW we have a lot of headings hierarchy failures (easy to fix) and aXe seems a bit zealous about ARIA landmarks.

If the failures are about skipped levels, then...it's not a hard WCAG failure (see https://developer.paciellogroup.com/blog/2020/03/heading-off-confusion-when-do-headings-fail-wcag/)

As for landmarks, historically axe/deque have been indeed quite keen to see lack of landmarks as a hard failure, but again the reality is not quite as clear cut (will need to see specifically which things are being flagged though...admittedly, i'm only commenting based on past experience with these tools)

ffoodd:
@patrickhlauke aXe is reporting multiple navigation landmarks as an error, specifying `html > body > nav`. Since they're properly labelled with `aria-label`, I don't think this should be considered as a failure…

Let us know if you have any time to browser this new report. I'll try to tackle a few things when I can, too.

patrickhlauke:
axe is correct that we do have two `<nav>` elements with the *same* `aria-label="Secondary navigation">` (the search bar at the top, and the TOC)

screenshot from devtools showing the two navs

patrickhlauke:
"Error: Heading levels should only increase by one" as note that's not actually a hard failure. should be a warning at best. we should ignore/exclude this particular test

patrickhlauke:
"Error: All page content must be contained by landmarks"

that's best practice, not a hard failure (in fact, looking at https://dequeuniversity.com/rules/axe/3.5/region?application=axeAPI and also the above heading thing https://dequeuniversity.com/rules/axe/3.5/heading-order?application=axeAPI axe does call this WCAG Best practice rather than noting them as specific SC failures ... but it seems pa11y doesn't differentiate here and calls them errors regardless - wonder if there's some way to easily say "just error on hard failures, not best practices")

patrickhlauke:
ditto "Error: Aside must not be contained in another landmark" - best practice, not failure

build/.pa11yci.json
```diff
@@ -0,0 +1,11 @@
+{
+  "standard": "WCAG2AA",
+  "level": "error",
+  "concurrency": 4,
+  "defaults": {
+    "runners": [
+      "axe"
+    ],
```
patrickhlauke: worth adding `"ignore": [ "notice", "warning" ]` here and seeing if that stops the various "best practice" issues we're now seeing flagged up by axe

XhmikosR:
All right, I cleaned up the branch.

Notes:

1. I switched back to htmlcs; it's faster and doesn't have the landmarks FP
2. I haven't split the accessibility to a separate Action because it needs the doc built files
3. We could go further and run pa11ci for js/tests/visual; we already run these through the HTML validator, but maybe it's an overkill
4. I don't love that the results are not logged in CLI. I still need to see if I can find a way to have both
5. Finally, there are still some valid failures even with our current ignores, so perhaps we shouldn't land this until we sort these out in master. Not a lot, though, so that's something regarding v5.0 :)

patrickhlauke:
i have a tiny bit of time this week (as I'm in-between jobs with a few days off) ... so shout if you need me to look at anything here (and maybe point me to what it is I should look at...I always lose track of where we're up to / where the logs are / etc)

XhmikosR:
@patrickhlauke the results are uploaded in the Docs Action. https://github.com/twbs/bootstrap/actions?query=workflow%3ADocs+branch%3Amaster-xmr-pa11y-ci Then you visit the failed action and you get the zip with the HTML results.

patrickhlauke:
just did a run-through of the pa11y results. all the color contrast issues are legit, and gel with some of the problems i found manually last week while going over some of the v5 pages.

beyond that, i'd still say the following two rules can be ignored:

- WCAG2AA.Principle3.Guideline3_2.3_2_2.H32.2 (forms without an explicit submit button)
- WCAG2AA.Principle1.Guideline1_3.1_3_1.H71.NoLegend (fieldsets without a legend)

plus that one failure that appears to be a false positive about the duplicate `id="constructor"` which i just can't find when manually verifying the page

mdo:
What's the status on this one? Still ongoing or closable?

Also, lol-ing at the dates here.

Screen Shot 2020-09-16 at 9 51 05 PM

ffoodd:
It's been a long walk, you know 😄

We'll need to rebase and check report. FYI I manages to make it green and required on Boosted so I really think we should keep going.

XhmikosR:
From my side these are the things to think about:

1. aXe or html_codesniffer; last time I tried aXe, we were flooded with issues that might have been FPs
2. I don't want the huge puppeteer (which in turn downloads Chromium) devDependency, ~143MB... Not sure how to work around this, we could probably move the needed packages in the Action that needs them, but it's not a perfect solution either
3. I can't seem to be able to find a way to display the CLI results and also redirect them at the same time

patrickhlauke:
Is this PR still worth pursuing at this point, or do we need a fresh take on it?

XhmikosR:
There's no need for a fresh take in it, I can rebase it. We need to solve the TODO I have in the OP.

PR #29320, opened 2019-08-28
 Fix Modal compensation padding

### LINKED ISSUE #28101: Padding problem when there is a forced scrollbar
There is an issue with the way modals add margins to account for the scroll bar. This issue occurs in the most recent releases of both Bootstrap 3 and 4.

When a browser is forced to display the scrollbar when the vertical content does not require the scrollbar, a padding-right is added to the body when it is not necessary.

Two jsfiddle examples are below that show the problem.

Bootstrap 4:
https://jsfiddle.net/ziluckmichael/fh71sgo0/

Bootstrap 3:
https://jsfiddle.net/ziluckmichael/nvz0ybux/

As you can see it adds padding-right to the browser when it is not required. Both examples use the default code almost exactly as it is off of the demo page.

Browsers Always Affected:
- Microsoft Edge 40.15063.674.0
- Google Chrome 71.0.3578.98 (Official Build) (64-bit)
- Firefox 64.0.2

Browsers Sometimes Affected:
- Internet Explorer 11.1508.15063.0

For IE 11, the issue does not occur on my test/demo of the problem, but it does occur on the actual production website where I discovered the issue originally. It appears this is because IE overlays the scrollbar and has it slowly disappear, while every other browser has it permanently on the screen.

cjweigle:
Modals prevent shifting horizontally by compensating for the vertical scrollbar with padding. There are several issues in the detection logic for detecting if a vertical scrollbar is present.

Fixes #28101

## CHANGES ##
* When checking if the body is overflowing, use the scrollHeight of `<html>`, unless the body has overflow-y set to hidden. In that case, use the bottom of the boundary rect.
* When showing the modal, call _adjustDialog after display has been set to 'block'
* In _adjustDialog, if compensation padding is no longer required due to window resize, remove previous compensation padding.

###### 1 ######
Currently _checkScrollbar checks for horizontal overflow to detect if the vertical scrollbar is present. The behavior was changed in commit 32f62bc565882b989fcef62874c0a90b0241dd1e to work around an IE quirk. To check for the presence of a vertical scrollbar, the content height must be compared to the window height, not the widths.

###### 2 ######
Currently when setting _isBodyOverflowing, the width of the body including offset is calculated by adding the left position and the right position of the boundary box. This includes the left
position twice as the right position includes the left position. The correct comparison checks only the right position. This comparison (modified to check vertical dimensions) is used in
the patch when checking if the body is overflowing and the body has overflow-y set to hidden.

This behavior can be observed in the test:
_"should not add compensation padding when body boundary width (including negative margin-left) is greater than window width with no vertical scrollbar before showing and no vertical scrollbar after showing"_

###### 3 ######
Currently when setting _isBodyOverflowing, the `<body>`'s boundary box is checked. The boundary box does not include overflowing descendant nodes. Checking `<html>`'s scrollHeight
includes the height of the content as well as margin on the `<body>`.

This behavior can be observed in the test:
_"should add compensation padding when body boundary height is less than or equal to window height with vertical scrollbar before showing due to tall content and no vertical scrollbar after showing"_

#### Initial _adjustDialog call ####

Currently _adjustDialog is called in the show() method before display is set to 'block'. Therefore the modals scrollHeight is always 0 when initially showing, which results in an incorrect value
for isModalOverflowing in _adjustDialog. When _adjustDialog is called due to a resize event the scrollHeight is available as the modal is displayed. As such, _adjustDialog has different effects
for similar conditions depending on whether it's the initial call or a resize call.

This behavior can be observed in the test:
_"should have consistent compensation padding when initially showing and after a resize event"_

#### _adjustDialog ####

Currently _adjustDialog adds compensation padding as needed. The compensation padding is removed when the modal is hidden. _adjustDialog never removes the compensation padding, even when it is no longer required.

This behavior can be observed in the test:
_"should remove compensation padding no longer needed"_  

###### 1 ######
In the tests related to Modal compensation padding, the Bootstrap css distribution file is loaded. I have confirmed that different test outcomes occur when loading the css file and
when not.

The synthetic toolbar used in the Modal tests does not accurately emulate the behavior of a browser scrollbar by showing and hiding as needed.

In _adjustDialog, the expression: `!this._isBodyOverflowing && isModalOverflowing` can be thought of as "no vertical scrollbar before showing and vertical scrollbar after showing".

The expression: `this._isBodyOverflowing && !isModalOverflowing` can be thought of as "vertical scrollbar before showing and no vertical scrollbar after showing".

The tests depend on the state transition of the scrollbar's visibility which is not emulated by the synthetic scrollbar.

###### 2 ######
The following tests required a 'tallContent' element to be added to correctly test their functionality now that compensation padding is correctly working:
_"should toggle a modal"
"should adjust the inline padding of fixed elements when opening and restore when closing"
"should adjust the inline margin of sticky elements when opening and restore when closing"
"should properly restore non-pixel inline body padding after closing"_

#### Current ####
current

#### Patched ####
patched

Preview: <https://deploy-preview-29320--twbs-bootstrap.netlify.app/docs/5.0/components/modal/>


js/src/modal/modal.js
```diff
@@ -432,8 +439,20 @@ class Modal {
   }
 
   _checkScrollbar() {
-    const rect = document.body.getBoundingClientRect()
-    this._isBodyOverflowing = rect.left + rect.right < window.innerWidth
+    let contentHeight = null
+
+    if (!bodyStyle) {
+      bodyStyle = window.getComputedStyle(document.body)
```
Johann-S: bodyStyle can change over the time by doing that you freeze it, so you should remove the global var and query the body style each time it's needed

js/src/modal/modal.js
```diff
@@ -432,8 +439,20 @@ class Modal {
...
+    }
+
+    if (bodyStyle.overflowY === 'hidden') {
+      const bbox = document.documentElement.getBoundingClientRect()
```
Johann-S: `const { bottom } = document.documentElement.getBoundingClientRect()`

js/tests/karma.conf.js
```diff
@@ -106,7 +106,8 @@ if (browserStack) {
   conf.browsers = browsersKeys
   reporters.push('BrowserStack', 'kjhtml')
   files = files.concat([
-    { pattern: 'js/src/**/*.spec.js', watched: false }
+    { pattern: 'js/src/**/*.spec.js', watched: false },
+    { pattern: 'dist/css/bootstrap.css', watched: false, included: false, served: true }
```
Johann-S: please do not include the whole Bootstrap css, just had what's needed to be include

js/src/modal/modal.spec.js
```diff
@@ -66,6 +75,11 @@ describe('Modal', () => {
       const originalPadding = '0px'
 
       document.body.style.paddingRight = originalPadding
+      const tallContent = document.createElement('div')
```
Johann-S: can you make a local method to create `tallContent` ?

js/src/modal/modal.spec.js
```diff
@@ -602,6 +632,1493 @@ describe('Modal', () => {
...
+      document.documentElement.style.paddingRight = null
+      fixtureEl.style.position = 'relative'
+      fixtureEl.style.left = null
+      fixtureEl.style.right = null
+      fixtureEl.style.top = null
+      fixtureEl.style.bottom = null
+      fixtureEl.style.width = null
+      fixtureEl.style.height = null
+      dist.addEventListener('load', () => {
+        fixtureEl.innerHTML = '<div class="modal"><div class="modal-dialog" ><div class="modal-content">\n' +
```
Johann-S: can you follow how we build our fixtures? Using an array and call `.join`

js/src/modal/modal.spec.js
```diff
@@ -602,6 +632,1493 @@ describe('Modal', () => {
 
       modal.show()
     })
+
+    it('should not add compensation padding when body boundary width is less than window width with no vertical scrollbar before showing and no vertical scrollbar after showing', done => {
+      const dist = document.createElement('link')
```
Johann-S: I prefer if you can create a local method which inject just the CSS needed not the whole CSS please

js/src/modal/modal.spec.js
```diff
@@ -602,6 +632,1493 @@ describe('Modal', () => {
...
+      dist.rel = 'stylesheet'
+      dist.id = 'bootstrap-dist'
+      dist.href = '/base/dist/css/bootstrap.css'
+      document.head.appendChild(dist)
+
+      document.body.style.width = window.innerWidth - 1 + 'px'
+      document.body.style.height = '1px'
+      document.documentElement.style.paddingRight = null
+      fixtureEl.style.position = 'relative'
```
Johann-S: If you change `fixtureEl` style please applied the default style after each of your unit tests

Johann-S (changes requested):
Hi @cjweigle,

Thanks for tackling this issue 👍 

You'll find some changes to do and after that, I'll have to dig a bit more in your changes

js/src/modal/modal.js
Johann-S: it's a possible case, but user can change body styles too

js/src/modal/modal.spec.js
```diff
@@ -38,6 +222,16 @@ describe('Modal', () => {
     })
 
     document.body.style.paddingRight = '0px'
+
```
Johann-S: You should remove the modal CSS added and when you made changes on `document.body` style

js/src/modal/modal.spec.js
```diff
@@ -3,7 +3,191 @@ import EventHandler from '../dom/event-handler'
 import { makeArray } from '../util/index'
 
 /** Test helpers */
-import { getFixture, clearFixture, createEvent, jQueryMock } from '../../tests/helpers/fixture'
+import { getFixture, clearFixture, unstyleFixture, createEvent, jQueryMock } from '../../tests/helpers/fixture'
+
+const getModalCss = () => {
+  return [
```
Johann-S: it's a lot of CSS, is it really needed? If it's just to emulate the overflow for me it's a bit too much

js/src/modal/modal.spec.js
```diff
@@ -66,6 +260,7 @@ describe('Modal', () => {
       const originalPadding = '0px'
 
       document.body.style.paddingRight = originalPadding
+      document.body.appendChild(createTallContent())
```
Johann-S: `should toggle a modal` test is created like that because most of the times a modal doesn't have a tall content.
What's the reason to add a tall content here?

js/src/modal/modal.spec.js
```diff
@@ -602,6 +800,751 @@ describe('Modal', () => {
...
+      document.body.style.height = '1px'
+      document.documentElement.style.paddingRight = null
+      unstyleFixture()
+
+      fixtureEl.innerHTML = getModalInnerHtml()
+
+      const modalEl = fixtureEl.querySelector('.modal')
+      const modal = new Modal(modalEl)
+
+      modalEl.addEventListener('show.bs.modal', e => {
```
Johann-S: what's the interest of checking if the `e` is defined? Is it something you want to test?

js/src/modal/modal.spec.js
```diff
@@ -602,6 +800,751 @@ describe('Modal', () => {
...
+      document.body.style.marginLeft = '-100px'
+      document.documentElement.style.paddingRight = null
+      unstyleFixture()
+
+      fixtureEl.innerHTML = getModalInnerHtml()
+
+      const modalEl = fixtureEl.querySelector('.modal')
+      const modal = new Modal(modalEl)
+
+      modalEl.addEventListener('show.bs.modal', e => {
```
Johann-S: Ok it seems this `show` event handler is something you copy/pasted please remove it when it's not needed

Johann-S (changes requested):
Thanks again you improved things a lot 👍 

Few things to change, but we move in the good direction  👍

js/src/modal/modal.spec.js
cjweigle: No it's not all needed, I have it narrowed down.

js/src/modal/modal.spec.js
cjweigle: The modal css is being removed below on line 233

js/src/modal/modal.spec.js
cjweigle: Within `should toggle a modal` it expects the original body padding to be stored in the data-padding-right attribute, which is only set when the body is overflowing. `should toggle a modal` can't both check the case that a modal doesn't have tall content, and also check that the original body padding is preserved.

js/src/modal/modal.spec.js
cjweigle: That's correct

js/src/modal/modal.spec.js
Johann-S: Ok make sense to me 👍

js/src/modal/modal.spec.js
cjweigle: Body style is reset on line 216

js/src/modal/modal.spec.js
Johann-S: Thanks that's better! 👍

js/src/modal/modal.spec.js
Johann-S: image

I don't see anything which clears the body style 😟

Johann-S: Ok I saw it thanks 👍

Johann-S:
Ok in terms of code everything is fine now; thanks for your hard work 👍 
Now I'll focus on if it works fine (at a first sight it works fine 👍)

XhmikosR:
This needs a rebase @Johann-S

js/src/modal.js
```diff
@@ -248,12 +246,15 @@ class Modal {
     const modalBody = SelectorEngine.findOne(Selector.MODAL_BODY, this._dialog)
 
     if (!this._element.parentNode ||
-        this._element.parentNode.nodeType !== Node.ELEMENT_NODE) {
+      this._element.parentNode.nodeType !== Node.ELEMENT_NODE) {
```
XhmikosR: Please revert the space changes in this file

js/tests/unit/modal.spec.js
```diff
@@ -117,7 +184,7 @@ describe('Modal', () => {
         '<div class="sticky-top" style="margin-right: 0px;"></div>',
         '<div class="modal"><div class="modal-dialog" /></div>'
       ].join('')
-
+      document.body.appendChild(createTallContent())
```
XhmikosR: Needs a newline before

XhmikosR (changes requested):
Please also rebase and squash your patches into one

XhmikosR:
@cjweigle can you rebase this please? Not merge, rebase it.

XhmikosR:
Actually, scratch that. If it's OK, we can just squash it when merging.

XhmikosR:
Unfortunately, since I haven't received any feedback, I'm gonna have to remove this PR from 4.4.0.

I want us to release v4.4.0 by next week top.

cjweigle:
Patches squashed

mdo:
@Johann-S @XhmikosR Thoughts on this one? Tempted to slate for alpha 2 if there's still interest here.

cjweigle:
Updated with master

XhmikosR:
@cjweigle thanks but unfortunately that's not a rebase; you have upstream commits while you should have only your commits. :/

XhmikosR:
Rebased the branch. There was one conflict with main; the `Math.round` addition.

mdo:
@twbs/js-review Is this still worth keeping around and shipping? It's out of date and has conflicts.

XhmikosR:
The issue is still open, so we need to fix it in both branches.

mike-koala-bear (commented):
good job

mdo:
Closing as stale.

PR #29322, opened 2019-08-29
v5: .form-check layout changes

### LINKED ISSUE #28242: custom-switch on the right side of the label
I would like the switch on the right side of the label.
Also have a spacing between them label and the switch so the label is on the left and the switch is on the right edge.

### LINKED ISSUE #28524: Official checkbox in list or menu item
Is there an official/sanctioned/blessed/Bootstrap best practices solution for having checkboxes (natural or custom) in a menu like list of items, wrapped in a collapsible drop down (see screen shot)? Clicking anywhere on the item toggles the checkbox. There are various implementations around but all seem a bit kludgy and bloated - including mine that stuffs everything in the checkbox label to preserve their click relationship. 

[https://jsfiddle.net/dragontheory/78aj5dhr/](https://jsfiddle.net/dragontheory/78aj5dhr/)

Is it possible to create a **new feature** that demonstrates best practices for implementation? Any guidance would be appreciated. Thank you!

BTW, your hard work has improved UX for millions of users. No pressure! Doing great!

checkbox_radio_btn_example_01

### LINKED ISSUE #28603: custom checkboxes inside input groups
You think of allowing custom checkboxes inside input groups, this example is bugged when using `form-control-sm`

https://codepen.io/claudioalmeiida/pen/XQmRYE

image

mdo:
Screen Shot 2019-08-28 at 9 08 36 PM

This PR does a few things:

- Scopes `.form-check-input` layout properties to the `.form-check` parent, allowing easy placement of the `.form-check-input` in more places.
- Updates the no-label checks. Current version recommends static positioning, but we use floats, so this is kinda busted right now.
- Updates the checkbox/radio in input group example to use the v5 style `.form-check-input`.
- Documents how to use `.form-check` within list groups. This also includes the addition of `margin-bottom: 0` on list group items so you can use the `<label>` element.

Fixes #28524, closes #28603 (though this is outdated by v5's combined `.form-check`). Also closes #28242.

Preview: https://deploy-preview-29322--twbs-bootstrap.netlify.com/


XhmikosR:
It seems this isn't allowed:

>Element "label" not allowed as child of element "ul" in this context

ysds (approved):
wow!

MartijnCuppens:
The first example without the text being clickable feels weird, not sure when those would be useful?

mdo:
> The first example without the text being clickable feels weird, not sure when those would be useful?

It leaves the affordance for when the text is hyperlinked or another tap target.

MartijnCuppens:
Links in labels are valid and the checkbox won't be checked if the link is clicked. Not sure what other tap targets could be present in the list group? inputs? I think it's quite uncommon to have those in combination with checkboxes inside a list group?

mdo:
Since it's only an example in the docs, I'm going to leave it for now as a demonstration that `<label>`s aren't required as parent elements. We can definitely do something else, as I could see folks wanting to do a checkbox, heading as a label, and some additional elements as supporting text. In desktop cases, this can easily be multiple tap targets, but less so on mobile/tablet.

CaninoDev:
Is [# 28242.](https://github.com/twbs/bootstrap/issues/28242) no longer considered to being incorporated? How was it resolved?

Issue #28304, opened 2019-02-19
Look if we need to add a fallback for `word-break`


XhmikosR:
Currently we are using `word-break: break-word;` in various places like `code`.

This isn't a standard property, at least not yet.

See if we need to provide a fallback.

/CC @mdo @MartijnCuppens

bardiharborow:
See #28381 for a possible solution.

XhmikosR:
It's not a browser bug... this isn't a standard yet.

bardiharborow:
Browser support, is I guess what I meant. I don't know if we should have a label for that or not, but that's another issue.

XhmikosR:
We don't, thus how I actually noticed the issue. Using vnu.jar's web interface, which can also lint CSS, I saw the `break-word` errors and then looked into it.

IMO we shouldn't use non-standard properties when possible, and when we do we need to make sure we have fallbacks.

Because something works in Chorme/Chromium, certainly doesn't mean is standard, and we shouldn't endorse this.

XhmikosR:
/CC @twbs/css-review, not sure if this is still applicable.

PR #29325, opened 2019-08-29
Use word-wrap in .text-break for IE and Edge compatibility.

### LINKED ISSUE #29319: Docs - Word break
I was trying to apply the `text-break` class in order to wrap text. Everything was working fine in Chrome. But once I moved over to IE and Edge, things were all over the place. I then visited the docs site and saw that even the docs site doesn't exhibit the intended behavior.

image

**Mandatory Information:**
Windows 10 Version 1709 / 16299.1331
Internet Explorer v. 11.1331.16299.0 / 11.0.140 (KB4511872)
Microsoft Edge 41.16299.1004.0 / Microsoft EdgeHTML 16.16299

florianlacreuse:
Fix #29319

`word-break` property is mostly use for East Asian languages, for soft wrap opportunities between letters commonly associated with languages like Chinese, Japanese, etc.

When the `overflow-wrap` property is not working (IE, Edge), we need to use the old `word-wrap` property as a fallback.

Note: If accepted and merged, this fix should be backported to v4.

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -434,7 +434,7 @@ $utilities: map-merge(
       values: italic normal
     ),
     "overflow-wrap": (
-      property: overflow-wrap word-break, // word-break for IE & < Edge 18
+      property: overflow-wrap word-wrap, // word-wrap for IE & < Edge 18
       class: text,
       values: (break: break-word)
     ),
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -433,8 +433,8 @@ $utilities: map-merge(
       class: font,
       values: italic normal
     ),
-    "overflow-wrap": (
-      property: overflow-wrap word-wrap, // word-wrap for IE & < Edge 18
+    "word-wrap": (
+      property: word-wrap,
       class: text,
       values: (break: break-word)
     ),
--- a/scss/mixins/_reset-text.scss
+++ b/scss/mixins/_reset-text.scss
@@ -1,6 +1,6 @@
 @mixin reset-text {
   font-family: $font-family-base;
-  // We deliberately do NOT reset font-size or word-wrap.
+  // We deliberately do NOT reset font-size or overflow-wrap / word-wrap.
   font-style: normal;
   font-weight: $font-weight-normal;
   line-height: $line-height-base;
--- a/site/content/docs/4.3/utilities/text.md
+++ b/site/content/docs/4.3/utilities/text.md
@@ -45,7 +45,7 @@ Prevent text from wrapping with a `.text-nowrap` class.
 
 ## Word break
 
-Prevent long strings of text from breaking your components' layout by using `.text-break` to set `overflow-wrap: break-word` (and `word-wrap: break-word` for IE & Edge compatibility).
+Prevent long strings of text from breaking your components' layout by using `.text-break` to set `word-wrap: break-word`. We use `word-wrap` instead of the more common `overflow-wrap` for wider browser support.
 
 {{< example >}}
 <p class="text-break">mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm</p>
--- a/site/content/docs/4.3/utilities/text.md
+++ b/site/content/docs/4.3/utilities/text.md
@@ -45,7 +45,7 @@ Prevent text from wrapping with a `.text-nowrap` class.
 
 ## Word break
 
-Prevent long strings of text from breaking your components' layout by using `.text-break` to set `overflow-wrap: break-word` (and `word-break: break-word` for IE & Edge compatibility).
+Prevent long strings of text from breaking your components' layout by using `.text-break` to set `overflow-wrap: break-word` (and `word-wrap: break-word` for IE & Edge compatibility).
 
 {{< example >}}
 <p class="text-break">mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm</p>


XhmikosR:
This won't apply clean on v4-dev, we'll need to manually backport it.

Also, note that I have #28304 for months now because of this. We should deal with it once and for all.

florianlacreuse:
I agree about the backport to `v4-dev`, I could do it in another PR if needed.

Good point about other components using property to break words. I've gone through the scss files in v5:

* `word-wrap: break-word;` is used in: `code`, `.card`, `.popover` and `.tooltip`.
* `word-break: normal;` is used in: `reset-text` mixin, `pre code`

`word-break: break-all` is only used in `.text-break` (v5). It seems that #28304 is no longer applicable.

It looks like you guys decide to use `word-wrap: break-word` and not `overflow-wrap`. Should I remove `overflow-wrap` and `word-break` from `.text-break`?

MartijnCuppens:
https://github.com/twbs/bootstrap/blob/7c8a1256238be5a5236aab4c47661e10dbb56c11/scss/_utilities.scss#L426

Do we need 3 properties here? Could you explain why?

We should do some crossbrowser testing to tackle this, I know IE, FF and chrome behave different on each of these properties. @florianlacreuse, could you make a demo with the effects?

florianlacreuse:
First of all, with this PR I had the idea to fix the `overflow-wrap` issue with IE / Edge without dropping properties or breaking things.

About the 3 properties, it really depends. I read several articles or messages, some recommend to use `overflow-wrap` and `word-wrap`, some only use `word-wrap`. Same with `word-break`, some use it, some don't.

AFAIK, `overflow-wrap` and `word-wrap` are the same, `overflow-wrap` is the new name and it's the new standard in CSS specs. Unfortunately, some browsers (IE and old Edge) require the legacy name `word-wrap` to work (they don't know `overflow-wrap`).

That's why you can see many times `overflow-wrap` and `word-wrap` used at the same time, people use `overflow-wrap` as new standard and also put `word-wrap` as fallback. That was the original goal of this PR.

Currently we could only use `word-wrap` and the result will be the same (unless some browsers don't support `word-wrap` but do support `overflow-wrap`? - sounds unlikely). As I said in my previous message, Bootstrap v5 only use `word-wrap` in several components, we can do the same with `.text-break`, i'm fine with it.

About `word-break` property: [this page from mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) explains the property with some examples. 1. `break-word` value is deprecated. 2. `break-all` value is used for CJK text, but I don't know more about it. In Bootstrap v5, `word-break` is only used in `.text-break` with deprecated `break-word` value. IMHO we should drop it.

@MartijnCuppens https://codepen.io/florianlacreuse/pen/BaBZMdJ

florianlacreuse:
@MartijnCuppens I know you guys are busy but it would be nice if we could chose what to do here. I still think we should drop `word-break` property, and keep `overflow-wrap` + `word-wrap` (or only `word-wrap` for consistency with other BS components / classes).

MartijnCuppens:
@florianlacreuse, I'll have a look at this somewhere this week. Thanks for (re)pinging me here, this is indeed something which should be tackled.

MartijnCuppens:
Solid research you did there, @florianlacreuse. Imo we should:
- Drop `word-break`, since this doesn't really add anything
- Also drop `overflow-wrap`. Although it is the standardised property of `word-wrap`, there's no difference if have it or not and it's very unlikely browsers will suddenly drop support for `word-wrap`.

@florianlacreuse, could you change this in your PR and maybe make a separate PR for `v4-dev`?

XhmikosR:
We should use whatever the standard is IMO. Later they might follow it.

MartijnCuppens:
> We should use whatever the standard is IMO. Later they might follow it.

I agree in most cases, but in this case the property is just completely redundant.

florianlacreuse:
For now, we agree that we should remove `word-break` property, that's good!

`overflow-wrap` browsers support : https://caniuse.com/#feat=wordwrap
`word-wrap` browsers support: https://caniuse.com/#feat=mdn-css_properties_word-wrap

@XhmikosR `overflow-wrap` is indeed the new standard but it doesn't have a full support in IE11 and Edge < 18. IMHO we should use `overflow-wrap` and `word-wrap` (as a fallback) or `word-wrap`, no strong opinion about these two choices.

XhmikosR:
Like I said, we should make sure we support the standard so that we are future-proof. IE is a dead browser and it will just use the workaround fine. The other evergreen browsers don't fall into the same category.

florianlacreuse:
@MartijnCuppens Thank you for your feedback! PR has been updated to remove `word-break`.

The other issue is about other BS components / css classes using only `word-wrap`, like:

https://github.com/twbs/bootstrap/blob/4e59997d066b26839c20bf4e54889aaadedc12e5/scss/_card.scss#L5-L11
https://github.com/twbs/bootstrap/blob/4e59997d066b26839c20bf4e54889aaadedc12e5/scss/_popover.scss#L1-L13
https://github.com/twbs/bootstrap/blob/4e59997d066b26839c20bf4e54889aaadedc12e5/scss/_reboot.scss#L281-L284
https://github.com/twbs/bootstrap/blob/4e59997d066b26839c20bf4e54889aaadedc12e5/scss/_reboot.scss#L414-L415
https://github.com/twbs/bootstrap/blob/4e59997d066b26839c20bf4e54889aaadedc12e5/scss/_tooltip.scss#L2-L12

For consistency, we should also use `overflow-wrap`, no?

MartijnCuppens:
@XhmikosR, from the w3c specification (https://www.w3.org/TR/css-text-3/#overflow-wrap-property):
> For legacy reasons, UAs must treat `word-wrap` as an legacy name alias of the `overflow-wrap` property.

So it's officially more an alias for `overflow-wrap`.

MartijnCuppens (dismissed):

XhmikosR:
@MartijnCuppens feel free to merge and backport it to a v4-dev branch

MartijnCuppens (approved):

MartijnCuppens:
Discussed this on Slack with @XhmikosR, we agreed on removing `overflow-wrap` since `word-wrap` is part of the specs, covers everything `overflow-wrap` covers and it's not worth adding it to all components @florianlacreuse listed in https://github.com/twbs/bootstrap/pull/29325#issuecomment-565495863.

florianlacreuse:
Using only `word-wrap` is fine too, everything is consistent so I'm good. Thanks for your time!

PR #29326, opened 2019-08-29
theming.md: automate _root.scss.

### LINKED ISSUE #29324: Automate _root:scss in theming page
It'll probably need some string handling and reading the generated CSS with Hugo. Not sure if it's worth it, but we should look into it.

Refs #29323

XhmikosR:
TODO: fix font variables.

Any ideas welcome.

Fixes #29324 

Preview: <https://deploy-preview-29326--twbs-bootstrap.netlify.com/docs/4.3/getting-started/theming/#available-variables>

### DIFF
--- a/site/content/docs/4.3/getting-started/theming.md
+++ b/site/content/docs/4.3/getting-started/theming.md
@@ -370,6 +370,10 @@ Here are the variables we include (note that the `:root` is required). They're l
 
 {{< highlight css >}}
 {{< root.inline >}}
+{{- $variables := readFile "scss/_variables.scss" -}}
+{{- $fontSansSerif := findRE "(?:\\$font-family-sans-serif:\\s+)(.*)(?: !default)" $variables -}}
+{{- $fontMonospace := findRE "(?:\\$font-family-monospace:\\s+)(.*)(?: !default)" $variables -}}
+
 :root {
   {{- range $.Site.Data.colors }}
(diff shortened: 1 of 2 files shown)


site/content/docs/4.3/getting-started/theming.md
```diff
@@ -370,15 +370,19 @@ Here are the variables we include (note that the `:root` is required). They're l
 
 {{< highlight css >}}
 {{< root.inline >}}
+{{- $variables := readFile "scss/_variables.scss" -}}
+{{- $fontSansSerif := findRE "(\\$font-family-sans-serif:\\s+)(.*)(!default)" $variables -}}
```
XhmikosR: @bep: do you have any suggestions how one can get a capturing group from `findRE`?

XhmikosR:
OK, this works.

Not sure about the performance penalty, but I can't find another way to automate this...

MartijnCuppens:
> Not sure about the performance penalty

`:root` is defined quite on top of the file, so it should be fine I guess

MartijnCuppens (approved):

XhmikosR:
>`:root` is defined quite on top of the file, so it should be fine I guess

We still have to read the whole file and parse it.

MartijnCuppens:
> We still have to read the whole file and parse it.

Yup, I know, worth it imo

PR #29327, opened 2019-08-29
docs: remove swatches for theme colors.

XhmikosR:
Use the built-in utilities instead.


MartijnCuppens:
The `.text-dark` utility is probably moot since it can inherit the color.

XhmikosR:
This snippet is used in many places so I went for consistency.

MartijnCuppens (approved):

PR #29329, opened 2019-08-29
Merge lint scripts

XhmikosR:
It seems ESLint is faster for me this way, stylelint is slower though.

Note that you should remove the `.cache` directory prior to making any tests.

/CC @twbs/team


MartijnCuppens:
I'm ok with this. Personally I would prefer `\"**/*.{css,scss}\"` (not sure if the curly brackets need to be escaped) instead of `\"**/*.scss\" \"**/*.css\"`

XhmikosR:
Merge those, good call.

MartijnCuppens (approved):

XhmikosR:
@Johann-S can you try this too?

Johann-S:
It works well for me too 👍

PR #29330, opened 2019-08-29
add helpers to add, remove and contains classes

Johann-S:
Preview: <https://deploy-preview-29330--twbs-bootstrap.netlify.app/>


js/src/dom/class-list.js
```diff
@@ -0,0 +1,27 @@
+function addRemoveClass(add, element, className, classNameList) {
+  if (!classNameList.length) {
+    element.classList[add ? 'add' : 'remove'](className)
+
+    return
+  }
+
+  const classList = [className].concat(classNameList)
+
+  classList.forEach(addClassName => element.classList[add ? 'add' : 'remove'](addClassName))
```
mvanroon, line 10: could this be rewritten as

`classList.forEach(item => element.classList.toggle(item, add))` ?

js/src/dom/class-list.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function addClass(element, className, ...classNameList) {
```
mvanroon, line 13: maybe I'm missing something here, but why separate the first className parameter from the spread?
won't `addClass(element, ...classNameList)` work? It will save you having the if statement on line 2

mvanroon (approved):
The class-list.js util could be groomed a bit, but otherwise 👍

XhmikosR:
@Johann-S could you try rebasing the branch so that we see where we are at?

js/src/dom/class-list.js
Johann-S: I prefer not to do that because if someone call `removeClass` on an element which doesn't possess the class it'll add it

js/src/dom/class-list.js
Johann-S: nice catch thanks! 👍

Johann-S:
Done @XhmikosR 👍

XhmikosR:
So, I thought about this and I'm in between. I'm not sure if it's worth it or not.

@rohit2sharma95 what are your thoughts?

rohit2sharma95:
But these all methods are available natively in the classList property of the DOM element 🤔

XhmikosR:
@Johann-S are you OK with dropping this patch?

Johann-S:
yep no problem 👌

PR #29339, opened 2019-08-30
v5: Simplify navbars by requiring containers

mdo:
Closes #25274 by @zalog.

Unsure if we want to go with this, but I believe @MartijnCuppens suggested it somewhere. It does simplify behavior of the navbar a bit, and it ensures horizontal alignment from @zalog's original reported issue.

Thoughts @twbs/css-review?

TODO:

- [x] Mention the change in migration page

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -833,7 +833,7 @@ $nav-pills-link-active-bg:          $component-active-bg !default;
 // Navbar
 
 $navbar-padding-y:                  $spacer / 2 !default;
-$navbar-padding-x:                  $spacer !default;
+$navbar-padding-x:                  null !default;
 
 $navbar-nav-link-padding-x:         .5rem !default;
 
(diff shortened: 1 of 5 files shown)


scss/_variables.scss
```diff
@@ -114,8 +114,8 @@ $escaped-characters: (
 
 $enable-caret:                                true !default;
 $enable-rounded:                              true !default;
-$enable-shadows:                              false !default;
-$enable-gradients:                            false !default;
+$enable-shadows:                              true !default;
+$enable-gradients:                            true !default;
```
MartijnCuppens: Could you revert this?

MartijnCuppens:
I like this idea, was thinking about doing this myself one day

XhmikosR:
Rebased and addressed @MartijnCuppens's comment.

XhmikosR:
Shouldn't we mention this in migration?

MartijnCuppens:
> Shouldn't we mention this in migration?

We should

MartijnCuppens (approved):

zalog:
Love it! Great work guys

PR #29348, opened 2019-08-31
v5: Update colors to add shades and tints

mdo:
Fresh pass at this to replace #28268. This PR is entirely aimed at providing additional variables, and not any more classes (save for perhaps light `background-color`s and `color`s in another PR). I truly wish we could generate the variables another way, but for now, this is a huge boost to customizability purely by color.

- Dynamic variables aren't possible in Sass, so this manually adds nine variations of each color.
- Adds two new functions to `shade` (lighten) and `tint` (darken) each color.
- From there, this documents the changes with docs specific classes to help communicate that these classes aren't part of Bootstrap proper.

/cc @twbs/css-review 

Preview: <https://deploy-preview-29348--twbs-bootstrap.netlify.com/docs/4.3/getting-started/theming/#all-colors>

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -100,3 +100,11 @@
 
   @return mix($color-base, $color, $level * $theme-color-interval);
 }
+
+@function tint-color($color, $level) {
+  @return mix(white, $color, $level * $theme-color-interval);
+}
+
+@function shade-color($color, $level) {
+  @return mix(black, $color, $level * $theme-color-interval);
+}
--- a/site/content/docs/4.3/getting-started/theming.md
+++ b/site/content/docs/4.3/getting-started/theming.md
@@ -242,14 +242,30 @@ Many of Bootstrap's various components and utilities are built through a series 
 
 ### All colors
 
-All colors available in Bootstrap 4, are available as Sass variables and a Sass map in `scss/_variables.scss` file. This will be expanded upon in subsequent minor releases to add additional shades, much like the [grayscale palette](#grays) we already include.
+All colors available in Bootstrap 5, are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we do not create classes for each of these variables.
+
+Sass cannot programmatically generate variables, so we must manually create them ourselves. We specify the midpoint value (`500`) and use custom color functions to tint (lighten) or shade (darken) our colors via Sass's `mix()` color function. Using `mix()` is not the same as `lighten()` and `darken()`—the former blends the specified color with white or black, while the latter only adjusts the lightness value of each color. The result is a much more complete suite of colors, as [shown in this CodePen demo](https://codepen.io/emdeoh/pen/zYOQOPB).
+
+Our `tint-color()` and `shade-color()` functions use `mix()` alongside our `$theme-color-interval` variable, which specifies a stepped percentage value for each mixed color we need. See the `scss/_functions.scss` and `scss/_variables.scss` files for the full source code.
 
 <div class="row">
   {{< theme-colors.inline >}}
   {{- range $.Site.Data.colors }}
     {{- if (and (not (eq .name "white")) (not (eq .name "gray")) (not (eq .name "gray-dark"))) }}
-    <div class="col-md-4">
-      <div class="p-3 mb-3 swatch-{{ .name }}">{{ .name | title }}</div>
+    <div class="col-md-4 mb-3 font-monospace">
+      <div class="p-3 mb-2 swatch-{{ .name }}">
+        <strong class="d-block">${{ .name }}</strong>
+        {{ .hex }}
+      </div>
+      <div class="p-3 bd-{{ .name }}-100">${{ .name }}-100</div>
+      <div class="p-3 bd-{{ .name }}-200">${{ .name }}-200</div>
+      <div class="p-3 bd-{{ .name }}-300">${{ .name }}-300</div>
+      <div class="p-3 bd-{{ .name }}-400">${{ .name }}-400</div>
+      <div class="p-3 bd-{{ .name }}-500">${{ .name }}-500</div>
+      <div class="p-3 bd-{{ .name }}-600">${{ .name }}-600</div>
+      <div class="p-3 bd-{{ .name }}-700">${{ .name }}-700</div>
+      <div class="p-3 bd-{{ .name }}-800">${{ .name }}-800</div>
+      <div class="p-3 bd-{{ .name }}-900">${{ .name }}-900</div>
     </div>
     {{ end -}}
   {{ end -}}
@@ -260,6 +276,10 @@ Here's how you can use these in your Sass:
 
 {{< highlight scss >}}
 .alpha { color: $purple; }
+.beta {
+  color: $yellow-300;
+  background-color: $indigo-900;
+}
 {{< /highlight >}}
 
 [Color utility classes]({{< docsref "/utilities/colors" >}}) are also available for setting `color` and `background-color`.
--- a/site/content/docs/4.3/getting-started/theming.md
+++ b/site/content/docs/4.3/getting-started/theming.md
@@ -250,22 +250,16 @@ Our `tint-color()` and `shade-color()` functions use `mix()` alongside our `$the
 
 <div class="row">
   {{< theme-colors.inline >}}
-  {{- range $.Site.Data.colors }}
-    {{- if (and (not (eq .name "white")) (not (eq .name "gray")) (not (eq .name "gray-dark"))) }}
+  {{- range $color := $.Site.Data.colors }}
+    {{- if (and (not (eq $color.name "white")) (not (eq $color.name "gray")) (not (eq $color.name "gray-dark"))) }}
     <div class="col-md-4 mb-3 font-monospace">
-      <div class="p-3 mb-2 swatch-{{ .name }}">
-        <strong class="d-block">${{ .name }}</strong>
-        {{ .hex }}
+      <div class="p-3 mb-2 swatch-{{ $color.name }}">
+        <strong class="d-block">${{ $color.name }}</strong>
+        {{ $color.hex }}
       </div>
-      <div class="p-3 bd-{{ .name }}-100">${{ .name }}-100</div>
-      <div class="p-3 bd-{{ .name }}-200">${{ .name }}-200</div>
-      <div class="p-3 bd-{{ .name }}-300">${{ .name }}-300</div>
-      <div class="p-3 bd-{{ .name }}-400">${{ .name }}-400</div>
-      <div class="p-3 bd-{{ .name }}-500">${{ .name }}-500</div>
-      <div class="p-3 bd-{{ .name }}-600">${{ .name }}-600</div>
-      <div class="p-3 bd-{{ .name }}-700">${{ .name }}-700</div>
-      <div class="p-3 bd-{{ .name }}-800">${{ .name }}-800</div>
-      <div class="p-3 bd-{{ .name }}-900">${{ .name }}-900</div>
+      {{ range (seq 100 100 900) }}
+      <div class="p-3 bd-{{ $color.name }}-{{ . }}">${{ $color.name }}-{{ . }}</div>
+      {{ end }}
     </div>
     {{ end -}}
   {{ end -}}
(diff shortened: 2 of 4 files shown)


XhmikosR:
Not sure how to bypass fusv here.

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -248,8 +248,20 @@ All colors available in Bootstrap 4, are available as Sass variables and a Sass
...
   {{- range $.Site.Data.colors }}
     {{- if (and (not (eq .name "white")) (not (eq .name "gray")) (not (eq .name "gray-dark"))) }}
-    <div class="col-md-4">
-      <div class="p-3 mb-3 swatch-{{ .name }}">{{ .name | title }}</div>
+    <div class="col-md-4 mb-3 font-monospace">
+      <div class="p-3 mb-2 swatch-{{ .name }}">
+        <strong class="d-block">${{ .name }}</strong>
+        {{ .hex }}
+      </div>
+      <div class="p-3 bd-{{ .name }}-100">${{ .name }}-100</div>
```
XhmikosR: This could be simplified with a `range`, ping me if I forget.

scss/_functions.scss
```diff
@@ -100,3 +100,19 @@
...
+//
+//   @return mix($color-base, $color, $level * $theme-color-interval);
+// }
+
+@function tint-color($color, $level) {
+  @return mix(black, $color, $level * $theme-color-interval);
+}
+
+@function shade-color($color, $level) {
+  @return mix(white, $color, $level * $theme-color-interval);
```
alecpl: tint-color() should use white and shade-color() should use black. At least that's how it works in Less.

site/static/docs/4.3/assets/scss/_colors.scss
```diff
@@ -15,3 +15,103 @@
...
+
+.bd-cyan-100 { color: color-yiq($cyan-100); background-color: $cyan-100; }
+.bd-cyan-200 { color: color-yiq($cyan-200); background-color: $cyan-200; }
+.bd-cyan-300 { color: color-yiq($cyan-300); background-color: $cyan-300; }
+.bd-cyan-400 { color: color-yiq($cyan-400); background-color: $cyan-400; }
+.bd-cyan-500 { color: color-yiq($cyan-500); background-color: $cyan-500; }
+.bd-cyan-600 { color: color-yiq($cyan-600); background-color: $cyan-600; }
+.bd-cyan-700 { color: color-yiq($cyan-700); background-color: $cyan-700; }
+.bd-cyan-800 { color: color-yiq($cyan-800); background-color: $cyan-800; }
+.bd-cyan-900 { color: color-yiq($cyan-900); background-color: $cyan-900; }
```
MisatoTremor: Could use something like this (colors listed explicit as ``$colors`` contains grays)
```scss
@each $color in ("blue", "indigo", "purple", "pink", "red", "orange", "yellow", "green", "teal", "cyan") {
  @for $i from 1 through 9 {
    .bd-#{$color}-#{$i * 100} {
      color: color-yiq(#{#{$color}-100});
      background-color: #{#{$color}-100};
    }
  }
}
```

MartijnCuppens: I don't think this 'll work. But if it does, could you provide a demo?

MisatoTremor: You're right ... I dunno how, but while writing this I must have imagined it working at some point :blush: Sorry for that!

Only posibility I'd see is to use a map instead of plain variables. That drops overridability of the single shades and tints. On the other hand I'm asking myself how often people really would want to change a single color variant when using these shades & tints, so they could just override the whole map.

Actually working example:
```scss
@function shades-and-tints($colors: $colors) {
  $shadesAndTints: ();

  @each $name, $color in $colors {
    $shadesAndTints: map-merge(
      (
        "#{$name}": (
          "100": shade-color($color, 8),
          "200": shade-color($color, 6),
          "300": shade-color($color, 4),
          "400": shade-color($color, 2),
          "500": $color,
          "600": tint-color($color, 2),
          "700": tint-color($color, 4),
          "800": tint-color($color, 6),
          "900": tint-color($color, 8)
        )
      ),
      $shadesAndTints
    );
  }

  @return $shadesAndTints;
}

$colorVariants: shades-and-tints() !default;

@each $colorName, $variants in $colorVariants {
  @each $variant, $colorValue in $variants {
    .bd-#{$colorName}-#{$variant} {
      background-color: #{$colorValue};
      color: color-yiq($colorValue);
    }
  }
}
```

joneff:
Depending on how complicated you want it to be, you could go for material's [palette](https://github.com/angular/components/blob/master/src/material/core/theming/_palette.scss) and [theming](https://github.com/angular/components/blob/master/src/material/core/theming/_theming.scss) approach:

* there is a `_palette.scss` file that contains both predefined colors and color maps *with* contrasting colors
* in addition, there is `_themeing.scss` file that deals with the mundane tasks, like color matching, retrieving colors etc.

Trough some heavy lifting, at the end, there is a "theme" palette that consists of primary, secondary sub palettes (containing base, darker, lighter and contrast colors), complimentary sub palette (colors for text, body / application background etc) and a `warn` color.

White the approach is definitely not for everyone, it actually makes sense to have a "source of truth" palette which you query, when you need a theme color. Of course, should one want to deviate, the rest of the colors are there.

Or to put in some sort of code:

```scss
// Not exactcly code
@function make-palette($base, $darker: null, $ligther: null, $contrast-values: () ) {
    // if darker is not provided use shade($base) --> less way shade
    // if ligther is not provided use tint($base) --> less way tint
    // if contrast-values are not provided use contrast function
    // some code
    @return ( $default, $darker, $lighter, $default-contrast, $darker-contrast, $lighter-contrast);
}

// some transformations ...

$theme: (
    "primary": (
        default: ... , // use for, say link text color or button background color
        darker: ... , // use for, say link hover text color or button hover background color
        lighter: ... ,
        default-contrast: ... , // use for, say button text color
        darker-contrast: ... , // use for, say button hover text color
        lighter-contrast: ...
    ),
    "secondary": (
        default: ... , // use for, say secondary button background color
        darker: ... , // use for, say secondary button hover background color
        lighter: ... ,
        default-contrast: ... , // use for, say secondary button text color
        darker-contrast: ... , // use for, say secondary button hover text color
        lighter-contrast: ...
    ),
    "complimentary": (
        body-bg: ... ,
        body-text: ... ,
        component-bg: ... ,
        component-text: ... 
        // more variables probably 
    )
);
```

If you are trying to create a theme style guide, using the palette would be quite efficient. In addition, having the ability to manually specify colors for the theme palette makes it possible to fulfill designers request, because some times plain white and plain black are not what they want for contrasting color.

mdo:
Just pushed a few updates—sorry for the confusion, folks.

I've fixed the language used for all the tints and shades: `tint()` is now for lighter colors and `shade()` is now for darker colors. We're not using the `lighten()` and `darken()` functions because those do not accomplish what we want here—those functions only adjust the color's `lightness` level. What we really want is a blend of our specified color with white or black, which can be done with `mix()`.

[See this CodePen example](https://codepen.io/joshmcrty/pen/OVZbBr) for how each function generates different outputs.

I've added details around this to the documentation so that folks understand the reasons for our custom functions. Can probably make some more improvements here, but this should definitely help explain some key details around our color system.

Some additional reading and resources I found:
- http://scg.ar-ch.org
- https://thoughtbot.com/blog/controlling-color-with-sass-color-functions

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -242,14 +242,30 @@ Many of Bootstrap's various components and utilities are built through a series
 
 ### All colors
 
-All colors available in Bootstrap 4, are available as Sass variables and a Sass map in `scss/_variables.scss` file. This will be expanded upon in subsequent minor releases to add additional shades, much like the [grayscale palette](#grays) we already include.
+All colors available in Bootstrap 4, are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we do not create classes for each of these variables.
```
MartijnCuppens:
```suggestion
All colors available in Bootstrap 5, are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we do not create classes for each of these variables.
```

scss/_functions.scss
```diff
@@ -100,3 +100,19 @@
 
   @return mix($color-base, $color, $level * $theme-color-interval);
 }
+
+// @function color-level($color-name: "blue", $level: 0) {
```
MartijnCuppens: This comment block can be removed

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -242,14 +242,30 @@ Many of Bootstrap's various components and utilities are built through a series
...
+
+Sass cannot programmatically generate variables, so we must manually create them ourselves. We specify the midpoint value (`500`) and use custom color functions to tint (lighten) or shade (darken) our colors via Sass's `mix()` color function. Using `mix()` is not the same as `lighten()` and `darken()`—the former blends the specified color with white or black, while the latter only adjusts the lightness value of each color. The result is a much more complete suite of colors.
```
MartijnCuppens: Maybe link to the Codepen demo mentioned [here](https://github.com/twbs/bootstrap/pull/29348#issuecomment-535759046)?

MartijnCuppens:
LGTM now, the only thing left to tackle is finding a way to bypass fusv. Probably we'll need to implement something like `// fusv-enable` and `// fusv-disable` comments upstream?

XhmikosR:
BTW this needs a rebase and cleanup of the patches.

As for fusv, PRs welcome as usual :)

site/content/docs/4.3/getting-started/theming.md
MartijnCuppens: Ping? 😄

XhmikosR: I'm gonna have a look soon-ish.

XhmikosR:
I cleaned up the branch, fixed tests and used a range.

I still feel this has a lot of duplication. Can't we use loops in Sass?

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -242,14 +242,24 @@ Many of Bootstrap's various components and utilities are built through a series
...
+All colors available in Bootstrap 5, are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we do not create classes for each of these variables.
+
+Sass cannot programmatically generate variables, so we must manually create them ourselves. We specify the midpoint value (`500`) and use custom color functions to tint (lighten) or shade (darken) our colors via Sass's `mix()` color function. Using `mix()` is not the same as `lighten()` and `darken()`—the former blends the specified color with white or black, while the latter only adjusts the lightness value of each color. The result is a much more complete suite of colors, as [shown in this CodePen demo](https://codepen.io/emdeoh/pen/zYOQOPB).
+
+Our `tint-color()` and `shade-color()` functions use `mix()` alongside our `$theme-color-interval` variable, which specifies a stepped percentage value for each mixed color we need. See the `scss/_functions.scss` and `scss/_variables.scss` files for the full source code.
```
XhmikosR: We could/should probably link to the repo files directly here.

XhmikosR (approved):

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -242,14 +242,24 @@ Many of Bootstrap's various components and utilities are built through a series
...
-    <div class="col-md-4">
-      <div class="p-3 mb-3 swatch-{{ .name }}">{{ .name | title }}</div>
+  {{- range $color := $.Site.Data.colors }}
+    {{- if (and (not (eq $color.name "white")) (not (eq $color.name "gray")) (not (eq $color.name "gray-dark"))) }}
+    <div class="col-md-4 mb-3 font-monospace">
+      <div class="p-3 mb-2 swatch-{{ $color.name }}">
+        <strong class="d-block">${{ $color.name }}</strong>
+        {{ $color.hex }}
+      </div>
+      {{ range (seq 100 100 900) }}
```
mdo: Hugo syntax is so alien to me sometimes, thanks for doing this!

PR #29358, opened 2019-09-02
Examples: use our utilities more.

XhmikosR:


site/content/docs/4.3/examples/floating-labels/floating-labels.css
```diff
@@ -36,7 +36,6 @@ body {
   display: block;
   width: 100%;
   margin-bottom: 0; /* Override default `<label>` margin */
-  line-height: 1.5;
```
XhmikosR: I didn't specify `lh-base` in this case because 1.5 is out base line-height and it seems it's inherited.

MartijnCuppens (approved):

PR #29359, opened 2019-09-02
Tweak form validation snippet.

XhmikosR:
Non-whitespace diff: <https://github.com/twbs/bootstrap/pull/29359/files?w=1>
Preview: <https://deploy-preview-29359--twbs-bootstrap.netlify.com/docs/4.3/examples/checkout/>

### DIFF
--- a/site/content/docs/4.3/examples/checkout/form-validation.js
+++ b/site/content/docs/4.3/examples/checkout/form-validation.js
@@ -9,7 +9,7 @@
   Array.prototype.slice.call(forms)
     .forEach(function (form) {
       form.addEventListener('submit', function (event) {
-        if (form.checkValidity() === false) {
+        if (!form.checkValidity()) {
           event.preventDefault()
           event.stopPropagation()
         }
--- a/site/content/docs/4.3/examples/checkout/form-validation.js
+++ b/site/content/docs/4.3/examples/checkout/form-validation.js
@@ -3,17 +3,18 @@
   'use strict'
 
   // Fetch all the forms we want to apply custom Bootstrap validation styles to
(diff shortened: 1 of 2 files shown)


site/content/docs/4.3/examples/checkout/form-validation.js
```diff
@@ -2,20 +2,18 @@
...
 
-    // Loop over them and prevent submission
-    Array.prototype.filter.call(forms, function (form) {
-      form.addEventListener('submit', function (event) {
-        if (form.checkValidity() === false) {
-          event.preventDefault()
-          event.stopPropagation()
-        }
+  // Loop over them and prevent submission
+  Array.prototype.filter.call(forms, function (form) {
```
Johann-S: Can you remove filter and use `forEach` instead ? There is no filtering inside...

site/content/docs/4.3/examples/checkout/form-validation.js
```diff
@@ -2,20 +2,18 @@
...
+    form.addEventListener('submit', function (event) {
+      if (form.checkValidity() === false) {
```
Johann-S: `if (!form.checkValidity()) {`

XhmikosR:
@Johann-S: those should happen in a separate PR. Feel free to tackle it later.

Johann-S:
Too bad to handle that later when we are able to fix that now

XhmikosR:
It's a different kind of change. It should be done separately.

Johann-S:
Yep, but different commits can be a good solution to fix that with just one PR

Johann-S (approved):

PR #29366, opened 2019-09-03
Removed second instance of "dependencies" and "peerDependencies"

JeremieLeblanc:
"dependencies" and "peerDependencies" were mentioned twice in the package.json file.


Johann-S:
Hi @JeremieLeblanc,

it was intended to be like that, because `jspm` need dependencies and peerDependencies too
https://github.com/twbs/bootstrap/blob/master/package.json#L158-L175

PR #29309, opened 2019-08-27
v4-dev backports

XhmikosR:
Preview:

* <https://deploy-preview-29309--twbs-bootstrap.netlify.com/docs/4.3/components/card/#grid-cards>
* <https://deploy-preview-29309--twbs-bootstrap.netlify.com/docs/4.3/layout/grid/#row-columns>

PR #29367, opened 2019-09-03
Add make-col-auto mixin

### LINKED ISSUE #29362: How to create a col-auto?
Hi, how can I create a col-auto with mixin?

I've tried with `@include make-col(auto)` but this doesn't work.

Can u help me?
Thanks

mdo:
Fixes #29362.

### DIFF
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -41,9 +41,7 @@
       }
 
       .col#{$infix}-auto {
-        flex: 0 0 auto;
-        width: auto;
-        max-width: 100%; // Reset earlier grid tiers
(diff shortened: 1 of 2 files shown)


XhmikosR:
Note that #29309 is still open and need a thorough review due to the backport patches.

MartijnCuppens:
We should replace these lines with the mixin, to prevent duplicate code:
https://github.com/twbs/bootstrap/blob/b4e957d360899c2555cb29a3ae4589405883bba9/scss/mixins/_grid-framework.scss#L43-L47

Apart from that, any other reasons why this is in draft, @mdo?

MartijnCuppens (approved):
LGTM now!

PR #29370, opened 2019-09-04
create a base component

Johann-S:
Create a base component which will allow us to mutualize logics across our components

TODO:

- [x] switch to [`toBeInstanceOf`](https://jasmine.github.io/api/edge/matchers.html#toBeInstanceOf)
- [x] Move constructor in base component
- [x] Move `getInstance` in the base component
- [x] Move `dispose` in base component
- [x] Move `VERSION` in base component
- [x] Address #32264

Preview: https://deploy-preview-29370--twbs-bootstrap.netlify.app/


XhmikosR:
This increases the dist files. I'm not sure about any performance impact either.

We should try and get some benchmarks in before going further with this kind of changes IMO...

Johann-S:
This change doesn't impact perfs, it's just a bit of refactoring to keep our codebase clean. Inheritance is something we already use (Popovers inherit from Tooltip)

Johann-S:
BTW I plan to add a `createInstance` method in the base component to allow our users to create instances without using `new` when they don't need to store an instance

mdo:
Is this worth revisiting or is it too stale/old?

Johann-S:
We should investigate a bit more to move the `_jQueryInterface` inside the `BaseComponent`

Johann-S:
@XhmikosR I worked on that and was able to move `dispose` in the `BaseComponent`

But currently I don't think we can do more than that, so this PR as ready to be reviewed

rohit2sharma95:
IMO, this PR now gives benefit to:
1. #32245 (Logic for CSS selectors can be written in BaseComponent)
2. #32180 (Includes tests for `getInstance` method)
3. #32184 (Event listeners can be written in BaseComponent 🤔)

js/src/base-component.js
```diff
@@ -0,0 +1,36 @@
...
+  /** Static */
+
+  static getInstance(element) {
+    return Data.getData(element, this.DATA_KEY)
+  }
+
+  static get DATA_KEY() {
+    return null
+  }
+}
```
rohit2sharma95, line 34: What if you add a getter here for `VERSION` also and keep the version in the base component because it is same in all the components 🤔
@Johann-S

XhmikosR: @rohit2sharma95 can you try it and see how it goes? If it works, you can make a PR against this branch :)

js/src/base-component.js
```diff
@@ -0,0 +1,36 @@
...
+    this._element = null
+  }
+
+  /** Static */
+
+  static getInstance(element) {
+    return Data.getData(element, this.DATA_KEY)
+  }
+
+  static get DATA_KEY() {
```
XhmikosR, line 31: Does this make sense to have it here if we are just returning null and we add `GET_DATA` is all components?

js/src/base-component.js
Johann-S: Indeed adding `VERSION` would be great 👌

js/src/base-component.js
Johann-S: I created the BaseComponent as an abstract class that's why I added that to be consistent but that's true currently abstract class doesn't exist in JavaScript

XhmikosR: Maybe we should remove this part then? The rest look good to me :)

Johann-S: done

XhmikosR (approved):

XhmikosR:
@Johann-S @rohit2sharma95 should we build the base plugin too? Currently it's inlined in each plugin's dist file in js/dist.

PR #29377, opened 2019-09-06
v5 docs: example shortcode regex

morrissey-ingenious:
example shortcode: see if we can use a regex

Fixes #29351

/cc @XhmikosR 

Preview: <https://deploy-preview-29377--twbs-bootstrap.netlify.com/>

PR #29379, opened 2019-09-06
v5 docs: added wrapping class for rounded utils

### LINKED ISSUE #29343: Tweak /utilities/borders/#border-radius
https://twbs-bootstrap.netlify.com/docs/4.3/utilities/borders/#border-radius

This doesn't look so good, and also on mobile the images don't have any vertical margins. Could be improved IMO.

morrissey-ingenious:
Tweak /utilities/borders/#border-radius

Fixes #29343

/cc @XhmikosR 

Preview: <https://deploy-preview-29379--twbs-bootstrap.netlify.com/docs/4.3/utilities/borders/#border-radius>


mdo:
Should be able to add utility classes to this instead of writing custom CSS—adding `.mb-1` or `.mb-2` to update them.

morrissey-ingenious:
@mdo I agree, I did it this way as to keep the example code clean, this pull follows the pattern of the border util above this if you look at the source.

I just wrote a regex for the example shortcode (pull #29377) to strip out the `bg-placeholder-img`.  

If you want to keep the docs clean and not use custom css, and keep the examples clean, can i suggest we update the example regex to strip all classes before and including say `example-strip-before`, and an after for `example-strip after`.

XhmikosR:
I agree, we should use our utils here. Let's wait until we are done with the shortcode.

XhmikosR:
@mdo: how do you want to proceed with this? I don't mind using CSS for just this case since it does fixes the issue.

site/assets/scss/_component-examples.scss
```diff
@@ -228,6 +228,12 @@
   }
 }
 
+.bd-example-rounded-utils {
```
mdo: Maybe change this to the generic `.bd-example` class and use a child selector?

```scss
.bd-example > [class^="rounded"] { ... }
```

Would avoid a new selector and parent class.

XhmikosR: Doesn't it feel stricter with the extra parent class? TBH I think it's better this way...

mdo: The suggestion I made matches most of the `.bd-example` customizations in this file. I'd like to take a crack at updating it after this because most of it is triggering off the parent class, which is fine to me and avoids writing extra HTML just to have our docs.

XhmikosR:
We are not writing extra HTML,we just add one class. See my last patch and
the preview

PR #29387, opened 2019-09-09
Docs for close icon

ysds:
Change the name to a "close button" for clarity and consistency.

It is free to backport this to v4.


XhmikosR:
This is not an icon, though? It can be used both ways.

/CC @mdo

mdo (approved):

PR #29400, opened 2019-09-11
Switch to chokidar-cli for watching.

XhmikosR:
~~The main difference is that by default it won't run the referenced subscripts, only when a file changes.~~

chokidar seems to be lighter than nodemon:

<https://packagephobia.now.sh/result?p=chokidar-cli> vs <https://packagephobia.now.sh/result?p=nodemon>

I thought about backporting this to v4-dev too, but we need to use chokidar-cli 1.x there, so I'll probably leave it as is for v4-dev.

Please try it on macOS and Linux and let me know your thoughts.

/CC @mdo @Johann-S @MartijnCuppens @ysds


ysds (approved):

XhmikosR:
Waiting for an upstream update to chokidar 3.1.0.

XhmikosR:
So, there's nodemon v2.0.0 out which is a lot smaller than before since they upgraded chokidar.

I'm gonna close this for now and I will revisit if needed later.

PR #29409, opened 2019-09-13
Grid card example tweaks

ysds:
Fix for https://github.com/twbs/bootstrap/pull/29309#issuecomment-527167585

image

Preview: <https://deploy-preview-29409--twbs-bootstrap.netlify.com/docs/4.3/components/card/#grid-cards>

### DIFF
--- a/site/content/docs/4.3/components/card.md
+++ b/site/content/docs/4.3/components/card.md
@@ -603,10 +603,10 @@ Just like with card groups, card footers in decks will automatically line up.
 
 ### Grid cards
 
-Use the Bootstrap grid system and its [`.row-cols` classes]({{< docsref "/layout/grid#row-columns" >}}) to control how many grid columns (wrapped around your cards) you show per row. For example, here's `.row-cols-md-2` splitting four cards to equal width across multiple rows, from the medium breakpoint up.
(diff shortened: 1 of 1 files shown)


XhmikosR:
@ysds did you try this on my v4-dev-xmr branch? I mean, the issue is mostly visible on v4-dev.

I'll try it tomorrow myself too, thanks!

XhmikosR:
I tested this and fixes the smaller than md issue but on large screens the images do look they have a wrong aspect ratio. We use `100%` for width and `140` for height.

Maybe we should just use 180 like the examples above?

/CC @MartijnCuppens

MartijnCuppens:
Yup, 180 feels better indeed

XhmikosR:
Alright, updated to 180px.

XhmikosR:
Maybe @mdo chose 140 for a reason though. I mean, for master, 140px doesn't look bad, but for v4-dev it doesn't look good.

https://deploy-preview-29309--twbs-bootstrap.netlify.com/docs/4.3/components/card/#grid-cards

I'm open to suggestions. Maybe we should use 180 in v4-dev only.

XhmikosR:
Alright, I reverted the height change. I will make it only in v4-dev later.

XhmikosR (approved):

Issue #29245, opened 2019-08-13
Vertical offset of progress bar label on IE11


Trainmaster:
- Operating system and version: Windows 7
- Browser and version: Internet Explorer 11 (11.0.9600.19399)

In https://github.com/twbs/bootstrap/pull/22703 the line `line-height: $progress-height;` was removed. But it seems like it's not really unnecessary as the commit message says, at least not in IE11.

Just navigate to https://getbootstrap.com/docs/4.3/components/progress/#labels and change
```html
<div class="bd-example">
```

to something like 

```html
<div class="bd-example" style="line-height: 30px;">
```

![progress-bar-label-offset-in-IE11](https://user-images.githubusercontent.com/2514421/62930626-1331a500-bdbd-11e9-8cbb-42426a06ccf1.png)

PR #29413, opened 2019-09-16
Prevent vertical offset in IE 11

Trainmaster:
Fix https://github.com/twbs/bootstrap/issues/29245


MartijnCuppens:
I think a `line-height: 0` would be more flexible than a fixed value here.

MartijnCuppens (approved):

Jaygiri:
Thanks to IE 11. This is a breaking change for many.

MartijnCuppens:
> Thanks to IE 11. This is a breaking change for many.

Please open a ticket with a demo if this is the case.

PR #29414, opened 2019-09-16
index path fix

did1335:
1. bundle.js and index.html are in the same folder so --><script src="bundle.js"></script>
2.
The popper.js path is wrong
import '../../../node_modules/popper.js/dist/umd/popper.min.js'


Johann-S:
Hi @did1335,

- The version in the coverage folder of `bundle.js` is the transpiled one that's why we use that one.
- About Popper.js no need to specify the path because it's a deps of our project

PR #29420, opened 2019-09-18
Remove "extra" section from composer.json

andreasschroth:
This section is meaningless and doesn't serve any purpose.

To give some more context: The branch alias is good for mapping `dev-master` (which is the latest dev version of a project) to a specific development branch. But if your `master` branch is always the latest dev version (as it is the case with Bootstrap), no mapping is needed.

Only problem could be, if another package is requiring a specific version and that conflicts with `dev-master` (on the `master`) branch if there's no alias set. But I would say this is VERY unlikely with a project like Bootstrap.

See also https://getcomposer.org/doc/articles/aliases.md#branch-alias .


XhmikosR:
In the past, i.e. when v4 was being developed, development happened on `v4-dev` and `master` was behind.

Do you think this could break things if we remove it?

andreasschroth:
No, unless someone would require another package that depends on the version being developed, e.g. `4.x`AND it's added in the project as `dev-master`. But why would another package depend on an unreleased version...?

I would say it's absolutely safe to remove it.

mdo (approved):

PR #29427, opened 2019-09-20
Sass: remove redundant stylelint inline suppressions.

XhmikosR:
`selector-max-universal` is set to `1` in our stylelint config  upstream BTW.

`-rd` works fine as long as one clears the local cache folder. But perhaps we should enable it anyway?

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -176,7 +176,6 @@
   // Place margin between footer elements
   // This solution is far from ideal because of the universal selector usage,
   // but is needed to fix https://github.com/twbs/bootstrap/issues/24800
-  // stylelint-disable-next-line selector-max-universal
   > * {
     margin: $modal-footer-margin-between / 2;
   }
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -1,5 +1,3 @@
-// stylelint-disable selector-no-qualifying-type
-
 //
 // Textual form controls
 //
--- a/scss/helpers/_position.scss
+++ b/scss/helpers/_position.scss
@@ -1,5 +1,3 @@
-// stylelint-disable declaration-no-important
-
 // Shorthand
 
 .fixed-top {
--- a/site/assets/scss/_masthead.scss
+++ b/site/assets/scss/_masthead.scss
@@ -1,5 +1,3 @@
-// stylelint-disable declaration-no-important
-
 .bd-masthead {
   padding: 3rem 0;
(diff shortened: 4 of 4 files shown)


MartijnCuppens:
What is `-rd`?

XhmikosR:
```
    --report-needless-disables, --rd

      Also report errors for stylelint-disable comments that are not blocking a lint warning.
      The process will exit with code 2 if needless disables are found.

    --report-invalid-scope-disables, --risd

      Report stylelint-disable comments that used for rules that don't exist within the configuration object.
      The process will exit with code 2 if invalid scope disables are found.
```

MartijnCuppens:
> selector-max-universal is set to 1 in our stylelint config upstream BTW.

Any idea why stylelint isn't failing with the current scss than? We do have more than one universal selector.

XhmikosR:
>We do have more than one universal selector.
In the same selector?

MartijnCuppens:
Ok, never mind, the documentation states:
> Limit the number of universal selectors **in a selector**.

I thought it was about the global number of universal selectors, now I get it why that statement was moot.

> `-rd` works fine as long as one clears the local cache folder.

What happens if the local cache folder isn't cleared? Can we get false results?

XhmikosR:
Yeah, I definitely get false results. I mean, stuff that can be removed but are not flagged as such. But on CI this will fail properly.

package.json
```diff
@@ -24,7 +24,7 @@
     "css": "npm-run-all css-compile css-prefix css-minify",
     "css-compile": "node-sass --output-style expanded --source-map true --source-map-contents true --precision 6 scss/ -o dist/css/",
     "css-lint": "npm-run-all --continue-on-error --parallel css-lint-*",
-    "css-lint-stylelint": "stylelint \"**/*.{css,scss}\" --cache --cache-location .cache/.stylelintcache",
+    "css-lint-stylelint": "stylelint \"**/*.{css,scss}\" --cache --cache-location .cache/.stylelintcache--report-needless-disables",
```
ysds: Isn't a space necessary before the `--report-needless-disables`?

I tried run `npm run lint` after fix this and remove my `.cache` folder, but stylelint reported a lot of errors:

image
and more.

ysds: refs: https://github.com/stylelint/stylelint/issues/4203

XhmikosR: Ooops, the space should be there.

XhmikosR: But yeah, I see now the issue, hence why i hadn't enabled it before :)

I'll drop it for now.

ysds (approved):

PR #29436, opened 2019-09-24
Fix inconsistent hr height with size attribute

ashfahan:
When hr used with size attribute results in height with 2px less than size specified, this is due to removal of border on hr elements.

Instead we can set its border-style to solid without remove it

[Code pen with fix](https://codepen.io/ashfahan/pen/bGbzqrQ)

Browser : Chrome [78.0.3904.21 Dev] , Chrome [Latest Stable]
OS: Windows 10 1903

_sorry about duplicate pull request previous one got overridden by one of my plugin_

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -70,12 +70,13 @@ hr {
   margin: $hr-margin-y 0;
   color: $hr-color; // 1
   background-color: currentColor;
-  border: 0;
+  border-style: solid none;
   opacity: $hr-opacity;
 }
 
 hr:not([size]) {
   height: $hr-height; // 2
+  border: 0;
 }
 
 


XhmikosR:
You should also try Firefox and IE 11.

ashfahan:
@XhmikosR tested on IE11 and firefox with Windows
However when i tested it on chrome 77 Mojave the results were unexpected

Height on hr was 26px when 30 size was specified , That is default `box-sizing : content-box` seems to work better there 🤔 💭

ashfahan:
After testing on Chrome and IE and Firefox

Size attribute in IE and firefox assigned the same height as size attribute unlike Chrome which applies 2px less than size attribute.

So even though if we do changes , one's behavior might break other browser behavior.

I think its better to leave it as issue than PR.

PR #29440, opened 2019-09-25
workflows/test.yml: specify `CI=true`

XhmikosR:
Would be nice if we could set this globally, but I couldn't find a way.

This will suppress some postinstall messages.


Johann-S (approved):

XhmikosR:
Does anybody knows if we can set this env var globally without having to specify it for each step?

PR #29441, opened 2019-09-25
Better radio input

codecalm:
Hi! Radio inputs looks different than older Bootstrap versions. Small upgrade with fix.

Screenshot 2019-09-25 at 16 21 23

Demo:
https://deploy-preview-29441--twbs-bootstrap.netlify.com/docs/4.3/forms/checks/#radios


florianlacreuse:
According to [this codepen](https://codepen.io/emdeoh/pen/LMeYxe) made by @mdo, we could also keep the current background image svg (with radius 3) and change the `$form-check-input-checked-bg-size` to `calc($form-check-input-width - .5em)`. The look seems a little bit better with big font-size values.

mdo (approved):

PR #29445, opened 2019-09-25
Remove outline from select box in FF

codecalm:
Firefox added a gray border to custom select. This fix removes it.

More info here: https://stackoverflow.com/questions/3773430/remove-outline-from-select-box-in-ff/18853002#18853002

**Before:**
Screenshot 2019-09-25 at 23 32 05

**After:**
Screenshot 2019-09-25 at 23 31 52

### DIFF
--- a/scss/forms/_form-select.scss
+++ b/scss/forms/_form-select.scss
@@ -61,7 +61,7 @@
   // Remove outline from select box in FF
   &:-moz-focusring {
     color: transparent;
-    text-shadow: 0 0 0 $black;
+    text-shadow: 0 0 0 $form-select-color;
   }
 }
 
--- a/scss/forms/_form-select.scss
+++ b/scss/forms/_form-select.scss
@@ -57,6 +57,12 @@
   &::-ms-expand {
     display: none;
   }
+
+  // Remove outline from select box in FF
+  &:-moz-focusring {
+    color: transparent;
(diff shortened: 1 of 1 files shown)


scss/forms/_form-select.scss
```diff
@@ -57,6 +57,12 @@
   &::-ms-expand {
     display: none;
   }
+
+  // Remove outline from select box in FF
+  &:-moz-focusring {
+    color: transparent;
+    text-shadow: 0 0 0 #000;
```
XhmikosR: Please use the `$black` variable.

XhmikosR (dismissed):

codecalm:
fixed

ysds:
reproducible on v4

ysds:
I found a duplicate issue #19398.

I agree to remove this dot outline in our custom select, but I haven't checked yet if this is the best solution because mdn is saying "Do no use `:moz-focusring`".
https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-focusring

/cc @patrickhlauke

mdo:
@ysds They show that anytime something's "non-standard" I believe. I don't see any new problems for users here, so I think it's safe?

ysds:
@mdo yep, that style LGTM :)

MartijnCuppens (approved):

PR #29448, opened 2019-09-26
Split GitHub Actions

XhmikosR:
This is mostly an experiment. If it's faster (even without caching) we could do things this way since Actions run in parallel.

TODO:

- [x] Fix caches
- [ ] Adapt master branch protection rules
- [ ] Backport it to v4?

### DIFF
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   CI: true
-  NODE: 10.x
+  NODE: 12.x
 
 jobs:
   browserstack:
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -1,5 +1,14 @@
 name: BrowserStack
-on: [push]
+
+on:
+  push:
+    branches:
+      - master
+      - "**-ci"
+  pull_request:
+    branches:
+      - "**"
+
 env:
   CI: true
   NODE: 10.x
--- a/.github/workflows/browserstack.yml
+++ b/.github/workflows/browserstack.yml
@@ -1,0 +1,39 @@
+name: BrowserStack
+on: [push]
+env:
+  CI: true
+  NODE: 10.x
+
+jobs:
+  browserstack:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set Node.js version
+        uses: actions/setup-node@v1
+        with:
+          node-version: "${{ env.NODE }}"
+
+      - name: Set up npm cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/browserstack.yml') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/browserstack.yml') }}
+            ${{ runner.OS }}-node-v${{ env.NODE }}-
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run dist
+        run: npm run dist
+
+      - name: Run BrowserStack tests
+        run: npm run js-test-cloud
+        env:
+          BROWSER_STACK_ACCESS_KEY: "${{ secrets.BROWSER_STACK_ACCESS_KEY }}"
+          BROWSER_STACK_USERNAME: "${{ secrets.BROWSER_STACK_USERNAME }}"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,35 @@
+name: Docs
+on: [push, pull_request]
+env:
+  CI: true
+  NODE: 10.x
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set Node.js version
+        uses: actions/setup-node@v1
+        with:
+          node-version: "${{ env.NODE }}"
+
+      - run: java -version
+
+      - name: Set up npm cache
+        uses: actions/cache@v2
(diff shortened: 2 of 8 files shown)


XhmikosR:
Some notes so far:

1. we end up with duplicate (push and PR) runs for some of the tasks when one of the members makes a PR. I'm not sure if we can improve this
2. `Tests` is basically the JS Tests only, so we should rename it if we go ahead with this
3. The README.md badges, not sure which we should add
4. The branch protection rules will need to be adapted if we go ahead with this

.github/workflows/browserstack.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: npm ci
+        env:
+          CI: true
+
+      - name: Run dist
+        run: npm run dist
+
+      - name: Run BrowserStack tests
+        run: npm run js-test-cloud
+        if: github.repository == 'twbs/bootstrap'
```
XhmikosR: Not sure if this check makes sense anymore?

XhmikosR:
@mdo: I definitely like this approach, but I don't like the duplicate runs when we make a PR ourselves

.github/workflows/browserstack.yml
MartijnCuppens: Why was this added in the first place?

XhmikosR:
The secrets are not available to other forks

XhmikosR:
Perhaps we should change the actions to fire up on `pull_request` only? Assuming we always use PRs, which we do.

MartijnCuppens:
> Perhaps we should change the actions to fire up on pull_request only? Assuming we always use PRs, which we do.

Fine by me, I assume the test rerun when we push something to the PR?

XhmikosR:
Not sure TBH.

Let me think more about it. I know for sure this is faster, though.

MartijnCuppens:
The tests rerun when we push to an existing PR with only `pull_request` (tested this with another project, https://github.com/Intracto/buildozer/pull/34)

XhmikosR:
Yeah but then they don't run when merged.

We could use some help, maybe I should ask on the community forum.

mdo:
@XhmikosR Send me anything via Slack DM if I can ask internally.

XhmikosR:
So, I renamed "Tests" to "JS Tests". I just link to this in README.md.

1. we need to agree on the workflow names
2. we need to make sure we are not missing any tests
3. I'd still like to not run everything twice when one of the team members open a PR; I guess we'll live with that
4. **just before the merge** we need to adapt the master branch protection rules
5. after the merge I need to backport this in v4-dev

MartijnCuppens:
The names lgtm and it looks like everything we covered before also is covered in this PR.

XhmikosR:
Only thing I don't like is the double actions when a team member makes a
PR. We could live with it, but there's a limit in parallel actions which we
will hit. It's not the end of the world, it's just not optimal.

MartijnCuppens (approved):

XhmikosR:
Alight, I'm not sure how to skip the duplicate runs, but this is still a lot faster than what we had.

mdo (approved):

iamandrewluca:
You can also limit the runs only when some specific files were modified.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths

mdo:
This okay to merge @XhmikosR?

XhmikosR:
> This okay to merge @XhmikosR?

I'm still not 100% happy with the duplicate runs... Also, this approach is a little "dangerous" in the sense that we need to be 100% careful to not miss testing something.

Regardless, I'll rebase this later. I also added a dart sass build action.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
...
+          path: ~/.npm
+          key: ${{ runner.os }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/docs.yml') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}-${{ hashFiles('.github/workflows/docs.yml') }}
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Test docs
+        run: npm run docs
```
ffoodd: Shouldn't we generate dist files before running the docs task, to ensure the build is accurate?

ffoodd:
> Yeah but then they don't run when merged.

You can restrict workflow on push events to specific branches (like `master`, or, say, `v5-dev`).

Since you're basically always working through PRs, a push event on `v5-dev` should always match a PR merge:


```sh
on:
  push:
      branches:
        - v5-dev
```

You can also target pull_request with a closed action and a `merged` argument equaling true:

```sh
on:
  pull_request:
    types:
        closed
    branches:
      - v5-dev

jobs:
  merge:
      - name: Just merged
        if: github.event_name == 'pull_request' && github.event.pull_request.action == 'closed' && github.event.pull_request.merged == true
```

I'm not quite sure if all of this work as intended, but currently trying…

.github/workflows/docs.yml
XhmikosR: Not really since the docs tests don't depend on the dist files.

XhmikosR:
Yeah, I know all this. The problem is that I personally find it handy to have CI run on push before I proceed with a PR.

ffoodd:
You'd want CI to run on any PRs, but also on push to `master`, am I right?

If so, I think this should do the job:

```yml
name: BrowserStack
on:
  push:
    paths:
      - 'js/**'
    branches:
      - master
  pull_request:
    paths:
      - 'js/**'
```

Restricting path would prevent to run if no relevant file is changed, and restricting to the `master` branch for `push` event seems to prevent the duplicate runs.

My current repo is way more quieter than this one obviously, but it seems to work fine in my case.

XhmikosR:
I thought about this, and I think we should only run tests on PRs, since we all create PRs to merge something.

That being said, my only concern is missing out testing some parts in case our scripts change...

But it definitely is faster.

Let me know your thoughts again and how do you think we should proceed.

XhmikosR:
@twbs/team this is ready for review. @mdo if this get approved, before the merge we need the master branch protection adapted.

If it's merged, I'll monitor things and if everything works good, I'm going to backport it.

mdo (approved):

PR #29449, opened 2019-09-26
remove .form-group leftovers

wojtask9:
`.form-group` was removed in https://github.com/twbs/bootstrap/pull/28450/

NOTE: Also removed whole `label` styling inside `.form-inline` and `vertical-align` from `.form-control` but I'm not sure about this. My tests shows that everything is OK but maybe my thinking is wrong.

### DIFF
--- a/scss/forms/_layout.scss
+++ b/scss/forms/_layout.scss
@@ -22,9 +22,6 @@
 // Make forms appear inline(-block) by adding the `.form-inline` class. Inline
 // forms begin stacked on extra small (mobile) devices and then go inline when
 // viewports reach <768px.
-//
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
> NOTE: Also removed whole label styling inside .form-inline and vertical-align from .form-control but I'm not sure about this. My tests shows that everything is OK but maybe my thinking is wrong.

Leave these changes for now, we'll probably gonna rely on the new grid to replace `.form-inline` (See #28517 or https://deploy-preview-28517--twbs-bootstrap.netlify.com/docs/4.3/forms/layout/#inline-forms)

MartijnCuppens (approved):
I removed the changes to `.form-inline` for now

Issue #29428, opened 2019-09-21
Split variables into different files, at least only for grid


SilencerWeb:
I need only grid from the bootstrap, but for this, I need to import `_variables.scss` with over 400 variables and some of my variables overwrite bootstrap variables (like font-size and line-height variables that are not related to the grid at all) and because of this, everything fails. It would be much better if I could import only variables for the grid.

![image](https://user-images.githubusercontent.com/20713191/65370402-cbcaee00-dc71-11e9-959f-c46e7f287a02.png)

florianlacreuse:
I don't know about splitting variables file but see #28240 as it may solves your issue.

SilencerWeb:
@florianlacreuse yeah, I've already seen this issue, it won't help in my case

InsanityMeetsHH:
@SilencerWeb to avoid this error you could use a variable prefix / namespace. I use the term "theme".
e.g. $theme-font-size, $theme-line-height in your case.

mdo:
Closing as a won't fix after closing #29451.

PR #29451, opened 2019-09-27
v5: Sass file structure idea

mdo:
Spiking out an idea to bring some clarity to our Sass files, specifically around the growing number of variables, functions, and settings. This is meant to invite further conversation, and shouldn't be reviewed yet.

## Context

My idea here is that there can be greater clarity around the layers of Sass code. We have a few features to how we build this project, including:

- Several global settings (via `$enable-` variables)
- An entire color system (soon to be expanding with #)
- Layout and component settings (via our breakpoints, grid configuration, spacer ranges, etc)
- And lastly, more typical variables for nearly every `property: value` pairing we want to support customization for

Putting all that in `_variables.scss` makes for a conflated and potentially confusing file.

## Changes
This PR stubs out at least two changes:

- Adds a new `_colors.scss` file for our growing list of color variables
- Adds a new `_settings.scss` file for our `$enable-` variables for global configuration

I'm thinking our other configuration variables can go in `_settings.scss`) as well (e.g., breakpoints, grids, containers, spacers, escaped characters). The rest of our variables, including the additional Sass map here and there, can remain in the original file as they support our components.

## Minimizing requirements for other builds
Our grid dist files currently bring in the entire Sass file, as mentioned in #29428. This fix would only require `_settings.scss`, `_functions.scss`, and `_mixins.scss`. This is a considerable reduction and more specific, while not overly complicated.

## Split it all?
I've thought about this for a bit and believe we shouldn't split the variables files by every component. This makes our lives more difficult (looking across more files if variables are reusing other variables), and takes away the option of generating or reading from a single file (a la several customizers out there).

---

Thoughts? Alternate ideas? Additional concerns?

## TODOs

- [ ] Fix imports


MartijnCuppens:
I like it. The only downside is that we'll have to include all files separately. Maybe we can go fore the same approach as we do with our mixins (one file which imports all mixin files in the `utilities` folder)?

```text
scss/
├── _variables.scss // Includes imports to all files in variables folder
├── variables/
│   ├── _colors.scss
│   ├── _settings.scss
│   └── _variables.scss // Maybe change this name, because another file named `_variables.scss` might be a little confusing? Maybe `_config.scss`?
└── ...
```

mdo:
I started off writing this as not in favor, but I could see that working well if we continue to break apart the variables file. This adds another file effectively, so while it saves on the import list, it does add another file to the directory.

Going to wait for more feedback for a bit before deciding.

manuchali (approved):

XhmikosR:
I agree, we are better off not importing one file, i.e. this initial patch.

astrahov:
It is possible to make publicly the sections on the site:

```
scss/
├── components/
│   ├── _components.scss // includes: '@import "alert";'...
│   ├── forms/
│   ├── _alert.scss
│   ├── _badge.scss
│   ├── _forms.scss
│   └── ...
│
├── content/
│   ├── _content.scss // includes: '@import "tables";'...
│   ├── _tables.scss
│   ├── _type.scss
│   └── ...
│
├── mixins/
│   ├── _mixins.scss // includes: '@import "border-radius";'...
│   ├── _border-radius.scss
│   ├── _box-shadow.scss
│   └── ...
│
├── layout/
│   ├── _layout.scss // includes: '@import "grid";'...
│   ├── _grid.scss
│   └── ...
│
├── variables/
│   ├── _variables.scss // includes: '@import "colors";'...
│   ├── _colors.scss
│   ├── _cards.scss
│   ├── _buttons-and-forms.scss
│   ├── _buttons.scss
│   ├── _forms.scss
│   └── ...
│
├── vendor/
│   ├── _rfs.scss
│   └── ...
│
├── utilities/
│   ├── _utilities.scss // includes: '@import "align";'...
│   ├── _align.scss
│   ├── _background.scss
│   └── ...
│
├── _functions.scss
└── bootstrtrap.scss // includes: '@import "variables/variables";', '@import "layout/layout";'...
```

You can think of file names.

MartijnCuppens:
I was also thinking about reordering our file structure like @astrahov mentions, I would make sense, since our docs also have this structure. We might need to move the forms folder one level up, since right now we treat them as another group next to components.

XhmikosR:
Maybe we should take into consideration the new Sass module feature while at it? Like have an index.scss per folder or something like that.

astrahov:
```
scss/
├── components/
│   ├── _index.scss // includes: '@import "alert";'...
│   ├── _alert.scss
│   ├── _badge.scss
│   └── ...
│
├── content/
│   ├── _index.scss // includes: '@import "tables";'...
│   ├── _tables.scss
│   ├── _type.scss
│   └── ...
│
├── forms/
│   ├── _index.scss // includes: '@import "form-check";'...
│   ├── _form-check.scss
│   ├── _form-control.scss
│   └── ...
│
├── mixins/
│   ├── _index.scss // includes: '@import "border-radius";'...
│   ├── _border-radius.scss
│   ├── _box-shadow.scss
│   └── ...
│
├── layout/
│   ├── _index.scss // includes: '@import "grid";'...
│   ├── _grid.scss
│   └── ...
│
├── variables/
│   ├── _index.scss // includes: '@import "colors";'...
│   ├── _colors.scss
│   ├── _cards.scss
│   ├── _buttons-and-forms.scss
│   ├── _buttons.scss
│   ├── _forms.scss
│   └── ...
│
├── vendor/
│   ├── _index.scss // includes: '@import "rfs";'...
│   ├── _rfs.scss
│   └── ...
│
├── utilities/
│   ├── _index.scss // includes: '@import "align";'...
│   ├── _align.scss
│   ├── _background.scss
│   └── ...
│
├── _functions.scss
└── bootstrtrap.scss // includes: '@import "variables";', '@import "layout";'...
```

swaibat (approved):
great job

scss/_settings.scss
```diff
@@ -0,0 +1,23 @@
...
+$enable-transitions:                          true !default;
+$enable-prefers-reduced-motion-media-query:   true !default;
+$enable-grid-classes:                         true !default;
+$enable-pointer-cursor-for-buttons:           true !default;
+$enable-responsive-font-sizes:                false !default;
+$enable-validation-icons:                     true !default;
+$enable-deprecation-messages:                 true !default;
+
+// Characters which are escaped by the escape-svg function
+$escaped-characters: (
```
MartijnCuppens: I think this should be moved back to the variables file?

mdo:
I've pushed an update here to bring in the latest changes (specifically the color theme variables I made months ago). Something still TBD I think is how much goes into the new settings file. Right now this is all `$enable-` variables, but I'd like to include the grid and spacers I think. These variables are more configuration than style, but I might be forcing an unnecessary relationship between parts of the codebase.

Thoughts?

Inchworm333:
Coming from the theming side, I feel like the colors could possibly be split up even further with a _colors.scss and possibly a _theme-colors.scss. My reasoning is that currently if someone wants to add lets say `$neon-green` and then change `$primary: $red` you would currently need to redefine `$red` otherwise sass won't be able to find it. I also agree with @astrahov although I might suggest adding a subfolder and further splitting up the `$form-` variables because there are quite a few of those.

mdo:
Closing for now.

PR #29454, opened 2019-09-27
Remove duplicated td selector

wojtask9:
the same rule at https://github.com/twbs/bootstrap/blob/master/scss/_tables.scss#L15

Introduced by #28450


MartijnCuppens (approved):
Thanks!

PR #29456, opened 2019-09-27
Added a class for a new feature to fix table header on top

ed1nh0:
Table header with the `.thead-fixed` class applied fix its cells on top allowing only `<tbody>` to scroll.


MartijnCuppens:
Hi @ed1nh0,

- This needs some documentation.
- I would increase the `z-index`, using `$zindex-sticky` would make sense.
- The top property should be adjustable with a variable.
- We should probably increase the top value for our docs because we have a sticky header for larger viewports.

ed1nh0:
> This needs some documentation.

What can I do to improve it?

> I would increase the `z-index`, using `$zindex-sticky` would make sense.

You mean create a variable to parameterize it?

> The top property should be adjustable with a variable.

Yeah sure! I guess this could help.

> We should probably increase the top value for our docs because we have a sticky header for larger viewports.

Are you sure? I didn't find anything that makes the `<thead>` fixed in someway.

scss/_tables.scss
```diff
@@ -187,3 +187,19 @@
...
   }
 }
+
+
+// Fixed header
+//
+// Table header with the `.thead-fixed` class applied fix its cells on top
+// allowing only <tbody> to scroll.
+
+.table {
```
XhmikosR: This should be move into the previous block. No need to duplicate the selector. See the lint error.

XhmikosR (changes requested):

scss/_tables.scss
XhmikosR: That's not a fix.

ed1nh0: Rather than move the block upwards I added the
```// stylelint-disable-next-line no-duplicate-selectors```
comment line right before it.

Comment (mdo) [reply scss/_tables.scss]: He's saying not to do that Stylelint override, and instead move this into an existing `.table` ruleset.

File: scss/_tables.scss [hunk 2]
```diff
@@ -187,3 +187,20 @@
...
+//
+// Table header with the `.thead-fixed` class applied fix its cells on top
+// allowing only <tbody> to scroll.
+
+// stylelint-disable-next-line no-duplicate-selectors
+.table {
+  .thead-fixed {
+    th {
+      position: sticky;
+      top: $table-head-fixed-margin-top;
```
mdo: This isn't a `margin-top`, it's `top`. Variable name should reflect that: `$table-head-sticky-top`.

scss/_tables.scss
```diff
@@ -187,3 +187,20 @@
...
+
+
+// Fixed header
+//
+// Table header with the `.thead-fixed` class applied fix its cells on top
+// allowing only <tbody> to scroll.
+
+// stylelint-disable-next-line no-duplicate-selectors
+.table {
+  .thead-fixed {
```
mdo: Is it fixed, or is it sticky?

scss/_tables.scss
```diff
@@ -187,3 +187,20 @@
...
+    th {
+      position: sticky;
```
mdo: This needs to be wrapped in `@supports (position: sticky) { ... }`

mdo (changes requested):

MartijnCuppens:
> **This needs some documentation.**
> What can I do to improve it?

You'll need to edit the documentation markdown files, in this case [tables.md](https://github.com/twbs/bootstrap/blob/master/site/content/docs/4.3/content/tables.md)

> **We should probably increase the top value for our docs because we have a sticky header for larger viewports.**
> Are you sure? I didn't find anything that makes the <thead> fixed in someway.

I'm talking about the sticky header we have when we scroll. See https://twbs-bootstrap.netlify.com/docs/4.3/content/tables/#overview
This is something you'll need to add in our documentation CSS.

scss/_tables.scss
ed1nh0: Done.

scss/_tables.scss
ed1nh0: It's `sticky`. Done.

scss/_tables.scss
ed1nh0: Done.

scss/_tables.scss
ed1nh0: Make sense. Done.

ed1nh0 (commented):
Everything done.

site/content/docs/4.3/content/tables.md
```diff
@@ -644,6 +644,76 @@ Similar to tables and dark tables, use the modifier classes `.thead-light` or `.
 </table>
 {{< /example >}}
 
+### Sticky header
+
+Add `.thead-sticky` to stick it to the top of the table's parent container. Doing so only the body of the table will scroll in case it exceeds its height.
```
mdo:
```suggestion
Add `.thead-sticky` to a `<thead>` to make it stick to the top of the table's parent container while the `<tbody>` contents continue to scroll.
```

scss/_tables.scss
```diff
@@ -129,6 +129,21 @@
...
+  //
+  // Table header with the `.thead-sticky` class applied sticks its cells on top
+  // allowing only <tbody> to scroll.
+
+  @supports (position: sticky) {
+    .thead-sticky {
+      th {
+        position: sticky;
+        top: $table-head-sticky-top;
+        z-index: $zindex-sticky;
```
mdo: Checking this out on the built Netlify preview, I'm wondering if this is the right `z-index` level. https://deploy-preview-29456--twbs-bootstrap.netlify.com/docs/4.3/content/tables/#sticky-header

This stuff shouldn't overlay any other fixed, stickied, absolute, etc navbars and other items. I feel like this just needs to overlay the content that comes after it. Thoughts @twbs/css-review?

MartijnCuppens: I think we better set a custom offset in our docs, so that the sticky header is displayed below the sticky header. If we just decrease the `z-index`, the documentation won't be clear because we won't see the sticky table head.

And we should of course document the `$table-head-sticky-top` too.

ed1nh0: > I think we better set a custom offset in our docs, so that the sticky header is displayed below the sticky header. If we just decrease the `z-index`, the documentation won't be clear because we won't see the sticky table head.
Make totally sense. Using `$zindex-sticky` for more than one sticky component could and should end up with conflict. I'm not feeling comfortable to suggest another var name.

mdo: Left you hanging here @ed1nh0. I'd suggest we do `$table-head-sticky-zindex` if you're still down to iterate here.

ed1nh0: > Left you hanging here @ed1nh0. I'd suggest we do `$table-head-sticky-zindex` if you're still down to iterate here.

Yeah sure! But as for syntax standard I'd suggest swap its end, therefore `$table-head-zindex-sticky`, what do you think?

tmorehouse:
If you are interested, here is what we (BootstrapVue) had to do to get sticky headers working with Bootstrap tables, specifically when `<tr>`, `<tbody>`,  `<thead>` and `<tfoot>` have table or background color variants applied to them (and not the cells).

https://github.com/bootstrap-vue/bootstrap-vue/blob/dev/src/components/table/_table.scss#L89-L250

```scss
// --- Table sticky header styling ---

@if $bv-enable-table-sticky {
  .b-table-sticky-header,
  .table-responsive,
  [class*="table-responsive-"] {
    // Move the table bottom margin to the wrapper
    margin-bottom: $spacer;

    > .table {
      // Reset `margin-bottom` to we don't get a space after
      // the table inside the scroll area
      margin-bottom: 0;
    }
  }

  .b-table-sticky-header {
    overflow-y: auto;
    // Annoyingly, when overflow-y is set, browsers convert
    // 'overflow-x: visible' to 'overflow-x: auto' - so it becomes
    // responsive in the x axis automatically
    // Default `max-height` before a scrollbar will show
    // We don't use `height` as table could be shorter than this value
    max-height: $b-table-sticky-header-max-height;
  }

  @media print {
    // Overide any styles (including inline styles)
    // when printing
    .b-table-sticky-header {
      overflow-y: visible !important;
      max-height: none !important;
    }
  }

  @supports (position: sticky) {
    // Positioning of sticky headers
    .b-table-sticky-header > .table.b-table > thead > tr > th {
      // Header cells need to be sticky on top
      position: sticky;
      top: 0;
      z-index: 2;
    }

    // Positioning of sticky columns
    // Sticky columns only work when table has sticky
    // headers and/or is responsive
    .b-table-sticky-header,
    .table-responsive,
    [class*="table-responsive-"] {
      > .table.b-table {
        > thead,
        > tbody,
        > tfoot {
          > tr > .b-table-sticky-column {
            position: sticky;
            left: 0;
          }
        }

        > thead {
          > tr > .b-table-sticky-column {
            // z-index needs to be higher than sticky columns and
            // sticky headers for correct layering
            z-index: 5;
          }
        }

        > tbody,
        > tfoot {
          > tr > .b-table-sticky-column {
            // z-index needs to be lower than sticky header that
            // is also a sticky column
            z-index: 2;
          }
        }
      }
    }

    // Default theme color background for table cells that are sticky
    // Applied only when no variant is applied to the rows, or no head-variant
    // Needed because Bootstrap v4 does not have table child elements set up
    // to inherit their background color from parent element by default
    //
    // An issue made at twbs/bootstrap repo for table
    // background color inheritance:
    //   https://github.com/twbs/bootstrap/issues/29244
    // If implemented, would negate the need for all the below SCSS
    .table.b-table {
      > thead,
      > tbody,
      > tfoot {
        > tr > .table-b-table-default {
          // Default cell color
          color: $table-color;
          // `$table-bg` is null by default in Bootstrap v4 variables
          // but could have a value set by the consumer
          background-color: if($table-bg, $table-bg, $body-bg);
        }
      }

      &.table-dark {
        > thead,
        > tbody,
        > tfoot {
          > tr > .bg-b-table-default {
            // Default cell color in table dark mode
            color: $table-dark-color;
            // Default cell background color in table dark mode
            background-color: $table-dark-bg;
          }
        }
      }

      // Handle case of zebra striping
      &.table-striped {
        // "fake" zebra striping via use of a transparent background image
        > tbody > tr:nth-of-type(#{$table-striped-order}) > .table-b-table-default {
          // `$table-accent-bg` (used for striping) default is a very transparent black
          // We overlay it over the background color to achieve the same color
          // effect while keeping the background solid.
          background-image: linear-gradient($table-accent-bg, $table-accent-bg);
          background-repeat: no-repeat;
        }

        &.table-dark {
          > tbody > tr:nth-of-type(#{$table-striped-order}) > .bg-b-table-default {
            // `$table-dark-accent-bg` (used for striping) default is a very transparent white
            // We overlay it over the background color to achieve the same color
            // effect while keeping the background solid.
            background-image: linear-gradient($table-dark-accent-bg, $table-dark-accent-bg);
            background-repeat: no-repeat;
          }
        }
      }

      // Handle case of hover
      &.table-hover {
        // "fake" hover via use of a transparent background image
        > tbody > tr:hover > .table-b-table-default {
          color: $table-hover-color;
          // `$table-hover-bg` default is a very transparent black
          // We overlay it over the background color to achieve the same color
          // effect while keeping the background solid.
          background-image: linear-gradient($table-hover-bg, $table-hover-bg);
          background-repeat: no-repeat;
        }

        &.table-dark {
          > tbody > tr:hover > .bg-b-table-default {
            color: $table-dark-hover-color;
            // `$table-dark-hover-bg` default is a very transparent white
            // We overlay it over the background color to achieve the same color
            // effect while keeping the background solid.
            background-image: linear-gradient($table-dark-hover-bg, $table-dark-hover-bg);
            background-repeat: no-repeat;
          }
        }
      }
    }
  }
}
```

Setting up background color inheritance for table child elements (tbody, thead, tfoot, tr, td, and th) would have simplified the above CSS greatly, but our CSS is loaded after Bootstrap CSS, so we couldn't define it in our SCSS (as it would have overridden the table/background variant definitions).

tmorehouse:
Also note that sticky header table needs to be wrapped in a containing DIV that has a max-height set, otherwise the sticky header sticks to the top of the page, rather than making the table body scrollable. 

See https://bootstrap-vue.js.org/docs/components/table#sticky-headers for an example of how we got it to work

scss/_tables.scss
tmorehouse: What if the table is wrapped in a vertically scrollable div...  Where the sticky headers should appear under any sticky top navbars when the page is scrolled?

I would expect this (table in a scrollable div) to be the more common use case than having the table header stick to the top of the page.

tmorehouse: We found that a z-index of 2 to 5 works well.

mdo:
Coming back to this again, this needs a different z-index that doesn't overlay our navbars. It's at least overlaying the docs subnav right now, so maybe an issue with the docs.

Would you be willing to update this @ed1nh0?

ed1nh0:
> Coming back to this again, this needs a different z-index that doesn't overlay our navbars. It's at least overlaying the docs subnav right now, so maybe an issue with the docs.
> 
> Would you be willing to update this @ed1nh0?

For sure.
Today the `_variables` file has these `z-index` levels:

```
$zindex-dropdown:                   1000;
$zindex-sticky:                     1020;
$zindex-fixed:                      1030;
$zindex-modal-backdrop:             1040;
$zindex-modal:                      1050;
$zindex-popover:                    1060;
$zindex-tooltip:                    1070;
```

The z-index for `.thead-sticky` is set to `2`.

mdo:
Per @tmorehouse's comments earlier, I think a z-index of 3 or 5 would be enough? We use 1, 2, and 3 the most I think.

ed1nh0:
I've opened a new PR for that.

mdo:
Superseded by #36125.

Issue #29457, opened 2019-09-28
Fix Coveralls Action


XhmikosR:
After we switched to GitHub Actions CI, the branch isn't passed on coveralls.

One way I've been experimenting is using `COVERALLS_GIT_BRANCH: "${{ github.ref }}"` but this results in `refs/head/branch` instead of `branch`, so it needs further manipulation.

There's an official action for coveralls, but it seems it leaves comments even when the setting on coveralls is disabled.

I'll try to find a solution soon.

PR #29458, opened 2019-09-28
coveralls: Add `COVERALLS_GIT_BRANCH`

XhmikosR:
This does pass the ref, but not just the branch name, the whole ref, i.e `refs/head/master`.

This is an short-term fix until we switch to the official action.

Refs #29457


XhmikosR:
I switched to the coveralls action. Works properly, but the build ID is not a number anymore. But a fixed thing is that coveralls now will again run for PRs.

We can remove the secret also from the repo settings.

I'll keep this open for a while, but we should switch to the official action as soon as the issues are fixed upstream.

XhmikosR:
I'm gonna split the action switch to a new PR

Johann-S (approved):

XhmikosR:
@mdo can you check why the merge button is disabled here?

PR #29460, opened 2019-09-28
Trim trailing whitespace from markdown files

MartijnCuppens:
We don't have trailing whitespaces for any markdown file. It's also quite annoying  the trailing whitespaces aren't removed when writing documentation.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,3 @@ indent_size = 2
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
-
-[*.md]
-trim_trailing_whitespace = false


XhmikosR:
LGTM but this is incomplete anyway, since we don't filter the allowed HTML elements (so as to allow `<br>` or not).

I need to check all markdown files just in case though.

MartijnCuppens:
I searched for occurrences of trailing whitespaces in our codebase and it only seems to be present in `clipboard.min.js` (opened https://github.com/webpack/webpack/pull/9756 to tackle this issue upstream). The markdown files should be ok.

XhmikosR (approved):

PR #29466, opened 2019-10-01
Group line-height variables

astrahov:


scss/_breadcrumb.scss
```diff
@@ -1,6 +1,7 @@
 .breadcrumb {
   display: flex;
   flex-wrap: wrap;
+  font-size: $breadcrumb-font-size;
```
MartijnCuppens: Use the `font-size()` mixin here (that's the reason the tests are failing)

scss/_variables.scss
```diff
@@ -1044,6 +1044,8 @@ $figure-caption-color:              $gray-600 !default;
 
 // Breadcrumbs
 
+$breadcrumb-font-size:              $font-size-base !default;
```
MartijnCuppens: You can use `null` for this, this will not output any css by default, but will make it possible to change the font size if needed.

MartijnCuppens:
Could you exclude the commits from #29467 here?

scss/_breadcrumb.scss
astrahov:
```suggestion
  @include font-size($breadcrumb-font-size);
```

astrahov (commented):
fix

MartijnCuppens (approved):

PR #29467, opened 2019-10-01
$breadcrumb-font-size

astrahov:


scss/_breadcrumb.scss
```diff
@@ -1,6 +1,7 @@
 .breadcrumb {
   display: flex;
   flex-wrap: wrap;
+  font-size: $breadcrumb-font-size;
```
MartijnCuppens: Use the `font-size()` mixin here (that's the reason the tests are failing)

scss/_variables.scss
```diff
@@ -1044,6 +1044,8 @@ $figure-caption-color:              $gray-600 !default;
 
 // Breadcrumbs
 
+$breadcrumb-font-size:              $font-size-base !default;
```
MartijnCuppens: You can use `null` for this, this will not output any css by default, but will make it possible to change the font size if needed.

scss/_breadcrumb.scss
```diff
@@ -3,6 +3,7 @@
   flex-wrap: wrap;
   padding: $breadcrumb-padding-y $breadcrumb-padding-x;
   margin-bottom: $breadcrumb-margin-bottom;
+  font-size: font-size($breadcrumb-font-size);
```
MartijnCuppens: You got to use the mixin, so:
```suggestion
  @include font-size($breadcrumb-font-size);
```

MartijnCuppens (approved):

Issue #29442, opened 2019-09-25
Disabled input form control should be without focus status


vraajesh:
![image](https://user-images.githubusercontent.com/7411711/65612566-1f22a080-dfd2-11e9-80d7-8bd27539c24a.png)

From the user experience point disabled / read only input form control should not have focus state, this might confuse the user.

patrickhlauke:
read-only form controls are focusable by default in browsers. forcing them NOT to be focusable would be an opinionated change that goes against browser conventions.

mdo:
Closing per last comment.

PR #29469, opened 2019-10-01
WebKit styles fix for form-control-plaintext

astrahov:


MartijnCuppens:
Bug reports must include a live demo of the problem. Per our [contributing guidelines](https://github.com/twbs/bootstrap/blob/master/.github/CONTRIBUTING.md), please create a reduced test case via [CodePen](https://codepen.io/)/[JS Bin](https://jsbin.com/) or [Stackblitz](https://stackblitz.com/) and report back with your link, Bootstrap version, and specific browser and OS details.

*This is an automated reply*

astrahov:
OS X: 10.14.6
Chrome: 77.0.3865.75
Bootstrap: 4.3.1

1. Click to plaintext field

https://codepen.io/alexstrakhov/pen/jONjxva

MartijnCuppens:
Not sure if we should set an alternative focus state here. @patrickhlauke, what do you think?

florianlacreuse:
See #29442

ysds:
Since accessibility reason, focusable elements should have focus style like blue focus ring. The `:focus-visible` pseudo-class would solve the uncomfortable style of the time the target element is clicked, but it is not yet available. https://caniuse.com/#search=%3Afocus-visible

So closing this.

PR #29470, opened 2019-10-01
Code readability changed

sonam-st:
Multiple selectors in each line improve code readability.


MartijnCuppens:
This code is generated with Sass, so your changes will be overridden in the next build.

PR #29475, opened 2019-10-02
Update button

eashsaxena:


MartijnCuppens:
Please provide more info about in an issue why you want these changes. Enabling underline for buttons doesn't feel like a good idea.

PR #29477, opened 2019-10-02
Restore original js src structure to avoid breaking modularity

### LINKED ISSUE #29471: Recent commit breaks modularity
Ever since this commit: https://github.com/twbs/bootstrap/commit/114a0ee193ffe7869c8369f25ef4dcc44929fbf5#diff-5160056bcd6d0f990ba7775663f6d27f

I have been unable to include bootstrap components in a modular way.

Instead of `require('bootstrap')` which includes every single module contained in bootstrap, I just want the modal: `require('bootstrap/js/dist/modal.js')` (or perhaps there's a better include I'm supposed to be using? I couldn't find it)

I'm using encore webpack, but it's unable to find the dependencies in `js/dist/modal.js` or the src one after this commit.

Johann-S:
- Restore old js src structure to fix the broken modularity
- Added an integration test to be sure to not broke the modularity again

Fixes: #29471


XhmikosR (approved):
Since tests pass, we should be OK

Johann-S:
@XhmikosR we'll have to do another build of our dist after the merge of this PR

PR #29478, opened 2019-10-02
Switch to the Coveralls Action

### LINKED ISSUE #29457: Fix Coveralls Action

XhmikosR:
This is the official way of using Coveralls with Actions. Also this brings back Coveralls for PRs.

~~Requires #29458 to be merged.~~

Fixes #29457


mdo:
> * Remove the repo secret (`COVERALLS_REPO_TOKEN `) before merging this

Will that break the new integration? If not, let's keep tests from failing and remove afterwards maybe?

XhmikosR:
This will only break for PRs or branches that might still use the old way. For master we won't have any problem. For v4-dev, I will cherry pick it in #29482 after this PR is merged.

XhmikosR:
We are also getting rid of one devDependency BTW.

XhmikosR:
Ping @mdo

XhmikosR:
There's currently a [bug](https://github.com/coverallsapp/github-action/issues/5) which I really hope gets fixed soon because we end up with comments on PRs.

I've already switched to this action in a few of my other projects, it works well apart from the aforementioned issue.

I guess we can live for a while with a few comments?

XhmikosR:
@mdo I really want to move forward with this. Can you get someone from GH to look into helping upstream to solve the extra comments issue https://github.com/coverallsapp/github-action/issues/5?

XhmikosR:
This is blocked until the extra comments issue is solved

XhmikosR:
The comments issue seems [fixed](https://github.com/coverallsapp/github-action/issues/5#issuecomment-553555754) so it should be OK to switch to this.

XhmikosR:
@mdo you can remove the `COVERALLS_REPO_TOKEN` secret.

Johann-S (approved):

PR #29481, opened 2019-10-02
Add `semantic` class for reduced markup

jonathanhefner:
By design, Bootstrap supports opt-in styling via explicit classes, rather than implicit styling via semantic HTML.  Opt-in styling is more robust, because implicit styling can be difficult to undo / opt out of, and may cause conflicts with third-party code.  However, styling entirely via explicit classes can result in more verbose markup.  The `semantic` class provides a middle ground between the two approaches, enabling implicit styling in limited scopes.  It changes nothing about the way Bootstrap currently works, but adds a way for users to style their content via semantic HTML.

For example, styling breadcrumbs via explicit classes:

```html
<ol class="breadcrumb">
  <li class="breadcrumb-item"><a href="#">Home</a></li>
  <li class="breadcrumb-item"><a href="#">Library</a></li>
  <li class="breadcrumb-item active" aria-current="page">Data</li>
</ol>
```

Versus styling breadcrumbs via the `semantic` class:

```html
<ol class="breadcrumb semantic">
  <li><a href="#">Home</a></li>
  <li><a href="#">Library</a></li>
  <li class="active" aria-current="page">Data</li>
</ol>
```

The list of semantic implication rules in this commit is small.  It is intended to grow with community input.

---

TODO:
- [ ] docs?
- [ ] tests?
- [ ] relocate Sass variables?
- [ ] additional implication rules?

### DIFF
--- a/scss/_semantic.scss
+++ b/scss/_semantic.scss
@@ -1,0 +1,30 @@
+$semantic-selector: ".semantic" !default;
+
+$semantic-implications: () !default;
+// stylelint-disable-next-line scss/dollar-variable-default
+$semantic-implications: map-merge(
+  (
+    ".blockquote > footer": ".blockquote-footer",
+    "blockquote > footer": ".blockquote-footer",
+    ".list-inline > li": ".list-inline-item",
+    ".figure > img": ".figure-img",
+    "figure > img": ".figure-img",
+    ".figure > figcaption": ".figure-caption",
+    "figure > figcaption": ".figure-caption",
+    ".breadcrumb > li": ".breadcrumb-item",
+    ".list-group > li": ".list-group-item",
+  ),
+  $semantic-implications
+);
+
+
+@each $selectors, $implied in $semantic-implications {
+  @each $selector in $selectors {
+    $i: str-index($selector, " ") or -1;
+    $selector: str-insert($selector, $semantic-selector, $i);
+
+    #{$selector} {
+      @extend #{$implied};
+    }
+  }
+}
--- a/scss/bootstrap.scss
+++ b/scss/bootstrap.scss
@@ -52,3 +52,7 @@
 // Utilities
 
 @import "utilities/api";
+
+
+// Semantic styling
+@import "semantic";


mdo:
What you're after isn't about semantic vs explicit, it's strictly about implicit vs explicit.

Both options are 100% semantically correct—you have an list of items who's order matters. This is the correct markup regardless, and the selectors created for it—whether classes or elements—carry no semantic value. HTML has semantic value, CSS does not.

This is purely a difference in implementation and project-wide approach. To us, based on our experiences, an explicit set of classes is the wiser move for writing frontend code. Yes, there are options to reduce the number of classes, but flipping between them seems like it'd only add confusion for folks.

MartijnCuppens:
I agree with @mdo. We could provide options to change our code methodologies, but it ain't worth the additional complexity we'll introduce in our codebase.

jonathanhefner:
> What you're after isn't about semantic vs explicit, it's strictly about implicit vs explicit.

Yes, I was trying to convey explicit (which could be done with all `div` tags, if desired) vs implicit (which would only be done with semantic HTML).  I'm sorry that my post was worded poorly.  If the class was renamed from `semantic` to `implicit`, would that make this proposal worth more consideration?  Otherwise, could you expand on "there are options to reduce the number of classes"?

To be clear, I think it's valuable that Bootstrap can style any given markup, using explicit classes, because that ensures it can work with any third-party libraries.  But, on the other side of the coin, (optionally) fewer classes would make Bootstrap plug-and-play with many third-party libraries which generate semantic HTML.

rafalp:
What you are proposing is how Bootstrap before v4 worked: you've added, say, `dropdown-menu` to `ul` and all `ul li a` turned into a nicely styled dropdown menu links... but if you wanted to include `li` with something else, you've had to write extra CSS to reset styles that `dropdown-menu` added. If you wanted to reuse styling elsewhere, you had to copy-paste it over. It may not see like much, but this stuff was becoming tiresome real fast, so people started asking for more explicit approach that would make it easier to customize, reuse and extend Bootstrap.

jonathanhefner:
@rafalp I believe this proposal is different than how Bootstrap previously worked, because here the cascading styles would be opt-in via the `semantic` class.  In other words, if you don't want `ul > li` automatically styled, simply don't include the `semantic` class on the `ul`.  (No need to do a reset.)

PR #29482, opened 2019-10-02
V4 dev backports and updates

XhmikosR:


XhmikosR:
It seems BrowserStack tests fail. I'm gonna restart them but if it persists we need to figure out why they fail.

MartijnCuppens (approved):

XhmikosR:
@Johann-S can you check why Safari keeps failing on BrowserStack?

```
Mobile Safari 11.0.0 (iOS 11.0.1) modal should restore focus to toggling element when modal is hidden after having been opened via data-api FAILED
	toggling element is once again focused
	Expected: true
	Actual: false
	
............
Mobile Safari 11.0.0 (iOS 11.0.1) modal should not restore focus to toggling element if the associated show event gets prevented FAILED
	focus returned to toggling element
	Expected: true
	Actual: false
```

XhmikosR:
@Johann-S I pinpointed the test failures to qunit 2.9.3. Can you check if we need to adapt those 2 tests and the changelog https://github.com/qunitjs/qunit/compare/2.9.2...master

For the time being I locked qunit to 2.9.2.

Johann-S:
I don't see anything special in the new release of QUnit 😲 that's pretty weird our unit tests failed because of that

XhmikosR:
I just pushed qunit to 2.9.3. Let's see if it will fail again.

XhmikosR:
Yup, definitely fails @Johann-S

PR #29483, opened 2019-10-02
IE image fix

### LINKED ISSUE #29356: Fix homepage bootstrap-stack.svg display on IE
So, apparently, this is broken for quite some time now :)

It seems it first broke on https://github.com/twbs/bootstrap/commit/cb2b21bb88604dee22bd4d0f2804d8366b2d65e3

The issue is present on the current live site, v4-dev and master branches. For the master branch it seems removing `align-items-start` from the parent row fixes the issue (plus it seems the class is redundant anyway).

For v4-dev, so far, I have found 2 ways to solve this:

1. specify the height in the CSS for >=768px, and width set to auto
2. remove the `img-fluid` class from the SVG

This is how the live site https://getbootstrap.com/ looks like right now:

/CC @MartijnCuppens @ysds for thoughts. I'll make a PR for master to remove the `align-items-start` class anyway.

MartijnCuppens:
Fixes #29356

`v4-dev` implementation of https://github.com/twbs/bootstrap/issues/29356

This aligns the image to the top for tabletish devices, which looks fine imo.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -4,7 +4,7 @@ layout: home
 
 <main class="bd-masthead" id="content" role="main">
   <div class="container">
-    <div class="row align-items-center">
+    <div class="row">
       <div class="col-6 mx-auto col-md-6 order-md-2">
         {% include icons/bootstrap-stack.svg width="512" height="430" class="img-fluid mb-3 mb-md-0" %}
       </div>


XhmikosR:
I didn't want to go this way in order to keep this the same, but it's @mdo's call.

mdo (approved):
Looks good to me!

Screen Shot 2019-10-07 at 10 19 30 AM

PR #29489, opened 2019-10-03
v5 docs: example shortcode regex

### LINKED ISSUE #29351: example shortcode: see if we can use a regex
https://github.com/twbs/bootstrap/blob/919b526f17a08cefba51c42b831c922adecb4a9f/site/layouts/shortcodes/example.html#L35-L38

The above basically extracts the class, removes `bd-placeholder-img` and `bd-placeholder-img-lg`.

I wonder if we can use `findRE` along with `replaceRE` to achieve the same thing.

Not a huge deal, it just feels like a more elegant solution.

Similarly we could try doing the same for:

https://github.com/twbs/bootstrap/blob/919b526f17a08cefba51c42b831c922adecb4a9f/site/layouts/shortcodes/example.html#L27-L29

This just extracts the `<svg></svg>`.

XhmikosR:
example shortcode: see if we can use a regex

Fixes #29351

Preview: <https://deploy-preview-29489--twbs-bootstrap.netlify.com/>

/CC @morrissey-ingenious to review the last patch.


XhmikosR:
I think it would help if we added a link to a https://regex101.com/ example as a comment

XhmikosR:
@morrissey-ingenious your thoughts about the last patch?

XhmikosR:
@morrissey-ingenious can you review my last patch?

morrissey-ingenious (approved):
I'm fine with these updates.  Removing some of the items I had in place makes the regex slightly less robust, but as long as the docs are written cleanly then there won't be an issue.

morrissey-ingenious (approved):

morrissey-ingenious (approved):

morrissey-ingenious (approved):

morrissey-ingenious (approved):

XhmikosR:
I guess I'm gonna merge this as is.

Johann-S (approved):

Issue #26892, opened 2018-07-16
custom-checkboxes not visible whith high contrast on windows 


liyokuna:
Hello,

We discovered an issue about the custom checkbox and radio button when using a different contrast.
On basic mode, we can see [this](https://getbootstrap.com/docs/4.1/components/forms/#checkboxes), image below. ![forms-checkboxes-boostrap](https://user-images.githubusercontent.com/11217936/42750721-83f4fdec-88e8-11e8-8a12-4b842c643b45.PNG)

When changing the contrast and pick in my case the second contrast on windows, as shown on the image below.
![contraste-picked-2](https://user-images.githubusercontent.com/11217936/42750809-df16ae3c-88e8-11e8-8ca9-ae4ecf59c264.PNG)

The result of this change affects a lot the custom checkbox and when can't see anything at all, as you can see on the image below. 
![forms-checkboxes-boostrap-contrast](https://user-images.githubusercontent.com/11217936/42750843-01917690-88e9-11e8-9ac8-e2a5a23ef7f3.PNG)

Just by following those steps, you can reproduce that issue.

My compagny uses a fork of Bootstrap, named [Boosted](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap), and a colleague raises this issue, [here's the link](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/87).

Plenty of idea crossed my mind, the first one is to create a class checked and it will be manageable by JS. In this  JS file, we can checked the contrast by using Modernizr. [As this example shows the use of Modernizr with contrast](https://www.ffoodd.fr/un-test-modernizr-pour-les-contrastes-eleves/)
The second is to edit the css class in the custom-forms.scss.

If you have another idea to solve it, share it :) !

- Operating system and version (Windows 7)
- Browser and version (Firefox 52.9.0 , IE 11)

liyokuna:
Hello,
I solved by editing the CSS classes used for checkboxes and radio button. Classes using attributes :before and :after.
I set the size border for the border and the tick for those two different elements. 
the result is this :

![](https://user-images.githubusercontent.com/11217936/42954408-ea5cd53c-8b7c-11e8-8642-efffd3ea4989.PNG)

mdo:
Confirming these are fixed, I see borders and background images in our docs in high contrast. This was addressed at some point when we added borders to the form elements. Closing out.

Issue #21269, opened 2016-12-01
SVG background icons not shown in Windows High Contrast Mode


patrickhlauke:
successor to https://github.com/twbs/bootstrap/pull/21268 for v4. while the way icons (such as the hamburger menu) are coded is different (SVG background rather than 3 empty spans), these backgrounds (likely in all other places where SVG backgrounds are currently used) still disappear in WHCM.

<img width="960" alt="capture" src="https://cloud.githubusercontent.com/assets/895831/20806645/219a421a-b7fc-11e6-9d92-75f8e48d600d.PNG">

joe-watkins:
Tnx for firing this up @patrickhlauke :) 

Are foreground `.svg` images an option for critical controls like this?

patrickhlauke:
they'll probably have to be. or we provide some high contrast media query stuff https://msdn.microsoft.com/en-us/library/windows/apps/hh465764.aspx or similar...

joe-watkins:
CSS generated content with the media query could be an option if foreground images aren't possible. May be too opinionated for a generic control.. maybe a warning comment in the docs would do?

``` CSS
@media screen and (-ms-high-contrast: active) {
    .navbar-toggler {
        width: auto;
    }
    .navbar-toggler:after {
        content: "menu";
    }
}
```

![mq-hack](https://cloud.githubusercontent.com/assets/3695795/20846863/783725dc-b880-11e6-8bb9-60f5b5d7d4af.png)

patrickhlauke:
Edging towards a possible solution. First problem is that SVG colors are not automagically inverted when in WHCM. This then results in semi-transparent black hamburger icon on a black background.

http://codepen.io/patrickhlauke/pen/KNoNwd

patrickhlauke:
With some noodling, I tried moving the stroke-related styles to an actual `<style>` block in the SVG, and adding an explicit `-ms-high-contrast` media feature query. The result is semi-promising (though IE refuses to show any background images in WHCM, so a different approach, like CSS generated image content, would have to be used) http://codepen.io/patrickhlauke/pen/WozRyM

patrickhlauke:
![capture](https://cloud.githubusercontent.com/assets/895831/20890076/881b7708-bafd-11e6-9610-f89db8774126.PNG)

patrickhlauke:
more progress using a subtly different approach...but this fails annoyingly in Firefox http://codepen.io/patrickhlauke/pen/bBMbXp

[...]

PR #29490, opened 2019-10-04
badge accessible in high contrast mode

julians300:
This PR gives a transparent border to badges so they give better context in high contrast mode.

Before

Screen Shot 2019-10-03 at 10 20 08 PM
Screen Shot 2019-10-03 at 10 20 34 PM

After

Screen Shot 2019-10-03 at 10 20 14 PM
Screen Shot 2019-10-03 at 10 20 38 PM

Preview: <https://deploy-preview-29490--twbs-bootstrap.netlify.app/docs/5.0/components/badge/>


MartijnCuppens:
This will indent each badge with one pixel. Maybe there's any way to detect when we're in high contrast mode?

patrickhlauke:
these could be scoped to just `-ms-high-contrast:active` media query - would still result in the indent, but only in HC mode.

note that Windows HC is quite unique in that it munges/reinterprets/overrides a site's CSS. also, HC likely doesn't currently work with a lot of other components (not tested, but off the top of my head thinking any fancy checkbox/radio button/form control constructs that use clever "faked" CSS / rely on background images and borders. https://github.com/twbs/bootstrap/issues/26892 + https://github.com/twbs/bootstrap/issues/21269

related https://developer.paciellogroup.com/blog/2016/12/windows-high-contrast-mode-the-limited-utility-of-ms-high-contrast/ and http://www.gwhitworth.com/blog/2017/04/how-to-use-ms-high-contrast/

XhmikosR:
Unless we do this across the codebase, I don't think it makes sense to do it just for one component.

MartijnCuppens:
> This will indent each badge with one pixel. Maybe there's any way to detect when we're in high contrast mode?

Seems I was wrong about this, the border just adds a px.

I'm ok with this change, as long as we add a comment in the source code since this is not self-explanatory.

mdo:
I'm inclined to leave this or scope it to high contrast mode only. Could be skipped entirely if we figure out the right theme/color scheme approach.

patrickhlauke:
adding some more resources:

* https://sarahmhigley.com/writing/whcm-quick-tips/
* https://www.w3.org/2019/09/Meetup/speaker-melanie.html

would be good to have a run through the entirety of Bootstrap to see what needs to be tweaked. it should be possible to make things work ok in high contrast without affecting the non-high-contrast look and feel

Issue #26716, opened 2018-06-16
New 'Content' component feature


sts-ryan-holton:
How many times have you come across a bunch of content within a `<div>` or an `<article`> tag with a bunch of `<p>`, `<ul>` and possibly even **headings** and thought to yourself if only there was a simple **classname** I can add to the div that would prevent me having to add specific classes to all of these elements?

This is an issue I've come across, and whilst I'm sure Bootstrap provides some base default styling, I'm not totally convinced that this is appropriate in most projects.

I was thinking this...

To add a new component under the **'Components'** heading on the docs (https://getbootstrap.com/docs/4.1/getting-started/introduction/) called something like: **'Content'**

The purpose of this component is to provide a simple method for styling elements within a div or article element, all the user would need to do is add a class called `.content` to their parent element, and elements within would be styled. **I was thinking of adding support for the following HTML elements:**

- `<p>` paragraphs
- `<ul>`, `<ol>` and `<dl>` lists
- `<h1>` to `<h6>` headings
- `<blockquote>` quotes
- `<em>`, `<b>` and `<strong>` tags
- `<table>` with _relevant child elements of a table_

What's people's opinion on this?

herbalite:
I've used jquery-ui before and am right now adapting a project to bootstrap. So much where I am coming from with my answer. jquery-ui does what you suggested.

If that component feature should be implemented, make it an entirely optional plug-in. 

In short there is too much JS overhead involved, losing design flexibility as well. Also there are always CSS hacks needed to override certain features. Usually I change the component (often omit something, that a JS component would introduce) to accommodate the projects needs

And if JS is off then things look worse, because nothing has the required CSS classes. A hard coded version still looks decent. And then is the issue if something changes.

A better way IMO is to make code pieces of common Bootstrap HTML components in ones preferred editor. Copy the component in, and add the specifics. If things change, change the template. Everything else I found when maintaining projects is too messy.

sts-ryan-holton:
@herbalite I think you’ve gotten a little confused with what I’m suggesting here.

creating a CSS class for content wouldn’t require any JS at all, not quite sure whether youve got confused with something else

MartijnCuppens:
@sts-ryan-holton "I think" what you're suggesting is something like Bulma is doing here: https://bulma.io/documentation/elements/content/

The Bootstrap docs also use a `.bd-content` class to control spacing between headings:
[...]

PR #29491, opened 2019-10-04
Indent the last element, for example

alexeymelekhov-dev:
Indent the last element, for example, for the `<p>` tag

### DIFF
--- a/scss/_alert.scss
+++ b/scss/_alert.scss
@@ -8,6 +8,10 @@
   margin-bottom: $alert-margin-bottom;
   border: $alert-border-width solid transparent;
   @include border-radius($alert-border-radius);
+  
+  > *:last-child {
+    margin-bottom: 0;
+  }
 }
 
 // Headings for larger alerts
--- a/scss/_alert.scss
+++ b/scss/_alert.scss
@@ -8,7 +8,7 @@
   margin-bottom: $alert-margin-bottom;
   border: $alert-border-width solid transparent;
   @include border-radius($alert-border-radius);
-  
+
   > *:last-child {
     margin-bottom: 0;
   }


MartijnCuppens:
I was thinking for a more global approach on this. We could add for example a `.content` class in which the margins of the last elements (`ul`, `p`,...) are removed:
```
.content {
  > :last-child {
    margin-bottom: 0;
  }
}
```
This class could then be used for as well cards, alerts or other components.
Toughts @twbs/css-review?

mdo:
@MartijnCuppens Yeah, something like that for more typographic control (larger `margin-top` on headings, enhanced lists, etc) could be really cool. I think there was a PR or issue somewhere, maybe from you even lol?

I'm unsure about a general content class yet for removing bottom margins though—that doesn't really feel like a content thing. That feels like a common product/UI issue more than a (marketing, editorial, blog) content thing.

MartijnCuppens:
> I think there was a PR or issue somewhere, maybe from you even lol?

We once talked about that here: https://github.com/twbs/bootstrap/issues/26716. Bulma has a `.content` class which adds top margins and prevents the last element from adding an additional bottom margin: https://bulma.io/documentation/elements/content/#sizes

> I'm unsure about a general content class yet for removing bottom margins though—that doesn't really feel like a content thing. That feels like a common product/UI issue more than a (marketing, editorial, blog) content thing.

I do use a custom `.content` class a lot myself, mostly for just removing the bottom margin if I can't add a utility class to the last `<p>` or `<ul>` tag. I wasn't sure about the name either (also used `.body` in the past), but I haven't found a better name so far.

mdo:
Per the conversation here, let's punt this to a separate effort. Arbitrary margin changes like this are a recipe for unintended consequences. We should avoid them where possible.

PR #29495, opened 2019-10-05
Upgrade few transitive dependencies

rmkanda:
## Upgrade few transitive dependencies
Some of the transitive dependencies referred are not latest. Hence updated to the latest version.


XhmikosR:
Thanks but this is already in #29447

PR #29496, opened 2019-10-05
Drop support for Node.js 8.

XhmikosR:
By the time v5 stable will be out, 8 will be EOL (December 2019).

TODO:

- [x] Adapt master protection rules before merging this


XhmikosR:
Notes:

1. in order to make Netlify use Node.js 10, we need to use a `.nvmrc` file. Adding this file might have other implications which I'm not 100% sure since I don't use nvm myself
2. we need to remove Node.js 8 from the master protection rules before merging this

XhmikosR:
I dropped the .nvmrc addition. Maybe netlify will properly read the engines version after this is merged on master. If not we add the .nvmrc file then.

mdo (approved):

XhmikosR:
@mdo: if we merge this, I need Node.js 8 removed from the master branch protection rules :)

@Johann-S: I assume you agree?

mdo:
Done!

Johann-S (approved):

PR #29498, opened 2019-10-05
Addition to unacceptable behavior

nabihahzaheer:
Added another point to unacceptable behavior to ensure more clarity


mdo:
Code of conduct is from another project, Code Covenant. Please submit patches there so everyone can review and ours doesn’t become mismatched.

PR #29501, opened 2019-10-08
Fix sass backward dependency

Logicus03:
Fix Sass Backward dependency issue on _root.scss file


MartijnCuppens:
Hi @Logicus03 

What does this solve exactly? Also, the generated CSS should not be included in a PR.

mdo:
I believe some compilers have a hard time with the CSS custom properties, but not something we focus in on given Libsass and our compilers work as-is.

Logicus03:
After install bootstrap in an Angular projects, some issues were solved after modify your files (scss files) and the reason was a problem with latest sass backward dependency. I made the changes and share with you.

PR #29503, opened 2019-10-08
Rename "js/tests/units" to "js/tests/unit".

XhmikosR:
We could probably renamed `tests` to `test` too, but I'm not sure about that.


Johann-S:
No opinion on that but I would love to move `helpers` in (currently) `units`

Johann-S (approved):

PR #29509, opened 2019-10-09
Netlify: force Node.js 10 version via .nvmrc

### LINKED ISSUE #29507: Netlify: use Node.js 10 for master
It seems it keeps using version 8 for some reason.

XhmikosR:
Not sure if there's a better way. It doesn't seem we can do this via an environment variable.

Fixes #29507 

BTW another approach I experimented with was using `$npm_package_engines_node`. This works as long as the node version isn't a range, i.e. if it is `10` it works fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ Thumbs.db
 # Folders to ignore
 /js/coverage/
 /node_modules/
+/.nvmrc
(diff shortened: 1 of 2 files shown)


XhmikosR:
Actually, I think I made it work with `$npm_package_engines_node`

XhmikosR:
Actually it seems we can just use `NODE_VERSION`. Closing this.

Issue #26704, opened 2018-06-12
Feature Request: Add animation to modal when its backdrop is static.


webframing:
When the modal has the property:

```javascript
{backdrop: 'static', keyboard: false}
```

Or when it has the attributes:

```html
data-backdrop="static" data-keyboard="false"
```

Obviously, it would not be close if the user tries to click outside of it and also using <kbd>esc</kbd> key. Then it should show the user that anything is important in modal (user needs to check modal) by showing any animation like `scale` to get the user attention there. Like the chrome's clear browser data modal:

![ubhyfdbl4s](https://user-images.githubusercontent.com/39026124/41271037-e80f45bc-6e2b-11e8-9c55-8a0da0cd725d.gif)

bkdotcom:
What if bootstrap fired a `hidePrevented.bs.modal` event?
You could then control what happens...   perhaps bootstrap could do the animation unless you call `preventDefault()` on the event

webframing:
Yes! It may a better solution to handle this situation more easily. Then user will have more options to animate the modal. Like using `animate.css`.

rjoly:
@bkdotcom this kind of event should be great to do whatever the developer wants to give any feedback.

florianlacreuse:
Agree. In applications with both static and non-static modals, user may be confused with these two different behaviors without feedback (on hide prevented).

It looks like we could fire the event from here: 

https://github.com/twbs/bootstrap/blob/32ab52ba2e9dc2549304fcef14dede7806384f34/js/src/modal.js#L355-L359

florianlacreuse:
@Johann-S I just wanted to know if there was a slight chance of getting this new feature for v5? Just asking, I know you guys are busy with this next major release.

Johann-S:
Yep why not, it seems useful to me, btw we are open for PRs 😄

anjoshigor:
Can I try to resolve this issue?

magicmb:
@webframing I'm confused about the status of this request. It's still showing "Assignees: No one assigned" on v4.4.0 To do list yet it seems that the request has moved along since and appears resolved now by @anjoshigor   #29516

XhmikosR:
This is in master and will be in v4.4.0 too.

PR #29516, opened 2019-10-12
Added animation when modal backdrop is static

anjoshigor:
# The problem
When the modal backdrop is static, it should not hide when user clicks outside it or press escape key. But the current behavior wasn't in agreement with this requisite.

more info on issue #26704

# The solution
+ When the event `KEYDOWN_DISMISS` is fired, the static property is verified and another event `HIDE_PREVENTED` is fired.
+ The event, if not prevented, is handled and a 1.2 scale transition is performed.
+ If anyone wants to ovewrite this behavior, he/she should listen to the new `HIDE_PREVENTED` event and handle it accordingly.

# Evidence
backdrop

Any doubts and suggestion, please contact! 🤘 

P.S: I'm planning to update doc in another PR.

Preview: <https://deploy-preview-29516--twbs-bootstrap.netlify.com/docs/4.3/components/modal/#static-backdrop>


MartijnCuppens:
Hi @anjoshigor,

Thanks for this contribution. I think we better add an example of this behaviour to our docs, just to make it clearer.

I would also make the classname less attached to the appearance. Something like "highlighted" or "focused" would be more appropriate, although those names aren't perfect yet. Open for suggestions here.

XhmikosR:
We definitely need the docs updated in this PR.

js/src/modal.js
```diff
@@ -307,8 +309,22 @@ class Modal {
     if (this._isShown && this._config.keyboard) {
       EventHandler.on(this._element, Event.KEYDOWN_DISMISS, event => {
         if (event.which === ESCAPE_KEYCODE) {
-          event.preventDefault()
-          this.hide()
+          if (this._config.backdrop === 'static') {
+            const hideEvent = EventHandler.trigger(this._element, Event.HIDE_PREVENTED)
```
Johann-S: Can you create a separate function for your `if` condition, btw it seems it's the same code used a few lines later, so it should be better to just call your function two times

Johann-S (changes requested):
Thanks @anjoshigor for your work 👍 

Just one comment to fix, and yes we need the docs in this PR 😄

rohit2sharma95:
@MartijnCuppens What about if the class name would be **static** or **modal-static**. Because the class is only added when the modal is static. I am just getting inspiration from the class **modal-open** that is added in the body, when the modal is open.

MartijnCuppens:
`.modal-static` would be better, thanks for the suggestion @rohit2sharma95!

florianlacreuse:
Thanks @anjoshigor, that was a feature asked by many people (including me)!

From what I see through the gif, clicking inside the modal will trigger the hide prevented animation? It seems weird to me at first sight. I would expect to have the animation only if I click in the modal backdrop area.

js/src/modal.js
anjoshigor: Sure thing! 👍

anjoshigor:
@florianlacreuse actually, the gif is kind of misleading. When the cursor is inside the modal, I press escape and that's why the animation is triggered 😅

@MartijnCuppens @rohit2sharma95 I'll use `modal-static` as classname 👍 

@XhmikosR @Johann-S I'll add an explanation to the docs, thanks!

florianlacreuse:
@anjoshigor Oh ok, good then! Thanks for your work.

XhmikosR:
@Johann-S shouldn't we backport this to v4-dev?

rohit2sharma95:
@anjoshigor I just have seen the deploy preview [here](https://deploy-preview-29516--twbs-bootstrap.netlify.com/docs/4.3/components/modal/#events):-
> This event is fired when the modal is shown, its backdrop is static and a click outside the modal or a scape key press is performed.

In the above quotation **scape** should be **escape**  😊

Johann-S:
Yep I think we can do that 👍

XhmikosR:
It won't apply clean probably, and we need to change any Hugo specific code to Jekyll, but it's doable. Feel free to approve and we tackle the backport in another PR.

Johann-S (changes requested):
Can you remove the JS comments? The code is pretty clear here 👍 
And it'll be good for your PR

anjoshigor:
@Johann-S removed the comments, could you please accept the PR now? 🙏

Johann-S (approved):

scss/_modal.scss
```diff
@@ -48,6 +48,11 @@
   .modal.show & {
     transform: $modal-show-transform;
   }
+
+  // When trying to close, animate focus to scale
+  .modal.modal-static & {
```
mdo: Only nit is that this could be simplified to `.modal-static &` as long as we're specifying where it goes in the docs. `.modal.show` is chained like that because `.show` is used throughout the project.

scss/_modal.scss
```diff
@@ -48,6 +48,11 @@
...
+    transform: scale(1.02);
```
MartijnCuppens: We should make this value variable.

XhmikosR:
@anjoshigor can you take care of the above comments please?

scss/_modal.scss
ysds: Unfortunately, the `.modal` (or any class selector) is required for override the above transform definition: 

image

anjoshigor:
@XhmikosR just pushed some changes. You can take a look to see if it fulfils all requirements

MartijnCuppens (approved):

mdo (approved):
Seems cool!

site/content/docs/4.3/components/modal.md
```diff
@@ -859,6 +918,10 @@ Bootstrap's modal class exposes a few events for hooking into modal functionalit
       <td>hidden.bs.modal</td>
       <td>This event is fired when the modal has finished being hidden from the user (will wait for CSS transitions to complete).</td>
     </tr>
+    <tr>
+      <td>hidePrevented.bs.modal</td>
+      <td>This event is fired when the modal is shown, its backdrop is <code>static</code> and a click outside the modal or a scape key press is performed.</td>
+    </tr>
```
rohit2sharma95: On line 923, It should be **or an escape keypress is performed**
:)

XhmikosR: Fixed in #29621 but please remember you can always submit a PR yourself

PR #29521, opened 2019-10-13
Set vertical-align on .form-check-input

jonathanhefner:
This is a follow-up to #29322, specifically "allowing easy placement of the `.form-check-input` in more places."  In my testing in Firefox and Chrome, setting `vertical-align: top` removes the need to add `float-left` to vertically align the checkbox and label text.  And doing so does not visually affect examples which do apply `float: left` via nesting in a `.form-check`.


mdo (approved):
LGTM. Anything from @twbs/css-review?

MartijnCuppens (approved):
Nope, LGTM!

PR #29522, opened 2019-10-14
The HTML5 Boilerplate Print CSS moved to a separate repo

coliff:
The HTML5 Boilerplate CSS has moved to a separate GitHub repo (h5bp/main.css) so the source link referenced here no longer works. This PR updates the link to the new correct location.


mdo:
@twbs/css-review I believe this would be a v4 fix and not for v5, correct? We're dropping nearly all print styles in v5 I think...

coliff:
yep, print styles aren't in v5 so I made this PR against the `v4-dev` branch

XhmikosR (approved):

Issue #29405, opened 2019-09-13
keydown navigation in dropdown broken if there are hidden items


msh-odoo:
Bootstrap v4.3.1

When navigating on dropdown using keyboard and if there are hidden(display: none) dropdown item in between then keyboard navigation stops.

Issues seems from: ```_dataApiKeydownHandler``` method, it doesn't skip hidden element, it directly set focus on next element while next element can be hidden and focus does not set on hidden element and navigation gets broken.

Johann-S:
Bug reports must include a live demo of the problem. Per our [contributing guidelines](https://github.com/twbs/bootstrap/blob/v4-dev/.github/CONTRIBUTING.md), please create a reduced test case via [CodePen](https://codepen.io/)/[JS Bin](https://jsbin.com/) or [Stackblitz](https://stackblitz.com/) and report back with your link, Bootstrap version, and specific browser and OS details.

*This is an automated reply*

msh-odoo:
@Johann-S here is the CodePen link: https://codepen.io/mishekha/pen/ExYREam

Open dropdown and press Down key, focus goes to first item and then press Down key again, it doesn't go to next item in dropdown, it is because we have ```display: none``` item in between.

Bootstrap version I am using: 4.3.1, Browser Chrome: 76.0.3809.132 OS: macOS Mojave.

Hope this will help, let me know if you need further details.

Johann-S:
it seems a normal behavior here 🤔 I think it's for an accessibility reason, maybe @patrickhlauke have some hints on that

patrickhlauke:
@Johann-S that's broken behavior: if an item isn't there (as `display:none` effectively removes it both visually and from the accessibility tree), it should not receive focus (or attempt to receive focus). the script needs to be modified to be more discerning about only trying to attempt to go to items that are actually there...or @msh-odoo should remove items completely from the DOM tree, rather than just hiding them via CSS.

Johann-S:
Hmm I see 🤔 we have a util helper to check `isVisible` we should use it

msh-odoo:
@patrickhlauke @Johann-S I can not remove item from dropdown in my case, I did workaround fix to put all hidden items at last so that navigation works but it should be fixed in dropdown itself, before going to next item it should check whether that item is visible or not, if it is not visible then focus is not possible on that item so we should go for next item and so on.

msh-odoo:
@Johann-S @patrickhlauke Any news on this?

[...]

PR #29523, opened 2019-10-14
Skip hidden dropdowns while focusing

jeremyvii:
This PR fixes issue #29405 by filtering out dropdown items set to `display: none` when focusing between items. The `isVisible()` utility method filters out the hidden items.


patrickhlauke:
not had a chance to test, but i'll just throw out that you might also need to check for visible, but disabled, ones as well

jeremyvii:
I think disabled items are ignored per this dropdown selector [here](https://github.com/twbs/bootstrap/blob/master/js/src/dropdown.js#L67), but I will double check.

Johann-S (changes requested):
Hi @jeremyvii,

Thanks for your PR!

jeremyvii:
While adding this unit test, I had noticed that `isVisible()` doesn't account for the `d-none` class. I added a check for this class in `isVisible()`.

jeremyvii:
I was just reviewing the change I made, and think it may be better to use `getComputedStyle()` for checking if an element is hidden. Rather than checking for the `d-none` class explicitly. That way if a user has some rouge class that hides an element it will still be skipped while focusing between items.

js/tests/unit/dropdown.spec.js
```diff
@@ -1292,6 +1292,43 @@ describe('Dropdown', () => {
...
+
+      const triggerDropdown = fixtureEl.querySelector('[data-toggle="dropdown"]')
+      const dropdown = fixtureEl.querySelector('.dropdown')
+
+      dropdown.addEventListener('shown.bs.dropdown', () => {
+        const keyDown = createEvent('keydown')
+        keyDown.which = 40
+
+        triggerDropdown.dispatchEvent(keyDown)
+        triggerDropdown.dispatchEvent(keyDown)
```
Johann-S: why did you dispatched the same events twice ?

jeremyvii: That's an oversight on my part. I have removed the extra `dispatchEvent()` in the latest commit.

Johann-S (approved):

PR #29540, opened 2019-10-17
Small changes

Tarang-singhal:
Added margin left and right in p selector,Changed color theme of anchor tag


patrickhlauke:
aeh...no, but thanks.

PR #29545, opened 2019-10-18
v5: Add Customize section to docs?

mdo:
Draft PR to get some feedback first, no need to review yet.

I've been thinking about this for awhile now and one thing I'd like to better showcase is that Bootstrap is extremely customizable, starting with v5. A "theming" page within the `Getting started` section just doesn't feel right anymore. The content represents more than theming, and it's also not really a "getting started" subject.

I swear I'm not making everything a top level nav group. This one feels more important to me with the level of customization we provide, the conversations in other communities around theming (looking at you, React), and the still thriving themes marketplaces out there (not just ours).

Thoughts? What's missing from this section/subject?

---

### Summary of changes

- Deleted the Theming page and expanded it unto an entire section called Customize
- New Customize > Overview page exists just to summarize this section. 
- Sass page includes some updated headings, and adjusted content around Sass map modifications.
- Options page doesn't add much new, but I have renamed two of the super long options to better match the others.
- Color page moves theme colors to the top, combines the previous iteration of the All colors and Grays sections, and adds two swatches for white and black.

---

### After discussion todos
Once I hear from folks, I'll get to work on the following:

- [x] Redirects from old page
- [x] Audit and rewrite content
- [x] Fix broken options table
- [x] Update lists of options
- [x] Update Color page after #29348 merges

### New discussion todos after 3/1
- [ ] Rename the overview into an index? Make the index the default for each section with this
- [ ] Components page likely needs to be moved to another page, or have it's contents changed—this feels more like an approach doc right now.
- [ ] Add new section for extending via plain CSS?
- [ ] Finish optimize page


site/content/docs/4.3/customize/components.md
```diff
@@ -0,0 +1,47 @@
...
+
+// Generate `.bg-*` color utilities
+@each $color, $value in $theme-colors {
+  @include bg-variant('.bg-#{$color}', $value);
+}
+{{< /highlight >}}
+
+## Responsive
+
+These Sass loops aren't limited to color maps, either. You can also generate responsive variations of your components or utilities. Take for example our responsive text alignment utilities where we mix an `@each` loop for the `$grid-breakpoints` Sass map with a media query include.
```
MartijnCuppens: We use the utility API for this nowadays, so this this needs some rewording or another example.

MartijnCuppens:
I agree

mdo:
Adding a page in the latest commits here for optimizing Bootstrap. Lots to be written, and I can handle a decent amount of that, but would love some direction and text for how to customize and optimize our JS. For example:

- How do folks exclude some plugins? Is it include only?
- What do folks need to keep in mind when building JS from source on their own asset pipelines?
- We can push for folks to not include jQuery, anything else?

And lastly, anything else on the CSS side?

site/content/docs/4.3/customize/color.md
```diff
@@ -0,0 +1,111 @@
...
+
+## All colors
+
+All Bootstrap colors are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we don't create text or background color classes for each of these variables. Instead, we choose a subset of these colors for a [theme palette](#theme-colors).
+
+<div class="row">
+  {{< theme-colors.inline >}}
+  {{- range $color := $.Site.Data.colors }}
+    {{- if (and (not (eq $color.name "white")) (not (eq $color.name "gray")) (not (eq $color.name "gray-dark"))) }}
+    <div class="col-md-4 mb-3 font-monospace">
```
MartijnCuppens: `font-monospace` can be moved to the row to avoid repetition

mdo:
@MartijnCuppens I rewrote some docs changes that come from #28508:

- Updated the add to a map by using the map merge, instead of redeclaring the original map. 
- Updated the removing a map value to keep both options. Maybe too much?

Cross-referencing with #28506 again because of the map merge thing for adding values. Mostly need to gut check that.

mdo:
Phew, should be rebased now. Hoping I didn't miss anything. Going to jump into the content and more soon.

mdo:
Just squashed a bunch of commits here, down to 4 from 18. Reviewing this once more and then will merge after the XXL grid tier gets added. I'd like to have the bulk of this in our first alpha.

site/content/docs/4.3/customize/color.md
```diff
@@ -0,0 +1,133 @@
...
+
+Bootstrap's source Sass files include three maps to help you quickly and easily loop over a list of colors and their hex values.
+
+- `$colors` lists all our available base (`500`) colors
+- `$theme-colors` lists all semantically named theme colors (shown below)
+- `$grays` lists all tins and shades of gray
+
+Within `scss/_variables.scss`, you'll find Bootstrap's color variables and Sass map. Here's an example of the `$colors` Sass map:
+
+{{< highlight scss >}}
```
MartijnCuppens, line 101: We should use `scss-docs` here

MartijnCuppens:
The table shortcode will conflict with https://github.com/twbs/bootstrap/pull/30466/files. Also, I'm not sure this is the best way to apply the table styles.

site/content/docs/4.3/customize/color.md
```diff
@@ -0,0 +1,133 @@
...
+    <div class="col-md-4">
+      <div class="p-3 mb-3 bg-{{ .name }} {{ if or (eq .name "light") (eq .name "warning") }}text-dark{{ else }}text-white{{ end }}">{{ .name | title }}</div>
+    </div>
+  {{ end -}}
+  {{< /theme-colors.inline >}}
+</div>
+
+All these colors are available as a Sass map, `$theme-colors`.
+
+{{< highlight scss >}}
```
MartijnCuppens, line 25: We should use `scss-docs` here

mdo:
> The table shortcode will conflict with https://github.com/twbs/bootstrap/pull/30466/files. Also, I'm not sure this is the best way to apply the table styles.

Can definitely rename it, but if there are better approaches, I'm happy to hear them. Any thoughts?

site/content/docs/4.3/customize/color.md
```diff
@@ -0,0 +1,106 @@
...
+
+All these colors are available as a Sass map, `$theme-colors`.
+
+{{< scss-docs name="theme-colors-map" file="scss/_variables.scss" >}}
+
+Check out [our Sass maps and loops docs]({{< docsref "/customize/sass#maps-and-loops" >}}) for how to modify these colors.
+
+## All colors
+
+All Bootstrap colors are available as Sass variables and a Sass map in `scss/_variables.scss` file. To avoid increased file sizes, we don't create text or background color classes for each of these variables. Instead, we choose a subset of these colors for a [theme palette](#theme-colors).
```
ffoodd, line 31: This part (and maybe the whole page) will evolve a lot, you may need help to rebase this later :innocent:

site/content/docs/4.3/customize/components.md
```diff
@@ -0,0 +1,37 @@
...
+
+These Sass loops aren't limited to color maps, either. You can also generate responsive variations of your components. Take for example our responsive alignment of the dropdowns where we mix an `@each` loop for the `$grid-breakpoints` Sass map with a media query include.
+
+{{< scss-docs name="responsive-breakpoints" file="scss/_dropdown.scss" >}}
+
+Should you modify your `$grid-breakpoints`, your changes will apply to all the loops iterating over that map.
+
+{{< scss-docs name="grid-breakpoints" file="scss/_variables.scss" >}}
+
+For more information and examples on how to modify our Sass maps and variables, please refer to [the Sass section of the Grid documentation]({{< docsref "/layout/grid#sass" >}}).
```
ffoodd, line 37: **Extremely clear and comprehensive explanation**. 
I'd only add a paragraph about creating a component, or maybe reword things a bit to suggest this is the way to do — and not only the way we did.

Maybe even mention some linter rules related to this?

site/content/docs/4.3/customize/css-variables.md
```diff
@@ -0,0 +1,42 @@
+---
+layout: docs
+title: CSS variables
+description: Use Bootstrap's CSS custom properties for fast and forward-looking design and development.
+group: customize
+toc: true
+---
+
+Bootstrap includes around two dozen [CSS custom properties (variables)](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) in its compiled CSS. These provide easy access to commonly used values like our theme colors and primary font stacks when working in your browser's Inspector, a code sandbox, or general prototyping.
```
ffoodd, line 9: Would it be the place to explain how / when we use custom properties?

Still in discussion though, but would make sense, IMHO.

site/content/docs/4.3/customize/optimize.md
```diff
@@ -0,0 +1,40 @@
...
+When using Sass in your asset pipeline, make sure you optimize Bootstrap by only `@import`ing the components you need. Your largest optimizations will likely come from the `Layout & Components` section of our `bootstrap.scss`.
+
+{{< scss-docs name="import-stack" file="scss/bootstrap.scss" >}}
+
+
+If you're not using a component, comment it out or delete it entirely. For example, if you're not using the carousel, remove that import to save some file size in your compiled CSS. Keep in mind there are some dependencies across Sass imports that may make it more difficult to omit a file.
+
+## Lean JavaScript
+
+Bootstrap's JavaScript includes every component in our primary dist files (`bootstrap.js` and `bootstrap.min.js`), and even our primary dependency (Popper.js) with our bundle files (`bootstrap.bundle.js` and `bootstrap.bundle.min.js`). While you're customizing via Sass, be sure to remove related JavaScript.
```
ffoodd, line 20: Missing a "how-to" for component removal (could be almost the same than Sass imports, I guess) — it's very likely we'll have this question several times, isn't it?

site/content/docs/4.3/customize/optimize.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Autoprefixer browserslist.rc
+
+Bootstrap depends on Autoprefixer to automatically add browser prefixes to certain CSS properties. Prefixes are dictated by our `browserslist.rc` file, found in the root of the Bootstrap repo. Customizing this list of browsers and recompiling the Sass will automatically remove some CSS from your compiled CSS, if there are vendor prefixes unique to that browser or version.
+
+## Use Purge CSS
+
+_Help wanted with this section, please consider opening a PR. Thanks!_
```
ffoodd, line 28: A few things to consider with this:
- we'd have to maintain an exclusion list of some sort (PurgeCSS WhiteList) since it basically parses HTML and JS to determine which selectors can be applied or not, it's usually missing some (that could be dynamically computed in JS, or not in the initial DOM for some reason). This would need some tests and context, I guess.
- mention various extractors too, to prevent users from struggling with basic configuration.
- I think we can search for posts and references, this has probably been covered several times.

site/content/docs/4.3/customize/optimize.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Minify and gzip
+
+Whenever possible, be sure to compress all the code you serve to your visitors. If you're using Bootstrap dist files, try to stick to the minified versions (indicated by the `.min.css` and `.min.js` extensions). If you're building Bootstrap from the source with your own build system, be sure to implement your own minifiers for HTML, CSS, and JS.
```
ffoodd, line 32: Could we mention how we minify things, here?

Might it be valuable to explicit our build depenndencies?

site/content/docs/4.3/customize/optimize.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Nonblocking files
+
+_Help wanted with this section, please consider opening a PR. Thanks!_
+
+## Always use https
+
+_Help wanted with this section, please consider opening a PR. Thanks!_
```
ffoodd, line 40: Would you be interested in a `@font-face` pattern here? I made it (at some point) ni Boosted.

Also we should try to mention a CDN strategy:
1. using resource hint like `dns-prefetch` or `preconnect`,
2. use SRI as much as possible,
3. `defer` / `async` scripts, maybe?

I do realize that this is pretty heavy and would require both writing and testing, I'll try to help as much as possible if anything makes sens to you here.

ffoodd (commented):
General overview: **I really like it,** it feels very **consistent** and **workable**.

A general comment here, though: would a section about forks (how to maintain and evolve as a fork) would have any interest? This is a step further customization, but Bootstrap currently have 68.6k forks and while most of them are probably dead, I guess at least a few hundreds aren't. I can share my own experience, but the whole team overview about this would be very valuable, as well as opening a dedicated section to review / PR by anyone to share about experiences (patterns, best practices, versionning, etc.).

I do realize that all those comments and questions are pretty heavy and would require both writing and testing, I'll try to help as much as possible if anything makes sens to you here, but I find this very engaging :)

site/content/docs/4.3/customize/components.md
mdo: Adding a section on creating your own, but not sure what linter options we'd specify here. Can tackle in a follow-up.

site/content/docs/4.3/customize/optimize.md
mdo: Adding links and articles for now. This could be better suited in the starter npm project.

site/content/docs/4.3/customize/optimize.md
mdo: That'd be rad! Feel free to take a pass at that :). I'm likely going to merge this PR shortly after addressing your other comments :).

mdo:
> A general comment here, though: would a section about forks (how to maintain and evolve as a fork) would have any interest? This is a step further customization, but Bootstrap currently have 68.6k forks and while most of them are probably dead, I guess at least a few hundreds aren't. I can share my own experience, but the whole team overview about this would be very valuable, as well as opening a dedicated section to review / PR by anyone to share about experiences (patterns, best practices, versionning, etc.).

I don't think we need a full page, but something about this would be super welcome, yes :D.

site/content/docs/4.3/customize/optimize.md
mdo: I think we'll leave this as-is for now and revisit during alphas/betas depending on what folks ask about.

PR #29552, opened 2019-10-19
Implemented color-scheme mix-ins

### LINKED ISSUE #29551: Dark/Light Mode Mixins
I'd like to help push forward with #27514 by implementing a mix-in for dark mode or light mode that can be used by users and throughout the Bootstrap source code. I will file a pull request with my improvements when I'm finished (if this sounds OK).

josephgeis:
Closes #29551

Adds two new mix-ins `dark-scheme` and `light-scheme` that take a block and wrap them in a media query for `prefers-color-scheme`.


MartijnCuppens:
I'm ok with this. Would add two more things:
- This should be mentioned in our theming documentation: https://twbs-bootstrap.netlify.com/docs/4.3/getting-started/theming/
- We should mention this doesn't work in IE and Edge <= 18

juniorRubyist:
Sounds good. I can add them to the docs tomorrow when I have the chance.

MartijnCuppens:
Hi @juniorRubyist,

Could you have a look at updating the docs?

ysds:
I'm not very interested in this. This is a just shorthand of the same fixed phrase no matter who writes.　Your request is the same as adding like follows:

```scss
@mixin reduce-motion {
    @content;
  }
}

@mixin print {
  @media print {
    @content;
  }
}
```

juniorRubyist:
@ysds I think that a shorthand for this would make the code cleaner and create less mistakes. (That second example is just flat out redundant.) If you think about it, you have a mixin for the breakpoints. `media-breakpoint-up(md)` beats having to write `@media (min-width: map-get($grid-breakpoints, md))`.

juniorRubyist:
@MartijnCuppens Sorry bout that. I pushed changes to the docs.

MartijnCuppens:
>  If you think about it, you have a mixin for the breakpoints. `media-breakpoint-up(md)` beats having to write `@media (min-width: map-get($grid-breakpoints, md))`

We use these mixins quite a lot in our codebase, and we also have `media-breakpoint-down()` `media-breakpoint-between()` mixins which are a little more complex, so it makes sense to have the `up` variant.

I think I'm gonna agree with @ysds here, we got to watch out we don't add too many mixins we don't use ourselves.

PR #29570, opened 2019-10-22
Remove empty tags when popover title is not defined

### LINKED ISSUE #29127: empty h3 in popover when title is not defined
**Operating system**
Windows
**Browser and version**
Chrome 73
**Reduced test case**
https://getbootstrap.com/docs/4.3/components/popovers/#four-directions

On popovers that do not have a title defined, the h3 is still there as empty. I feel this could be an accessibility problem having an empty h3 on the page : 
image

I would make sure the h3 disappears if the title is not defined.

charlottedestrait:
### **The problem**

When popovers are defined without title, the h3 tag is still there as empty.

Fixes #29127

### **The solution**

The solution I found was simply to display another template when title is not defined.

Feel free to comment if you have any suggestions :)


js/src/popover.js
```diff
@@ -110,12 +110,21 @@ class Popover extends Tooltip {
 
     // we use append for html objects to maintain js events
     this.setElementContent(SelectorEngine.findOne(Selector.TITLE, tip), this.getTitle())
+    // if the popover does not have a title we delete the .popover-header tag
+    if (!this.getTitle() && SelectorEngine.findOne(Selector.TITLE, tip)) {
```
Johann-S: you should store the result of `SelectorEngine.findOne(Selector.TITLE, tip)`, it'll allow you to use it in your `if` case and it's better in terms of performances

js/src/popover.js
```diff
@@ -110,12 +110,21 @@ class Popover extends Tooltip {
...
+    }
+
     let content = this._getContent()
     if (typeof content === 'function') {
       content = content.call(this.element)
     }
 
     this.setElementContent(SelectorEngine.findOne(Selector.CONTENT, tip), content)
+    // if the popover does not have a content we delete the .popover-body tag
+    if (!this._getContent() && SelectorEngine.findOne(Selector.CONTENT, tip)) {
```
Johann-S: same here

Johann-S (changes requested):
Hi @charlottedestrait,

Thanks for the PR 👍 

Can you add unit tests please?

XhmikosR:
I think we shouldn't add the title in this case, not add it and then remove it.

charlottedestrait:
> I think we shouldn't add the title in this case, not add it and then remove it.

Yes that was what I wanted to do initially but the solution had a big impact. But I am going to modify my code and use this solution which is better. Thanks!

js/src/popover.js
```diff
@@ -121,6 +127,60 @@ class Popover extends Tooltip {
...
+      NAME,
+      config,
+      this.constructor.DefaultType
+    )
+
+    if (config.sanitize) {
+      config.template = sanitizeHtml(config.templateWithHeader, config.whiteList, config.sanitizeFn)
+      console.log(this.element.getAttribute('title') === null)
+      console.log(config.title === null)
+      console.log('--------------------------')
```
XhmikosR: Weird that this passes linting. I'll make an issue but until then drop this

charlottedestrait: Sorry for this oversight! I deleted these dump lines.

js/src/popover.js
charlottedestrait: I changed my solution so this part was removed

js/src/popover.js
charlottedestrait: This part was removed too

charlottedestrait:
> Hi @charlottedestrait,
> 
> Thanks for the PR 
> 
> Can you add unit tests please?

Thanks for the remark, I added them

mdo:
Is this too stale to bring back? Any thoughts @twbs/js-review on how to proceed?

XhmikosR:
@mdo I think we need a simpler solution.

XhmikosR:
Thanks for the PR @charlottedestrait . Closing for now since we'd like to find a simpler solution. If you find a simpler solution, feel free to make a new PR. 🙂

PR #29581, opened 2019-10-24
refactor(button-group): cleaning up

ffoodd:
Two changes:
1. the `hover()` mixin used to apply the exact same styles than `:focus, :active, .active` stack above — which  doesn't use dedicated mixins… Removing mixin here seems a good thing since we can simplify selectors stack and reduce complexity a bit;
2. a stylelint check is disabled for useless qualifying selector: `[type="radio"]` and `[type="checkbox"]` obviously applies to `input` elements, and removing this element selector doesn't prevent selector to match.

Tried this locally without seeing any regression. Did I miss something?

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -1,5 +1,3 @@
-// stylelint-disable selector-no-qualifying-type
-
 // Make the div behave like a button
 .btn-group,
 .btn-group-vertical {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
1. We need to keep the `hover()` mixin in `v4` for backwards compatibility.
2. We also use the qualifying type in reboot. If we change this here, we need to adjust this globally. Changing this in reboot influences specificity and might break things so we'll pass on this.

ffoodd:
Thanks for explanations :)

PR #29585, opened 2019-10-24
Enable eslint no-console rule except for build directory

### LINKED ISSUE #29583: ESLint: enable `no-console` rule
We might need to suppress it in some cases but it's better this way

pwolaq:
Resolves #29583 

I have added the following config to the global `.eslintrc.json` file:
```
"no-console": "error"
```
As a result there were few errors inside `build` directory, which are now fixed by disabling this rule inside directory.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,6 +9,7 @@
     "xo/browser"
   ],
   "rules": {
+    "no-console": "error",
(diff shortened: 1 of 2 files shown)


XhmikosR (approved):
Thanks!

XhmikosR:
I'll manually backport this to v4-dev.

PR #29595, opened 2019-10-26
remove superflous transition parameter

midzer:
Hi,

only `color` is transitioned on `:hover`.

Have a nice weekend
midzer


MartijnCuppens (approved):
Thanks Midzer!

PR #29597, opened 2019-10-27
Enable RFS for font resizing

MartijnCuppens:
This PR only tackles the font size part of https://github.com/twbs/bootstrap/pull/29152, as [requested](https://github.com/twbs/bootstrap/pull/29152#issuecomment-527931753) by @ysds. 

- Enable RFS
- Add documentation
- Disable doc font size overrides

Preview: <https://deploy-preview-29597--twbs-bootstrap.netlify.com/>

**TODO:**

- [x] Link to versioned repo

PR #29598, opened 2019-10-27
Link to versioned RFS repo

MartijnCuppens:
Bootstrap 4 is using RFS 8 instead of `v9` and it looks like this can confusing for some people (https://github.com/twbs/rfs/issues/166).

I will tackle this for master in https://github.com/twbs/bootstrap/pull/29597


XhmikosR:
I'd rather this not be hardcoded, but assuming there won't be any new 8.x versions, I won't block it.

mdo (approved):

PR #29601, opened 2019-10-28
Remove redundant "Navbar divider" from "Contents"

MatiasPuhakka:
I noticed this was left unnoticed after navbar divider mixin was dropped earlier


MartijnCuppens (approved):

MartijnCuppens:
Thanks!

Issue #29584, opened 2019-10-24
Typography: <u> should not be used to underline text in HTML5?


florianlacreuse:
According the the HTML5 specs and as explained [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u), `<u>` has been previously deprecated and reintroduced with a new meaning.

> This element used to be called the "Underline" element in older versions of HTML, and is still sometimes misused in this way. To underline text, you should instead apply a style that includes the CSS `text-decoration` property set to **underline**.

However, the BS docs suggests to use `<u>` element to underline text as it's used to be in the past (as I understand it): https://getbootstrap.com/docs/4.3/content/typography/#inline-text-elements

One more thing, from [this ressource from w3](https://www.w3.org/TR/REC-html40/present/graphics.html#h-15.2.1) about `<b>`, `<i>`, etc.:

> The following HTML elements specify font information. Although they are not all deprecated, their use is discouraged in favor of style sheets.

May be BS should at least provide an utility class to underline text? We can already use `.font-weight-bold` instead of `<b>` and `.font-italic` instead of `<i>`.

XhmikosR:
Yeah, I think this would be a good addition. Personally I almost have to use `u`, `b` or `i`

XhmikosR:
Fixed in #29604

PR #29604, opened 2019-10-28
Change deprecated html tags to text decoration classes

mattvgn:
https://github.com/twbs/bootstrap/issues/29584

The tags `<u>` and `<s>` are deprecated in html 4.01  : https://www.w3.org/TR/REC-html40/present/graphics.html#h-15.2.1

Demo: https://deploy-preview-29604--twbs-bootstrap.netlify.com/docs/4.3/utilities/text/#text-decoration

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -426,7 +426,7 @@ $utilities: map-merge(
     ),
     "text-decoration": (
       property: text-decoration,
-      values: none
+      values: none underline line-through
     ),
     "font-style": (
       property: font-style,
--- a/site/content/docs/4.3/content/typography.md
+++ b/site/content/docs/4.3/content/typography.md
@@ -143,17 +143,19 @@ Styling for common inline HTML5 elements.
 <p><em>This line rendered as italicized text.</em></p>
 {{< /example >}}
 
-Beware that those tags should be use for semantic purpose: 
-`<mark>` represents text which is marked or highlighted for reference or notation purposes.
-`<small>` represents side-comments and small print, like copyright and legal text.
-`<s>` represents element that are no longer relevant or no longer accurate.
-`<u>` represents a span of inline text which should be rendered in a way that indicates that it has a non-textual annotation.
+Beware that those tags should be used for semantic purpose: 
 
-If you want to style your text, you should use those css classes instead:
-`.mark` will apply the same styles as `<mark>`.
-`.small` will apply the same styles as `<small>`.
-`.text-decoration-underline` will apply the same styles as `<u>`.
-`.text-decoration-line-through` will apply the same styles as `<s>`.
+* `<mark>` represents text which is marked or highlighted for reference or notation purposes.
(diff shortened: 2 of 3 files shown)


MartijnCuppens (dismissed):
Thanks!

florianlacreuse:
Thanks @mattvgn!

It would be perfect if this could be backported to v4, if possible.

mdo:
These elements were restored in HTML5.

> Along with other pure styling elements, the original HTML Underline (<u>) element was deprecated in HTML 4; however, <u> was restored in HTML 5 with a new, semantic, meaning: to mark text as having some form of non-textual annotation applied.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u

XhmikosR:
Yup, they are not deprecated any more; their meaning is different.

mattvgn:
This is not deprecated, but the way it's beeing used here is 

> You should not use <u> to simply underline text for presentation purposes, or to denote titles of books.

About the `<s>` nothing in the documentation is telling to not use it : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s

What about removing the `<u>` in the "Inline text elements" section on Bootstrap and let the `<s>`, but still add this two classes : `.text-decoration-underline` `.text-decoration-line-through` ?

mattvgn:
> This is not deprecated but people might use it wrongly :
> 
> > You should not use  to simply underline text for presentation purposes, or to denote titles of books.
> 
> About the `<s>` nothing in the documentation is telling to not use it : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s
> 
> What about removing the `<u>` in the "Inline text elements" section on Bootstrap and let the `<s>`, but still add this two classes : `.text-decoration-underline` `.text-decoration-line-through` ?

@mdo @MartijnCuppens @XhmikosR What do you guys think ?

MartijnCuppens:
We might need to add some documentation about `<u>`, `<s>` (& `<b>` while we're at it) to clarify their meaning.

I think we still need the newly added utilities here, to separate semantic elements from styling classes.

MartijnCuppens (approved):

mattvgn:
Capture d’écran 2019-11-26 à 21 57 39
Here is the screenshot of the documentation based on your feedback

florianlacreuse:
@mattvgn May be you already know that but the documentation is built for each PR, see deploy notifiy check / job in the below approval section. Current direct link for instance: https://deploy-preview-29604--twbs-bootstrap.netlify.com/docs/4.3/content/typography/#inline-text-elements

XhmikosR (approved):

XhmikosR:
Do we want to backport this in v4-dev @mdo @MartijnCuppens?

MartijnCuppens:
> Do we want to backport this in v4-dev @mdo @MartijnCuppens?

This can be seen as a new feature, not sure if we want to release `v4.5`?

XhmikosR:
NP, let's leave it in v5 for now.

PR #29609, opened 2019-10-30
Update normalizeDataKey to match the spec

MisatoTremor:
See https://html.spec.whatwg.org/multipage/dom.html#dom-dataset

### DIFF
--- a/js/src/dom/manipulator.js
+++ b/js/src/dom/manipulator.js
@@ -26,7 +26,7 @@ function normalizeData(val) {
 }
 
 function normalizeDataKey(key) {
-  return key.replace(/[A-Z]/g, chr => chr.toLowerCase())
+  return key.replace(/[A-Z]/g, chr => `-${chr.toLowerCase()}`)
 }
 
 const Manipulator = {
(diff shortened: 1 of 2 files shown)


XhmikosR:
This is not right, because currently we add the hyphen after we call this.

MisatoTremor:
Afaik this does not happen.
The `Manipulator` functions use `normalizeDataKey` as is and would result in wrong attribute names if ever called with key names as coming from a DOMStringMap (`dataset`).
I know that internal calls to these functions already use camel-case, but that would make this function obsolete at best.
My thought was that this function exists as a safe guard to ensure valid attribute names (as the Manipulator could be used by plugin authors etc.) and should at least function as such.

XhmikosR:
https://github.com/twbs/bootstrap/blob/639c405c6510a286a3cfcfd6d733d28d0e7baf92/js/src/dom/manipulator.js#L34

XhmikosR:
You will need to adapt tests then.

Johann-S (approved):
Nice catch @MisatoTremor 👍 
LGTM 👍

XhmikosR:
Thanks!

PR #29624, opened 2019-11-02
config.yml: update popper.js to v1.16.0

XhmikosR:
It seems cdnjs has some issues and doesn't have popper.js 1.16.0. I switched to jsdelivr so that we are consistent with the version we use.


MartijnCuppens (approved):

ayoubkhan558:
hack

XhmikosR:
FYI https://github.com/cdnjs/cdnjs/issues/13524

If this changes soon (like in the next 2 days) we might not need to switch CDNs.

I'm gonna wait a few days. I just want this in before we release v4.4.0.

XhmikosR:
I think I'm going to merge this since I don't see it fixed upstream. And we should probably revisit the other cdnjs links later, if this isn't resolved soon.

FezVrasta:
Note that the only officially supported Popper CDN is unpkg.com, I don't have any way to control other CDNs.

XhmikosR:
NP jsdelivr is automatically updated anyway

PR #29625, opened 2019-11-02
Introduce posibility to disable !important suffix for utility classes

### LINKED ISSUE #29606: Utilities 'strongly written' settings
I would like to decide if utilities are written with `!important` or not.

Bootstrap utilities are super useful but sometimes their use is difficult or generates some doubts because they are written with `!important`. For example, working in a configurable tool like [Liferay](https://github.com/liferay/liferay-portal), we cannot use easily classes written with `!important` because the user should have the last word about styles. For example, if we use an `m-5` and the final user wants a different `margin-bottom` they will need to use `!important` in her custom CSS

Utilities assume you have complete access to the markup but sometimes it is not so easy, or you just want to customize by modifying CSS.

**Proposed solution:**
Add a variable to configure if Utilities are strongly written with `!important` or written without it.  Something like: `$utilities-strongly: true;` or `$utilities-important: true;`

**Extra info / examples**
Some examples of Issues we had by following Utilities:
https://github.com/liferay/clay/issues/377
https://github.com/liferay/clay/issues/381

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/issues/29606. Also tested in Dart Sass.


XhmikosR:
We need a warning next to this, like `Do not enable this unless you understand what it does`.

Also, let's wait for @mdo to chime in.

MartijnCuppens:
Imo, it's quite obvious what the effect is.

XhmikosR:
You'd be surprised by what people think obvious and what not :p But anyway
I guess this is fine here since not everyone will have to fiddle with this
stuff.

MartijnCuppens:
> I guess this is fine here since not everyone will have to fiddle with this

Indeed.

Need a review here, @twbs/css-review

XhmikosR (approved):

ysds (approved):

PR #29633, opened 2019-11-04
Make check label cursor customizable

GaryPEGEOT:
Use a cursor pointer on custom form labels to indicate that the element is "clickable"

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -76,6 +76,7 @@
   margin-bottom: 0;
   color: $custom-control-label-color;
   vertical-align: top;
+  cursor: pointer;
 
   // Background-color and (when enabled) gradient
   &::before {
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -76,7 +76,7 @@
   margin-bottom: 0;
   color: $custom-control-label-color;
   vertical-align: top;
-  cursor: pointer;
+  cursor: $custom-control-cursor;
 
   // Background-color and (when enabled) gradient
   &::before {
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -526,6 +526,7 @@ $custom-forms-transition:               background-color .15s ease-in-out, borde
 
 $custom-control-gutter:                 .5rem !default;
 $custom-control-spacer-x:               1rem !default;
+$custom-control-cursor:                 null !default;
 
 $custom-control-indicator-size:         1rem !default;
 $custom-control-indicator-bg:           $input-bg !default;


MartijnCuppens:
@patrickhlauke, is this a good idea?

patrickhlauke:
this is a constant debate. by default browsers don't show do it. this would be an opinionated change. 
we already had lengthy back and forths about setting `cursor:pointer` for buttons, and in the end resorted to making a new variable for it. we could create a new variable, or rename/expand the current `$enable-pointer-cursor-for-buttons` to cover more "browsers don't normally do this, but it's clickable, so we're forcing a different cursor". but i personally wouldn't go for this, as some users may fear that clicking on the label when it shows a hand cursor may actually take them somewhere else/out of the form altogether (like it's a link or something)

GaryPEGEOT:
> 
> 
> this is a constant debate. by default browsers don't show do it. this would be an opinionated change.
> we already had lengthy back and forths about setting `cursor:pointer` for buttons, and in the end resorted to making a new variable for it. we could create a new variable, or rename/expand the current `$enable-pointer-cursor-for-buttons` to cover more "browsers don't normally do this, but it's clickable, so we're forcing a different cursor". but i personally wouldn't go for this, as some users may fear that clicking on the label when it shows a hand cursor may actually take them somewhere else/out of the form altogether (like it's a link or something)

On the other hand, right now some user don't know they can click on the switch / checkbox, so an option might be interesting to let people decide if they want it to be displayed as a cursor, WDYT ?

patrickhlauke:
there's a lot of things users don't know ... there's a fine line between helping users vs diverging too much from what browsers do by default on all other sites...

there's no right or wrong around this question (despite strong opinions one way or another like https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b) ... so having a variable strikes a certain balance at least

GaryPEGEOT:
@patrickhlauke  what's the prefered way? Create a new variable or rename & expand the current one?

MartijnCuppens:
I wouldn't use a new `$enable-*` variable here but just add a `$custom-control-cursor` variable which is `null` by default.

XhmikosR:
This targets the wrong branch.

GaryPEGEOT:
> 
> 
> This targets the wrong branch.

Ok, so which one should I use ? (It's my first "contribution")

XhmikosR:
The master branch is the default branch.

GaryPEGEOT:
> 
> 
> The master branch is the default branch.

@XhmikosR neither the file nor the class exist on the master branch.

XhmikosR:
Maybe this is only v4 specific then. In that case I'd say I'm against it since it can be a breaking change.

GaryPEGEOT:
@XhmikosR the default value is null, as of the actual behavior, so I don't see how it is a BC.

XhmikosR:
Yeah, I already edited my comment.

mdo:
Custom forms are no longer custom in v5, which is why these styles don't appear there. They're all consolidated into the default form controls so we have a single set of forms. I'm good with the null variable here.

XhmikosR:
@mdo: does this patch apply in master, though?

MartijnCuppens (approved):
Let's do this

PR #29638, opened 2019-11-04
Dist

XhmikosR:
@twbs/css-review please have a look in case we missed something


Johann-S (approved):

XhmikosR:
The only thing I don't get is why those `webkit` prefixes are added. Probaly some browser's usage changed.

`npx browserslist` should show more info @MartijnCuppens

PR #29639, opened 2019-11-04
Add safety check for falsy sanitizer input.

### LINKED ISSUE #29474: `tooltip#sanitizeHtml()` missing `null` check
Operating system: Windows 10
Browser: Chrome 77

Test case:
In update to 3.4.1 community added `sanitizeHtml` feature to `tooltip.js`. Code itself works fine, but it is missing _null check_ before attempt to sanitize string/content/title, which breaks (at least in my project) existing code.

Proposed fix:
```
function sanitizeHtml(unsafeHtml, whiteList, sanitizeFn) {
    // if (unsafeHtml.length === 0) {
    if (!unsafeHtml || unsafeHtml.length === 0) {
      return unsafeHtml
    }
...
}
```

peterblazejewicz:
When the input of sanitizer evaluates to false:
- do not rely on assumed string length
- default to empty string on falsy input

With that change the falsy input like null, undefined or empty string
should return the same results: empty string.

Thanks!

Fixes #29474


js/src/util/sanitizer.js
```diff
@@ -92,8 +92,8 @@ export const DefaultWhitelist = {
 }
 
 export function sanitizeHtml(unsafeHtml, whiteList, sanitizeFn) {
-  if (!unsafeHtml.length) {
-    return unsafeHtml
+  if (!unsafeHtml || !unsafeHtml.length) {
```
XhmikosR: How about we just do `!Boolean(unsafeHtml)`?

js/src/util/sanitizer.js
```diff
@@ -92,8 +92,8 @@ export const DefaultWhitelist = {
...
+    return unsafeHtml || ''
```
XhmikosR: Any specific reason to return an empty string instead of `false`?

XhmikosR:
That issue is for v3 BTW. @Johann-S we should edit to mention if it happens in v4 and v5 (remove 3 since it's EOL) and mark this for backport

js/src/util/sanitizer.js
peterblazejewicz: Good point, done the reading and now I knew Boolean(...) is alrady used here in the source base. To be amended

peterblazejewicz: So that way:
```js
if (!Boolean(unsafeHtml) || !unsafeHtml.length) {
```
cannot be written, as there is a rule in ESLint config `Redundant Boolean call.eslint(no-extra-boolean-cast)`, which fixes to expression I've used. Not sure what to do with this to not over engineer simple statement

js/src/util/sanitizer.js
peterblazejewicz: As there is no JSDoc, I tried to read intent from:
- empty string (String.lenght) returns empty string
- `sanitizeFn` from docs recommends DOMPurify, which returns empty string always (we are using DOMPurify in one product, pretty sure here)
- the return from the function itself, here is the string (`innerHtml`)

Johann-S (approved):
LGTM 👍 
And yes we can backport that in v4 too

js/src/util/sanitizer.js
XhmikosR: I guess we can just do `!unsafeHtml`? @Johann-S what do you prefer?

Johann-S:
Hi @peterblazejewicz,

It seems it's impossible to pass `undefined` or `null` to the sanitizer because, our sanitizer is bundled in our components see: https://github.com/twbs/bootstrap/pull/29733

XhmikosR:

*This is an automated reply*

PR #29644, opened 2019-11-05
adjust card overflow settings

### LINKED ISSUE #29643: Set overflow:hidden on a .card
When a card has a fairly large border radius, the background color can be seen outside of the card. Adding `overflow:hidden` to a card will fix this.

Here is an example with just changing the border radius of a card to 2rem instead of 0.25rem
image

Mac OSX - Mojave -- Safari 12.1.1 and Chrome 78.0.3904.70

korthjp17:
Add overflow:hidden to all instances of card (removed from the specific version of accordion implementation)

fixes #29643

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -12,6 +12,7 @@
   background-color: $card-bg;
   background-clip: border-box;
   border: $card-border-width solid $card-border-color;
+  overflow: hidden;
   @include border-radius($card-border-radius);
 
   > hr {
@@ -243,7 +244,6 @@
 
 .accordion {
   > .card {
-    overflow: hidden;
 
     &:not(:last-of-type) {
       border-bottom: 0;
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
This change could simplify the card a little more (like the removal of `$card-inner-border-radius`).

The only downside will be that we won't be able to add elements which overflow the card like:


(You could achieve this by adding the badge to a wrapper though)

Anyway, imo this is something we can do if:
- we document this in our migration guide
- we make the overflow value a variable instead of a static value
- we remove the `$card-inner-border-radius` css

The last point is the only thing I'm a bit concerned about, if we remove this, it will reintroduce the overflow issues we have when the overflow is reset.

@twbs/css-review, what are your opinions?

mdo:
Sorry, but we can’t do this. Breaks too many things for folks.

ysds:
Thank you thought we'll pass on this per above comment.

PR #29653, opened 2019-11-07
Fixed input-height-sm and input-height-lg calculations

SamDuvall:
I believe that `input-height-sm` should use `input-padding-y-sm` and NOT `input-btn-padding-y-sm`, because one can override `input-padding-y-sm` which defaults to `input-btn-padding-y-sm`.

The same goes for the `input-height-lg` calculation


MartijnCuppens (approved):
Yup, thanks!

PR #29664, opened 2019-11-12
Update tooltip.js

cherrol:
add custom class.


XhmikosR:
Thanks for the patch but v3 is EOL.

TriteLove:
okay, thank for you reply.

Issue #29701, opened 2019-11-21
Add additional test to to check if dart Sass or node-sass with compressed output are still working


MartijnCuppens:
We only test for the `expanded` output style, but the compressed style could potentially break things.

We should also add tests for Dart Sass.

First step is to just check if compilation works, but we might also check if the output is the same (this can get tricky since the output ain't always the same, and I thought the way Dart Sass and lib sass handle `@extend`s is different).

XhmikosR:
I don't think it's our job to catch such issues, though. I mean, if it's a compiler bug, that's the place they test such things.

MartijnCuppens:
Agreed.

Closing this, since we have tests for Sass and node sass in the meantime

PR #29702, opened 2019-11-21
Compressed Sass output support

MartijnCuppens:
If the output style is changed from `expanded` to `compressed` here
https://github.com/twbs/bootstrap/blob/f1f320461066aec558a2a11cfb2f73c841434732/package.json#L24

the build fails, because keywords like `white` are converted to `#fff`. By converting the keys to strings, Sass won't recognize the keys as colors anymore.

I've opened https://github.com/twbs/bootstrap/issues/29701 to have a look at testing this by default.

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -77,7 +77,7 @@ $utilities: map-merge(
     "border-color": (
       property: border-color,
       class: border,
-      values: map-merge($theme-colors, (white: $white))
+      values: map-merge($theme-colors, ("white": $white))
     ),
     // Sizing utilities
     "width": (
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

XhmikosR:
Do we have any other places?

XhmikosR:
BTW this would maybe make sense to have it as a rule in stylelint or stylelint-scss. Not sure where, and how hard it will be to implement.

MartijnCuppens:
Yeah, I've been thinking about that too, probably something for `stylelint-scss` since it's about Sass maps.

XhmikosR:
Please make a feature request in stylelint-scss so that we track this and we'll see if anyone will jump in to provide a patch :)

MartijnCuppens:
It seems this is already a thing, I've opened https://github.com/twbs/stylelint-config-twbs-bootstrap/issues/35 to have a look at this, but this will be quite a bulky change.

mdo (approved):

PR #29708, opened 2019-11-21
Remove Firefox fieldset bug callout

EricDunsworth:
* #21262 removed v4's copy of this callout.
* The bug the callout covered was fixed in late 2016.


EricDunsworth:
Closing since I wasn't aware that Bootstrap 3 had reached EOL.

PR #29709, opened 2019-11-22
Use double quotes in `.stylelintrc`

MartijnCuppens:
This rule should also be backported in `v4-dev`. Shall we do a new PR or cherry-pick https://github.com/twbs/bootstrap/pull/29612 and this PR, @XhmikosR?

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -12,10 +12,10 @@
       "transition"
     ],
     "function-blacklist": ["calc"],
-    'scss/dollar-variable-default': [
+    "scss/dollar-variable-default": [
       true,
       {
-        'ignore': 'local'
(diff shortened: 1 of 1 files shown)


XhmikosR:
We just need to cherry pick the quotes change if it applies on the current v4-dev. Newer stylelint-config-twbs-bootstrap versions don't work with Node.js 6.

XhmikosR (approved):

XhmikosR:
AFAICT there's nothing to backport in v4, so we are good.

MartijnCuppens:
> Newer stylelint-config-twbs-bootstrap versions don't work with Node.js 6.

Ow yeah, I forgot. looks good indeed then

Issue #28247, opened 2019-02-13
The `disabled` property value of the custom radio button is not reflected in the display on Microsoft Edge


4dn-oss:
On Microsoft Edge, on changing the `disabled` property value of a custom radio button from JavaScript, the display of radio button is not updated. The display of radio button is updated correctly on  Firefox (65), Chrome (72.0.3626.96), IE11 (11.590.17134.0).

OS: Windows 10 (1803)
Browser: Microsoft Edge (42.17134.1.0)

Sample: https://codepen.io/4dn-oss/pen/KJeOdz

Expected Behavior: Every time the 'disable/enable' button is clicked, radio buttons will be disabled/enabled.
Result: On Microsoft Edge, radio buttons are always displayed as if it is enabled.

ysds:
Confirmed on Windows 10 (1809), Microsoft Edge 44.17763.1.0

![edge](https://user-images.githubusercontent.com/4065765/52720050-3e902e80-2fea-11e9-92c8-62c90a0a61a4.gif)

I found a similer report: https://github.com/IBM/carbon-components/issues/1536

tmorehouse:
I wonder if calling `vm.$forceUpdate()` on the checkbox after changing the disabled state would get around this issue? If it does, it would be a bit of a hack (as it would force the radio to re-render).  In carbon, it looks like they used a CSS hack to fix it.

4dn-oss:
Adding the CSS style `[disabled] ~ _ {}` introduced in the https://github.com/IBM/carbon-components/issues/1536 seems to solve this issue.

https://codepen.io/4dn-oss/pen/LqJGWQ

4dn-oss:
In my project, `[disabled] ~ _ {}` has deleted by webpack production build process. Dummy style such like  `[disabled]~_{font-size:inherit}` solved this probrem.

PR #29740, opened 2019-11-26
Use pseudo-class too for disabled check/radio

### LINKED ISSUE #29739: Check/radio doesn't consider disabled fieldset
As disabled attribute is used instead of :disabled pseudo-class, check/radio inputs are not rendered as disabled when they are contained in a disabled fieldset (see: https://jsfiddle.net/o17mLfda/1/).

jeromelebleu:
In order to consider check/radio inputs inside a disabled fieldset, this first use the :disabled pseudo-class. By the way, the URL of the relevant issue on M$ Edge seems broken or needs at least an account to track its progress...

fixes #29739


MartijnCuppens:
Maybe we could link to https://github.com/twbs/bootstrap/issues/28247 instead?

MartijnCuppens (changes requested):
This should also be changed in `_form-file.scss`

jeromelebleu:
Thanks for the link and the review. I have updated the patch to use the suggested workaround for Edge instead to keep only the `:disabled` pseudo-class for the ruleset. However, I don't have Edge to validate that it is still working on it. Could someone check that please?

ysds:
I am not a fan of such CSS hack 😕. I think the previous change (both `:disabled` and `[disabled]` ) is better.

jeromelebleu:
@ysds I agree with you, besides it seems to come out of nowhere... I have switched to this hack as it has been tested on Edge and implemented in other projects regarding #28247, and because I cannot confirm that using both `:disabled` and `[disabled]` will work on it too. But it is as you want for sure.

MartijnCuppens:
I agree with @ysds here

ysds:
I want to include this to 4.4.1 if we can work quickly.

jeromelebleu:
I have made changes accordingly, if someone can confirm that it is working on Edge please.

ysds:
I created a PR for v4: #29762

ysds:
So no longer need to rush this PR :)

ysds:
I checked the following test cases:

- Current (master) https://codepen.io/fellows3/pen/jOENWQe
- After https://codepen.io/fellows3/pen/xxbKZyb

ysds (approved):

MartijnCuppens (approved):

PR #29744, opened 2019-11-26
Add font style normal

### LINKED ISSUE #29590: Add `font-style-normal`
When you need to change the font style to italic, you can use `font-italic`. However, sometimes you need to change it back to normal (say all of your `.blockquote`s are italic and you need to make one of them normal). We already have `font-weight` utilities which names start with `font-weight`. Probably `font-style` utilities naming should be corrected as well: 

`font-style-normal`
`font-style-italic`

mattvgn:
Fixes #29590

I was not confident about changing `font-italic` to `font-style-italic` because it would break some people style.


site/content/docs/4.3/utilities/text.md
```diff
@@ -76,6 +76,7 @@ Quickly change the weight (boldness) of text or italicize text.
 <p class="font-weight-light">Light weight text.</p>
 <p class="font-weight-lighter">Lighter weight text (relative to the parent element).</p>
 <p class="font-italic">Italic text.</p>
+<p class="font-normal">Normal text.</p>
```
MartijnCuppens: Not sure if `Normal text` is clear enough, maybe we should go for `Text without font style`?

mattvgn: Good idea ! :)

MartijnCuppens (approved):
Thanks @mattvgn!

PR #29750, opened 2019-11-27
Update hugo

XhmikosR:
goldmark notes:

1. I see a difference when we use the `markdownify` function; we no longer have smartypants. I'm not sure if there's another function for this
2. there are some cases the resulting HTML isn't valid with goldmark

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -3,9 +3,12 @@ title:                  "Bootstrap"
 baseURL:                "https://getbootstrap.com"
 enableInlineShortcodes: true
 
-pygmentsUseClasses:     true
-# you can use ```lang if you want too
-pygmentsCodeFences:     true
+markup:
+  goldmark:
+    renderer:
+      unsafe:           true
+  highlight:
+    noClasses:          false
 
 buildDrafts:            true
 buildFuture:            true
--- a/package-lock.json
+++ b/package-lock.json
@@ -6096,9 +6096,9 @@
       }
     },
     "hugo-bin": {
-      "version": "0.47.1",
-      "resolved": "https://registry.npmjs.org/hugo-bin/-/hugo-bin-0.47.1.tgz",
-      "integrity": "sha512-WJ3xMDTe3Qi9vcDguzSOWUVUEIVnPahlvex/Em1k7ig39IZuGAbDqIjkVzCCbtSZMUCjNHDzO35P7xyH7WX7rg==",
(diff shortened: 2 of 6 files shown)


XhmikosR:
For the second point, I think goldmark has an issue if there's a space in the inline HTML. I'll need to dig in more and report the issues upstream.

In the mean time we could probably update to the latest version but keep using blackfriday until all issues are sorted or until we find a workaround.

XhmikosR:
@bep: only thing left AFAICT is the smartypants issue when we use the `markdownify` function. Now we no longer get them. You can see this on:

* https://twbs-bootstrap.netlify.com/docs/4.3/getting-started/introduction/
* https://deploy-preview-29750--twbs-bootstrap.netlify.com/docs/4.3/getting-started/introduction/

https://www.diffchecker.com/7neZRrYz

MartijnCuppens:
@XhmikosR, you really lost me with this PR. What are `goldmark`, `smartypants`, `blackfriday` or `markdownify` exactly? Or what is the consequence of upgrading?

The only thing I notice is the table of contents looks different, this might need some more padding on the left.

XhmikosR:
The new Hugo version has switched the default Markdown library to goldmark
which has a different behavior in some cases like the invalid HTML.

MartijnCuppens:
It looks like you managed to fix the invalid HTML? What markup was generated before and is this a problem which will be detected by our tests if it occurs again?

XhmikosR:
No test can detect this apart from the HTML validator.

XhmikosR:
https://github.com/twbs/bootstrap/pull/29750/commits/95db6459a2afb13d3d98848792e367a095045100

XhmikosR:
The first issue I guess we'll live with it for the time being unless anyone has any strong opinion.

MartijnCuppens:
> The first issue I guess we'll live with it for the time being unless anyone has any strong opinion.

Not really, but are there any benefits we gain from updating then?

XhmikosR:
For the same reason we update anything.

MartijnCuppens (approved):
This is ok for me, but I would like to have another review before this gets merged, especially for the smartypants change.

XhmikosR:
I'm gonna merge this tomorrow since I'd rather follow the hugo versions than sticking to an old one. The smartypants issue is reported upstream.

PR #29762, opened 2019-11-28
V4: Add :disabled for disabled fieldset

ysds:
This PR is #29740 for v4, targets the `v4-dev-xmr` branch.

Before (4.4.0): https://codepen.io/fellows3/pen/LYYwejp
After: https://codepen.io/fellows3/pen/pooMpWQ

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -54,9 +54,9 @@
     @include box-shadow($custom-control-indicator-active-box-shadow);
   }
 
-  // Use disabled attribute instead of :disabled pseudo-class
-  // Workaround for: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11295231
-  &[disabled] {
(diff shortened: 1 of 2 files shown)


XhmikosR:
@ysds does this apply to master too or only v4?

ysds:
Only v4!

XhmikosR:
Hmm, then we need @MartijnCuppens and @mdo to review too.

MartijnCuppens (approved):

MartijnCuppens:
If we're going to include this in 4.4.1, we'll need to update our blogpost a bit

XhmikosR:
If you guys think it's safe, we can include it since have the patch ready.

MartijnCuppens:
I think we're save with this one

MartijnCuppens:
I've added this in https://github.com/twbs/blog/pull/123

PR #29766, opened 2019-11-28
docs: `$enable-rfs` is true by default

cccabinet:
If so, `$ enable-rfs` in ”Sass options" list must default to `true`.


MartijnCuppens (approved):
Yup indeed, thanks!

PR #29780, opened 2019-12-01
update to v4.4.1 in README

tomByrer:
There prob should be a way to automate this, esp since some release portals are staggered (eg docs still have only v4.3).
But for now, here it is.


XhmikosR:
Thanks, but there's no point in this. master has v5.0. We'll change everything when we release it.

PR #29793, opened 2019-12-02
Drop `.text-justify` class

### LINKED ISSUE #29679: No responsiveness for text-justify?
According to the docs @ https://getbootstrap.com/docs/4.3/utilities/text/#text-alignment, we can create breakpoints for aligning text i.e. `text-left`, `text-sm-left`, `text-xl-center` etc.

There is a `text-justify` class for all breakpoints, but there are no individual break points? Why hasn't this been added with the other alignment classes?

For example, on mobiles a block of text looks better to be centred, but anything larger should be justified. This could easily be achieved by using `text-center text-sm-justify`, to work along the lines of all the other alignment classes.

MartijnCuppens:
Closes https://github.com/twbs/bootstrap/issues/29679. 

Please check if the copy is ok or tweak where needed.

https://deploy-preview-29793--twbs-bootstrap.netlify.com/docs/4.3/utilities/text/#text-alignment


site/content/docs/4.3/utilities/text.md
```diff
@@ -27,6 +21,10 @@ For left, right, and center alignment, responsive classes are available that use
 <p class="text-xl-left">Left aligned text on viewports sized XL (extra-large) or wider.</p>
 {{< /example >}}
 
+{{< callout info >}}
+Notice we don't provide utility classes for justified text. While, aesthetically, justified text can look more appealing, it does make word-spacing more random and therefor harder to read.
```
razh: therefor &rarr; therefor**e**

MartijnCuppens: Thanks!

site/content/docs/4.3/migration.md
```diff
@@ -64,6 +64,7 @@ Changes to Reboot, typography, tables, and more.
 - Simplified table styles (no more 2px border on `thead > th` elements) and tightened cell padding.
 - Dropped `.pre-scrollable` class. [See #29135](https://github.com/twbs/bootstrap/pull/29135)
 - `.text-*` utilities do not add hover and focus states to links anymore. `.link-*` helper classes can be used instead. [See #29267](https://github.com/twbs/bootstrap/pull/29267)
+- Drop `.text-justify` class.
```
XhmikosR: Maybe add an explanation here?

XhmikosR (approved):

XhmikosR (approved):

srsteinberg:
Very disappointed

Masri-Programmer:
why

lollyca:
I agree, why is this class being dropped? Seems useful to me

ranjan3nov:
> I agree, why is this class being dropped? Seems useful to me

WHY??? can u please explain

Siddhesh-Sarfare:
ok then is there any substitute class for this.?

realsolutio:
Removing a class without a replacement, why?

diamyndwallet:
The text-justify is needed and widely used by many sites... What's the replacement or orientation for projects that need text-justify?

Hasenpriester:
> > I agree, why is this class being dropped? Seems useful to me
> 
> WHY??? can u please explain .. How this can be helpful

It is useful, if you want to display text justified.

skyndfly:
Return this class back

tapmeppe:
Why has this class been dropped?

FerJoTello:
Well, if anybody else just like me just discovered that text-justify does not exists anymore the only explanation that i found was on the docs (https://getbootstrap.com/docs/5.3/utilities/text/#text-alignment), check the note!
image


Opinion: It's really funny because they thought that justified text it's hard to read so removing the class was the best option (?). Maybe a striked out text is hard to read, but the class actually exists, so what's the point?

shapeh:
I find this a bit odd too. Thankfully, it is very easy to fix - solved by adding this to my SASS file (scss /.css):

**CSS:**

    .text-justify {
        text-align: justify;
        text-justify: inter-word;
    }

**HTML:**

     `<div class="text-justify">Blah blah</div>`

NullUserException:
> Opinion: It's really funny because they thought that justified text it's hard to read so removing the class was the best option (?). Maybe a striked out text is hard to read, but the class actually exists, so what's the point?
> 
> You should provide the class, it's the decision from the people who is implementing it to determine if it results hard to read or don't, but you guys simply removed it, and at least, I don't understand why. Lol

Justified text often creates unsightly typographic rivers that make text very hard to read, specially if the reader is dyslexic, while serving no semantic purpose -- unlike strikethrough text.

PR #29798, opened 2019-12-03
`active` classes in `.btn-link`

codecalm:
`.btn-link` also should have `active` variants


MartijnCuppens:
Not sure what the purpose is of using the same active state as the hover state?

mdo:
Given `.btn-link` is meant to make a button appear as a hyperlink, we'll pass on this. Our default link styles have no active styles on purpose.

Issue #27805, opened 2018-12-09
Non-horizontal form-group using fieldset renders legend font-size different than label


tmorehouse:
When using a `<fieldset>` as the root element of a non-horizontal `.form-group` (i.e. no `.form-row` or `.row`), the `<legend>` element renders differently than a `<label>` (specifically the font-size being set to `1.5rem` via reboot)

```html
<div class="form-group">
  <label for="input">Label</label>
  <input type="text" class="form-control">
</div>

<fieldset class="form-group">
  <legend>Foobar</legend>
  <div role="group">
    <div class="form-check form-check-inline">
      <input id="check1" class="form-check-input" type="checkbox" name="checks" value="A">
      <label for="check1" class="form-check-label">Radio A</label> 
    </div>
    <div class="form-check form-check-inline">
      <input id="check2" class="form-check-input" type="checkbox" name="checks" value="B">
      <label for="check2" class="form-check-label">Radio B</label> 
    </div>
  </div>
</fieldset>
```

Renders:
![image](https://user-images.githubusercontent.com/2781561/49701433-b4c77d80-fbc2-11e8-8840-e4f88d70d1a8.png)

Manually adding a style `font-size: inherit;` or `font-size:  1rem;` to the `<legend>` appears to fix this issue, but it would be preferential not to have to add the additional styling.

Proposed solution (to mimic `<label>` default) in `/scss/_forms.scss` before the `.col-form-label` rule definition:
```css
.form-group legend {
  font-size: inherit;
}
```

Which renders the following (using the same above HTML):
![image](https://user-images.githubusercontent.com/2781561/49701608-7c28a380-fbc4-11e8-9dc6-f4cd470c9f25.png)

mdo:
Every `<legend>` has that font-size though, it's not a matter of being in a particular form layout. See https://codepen.io/anon/pen/BvKwNV for an example. I don't think this is something we'd change.

Thoughts @MartijnCuppens or @andresgalante?

tmorehouse:
When creating a non-horizontal form composed of multiple inputs and, say, a group of radios or checkboxes, the checkboxes and radios should be placed in a fieldset for proper semantic markup (and for screen readers).  Labels cant be associated with fieldsets, so one uses a legend.  

The default behaviour breaks the cadence of the form "labels" as the `<legend>` pops out visually at 50% larger font size than a `<label>`.

mdo:
[...]

PR #29801, opened 2019-12-04
Backport legend theming options to v4 (#27805)

akrawchyk:
Added the `$legend-*` variables and updated Reboot to use them (see https://github.com/twbs/bootstrap/pull/28917). I also updated the docs to provide an example of the legend in a form.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -414,8 +414,9 @@ legend {
   width: 100%;
   max-width: 100%; // 1
   padding: 0;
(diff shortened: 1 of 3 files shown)


MartijnCuppens:
We're not planning to add new features to `v4` anymore. @twbs/team, correct me if I'm wrong here.

XhmikosR:
Yeah, agreed. For the time being v4-dev is in bug-fix mode only.

PR #29803, opened 2019-12-04
Fix: avoid ignore plain text in input-group-$size

### LINKED ISSUE #29605: form-control-plaintext ignores input-group-$size
Bug reports must include:

- macOS 10.14.6 (18G95)
- Chrome Version 77.0.3865.120 (Official Build) (64-bit)
- Compare the difference here:

```html
<div class="container">
  <div class="row">
    <div class="col">
      <div class="input-group input-group-sm">
          <button class="btn btn-primary">Button</button>
        </div>
        <input class="form-control" value="hello world" readonly />
      </div>
    </div>
  </div>
  <div class="row">
    <div class="col">
      <div class="input-group input-group-sm">
          <button class="btn btn-primary">Button</button>
        </div>
        <input class="form-control-plaintext" value="hello world" readonly />
      </div>
    </div>
  </div>
</div>
```

You'll see that the `.form-control-plaintext` has issues with both width and sizing. The fix is to add `.form-control-sm` to `.form-control-plaintext`.

mendozagioo:
Fixes #29605

### DIFF
--- a/scss/_input-group.scss
+++ b/scss/_input-group.scss
@@ -20,11 +20,17 @@
     min-width: 0; // https://stackoverflow.com/questions/36247140/why-dont-flex-items-shrink-past-content-size
     margin-bottom: 0;
 
+    + .form-control-plaintext,
     + .form-control,
     + .custom-select,
     + .custom-file {
       margin-left: -$input-border-width;
     }
+  }
+
+  // It add padding in plain text for default size
+  > .form-control-plaintext {
+    padding: $input-padding-y $input-padding-x;
   }
 
   // Bring the "active" form control to the top of surrounding elements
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Issue: https://codepen.io/MartijnCuppens/pen/dyPoOKM
Fix: https://codepen.io/MartijnCuppens/pen/KKwpNBx

It seems the inputs aren't aligned due to the border. Is this issue also present in master?

ysds:
@MartijnCuppens Dropped in #28972

mendozagioo:
Hey @MartijnCuppens, in master branch is worst, the input is out of line. The problem here is that `form-control-plaintext` class is considered just for remove border and color but not is considered to use like child of `input-group` class. The fix, just it add `form-control-plaintext` class to be used inside of `input-group` class

MartijnCuppens:
This PR need some additional tweaks to fix the issues with the border. Not sure if it's worth fixing this if we drop this if we drop this in master anyway.

\CC @twbs/css-review

mdo:
> This PR need some additional tweaks to fix the issues with the border. Not sure if it's worth fixing this if we drop this if we drop this in master anyway.

It is indeed not worth it :).

PR #29808, opened 2019-12-06
V4 dev backports and updates

XhmikosR:


XhmikosR:
OK caching seems to do miracles in v4-dev :) Minus Node.js 6.x but we'll figure something out

MartijnCuppens (approved):

PR #29813, opened 2019-12-06
Adjust card spacing variables

### LINKED ISSUE #29812: .card-body padding only uses $card-spacer-x
[Source Code](https://github.com/twbs/bootstrap/blob/9894000b42472c8bb2deb514ecc8c896093f945e/scss/_card.scss#L42)

I considered including OS, Browser, and Test Case, but I don't think any of that affects this issue.  The fix is to add $card-spacer-y to the padding.

bheus:
Closes https://github.com/twbs/bootstrap/issues/29812

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -71,7 +71,7 @@
 //
 
 .card-header {
-  padding: ($card-spacer-y / 2) $card-spacer-x;
+  padding: $card-spacer-y $card-spacer-x;
   margin-bottom: 0; // Removes the default margin-bottom of <hN>
   color: $card-cap-color;
   background-color: $card-cap-bg;
@@ -89,7 +89,7 @@
 }
 
 .card-footer {
-  padding: ($card-spacer-y / 2) $card-spacer-x;
+  padding: $card-spacer-y $card-spacer-x;
   background-color: $card-cap-bg;
   border-top: $card-border-width solid $card-border-color;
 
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
This does influences the appearance of the card somewhat:
**Before:**


**After:**


We might better rename our variables a bit.

beedaan:
I've updated the default value of `$card-spacer-y` and it's usage so that the original appearance is preserved.

MartijnCuppens:
Hi @beedaan,

Thanks for the changes. Maybe we should introduce new variables like `$card-cap-padding-x` and`$card-cap-padding-y` to make things a little more flexible?

beedaan:
It looks like `$card-cap-*` is used for the [header](https://github.com/twbs/bootstrap/blob/master/scss/_card.scss#L76) and the [footer](https://github.com/twbs/bootstrap/blob/master/scss/_card.scss#L93).  The `$card-spacer-y` is used for the `.card-body`, `.card-title`, `.card-subtitle`, `.card-header`, `.card-footer`, `.card-header-tabs`.

We could use the `$card-cap-padding-y` for the header and the footer, but what about the other elements like the title, subtitle, and header-tabs?

beedaan:
I introduced `$card-cap-padding-y` and `$card-cap-padding-x`.  They are used to adjust

- card-title
- card-subtitle
- card-header
- card-footer
- card-header-tabs

The `$card-spacer-y` is only used to adjust 

- card-body

MartijnCuppens:
These were indeed the changes I had in mind, thanks @beedaan. 

The usage of `$card-cap-padding-y` for the title margin still feels a bit odd. Maybe we should add a new variable to control the these margins?

beedaan:
I've created the variable `$card-title-spacer-y`, which is used to control the title and the subtitle.

MartijnCuppens (approved):
Looks good! Thanks for this contribution and follow-up, @beedaan!

PR #29834, opened 2019-12-10
Documentation: Add slash key event to focus the input search.

chuckrincon:


MartijnCuppens:
I've been thinking about this too! Two things which should be fixed:
- The input is also focused when typing in an input field or textarea
- We should make it clearer we have this hotkey, like GitHub does

mdo:
Moving to #31702.

Issue #29774, opened 2019-11-29
Custom control border color missing on focus and checked


QballDevelop:
Focus border color is missing when a custom checkbox is checked.

It's because of the following line of code:
```
&:focus:not(:checked) ~ .custom-control-label::before {
  border-color: $custom-control-indicator-focus-border-color;
}
```

It should be without
`
:not(:checked)
`

Resulting in:
```
&:focus ~ .custom-control-label::before {
  border-color: $custom-control-indicator-focus-border-color;
}
```

mdo:
Bug reports must include a **live demo** of the issue. Per our [contributing guidelines](https://github.com/twbs/bootstrap/blob/master/.github/CONTRIBUTING.md), please create a reduced test case via [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/) and report back with your link, Bootstrap version, and specific browser and operating system details.

---

_This is a saved reply._

QballDevelop:
Here you go:

https://codepen.io/dnl-code/pen/mdydgJN

In all browsers (except safari and IE) and all operating systems.

MartijnCuppens:
You have removed the focus box shadow, which we add for accessibility reasons (some people don't notice the subtle border color change).

TAOlijun:
I&nbsp;Received，thank you!


------------------&nbsp;原始邮件&nbsp;------------------
发件人:&nbsp;"Martijn Cuppens"<notifications@github.com&gt;;
发送时间:&nbsp;2019年12月10日(星期二) 下午3:42
收件人:&nbsp;"twbs/bootstrap"<bootstrap@noreply.github.com&gt;;
抄送:&nbsp;"Subscribed"<subscribed@noreply.github.com&gt;;
主题:&nbsp;Re: [twbs/bootstrap] Custom control border color missing on focus and checked (#29774)




Closed #29774.

QballDevelop:
@MartijnCuppens that doesn't mean that the focus border should be omitted. Why prohibit the possibility to change this? People can easily increase the border thickness which I did and then they will notice the border color change.

I've intentionally removed the box shadow so you would see more clearly what's going on. I know I can change it with my own css but it took me a while to figure out what's wrong here because of the following line in the variables:
`
$custom-control-indicator-focus-border-color:   $input-focus-border-color !default;
`

Please reopen this issue.
Let me fix this and I will create a merge request.

TAOlijun:
[...]

PR #29836, opened 2019-12-10
Fixed checked custom control focus border color (#29774)

QballDevelop:
#29774


mdo:
Given this component was rewritten for v5, and since you can accomplish this by modifying the focus styles, I'm closing as a won't fix. Here's an example of your turquoise applied to the `box-shadow`.

Screen Shot 2020-02-14 at 5 34 36 PM

PR #29843, opened 2019-12-12
Wrong value for word-break property in the .text-break class.

a92gerasimov:
Wrong value for word-break property in the .text-break class. Affects IE & Edge.


florianlacreuse:
Please see PR #29325 and [my message about these properties ](https://github.com/twbs/bootstrap/pull/29325#issuecomment-526357045) as I think this is the way to go.

IMHO using `break-all` is a bad move.

Also do not modify the `dist` files.

XhmikosR:
Closing for the aforementioned reasons

PR #29851, opened 2019-12-13
Switch to @rollup/plugin-node-resolve

XhmikosR:
We might need to adapt our configs. Any help is welcome.


Johann-S:
Seems good to me 👍 What do you mean by: `We might need to adapt our configs`? I'm a bit outdated 😄

XhmikosR:
I said that being it's a major version bump but from the changelog it seems we are fine :)

https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#600

MartijnCuppens (approved):

PR #29872, opened 2019-12-17
docs: Add OpenSearch support

### LINKED ISSUE #29949: Enable OpenSearch
It would be nice to use open search to search through the docs. I made a PR (https://github.com/twbs/bootstrap/pull/29872), but first we got to figure out how to deal with caching and have a solution for the absolute urls in the config xml. Maybe we should also make a separate search page in the docs.

MartijnCuppens:
## How does this work?

I've only tested this in Chrome so far.

- Navigate to https://deploy-preview-29872--twbs-bootstrap.netlify.app/
- Open a new tab
- Enter https://deploy-preview-29872--twbs-bootstrap.netlify.app/ in the address bar, but do not submit
- You'll see this message in the address bar:
image
- Press tab
- Search for `card`
- You'll land on https://deploy-preview-29872--twbs-bootstrap.netlify.app/docs/5.0/getting-started/introduction/#search=card with the search prefilled

Maybe we should create a dedicated search page for this?

Also, I need to make the urls variable, no clue how to do that with xml in Hugo. @XhmikosR, could you help me out with this?

---

TODO:

- [x] Add `InputEncoding`
- [x] Add a 16x16 PNG file
- [ ] Add a proper 64x64 PNG file
- [x] Revert back to the favicon.ico?
- [ ] Add the `link` tag in all pages?
- [ ] Shouldn't the param be a query and not a hash? https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/search
- [ ] We should escape any params
- [ ] See if we can achieve the same without an output format in Hugo
- [ ] add the `link` in all pages
- [ ] Make sure we are not missing anything from https://developer.mozilla.org/en-US/docs/Web/OpenSearch
- [ ] Update images
- [ ] keywords like `@bootstrap`?

Fixes #29949


XhmikosR:
Not sure what you mean the URLs variable?

I'll be able to have a look tomorrow and read about this

MartijnCuppens:
These urls:

https://github.com/twbs/bootstrap/blob/8f5c09a1705fa9ac11fb95c6177c003304a89634/site/static/documentation-search.xml#L5-L6

XhmikosR:
Some notes

1. we don't have access to the build URL as things are now
2. I still can't make this work locally with your instructions, both with FF and CHrome
3. I do see Firefox offering to add the site to the search engines and it does work locally

MartijnCuppens:
> we don't have access to the build URL as things are now

We should definitely get rid of the versioned urls if we want to merge this. 

> I still can't make this work locally with your instructions, both with FF and CHrome

It doesn't seem to work locally here either.

> I do see Firefox offering to add the site to the search engines and it does work locally

I'm not sure how every browser handles this, in Chrome/Brave it worked like described in the PR description.

XhmikosR:
> We should definitely get rid of the versioned urls if we want to merge this.

I don't see how it's related.

MartijnCuppens:
> I don't see how it's related.

Because, if we implement this, people can get a cached version of the Opensearch config which will lead them to the old docs.

XhmikosR:
Well, then let's drop this altogether until (if ever) we have the proper patch.

MartijnCuppens:
Closing this for now for the issues mentioned above.

mdo:
Should we close as stale here?

XhmikosR:
Nope let's leave it open; it does work, I just don't know if it's the proper way to do it in Hugo.

mdo:
Closing as stale.

PR #29882, opened 2019-12-19
Shorthand properties fixes

MartijnCuppens:
Discussed in https://github.com/twbs/bootstrap/pull/29864. Each background property is now split into separate properties (which is way more clear). Also added some comments in this PR to clarify some decisions.

Advantages:

- Clarity
- CSS variables support (needed in the future)

Do not squash this PR


scss/mixins/_gradients.scss
```diff
@@ -20,33 +17,31 @@
 // Creates two color stops, start and end, by specifying a color and position for each color stop.
 @mixin gradient-x($start-color: $gray-700, $end-color: $gray-800, $start-percent: 0%, $end-percent: 100%) {
   background-image: linear-gradient(to right, $start-color $start-percent, $end-color $end-percent);
-  background-repeat: repeat-x;
```
MartijnCuppens: I removed the background repeats from these mixins, since these gradients always stretch over the whole element, see https://www.w3schools.com/css/tryit.asp?filename=trycss3_gradient-linear.

scss/mixins/_buttons.scss
```diff
@@ -43,21 +43,18 @@
   &:disabled {
     color: $color;
     background-color: $background;
-    border-color: $border;
     // Remove CSS gradients if they're enabled
-    @if $enable-gradients {
-      background-image: none;
-    }
+    background-image: if($enable-gradients, none, null);
```
MartijnCuppens: Reason I did this is because the property order will be respected if we use this notation (See inconsistency with the block below)

scss/_variables.scss
```diff
@@ -701,12 +701,11 @@ $form-select-disabled-color:        $gray-600 !default;
 $form-select-bg:                    $input-bg !default;
 $form-select-disabled-bg:           $gray-200 !default;
 $form-select-disabled-border-color: $input-disabled-border-color !default;
+$form-select-bg-position:           right $form-select-padding-x center !default;
 $form-select-bg-size:               16px 12px !default; // In pixels because image dimensions
 $form-select-indicator-color:       $gray-800 !default;
 $form-select-indicator:             url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'><path fill='none' stroke='#{$form-select-indicator-color}' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' d='M2 5l6 6 6-6'/></svg>") !default;
 
-$form-select-background:            no-repeat right $form-select-padding-x center / $form-select-bg-size !default; // Used so we can have multiple background elements (e.g., arrow and feedback icon)
```
MartijnCuppens: I split this into separate variables for clarity.

XhmikosR:
Please don't do this...

MartijnCuppens:
FYI:

File size `bootstrap.min.css` before this PR: 21398 bytes (gzipped)
File size `bootstrap.min.css` after this PR:  21378 bytes (gzipped)

XhmikosR:
@MartijnCuppens this needs a rebase when you can

MartijnCuppens:
> @MartijnCuppens this needs a rebase when you can

Done

XhmikosR:
Generally it looks good apart from the 2 cases we can actually just use the shorthand (where we set all properties). The size comparison isn't fair because you removed background repeats from gradients. I bet that if you switch back those 2 cases then it'll be smaller than using the individual properties.

MartijnCuppens:
I'll rebase this PR this week. Another reason to do this is custom property support (which we'll have in the future, also see https://github.com/twbs/bootstrap/pull/30121).

> Generally it looks good apart from the 2 cases we can actually just use the shorthand (where we set all properties).

Nope, we don't. The `background` property is a shorthand for 8 properties (https://developer.mozilla.org/en-US/docs/Web/CSS/background). 

> The size comparison isn't fair because you removed background repeats from gradients. I bet that if you switch back those 2 cases then it'll be smaller than using the individual properties.

Indeed

XhmikosR (approved):

ysds (approved):

PR #29888, opened 2019-12-19
Cleanup algolia theming

MartijnCuppens:
We load some algolia CSS from a CDN and than completely override it with `!important`s which is far from ideal.

I've themed the search results based on classes we get from algolia, which looks a lot cleaner.


XhmikosR:
BTW this is what I try to fix in #29283, where we can load thirdparty files and build them with hugo, but there's a bug in Hugo and we can't switch to that yet.

XhmikosR (approved):

XhmikosR:
LGTM AFAICT

PR #29892, opened 2019-12-20
add dark mode to pagination

RaquelAM:
Enhancement based on #29635
Adding dark mode to pagination by using `.pagination-dark` class
And also adding this new feature to the documentation

Issue #29635, opened 2019-11-04
Dark theme for pagination


AmirHosseinKarimi:
Hi, Many components of Bootstrap has both light and dark theme,
I found Pagination component does not have dark theme variant.
Please make Pagination component compatible with dark colors.

![Screenshot from 2019-11-04 15-08-10](https://user-images.githubusercontent.com/17345129/68118332-f799ee80-ff14-11e9-919a-6e38ef7be6da.png)

Arvind-Suthar:
You can try making one yourself and making a pull request

aapljack:
I would like to take this on if it hasn't been started yet.

AmirHosseinKarimi:
@aapljack Good news! I did not work on this.

RaquelAM:
Hi!
Did anyone work on this or can I take it?

AmirHosseinKarimi:
@RaquelAM Hi, Looks like no.

RaquelAM:
@AmirHosseinKarimi  ok, I will work on this
Thanks!

aapljack:
@RaquelAM I have started on this feature, but do not have a PR ready yet.

RaquelAM:
@aapljack Oh! I'm so sorry but I solved the issue, I will send the PR to verify my solution, you could do that too, best wishes

aapljack:
@RaquelAM No worries. I took a slightly different approach with the colors and hover states so I went ahead and submitted my PR also. Sorry for the confusion. I had only started to work on this last night. :)

aapljack:
#29893

MartijnCuppens:
Closing this for now. In the future we'll have a look at providing a solution for all components, but let's wait until we drop IE support and be able to use CSS custom properties.

chrisluedtke:
also hoping for a `class="pagination-primary"` `class="pagination-secondary"` etc.

danger89:
Or `paginationdark`

PR #29893, opened 2019-12-21
Added a dark theme for the pagination.

aapljack:
Originating from issue [#29635](https://github.com/twbs/bootstrap/pull/29892)

Added CSS to support a dark version of the pagination.
Updated the documentation to support this change.

Demo: https://deploy-preview-29893--twbs-bootstrap.netlify.com/docs/4.3/components/pagination/#dark-color-scheme


aapljack:
I know another user has submitted a PR similar to this one. I had already started this work when they also started working on the issue. Since my solution was slightly different, I went ahead and submitted what I had done.

jawadulhassan (approved):

XhmikosR:
I think we need to revisit how we are going to proceed with this across all components. I think the nest approach might be to use CSS variables, but IE11 doesn't support them.

Otherwise if we end up doing this for all components, the CSS size will increase a lot more.

If anyone has any better solutions let us know.

MartijnCuppens:
> I think we need to revisit how we are going to proceed with this across all components. I think the nest approach might be to use CSS variables, but IE11 doesn't support them.
>
> Otherwise if we end up doing this for all components, the CSS size will increase a lot more.

I agree, closing this.

NicolasPereira:
Is this PR in the new bootstrap versions?

I looked in the documentation and didn't find it, I had to create my own css rule to meet my demand

PR #29908, opened 2019-12-24
package.json: remove engines property

### LINKED ISSUE #29821: Remove engines version from package.json
I was thinking about it and I *think* it makes sense if we actually dropped this.

Consumers of the npm package just get the dist files so they aren't affected by this AFAICT. What we use to build and develop Bootstrap itself does not affect consumers. So having a version could be considered wrong.

This will allow us to use whatever Node.js version we want to use without having to support unsupported Node.js versions anymore. Which means, we could just stop testing Node.js 6.x in v4-dev and be able to update the devDependencies. I'm not sure if this can be considered a breaking change, though, so please chime in in case I'm missing something.

People willing to contribute can always check what we test on CI for each branch. Or we can just mention it in README.md or CONTRIBUTING.md.

@twbs/team: thoughts?

XhmikosR:
This property doesn't, or shouldn't, affect consumers of the npm package since we distribute the built files.

IMO we should definitely backport this to v4-dev, that's the whole point, but it could be considered a breaking change, although I don't see exactly how. So, please chime in if you think this could break something out there.

Closes #29821 

TODO:

- [x] verify Netlify picks up the proper Node.js version - it's using Node.js 10.x AFAICT, like before
- [ ] decide if we should backport this to v4-dev


MartijnCuppens (approved):

Johann-S:
I don't think it'll break something we do not publish a Node.js package, we are a frontend framework so it's totally fine to me 👍

Johann-S (approved):

PR #29910, opened 2019-12-24
Adding class to center caption on carousel

### LINKED ISSUE #29250: Carousel caption flex alignment
Hello,
I wanted to center the carousel caption vertically inside the carousel but I noticed the caption has an absolute position. 
Is it somehow possible to use flex with the carousel to align the caption?

Regards
Andi

mendozagioo:
Fixes #29250
It adding a new class for center `carousel-caption`

### DIFF
--- a/scss/_carousel.scss
+++ b/scss/_carousel.scss
@@ -194,4 +194,13 @@
   padding-bottom: $carousel-caption-padding-y;
   color: $carousel-caption-color;
   text-align: center;
+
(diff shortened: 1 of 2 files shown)


mendozagioo:
@XhmikosR here is the new PR to master branch

MartijnCuppens:
I would prefer an implementation which just uses utility classes instead of additional CSS.

PR #28591, opened 2019-04-02
Docs unordered lists dropdowns

Katherraptor:
Fixes #25728.

Documentation should suggest semantic `ul` and `li` elements in the dropdowns and navbar components to encourage semantic HTML markup. Added a mention that while examples here depict semantic `ul` elements custom markup is supported.

PR #29927, opened 2019-12-29
Docs: unordered lists dropdowns, more

cccabinet:
In #28591, the drop menu for documents and the drop menu for Navbar were fixed in an unordered list, but input-group, Nav, Button-group, and Scrollspy did the same.


MartijnCuppens:
There are still some occurrences left in the examples and in the tests.

MartijnCuppens (approved):
I've processed the examples and test files, LGTM now

Issue #29928, opened 2019-12-29
[feature] Easy multiple items slider,  supporting `grid` and `flex`, with native fallback on mobile. need discussion


wtorsi:
Hello everyone.

In the last project i thought about horizontal/vertical slider for thumbs, partners and other things which everybody use for scrollable lists, and which can be slid right and left/up and down, and i hope i found a easy solution. 

So, my idea was to provider the
- Minimal weight slider supporting bootstrap grid system. 
I want `carousel` behaviour (not all behaviour for now), but for multiple items. As a note, there are a lot of supporting Google queries which make standard bootstrap carousel works as I want, but they a very complicated and not unique - you always need cheating solutions, need more code. 
I want to use usual `grid` and `flex` utilities to create the slider which you need. 
At this moment I've dropped infinite loop at all, because I don't have tests which passing this option in usability. Those kind of elements usually come from _design_ and _common things for nowage web_. 

- No nav slider utils on mobile (nav is redundant when you have swipes). I want native browser swiping (because it's cheaper for fx). 
- I want easy and fast customisation with `html` only, because it's independent from other languages.
 
As the result, I came up to the following solution, and I want to discuss it (it has bugs). Because it needs more work, but I think it can become a new component.

https://github.com/twbs/bootstrap/pull/29929

Here is the fiddle:
https://jsfiddle.net/wtorsi/2agmz9cp/9/

And here is the vertical:
https://jsfiddle.net/wtorsi/7g0zdyv5/5/

MartijnCuppens:
We already have the carousel, I don't think we need another component like this?

/CC @twbs/team

wtorsi:
Hi, @MartijnCuppens, thanks for you attention.

Yes your are right, but it was not my idea.
At this moment carousel is good only for **one** item.
My idea is to add support of multiple `grid` items with simple and native/browser markup.
It's better to compare my idea with https://owlcarousel2.github.io/OwlCarousel2/ or smth kind of, rather than with bootstrap `carousel`.

XhmikosR:
Ideally this shouldn't be a new component. I'm definitely in favor of adding the feature, see an old attempt of mine in #28210 and we track this in #29130

mdo:
Closing as a won't fix—`.carousel-item`s can contain just about any HTML and CSS you want, so this is possible. Leaving #29130 open for now, though.

PR #29929, opened 2019-12-29
[Feature] Easy & small carousel for multiple items

wtorsi:
Hello everyone.

In the last project i thought about horizontal/vertical slider for thumbs, partners and other things which everybody use for scrollable lists, and which can be slid right and left/up and down, and i hope i found a easy solution. 

So, my idea was to provider the
- Minimal weight slider supporting bootstrap grid system. 
I want `carousel` behaviour (not all behaviour for now), but for multiple items. As a note, there are a lot of supporting Google queries which make standard bootstrap carousel works as I want, but they a very complicated and not unique - you always need cheating solutions, need more code. 
I want to use usual `grid` and `flex` utilities to create the slider which you need. 
At this moment I've dropped infinite loop at all, because I don't have tests which passing this option in usability. Those kind of elements usually come from _design_ and _common things for nowage web_. 

- No nav slider utils on mobile (nav is redundant when you have swipes). I want native browser swiping (because it's cheaper for fx). 
- I want easy and fast customisation with `html` only, because it's independent from other languages.
 
As the result, I came up to the following solution, and I want to discuss it (it has bugs). Because it needs more work, but I think it can become a new component.

Here is the request.


XhmikosR:
Hello.

Not sure if we'll even go ahead with this, but the master version does not use jQuery. Also always make sure tests pass.

XhmikosR:
Let alone the folder 4.1 shouldn't exist. I'm tempted to just close this one and when you are ready to make a proper PR do it then.

wtorsi:
Sure, it's just a template for https://github.com/twbs/bootstrap/issues/29928 issue

patrickhlauke:
just skimming the code, that example would only work for mouse users. no touchscreen, keyboard, or assistive technology support. and particularly for AT on touch, currently the only proper way to support a slider correctly is to have an actual range input and hijacking that (e.g. https://codepen.io/patrickhlauke/pen/byWPMX) or, at that stage, just using some CSS to style an actual range input.

wtorsi:
@patrickhlauke 
First of all, happy new year! 🎄

Just checked, touch and keyboard is supported by default (Chrome) because it's a native scrollable container. You can check `dev` version https://rvio.itsd.one/ (first row after accordeon, and the last row of logos before footer), maybe it will be more informative.

At this moment there are no AT in the example, because I'm not good in the technology, but I hope this moment can be resolved with smbd help. 
For now I don't see any reasons to use `range` inputs in the slider, but at this moment I need more investigation as well, maybe you can describe the reasons more precisely?

wtorsi:
Ok, i think this is a misunderstanding, according to https://www.w3.org/TR/wai-aria-practices-1.1/#carousel  
it's not a slider, it's a carousel. There is no need in range inputs here.

PR #29935, opened 2019-12-31
Year bump to 2020

abcdan:
Decided it was time to bump the licenses to 2020 :)


XhmikosR:
Thanks but we'll deal with it later ourselves. 2020 isn't even here yet, I don't see why the rush...

PR #29938, opened 2019-12-31
reboot.md: Fix rendering for the recent Hugo versions

XhmikosR:
Not sure if we have more cases but from a quick look I couldn't spot any other cases.

Preview: <https://deploy-preview-29938--twbs-bootstrap.netlify.com/docs/4.3/content/reboot/#forms>


XhmikosR:
@bep: could this be https://github.com/yuin/goldmark/issues/81? If so maybe an update would fix the issue?

MartijnCuppens (approved):

PR #29940, opened 2019-12-31
Docs: always show anchors

XhmikosR:
I think this is better UE. We might want to change the color though. Or show the anchor before the header.

Preview: <https://deploy-preview-29940--twbs-bootstrap.netlify.com/docs/4.3/getting-started/introduction/>


mdo:
I'm in favor of keeping them as-is.

PR #29944, opened 2020-01-01
Bump copyright year to 2020.

XhmikosR:
I hope this prevent people from submitting bump year PRs...

PR #29945, opened 2020-01-01
Bump website copyright year

zlatanvasovic:


yakov116:
#29944

zdroid:
@yakov116 that's `master` branch, this is `gh-pages`.

XhmikosR:
I don't think this gives any value and in fact we should just remove the meta from the gh-pages branch, unless we link to these files. (LICENSE, LICENSE-DOCS)

/CC @mdo

mdo:
We can remove them from `gh-pages`.

XhmikosR:
Made #29959 which removes these files.

PR #29948, opened 2020-01-02
Docs navbar tweaks

### LINKED ISSUE #29856: IE: docs: header doesn't seem to be aligned
There seemed to be more bottom padding applied to the header in IE.
image

MartijnCuppens:
Since the containers are now required, we can simplify our navbar.

- Remove redundant classes
- Minor CSS optimalizations
- IE alignment fix (Closes https://github.com/twbs/bootstrap/issues/29856)


XhmikosR:
Can you check if any of these apply to v4-dev?

BTW maybe we should be navbar-scroll taller?

MartijnCuppens:
> Can you check if any of these apply to v4-dev?

The containers are only required since `v5` and the navbar has been rethemed.

> BTW maybe we should be navbar-scroll taller?

What do you mean exactly?

XhmikosR:
https://github.com/twbs/bootstrap/compare/master-xmr-navbar-scrolll

MartijnCuppens:
@XhmikosR could you explain why you would make the navbar taller?

XhmikosR:
Because there's not enough room to horizontally scroll on mobile. It's
mostly noticeable on Firefox.

MartijnCuppens:
Ok, it'll be better to increase the padding instead of messing around with magic height numbers. Just pushed a change, LGTY now?

XhmikosR (approved):

ysds (approved):

PR #29952, opened 2020-01-03
Improve partial caching logic

regisphilibert:
More partials are cached by adding "variations" arguments when possible.

See comments in files.

### DIFF
--- a/site/layouts/_default/404.html
+++ b/site/layouts/_default/404.html
@@ -6,7 +6,7 @@
   <body class="d-flex flex-column min-vh-100">
     {{ partialCached "skippy" . }}
 
-    {{ partial "docs-navbar" . }}
+    {{ partialCached "docs-navbar" . "404" }}
 
     <main class="my-auto p-5" id="content" role="main">
       {{ .Content }}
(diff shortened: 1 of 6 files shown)


regisphilibert:
@XhmikosR I ran local tests to make sure this was proper, but not being too familiar with the project, I will let you run the production builds etc... to make sure everything is good.

XhmikosR:
Thanks @regisphilibert for the patch :)

Now, looking at it, it doesn't appear we save a lot. I'm afraid this might cause issues with not so experienced Hugo people, and I'm wondering what's your opinion about maybe not using parialCached at all, after all? If the gains were big, I wouldn't even discuss it, but being that they are not big, maybe this will cause us more issues in the future?

regisphilibert:
Yes a new comers would have to be familiar with the current caching logic. I would not recommend going through that process if there's no gain. But if it ever comes to really needing this, you could always get back to that code and review.

XhmikosR:
Let me merge this to my branch so that we have everything is one place in case we need this in the future.

Thanks again for the patch. I hope this doesn't put you off to help us out more :)

PR #29955, opened 2020-01-05
Fix logo alignment for IE

MartijnCuppens:
The stack logo is misaligned in IE because IE doesn't handle `height: auto` all too well. With `preserveAspectRatio` we can align the SVG content within, which looks the same visually.

I also added a fix for newlines being added by IDEs each time an SVG is edited.


XhmikosR:
Can you try adding `align-items-center` to the parent row? IE shows the image shrunk and I believe it will look better this way as long as we can make IE behave.

From my quick testing the only way I found to make IE work is to add width and height in the CSS, but this might break responsive.

XhmikosR:
Oh, and the newlines patch shouldn't be in this branch nor in the repo IMO. Not all SVG files are minified completely.

MartijnCuppens:
> Oh, and the newlines patch shouldn't be in this branch nor in the repo IMO. Not all SVG files are minified completely.

Webstorm (and probably other IDEs) automatically respects the `.editorconfig` and add a EOF newline on save. So whenever I want to edit the SVGs I need to temporary disable it.

> Not all SVG files are minified completely.

Could be, but adding this won't do harm in any case.

XhmikosR:
>Could be, but adding this won't do harm in any case.

True, but I'd rather have the newlines at the end for the files that are not completely minified. I can add a `chomp` or `trim` around the SVG includes in Hugo to remove any newlines anyway.

Either way this change does not belong in this branch.

Maruda1986:
x

ysds (approved):
LGTM in the case.

MartijnCuppens:
Dropped the newlines patch

PR #29956, opened 2020-01-06
Add Microsoft Edge for MacOS to supported browsers

coliff:
The new Chromium-based Edge has been available as a beta for sometime and general availability of the final version is expected to ship from Jan 15. I think it makes sense to add it as a supported browser once it's out of beta.
REF: https://blogs.windows.com/windowsexperience/2019/11/04/introducing-the-new-microsoft-edge-and-bing/

(made this change against the v4 branch so it can be updated in the docs for the v4.4.2 release though I realise this would also need to be ported to the v5 branch)

### DIFF
--- a/site/content/docs/4.3/getting-started/browsers-devices.md
+++ b/site/content/docs/4.3/getting-started/browsers-devices.md
@@ -87,7 +87,7 @@ Similarly, the latest versions of most desktop browsers are supported.
       <td>Supported</td>
       <td>Supported</td>
       <td class="text-muted">&mdash;</td>
-      <td class="text-muted">&mdash;</td>
+      <td>Supported</td>
       <td>Supported</td>
       <td>Supported</td>
     </tr>


XhmikosR:
Everything should go into the master branch and if need to we will cherry pick in v4-dev.

Although TBH I see no value mentioning specifically the "new" Edge being it's just Chromium. We should probably just reword this part and not mention Opera specifically, either.

/CC @mdo

mdo:
If we end up listing browser engines, I'd think it would still be super helpful for some folks out there to see an exact browser/version list.

XhmikosR:
I think we should keep it simple. The only reason we mention Edge is because it was using a different Engine. The same goes for Opera before they switched to Chromium.

We should reword this section. Edge should still be listed though, based on our .browserlistrc and the karma browsers we are using.

mdo:
Understand the feedback that we can reword this section perhaps, but for the time being, we can fix one small problem here and show that it's supported on Mac :). I'll open an issue to revisit the docs more thoroughly later.

mdo (approved):

XhmikosR:
This can land as long as we have an issue to track the rewording part.

mdo:
> This can land as long as we have an issue to track the rewording part.

Done in #30194.

XhmikosR:
Though it should have been targeting the master branch.

XhmikosR:
Alright, I did this manually but please next time target the master branch first.

coliff:
> Alright, I did this manually but please next time target the master branch first.

will do. thanks for merging. 👍

Issue #25110, opened 2017-12-29
Missing border radius on input group with validation feedback


MoritzBru:
With new beta 3 and using the new input-group stuff, inputs inside input-groups with some validation feedback after them are missing the border-radius.

![formbug](https://user-images.githubusercontent.com/23639869/34426369-e70a858c-ec35-11e7-86f4-f178276e1f6e.png)
red circle: without border-radius
green circle: with border-radius

This is an example from the [bootstrap documentation](https://getbootstrap.com/docs/4.0/components/forms/#custom-styles)

mdo:
Damn, I'm super disappointed in myself for not catching this during Beta 3. I thought I had this all locked up with that rewrite. This is going to be tough.

/cc @ysds in case he has ideas having looked at some of this code recently

mdo:
I spent an hour on this last night and had nothing, but just had an idea to use the validation state.

```scss
  .form-control,
  .custom-select {
    &:not(:last-child) { @include border-right-radius(0); }
    &:not(:first-child) { @include border-left-radius(0); }

    &:not(:last-child):required,
    &:not(:last-child).is-valid,
    &:not(:last-child).is-invalid {
      @include border-right-radius($input-border-radius);
    }
  }
```

This works on the particular docs example, but I need to see if it doesn't work elsewhere.

ysds:
@mdo I am terribly sorry for the lack of confirmation. There is still a issue. e.g. appended addon:

```html
<div class="input-group mb-3">
  <input type="text" class="form-control is-invalid" placeholder="Recipient's username">
  <div class="input-group-append">
    <span class="input-group-text">@example.com</span>
  </div>
  <div class="invalid-feedback">
    Please provide a valid zip.
  </div>
</div>
```

I created various test cases here. https://codepen.io/fellows3/pen/OzmMoz

mdo:
Nice, thanks for the test case. I've forked your pen to:

- Add my latest built CSS from my local changes I included in my last comment
- Remove validation on multiple inputs—there's no way to do this, and we already explicitly do not support this.

See https://codepen.io/emdeoh/pen/OzmNNw?editors=1100. Have to improve some custom select and custom file, as well as dealing with the appended as you noted.

ysds:
We also have to suppose a case that `.is-invalid` or `.is-valid` .class does not exist as the initial. Like https://getbootstrap.com/docs/4.0/components/forms/#custom-styles

mdo:
[...]

PR #29958, opened 2020-01-06
Update Username input radius _border-radius.scss

SylvainBTL:
Hello 👋, i noticed somes username input don't have same border radius :

- id="validationCustomUsername"
- id="validationServerUsername"
- id="validationTooltipUsername"

->https://getbootstrap.com/docs/4.4/components/forms/

I'm UX/UI designer so .. if you have better fix let me know ;)

Thanks,
Sylvain

### DIFF
--- a/scss/mixins/_border-radius.scss
+++ b/scss/mixins/_border-radius.scss
@@ -17,12 +17,6 @@
   }
 }
 
-@mixin border-right-radius($radius) {
-  @if $enable-rounded {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
We are aware of this, but it's not that easy to fix since we don't know when an input element is the last visible child.

This is a duplicate of https://github.com/twbs/bootstrap/issues/25110, so I'm closing this

PR #29967, opened 2020-01-07
Remove the jasmine-core devDependency.

XhmikosR:
karma-jasmine already depends on jasmine-core.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "hammer-simulator": "0.0.1",
     "hugo-bin": "^0.50.0",
     "ip": "^1.1.5",
-    "jasmine-core": "^3.5.0",
     "karma": "^4.4.1",
     "karma-browserstack-launcher": "1.4.0",
     "karma-chrome-launcher": "^3.1.0",


Johann-S:
You're right 👍

Johann-S (approved):

floreengrad:
I now get a warning on karma-jasmine-html-reporter having a non met jasmine-core dependency.

XhmikosR:
It's something on your side, there's nothing wrong with this change.

floreengrad:
indeed I was using yarn.
yarn install seems to display some warnings that I don't see with npm.

XhmikosR:
Since karma-jasmine depends on jasmine-core, it's redundant to have jasmine-core in the devDependencies, which is what this patch was about.

I don't use yarn or any other package managers, nor do we test them on CI.

floreengrad:
I totally get the dep removal, no prb with that, it makes total sense.
I was trying to catch up with the last 100 commits you did since October and noticed this change, that's all.

PR #29968, opened 2020-01-07
sanitizer.js: Add srcset in the allowed attributes

XhmikosR:
Not sure if this should be backported


Johann-S:
It's sanitized by Angular too here: https://github.com/angular/angular/blob/7.2.4/packages/core/src/sanitization/html_sanitizer.ts#L67

so it's a good change 👍 

It depends if we plan to ship a new v4 release or not

Johann-S (approved):

XhmikosR:
I think I will try backporting this along with #29969. We do need to release a new 4.4.x version anyway.

PR #29969, opened 2020-01-07
Use regex.test() when we want to check for a Boolean.

XhmikosR:
Not sure if we should backport this

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -747,9 +747,8 @@ class Tooltip {
   _cleanTipClass() {
     const tip = this.getTipElement()
     const tabClass = tip.getAttribute('class').match(BSCLS_PREFIX_REGEX)
-    if (tabClass !== null && tabClass.length) {
-      tabClass
-        .map(token => token.trim())
+    if (tabClass !== null && tabClass.length > 0) {
+      tabClass.map(token => token.trim())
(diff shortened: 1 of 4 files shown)


js/src/tooltip.js
```diff
@@ -747,9 +747,8 @@ class Tooltip {
   _cleanTipClass() {
     const tip = this.getTipElement()
     const tabClass = tip.getAttribute('class').match(BSCLS_PREFIX_REGEX)
-    if (tabClass !== null && tabClass.length) {
-      tabClass
-        .map(token => token.trim())
+    if (tabClass !== null && tabClass.length > 0) {
```
Johann-S: `tabClass.length` is equal too `tabClass.length > 0` not really needed that change

Johann-S (approved):
LGTM 👍 

I don't think it's really needed to backport that

js/src/tooltip.js
XhmikosR: It's what we do in poper.js; this code is exactly the same

Johann-S: So we can change that in `popover.js` too 😄 but it's nitpicking

XhmikosR: Fine by me, I'd just prefer if we enforced this with a rule, but I cannot find one, only for the opposite `unicorn/explicit-length-check`

PR #29971, opened 2020-01-07
Update docs-versions.html

XhmikosR:
Remove old versions

Alternatively, we could just get rid of all versions minus the latest and just link to the versions page for the rest.

Preview: <https://deploy-preview-29971--twbs-bootstrap.netlify.com/docs/4.3/getting-started/introduction/>


XhmikosR:
Or maybe also skip the minor and patch number and always use `.x`, i.e `4.x`, `3.x` etc?

mdo:
Can you also update the `latest` badge on the  versions page in this PR?

XhmikosR:
That's done automatically it's fixed in the v5 PR

XhmikosR:
So, which solution do you guys prefer?

MartijnCuppens:
> Alternatively, we could just get rid of all versions minus the latest and just link to the versions page for the rest.

I do think people want to switch to some older versions easily.

> Or maybe also skip the minor and patch number and always use .x, i.e 4.x, 3.x etc?

Explicitly displaying the minor version will somehow warn the user the docs can be different if they have another version.

I would also like to keep the link to `v2`, just in case someone wants to switch.

XhmikosR:
I don't agree. Someone might want to switch to 1.x too, but we don't have it there. Every version is listed in the versions page and the toggler doesn't really "switch" the current page.

We should keep things simple and we shouldn't really promote old versions since we don't support them.

Regardless, let's simplify the toggler because it makes no sense to me as it is.

XhmikosR:
But by the same logic 3.x shouldn't be listed either, so I guess we can bring back 2.x

MartijnCuppens:
> we shouldn't really promote old versions

It's not about promoting old versions, it's about providing the docs to developers with legacy projects. I doubt people will install v2 because it's still present in the switcher. 

> But by the same logic 3.x shouldn't be listed either, so I guess we can bring back 2.x

👍

So this is ready for review then?

XhmikosR:
Should we backport this to v4-dev if it lands on master?

MartijnCuppens:
Hmm, this reminds me we were thinking about loading this dropdown with ajax to make it more maintainable. Maybe we focus on that instead of implementing this change?

XhmikosR:
This is a lot easier for now and doing that will require some thought and secondly modifying the existent gh-pages files, which can be a pain in the ass.

MartijnCuppens:
> This is a lot easier for now and doing that will require some thought and secondly modifying the existent gh-pages files, which can be a pain in the ass.

Yeah probably.

> Should we backport this to v4-dev if it lands on master?

Don't think so then.

coliff:
I vote to backport this change to v4. Previous versions of v3 and v4 have JavaScript security vulnerabilities so I think it's better to de-promote links to those old docs. Those really old v4 Alpha docs aren't helpful to anyone now and could be potentially be confusing to users coming to the site for the first time (maybe they think its a preview of a newer version).

Also, there's still that problem with old versions appearing prominently in search results, and reducing the number of pages with links to those old docs may mean that search engines may take this in to account and downrank them. (worth a try!)

XhmikosR:
Backporting to v4-dev won't magically change the already built docs :P

coliff:
yeah I know, but if that change makes it to the next 4.4.2 release when the v4 docs get rebuilt then I think it's worth it. It's a really good change I think....

MartijnCuppens (approved):

PR #29988, opened 2020-01-08
Add nav link transition

MartijnCuppens:
The nav links do not transition while tabs do, this felt a bit odd.

https://deploy-preview-29988--twbs-bootstrap.netlify.com/docs/4.3/components/navs/#javascript-behavior


mdo (dismissed):

XhmikosR:
Are we 100% sure about this? It feels a little weird to me. Maybe we shouldn't have a transition for tabs either?

MartijnCuppens:
> Maybe we shouldn't have a transition for tabs either?

The transition can be removed by removing the `fade` class.

> Are we 100% sure about this? It feels a little weird to me.

It felt weird to me the buttons had a transition while these nav links hadn't.

XhmikosR:
I'm still not convinced about this. I'm not saying it's wrong but it just
feels weird.

MartijnCuppens:
I guess it would be fine to leave this as is since I seem to be the only one in the world whining about this 😄

XhmikosR:
Let's wait until @ysds or @mdo comment too :)

ysds:
I can not find the branch...

XhmikosR:
@ysds check again please, I restored it (you could do it also via the button :))

scss/_variables.scss
```diff
@@ -818,6 +818,7 @@ $zindex-tooltip:                    1070 !default;
 
 $nav-link-padding-y:                .5rem !default;
 $nav-link-padding-x:                1rem !default;
+$nav-link-transition:               color .3s ease-in-out, background-color .15s ease-in-out, border-color .15s ease-in-out !default;
```
ysds: The `color`'s transition duration is different from the others, so I felt my eyes irritated. Perhaps all duration should be `.15s` same as the buttons. If it is fixed, this looks good to me.

BTW, the transition mixin generates unnecessary code for prefer reduce motion when given `null`. I will create a PR to fix this at a later date.

MartijnCuppens:
Changed the transition duration to `.15s`, which feels better.

ysds (approved):

XhmikosR:
I'd like @mdo's opinion before merging this

PR #29997, opened 2020-01-10
Fix padding calculation for modal footer

karlsve:
This solves padding assignment for list type values regarding the modal-footer padding since the last change broke existing usages as mentioned in https://github.com/twbs/bootstrap/pull/25103#issuecomment-572949386.


MartijnCuppens:
This solution is way too overengeneerd. By just adding `$modal-inner-padding-x` and `$modal-inner-padding-y` (like we do with other components too), we should be covered.

mdo:
Closing per last comment. Thanks though!

Issue #29994, opened 2020-01-09
Compiling negative border-radius by "0px"


endcoreAK:
If i change the variable

`$border-radius: .25rem !default;` to  `$border-radius: 0px !default;`

all the variables wich contain **subtract** like

`$card-inner-border-radius: subtract($card-border-radius, $card-border-width) !default;`

will result in a negative border-radius wich means an invalid property value:

`.card-header:first-child { border-radius: -1px -1px 0 0; }`

mendozagioo:
@endcoreAK if you need to remove all borders, just modify: `$enable-rounded: false`

mendozagioo:
Anyway I added a conditional to avoid negative values ​​with the subtract function

mendozagioo:
So, I return to my original idea: this is not an issue (to me), because if you target is remove rounded, you just need to change to `false` the variable `$enable-rounded` to avoid negative values and it create less code

endcoreAK:
Yes, I usually go this way too, but I forgot to set it to false in one project and that's when I noticed it. Your solution with the additional query is great! :)

PR #29999, opened 2020-01-11
Avoid negative CSS values with subtract function

mendozagioo:
Fix issue: #29994


MartijnCuppens:
The function is still called `subtract()`. I woudn't expect the output of `subtract(1, 3)` would be 0.

mendozagioo:
After creating the PR, I kept thinking about that. Just ignore it.

PR #30006, opened 2020-01-12
v4: drop Node.js < 10 and dev dependency updates

XhmikosR:
This drops the engines version from package.json and updates all devDependencies. Also I updated Jekyll to v4.0.0 which is only slightly faster but better than nothing.

In the past we considered such changes breaking ones. I don't think they should be considered breaking ones since they are development only changes. And from now on we should be more lax on master branch anyway.

/CC @twbs/team for feedback.

PS. I will probably rebase this later with further dep updates.

Refs #29821 

TODO:

- [x] Adapt v4-dev branch protection rules
- [x] Switch to terser
- [x] look if we can update jekyll-toc
- [x] Make sure docs are fine: <https://deploy-preview-30006--twbs-bootstrap.netlify.com/>

PR #30008, opened 2020-01-13
Add dependabot config for v4

XhmikosR:
Theoretically this should work, not sure how to test it without merging, though.

We need to first merge #30006 and then this one and it might need further tweaks later.


XhmikosR:
I changed the interval to monthly since v4-dev is in maintenance mode (mostly). We can always run it manually if we need to.

MartijnCuppens (approved):

PR #30026, opened 2020-01-15
[idea] Add inherit border color to Utils

k-utsumi:
Add class to change border color Easy with color.

_e.g._
```
<div class="border border-color-inherit text-warning">
<div class="border border-color-inherit text-white-50">
<div class="border border-color-inherit" style="color:#563d7c">

<div class="text-warning">
  <div class="border border-color-inherit">
```

Playground: https://jsfiddle.net/kpsgh75t/

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -77,7 +77,7 @@ $utilities: map-merge(
     "border-color": (
       property: border-color,
       class: border,
-      values: map-merge($theme-colors, ("white": $white))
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
I don't think this will be a utility which will be used a lot.

/CC @twbs/css-review for second opinions

mdo:
Agreed w/ @MartijnCuppens. I've used this quite sparingly myself, so I think it's something to leave to the user to customize.

PR #30032, opened 2020-01-16
Fixed vague [type="button"] selector in _reboot.

### LINKED ISSUE #30031: [type="button"] styles apply to anything, not just inputs
Not a serious issue, but one I found on accident.

With Bootstrap, you can give any element the attribute `type="button"` and it will be given the CSS property `-webkit-appearance: button`. I don't know whether this is intentional or not, but that doesn't seem to be the case. If there is some benefit to this, please let me know!

[JSPen example](https://jsbin.com/gocotedage/edit?html,css,output)

Tested on Firefox 72 & Chrome 79 on Ubuntu Linux 19.10.

I'll be creating a pull request to fix this.

josephodom:
Fixes #30031


MartijnCuppens:
Hi @josephodom,

This change is moot, `[type="button"]` is only allowed on inputs (and buttons), so there's not really a reason to add CSS and increase specificity.

Also, avoid adding generated CSS to PRs.

PR #30035, opened 2020-01-17
Add utility class for the user-select property

### LINKED ISSUE #18176: Add utility class for user-select:none ?
Is there a reason why the text selection isn't disabled on the `.checkbox label` element? It's kind of annoying when you're clicking fast on the element, see the image:

screen shot 2015-11-05 at 13 56 04

Adding these styles fixes the issue:

```
.checkbox label{
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

<hr>

Or maybe creating, for example, a `.disable-user-select` class, like:

```
.disable-user-select{
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
}
```

```
<div class="checkbox">
    <label class="disable-user-select">
        <input type="checkbox"> Check me out
    </label>
</div>
```

<hr>

However, I've noticed that you using the `user-select` style on the `.btn` class. Should I add `.btn` class to the `.checkbox label` element? Is that the "bootstrap" way for dealing with it?

```
<div class="checkbox">
    <label class="btn">
        <input type="checkbox"> Check me out
    </label>
</div>
```

JensLuyten:
Closes #18176


JensLuyten:
Provided a docs page for Behavior utility classes. 
There didn't exist a utility category where the user-select utility class fitted, and this page can serve for possible other behavior utilities (e.g. pointer-events).

MartijnCuppens (dismissed):
LGTM, thanks!

mdo:
Why a page named `Behaviors`? Should this instead be simply `User select`, or would something else go on this new page that we should be thinking about

MartijnCuppens:
> Provided a docs page for Behavior utility classes.
There didn't exist a utility category where the user-select utility class fitted, and this page can serve for possible other behavior utilities (e.g. `pointer-events`).

I was thinking about adding these `pointer-events` later, that's why I suggested a more general page.

MartijnCuppens:
Ping @mdo 

I'm thinking about adding `pointer-events` utilitiesone of these days, so we'll need to decide whether we document it on the same page as these `user-select` utilities or create separate pages.

XhmikosR:
Behaviors doesn't seem to be self-explanatory. Not sure where this should belong, but behaviors doesn't seem good.

mdo:
Eh, behaviors might be the best group option if we don't want to have each utility on its own page. Some more ideas related to that one if we're open to brainstorming:

- Mouse events
- Mouse behaviors
- Interactions

MartijnCuppens:
Maybe _Interactions_ then, since "mouse" doesn't apply to touch-only devices?

ysds:
I prefer "Interactions" 😉. [Microsoft guidelines](https://docs.microsoft.com/en-us/windows/uwp/design/input/) (Interactions > Selecting text and images) have crossed my head.

MartijnCuppens:
@JensLuyten, could you change the category to `Interactions`?

mdo (approved):

mdo:
@MartijnCuppens I pushed a change to their branch for them, and edited the title description while I was there. Let me know if I missed anything with a quick review, please.

MartijnCuppens (approved):

PR #30039, opened 2020-01-17
Nested list-group-horizontal in list-group-flush border fix

### LINKED ISSUE #30027: Nested list-group-horizontal inside list-group-flush renders without proper borders
Bug reports must include:

- Operating system and version (Windows)
- Browser and version (Tested on Chrome and Firefox)
- Test Case : https://codepen.io/jarodhanko/pen/vYEaoKr
- Bootstrap Version: 4.4.1

Perhaps this is a bit of an edge case or simply unsupported behavior, but if you attempt to nest a list-group-horizontal inside of a custom list-group that is using list-group-flush, the list-group-horizontal renders with the inner borders, the side borders, or the top border of the first item.  It is inheriting the css from .list-group-flush .list-group-item.  Any nested list groups that are not horizontal list groups render as list-group-flush as well.

It works fine with a normal list group.

PDerid:
Added styles only to the top level element of the list. 
Removed following lines:

https://github.com/twbs/bootstrap/blob/dfc1149597774bb3549c2cb42c73cf89b98c6615/scss/_list-group.scss#L145-L149

and added the border width of 0 to the last `.list-group-item` child.

closes #30027
closes #27126 because the double border does not occur in the code update.

[Netlify link](https://deploy-preview-30039--twbs-bootstrap.netlify.com/docs/4.3/components/list-group/#nested)

### DIFF
--- a/scss/_list-group.scss
+++ b/scss/_list-group.scss
@@ -132,7 +132,7 @@
 // useful within other components (e.g., cards).
 
 .list-group-flush {
-  .list-group-item {
+  > .list-group-item {
     border-right-width: 0;
     border-left-width: 0;
     @include border-radius(0);
@@ -140,10 +140,8 @@
     &:first-child {
       border-top-width: 0;
     }
-  }
 
-  &:last-child {
-    .list-group-item:last-child {
+    &:last-child {
       border-bottom-width: 0;
     }
   }
--- a/site/content/docs/4.3/components/list-group.md
+++ b/site/content/docs/4.3/components/list-group.md
@@ -112,6 +112,58 @@ Add `.list-group-horizontal` to change the layout of list group items from verti
 {{< /list-group.inline >}}
 {{< /example >}}
 
+## Nested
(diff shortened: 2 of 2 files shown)


scss/_list-group.scss
```diff
@@ -132,18 +132,16 @@
...
     @include border-radius(0);
 
     &:first-child {
       border-top-width: 0;
     }
-  }
 
-  &:last-child {
-    .list-group-item:last-child {
+    &:last-child {
```
MartijnCuppens: This breaks the kitchen sink example:
https://deploy-preview-30039--twbs-bootstrap.netlify.com/docs/4.3/components/card/#kitchen-sink

MartijnCuppens (changes requested):
Card kitchen sink example is broken by this change:
 https://deploy-preview-30039--twbs-bootstrap.netlify.com/docs/4.3/components/card/#kitchen-sink

PR #30042, opened 2020-01-17
Provide a standalone accordion component

### LINKED ISSUE #25811: .accordion-flush
It might be interesting to add an `accordion-flush` class like `list-group-flush`.

### LINKED ISSUE #28134: Accordion: card-header border bottom hidden by card-body.
When an accordion element is open, the border bottom of the `card-header` element is hidden by the `card-body` element (in case of non-transparent background). See here the actual comportment : https://codepen.io/anon/pen/GzjNYL.
This behavior is due to this :

https://github.com/twbs/bootstrap/blob/e65dccfc95c04aba3de66d9880ad1e6852b659f2/scss/_card.scss#L286

This addition is due to those issues : #27124, #27108.
The bug reported here was the double `border-bottom` between the accordion elements when closed.

See the result when removing the rule over the `margin-bottom` : https://codepen.io/anon/pen/XOjNGz
Unfortunately the double `border-bottom` is back but the behavior of the `card-header` of an open element seems correct with a `border-bottom`.

### LINKED ISSUE #28873: position:sticky inside accordion
- Operating system and version: Linux
- Browser and version: Chrome 73 (but reproducible on all browsers)
- [Reduced test case with fix](https://codepen.io/anon/pen/NVmZdK?editors=1100#0)

In #27133 to remove the double border at the bottom, cards inside an accordion got overflow: hidden done. This means that `position: sticky` doesn't work (my guess is that because the browser is hiding things that are overflowing, sticky doesn't trigger?).

The codepen attached replaces that with a removal of the border - since that's what we're actually trying to achieve but I'm not 100% sure that's the right solution.

### LINKED ISSUE #30015: Create standalone accordion component
The accordion now is a mixture of cards with a header made of a button and some utility classes. I think it should be a standalone component which is configurable with some Sass variables.

Maybe we should also have a striped (like the [tables](https://getbootstrap.com/docs/4.4/content/tables/#striped-rows)) or flushed (like the [list group](https://getbootstrap.com/docs/4.4/components/list-group/#flush)) variant.

gijsroge:
This adds a standalone accordion component with its own settings and markup.

Preview: https://deploy-preview-30042--twbs-bootstrap.netlify.com/docs/4.3/components/accordion/

Also added a **[flush](https://deploy-preview-30042--twbs-bootstrap.netlify.com/docs/4.3/components/accordion/#accordion-flush-example)** and **[striped](https://deploy-preview-30042--twbs-bootstrap.netlify.com/docs/4.3/components/accordion/#accordion-striped-example)** variation.

Fixes #30015 

Fixes #28134 by moving the border bottom from the header to a border top on the body we fix the overlapping issue where the border would be invisible if the body had a background color. This also allowed me to also remove the `overflow: hidden;` on the `.accordion-item` which fixes #28873

Also fixes #25811

Currently this is still a work in progress.
- [x] Add active state
- [x] Add focus state
- [x] Add icon based on collapsed state
- [ ] Double check design
- [ ] Double check code
- [ ] Check if all conditions described in https://www.w3.org/TR/wai-aria-practices-1.1/#accordion are met
- [ ] If these conditions are met, check if https://github.com/twbs/bootstrap/blob/master/site/content/docs/4.3/components/collapse.md#L130 is still relevant


scss/_accordion.scss
```diff
@@ -0,0 +1,146 @@
...
+  width: 100%;
+  padding: $accordion-button-padding-y $accordion-button-padding-x;
+  color: $accordion-button-color;
+  text-align: left;
+  background-color: $accordion-button-bg;
+  border: $accordion-border-width solid $accordion-border-color;
+  border-bottom: none;
+  @include border-radius(0);
+
+  &[aria-expanded="true"] {
```
MartijnCuppens: I would prefer classes here. In general we use the aria labels for accessibility and classes for theming.

MartijnCuppens:
image
Because the box-shadow overlaps the border, it looks like the bottom shadow is a bit smaller

scss/_accordion.scss
```diff
@@ -0,0 +1,146 @@
...
+  padding: $accordion-button-padding-y $accordion-button-padding-x;
+  color: $accordion-button-color;
+  text-align: left;
+  background-color: $accordion-button-bg;
+  border: $accordion-border-width solid $accordion-border-color;
+  border-bottom: none;
+  @include border-radius(0);
+
+  &:not(.collapsed) {
+    color: $accordion-button-active-color !important; // stylelint-disable-line declaration-no-important
```
MartijnCuppens: Any reason why these `!important`s where added? Seems to work fine without?

scss/_accordion.scss
```diff
@@ -0,0 +1,146 @@
...
+    background-color: $accordion-button-active-bg !important; // stylelint-disable-line declaration-no-important
+    &::after {
+      background-image: escape-svg($accordion-button-active-icon);
+      transform: rotate(180deg);
```
MartijnCuppens: Would make this variable, could be people want to rotate 90 degrees, or flip the chevron

scss/_accordion.scss
```diff
@@ -0,0 +1,146 @@
...
+    content: "";
+    background-image: escape-svg($accordion-button-icon);
+    background-repeat: no-repeat;
+    background-position: center;
+    background-size: 1.5rem;
+    @include transition($accordion-icon-transition);
+  }
+
+  &:focus,
+  &.focus {
```
MartijnCuppens: I don't think the `.focus` class is needed here, you probably copied this from the buttons? They are used there for the button js plugin (not sure if this is even applicable these days, but that 's something else)

scss/_accordion.scss
MartijnCuppens: Aha, probably because of the zebra stripes? Maybe then just use the `:not()` construction there too? This way, people will still be able to override the behaviour with utilities.

gijsroge: Yes this was one of the things I was planning on refactoring, will look at it hopefully this weekend :D

scss/_accordion.scss
```diff
@@ -0,0 +1,146 @@
...
+  border: $accordion-border-width solid $accordion-border-color;
+  border-bottom: none;
+  @include border-radius(0);
+
+  &:not(.collapsed) {
+    color: $accordion-button-active-color !important; // stylelint-disable-line declaration-no-important
+    background-color: $accordion-button-active-bg !important; // stylelint-disable-line declaration-no-important
+    &::after {
+      background-image: escape-svg($accordion-button-active-icon);
+      transform: rotate($accordion-button-active-icon-rotation);
```
MartijnCuppens: Maybe make the whole transform value variable, so that flips can also be used?

mdo:
Assigning myself here to review the design after we get our first v5 alpha out, then this is in for alpha 2. Thanks!

scss/_variables.scss
```diff
@@ -959,6 +959,36 @@ $card-img-overlay-padding:          1.25rem !default;
...
+$accordion-body-padding-x:          $accordion-padding-x !default;
+$accordion-button-color:            $accordion-color !default;
+$accordion-button-bg:               $accordion-bg !default;
+$accordion-border-width:            $border-width !default;
+$accordion-border-radius:           $border-radius !default;
+$accordion-border-color:            rgba($black, .125) !default;
+$accordion-icon-color:              $accordion-color !default;
+$accordion-icon-active-color:       $accordion-button-active-color !default;
+$accordion-icon-transition:         $transition-base !default;
+$accordion-button-icon-rotation:    0deg !default;
```
XhmikosR: Is the unit really needed here?

site/content/docs/4.3/components/accordion.md
```diff
@@ -0,0 +1,139 @@
+---
+layout: docs
+title: Accordion
+description: Provides an accordion to collapse content vertically.
+group: components
+aliases:
```
XhmikosR: this isn't wrong, please remove it

XhmikosR: Or if you keep it, remove the redirects from the alerts page

scss/_variables.scss
gijsroge: Yeah not needed indeed, but like @MartijnCuppens said I will change the rotation to a generic transform property so the user can decide what he does with the icon instead of forcing them to use a rotation.

site/content/docs/4.3/components/accordion.md
XhmikosR: I remove the redirects from the alerts page and kept these ones.

gijsroge: Thanks @XhmikosR

mdo:
@gijsroge Would you be willing to resolve conflicts here and help bring this back?

mdo:
Picking this up in #32013.

gijsroge:
@mdo thanks for picking this up, I did not have the time due to personal stuff. I hope my PR was still somewhat useful :D

mdo:
It most assuredly was!

PR #30046, opened 2020-01-17
Set background hover color for each table color variant

### LINKED ISSUE #28160: Suggestion for .table-hover

}

Cinnead:
Instead of setting a background color on the tr tag, we set the background-color on the th or td tags below the tr tag. This way we can set the correct background-color for the th and td tags that also have color variant class (table-primary, table-secondary, ..)

closes #28160

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -106,6 +106,7 @@
     }
   }
 }
+// stylelint-enable selector-max-type
 
 
 // Table backgrounds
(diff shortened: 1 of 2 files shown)


scss/_tables.scss
```diff
@@ -86,11 +86,23 @@
 //
 // Placed here since it has to come after the potential zebra striping
 
+// stylelint-disable selector-max-type
```
XhmikosR: Please enable the rule again after the block you want to suppress the warning

Cinnead: Sorry about that, I have re-enabled the selector-max-type rule after the block.

mdo:
Punting on this for a `background-image: linear-gradient()` hack of sorts in #30342. Thanks!

PR #30048, opened 2020-01-17
Utility class border radius

### LINKED ISSUE #29989: Confusing classes `rounded-sm` and `rounded-lg`
Hello, 

JensLuyten:
Closes #29989

Updated the docs pages by searching for the `.rounded` classes and replacing them by the new `.border-radius` classes.

See 
https://deploy-preview-30048--twbs-bootstrap.netlify.com/docs/4.3/utilities/borders/#border-radius
https://deploy-preview-30048--twbs-bootstrap.netlify.com/docs/4.3/utilities/borders/#sizes

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -445,6 +445,7 @@ $utilities: map-merge(
     ),
     "border-radius": (
       property: border-radius,
+      class: br,
       responsive: true,
       values: (
         null: $border-radius,
@@ -458,22 +459,22 @@ $utilities: map-merge(
     ),
     "rounded-top": (
       property: border-top-left-radius border-top-right-radius,
-      class: rounded-top,
+      class: br-top,
       values: (null: $border-radius)
     ),
     "rounded-right": (
       property: border-top-right-radius border-bottom-right-radius,
-      class: rounded-right,
+      class: br-right,
       values: (null: $border-radius)
     ),
     "rounded-bottom": (
       property: border-bottom-right-radius border-bottom-left-radius,
-      class: rounded-bottom,
+      class: br-bottom,
       values: (null: $border-radius)
     ),
     "rounded-left": (
       property: border-bottom-left-radius border-top-left-radius,
-      class: rounded-left,
+      class: br-left,
       values: (null: $border-radius)
     ),
     "visibility": (
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -443,16 +443,17 @@ $utilities: map-merge(
       class: font,
       values: (monospace: $font-family-monospace)
     ),
-    "rounded": (
+    "border-radius": (
       property: border-radius,
-      class: rounded,
+      responsive: true,
       values: (
         null: $border-radius,
-        sm: $border-radius-sm,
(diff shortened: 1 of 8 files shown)


MartijnCuppens (dismissed):

scss/_utilities.scss
```diff
@@ -443,16 +443,17 @@ $utilities: map-merge(
       class: font,
       values: (monospace: $font-family-monospace)
     ),
-    "rounded": (
+    "border-radius": (
```
mdo: Thoughts from @twbs/css-review for using `radius`, `br`, or something else shorter here? If we go with `br`, I’d suggest we change `border` to  ` b` as well. Good example of this is `bg` for `background-color`.

MartijnCuppens: Yup, `br` and `b` make sense.

JensLuyten:
I have shortened the new `.border-radius` class to `.br` and updated the docs with the new information.

tpaksu:
Then "border radius bottom right" will be `br-br`, right? :)

JensLuyten:
> Then "border radius bottom right" will be `br-br`, right? :)

The border radius class will apply to all 4 corners, so there won't be a `br-br-2` class to just set the bottom right border radius with the current implementation.

tpaksu:
> 
> 
> > Then "border radius bottom right" will be `br-br`, right? :)
> 
> The border radius class will apply to all 4 corners, so there won't be a `br-br-2` class to just set the bottom right border radius with the current implementation.

Yes, I guess I saw the implementation in tailwindcss. That came to my mind and wanted to ask. BTW the clarification with `.border-radius` is gone with `.br`.  IMO, of course. If you want to add a line break class later for example .br { margin-bottom: 1em } this would clash. But there's also `.mb-3`, right. Well you know it better. Just my thoughts.

MartijnCuppens:
Not sure about the `br-top` syntax, in general we concat the properties (eg. `pt` for `padding-top`). This would mean we should use `brt` instead.

Not sure about that syntax either though. @mdo, what do you think?

ysds:
This PR does not include rename the border class name, but as @mdo commented above, the new name for border must also be considered. (This does not mean to include it in this PR.)
 
We should avoid conflicting `br` (border-radius and border-right). The best solution here is to match [Zen Cording](https://code.google.com/archive/p/zen-coding/wikis/ZenCSSPropertiesEn.wiki
). 

e.g.

* `bd-1` : `border: 1px solid $border-color`;
* `bdt-1`: `border-top: 1px solid $border-color;`
* `bdr-1`: `border-right: 1px solid $border-color;`
* `bdrs-1`: `border-radius: $border-radius-sm:`

This also compatible with the spacing classes.

ysds:
@JensLuyten Could you change the class names to `bdrs`?

mdo:
@ysds At that point, we might as well use `radius` instead of abbreviating :). I don't mind that fwiw.

mdo:
@JensLuyten Could you please take a pass at updating this to reflect the latest comments? We're backpedaling on the abbreviations and are opting for `radius` as the clearest option.

PR #30049, opened 2020-01-17
Prevent grid with default cols from breaking when large pre is present

### LINKED ISSUE #25410: Pre with horizontal scrollbar causes grid to stack
If a pre has a horizontal scrollbar it will stack thus causing the grid to malfunction.

**Operating system**
Windows  7 (Microsoft Windows [version 6.1.7601])

**Browser and version**
Chrome 61.0.3163.100

**Reduced test cases**
https://codepen.io/stephenbe/pen/zpXWyy

In this codepen, you can see that the first examples stack while the other without a horizontal scrollbar it works just fine.

Thanks!

JensLuyten:
Closes #25410 (Codepen with bug present in issue)

Tested this in Codepen: https://codepen.io/jensluyten/pen/povxyPe. 
The `min-width: 0;` on the `.col` actually prevents the col from taking up the entire row space.

This was only an issue on the default cols, because specified cols have a fixed `flex-basis` property that prevents this issue from happening.


MartijnCuppens (approved):
Excellent, thanks!

XhmikosR:
Does this apply to v4 @MartijnCuppens? if so please add the label in the PRs you approve.

choldgraf:
Did this one get re-introduced? I noticed this behavior on 4.6.1 and saw this section had been changed:

https://github.com/JensLuyten/bootstrap/blob/af78de3cddb25d6fe3b3549cfe6812b6f7812363/scss/mixins/_grid-framework.scss#L24-L27

and didn't see any min-width here:

https://github.com/JensLuyten/bootstrap/blob/af78de3cddb25d6fe3b3549cfe6812b6f7812363/scss/mixins/_grid-framework.scss#L8-L13

PR #30050, opened 2020-01-17
Add RFS to utility API

MartijnCuppens:
In https://github.com/twbs/bootstrap/pull/29152, I tried combining too many things in one PR.

In this PR I just enabled the possibility to use RFS in the utility API. We 'll' see how we'll deal with the responsive font size and responsive spacing classes afterwards.

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -23,6 +23,8 @@ html {
 @import "mixins/grid";
 @import "mixins/utilities";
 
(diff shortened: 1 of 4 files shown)


mdo (approved):
Haven't tested this out myself, but I'm in favor of enabling the RFS scaling on utilities as folks want via the utility API.

MartijnCuppens:
Rebased & retested this

PR #30051, opened 2020-01-17
Add ::after element for .bd-navbar to make sure alignation is correct

### LINKED ISSUE #29856: IE: docs: header doesn't seem to be aligned

Cinnead:
There's a problem in IE when min-height and align-items are combined, this breaks the alignation of the header on the docs.

This code fixes it by adding an ::after as workaround

closes #29856


mdo:
Looks like #29948 will fix this without adding an pseudo-element. Thanks though!

PR #30066, opened 2020-01-23
Remove the unused babel-plugin-istanbul devDependency.

XhmikosR:
I need to confirm if it's redundant in v4-dev later.


XhmikosR:
CI is green so this does seem to be unused.

MartijnCuppens (approved):

PR #30067, opened 2020-01-23
ARIA roles in docs and examples

ffoodd:
Questionning documentation markup with accessibility in mind, I came across [Scott O'Hara's recent post about implicit landmarks](https://www.scottohara.me/blog/2019/04/05/landmarks-exposed.html) which basically says:

- `main` role is not necessary anymore since both NVDA and Jaws fill the gap in IE11;
- `contentinfo` role is useful (albeit not necessary) for macOS VoiceOver with WebKit / Chromium based browsers.

I also added the `search` role to the search form since it's very useful as a landmark, and obviously updated the `vnu-jar.js` ignored warning to match what's been done on markup.

There isn't any edge cases as far as I'm aware of, but it's still useful to share [the last WebAim Screen-reader users survey (anchored to the browser combos section)](https://webaim.org/projects/screenreadersurvey8/#browsercombos).

I guess @patrickhlauke should review this?


XhmikosR:
I don't think we need the contentInfo back.

XhmikosR:
I mean, this is a clear Webkit issue and we should pursue to be fixed instead of this workaround.

ffoodd:
The `main` role was redundant for the exact same reason, so I strongly disagree with the argument: we're talking about real users, not browsers.

However `contentinfo` only concerns additional content, so it's not essential. Since I used atomic commit, it'll be easy to only revert this part — so if @patrickhlauke agrees, let's revert this part. :)

XhmikosR:
>The main role was redundant for the exact same reason, so I strongly disagree with the argument: we're talking about real users, not browsers.

Well, IE is not the same as the other browsers :P

Milad-Abooali (approved):

patrickhlauke:
IE is still supported https://getbootstrap.com/docs/4.4/getting-started/browsers-devices/ so `role="main"` should probably stay.

identifying the main part of a page is more important for users than identifying the footer - historically, the biggest concern has always been for users to be able to easily find the main section. not so much about users needing to easily find the/a footer. so for `role="contentinfo"` i'd say it's not that important to patch it...but not a hill i'd die on. belt-and-braces doubling up is not really a problem in my view, and if we feel it does help some people...why not.,

ffoodd:
> IE is still supported https://getbootstrap.com/docs/4.4/getting-started/browsers-devices/ so `role="main"` should probably stay.

@patrickhlauke I'm aware of that; I'm refering to [Scott O'Hara's recent post "Landmarks exposed"](https://www.scottohara.me/blog/2019/04/05/landmarks-exposed.html) explainging that IE11 & Edge with both NVDA or JAWS now expose `main` landmark correctly — despite the fact that browsers didn't fix anything, ATs did.

Do you still want to get the explicit role back? It's a single commit to revert, please let me know :)

patrickhlauke:
reliant on latest AT. anyway, you do what you think is best, i just feel that it's such a small inconsequential thing that doesn't do any harm to keep

ffoodd:
I understand. You're right in that it doesn't do any harm — the only drawback is that it requires ot ignore a warning in HTML Validator (so I did).

Let's go with this, then.

build/vnu-jar.js
```diff
@@ -37,7 +37,13 @@ childProcess.exec('java -version', (error, stdout, stderr) => {
     'The “month” input type is not supported in all browsers.*',
     'The “color” input type is not supported in all browsers.*',
     'The “datetime-local” input type is not supported in all browsers.*',
-    'The “time” input type is not supported in all browsers.*'
+    'The “time” input type is not supported in all browsers.*',
+    // IE11 doesn't recognize <main> / give the element an implicit "main" landmark.
+    // Explicit role="main" is redundant for other modern browsers, but still valid.
+    'The “main” role is unnecessary for element “main”.',
```
XhmikosR: This is wrong now since IE is gone

ffoodd: Mistake while rebasing, my bad!

XhmikosR:
@patrickhlauke the final call is yours :) I'm not 100% sure it makes sense to support such workarounds, unless we've tried bugging upstream to fix the issues there and there's no response.

Also, we need to make sure our [current supported browsers](https://github.com/twbs/bootstrap/blob/master/.browserslistrc) are affected by this. If you read https://www.scottohara.me/blog/2019/04/05/landmarks-exposed.html it says:

>Please note that there the following information may no longer be accurate if testing with newer versions of browsers and screen readers than the ones noted in this article. For instance, since Safari 13 was released, VoiceOver will now correctly expose a footer a content information landmark.

ffoodd:
@XhmikosR @patrickhlauke  Up on this one: very low importance, however it's the last one coming from my fork (which I'd be proud to delete!).

It basically only adds `contentinfo` role on `<footer>`.

XhmikosR:
https://github.com/twbs/bootstrap/pull/30067#issuecomment-602512092

patrickhlauke:
ok, although i'm really not too fussed about this, my current thinking - take it for what it is...

as we're now ripping other more fundamental stuff out (in various places) that will end up actually breaking IE11, we can probably also do away with `role="main"`.

as for the `role="contentinfo"`, @scottohara rightly pointed out that if our supported browsers includes "Safari >= 10" (per https://github.com/twbs/bootstrap/blob/master/.browserslistrc) then that fix *should* be there. or we can modify the browser list to go for >= 13 onwards.

ffoodd:
I think only supporting Safari 13 would be quite too much, since it's pretty young (released in last september).

However until this PR isn't approved, it's open for discussion :)

patrickhlauke (approved):

XhmikosR:
I guess we can land it although it's not valid and we'll probably forget to remove it later, but, well.

ffoodd:
Validation is about syntax; this won't cause any issue. The `main` role was implemented this way and ignored from the Validator too.

Do you want me to leave a `@todo` comment about this? The switch here is already known: as soon as we stop supporting Safari previous to 13, we'll be safe removing this.

I don't know if there's some kind of reminder for things?

patrickhlauke:
not that it's not valid. defining the same role on an element is redundant, but not disallowed. it's the validator not differentiating between hard failures and "shouldn't need to do it"

ffoodd:
Let's dump this one, it's a very small improvement for an even smaller part of our users and end users.

I don't think it worths it.

PR #30071, opened 2020-01-23
Drop `.booticon` styles

ffoodd:
Flagged as `Temp` by @mdo, I couldn't find any occurrence of this in the entire repo.

Should be safe to remove, shouldn't it?


MartijnCuppens (approved):
Yup!

XhmikosR:
Yup, it was added because in the beginning the icons were supposed to be part of this repo.

XhmikosR (approved):

PR #30073, opened 2020-01-23
Documentation skip links

ffoodd:
Mostly usability improvements — at least some tries:

- new skip-link to the main navigation, in documentation page;
- new skip-link to search form (on pages that display it);
- refactored skip-links pattern, mainly to ease their use for screen magnifiers users (which can't see them appearing since when using it, they are likely to be on the top left corner of the document).

This is more convenient to use, and also simplify both markup and styles — however maybe opinionated?

Each of these is a separate commit so let me know if I should revert one of those.

Ping @patrickhlauke again :)

Preview: <https://deploy-preview-30073--twbs-bootstrap.netlify.com/>

### DIFF
--- a/site/assets/scss/_skippy.scss
+++ b/site/assets/scss/_skippy.scss
@@ -6,4 +6,8 @@
   padding: map-get($spacers, 2);
   color: $white;
   background-color: $bd-purple;
+
+  &:hover {
+    color: $white;
+  }
 }
--- a/site/assets/scss/_skippy.scss
+++ b/site/assets/scss/_skippy.scss
@@ -1,9 +1,9 @@
 .skippy {
   position: fixed;
-  top: map-get($spacers, 2);
-  left: map-get($spacers, 2);
+  top: .5rem;
+  left: .5rem;
   z-index: $zindex-fixed;
-  padding: map-get($spacers, 2);
+  padding: .5rem;
   color: $white;
   background-color: $bd-purple;
 
--- a/site/assets/scss/_skippy.scss
+++ b/site/assets/scss/_skippy.scss
@@ -1,17 +1,9 @@
 .skippy {
-  display: block;
-  padding: 1em;
+  position: fixed;
+  top: map-get($spacers, 2);
+  left: map-get($spacers, 2);
+  z-index: $zindex-fixed;
+  padding: map-get($spacers, 2);
   color: $white;
-  text-align: center;
   background-color: $bd-purple;
-  outline: 0;
-
-  &:hover {
-    color: $white;
-  }
-}
-
-.skippy-text {
-  padding: .5em;
-  outline: 1px dotted;
 }
--- a/site/content/docs/4.3/examples/blog/index.html
+++ b/site/content/docs/4.3/examples/blog/index.html
@@ -11,13 +11,13 @@ include_js: false
   <header class="blog-header py-3">
     <div class="row flex-nowrap justify-content-between align-items-center">
       <div class="col-4 pt-1">
-        <a class="text-muted" href="#">Subscribe</a>
+        <a class="link-secondary" href="#">Subscribe</a>
       </div>
       <div class="col-4 text-center">
         <a class="blog-header-logo text-dark" href="#">Large</a>
       </div>
       <div class="col-4 d-flex justify-content-end align-items-center">
-        <a class="text-muted" href="#" aria-label="Search">
+        <a class="link-secondary" href="#" aria-label="Search">
           <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="mx-3" role="img" viewBox="0 0 24 24"><title>Search</title><circle cx="10.5" cy="10.5" r="7.5"/><path d="M21 21l-5.2-5.2"/></svg>
         </a>
         <a class="btn btn-sm btn-outline-secondary" href="#">Sign up</a>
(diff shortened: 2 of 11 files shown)


Milad-Abooali (approved):

MartijnCuppens:
- I don't really like the theming changes, any reasons why you did this? Also don't remove the hover state, the link is now blue when hovered.
- Combining skippy and skippy-text is a good idea though
- Any reason why you don't link to the search input field itself?

ffoodd:
For the theming, I only moved them to the top left corners, which is generally where it's used by screen magnifiers users. It's not an absolute rule obviously, just a kind of convention.

As a matter of fact, previous theming using full width colored background might be a sufficient clue I guess, however I'd recommend obviousness to visual clues, when speaking about visual deficiencies.

I'm fixing the hover color, missed that; and you're right about linking to the search input, would be better!

Commits on their way, let me know if I should revert the theming part.

MartijnCuppens:
Ping @twbs/team for a second opinion on the theming changes. Personally I don't the blue outline we have now:
image

ffoodd:
@MartijnCuppens Those are default outlines from your browser (Chrome, I guess?) — and should be the same in the navbar, for example. 

Should I customize it? They don't seem to be customized for links at all in `_reboot.scss`…

XhmikosR:
@ysds what's your take on this?

ysds (approved):
I don't like UIs that cover logos because their identities seem to be ignored, but there is no rational reason to deny this proposal. The colors (bg, text and outlines) look good for me.

mdo:
Is this a common practice? I haven't seen this many tabs in front of my ability to get to a clearly focusable element. Also, tabbing to the search form applies the focus style, but it's not focused with my cursor—at least in Safari, I cannot type in the field with tab-tab-tab-enter. Feels like an incomplete implementation with that.

I'd also echo the feedback of this being a little tough to see given the colors. It matches, yes, but it could be more obvious.

XhmikosR:
I agree with the search focus comment. I too would expect it to focus the search input.

I like the change in the skip to main content part but only for desktop. For mobile it hides the logo. Not a big deal, and TBH, this might be a better approach since the page does not jump when the links show up?

ffoodd:
You're right about the search input, I guess it's somehow related to Algolia. I'll investigate on this.

About the design thing, let me know if you have any ideas: FYI I didn't change anything on the design part, except links being above the header instead of making it jumppy (and this is still easy to revert).

I seriously think it's more common to have skip links fixed above the header than moving around. Above the logo is not a concern to me, since people got to your page and saw it… I guess their comfortable usage of the page is more important than always see your logo :D 

The main goals of this PR are:
1. get skip links left aligned, since people using a screen magnifier start on the top left, whatever their screen resolution is;
2. add other skip links to ease navigating: to the search input, and to the documentation subnav.

Styling skip links wasn't to my mind so please elt me know if you think I can improve this too.

For now it's at least missing a correct behaviour for the search input getting focus.

ffoodd:
Hi there,

I'm getting very different and disappointing behaviours on this. It seems to work fine in Chrome, but not at all in Firefox — and I only tested those two, I guess trying with Edge, Safari, or mobile browsers would get things worse. I made [a reduced test case on CodePen](https://codepen.io/ffoodd/pen/VwLaGjE?editors=1111).

At this point, the skip to search input link would require some JavaScript to properly get the focus :/ I found several posts mentionning this kind of issues with skip links / anchors, all of them ending up with a JavaScript patch…

Since there's not so much tabs to get through before reaching the search input, I'm not sure it's worth it.

I'd recommend to simply remove the skip to search link (and still standing by for the design part) — do you have any opinion on this?

XhmikosR:
I'm still not sure if the second skip to main nav has any value since the nav it's in the content already.

/CC @mdo @patrickhlauke for your final words

patrickhlauke:
i think the additional skip link is quite useful for kbd users as, when navigating through the docs, it allows for reasonably quick access to the side navigation there without needing to constantly tab through the top navigation stuff.

i wonder though if rather than "main navigation", it should say "docs navigation" (i.e. "skip to docs navigation") as arguably the main navigation throughout the site is the one at the top? (though i see the `<nav>` itself also has an `aria-label="Main Navigation"`, so it's consistent with that i guess)

also, a complete aside - is there any issue opened about the disclosure widgets in the side navigation there not having `aria-expanded` and not having a button role?

XhmikosR:
> also, a complete aside - is there any issue opened about the disclosure widgets in the side navigation there not having `aria-expanded` and not having a button role?

The thing is that the docs don't use our core collapse plugin for the sidebar (I didn't implement this and it must got merged without noticing it). So, it's very basic and obviously lacks some stuff. I tried to switch to the collapse plugin for the sidebar, but I gave up since it didn't work properly.

ffoodd:
@XhmikosR @patrickhlauke I think I managed to make collapse work on sidenav: could you please check?

XhmikosR:
@ffoodd please try to keep your PRs clean, simple and short. Squash any patches when possible, don't make more than 1 core change. Please move the discussion about the sidebar to a new PR.

patrickhlauke:
there seems to be some slight delay now between activating the disclosure control and the submenu appearing.

additional complication/consideration: when in smaller screen view, where the secondary navigation is collapsed behind the hamburger menu, the "skip to docs navigation" won't work as expected when the hamburger is currently closed.

ffoodd:
@patrickhlauke Collapse behaviour is handled in #30230 from now on.

For skip to docs nav, maybe could we simply hide it on smaller screen with utility classes?

patrickhlauke:
> For skip to docs nav, maybe could we simply hide it on smaller screen with utility classes?

i guess that would be the least bad alternative, yes

ffoodd:
This is done. :)

XhmikosR:
@mdo @patrickhlauke: do you think this is ready to land in v5?

patrickhlauke:
@XhmikosR @ffoodd this looks good, yes.

on a completely separate note...the hamburger button on small screen lacks any kind of focus indication. is that already noted/addressed somewhere?

small screen docs navigation with hamburger menu button

XhmikosR:
>on a completely separate note...the hamburger button on small screen lacks any kind of focus indication. is that already noted/addressed somewhere?

Nope, we should fix it in a separate PR targeting master, good catch!

@ffoodd please rebase and clean up your patches when you have some time :)

Still wondering if we should backport this. Thoughts?

patrickhlauke:
> Nope, we should fix it in a separate PR targeting master, good catch!

i'll file a fresh issue

patrickhlauke:
> Still wondering if we should backport this. Thoughts?

would be nice if we could, but not a deal-breaker if we'd rather concentrate on more pressing matters

XhmikosR:
I added in the 4.4.2 project to track this and we can backport it later

ffoodd:
@XhmikosR Rebased!

XhmikosR (approved):

site/assets/scss/_skippy.scss
```diff
@@ -1,17 +1,13 @@
 .skippy {
-  display: block;
-  padding: 1em;
+  position: fixed;
+  top: map-get($spacers, 2);
```
XhmikosR: Actually, @ffoodd can you remove `map-get`s and use the values directly? We don't do this anywhere else.

XhmikosR (changes requested):

XhmikosR (approved):

XhmikosR:
@MartijnCuppens @ffoodd so what's it going to be with this? Currently we have the patch on master, and I'm tracking in v4-next, but I recall we had some second thoughts?

ffoodd:
Indeed, I'll try something else today.

PR #30079, opened 2020-01-24
Use `aria-current` where appropriate

ffoodd:
`aria-current` attribute has many advantages and a single drawback compared to visually hidden text like `<span class="sr-only">(current)</span>` — let me explain:

- [it's standardized as part of ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#aria-current)
- [it's widely supported by browsers and screen readers](http://design-patterns.tink.uk/aria-current/)
- it's interoperable and programmatic, so can be usedby any tool reading the markup
- internazionalition and localization is not a thing, assistive technologies take care of it
- it allows to fine tune the context with its diffrent values (`page`, `true`, `date`, `step`, `location`, etc.)
- it could be used as a style hook by using `[aria-current]`, thus replace the `.active` state (didn't push this much, but could if you like the idea) — and also as a JavaScript hook, obviously.

The single drawback is that is does not appear when CSS is missing, whereas the visually hidden text does.

I think `aria-current` wins, so here's a full PR making a few different things:
- add `aria-current` in components already using `.active` (only if appropriate, obviously — so not in JavaScript tabbed interface, for example) ;
- mention `aria-current` beside `.active` in thos components' documentation;
- update examples templates accordingly;
- apply `aria-current` in docs navigations as well (main nav, sidenav, versions dropdowns).

And as usually, each of those lives in its very own commit.


site/content/docs/4.3/components/dropdowns.md
```diff
@@ -560,12 +560,13 @@ You can also create non-interactive dropdown items with `.dropdown-item-text`. F
 
 ### Active
 
-Add `.active` to items in the dropdown to **style them as active**.
+Add `.active` to items in the dropdown to **style them as active**
+Regarding accessibility, the active state should also be indicated with the `aria-current` attribute — using the `page` value for current page, or `true` for current item in a set.
```
XhmikosR: This doesn't add a newline. It might be better to add a `.` and continue on the same line

site/layouts/partials/docs-sidebar.html
```diff
@@ -16,16 +16,16 @@
     {{- $active_group := eq $.Page.Params.group $group_slug }}
 
     <li class="bd-sidenav-group my-1{{ if $active_group }} active{{ end }} js-sidenav-group{{ if $group.pages }} has-children{{ end }}">
-      <a class="d-inline-flex align-items-center bd-sidenav-group-link" href="/docs/{{ $.Site.Params.docs_version }}/{{ $group_slug }}/{{ if $group.pages }}{{ $link_slug }}/{{ end }}">
-        <div>{{ $group.title }}</div>
```
XhmikosR: I remember I saw this before and I had removed but the patch probably never landed. I can't find any reason to keep the `div` here.

ffoodd: Neither do I, I searched across scripts and styles, at every breakpoints, and didn't find any usage of this `div`…

site/content/docs/4.3/components/dropdowns.md
```diff
@@ -560,12 +560,12 @@ You can also create non-interactive dropdown items with `.dropdown-item-text`. F
 
 ### Active
 
-Add `.active` to items in the dropdown to **style them as active**.
+Add `.active` to items in the dropdown to **style them as active**. Regarding accessibility, the active state should also be indicated with the `aria-current` attribute — using the `page` value for current page, or `true` for current item in a set.
```
patrickhlauke: Sounds a bit stilted. Suggest something more like

"Regarding accessibility, the active state should also be indicated with" > "To convey the active state to assistive technologies, use"

"for current" > "for the current"

site/content/docs/4.3/components/navs.md
```diff
@@ -14,12 +14,14 @@ The base `.nav` component is built with flexbox and provide a strong foundation
 
 {{< callout info >}}
 The base `.nav` component does not include any `.active` state. The following examples include the class, mainly to demonstrate that this particular class does not trigger any special styling.
+
+Regarding accessibility, the active state should also be indicated with the `aria-current` attribute — using the `page` value for current page, or `true` for current item in a set.
```
patrickhlauke: "Regarding accessibility, the active state should also be indicated with" > "To convey the active state to assistive technologies, use"

site/content/docs/4.3/components/navs.md
```diff
@@ -245,6 +247,8 @@ If you're using navs to provide a navigation bar, be sure to add a `role="naviga
 
 Note that navigation bars, even if visually styled as tabs with the `.nav-tabs` class, should **not** be given `role="tablist"`, `role="tab"` or `role="tabpanel"` attributes. These are only appropriate for dynamic tabbed interfaces, as described in the [<abbr title="Web Accessibility Initiative">WAI</abbr> <abbr title="Accessible Rich Internet Applications">ARIA</abbr> Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel). See [JavaScript behavior](#javascript-behavior) for dynamic tabbed interfaces in this section for an example.
 
+On the other hand, the `aria-current` attribute is useless on dynamic tabbed interfaces since our JavaScript handles the selected state by adding `aria-selected="true"` on the active tab.
```
patrickhlauke: "useless" > "not necessary"

patrickhlauke (changes requested):

ffoodd:
@patrickhlauke I made the requested changes, thanks for review :)

site/content/docs/4.3/components/navs.md
```diff
@@ -245,6 +247,8 @@ If you're using navs to provide a navigation bar, be sure to add a `role="naviga
 
 Note that navigation bars, even if visually styled as tabs with the `.nav-tabs` class, should **not** be given `role="tablist"`, `role="tab"` or `role="tabpanel"` attributes. These are only appropriate for dynamic tabbed interfaces, as described in the [<abbr title="Web Accessibility Initiative">WAI</abbr> <abbr title="Accessible Rich Internet Applications">ARIA</abbr> Authoring Practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel). See [JavaScript behavior](#javascript-behavior) for dynamic tabbed interfaces in this section for an example.
 
+On the other hand, the `aria-current` attribute is not necessary on dynamic tabbed interfaces since our JavaScript handles the selected state by adding `aria-selected="true"` on the active tab.
```
patrickhlauke: "On the other hand, " feels a bit out of place here. What's the one hand? i.e. this would make sense if in the previous sentence you'd just said "you must use `aria-current`. 

Anyway, I think it would make most sense to just start the sentence with "The `aria-current` attribute is not necessary...". could even add it to the end of the previous paragraph, rather than starting a new paragraph.

patrickhlauke (changes requested):

patrickhlauke (approved):

ihorzenich:
@ffoodd Keeping `<span class="sr-only">(current)</span>` together with `aria-current="page"` on parent cause double announcement of "current" page in NVDA screen-reader.
image

It is necessary to remove `<span class="sr-only">(current)</span>` from docs.

patrickhlauke:
@ihorzenich would have been better to open a new issue instead of necro-posting to a PR from february... https://github.com/twbs/bootstrap/issues/31891

PR #30080, opened 2020-01-24
Headings hierarchy in licence.md

ffoodd:
Very simple and specific improvement to the licence page, which stepped from `h1` to `h4`.

FYI, I also noticed a few other headings hierarchy failures when using some callouts, in which heading level is defined in Markdown without taking the context of the page in which they're included.

For example, calling `callout-danger-async-methods.md` after a `h2` will break headings hierarchy by stepping to `h4`.

I tried a bit but struggled with Hugo specificities: we're include a Markdown partial through a partial shortcode (why? is it only to pass context?) inside a callout shortcode. So passing an argument to handle heading level in partial isn't easy.

I don't know if I should try to handle arguments in the `partial` shortcode, or if I should move those callouts to `.html` and calling them directly with the `partial` method (without the partial shortcode in the middle).

I can't help more with this, but at least the Licence page is easy to imrpove :)

### DIFF
--- a/site/content/docs/4.3/about/license.md
+++ b/site/content/docs/4.3/about/license.md
@@ -7,18 +7,18 @@ group: about
 
 Bootstrap is released under the MIT license and is copyright {{< year >}} Twitter. Boiled down to smaller chunks, it can be described with the following conditions.
 
-#### It requires you to:
+## It requires you to:
 
 * Keep the license and copyright notice included in Bootstrap's CSS and JavaScript files when you use them in your works
 
-#### It permits you to:
+## It permits you to:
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Yeah, the headings hierarchy is something I did try to tackle in the past, but I didn't consider it worth it at the time, especially since we need to change so many places.

#28487

I'd say just don't bother with it, it's not so important. We first need to solve some other Hugo-related issues anyway.

ffoodd:
Seems legit. I'll keep an eye on this though, in case of an incoming brilliant idea :D

XhmikosR (approved):

PR #30081, opened 2020-01-24
Improve reduced-motion case for spinners

ffoodd:
Last but not least, this is a real feature request and not another docs enhancement.

Basically, it ensures that spinners honor the `prefers-reduced-motion` feature by stopping the animation — which kind of defeat the need for a visual spinner. The idea I came up with is to display the visually hidden text instead.

So it then litterally displays the `Loading…` text instead of the animated thing.

### DIFF
--- a/scss/_spinners.scss
+++ b/scss/_spinners.scss
@@ -53,3 +53,21 @@
   width: $spinner-width-sm;
   height: $spinner-height-sm;
 }
+
+@if $enable-prefers-reduced-motion-media-query {
+  @media (prefers-reduced-motion: reduce) {
+    .spinner-border,
+    .spinner-grow {
+      min-width: 10ch;
+      border: 0;
+      animation: none;
+
+      // `all` is safe here, since its support is better than the `prefers-reduced-motion` query's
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
I don't really like to see the `.sr-only` is overridden, also `.sr-only-focussable` should be taken into account. On the other hand this construction makes it hard to theme the `.sr-only` element, even with utilities (specificity of `1,0,2,0` vs `1,0,1,0`)

Also, I doubt this is really needed, quoting @patrickhlauke (https://github.com/twbs/bootstrap/pull/27986#issuecomment-454182392):
> I'd note that reduced motion is primarily about stuff that can cause trouble (read: feelings of nausea etc) to users with vestibular disorders, so mainly animated scrolling effect, accordion-type panels that animate open/close, etc.

The spinner animation is rather subtle.

ffoodd:
I understand, and globally agree with your arguments.
This will probably land in our fork so I preferred to suggest this directly here :)

Let me know if it still sounds interesting and needs a rework, though.

MartijnCuppens:
Gonna close this for the reasons I mentioned above. Thanks though

Issue #30062, opened 2020-01-22
<blockquote> tag being specified incorrectly


Glenn-Domin:
Wasn't sure quite where to report this, but it appears your HTML `<blockquote>` [example](https://getbootstrap.com/docs/4.4/content/typography/#blockquotes) isn't technically correct:
```
<blockquote class="blockquote">
  <p class="mb-0">Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere erat a ante.</p>
  <footer class="blockquote-footer">Someone famous in <cite title="Source Title">Source Title</cite></footer>
</blockquote>
```
[The spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-blockquote-element) for `<blockquote>` states:
> Attribution for the quotation, if any, must be placed outside the blockquote element.

There are two examples of how you could do this correctly in the spec:
```
<blockquote>
 <p>I contend that we are both atheists. I just believe in one fewer
 god than you do. When you understand why you dismiss all the other
 possible gods, you will understand why I dismiss yours.</p>
</blockquote>
<p>— Stephen Roberts</p>
```
and
```
<figure>
 <blockquote>
  <p>The truth may be puzzling. It may take some work to grapple with.
  It may be counterintuitive. It may contradict deeply held
  prejudices. It may not be consonant with what we desperately want to
  be true. But our preferences do not determine what's true. We have a
  method, and that method helps us to reach not absolute truth, only
  asymptotic approaches to the truth — never there, just closer
  and closer, always finding vast new oceans of undiscovered
  possibilities. Cleverly designed experiments are the key.</p>
 </blockquote>
 <figcaption>Carl Sagan, in "<cite>Wonder and Skepticism</cite>", from
 the <cite>Skeptical Inquirer</cite> Volume 19, Issue 1 (January-February
 1995)</figcaption>
</figure>
```
Personally I like the latter, though it's a bit verbose.
To be clear, the `<footer>` element can be used inside a `<blockquote>`, but not for the purpose of attribution.

MartijnCuppens:
Thanks! Feel free to open a PR to fix this.

ffoodd:
I'm going on a few PR in the docs, I can handle this one if you want, @Glenn-Domin?

tmorehouse:
I prefer the latter example with the figure/figcaption, as it automatically associates the attribution to the blockquote content (great for screen reader users)

mdo:
Fixed by #30814.

PR #30082, opened 2020-01-24
Blockquote sources

ffoodd:
Starting to work on #30062, still missing to fix the examples templates and probably double-check the rendering after this change.

@patrickhlauke Regarding accessibility, I had in mind that `figure` needed a  `group` role and implemented it that way: am I right?


ffoodd:
@Glenn-Domin @MartijnCuppens Sorry for the duplicate PR, I just saw #30085 (and obviously started working on this one earlier…).

I'll comment on #30085 for discussion :)

PR #30085, opened 2020-01-25
fixing blockquote footer examples HTML

Glenn-Domin:
@MartijnCuppens this is in relation to issue #30062. It amends the relevant HTML on the card, typography and reboot pages. This rectifies the technically incorrect `blockquote` HTML mentioned in the issue and replaces it with a `figure` example.

While appearance has been maintained in card and typography, the one on the reboot page appears slightly different because the current example was missing the `blockquote-footer` class.

I've also tweaked the language slightly in typography to suit the new code. Feel free to change if a different form of language is preferred.


MartijnCuppens:
Since we have dedicated classes here, we can make use of negative margins to fix the margins instead of using utilities, like https://codepen.io/MartijnCuppens/pen/ZEYPvqm.

Thoughts?

Glenn-Domin:
> we can make use of negative margins to fix the margins instead of using utilities

If you were starting from scratch I might agree with that. The problem there is that loads of people have coded their blockquotes with the `blockquote-footer` style as it currently exists. If we add a negative `margin-top` to it that may break appearance on many sites when they go to update Bootstrap.

Another consideration (I'm not sure if you want to cater to this crowd) is that many developers - without checking documentation - have a similar `blockquote` mental model to what's in the Bootstrap docs currently, and would do something like:
```
<blockquote class="blockquote">
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere erat a ante.
  <footer class="blockquote-footer">Someone famous in <cite title="Source Title">Source Title</cite></footer>
</blockquote>
```
In this instance, a negative top margin would cut in to the `blockquote` content.

Personally if I were putting this on a site (which I am, which is what drew this issue to my attention), in my template file I'd be coding the `figure`, `blockquote` and `figcaption` tags, then the content the client has entered in the back-end would populate it. I might do something like
```
.blockquote p:last-child {
  margin-bottom: 0;
}
```
That way the client wouldn't have to touch HTML and it works regardless of whether or not there's footer content. It would, however, be contingent on the WYSIWYG generating the `p` tags correctly.

ffoodd:
Hi there, while working on #30082 I came up with the exact same solution as @Glenn-Domin using utilities.

The only way I can imagine to solve this would be to use `figure.blockquote-footer` for this negative margin things, but not quite sure of it. @MartijnCuppens what's your opinion, can it solve the spacing issue?

Also @Glenn-Domin, there are other occurrences of blockquote in examples/masonry.

@MartijnCuppens I also came across a visual issue with moving the `blockquote-footer`, since it's sized in `em` and its context changes. I updated the relevant variables `$blockquote-small-font-size` to limit the difference — however since it was computed to `17.5px` I chose to stick to a `18px` equivalent.

I'll let you decide what to do with my PR, I'd be pleased to help on this one instead :)

Glenn-Domin:
Thanks @ffoodd for pointing out the masonry file. I've amended the PR with changes to that file.

@ffoodd the issue you're experiencing in your code relates to breaking the `blockquote-footer` class out of  the `blockquote` parent class. I'd advise you to keep `blockquote` as the parent class and `blockquote-footer` as the child. You'll avoid side-effects while still keeping the footer's font-size consistently smaller relative to the `blockquote` text. Changing the behaviour of the Sass variable could have adverse side-effects for existing projects that rely on `$blockquote-small-font-size` being tied to `$small-font-size`. I think if potential breaking changes can be avoided I'd prefer it.

I don't see an issue with using utility classes for editing the margins as they're frequently used in the existing examples. I think there's a much lower risk of misuse than with negative margins, which in my opinion should be strictly controlled due to the risk of overlapping content.

ffoodd:
Totally agree!

Fine for me, I guess our PRs will have the same result so let's go with yours!

mdo:
Addressed in #30814 in v5.

PR #30086, opened 2020-01-25
Fix file input

MartijnCuppens:
Broken: https://twbs-bootstrap.netlify.com/docs/4.3/forms/validation/#supported-elements

Fixed: https://deploy-preview-30086--twbs-bootstrap.netlify.com//docs/4.3/forms/validation/#supported-elements

### DIFF
--- a/site/content/docs/4.3/forms/validation.md
+++ b/site/content/docs/4.3/forms/validation.md
@@ -303,7 +303,10 @@ Validation styles are available for the following form controls and components:
 
   <div class="form-file">
     <input type="file" class="form-file-input" id="validationFormFile" required>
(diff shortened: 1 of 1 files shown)


XhmikosR:
When did this break exactly? Just to double check things.

XhmikosR (approved):

MartijnCuppens:
We switched from a pseudo element button to a `.form-file-button` and forgot to adjust the documentation here.

XhmikosR:
Can you check if there are more cases?

MartijnCuppens:
I did (searched for `custom-file` occurrences and this was the only place we missed)

XhmikosR:
BTW it would be nice if you referenced the commit or PR something broke in the future :)

PR #30088, opened 2020-01-26
change max container width to 1800px

gumtow:


MartijnCuppens:
Larger containers will make it harder to read text, passing on this

Issue #28581, opened 2019-03-31
Better link in documentation to the latest release


matronator:
One thing that's been driving me crazy ever since I started using Bootstrap was the fragmenting of documentation across multiple versions on Google. Every time you want to quickly lookup some component or whatever on Google, you get several results of the same page but from a different Bootstrap version and even worse most of the time the latest (stable) one isn't even in the results.

As annoying as this is, I can see the reason in keeping those older docs up in the results, as not everyone is always using the latest version. However what really gets me every time is when I click on the sticky top link in hopes of getting to the same page but on the latest release and end up on the main page instead, having to find the page again in the documentation

 I've done it so many times already but it doesn't matter, because the next time I land on an older version of the docs I will forget all about it and end up clicking the link anyways.

What is the reasoning behind it behaving the way it does? It's not very intuitive IMO. The link text doesn't give any indication where exactly you'll be taken, other than it will have something to do with a newer version of (currently) Bootstrap 4 and is kinda misleading that way.

Wouldn't it be more logical for the link to lead to the same page on the newer version, if it exists, otherwise the main page or changelog/release notes? Definitely more convenient at least.

coliff:
This issue has cropped up many times, and I too find it frustrating. If you Google 'Bootstrap 4 docs' 2 of the 3 first results are for v4.0 and v4.1 - and none for the latest v4.3. This leads some users to start using older versions of the docs and probably leads to extra 'support issues' being opened with users using old versions with bugs/features which have since been addressed.

@XhmikosR mentioned in the Slack channel the possibility of blocking Google from indexing the old versions of the docs with `robots.txt`. I think this would be the best option.
So it'd block:
```
/docs/4.2/
/docs/4.1/
/docs/4.0/
/docs/3.3/
```
Then if someone searches for 'Bootstrap 3 docs' they'll get the latest v3.4 docs and searching for 'Bootstrap 4 docs' will get the latest v4.3 docs.

chrisbo:
Disallowing only works for unindexed content.

coliff:
It can be requested to remove the unwanted URLs from the index via Google Search console https://search.google.com/search-console/welcome

XhmikosR:
Yeah it's my TODO to sort this out.

I need to start from somewhere probably by including the subsitemaps for the docs we want in the root sitemap.

And then block the other versions.

But while at it, I was thinking if maybe having a `/latest/` endpoint is a good idea. It has its pros and cons though.

[...]

PR #30090, opened 2020-01-26
Adding version number to documentation `<title>`

MikeRogers0:
When I've searched for bootstrap documentation it can be a little tricky to tell what version I'm about to click through to on search results ( this was mentioned in https://github.com/twbs/bootstrap/issues/28581 ). As I sometimes run different versions of Bootstrap it would be nice to search for "bootstrap v4 modal" & get the latest documentation.

A solution I'd like to purpose is to add the version number into the `<title>` tag going forward, I think this could make it easier to identify which links are for the latest documentation before I click through.


MartijnCuppens (approved):
I'm ok with this

mdo (approved):

mdo:
I'm good with backporting this to v4 btw.

PR #30106, opened 2020-01-29
Change file permissions to 644

XhmikosR:
I wonder if I should do the same for remaining 2 files:

* build/change-version.js
* build/ship.sh

For ship.sh it might be an issue, but I don't use the script myself.


XhmikosR:
I wonder if I should do the same for remaining 2 files:

* build/change-version.js
* build/ship.sh

For ship.sh it might be an issue, but I don't use the script myself.

MartijnCuppens (approved):

XhmikosR:
How about the other 2 files that have 0755?

MartijnCuppens:
The two files are still executable if permissions are switched to `664`.

XhmikosR:
Alright, I'll change the other two files tomorrow too, assuming you tried them yourself.

MartijnCuppens:
> Alright, I'll change the other two files tomorrow too, assuming you tried them yourself.

Yup, I did

PR #30110, opened 2020-01-30
Fix borders of list-group using horizontal & flush

mendozagioo:
Fix issue: #30027


mdo:
I don't see much use case in nesting horizontal list groups in vertical ones, nor much of an interest from folks in seeing this added. This complicates the selectors quite a lot as well. Given that, I'm inclined to pass on this for the time being. Thanks though!

PR #30121, opened 2020-02-03
Separate background color from background shorthand in custom select

Quy:
In Firefox and Chrome, the custom select displays correctly but not with Safari 13.0.4 and Microsoft EdgeHTML 18.18363. Moving background color to its own line after the background shorthand fixes the issue.

Before PR:
background

> background: var(--atum-bg-dark) url(../images/select-bg.svg) no-repeat right center/116rem;

After PR:
background-color

> background: url(../images/select-bg.svg) no-repeat right center/116rem;
background-color: var(--atum-bg-dark);

### DIFF
--- a/scss/forms/_form-select.scss
+++ b/scss/forms/_form-select.scss
@@ -14,7 +14,8 @@
   line-height: $form-select-line-height;
   color: $form-select-color;
   vertical-align: middle;
-  background: $form-select-bg escape-svg($form-select-indicator) $form-select-background;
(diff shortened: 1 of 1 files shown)


ffoodd:
You're using a custom property for `background-color`, that's why it breaks.

I don't think Bootstrap uses custom properties at this point — does it, @MartijnCuppens?

MartijnCuppens:
Nope, it doesn't yet. Anyway, https://github.com/twbs/bootstrap/pull/29882 will solve this issue too, so I'm gonna close this one.

PR #30072, opened 2020-01-23
V4 dev backports and updates

XhmikosR:
~~We need #29325 and #29986 to be manually backported to this branch~~

PR #30147, opened 2020-02-08
v4-dev updated nuspec for content files

supergibbs:


nuget/bootstrap.nuspec
```diff
@@ -2,16 +2,16 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>bootstrap</id>
-    <version>4.4.1</version>
+    <version>4</version>
```
XhmikosR: What's the rationale behind not using the full version here?

nuget/bootstrap.nuspec
```diff
@@ -20,16 +20,20 @@
       <dependency id="popper.js" version="[1.16.0,2)" />
     </dependencies>
     <tags>css mobile-first responsive front-end framework web</tags>
+	<contentFiles>
```
XhmikosR: Indentation seems off

XhmikosR:
We should just land one patch in master, cherry pick it and when v5.0.0 is out, we just change the version number in master.

nuget/bootstrap.nuspec
supergibbs: fixed

nuget/bootstrap.nuspec
supergibbs: I  added a comment to the file; the version is pulled from package.json so no need to maintain this

supergibbs:
@XhmikosR This is all cleaned up. Normally I'd agree to merge one and cherry pick the commit but the .nuspec files have already diverged so probably cleaner to just merge both this and #30148

XhmikosR:
@supergibbs can you check https://github.com/twbs/bootstrap/pull/30072/commits/0a3042e8808f3b3f6f87de044fc55d9602cf6711?

supergibbs:
Ya that looks good, if you want to cherry pick it and resolve the conflicts instead that's fine. It should include the `jQuery` dependency in `v4-dev`

XhmikosR:
Yeah, that's already done there, I just cherry picked the rest.

Thanks for the patch, it will be included in #30072

PR #30148, opened 2020-02-08
Updated nuspec content files and icon

### LINKED ISSUE #30045: Sass files are not added to the project, when installed bootstrap.sass nuget package
when installing bootstrap.sass into ASP.NET Core 3.1 project, the sass files are not added to the project.

The reason is described [here](https://stackoverflow.com/a/59712339/475727): 
> package authors should use contentFiles element instead of old content element when creating packages.
> The author do include those files in the contentFiles folder when generating nuget packages. However, he also need to specify the contentFiles element in xx.nuspec file like this [issue](https://stackoverflow.com/questions/34031650/nuspec-contentfiles-example). Only one folder whose name is contentFiles in xx.nupkg is not enough if we want to access those files in VS Solution Explorer

Steps to reproduce:

1. Create default asp.net core project with either CLI or Visual Studio
   `dotnet new webapp`
2. cd web-app
3. Install-Package bootstrap.sass

- .NET Core 3.1
- Operating system and version Win10
- Visual Studio 16.4 (optional)

supergibbs:
Fixes #30045


XhmikosR:
Hey, @supergibbs. Since v5 isn't out yet, I don't think we should land this. That being said, if you fix the indentation and ping me we can land the other parts.

supergibbs:
Oh ok, I thought master was now the v5 dev but I’ll revert it and we can update later. The version tag is required but isn’t used. The build script pulls it from package.json.

supergibbs:
@XhmikosR All fixed!

XhmikosR (approved):

PR #30161, opened 2020-02-11
change-version.js: Remove nuspec files

XhmikosR:
The version in the nuspec files is updated automatically from package.json through MyGet.ps1

/CC @supergibbs


MartijnCuppens (approved):
Approval based on @supergibbs' 👍

Issue #29439, opened 2019-09-25
.invalid-feedback doesn't show for .input-group with an invalid control


br3nt:
I'm reposting an outstanding bug that is documented in #23454 (closed).  That issue has received ALOT of attention over a two year period but has not received a resolution.

**Original issue:**

Example:

```html
<div class="form-group col-md-4">
  <label class="form-control-label is-invalid" for="valuation_value">Estimated Value</label>
  <div class="input-group is-invalid">
    <span class="input-group-addon">£</span>
    <input class="form-control is-invalid" min="0" type="number" value="" name="valuation[value]">
  </div>
  <div class="invalid-feedback">can't be blank, is not a number</div>
</div>
```

Results in:

<img width="217" alt="29313300-384e904a-81b1-11e7-962c-893d9d708c4c" src="https://user-images.githubusercontent.com/1089651/65596709-9823f700-dfd6-11e9-8c25-52e1156a9205.png">

I would expect to see `can't be blank, is not a number` under the input-group.

**Summary of working hacks:**

There are two working solutions (aka hacks) to get the `invalid-feedback` displaying for `custom-file` with an `input-group-append`.

1) Adding the `flex-wrap` class to the `custom-file` and adding the `w-100` class to the `invalid-feedback` element.

```html
    <div class="section measurements-area" style="display: none">
        <h4>Measure your test sample and upload the csv file</h4>
        <form id="measurements-form" class="needs-validation" novalidate>
            <div class="input-group input-group-sm">
                <div class="custom-file flex-wrap">
                    <input type="file" class="custom-file-input" id="measurements" required>
                    <div class="invalid-feedback w-100">
                        The measurements file is required.
                    </div>
                    <label class="custom-file-label" for="measurements" data-browse="Browse">Upload measurements</label>
                </div>
                <div class="input-group-append">
                    <button class="btn btn-outline-primary" type="submit" id="recalibrate">Recalibrate</button>
                </div>
            </div>
        </form>
    </div>
```

2)  Adding `style="position: absolute; top: 3.0em"` to the `invalid-feedback` element.  

```html
<form id="measurements-form" class="needs-validation" novalidate>
    <div class="input-group input-group-sm">
        <div class="custom-file">
            <input type="file" class="custom-file-input" id="measurements" required>
            <div class="invalid-feedback" style="position: absolute; top: 3.0em">
                The measurements file is required.
            </div>
[...]

PR #30170, opened 2020-02-13
V5 form validations

### LINKED ISSUE #28414: Validation feedback/tooltips lack association with their form control

ysds:
Drop feedback messages by sibling selector for V5.

There is no way to solve the #25110 and #29439 maintaining the current specification. If you are using a modern JavaScript framework, this removal of sibling selectors is not a problem, because feedback messages are almost always controlled by `if` directives (e.g. `ng-if`). Control styling with the `.is-valid/.is-invalid` has the same feature as v4.

Also this close #28414 and #30438.

image

Preview: https://deploy-preview-30170--twbs-bootstrap.netlify.app/docs/5.0/forms/validation/

PR for v4 is #30180


mdo:
So with this change, there is no built-in and functional form validation in Bootstrap? Because everything would need to be controlled by another JS library or framework?

ysds:
Does not deny the creation of new built-in JavaScript plugins. We can consider it after the alpha release.

At the least, the current  control by sibling selectors should be abolished.

XhmikosR:
Just keep in mind there's one snippet using inline JS to handle validation. If it's supposed just to be an example we should make it so, because currently it applies to the code.

ysds:
I rebased into the master and updated the first custom JS example.

mdo:
Any chance this could be rebased @ysds?

XhmikosR:
I don't want to hurry this @mdo. The tooltip example doesn't look right, the feeback is always shown.

ysds:
Sorry for being away for a while.

> The tooltip example doesn't look right, the feeback is always shown.

It's intentional that the feedback tooltip examples are always visible, but what do you think it should be?

XhmikosR:
>It's intentional that the feedback tooltip examples are always visible, but what do you think it should be?

I thought it was a mistake since previously they didn't show up nor in v4. If it's intentional then all good :)

ysds:
👍  As commented on https://github.com/twbs/bootstrap/pull/30170#issuecomment-586534806
, additional JS is required to toggle the display of feedback messages. I think the static sample is enough for now.

mdo:
@ysds Can you confirm something here before I attempt to resolve conflicts and merge this? Are we removing all feedback elements from the examples under the `## Server side` heading? Your earlier commit does that, and I may have screwed that up in my later commits.

mdo:
Hoping this can be replaced by #31666.

XhmikosR:
@mdo is this still valid? #31666 was closed in favor of #31953

mdo:
Closing as out of date—miss you @ysds! <3

PR #30171, opened 2020-02-14
Add .dropdown-menu-dark

### LINKED ISSUE #30105: dropdown-dark
In the current bootstrap version, dark mode is a painful choice for the dropdown menu (and a few other components).

Dark nav bar with white can create troubles for people sensitive to UI/UX.

There are existing workarounds, but the implementation gives us the feeling that there is a bug in the design of bootstrap, dropdown menu dark mode should be as accessible as navbar dark mode.

Plenty of people are proposing complicated workarounds of something that should as basic navbar-dark bg-dark **dropdown-dark**:

- https://stackoverflow.com/questions/44612813/how-to-change-bootstrap-4-drop-down-colors/44612898

- https://community.plot.ly/t/i-cant-change-color-of-dropdown-menu/28174

mdo:
Screen Shot 2020-02-13 at 5 35 41 PM

Fixes #30105.

Preview: <https://deploy-preview-30171--twbs-bootstrap.netlify.app/docs/5.0/components/dropdowns/#dark-dropdowns>


MartijnCuppens:
I'm a bit worried that once we're adding some dark components, people will keep asking for support for other elements (eg. modals, pagination (which I declined in the past, https://github.com/twbs/bootstrap/pull/29892), form elements, ect... )

Once we switch to CSS variables, it 'll be way easier to colorize components.

mdo:
Yeah, I agree on that concern, but we can still say no to most.

I mostly wanted to do this because of how prominent our dark navbar component is. Completing the style there and connecting the two components (navbar and dropdown) makes a lot of sense to me. The rest, I'm not as sold on, though we might as well discuss our options further beyond this.

MartijnCuppens:
Ok, fine by me. Could you also switch the dropdowns of the dark navbars then (make sure to tackle the examples too)?

mdo:
> Could you also switch the dropdowns of the dark navbars then (make sure to tackle the examples too)?

Know I `+1`ed this back in February, but coming back to this I don't know how much I want to actually replace the other dark navbars. I think since this is a modifier class, it can stay as an option documented under the dropdowns.

Any aversions to merging as-is without further examples or docs examples?

MartijnCuppens:
> Any aversions to merging as-is without further examples or docs examples?

Now that we have CSS custom properties, this might be a good use case to use them. I think in we 'll need to update our other components to CSS custom properties (where suitable), adding this like might cause double work.

mdo:
Thinking I'll leave this as-is for now and then come back in minor releases to CSS-variable-ify everything. Thoughts on that approach @MartijnCuppens?

MartijnCuppens (approved):
Job well done, @mdo!

ffoodd (approved):
LGTM

XhmikosR:
I like it, I just don't know how we are going to proceed with dark components for all components in the future, but we'll discuss it later :)

PR #30183, opened 2020-02-15
Remove unnecessary reduce motion when $enable-transition: false

ysds:
When `$enable-transition: false`, 
```scss
   transition: none;
}
```
is unnecessary.

### DIFF
--- a/scss/mixins/_transition.scss
+++ b/scss/mixins/_transition.scss
@@ -6,11 +6,11 @@
     } @else {
       transition: $transition;
     }
-  }
 
-  @if $enable-prefers-reduced-motion-media-query {
-    @media (prefers-reduced-motion: reduce) {
-      transition: none;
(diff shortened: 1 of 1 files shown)


XhmikosR:
@ysds don't forget to add the projects in the PRs :)

XhmikosR:
If this is supposed to be backported to v4, we need the 4.4.2 project too.

MartijnCuppens (approved):

XhmikosR:
Any reason why not include this in alpha-1 and 4.4.2 subsequently?

ysds:
I think this a minor bug. To speed up the release of Alpha1, I targeted this to Alpha2 not to increase our review work, but now the review was completed it is fine to include this in Alpha1.

I forgot to put it in 4.4.2 ^^;

Issue #21967, opened 2017-02-11
Auto wrap with col does not work in Safari


iFlash:
This code will not auto wrap in Safari 10.0.3  (OSX 10.12).

Basically using class="col" in order to auto wrap columns will not work. The columns are displayed outside page margins.

Chrome and FF behave as expected.

```
<div class="container-fluid">
    <div class="row">
        <div class="col"><small>Some text here</small></div>
        <div class="col"><small>Some text here</small></div>
        <div class="col"><small>Some text here</small></div>
        .... add more columns until they should wrap
        <div class="col"><small>Some text here</small></div>
        <div class="col"><small>Some text here</small></div>
    </div>
</div>
```

iFlash:
couldnt get the tags to show. sorry.

iFlash:
See http://stackoverflow.com/questions/42174385/safari-and-bootstrap-4-automatic-grid-breaks for example.

mdo:
Bug reports must include a live demo of the problem. Per our contributing guidelines, please create a reduced test case via JS Bin and report back with specific browser and OS details.

iFlash:
I made an example available. I don't have the time to mouth feed you to make you product bug free. Sorry.

mukadr:
Created [JSBin](https://jsbin.com/yuhubuziru/edit?html,output)

Reproduced this issue on the following browsers:

iOS 7, Safari 7
iOS 8, Safari 8
iOS 7 & 8, Chrome 47

mdo:
Turns out this is [flexbug 11](https://github.com/philipwalton/flexbugs#11-min-and-max-size-declarations-are-ignored-when-wrapping-flex-items), where Safari doesn't wrap flex items properly. As I understand it, Safari uses different calculations for determining how to wrap flex items. The only workaround is using `flex-basis` to size the columns :\. In this case, you'd need to add:

```css
.col {
  flex: 1 0 50%; // where 50% is the intended size of the items
}
```

This is the same problem #21620 faces as well. Digging into even more, a `border: 1px solid transparent` on the `.col` will also fix this. Will work on a docs update to suggest these in context.

dxh9845:
I got this as well. Site: danherzig.com. Happened on iOS browsers. Don't have an iPhone but got these screenshots. Fixed by reverting to prior columns (col-sm-12 and col-md-6). So the official fix was adding an invisible border?

pimisi:
[...]

PR #30189, opened 2020-02-16
Remove confusing "Equal-width multi-line" section from Grid docs.

mikelehen:
Reading the grid docs from top to bottom, this section made no sense to me. I didn't understand why you wouldn't just use multiple rows, and I had no idea what w-100 meant, or how you'd "mix" it with "responsive display utilities".

Then I continued reading and got to https://getbootstrap.com/docs/4.4/layout/grid/#column-breaks which explains the same concept but with a much better explanation and examples (including the responsive display utilities).

It looks like the "Equal-width multi-line" section was added in 9010978c1a to resolve to #21967 as a way to document the existence of a Safari bug.  But that bug has long since been fixed (Safari 10.1 came out in May 2017) and so I think it's reasonable to just drop this section and rely on the better docs further down in the guide.

### DIFF
--- a/site/content/docs/4.3/layout/grid.md
+++ b/site/content/docs/4.3/layout/grid.md
@@ -145,24 +145,6 @@ For example, here are two grid layouts that apply to every device and viewport, 
 </div>
 {{< /example >}}
 
-### Equal-width multi-line
-
-Create equal-width columns that span multiple lines by inserting a `.w-100` where you want the columns to break to a new line. Make the breaks responsive by mixing `.w-100` with some [responsive display utilities]({{< docsref "/utilities/display" >}}).
-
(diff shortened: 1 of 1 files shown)


mdo:
This can probably go away with `.row-cols` classes, right @twbs/css-review?

MartijnCuppens:
> Then I continued reading and got to getbootstrap.com/docs/4.4/layout/grid/#column-breaks which explains the same concept but with a much better explanation and examples (including the responsive display utilities).

Yup, this section is way clearer indeed.

> This can probably go away with `.row-cols` classes, right twbs/css-review?

Indeed, therefore it would also be better to remove this confusing section.

MartijnCuppens (approved):

MartijnCuppens:
Thanks for taking your time to make a PR for this, @mikelehen!

PR #30192, opened 2020-02-16
Add transparent background color to versions switcher

MartijnCuppens:
Fixes blue background:
image


mdo (approved):
Was wondering what I/we did to cause this haha, thanks!

XhmikosR (approved):

XhmikosR:
Must be a regression from the shorthand switch patch.

PR #30193, opened 2020-02-16
Remove `.table-dark` as table variant

### LINKED ISSUE #27879: .table-dark is used in two conflicting contexts
> version : `v4.1.3`

default

MartijnCuppens:
Remove `.table-dark` as [color variant](https://twbs-bootstrap.netlify.com/docs/4.3/content/tables/#variants). `.table-dark` was already used for [inverted tables](https://twbs-bootstrap.netlify.com/docs/4.3/content/tables/#inverted).

Closes https://github.com/twbs/bootstrap/issues/27879

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -101,7 +101,7 @@
 // Exact selectors below required to override `.table-striped` and prevent
 // inheritance to nested tables.
 
-@each $color, $value in $theme-colors {
+@each $color, $value in $table-theme-colors {
   @include table-row-variant($color, color-level($value, $table-bg-level), color-level($value, $table-border-level));
 }
 
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -478,6 +478,8 @@ $table-dark-hover-color:      $table-dark-color !default;
 $table-dark-hover-bg:         rgba($white, .075) !default;
 $table-dark-border-color:     lighten($table-dark-bg, 7.5%) !default;
 
+$table-theme-colors:          map-remove($theme-colors, dark) !default;
+
 $table-striped-order:         odd !default;
 
 $table-caption-color:         $text-muted !default;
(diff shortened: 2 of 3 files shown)


ysds:
I thought it would change either name to fix conflicts.

tomsommer:
> I thought it would change either name to fix conflicts.

I agree this would be a better solution

ysds:
Ideas of the name:
* `table-highlight-{theme}` and `table-active`
* `table-variant-{theme}` and `table-active`
* `table-data-{theme}` and `table-active`

tomsommer:
Why not just rename `.table-dark` to `.table-inverted`?

MartijnCuppens:
> Or even better: remove .table-dark entirely since you can do the same with other utility classes?

You can't use the current utility classes to change the border colors. On the other hand, it's not that user friendly if you need to add classes to each cell to override border colors.

> I thought it would change either name to fix conflicts.

Any suggestions on this approach? I chose to remove the dark variant since the `dark`, `active` and `secondary` table variants look quite alike.

ysds:
> Any suggestions on this approach? I chose to remove the dark variant since the dark, active and secondary table variants look quite alike.

Please see my above comment.

MartijnCuppens:
My point was: do we really need this many shades of gray? https://codepen.io/MartijnCuppens/pen/dyoOpNV

ysds:
The gray active and secondary are just by default colors. Users can customize the colors, but dark is dark for everyone.

ysds:
For example, we need to consider users who have active light blue and secondary pink.

mdo:
Wondering if we can nix the `.table-dark` that inverts things for a series of utilities:

- `.bg-dark` for the `background-color`
- `.text-white` for the `color` change
- Add new border color utilities

Thoughts?

mdo:
Alternate approach opened up at #30342.

PR #30221, opened 2020-02-18
Add .row-cols-{sm/md/lg/xl}-none

ZhangChengLin:
Add .row-cols-{sm/md/lg/xl}-none 
Meeting the need when quantity is no longer needed

### DIFF
--- a/scss/mixins/_grid-framework.scss
+++ b/scss/mixins/_grid-framework.scss
@@ -35,6 +35,11 @@
         max-width: 100%;
       }
 
+      .row-cols#{$infix}-none {
+        flex: auto;
+        max-width: none;
+      }
+
       @for $i from 1 through $grid-row-columns {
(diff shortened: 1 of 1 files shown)


XhmikosR:
Thanks, but we don't need 2 PRs.

zhangchenglin:
> Thanks, but we don't need 2 PRs.

yes
 
This is for 2 different branches. I see that the structures of v4 and v5 are different, so two PRs at different positions were submitted.
And I don't know the way I use, can they merge into one PR.

XhmikosR:
We can cherry pick the patch if we decide we want to backport it.

zhangchenglin:
> We can cherry pick the patch if we decide we want to backport it.


The network is very slow. It took me 18 minutes to refresh and see your information. Thank you, I got it.

PR #30230, opened 2020-02-18
Use collapse component in docs' sidenav

ffoodd:
Related to #30073, use collapse component in sidebar nav.

Preview: <https://deploy-preview-30230--twbs-bootstrap.netlify.com/docs/4.3/getting-started/introduction/>

### DIFF
--- a/site/assets/js/application.js
+++ b/site/assets/js/application.js
@@ -19,19 +19,6 @@
     return [].slice.call(list)
   }
 
-  makeArray(document.querySelectorAll('.js-sidenav-group'))
-    .forEach(function (sidenavGroup) {
-      var groupHasLinks = Boolean(sidenavGroup.querySelector('li'))
-      var groupLink = sidenavGroup.querySelector('a')
-
-      if (groupHasLinks) {
-        groupLink.addEventListener('click', function (e) {
-          e.preventDefault()
-          e.target.parentNode.classList.toggle('active')
-        }, true)
-      }
-    })
-
   // Tooltip and popover demos
   makeArray(document.querySelectorAll('.tooltip-demo'))
     .forEach(function (tooltip) {
--- a/site/assets/scss/_sidebar.scss
+++ b/site/assets/scss/_sidebar.scss
@@ -55,7 +55,7 @@
   flex-flow: column nowrap;
   padding-left: 1.25rem;
 
-  > li > a {
+  a {
     display: inline-block;
     padding: .25rem .5rem;
     @include font-size(.875rem);
@@ -70,7 +70,7 @@
     }
   }
 
-  [aria-current="page"] {
+  .active {
     font-weight: 600;
     color: rgba($black, .85);
   }
--- a/site/assets/scss/_sidebar.scss
+++ b/site/assets/scss/_sidebar.scss
@@ -16,10 +16,6 @@
   }
 }
 
-:not(.active) > .bd-sidenav {
-  display: none;
-}
-
 .bd-sidenav-group-link {
   padding: .25rem .625rem .25rem .5rem;
   font-weight: 600;
@@ -57,6 +53,7 @@
 
 // All levels of nav
 .bd-sidebar .nav {
+  flex-flow: column nowrap;
   padding-left: 1.25rem;
 
   > li > a {
--- a/site/assets/scss/_sidebar.scss
+++ b/site/assets/scss/_sidebar.scss
@@ -40,13 +40,11 @@
     content: escape-svg($sidebar-collapse-icon);
   }
 
-  &.active {
-    .bd-sidenav-group-link::before {
-      transform: rotate(90deg);
-    }
+  [aria-expanded="true"] {
+    color: rgba($black, .85);
 
-    > .bd-sidenav-group-link {
-      color: rgba($black, .85);
+    &::before {
+      transform: rotate(90deg);
     }
   }
 }
@@ -71,9 +69,7 @@
     }
   }
 
-  > .active > a,
-  > .active:hover > a,
-  > .active:focus > a {
+  [aria-current="page"] {
     font-weight: 600;
     color: rgba($black, .85);
   }
--- a/site/assets/scss/_sidebar.scss
+++ b/site/assets/scss/_sidebar.scss
@@ -1,72 +1,63 @@
-// stylelint-disable declaration-no-important
-
 .bd-links {
-  @include media-breakpoint-up(md) {
-    @supports (position: sticky) {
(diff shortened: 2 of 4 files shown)


XhmikosR:
This has the issue I also faced back then, the show/collapse is jaggy, or whatever the proper word is to describe this :P

XhmikosR:
My previous PR #29104

ffoodd:
@XhmikosR The struggling transitions was due  to `flex-column` and wrapping that didn't happen at the right time. I guess it's good now.

I chose to make it in `docs.css` instead of using utilities since it's pretty specific: should I change it to use utilities?

XhmikosR:
It's fine, thanks! But do we need the custom JS code to toggle the `active` class now? Can't we make use of another built-in class for this here and drop the custom JS toggling code?

I remember @mdo not linking the transition there, which was another reason I dropped my PR.

ffoodd:
You were right, I was able to drop the custom JS (and the associated class in markup) by simply using `aria` attributes as hooks for styles. It makes sense, to me.

ffoodd:
About transition, it's all about the collapse component: I think it's more consistent to keep it, but we should be able to reset it if you decide otherwise.

site/assets/scss/_sidebar.scss
```diff
@@ -44,19 +40,18 @@
...
-    .bd-sidenav-group-link::before {
-      transform: rotate(90deg);
-    }
+  [aria-expanded="true"] {
+    color: rgba($black, .85);
 
-    > .bd-sidenav-group-link {
-      color: rgba($black, .85);
+    &::before {
+      transform: rotate(90deg);
```
XhmikosR: If we keep the transition we might as well add a transition for the arrow. IIRC the default duration is `.35s`

But let's wait for @mdo to chime in first

mdo:
You know what, I can dig it.

Makes me think that we could also, in another PR, do a docs or dashboard example that uses this same kind of layout and sidebar. We always get folks asking to use pieces of our docs :).

site/assets/js/application.js
```diff
@@ -29,19 +29,6 @@
     checkbox.indeterminate = true
   })()
 
-  makeArray(document.querySelectorAll('.js-sidenav-group'))
```
XhmikosR: @ffoodd Can you check if `js-sidenav-group` is used anywhere else and if not remove it from the HTML too?

XhmikosR: Ah, NVM, you already have :)

site/assets/scss/_sidebar.scss
```diff
@@ -44,19 +40,18 @@
     content: escape-svg($sidebar-collapse-icon);
   }
 
-  &.active {
-    .bd-sidenav-group-link::before {
-      transform: rotate(90deg);
-    }
+  [aria-expanded="true"] {
```
MartijnCuppens: Not a fan of using attributes to style things. We should be able to style things based on the `.collapsed` class. The class should be added to the collapsed items by default though.

site/assets/scss/_sidebar.scss
MartijnCuppens: It feels weird to transition the height but not the rotation indeed.

site/assets/scss/_sidebar.scss
```diff
@@ -74,9 +69,7 @@
     }
   }
 
-  > .active > a,
-  > .active:hover > a,
-  > .active:focus > a {
+  [aria-current="page"] {
```
MartijnCuppens: Same here, add an `active` class in the docs-sidebar template instead.

XhmikosR: Well, the idea was to reduce the custom JS needed for this thing. Ideally we should rely on the classes the collapse plugin is using, but worse case scenario, we add back an active class.

MartijnCuppens: This can be achieved by changing the Hugo template.

XhmikosR: Yeah, well, I'm not sure it's simpler, so feel free to provide a patch. BTW I still want to add the page's class in body, but with our templates being messed up, it wasn't so easy last time I tried.

ffoodd:
Hi there, just to let you know: my second child is just bien, so don't expect me to work on this until mid-march 👶

MartijnCuppens:
Congrats @ffoodd!

site/assets/scss/_sidebar.scss
ffoodd: In fact it's lighter using attribute selector: the equivalent selector would be something like `&.has-children .bd-sidenav-group-link:not(.collapsed)`.

Would really prefer the latter?

Question, too: aren't `.bd-sidenav-group` and `.has-children` redundant? I can't find a case where one of those would be set without the other…

site/assets/scss/_sidebar.scss
ffoodd: @XhmikosR There is already a `.active` class on the parent in Hugo template, let's use it. I just move it to the link, since it's not used on the list item anyway…

ffoodd:
Here I am :)

From my view, the only remaining question si whether to keep `[aria-expanded="true"]` selector or to switch to `&.has-children .bd-sidenav-group-link:not(.collapsed)` (and adding `.collapsed` class in markup for first load).

@MartijnCuppens what's your opinion?

MartijnCuppens:
@ffoodd, in general classes are used for theming, aria labels for accesibility.

I just simplified the styles a little bit. Not sure if we should use a class name for the links. I think we can use the less strict approach here since all links have a lot of styles in common.

ffoodd:
@MartijnCuppens  `:not(;only-child)` is a  nice move!

XhmikosR (approved):

XhmikosR:
I like it! Less code, better for me :)

Waiting for @patrickhlauke's, @mdo's and @MartijnCuppens's approval.

MartijnCuppens (approved):

XhmikosR:
@ffoodd can you rebase this please?

ffoodd:
@XhmikosR Done :)

XhmikosR (approved):

XhmikosR:
I'm going to merge this and we can tweak it further later.

inwardmovement:
Did you guys consider using the accordions alternative?

At first sight I think it may be more practical and pleasant to use (+ resulting in a less long sidebar ; is it useful to have more than one section expanded at a time?)

XhmikosR:
@inwardmovement please make new issues, commenting on old stuff is hard to track.

PR #30236, opened 2020-02-19
Add .row-cols-{sm/md/lg/xl}-none > *

ZhangChengLin:
Add .row-cols-{sm/md/lg/xl}-none > *  
Meet the need when you no longer need to control the number of grid columns


ysds:
Thanks but This is included in #28517.

PR #30244, opened 2020-02-19
Fix centered modal scrolling issue

ysds:
Another solution of #30003.

You can see the issue in the [vertical modal demo](https://getbootstrap.com/docs/4.4/components/modal/#vertically-centered). This behavior occurs when the browser UI (address bar) is displayed.

ezgif com-optimize

This PR fix it by using the `min-content` property.

Preview: https://deploy-preview-30244--twbs-bootstrap.netlify.com/docs/4.3/components/modal/#vertically-centered


Achmed4:
I like your solution @ysds , Its more better and simple. Thanks 👍

MartijnCuppens (approved):

PR #30251, opened 2020-02-19
prevents .table-striped to be automatically applied to inner tables

lucianocosta:
It seems we had no clean way to remove the striped behaviour from inner tables once it's added to a parent table. Now it's possible to add it manually, if wanted.

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -76,7 +76,7 @@
 // Default zebra-stripe styles (alternating gray and transparent backgrounds)
 
 .table-striped {
-  tbody tr:nth-of-type(#{$table-striped-order}) {
+  & > tbody > tr:nth-of-type(#{$table-striped-order}) {
     background-color: $table-accent-bg;
   }
 }
@@ -146,7 +146,7 @@
   }
 
   &.table-striped {
-    tbody tr:nth-of-type(#{$table-striped-order}) {
+    & > tbody > tr:nth-of-type(#{$table-striped-order}) {
       background-color: $table-dark-accent-bg;
     }
   }


MartijnCuppens:
We should do this with all our table styles for consistency. The docs should be adjusted then too.

lucianocosta:
Hi @MartijnCuppens!

It seems to be just an unexpected behaviour for ".table-striped" class, I think there's nothing to change in the docs...

I'm trying to help by solving one problem at time, without side effects. Please point me what other styles you think we should fix to consistently solve this, I would be pleased to give back to Bootstrap!

lucianocosta:
The [doc for .table-striped](https://getbootstrap.com/docs/4.4/content/tables/#striped-rows) says:

> `Use .table-striped to add zebra-striping to any table row within the <tbody>.`

And this behaviour is currently being added for (inner) tables that I've **not added** the class ".table-striped"

MartijnCuppens:
> I'm trying to help by solving one problem at time, without side effects.

That's how we like it. However, we can't ship a version where the child combinator is only used for the striped tables.

> Use .table-striped to add zebra-striping to any table row within the <tbody>

"Any" means any, not only the direct children.

Don't get me wrong, I really love to see this, but then applied to all table styles to be consistent and with documentation changes + documentation in the migration guide.

lucianocosta:
> That's how we like it. However, we can't ship a version where the child combinator is only used for the striped tables.

I've started to work on that, but the more I see, the more think that's an issue just for striped tables! I mean, if I set a table to be a table-dark, table-bordered, table-borderless, table-hover or table-sm, it's *visually expected* that behaviour to be extended for any inner table, and it's easy to change if not. But when we set it to be a striped-table, the inner table's rows get weird because of the tr(odd) selector, wich mixes *any* table rows inside a striped-table...  

Please take a look to [this jsFiddle](https://jsfiddle.net/lucianocosta/2q8cxstf/17/) and note that I've changed the inner table class attr on second row (@fat), to be sm and borderless.

Thanks for your attention, and tell me if you think that we should really change it all, then I'll do it!

lucianocosta:
In time, just noticed that @mdo [changed that](https://github.com/lucianocosta/bootstrap/commit/547f16b58e54375b1d523cc6e2153993bf731875) 5 years ago

MartijnCuppens:
> ... the more think that's an issue just for striped tables ...

I don't think it it. For example, the hoverable table styles will leak to nested tables.

> and tell me if you think that we should really change it all

This issue might need more research and discussion. Could you open a new issue to discus this with some POC codepens (also with nested tables)?

PR #30262, opened 2020-02-20
$link-decoration: underline affects some components

### LINKED ISSUE #15304: Links (as part of regular text) not distinguishable enough
For aesthetic reasons, by default Bootstrap links have no underlines, relying instead just on a color difference from surrounding text.

This in itself violates WCAG 2.0 "1.4.1 Use of color" http://www.w3.org/TR/WCAG20/#visual-audio-contrast-without-color

There is a technique suggesting that sufficient difference in color contrast/luminance may be enough in certain conditions, but having some additional cue is preferred http://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140916/G183

There are also situations where links are even less distinguishable.

bootlink2

http://getbootstrap.com/components/#alerts-links - link is of course distinguishable from surrounding text, but the difference between the actionable link and the bold text at the start of the alert ("Well done!" etc) is again only down to a subtle variation in color ... this is confusing for users, as it's not immediately clear what is and isn't a link.

bootlink3

http://getbootstrap.com/components/#navbar-links No distinction at all between inert text and link, unless the user goes on a "let me just hover my mouse over every word" pixel hunt. All sighted users are affected by this (though at least keyboard users will more readily discover these links by virtue of tabbing onto them).

In short: I propose that by default, links should ideally have some subtle additional indication that they're links. Of course, for links in navigation bars etc, this does not apply...the context there is clear enough ("this is a series of actionable things"). But particularly for links that are just part of regular text, we should really do something.

Failing that, this needs to be noted in our accessibility guidance.

/cc @cvrebert

arthurshlain:
Removed redundant text-decoration when element is link with .btn class (a.btn) and $link-decoration variable is set to 'underline'.

.btn-link also works with this fix, because .btn-link declaration located below.

**Upd.**

Fixed text-decoration for:

- .btn (except .btn-link)
- .dropdown-item
- .list-group-item
- .page-link (looks like button)
- .nav-link


ysds (dismissed):
Nice catch!

ysds:
I'm wondering if this should be applied to other components.

ArthurShlain:
I know same behavior exist with pagination <a> elements.
Also 'underline' applied to .navbar-brand, but it's maybe ok...

MartijnCuppens:
I would prefer an if test to prevent generation of the property if not needed. Something like:
```
text-decoration: if($link-decoration == none, null, none);
```
(code not tested)

ArthurShlain:
> (code not tested)

It's works good

ArthurShlain:
.nav-link looks normal by default, except .active state

image

MartijnCuppens:
> Remove underline for any `.nav-link` or only for `.active` state?

I would just make sure the nav links look the same like in the current docs:

image

ArthurShlain:
Something wrong with this fix now?

ysds (approved):
LGTM

MartijnCuppens (approved):

patrickhlauke:
this closed #15304 but note that one of the scenarios is still there...alerts with links and strong text (mocked up here by taking one of the alerts in the docs and adding an `<a href="#" class="alert-link">...</a>` in there)

screenshot of alert with both bold strong text and a bold link

MartijnCuppens:
Didn't notice this was added, I've reopened #15304.

Issue #30052, opened 2020-01-17
Using SVGs inline


MartijnCuppens:
For the carousel indicators, the navbar toggler and the upcoming stand alone accordion components it would be handy to switch from svg background images to inline SVGs. This way we'll have more control over hover states and we could save some file size in our generated CSS.

Maybe we should switch to SVG for the close button to, this will make it easier to theme the focus state (see https://github.com/twbs/bootstrap/issues/27163)

- [ ] Convert carousel indicators (#30268)
- [ ] Convert navbar toggler
- [ ] Convert accordion icons
- [ ] Convert close button

ffoodd:
Also I think it'd be better for accessibility, since inline SVGs are not as much affected by Windows High Contrasts Mode as `background-image` (at least if they're filled with `currentColor`).

mdo:
Other than the form controls, I'm fine with the SVGs coming out of the CSS. For the carousel for example, it'll enable more customization that we cannot do with images in CSS (e.g., inheriting colors, `currentColor`, easier transitions/styling, etc).

alecpl:
I'm guessing this might be a BC break. In this case consider #27294 again, please.

Glenn-Domin:
> Also I think it'd be better for accessibility, since inline SVGs are not as much affected by Windows High Contrasts Mode as background-image (at least if they're filled with currentColor).

I wonder if this could be covered by the `prefers-color-scheme: dark` media query.

ffoodd:
BTW, using `width="1m" height="1em"` on those SVGs would allow us to size them using `font-size`— which could be very handy too.

mdo:
This is being punted on for v4 and v5 due to difficulties and inconsistencies in getting all our components to use a single approach that's not embedded in CSS. For example, inline SVGs aren't possible in the bulk of our form controls and as such, they must be used via CSS (either embedded or referenced externally).

PR #30268, opened 2020-02-22
Use embedded SVGs for carousel icons

MartijnCuppens:
Part of https://github.com/twbs/bootstrap/issues/30052.

@XhmikosR, do you know a way to easily indent the svgs in the example code?

### DIFF
--- a/js/tests/integration/index.html
+++ b/js/tests/integration/index.html
@@ -46,13 +46,15 @@
               </div>
             </div>
           </div>
-          <a class="carousel-control-prev" href="#carouselExampleCaptions" role="button" data-slide="prev">
-            <span class="carousel-control-prev-icon" aria-hidden="true"></span>
-            <span class="sr-only">Previous</span>
+          <a class="carousel-control-prev" href="#carouselExampleCaptions" role="button" data-slide="prev" aria-label="Previous">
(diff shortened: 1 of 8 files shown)


XhmikosR:
My main concern with this is that the SVGs shouldn't be in the docs since we don't distribute the docs.

MartijnCuppens:
> My main concern with this is that the SVGs shouldn't be in the docs since we don't distribute the docs.

From my perspective, the SVGs are just markup that are needed to make the carousel work.

XhmikosR:
>From my perspective, the SVGs are just markup that are needed to make the carousel work.

I agree, but people might argue that the SVGs should be included.

I'm not against this, obviously, just trying to think what issues we might face.

XhmikosR:
>@XhmikosR, do you know a way to easily indent the svgs in the example code?

Not sure about this right now, I'll need to play with it and see, because using `~` doesn't seem to have any effect.

ffoodd:
About markup VS distribution, just a reminder that for now our close button is text-based and thus requires to get the markup. I think it's the same.

Apart from that, it'd be very easy to make SVGs downloadable from the docs without distributing them in our package: might be another way around this.

mdo:
Closing for #31650.

PR #30273, opened 2020-02-23
Use `scroll-margin-top` instead of pseudo hack

MartijnCuppens:
I recently discovered the `scroll-margin-top` property and it turned out this was exactly what we needed instead of the hack we have. Some side notes:

- This doesn't work for IE, but since sticky itself doesn't work, this is a non-issue.
- This doesn't work for the old Edge, I'm not sure if that's really an issue since it's just for our own docs.

I've also wrapped it in a media query since the stick menu is only present above the `md` breakpoint.


site/assets/scss/_content.scss
```diff
@@ -5,20 +5,11 @@
...
-  > h4[id] {
-    pointer-events: none;
-
-    &::before {
-      display: block;
-      height: 6rem;
-      margin-top: -6rem;
-      content: "";
+  // Offset for the sticky header
+  > * {
```
mdo: I think we could still limit this to headings, no?

MartijnCuppens: I've now limited it to elements with an id, just in case someone links to another element.

XhmikosR:
Not working in Edge is an issue for sure. Not going to block it though, since we are talking about a small percentage anyway https://caniuse.com/#feat=mdn-css_properties_scroll-margin-top

MartijnCuppens:
> Not working in Edge is an issue for sure.

It just doesn't work in the old edge, caniuse is wrong here.

XhmikosR:
I personally meant the old Edge. Like I have mentioned in the past, browsers don't magically get updated on Windows :P Also, Microsoft hasn't attempted to replace the old Edge yet, let alone that even if they do, there are so many supported Windows versions out there.

Anyway, this is just a FYI, ideally we should have some kind of fallback, but even without it I guess we could try landing it later.

mdo (approved):

XhmikosR:
Can't we have a fallback of some kind?

ffoodd:
@XhmikosR by using `@supports()`, it should be pretty easy — however it means more code.

MartijnCuppens:
> @XhmikosR by using @supports(), it should be pretty easy — however it means more code.

This would be moot. The current implementation works 100% (but is pretty ugly), no reason to just add code.

XhmikosR:
Technically, this does **not** work 100% since it doesn't work in all of our supported browsers.

Anyway, https://github.com/twbs/bootstrap/pull/30273#issuecomment-590199201

XhmikosR (approved):

MartijnCuppens:
I feel the day I can convince @XhmikosR to drop legacy Edge support is coming closer...

ffoodd:
Just to be sure, couldn't this be applied to sticky / fixed navbars component too?

PR #30276, opened 2020-02-24
Fix commonmark syntax

satotake:
@XhmikosR 

related 
- https://github.com/twbs/bootstrap/pull/29750
- https://github.com/gohugoio/hugo/issues/6553


XhmikosR:
What does this fix exactly? The previous solution worked fine too.

satotake:
Actually, this leads to no change of your site's view
It is just about styling of code

> The previous solution worked fine too.

Of course, it is ok as it is.
If so, please close

I would like to share another solution to you and close gohugoio/hugo#6553

XhmikosR:
I have already removed the newlines to work around the issue. I'm not against this change, but I'll wait for someone else to chime in since it's just a cosmetic change :)

XhmikosR:
@MartijnCuppens what's your take on this? There's basically no functional change apart from the indentation in the source and output files.

MartijnCuppens (approved):
It's better than before since we don't have a sudden 4 space indention.

PR #30285, opened 2020-02-25
Harmonize usage of constants

Sparksx:
Use stored constants instead of strings for classes and selectors


MartijnCuppens (approved):
LGTM

XhmikosR:
Not sure it's an improvement but at least it's consistent :)

XhmikosR:
For what is worth, this increases the JS size. Not an issue with this patch per se, just generally using the constants as objects, since the minifier does not mangle the properties.

XhmikosR:
I think at some point we'll need to revisit this altogether. I can't see how this is better than just plain strings for constants.

XhmikosR (approved):

Issue #28423, opened 2019-03-07
Non ul or ol based .nav-link should require .nav-item class


ysds:
The non `ul` or `ol` based navs e.g.

```html
<nav class="nav">
  <a class="nav-link active" href="#">Active</a>
  <a class="nav-link" href="#">Link</a>
  <a class="nav-link" href="#">Link</a>
</nav>
```

should be require `.nav-item` class:

```html
<nav class="nav">
  <a class="nav-item nav-link active" href="#">Active</a>
  <a class="nav-item nav-link" href="#">Link</a>
  <a class="nav-item nav-link" href="#">Link</a>
</nav>
```

Because there are many snippets without `.nav-item` class in the docs, so it looks like it's not necessary, but the `.nav-item` class is often required. As an example, the pill style does not require `.nav-item`, but the tab style needs it (this usage isn't documented).

![image](https://user-images.githubusercontent.com/4065765/53945740-9d3a5b00-4105-11e9-99a5-44c5b7321e06.png)

demo: https://codepen.io/anon/pen/wOJWJo

And the [example in the navbar nav](https://getbootstrap.com/docs/4.3/components/navbar/#nav) use `.nav-item` with `.nav-link`.

![image](https://user-images.githubusercontent.com/4065765/53946273-c5768980-4106-11e9-853b-795b17148cd2.png)

This lack of consistency caused the problem of #28117.

I strongly recommend add `.nav-item` to all `.nav-link` of the non `<ul>` or `<ol>` based nav.

ms-yogi:
@MartijnCuppens I want to contribute to the repo. Can I take this issue?

MartijnCuppens:
@ms-yogi, sure, go ahead! For this case, we might maybe better more the `margin-bottom: -1px;` from `.nav-tabs .nav-item` to `.nav-tabs .nav-link`, this way we don't need to add this additional class. (I haven't tested if there are any downsides to this though)

ysds:
@MartijnCuppens 
My point is `.nav > a` requires both classes. Please see 
https://github.com/twbs/bootstrap/issues/28117#issuecomment-457549231

rohit2sharma95:
> My point is `.nav > a` requires both classes.

@ysds `.nav-item` class is required only if `.nav` has `.nav-justified` class.

As the border issue has been fixed (in V5) by adding margin in the `.nav-link` inside `.nav-tabs`
[...]

PR #30288, opened 2020-02-26
Minor `.nav-tabs` optimization

ms-yogi:
~In this PR, margin-bottom is added to the `.nav-tabs .nav-link` and it is removed from `.nav-item` to make the active tab behavior look consistent throughout.~

_Edited:_

The original issue seems to be a documentation issue, this PR just introduces a minor CSS optimization.


ysds:
I’m sure that the #28423 is not a CSS issue but a document issue.

MartijnCuppens:
Indeed, now I see. Still, I think we can merge this as minor CSS optimization.

MartijnCuppens (approved):

PR #30290, opened 2020-02-26
Use `<article>`s for cards

### LINKED ISSUE #30266: Use `<article>`s for cards when appropriate
The cards should be articles in our docs. We should have a look at this after https://github.com/twbs/bootstrap/pull/28517 is merged.

rubenRP:
Fixes #30266

Demo: https://deploy-preview-30290--twbs-bootstrap.netlify.com/docs/4.3/components/card/


MartijnCuppens:
thanks for looking into this, @rubenRP. I think we should only use `<article>`s for card with titles. And maybe also add a note to the docs that `<article>` should only be used if the card contains a title.

mdo:
Closing as stale. We have #30266 open to track this still as well.

PR #30317, opened 2020-02-29
Hide horizontal scrollbars on code snippets

mdo:
Works in Safari, Firefox, and Chrome so far. ~Pushing to test on my Windows machine.~ Edge and IE11 have full scrollbars always visible. Accessibility-wise, I'm not worried about folks not knowing they can scroll—the majority of our code snippets are quite long horizontally, so having them cut off is a clear indicator of _more_.

### DIFF
--- a/site/assets/scss/_component-examples.scss
+++ b/site/assets/scss/_component-examples.scss
@@ -235,14 +235,9 @@
 //
 
 .highlight {
-  padding: 1rem;
   margin-top: 1rem;
   margin-bottom: 1rem;
   background-color: $gray-100;
-
-  @include media-breakpoint-up(sm) {
-    padding: 1.5rem;
-  }
 }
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
What's the reason behind this change?

BTW

1. I remember we actually had worked around issues in IE/Edge in the past, but we lost the changes somewhere
2. I had this in my TODO list actually and I've been meaning to bring this up `* snippets on one line: horizontal scrollbar overlaps with content`

XhmikosR:
For what is worth, I don't really like this change personally. I find it confusing not seeing the scrollbar.

patrickhlauke:
yeah have to admit i'm not 100% sold on it either, purely visually (though I can understand it's "cleaner" looking to an extent)

image

mdo:
Tabling for now.

PR #30323, opened 2020-03-01
Fix `th` alignment for Safari

### LINKED ISSUE #30321: Table headers and text-align
Screen Shot 2020-02-29 at 9 12 03 AM

The Reboot docs talk about `text-align` being matched to the `<td>`, but as you can see above, `<th>`s aren't left-aligned. This is because we're not explicitly setting `text-align` anywhere up the DOM. We updated our styles while back I believe to lower the selector specificity to allow `.text-*` classes on `.table`, but we didn't reset the default value.

I'd like to fix this by doing one of the following:

- Setting `text-align` on the `<body>` to a default `left` value.
- Setting an explicit `text-align` on the `<th>` in Reboot (my preferred I think, given lowest specificity and an existing selector).
- Setting an explicit `text-align` on the `.table`.

Thoughts @twbs/css-review?

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/issues/30321.

Did some tests and it seemed only Safari had issues figuring out the default text alignment.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -344,11 +344,12 @@ caption {
   caption-side: bottom;
 }
 
-// Matches default `<td>` alignment by inheriting from the `<body>`, or the
-// closest parent with a set `text-align`.
+// Matches default `<td>` alignment by inheriting `text-align`.
+// 1. Fix alignment for Safari
 
 th {
   text-align: inherit;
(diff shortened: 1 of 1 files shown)


XhmikosR:
Shouldn't this be backported?

ysds (approved):

PR #30325, opened 2020-03-02
Update modal docs

### LINKED ISSUE #30055: Modal Documentation for Vertically Centered isn't Correct
The docs for the vertically centered don't appear to fully match what should be shown.  The "Launch Demo Modal" text should be "Vertically centered modal," but what really matters is that there's no class mentioned for `modal-dialog-scrollable` in the **Vertically Centered** section.
image

Very minor issue.

ysds:
* Simplify modal snippets
* Use id for data-target
* Unify the id positions

Closes #30055


MartijnCuppens:
Imo, this PR removes a little too much. I think people like it to just copy-paste the whole modal.

ysds:
I do not think so. The first snippet in the "Scrolling long content" section makes no sense. Also, the current snippets of the size options don't work already because there are no content in the `.modal-content.` To make all snippets that can be copied successfully need too long line.

XhmikosR:
I'm in between personally; I see @ysds's and @MartijnCuppens's points.

/CC @mdo to the rescue :P

patrickhlauke:
FWIW i don't mind the cuts here

mdo (approved):
Cut it! :D

MartijnCuppens (approved):
Fine by me, let's do this

PR #30326, opened 2020-03-02
Prevent overflowing static backdrop modal animation

### LINKED ISSUE #30322: Click on Static Modal Backdrop Scaling up the Modal Content and Adding Scrollbar to Backdrop
Using Bootstrap 4.4.1, I am facing an issue on clicking outside of static backdrop modal. As you can see in attached image when clicking outside of modal it is scaling up the size of Modal Content and eventually adding scrollbar to the backdrop. I have test this on Chrome and Firefox and same issue is happening. Can you please let me know how to stop this?
Here is the image link
Static

    ` <div class="container p-5">
      
      <button type="button" class="btn btn-primary" data-toggle="modal" data- 
        target="#staticBackdrop">
       Launch static backdrop modal
      </button>
      
     <div class="modal fade" id="staticBackdrop" data-backdrop="static" tabindex="-1" role="dialog" aria-labelledby="staticBackdropLabel" aria-hidden="true">
     <div class="modal-dialog modal-lg modal-dialog-centered" role="document">
     <div class="modal-content">
      <div class="modal-header">
        <h5 class="modal-title" id="staticBackdropLabel">Modal title</h5>
        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
          <span aria-hidden="true">&times;</span>
        </button>
      </div>
      <div class="modal-body">
        ...
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Understood</button>
      </div>
      </div>
     </div>
       </div>
         </div>`

ysds:
Fixes #30322

This is a JavaScript fix for respects the current style, but also can be changed to shrink animation by CSS.

demo: https://codepen.io/fellows3/pen/vYOZwqR


js/src/modal.js
```diff
@@ -416,10 +416,23 @@ class Modal {
         return
       }
 
+      const isModalOverflowing = this._element.scrollHeight > document.documentElement.clientHeight
```
MartijnCuppens: woudn't it be better to use the window height instead of the html element?

ysds: I copied this code from here without thinking deeply: 
https://github.com/twbs/bootstrap/blob/ca51b1eefd1d873dcfb8dcd11cf30af051456b11/js/src/modal.js#L448-L450

Johann-S: LGTM this way because we are taking care of the client style

Johann-S:
I would love to see a unit test @ysds, do you think it's possible? 🤔

ysds:
@Johann-S I tried add unit a test, but sorry I don't understand why fail on Firefox.

js/src/modal.js
```diff
@@ -410,10 +410,23 @@ class Modal {
...
+
+      if (!isModalOverflowing) {
+        this._element.style.overflowY = 'hidden'
+      }
+
       this._element.classList.add(CLASS_NAME_STATIC)
-      const modalTransitionDuration = getTransitionDurationFromElement(this._element)
+      const modalTransitionDuration = getTransitionDurationFromElement(this._dialog)
+      EventHandler.off(this._element, TRANSITION_END)
```
Johann-S, line 421: what's the point of adding `EventHandler.off(this._element, TRANSITION_END)` when we use `once` just after that?

XhmikosR:
@ysds ping :)

js/src/modal.js
ysds: Sorry for very late reply! This `EventHandler.off(this._element, TRANSITION_END)` is for addressing quick clicks (click backdrop again while the transition is not finished). If this line is missing, the scroll bar will be displayed.

Johann-S (approved):
LGTM 👍

PR #30328, opened 2020-03-03
Mark popper.js as optional peerDependency

felixfbecker:
This solves the log spam for people using Bootstrap without the JS functionality, with yarn. This can be annoying because not everybody on a large team knows these are harmless in their app, and it trains people to ignore other valid warnings. It only affects the display of warnings, so this is a safe change. Version compatibilities are still checked if the peer dependencies are installed.

```
[4/5] 🔗  Linking dependencies...
warning " > bootstrap@4.4.1" has unmet peer dependency "jquery@1.9.1 - 3".
warning " > bootstrap@4.4.1" has unmet peer dependency "popper.js@^1.16.0".
```

Reference:
https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md
https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md#1130

Related issue with complains about the log spam: https://github.com/twbs/bootstrap/issues/23557

Backport in #30329


XhmikosR:
I don't use yarn myself.

Also, this issue has its drawbacks. We can't know who's going to use the npm package for what exactly.

I'd rather not touch this right now.

felixfbecker:
> Also, this issue has its drawbacks.

What drawbacks do you see?

XhmikosR:
popper.js is **not** an optional peer dependency if one is going to use the JS files which don't bundle popper.js.

felixfbecker:
`peerDependencies` already are not installed automatically and need to be installed manually in addition, you only get a warning. Bootstrap also already errors when Popper is not loaded:

https://github.com/twbs/bootstrap/blob/481540800d3c242c8152bbfed7fd60dbf59437c6/js/src/tooltip.js#L134-L136

So the only difference is that instead of showing a warning on _every_ `yarn install`, for _everyone_, there is only an error for those who actually try to use the features dependent on it.

Existing users must already have popper installed (by the definition of `peerDependencies`) so nothing changes for them, it's purely about the first-install experience and the experience of everyone who doesn't use these features at all.

Please consider that this has been complained about by many people (see all the duplicate issues in https://github.com/twbs/bootstrap/issues/23557)

XhmikosR:
@Johann-S thoughts?

rafa-as:
I totally agree with @felixfbecker

Johann-S:
For me Popper.js should be a `dependency` because  it's used in 3 of our components and it's not nothing.
But people well keep complain because some of them wants to use Bootstrap as a CSS only framework.
IMO we have to make a decision and stick with it.

jquense:
FWIW npm will (did?) Add support for this as well so it's not yarn only. Also npm 7 will auto install peer deps again so the cost of keeping it an (optional) peer deps isn't very high and does help out cases, like react bootstrap where folks only install bootstrap for the css

rafa-as:
Some news here?

mdo:
Closing as stale.

PR #30329, opened 2020-03-03
Mark peer dependencies as optional

felixfbecker:
This solves the log spam for people using Bootstrap without the JS functionality, with yarn. This can be annoying because not everybody on a large team knows these are harmless in their app, and it trains people to ignore other valid warnings. It only affects the display of warnings, so this is a safe change. Version compatibilities are still checked if the peer dependencies are installed.

```
[4/5] 🔗  Linking dependencies...
warning " > bootstrap@4.4.1" has unmet peer dependency "jquery@1.9.1 - 3".
warning " > bootstrap@4.4.1" has unmet peer dependency "popper.js@^1.16.0".
```

Reference:
https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md
https://github.com/yarnpkg/yarn/blob/master/CHANGELOG.md#1130

Related issue with complains about the log spam: https://github.com/twbs/bootstrap/issues/23557

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,14 @@
     "jquery": "1.9.1 - 3",
     "popper.js": "^1.16.0"
   },
+  "peerDependenciesMeta": {
+    "jquery": {
+      "optional": true
+    },
+    "popper.js": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@babel/cli": "^7.8.4",
     "@babel/core": "^7.9.6",


MartijnCuppens:
I'm in favor of this, I sometimes only use the CSS or depend on jQuery/Popper.js libraries which I don't include via npm.

mdo:
We also don't require jQuery anymore as of v4.3.0, right? Can this be removed outright or am I missing something?

XhmikosR:
v4-dev (and thus 4.3.0) requires jQuery.

felixfbecker:
> v4-dev (and thus 4.3.0) requires jQuery.

Yep, that's why I opened two PRs targeting the two branches individually.

mdo:
Any reason to keep this on hold @XhmikosR @Johann-S?

XhmikosR:
@mdo see https://github.com/twbs/bootstrap/pull/30328. I understand it can be annoying but we don't use separate packages thus we can't know what one is going to do. No solution is perfect right now. We need to think how to proceed but for v4 it might be a breaking change.

PR #30330, opened 2020-03-03
Embedded examples cleanup

mdo:
Few visual tweaks to our docs' embedded examples and code snippets:

- Thinner, darker border around the demo
- Removed `margin-bottom` on all `:last-child` within
- Tightened vertical padding on `.highlight` on desktop viewports
- Fixed a busted class
- Nested more CSS

Screen Shot 2020-03-03 at 3 37 12 PM

### DIFF
--- a/site/assets/scss/_component-examples.scss
+++ b/site/assets/scss/_component-examples.scss
@@ -247,19 +247,7 @@
   @include media-breakpoint-up(sm) {
     padding: 1rem 1.5rem;
   }
-}
 
-.bd-content .highlight {
-  margin-right: (-$grid-gutter-width / 2);
-  margin-left: (-$grid-gutter-width / 2);
-
-  @include media-breakpoint-up(sm) {
-    margin-right: 0;
-    margin-left: 0;
-  }
-}
-
-.highlight {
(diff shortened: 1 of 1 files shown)


site/assets/scss/_component-examples.scss
```diff
@@ -99,7 +99,7 @@
   }
 
   + .highlight,
-  + .clipboard + .highlight {
+  + .bd-clipboard + .highlight {
     margin-top: 0;
   }
```
ysds: The `application.js` insert a `div.clipboard` block before every .`div.highlight` block. So the following `margin-top: 1rem` can be removed:

https://github.com/twbs/bootstrap/blob/36a68b1bc0482c3258d4fd81eb144a36e298b2e8/site/assets/scss/_component-examples.scss#L242-L244

Therefore, no need to override the `margin-top: 0` (L101-104) and the below too:

https://github.com/twbs/bootstrap/blob/36a68b1bc0482c3258d4fd81eb144a36e298b2e8/site/assets/scss/_clipboard-js.scss#L5-L12

ysds (approved):

PR #30331, opened 2020-03-04
Removed 'shrink-to-fit' from `viewport` meta

dehghani-mehdi:
Removed 'shrink-to-fit' from `viewport` meta.


dehghani-mehdi:
@ysds why?

PR #30333, opened 2020-03-04
xxl container of 1370px width for screens min-width 1440px

bhadresharya:


ysds:
We'll pass on this (see https://github.com/twbs/bootstrap/issues/20593#issuecomment-243200953)

PR #30336, opened 2020-03-04
Switch versions & docs search

MartijnCuppens:
Something I never got used to: the search field moved from the left side to the right in the `v5` docs.

In the `v4` docs, everything related to navigation was positioned on the left side (main menu in the upper corner, search field just below the navigation, and then the subnavigation).

Can we move the search back to the left?


mdo (approved):

mdo:
Such a small thing, bumping to v5. Merge at will!

Issue #29244, opened 2019-08-13
Table child elements (thead, tbody, tfoot, tr, td, th) should inherit parent background color


tmorehouse:
When working on "sticky header" and "sticky column" features for BootstrapVue (using Bootstrap v4.3.1 CSS), I had to generate a huge amount of SCSS (and conditionals in JS) to ensure that the cells had a background color (they were transparent by default, as table cells do not inherit their parent background color).

It would be nice if Bootstrap set up the table child elements to inherit their parent element's background color, so if a color is applied to a cell, row, or table they will have a solid background.

We also came up with a simple way to generate zebra striping and hover state with much less lines of CSS by cheating with a transparent solid gradient background image (which gets overlayed on the background color).

Something like this would be great in Bootstrap (v5), and it probably would decrease the amount of CSS rules generated (specifically for the them color variants):

```scss
.table {
  // Ensure table has a background color by default, plus handle reset
  // of color for nested tables
  background-color: if($table-bg, $table-bg, $body-bg);

  // Ensure child elements inherit their parent element's background color
  // As table child elements do not inherit bg-color by default.
  // Must be defined before any table theme colors are defined
  thead,
  tbody,
  tfoot,
  tr,
  th,
  td {
    background-color: inherit;
  }

  // ...
  // Define table theme background and border color stuff here
  // But no hover/striping stuff, as it is handled below
  // ...

  // Zebra striping (regardless of background color)
  // Also works when cells have their on theme color variant applied
  // table cell backgrounds remain `solid` so if sticky headers/columns is
  // used, the header cells won't show the scrolled table contents through them
  &.table-striped {
    > tbody {
      > tr:nth-of-type(#{$table-striped-order}) {
        // Could do this to reduce the number of rules generated
        // > :nth-child(1n) {
        > td,
        > th {
          // Simple way to "darken" a row, regardless of it's `.table-<variant>` color
          // using transparent black, which is overlayed on top of bg color
          // No need to generate special striping states for each theme color
          background-image: linear-gradient($table-accent-bg, $table-accent-bg);
          background-repeat: no-repeat;
        }
      }
    }

    // Dark mode Zebra striping
    &.table-dark {
      > tbody {
        > tr:nth-of-type(#{$table-striped-order}) {
          // Could do this to reduce the number of rules generated
          // > :nth-child(1n) {
          > td,
          > th {
            // Simple way to "lighten" a row, regardless of it's `.bg-<variant>` color,
[...]

PR #30342, opened 2020-03-05
De-dupe .table-dark, add new border color utilities, and update table docs

### LINKED ISSUE #24529: Inconsistant .table-active color

### LINKED ISSUE #27879: .table-dark is used in two conflicting contexts

### LINKED ISSUE #28160: Suggestion for .table-hover

### LINKED ISSUE #28480: v5: Revisit table styles
#26678 has conflicts and tried to tackle a ton, so let's revisit style scoping, Sass maps, and more in v5.

mdo:
Alternative to #30193. _This ended up being a couple changes in one, hence the length title, but the two are connected._

---

### De-duped .table-dark

Rather than deviate from the `$theme-colors` pattern we have elsewhere, this PR removes the `.table-dark` for changing the background, color, and border of a `.table`. Instead, this can be accomplished through all utilities.

Also removes thead-dark/thead-light modifiers for utilities.

### New border-color utils

To support the utilities-driven approach for replacing `.table-dark`, I've added a few new `border-color` utilities. These are a subset of our available grays, but I do believe they provide an improved set of options without including all the grays (and thus duplicating `$gray-900` and `$dark` options). The new options have been added to the borders utilities docs page.

### New background-color utils

To match the border utilities, I've added one tint lighter for two gray backgrounds (didn't add `$gray-100` as it's the same as `$light`). This way you can choose from additional gray backgrounds and still have a slight contrast in the border colors. Definitely interested in thoughts on this.

### Updated table docs

This also obviously updates the examples in the tables docs to reflect the new dark table approach, but also clarifies how inheritance comes into play with the table styles and the table borders. It also adds a new multiple `<tbody>` example and tweaks a few other bits and pieces.

https://deploy-preview-30342--twbs-bootstrap.netlify.com/docs/4.3/content/tables/

---

Closes #28480, fixes #27879, nullifies #30046, fixes #28160, and fixes #24529.

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -108,60 +108,6 @@
 @include table-row-variant(active, $table-active-bg);
 
 
-// Dark styles
-//
-// Same table markup, but inverted color scheme: dark background and light text.
-
-// stylelint-disable-next-line no-duplicate-selectors
-.table {
-  .thead-dark {
-    th {
-      color: $table-dark-color;
-      background-color: $table-dark-bg;
-      border-color: $table-dark-border-color;
-    }
-  }
-
-  .thead-light {
-    th {
-      color: $table-head-color;
-      background-color: $table-head-bg;
-      border-color: $table-border-color;
-    }
-  }
-}
-
-.table-dark {
-  color: $table-dark-color;
-  background-color: $table-dark-bg;
-
-  th,
-  td,
-  thead th {
-    border-color: $table-dark-border-color;
-  }
-
-  &.table-bordered {
-    border: 0;
-  }
-
-  &.table-striped {
-    tbody tr:nth-of-type(#{$table-striped-order}) {
-      background-color: $table-dark-accent-bg;
-    }
-  }
-
-  &.table-hover {
-    tbody tr {
-      &:hover {
-        color: $table-dark-hover-color;
-        background-color: $table-dark-hover-bg;
-      }
-    }
-  }
-}
-
-
 // Responsive tables
 //
 // Generate series of `.table-responsive-*` classes for configuring the screen
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -1,3 +1,5 @@
+// stylelint-disable comment-empty-line-before
+
 //
 // Basic Bootstrap table
 //
@@ -8,11 +10,17 @@
   color: $table-color;
   vertical-align: $table-cell-vertical-align;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
+  border-color: $table-border-color;
 
   th,
   td {
     padding: $table-cell-padding;
-    border-bottom: $table-border-width solid $table-border-color;
+    // Table cells will only inherit if you use longhand properties
+    /* clean-css ignore:start */
+    border-color: inherit;
+    border-bottom-style: solid;
+    border-bottom-width: $table-border-width;
+    /* clean-css ignore:end */
   }
 
   tbody {
@@ -21,11 +29,14 @@
 
   thead th {
     vertical-align: bottom;
-    border-bottom-color: $table-head-border-color;
   }
 
   tbody + tbody {
-    border-top: (2 * $table-border-width) solid $table-border-color;
+    /* clean-css ignore:start */
+    border-color: inherit;
+    border-top-style: solid;
+    border-top-width: (2 * $table-border-width);
+    /* clean-css ignore:end */
   }
 }
 
@@ -47,18 +58,14 @@
 // Add or remove borders all around the table and between all the columns.
 
 .table-bordered {
-  border: $table-border-width solid $table-border-color;
+  border-width: $table-border-width;
 
   th,
   td {
-    border: $table-border-width solid $table-border-color;
-  }
-
-  thead {
-    th,
-    td {
-      border-bottom-width: 2 * $table-border-width;
-    }
+    /* clean-css ignore:start */
+    border-style: solid;
+    border-width: $table-border-width;
+    /* clean-css ignore:end */
   }
 }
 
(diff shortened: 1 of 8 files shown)


mdo:
For next steps, I'm wondering if we also remove the `.table-*` variants and opt for all background, border, and color variations... Might be pushing it too much.

mdo:
Yeah we could perhaps ditch those and create a broader suite of utilities. It was a huge ask awhile back to have light and dark background options. Might be a v6 thing.

site/content/docs/4.3/content/tables.md
```diff
@@ -84,12 +119,10 @@ You can also invert the colors—with light text on dark backgrounds—with `.ta
 </table>
 {{< /example >}}
 
-### Striped rows
-
-Use `.table-striped` to add zebra-striping to any table row within the `<tbody>`.
+And because this happens with utilities, you have some flexibility in how you style your tables relative to their context. For example, if your page is already dark gray with lighter text, you don't need to set every utility on the `<table>`—just `.border-gray-700`.
```
MartijnCuppens: It took a while before I noticed what was going on here, I was just screening the page without reading the intro text and thought padding was somehow added to the table. I would emphasise the dark page notice more (bold text?) or add what utilities were added to the parent (because you can't see them in the snippet)

site/content/docs/4.3/content/tables.md
```diff
@@ -285,7 +358,43 @@ Add `.table-borderless` for a table without borders.
...
+      <td>@twitter</td>
+    </tr>
+  </tbody>
+</table>
+{{< /example >}}
+
+And also works with table color classes:
+
+{{< example >}}
+<table class="table table-striped">
```
MartijnCuppens: This table looks pretty ugly 😄

site/content/docs/4.3/content/tables.md
```diff
@@ -575,15 +720,106 @@ Regular table background variants are not available with the dark table, however
 {{< partial "callout-warning-color-assistive-technologies.md" >}}
 {{< /callout >}}
 
+## Nesting
```
MartijnCuppens: This needs more explanation. Also, I'm not a fan of the styles leaking to the nested table. But maybe we should have a look at this afterwards to prevent this PR from becoming too huge?

scss/_utilities.scss
```diff
@@ -77,7 +77,16 @@ $utilities: map-merge(
     "border-color": (
       property: border-color,
       class: border,
-      values: map-merge($theme-colors, ("white": $white))
+      values: map-merge(
+        $theme-colors,
+        (
+          "white": $white,
+          "gray-200": $gray-200,
```
MartijnCuppens: The default border color is 300, weird not to have it here:
https://github.com/twbs/bootstrap/blob/8abccff7b3939334c61286df448d0f730da85325/scss/_variables.scss#L318

site/content/docs/4.3/content/tables.md
mdo: Yeah I know lol... remove it? Find a better example?

site/content/docs/4.3/content/tables.md
mdo: Oh yeah, I added the nesting to show how it happens, and it basically convinced me we needed to figure that out. I'm thinking we could find a balance of styling a single first descendant?

Something like this maybe:

```scss
.table {
 > tbody  {
    td,
    th {
      ...
    }
  }
}
```

tmorehouse: > ```scss
>  .table {
>   > tbody  {
>      td,
>      th {
>        ...
>      }
>    }
>  }
> ```

Would still leak into nested table thead and tfoot (as well as tbody)

MartijnCuppens:
Wow, the `linear-background()` hack is a genius move!

tmorehouse:
The hover and striping background gradient rules need to target the inner `td/th` cells of the body  rows to correctly handle individual cell background colors (and preferably use child selectors to limit leaking into child tables... i.e. preventing hover effects on nested table thead/tfoot) (https://github.com/twbs/bootstrap/issues/29244#issuecomment-595610814)

tmorehouse:
One may even want to add in a `table-hover-light` modifier that uses a light translucent gradient when users have dark (almost black) background color.  It would just add two more generated rules and one new SASS variable

scss/_tables.scss
```diff
@@ -8,11 +8,15 @@
   color: $table-color;
   vertical-align: $table-cell-vertical-align;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
+  border-color: $table-border-color;
 
   th,
   td {
     padding: $table-cell-padding;
-    border-bottom: $table-border-width solid $table-border-color;
+    // Table cells will only inherit if you use longhand properties
```
tmorehouse: Would some CSS minifiers (i.e. cssnano, et al), potentially munge these by trying to merge them into a single shorthand property?

mdo: We use clean-css, which did in fact recombine these. I've disabled that rule for now.

mdo:
Rebased and recombulated the commits here in case we want to pull out the border and background utility changes.

ysds:
Can we use the [unset](https://developer.mozilla.org/en-US/docs/Web/CSS/unset) to nested tables if dropping IE11 in v5?

scss/_tables.scss
XhmikosR: Maybe we should disable this optimization globally @mdo?

MartijnCuppens:
> Can we use the unset to nested tables if dropping IE11 in v5?

We could, but I guess CSS specificity will be a PITA

ysds:
> We could, but I guess CSS specificity will be a PITA

Oh I forgot that..

MartijnCuppens:
CSS custom properties on the other hand... https://codepen.io/MartijnCuppens/pen/eYNVmzp

@mdo, we probably better switch to a CSS custom properties approach which will help us out with the nesting issues.

tmorehouse:
@MartijnCuppens Intereting idea. 

One might have to be careful which the CSS variable names.. perhaps `--table-bg-color`... if similar variable names were used for other things (such as input group background colors, alert background colors, etc) when they are nested inside a table.

MartijnCuppens:
@tmorehouse, of course. That was just a quick demo 😉

scss/_tables.scss
tmorehouse: Maybe there should be a note in the SCSS part of the docs about SASS compilation and minifiers possibly interfering with styles, as some users may be the source SCSS files and a different CSS minifier other than `clean-css` (i.e. `cssnano`, etc)

mdo: Would you mind opening an issue for that @tmorehouse?

mdo:
So I need to circle back around to this and check in on some things around inheritance and nesting. I don't think mixing the CSS vars and Sass makes sense necessarily? Maybe I'm missing something there.

scss/_tables.scss
tmorehouse: OK...  will do

tmorehouse: Opened #30429

MartijnCuppens:
I'm working on an alternative solution, will open another PR soon.

stgm:
Has the use of linear-gradient been performance-tested in some way? I have an example where this makes Safari rendering extremely slow when scrolling. Removing the gradient helps, but obviously kills my striping. If this wasn't expected, I can have a look at creating an example and posting an issue.

MartijnCuppens:
> If this wasn't expected, I can have a look at creating an example and posting an issue.

Please do if there are any performance issue

PR #30345, opened 2020-03-06
Fix `<legend>` wrapping bug

### LINKED ISSUE #29712: Flex container doesn't display correctly next to a legend
Since #28917, the `legend` element is no longer rendered with `display: block` but it has `float: left` instead. There is at least an issue when displaying a grid just next to `legend` as it is rendered at the right and not under (see: https://jsfiddle.net/du95trv3/2/). I let you see what is the best way to fix that, normalize.css for example defines `display: table` for `legend`...

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/issues/29712. Alternative for https://github.com/twbs/bootstrap/pull/29831 which works in Edge.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -500,8 +500,11 @@ fieldset {
   border: 0; // 2
 }
 
-// 1. By using `float: left`, the legend will behave like a block element
+// 1. By using `float: left`, the legend will behave like a block element.
+//    This way the border of a fieldset wraps around the legend if present.
 // 2. Correct the text wrapping in Edge.
+// 3. Fix wrapping bug.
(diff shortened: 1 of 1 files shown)


ysds:
What is a reason not to use `display: table` like normalize.css?

mdo:
Might be imagining it, but I feel like we ran into an issue with the table approach? Can't remember...

MartijnCuppens:
The reason is ease of theming:
v4: https://codepen.io/MartijnCuppens/pen/vYOdBeo
v5: https://codepen.io/MartijnCuppens/pen/VwLQZWN

mdo (approved):

XhmikosR:
@MartijnCuppens can you rebase this please?

PR #30353, opened 2020-03-06
Fix date inputs heights in safari

MartijnCuppens:
It seems https://github.com/twbs/bootstrap/issues/11266 is solved in the meantime and this was causing the inputs to shrink (date & time inputs):
image

### DIFF
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -15,6 +15,7 @@
   background-color: $input-bg;
   background-clip: padding-box;
   border: $input-border-width solid $input-border-color;
+  appearance: none; // Fix appearance for date inputs in Safari
 
   // Note: This has no effect on <select>s in some browsers, due to the limited stylability of `<select>`s in CSS.
   @include border-radius($input-border-radius, 0);
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -466,19 +466,6 @@ button,
   border-style: none;
 }
 
-// Remove the default appearance of temporal inputs to avoid a Mobile Safari
-// bug where setting a custom line-height prevents text from being vertically
-// centered within the input.
-// See https://bugs.webkit.org/show_bug.cgi?id=139848
-// and https://github.com/twbs/bootstrap/issues/11266
-
-input[type="date"],
-input[type="time"],
-input[type="datetime-local"],
-input[type="month"] {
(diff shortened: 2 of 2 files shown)


XhmikosR:
@MartijnCuppens which iOS versions did you test?

https://github.com/twbs/bootstrap/blob/d79830c4ee407efdafd754aaf4dd729a33fe3a37/.browserslistrc#L10

mdo:
Tangent, but saw the comment and figured I'd look up iOS distribution: https://developer.apple.com/support/app-store/. We can probably move to iOS 12+ given the browser share there. v4 covers us on older versions.

ysds:
@MartijnCuppens Could you create before and after test cases?

MartijnCuppens:
## Before
ios13:
image
ios12:
image
ios11:
image
ios9:
image

## After

ios13:
image
ios12:
image
ios11:
image
ios9:
image


I can't reproduce the initial issue from back in 2013.

XhmikosR:
Shouldn't this be also backported?

MartijnCuppens:
Yes, made a separate PR since reboot changed a lot in v5

ysds:
@MartijnCuppens Do you intend that this affect input style with `.form-control` class?

Before: https://codepen.io/fellows3/pen/BaNmvvN

image


After: https://codepen.io/fellows3/pen/RwPjEeO

image

MartijnCuppens:
In that case we should only change the appearance for `.form-control`s. I now pushed some changes and set the height of the `::-webkit-date-and-time-value` pseudo selector to prevent it from collapsing.

scss/forms/_form-control.scss
```diff
@@ -28,6 +29,11 @@
     border: 0;
   }
 
+  // Set the height of date inputs in Safari
+  &::-webkit-date-and-time-value {
+    height: if(unitless($input-line-height), $input-line-height * 1em, $input-line-height);
```
ysds: I'm not sure why this needed. I remove this and checked on iOS 13 Safari, but I can't see the effect.

MartijnCuppens: I thought we could remove the min-height when this is set. But suddenly I rememered https://github.com/twbs/bootstrap/pull/29124, so this is moot.

ysds (approved):

PR #30354, opened 2020-03-06
v5: Add new `xxl` grid tier and move ToC to a sidebar

mdo:
Screen Shot 2020-03-06 at 11 06 31 AM

Compared to v4, v5's docs have a fixed-width, center aligned container. I wanted to keep that look of a centered container, but add a third column to our docs for fitting in a sidebar table of contents. Lately, the ToC has been wildly unhelpful as the constant scrolling up and down to find a link isn't saving me anytime.

Getting that third column though means either sacrificing significant width from the content column, going fullscreen again (don't want to do), or adding a wider container. I chose the last option and rather than write CSS that limits an even wider container to just our docs, I'm adding it to the core.

Let the feedback rip!

Preview: <https://deploy-preview-30354--twbs-bootstrap.netlify.com/>


mdo:
Right now this comes with the need to update documentation on the number of grid tiers, changing bundlesize, etc. It's a big increase, and not just motivated by adding a sidebar ToC—lots of folks have asked for a larger device tier.

MartijnCuppens:
I'm not 100% sure it's a good idea to add a new grid tier just because we need it for our docs. However we can dig up this idea (which i liked very much, see my _I like this approach_ comment): https://github.com/twbs/bootstrap/pull/28500 and customize the grid breakpoints only for the docs.

I would also move the TOC next to the introduction, pages like the card documentation look a bit weird: https://deploy-preview-30354--twbs-bootstrap.netlify.com/docs/4.3/components/card/

image

florianlacreuse:
We already add some grid tiers in our projects. Therefore, I'm in favour of adding a new size.

My only concern is about the naming. Why `xxl` and not `2xl`? In our case we will need a extra grid tier (for a 1540px container for instance), and in this case `3xl` looks prettier (IMHO) than `xxxl`.

ffoodd:
+1, we already added a `-xxl` tier in Boosted in v4 :)

MartijnCuppens:
Current `bootstrap.min.css` size: `21.32kb` (gzip) 
This PR `bootstrap.min.css` size: `22.15kb` (gzip)

I don't really think this is a good idea. Containers will be `1320px` which is way to big if you do not have any sidebars. Reading text will be like watching a tennis game.

midzer:
Is it possible to write a function that extends our grid system?

Have current `$grid-breakpoints` and `$container-max-widths` as base and allow new variable `$extra-grid-tiers: null !default` to specify how many additional extra large tiers are generated on top of `xl` following scheme `2xl`, `3xl` etc as @florianlacreuse suggested.

mdo:
Wondering if we can get scrollspy to work with this table of contents nav. Thoughts @xhmikosr?

site/content/docs/4.3/layout/breakpoints.md
```diff
@@ -0,0 +1,207 @@
...
+
+// Example: Style from medium breakpoint and down
+@include media-breakpoint-down(md) {
+  .custom-class {
+    display: block;
+  }
+}
+{{< /highlight >}}
+
+These mixins take those declared breakpoints, subtract `.02px` from them, and use them as our `max-width` values. For example:
```
XhmikosR, line 154: Could we maybe add a link/explanation why we substract .02px?

XhmikosR:
> Wondering if we can get scrollspy to work with this table of contents nav. Thoughts @XhmikosR?

I remember I tried this in the past but I faced some issues with not being able to customize the ToC IDs.

Better split the ToC patches to a separate PR and we can discuss this further there 🙂

mdo:
I’m still okay with shipping the general ToC updates here. With or without the scrollspy, this is still an improvement :).

site/content/docs/4.3/layout/breakpoints.md
mdo: This is already there in the callout below. I've made that callout a warning instead of an info, added a bit of a title to it, and clarified some stuff.

MartijnCuppens:
@mdo, thoughts on switching to `2xl` like @midzer and @florianlacreuse suggest?

site/content/docs/4.3/components/list-group.md
```diff
@@ -102,7 +102,7 @@ Add `.list-group-horizontal` to change the layout of list group items from verti
 
 {{< example >}}
 {{< list-group.inline >}}
-{{- range $.Site.Data.breakpoints }}
+{{- range $.Site.Data.breakpoints -}}
```
XhmikosR, line 105: Is this really needed here?

site/content/docs/4.3/layout/z-index.md
```diff
@@ -0,0 +1,30 @@
...
+
+Several Bootstrap components utilize `z-index`, the CSS property that helps control layout by providing a third axis to arrange content. We utilize a default z-index scale in Bootstrap that's been designed to properly layer navigation, tooltips and popovers, modals, and more.
+
+These higher values start at an arbitrary number, high and specific enough to ideally avoid conflicts. We need a standard set of these across our layered components—tooltips, popovers, navbars, dropdowns, modals—so we can be reasonably consistent in the behaviors. There's no reason we couldn't have used `100`+ or `500`+.
+
+We don't encourage customization of these individual values; should you change one, you likely need to change them all.
+
+{{< highlight scss >}}
+{{< zindex.inline >}}
+{{- $file := readFile "scss/_variables.scss" -}}
```
XhmikosR, line 16: I guess we could/should switch this to the `scss-docs` shortcode instead of an inline one? Unless you want to reduce the spaces after the variables, but it's your call.

MartijnCuppens: We should switch to remain consistent.

mdo:
> thoughts on switching to `2xl` like @midzer and @florianlacreuse suggest?

I think I'm inclined to keep it as `xxl` for the time being. We don't have much information on how folks have customized things to date, and I've only heard a few asks for a `2xl` name that'd allow for even larger breakpoints.

mdo:
Any other reviews from @twbs/css-review or should I go ahead and merge? :)

MartijnCuppens:
Found some `lg` references in the docs and pushed some changes to fix https://github.com/twbs/bootstrap/pull/30354#issuecomment-596124195 gonna do a recheck one of these days

mdo:
Good call @MartijnCuppens, just found some more fixes and pushed them as well.

mdo:
Squashed the last several commits into one for updating docs across the board. Taking another spin through and then likely merging :D.

PR #30355, opened 2020-03-06
Remove babel-eslint

XhmikosR:
It doesn't seem this is needed


XhmikosR:
We might need to add this back if needed in the future, but right now it seems it's redundant since tests pass.

Johann-S (approved):

PR #30358, opened 2020-03-07
Switch to strings constants.

### LINKED ISSUE #20011: v5: JS constant declaration / minification

XhmikosR:
This allows the minifier to mangle the constants. It also allows the linter to find unused strings properly.

This is an update/continuation of #21711.

Fixes #20011

| File | Before | After | Diff |
| --- | -------- | ----- | ---- |
| bootstrap.bundle.min.js | 23.61 kB | 22.61 kB | -1.00 kB (-4.23 %) |
| bootstrap.min.js | 17.04 kB | 16.08 kB | -0.96 kB (-5.63 %) |

Preview: <https://deploy-preview-30358--twbs-bootstrap.netlify.com/>

TODO:

- [x] popover
- [x] tooltip
- [x] take care of any newly added `TODO` comments

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -38,7 +38,7 @@
     },
     {
       "path": "./dist/js/bootstrap.bundle.min.js",
-      "maxSize": "24 kB"
+      "maxSize": "23 kB"
     },
     {
       "path": "./dist/js/bootstrap.esm.js",
@@ -54,7 +54,7 @@
     },
     {
       "path": "./dist/js/bootstrap.min.js",
-      "maxSize": "17.5 kB"
+      "maxSize": "16.5 kB"
     }
   ]
 }
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -28,21 +28,15 @@ const DATA_KEY = 'bs.alert'
 const EVENT_KEY = `.${DATA_KEY}`
 const DATA_API_KEY = '.data-api'
 
-const Selector = {
-  DISMISS: '[data-dismiss="alert"]'
-}
+const SELECTOR_DISMISS = '[data-dismiss="alert"]'
 
-const Event = {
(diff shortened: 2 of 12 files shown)


XhmikosR:
I could use some help with popover and tooltip Event getter

Johann-S (approved):

PR #30361, opened 2020-03-07
v4: Remove appearance from date inputs

MartijnCuppens:
Backport #30353

Things are a little different for v4, since the `.form-control` class can also be added to `select`s. That's why I scoped the `appearance: none` to the date input fields here.


XhmikosR:
@MartijnCuppens please fix the commit message to reference what this changes.

ysds (approved):

PR #30377, opened 2020-03-11
v5: drop Internet Explorer support

XhmikosR:
Preview: <https://deploy-preview-30377--twbs-bootstrap.netlify.com/>


XhmikosR:
@Johann-S I think we could probably remove the `makeArray` function too? https://caniuse.com/#feat=mdn-api_nodelist_foreach

abarreraaponte:
Yes please

richlira:
Yes please!

site/content/docs/4.3/migration.md
```diff
@@ -11,7 +11,7 @@ toc: true
 
 See the browser and devices page for details on what is currently supported in Bootstrap 5. Since v4, here's what's changed to our browser support:
 
-- Dropped support for Internet Explorer NN
+- Dropped support for Internet Explorer 11
```
mdo:
```suggestion
- Dropped support for Internet Explorer 10 and 11
```

Johann-S:
before that we should maybe upgrade our browserlist to Edge >= 16, what do you think @XhmikosR ?

XhmikosR:
@Johann-S that's already done in #30378

Johann-S:
so I'm definitely in favor of removing `makeArray` 👍

XhmikosR:
I split the `makeArray` removal patch to another branch because it potentially affects more browsers than IE.

The PR needs another review for the doc changes, a final rebase/squash of some of the patches and then we can merge it.

site/content/docs/4.3/getting-started/browsers-devices.md
```diff
@@ -8,7 +8,7 @@ toc: true
 
 ## Supported browsers
 
-Bootstrap supports the **latest, stable releases** of all major browsers and platforms. On Windows, **we support Internet Explorer 11 / Microsoft Edge**.
+Bootstrap supports the **latest, stable releases** of all major browsers and platforms. On Windows, **we support Microsoft Edge**.
```
XhmikosR: Maybe we should remove the Windows-specific sentence since it's covered by the previous one?

site/content/docs/4.3/getting-started/theming.md
```diff
@@ -221,7 +221,7 @@ You can also specify a base color with our color map functions:
 
 #### Escape SVG
 
-We use the `escape-svg` function to escape the `<`, `>` and `#` characters for SVG background images. These characters need to be escaped to properly render the background images in IE.
+We use the `escape-svg` function to escape the `<`, `>` and `#` characters for SVG background images.
```
XhmikosR: This reminds me, maybe we should revisit the escape SVG function later.

site/content/docs/4.3/getting-started/browsers-devices.md
MartijnCuppens: Updated this: https://github.com/twbs/bootstrap/pull/30377/commits/642ff6b15b3da1b681ad5ce5713e1e7ab4088894

site/content/docs/4.3/getting-started/browsers-devices.md
```diff
@@ -8,7 +8,7 @@ toc: true
 
 ## Supported browsers
 
-Bootstrap supports the **latest, stable releases** of all major browsers and platforms. On Windows, **we support Internet Explorer 11 / Microsoft Edge**.
+Bootstrap supports the **latest, stable releases** of all major browsers and platforms. This also includes the latest version of Legacy Edge (EdgeHTML layout enginge).
```
XhmikosR:
```suggestion
Bootstrap supports the **latest, stable releases** of all major browsers and platforms. This also includes the latest version of Legacy Edge (EdgeHTML layout engine).
```

MartijnCuppens (approved):
Drop it like it's hot

ysds (approved):

mdo (approved):

rbonomo:
Supporting IE would be a wise and thoughtful decision. You'll end up with people using a fork that supports IE. Supporting IE is a small challenge to overcome.

mohe2015:
I think they want to keep supporting bootstrap 4.

mirap:
> Supporting IE would be a wise and thoughtful decision. You'll end up with people using a fork that supports IE. Supporting IE is a small challenge to overcome.
> 
> There is an ongoing Hacker News discussion that links here for everyone's context: https://news.ycombinator.com/item?id=22802003
> 
> **Today's reality is that there are IE users, and by not supporting them you are excluding them.**

Supporting Macromedia Flash would be a wise and thoughtful decision. Supporting Macromedia Flash is a small challenge to overcome.

Today's reality is that there are Macromedia Flash users, and by not supporting them you are excluding them.

withinboredom:
> Supporting IE would be a wise and thoughtful decision. You'll end up with people using a fork that supports IE. Supporting IE is a small challenge to overcome.
> 
> There is an ongoing Hacker News discussion that links here for everyone's context: https://news.ycombinator.com/item?id=22802003
> 
> **Today's reality is that there are IE users, and by not supporting them you are excluding them.**

I think it’s more useful to point out which demographic would be using internet explorer instead of just saying “some users use it.” Such as military, government, schools, corporations, etc. These demographics are not reflected well in caniuse due to the fact that they’re already excluded from most of the popular web. However, these demographics are the most likely to use a framework such as this one, albeit, internally. 

You may be shooting yourself in the foot here.

LegoStormtroopr:
To everyone saying, if you want to keep supporting IE just use Bootstrap4, [Bootstrap4 is End-Of-Life at the end of this year](https://github.com/twbs/release):

image

The decision to drop IE support is understandable, but as @withinboredom pointed out a lot of the education, government and big banking sectors still use IE (and do so with Microsofts blessing).
Also there are a lot of startups that support these industries, so its not always a case of saying "hey if they are so big, just change the policy".

Remove IE support from Bootstrap5, but at least give us *some* longer support for Bootstrap4.

keithn:
I think if you remove it, bootstrap no longer becomes a simple choice.  BS5 suddenly turns into a bet, where BS4 is a safer bet, but it will become unmaintained.  

In my experience, the majority of IE11 users are often the users with $s who want to pay for products. Mostly because they are bigger companies/organizations.

tough choice.

mdo:
This is the same issue as moving from v3 to v4 when we dropped support for IE9, and the same as moving from v2 to v3 when we dropped IE8. You can still use the older versions despite new ones coming out.

XhmikosR:
Just to give some context.

We went back and forth with this issue for quite some time. I, myself, was in favor of keeping IE support until it became an issue for us. In our last team call, we evaluated the pros and cons of keeping IE support and the pros outweigh the cons.

1. **v4 will be supported for quite some time** which means people who want to support IE can still use this version, like people could use v3 after v4 was released and we kept supporting it for a long time.
2. Dropping IE support isn't something we decided to do out of spite or something; it really allows us to improve things for everyone else. And by seeing the numbers, the people that will benefit are multiple times more.
3. The first clear gain of dropping IE is the use of CSS variables. Our CSS size (bootstrap.min.css) is already down by **10%**, and the decrease will keep getting bigger. Less CSS is always better for everyone.
4. The JS size is also down; there are almost no polyfills needed for the features we use, and let's not forget the jQuery removal in v5

Now, in an ideal world we could have all of the above. But our world is not ideal.

PR #30379, opened 2020-03-11
Set .btn padding for default sized input-group

toooni:
This fixes an issue if the default value of `$btn-padding-y` or `$btn-padding-x` has been changed.

When overwriting the default .btn paddings:
```
$btn-padding-y: .625rem;
$btn-padding-x: 1.125rem;
```

This simple `input-group`
```
<div class="input-group">
    <input type="text" class="form-control" placeholder="Name" aria-label="Name" aria-describedby="button-addon2">
        <button class="btn btn-primary" type="button" id="button-addon2">Button</button>
    </div>
</div>
```

Will result in the following:
image

The provided fix will result in this:
image

This issue does only happen on the default size of `.input-group` with a `.btn`. There is already a solution for the sized ones: `.input-group-lg` (https://github.com/twbs/bootstrap/blob/2a2b0b5abe7b44b519c9e19e8876b75608979dd0/scss/forms/_input-group.scss#L135) and `.input-group-sm` (https://github.com/twbs/bootstrap/blob/2a2b0b5abe7b44b519c9e19e8876b75608979dd0/scss/forms/_input-group.scss#L154)

It seems like it was also fixed for the non sized `.input-group-text` but the default size `.btn` was not being thought of: https://github.com/twbs/bootstrap/blob/2a2b0b5abe7b44b519c9e19e8876b75608979dd0/scss/forms/_input-group.scss#L102

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -70,7 +70,7 @@
   > .btn {
     padding: $input-padding-y $input-padding-x;
   }
-  
+
   // Ensure buttons are always above inputs for more visually pleasing borders.
   // This isn't needed for `.input-group-text` since it shares the same border-color
   // as our inputs.
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -67,6 +67,10 @@
 .input-group-append {
   display: flex;
 
+  > .btn {
+    padding: $input-padding-y $input-padding-x;
+  }
+  
   // Ensure buttons are always above inputs for more visually pleasing borders.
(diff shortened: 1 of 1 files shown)


mdo:
Input groups should grow to stretch their buttons and inputs. This screenshot is from changing the `padding` manually in the web inspector for now, but that shouldn't matter—input groups are flex containers and have no set heights.

Screen Shot 2020-09-18 at 3 20 22 PM

Regardless, this fix isn't acceptable given this creates issues for large and small buttons.

Thanks, though!

PR #30383, opened 2020-03-11
fix: ensure totype always return stringified null when null passed

### LINKED ISSUE #29996: Dropdown doesn't work with null popperConfig
Error Message will be 
  `  DROPDOWN: Option "popperConfig" provided type "window" but expected type "(null|object)". `

Could be caused by Angular, as i and other users use Bootsrtap in conjunction with Angular.
Other users have the same problem: 
https://stackoverflow.com/questions/59628601/dropdown-bootstrap-4-doesnt-work-in-production-of-angulars-application 

- Operating system and version (Windows, macOS, Android, iOS, Win10 Mobile)
Windows
Firefox 68 ESR
- [Reduced test case](https://css-tricks.com/reduced-test-cases/) and suggested fix using [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/)

Cause seems to be with the toType method, and the typecheck in Dropdown.

```
$ typeof null
> "object"
$ const toType = obj => ({}.toString.call(obj).match(/\s([a-z]+)/i)[1].toLowerCase())
$ toType(null)
> "window"
```

Expected type for popperConfig is [null | object ](https://getbootstrap.com/docs/4.4/components/dropdowns/#options)

For me it is not intuitive why toType(null) should return window, and not null

Johann-S:
fixes: #29996

### DIFF
--- a/js/tests/unit/util/index.spec.js
+++ b/js/tests/unit/util/index.spec.js
@@ -198,8 +198,9 @@ describe('Util', () => {
   })
 
   describe('typeCheckConfig', () => {
+    const namePlugin = 'collapse'
+
     it('should check type of the config object', () => {
-      const namePlugin = 'collapse'
       const defaultType = {
         toggle: 'boolean',
         parent: '(string|element)'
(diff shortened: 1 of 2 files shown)


neil-benn:
Just to check - this is not on the master because of 'This branch is out-of-date with the base branch' error?

XhmikosR:
> Just to check - this is not on the master because of 'This branch is out-of-date with the base branch' error?

No, it's because this is not merged yet.

js/tests/unit/util/index.spec.js
```diff
@@ -213,6 +214,34 @@ describe('Util', () => {
         Util.typeCheckConfig(namePlugin, config, defaultType)
       }).toThrow(new Error('COLLAPSE: Option "parent" provided type "number" but expected type "(string|element)".'))
     })
+
+    it('should return null to string when null passed', () => {
```
XhmikosR: Can we reword this and the other description a little better please? :)

Johann-S: reworded, do not hesitate if you have any suggestion of rewording

XhmikosR (approved):

Issue #30318, opened 2020-02-29
Small & Large Progress


harnishdesign:
1rem height is by default for Progress currently. if, We will want increase or decrease progress height. So, We will need to add inline css style. So, Here, I am request you to for add `.progress-sm` (Small progress) and `.progress-lg` (Large progress).

https://stackoverflow.com/questions/23414333/thinner-bootstrap-progress-bars/60463601#60463601

mdo:
All that's needed is to change the height as the docs mention: https://getbootstrap.com/docs/4.4/components/progress/#height. I don't think we need the variants?

harnishdesign:
> I don't think we need the variants?

I know. But, it have possible with inline css.

MartijnCuppens:
I agree with @mdo, the height of the default progress bar is configurable in Sass.


What we could do though is switching the height to a `em` height. This way we can size the progress bar and font size with font size utilities once they land.

mdo:
> What we could do though is switching the height to a `em` height. This way we can size the progress bar and font size with font size utilities once they land.

In this case, I think an explicit `height` is better and more bulletproof. Say you have a super small font-size and things get nested with `em` units—ultimately the progress bar could disappear or just be unusable.

mdo:
Closing per convo above.

PR #30387, opened 2020-03-12
Add sm and lg sizes for progress

Masashiotake696:
Issue #30318


mdo:
See https://github.com/twbs/bootstrap/issues/30318#issuecomment-598542842. Don't think this really is necessary if we have customization with the `height`.

MartijnCuppens:
I agree

PR #30388, opened 2020-03-12
enable button toggle on label when checkbox is inside

### LINKED ISSUE #30077: $().button('toggle')	not working in 4.4.1
$().button('toggle')

not working in 4.4.1, but working fine in 4.3.1

            <div class="btn-group-toggle" data-toggle="buttons">
                <label id="my-switch" class="btn btn-lg btn-light font-weight-bold ml-2 text-nowrap">
                <input type="checkbox" autocomplete="off" value="1"><span class="check">✓</span> <i class="far fa-clipboard"></i> <span class="d-none d-lg-inline">My</span>
                </label>
            </div>

    $("#my-switch").button('toggle');

4.4.1 https://codepen.io/DK568/pen/MWYxWMO

4.3.1 https://codepen.io/DK568/pen/ExaMaxd

the difference is that button is not checked by jquery

Lausselloic:
fix #30077 for V4 $().button('toggle') not working for checkbox inside label
I' tried with mouse, keyboard and jquery call on label and also click on input
add unit test to avoid further errors

Need to check on V5 without Jquery button plugin have the previous bug (checked state not change)

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -84,17 +84,13 @@ class Button {
               $(activeElement).removeClass(ClassName.ACTIVE)
             }
           }
-        } else if (input.type === 'checkbox') {
-          if (this._element.tagName === 'LABEL' && input.checked === this._element.classList.contains(ClassName.ACTIVE)) {
-            triggerChangeEvent = false
-          }
-        } else {
-          // if it's not a radio button or checkbox don't add a pointless/invalid checked property to the input
-          triggerChangeEvent = false
         }
 
         if (triggerChangeEvent) {
-          input.checked = !this._element.classList.contains(ClassName.ACTIVE)
+          // if it's not a radio button or checkbox don't add a pointless/invalid checked property to the input
+          if (input.type === 'checkbox' || input.type === 'radio') {
+            input.checked = !this._element.classList.contains(ClassName.ACTIVE)
+          }
           $(input).trigger('change')
         }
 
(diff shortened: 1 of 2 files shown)


Lausselloic:
@patrickhlauke is there a version of your code on the v5? or do I need to make a dedicated PR with your fix and this one?

XhmikosR:
I don't think we landed the PR that introduced the issue in master, IIRC

patrickhlauke (approved):
Tested in Chrome, Firefox, IE11 with mouse/keyboard/triggering the toggle method programmatically. indeed, this seems to work well.

XhmikosR:
@Johann-S LGTY?

Johann-S (approved):

mdo:
Possible regression in #30849?

XhmikosR:
Seems like it. @Lausselloic can you have a look please and CC me? This will probably need to PRs one if the issue is present in master too.

XhmikosR:
@Lausselloic friendly ping :)

Lausselloic:
Sorry I've miss the ping. Yes it's due to that PR. I will give a try to fix it.

Lausselloic:
@XhmikosR I've update the solution. Instead of stopping the event propagation if event is fired on checkbox inside the label, I detect if the event is not on this element before firing the toggle.

Tests are okay for I'll let js team check that don't break other stuff but looks better for me

PR #30389, opened 2020-03-12
v5: Underline links

### LINKED ISSUE #15304: Links (as part of regular text) not distinguishable enough

mdo:
This PR changes the `$link-text-decoration` from `none` to `underline` to help improve link discovery. It also includes a fix to the docs sidebar to override that for our desired look. I've kind of missed underlines on our links throughout the docs at least, so this feels like a win-win to me.

With the work done in #30262, this also doesn't affect any existing Bootstrap component. All the components you'd expect will nullify the underline thanks to the changes from that PR.

Fixes #15304.

---

Let me know your thoughts @patrickhlauke and @twbs/css-review.

Preview: <https://deploy-preview-30389--twbs-bootstrap.netlify.com/>


scss/_variables.scss
```diff
@@ -249,9 +249,9 @@ $body-text-align:           null !default;
 // Style anchor elements.
 
 $link-color:                              $primary !default;
-$link-decoration:                         none !default;
+$link-decoration:                         underline !default;
 $link-hover-color:                        darken($link-color, 15%) !default;
-$link-hover-decoration:                   underline !default;
+$link-hover-decoration:                   $link-decoration !default;
```
MartijnCuppens: We could `null`ify this. This way we could also remove the hover text-decorations from buttons.

site/assets/scss/_sidebar.scss
```diff
@@ -24,6 +24,7 @@
   padding: .25rem .625rem .25rem .5rem;
   font-weight: 600;
   color: rgba($black, .65);
+  text-decoration: none;
```
MartijnCuppens: By adding this, line 35 becomes redundant. Same for line 74.

MartijnCuppens:
The hover text-decorations covered in https://github.com/twbs/bootstrap/pull/30262/files can now be `null`ified if `$link-hover-decoration == null`

XhmikosR:
Should we remove the effect for ToC links?

Generally I can't tell if I like it or not, I guess it takes some time to evaluate this since I was used to no underline.

MartijnCuppens:
> Should we remove the effect for ToC links?

They are going to rethemed anyway: https://github.com/twbs/bootstrap/pull/30354

> Generally I can't tell if I like it or not, I guess it takes some time to evaluate this since I was used to no underline.

You'll get used to it. It's way clearer for visually impared people.

patrickhlauke:
from a cursory glance on the netlify preview...i like it (but yes should be rethemed/suppressed for the TOCs)

ysds:
I agree with @XhmikosR, but anyway, we have time before the official v5 release, so let's get feedback from users.

mdo:
Updated this the other day so it's ready to go I think.

XhmikosR:
I think we should wait a bit on this. For example, I see `text-muted` has no visual difference when focusing or hovering. Not sure if it's intentional. Also, there might be more such cases.

MartijnCuppens:
I've pushed some changes I mentioned in https://github.com/twbs/bootstrap/pull/30389#issuecomment-598388393.

> I see text-muted has no visual difference when focusing or hovering.

`text-muted` shouldn't have a hover state, the [colored links](http://localhost:9001/docs/4.3/helpers/colored-links/) should be used for this.

We might need to revisit the color change of hovered links later, since they are always darkened. See
https://github.com/twbs/bootstrap/blob/81223ee0412cc937cde108242e93b4407f6cede5/scss/helpers/_colored-links.scss#L8

MartijnCuppens (approved):

XhmikosR:
> `text-muted` shouldn't have a hover state, the [colored links](http://localhost:9001/docs/4.3/helpers/colored-links/) should be used for this.

We need to fix this in our docs then. I find it wrong that in the homepage, where we use text-muted, there's no focus/hover style.

patrickhlauke:
>  I find it wrong that in the homepage, where we use text-muted, there's no focus/hover style.

for focus, we at least have the default outline though, just to clarify.

patrickhlauke (approved):
not checked all possible variations/edge cases, but this looks good to me from looking over the deploy preview

XhmikosR:
>for focus, we at least have the default outline though, just to clarify.

True, TBH I didn't even notice it :P

Anyway, I'm not blocking the patch :)

MartijnCuppens:
I'll open a new PR to tackle the hover color of the link on the homepage, we can merge this if all checks succeed

patrickhlauke:
> I'll open a new PR to tackle the hover color of the link on the homepage

be good to check that various situations (are there any more, beyond `.text-muted`?) are also covered in general

ysds (approved):

PR #30390, opened 2020-03-12
Use `display: block;` for form elements

MartijnCuppens:
`.form-select` and `.form-file` have `display: inline-block` applied to. This results in an inconsistent behaviour in `col-auto` because the label length influences the col width, even when the label is smaller:
https://codepen.io/MartijnCuppens/pen/wvamMpQ

`.form-file` is a div, so several properties can be removed. They were added in the past because it used to be a label.

Not sure if we should backport this, could be a BC.

### DIFF
--- a/scss/forms/_form-file.scss
+++ b/scss/forms/_form-file.scss
@@ -1,9 +1,6 @@
 .form-file {
   position: relative;
-  display: inline-block;
-  width: 100%;
   height: $form-file-height;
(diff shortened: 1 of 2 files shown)


mdo (approved):

MartijnCuppens:
Not gonna backport this to v4, it would break implementations where the form elements are used inline.

PR #30391, opened 2020-03-14
transition() mixin prevent redundant 'transition: none' and pass `null`

ysds:
This PR includes three points:

**1.** `transition(null)`  and `transition(none)` output the following redundant CSS, but should be prevented.

```scss
  selector {
    transition: none;
  }
}
````

**2.**  Fix the problem where libsass(node-sass) can't handle `transition(null)` well:

```scss
selector {
  @include transition(null);
}
// output
selector {
  transition:;
}

```

**3.**  The transition mixin now show warning message  when use `none` or `null` with multiple arguments.

Before: https://www.sassmeister.com/gist/b4d2d3ce8601116308e72df97a0d309a
After: https://www.sassmeister.com/gist/7ddae4ab995e0b41bbb93abe6eff616f


scss/mixins/_transition.scss
```diff
@@ -1,13 +1,21 @@
...
-  @if $enable-transitions {
-    @if length($transition) == 0 {
-      transition: $transition-base;
-    } @else {
-      transition: $transition;
+  @if length($transition) == 0 {
+    $transition: $transition-base;
+  }
+
+  @if length($transition) > 1 {
```
MartijnCuppens: This could be a `@else if`?

ysds: It is fine in most cases, but `@if` can catch cases where invalid values ​​like `$transition-base: null, all 1s`.

ysds:
I found a problem, please don't review yet.

ysds:
OK, ready for review.

MartijnCuppens (approved):

XhmikosR:
Any reason we don't land this in v5?

PR #30394, opened 2020-03-14
`box-shadow()` mixin: allow 'null' and drop support 'none' with multiple args

ysds:
The `box-shadow()` mixin has been changed to handle `none` with multiple arguments in #27972 and #28004, but it should be `null` instead of `none`.

This PR make the `box-shadow()` mixin allow `null` and drop support `none` with multiple args. And add warning message when used `none` with multiple arguments.

before: https://www.sassmeister.com/gist/bdf41c41937a614861a026169e39279e
after: https://www.sassmeister.com/gist/56fdeffc72831664e546baf22f365cdf


MartijnCuppens (dismissed):

mdo:
In cases of `none, none`, should that be a single `none`?

MartijnCuppens:
> In cases of `none, none`, should that be a single `none`?

In cases of `none, none`, you're doing something wrong. Not sure if we should really try to tackle all possible errors, unless it's something which can occur by configuring the Sass variables (which is not the case)

ysds:
I agree with @MartijnCuppens. I think it should output incorrect value as it is, not implicitly fixing it.

site/content/docs/4.3/migration.md
```diff
@@ -42,6 +42,7 @@ Changes to our source Sass files and compiled CSS.
 - The `button-variant()` mixin now accepts 3 optional color parameters, for each button state, to override the color provided by `color-yiq()`. By default, these parameters will find which color provides more contrast against the button state's background color with `color-yiq()`.
 - The `button-outline-variant()` mixin now accepts an additional argument, `$active-color`, for setting the button's active state text color. By default, this parameter will find which color provides more contrast against the button's active background color with `color-yiq()`.
 - Ditch the Sass map merges, which makes it easier to remove redundant values. Keep in mind you now have to define all values in the Sass maps like `$theme-colors`. Check out how to deal with Sass maps on the [theming documentation]({{< docsref "/getting-started/theming#maps-and-loops" >}}).
+- The `box-shadow()` mixin now output as it is whatever when passed 'none' with multiple arguments. [See #30394](https://github.com/twbs/bootstrap/pull/30394).
```
XhmikosR: This might need some wording fixes. /CC @mdo

wojtask9:
I added this `none` checks because I was trying to fix https://github.com/twbs/bootstrap/issues/26478

Looking at code
https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_buttons.scss#L55
it supports setting variable `$btn-active-box-shadow`  to `none`.
But then this code 
https://github.com/twbs/bootstrap/blob/v4-dev/scss/_buttons.scss#L48
produces invalid css rule (for example `box-shadow: none, 1px 1px 1px black`).

MartijnCuppens:
I think I'm gonna change sides here, @ysds. It's actually quite easy to autofix the situations with multiple values and `null`:

https://www.sassmeister.com/gist/49f9c4e91db45677f45dfd2e8d5cf397

MartijnCuppens:
ping @ysds 

I think we should switch to the example I suggested, it will auto-fix the issue @wojtask9 is talking about. We can also leave out the `$btn-active-box-shadow != none` checks in other places.

ysds:
@MartijnCuppens Your sassmeister doesn't work in libsass.

I think the box-shadow mixin should not do autofix. 

```
box-shadow(none, 0 0 1px # 000);
box-shadow(0 0 1px # 000, none);
```

Add or remove shadow, not 100% sure which one the user intends. It should be outputted incorrect value as it is, not implicitly fixing it. So I will change the `mixins/_buttons.scss`.

ysds:
And also, `$btn-active-box-shadow: null` is absolutely correct than `$btn-active-box-shadow: none`.

MartijnCuppens:
> @MartijnCuppens Your sassmeister doesn't work in libsass(node-sass).

Seems to work fine with the Bootstrap setup which is node-sass based

> Add or remove shadow, not 100% sure which one the user intends.

The issue with `$btn-active-box-shadow: none` shows us we can run in to problems with the current config. We can of course fix this by adding an additional if test, but I'm not sure if it's a good idea to move that logic outside of the mixin. We now have to think about this every time we use the mixin, while it can easily be fixed for us.

> Add or remove shadow, not 100% sure which one the user intends.

In our case, we probably get into this situation when people try to remove one of the shadows that would be set, so in both cases `box-shadow: 0 0 1px #000;` is probably the result they want.

ysds:
> Seems to work fine with the Bootstrap setup which is node-sass based

See the none1 and 4 cases.

> In our case, we probably get into this situation when people try to remove one of the shadows that would be set,

Why not use $btn-active-box-shadow: null ?

ysds:
> In our case, we probably get into this situation when people try to remove one of the shadows that would be set,

That just in our case. The mixin should expect any cases.

MartijnCuppens:
> Seems to work fine with the Bootstrap setup which is node-sass based
>
> See the null1 and 4 cases.

I did something wrong, code doesn't work in node-sass indeed. 

> Why not use $btn-active-box-shadow: null ?

That doesn't work, you can't unset variables with `null`, they will be overridden by the Bootstrap variables (which is a pity).

ysds:
> That doesn't work, you can't unset variables with null,


I was forgotting that... I’ll reconsider this.

mdo (approved):

PR #30395, opened 2020-03-15
fix(css): adds a new opacity var

zalog:
This PR fixes beeing able to control `.is-valid` & `.is-invalid` input:focus opacity and removes hardcoded .25.

If it's ok, I can propagate this also on the other elements in `mixins/_forms.scss`.

Thanks for your time.


MartijnCuppens (approved):

zalog:
I made a rebase.

MartijnCuppens:
Thanks!

PR #30398, opened 2020-03-15
Removes `.text-reset` references

zalog:
This PR removes `.text-reset` references from docs and adds a line in migration. 

If we'll remove this class, please give me some feedback of what we need to write in migration doc.

[deploy-preview](https://deploy-preview-30398--twbs-bootstrap.netlify.com/docs/4.3/utilities/text/#monospace)

Thanks for your time.


zalog:
Oh crap, [I found](https://github.com/twbs/bootstrap/blob/master/scss/_utilities.scss#L393) that we still use this :)
Please ignore.

PR #30401, opened 2020-03-16
variables: adds empty `$utilities`

zalog:
This PR expose the possibility of adding new utility classes.

If it's ok, please give me some feedback of what comment description we should leave there.

@MartijnCuppens what do you think?

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -1,6 +1,5 @@
 // Utilities
 
-$utilities: () !default;
 // stylelint-disable-next-line scss/dollar-variable-default
 $utilities: map-merge(
   (
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -199,6 +199,12 @@ $escaped-characters: (
   (")","%29"),
 ) !default;
 
+
+// Utilities
+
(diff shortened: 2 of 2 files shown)


XhmikosR:
The branch is not based on the latest master branch @zalog :)

zalog:
Wow! My bad :)
Now it's ok.

MartijnCuppens:
Not sure about this, I think it's clearer to leave it in the separate file.

zalog:
Yeah, but I think exposing this on our most visible scss file, and leave there a comment, can open developers eyes. What do you think?

MartijnCuppens:
Hmm, not sure. It's weird to only have an empty variable. We have a documentation page dedicated to the Utility API (https://twbs-bootstrap.netlify.com/docs/4.3/utilities/api/) and I think this covers it well enough.

zalog:
Ok. Please think about this if we will need some sort of specific order in variables regarding this.

Tnks for your time.

MartijnCuppens:
We still have https://github.com/twbs/bootstrap/pull/29451 open, feel free to drop ideas there

PR #30407, opened 2020-03-17
Change Versions to the latest Bootstrap since now

zandiarash:
I've changed versions to the latest Bootstrap till now.
Also this 

> `site/assets/js/src/search.js`

was showing instead of this

> `site/static/docs/4.4/assets/js/src/search.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,14 +41,14 @@
 
 Several quick start options are available:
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Thanks, but this is taken care of in #29925

PR #30411, opened 2020-03-17
terser: use 2 passes instead of the default 1

XhmikosR:
Apparently this yields a ~~non-trivial size~~ decrease (gzip):

| File | Before | After | Diff |
| --- | -------- | ----- | ---- |
| bootstrap.bundle.min.js | 22.38 kB | 22.3 kB | -0.08 kB (-0.36 %) |
| bootstrap.min.js | 15.86 kB | 15.83 kB | -0.03 kB (-0.19 %) |

~I haven't tried passes > 2 yet, might be worth trying 3 or 5 too~. I tried passes=5 and it results in the same savings as 3. Also, if we conclude this is fine, we might as well backport this to v4-dev.

Preview: <https://deploy-preview-30411--twbs-bootstrap.netlify.com/>


Johann-S (approved):

XhmikosR:
Alright, I updated the results now after #30358. It seems the gain is now pretty small.

XhmikosR:
I think I'm going to drop this patch. It doesn't gain us a lot.

PR #30412, opened 2020-03-17
Clarify card group behavior in v4

### LINKED ISSUE #30366: Card-group class looses its display:flex; when screen < 576px
Hi,

After upgrading from `4.3.1` to `4.4.1` I've noticed that `.card-group` class was loosing it attribute `display: flex;` when resizing screen under the `576px`.

Here is the CSS class I get from Dev Tools on the documentation's bootstrap page  (https://getbootstrap.com/docs/4.4/components/card/#card-groups):

```css
@media (min-width: 576px) {
    .card-group {
        display:-ms-flexbox;
        display: flex;
        -ms-flex-flow: row wrap;
        flex-flow: row wrap
    }
   ...
}
```

In the documentation, it is said 
> Card groups use display: flex;
This may be confusing.

Operating system: macOs 10.15.3 (19D76)
Browsers: 
- Safari Version 13.0.5 (15608.5.11)
- Firefox 73.0.1 (64 bits)
- Brave Version 1.4.96 Chromium: 80.0.3987.132 (Official Build) (64-bit)

mdo:
Closes #30366.

### DIFF
--- a/site/docs/4.4/components/card.md
+++ b/site/docs/4.4/components/card.md
@@ -479,7 +479,7 @@ In addition to styling the content within cards, Bootstrap includes a few option
 
 ### Card groups
 
-Use card groups to render cards as a single, attached element with equal width and height columns. Card groups use `display: flex;` to achieve their uniform sizing.
+Use card groups to render cards as a single, attached element with equal width and height columns. Card groups start off stacked and use `display: flex; ` to become attached with uniform dimensions starting at the `sm` breakpoint.
 
 {% capture example %}
 <div class="card-group">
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

XhmikosR:
The test failures are unrelated to this PR; I have seen them in the past, I even tried using Ruby 2.6 but I still got similar errors.

I'll check tomorrow if there are any gem updates and make a v4-dev PR to update gems and deps.

MartijnCuppens (approved):
LGTM but we also need to add this in master

XhmikosR:
>LGTM but we also need to add this in master

If so, we should first target master and cherry pick the patch to v4-dev later.

XhmikosR:
I made #30416 which I plan to cherry pick later in v4-dev and close this PR.

PR #30414, opened 2020-03-17
V4 dep updates and backports

XhmikosR:
Preview: <https://deploy-preview-30414--twbs-bootstrap.netlify.com/>


Johann-S (approved):

XhmikosR:
@ffoodd @MartijnCuppens @patrickhlauke the #30073 backport is nice for v4-dev, but since the navbar occupies the whole width, the skip links overlap with it. Just pointing it out. I personally don't really like this :/

ffoodd:
I agree, it's more awkward in `v4-dev`.

I see two workarounds (that could match to `master` as well):

1. use a colored backplate and make it a sort of sticky-top element (just like Github's one!)
2. keep the previous pushing design, just left-aligning skip-links (being at the top left is the main goal of this PR) — may consider to make both skip-links visible at once on docs, in that case. This could look weird in v5 since the whole layout is centered.

Depending on what we consider to do, I'll open a PR against `master` or `v4-dev`.

Any thoughts?

XhmikosR:
@mdo @MartijnCuppens thoughts how to proceed with the skip links in this branch?

XhmikosR:
I removed the docs skip links patch for now and pushed it to https://github.com/twbs/bootstrap/tree/v4-dev-docs-skip-links

@MartijnCuppens mentioned some issues with master too so we better discuss this more and/or revert the patch from master in the meantime.

MartijnCuppens (approved):

PR #30415, opened 2020-03-18
Drop `@support (position: sticky)` media query

MartijnCuppens:


XhmikosR:
This should be reviewed thoroughly and if landed, it should be after #30377

ffoodd:
Just to be sure, dropping IE means we don't care about graceful degradation?

In the very specific case of `.position-sticky`, offsets with a `relative` fallback could lead to overlapping content in unsupporting browsers.

Is this tolerable?

XhmikosR:
>Just to be sure, dropping IE means we don't care about graceful degradation?

That's my concern too about this patch. But graceful degradation is broken for IE for sure (or rather it will be after #30377). For the rest of the browsers is something we should consider before landing a patch like this.

ffoodd:
FWIW if we don't care about IE, [other browsers should handle this correctly](https://caniuse.com/#search=sticky) – since most of unsupporting browsers are feature phone's ones.

I'm for dropping the feature query, however if we can handle a decent fallback with a single property it may worth it.

XhmikosR:
AFAICT this just saves us 0.02 kB from the minified file after compression.

So, we should really think about it before landing this for sure.

ysds (approved):

mdo:
I'm okay leaving it around if we think it'll avoid something potentially problematic for folks, but otherwise, let's roll with this being a more forward-focused iteration and drop it.

MartijnCuppens:
> ... let's roll with this being a more forward-focused iteration and drop it.

We should clean up legacy stuff as much as possible. Apart from IE, the browsers we support should be fine.

MartijnCuppens:
We already use custom properties and the difference between browsers that support  custom properties and not position sticky is negligible:

- https://caniuse.com/#feat=css-sticky
- https://caniuse.com/#feat=css-variables

PR #30425, opened 2020-03-20
Regenerate package-lock.json to fix npm vulnerabilities

XhmikosR:
3 left, 2 low and 1 high

I'm going to keep this open for a while in case we have further vulnerability fixes soon. I don't expect the high one to be fixed soon, but one can only hope :)


XhmikosR:
NVM, we need to fix and land #30447 first...

Johann-S (approved):

PR #30426, opened 2020-03-20
Make form-check label customizable

zalog:
This PR adds a new var for `$form-check-label`.

What do you think?

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -612,7 +612,7 @@ $form-check-input-width:                  1.25em !default;
 $form-check-min-height:                   $font-size-base * $line-height-base !default;
 $form-check-padding-left:                 $form-check-input-width + .5em !default;
 $form-check-margin-bottom:                .125rem !default;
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -612,6 +612,7 @@ $form-check-input-width:                  1.25em !default;
 $form-check-min-height:                   $font-size-base * $line-height-base !default;
 $form-check-padding-left:                 $form-check-input-width + .5em !default;
 $form-check-margin-bottom:                .125rem !default;
+$form-check-label-color:                  $gray-700 !default;
```
mdo: This should be `null` to start I'd think, since we don't style the `<label>`s to be another color—this gets inherited from the `<body>`.

zalog: Yep, you're right. `inherit` it is.

MartijnCuppens (approved):
Changed it to `null` @mdo.
Thanks for the suggestion @zalog!

PR #30427, opened 2020-03-20
Use a link helper instead of the text color utility

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/pull/30389#issuecomment-601569264

Preview: https://deploy-preview-30427--twbs-bootstrap.netlify.com/


XhmikosR:
Can you check if we do this in other places too?

MartijnCuppens:
> Can you check if we do this in other places too?

I've adjusted these too in the meantime.

XhmikosR (approved):

PR #30433, opened 2020-03-22
Fix display css for hiding element only on xs screen 

mohsin-mehmood:
Fixed the css class for hiding element only on xs screen.


XhmikosR:
There's no `.d-xs-none`, see https://getbootstrap.com/docs/4.4/utilities/display/#notation

PR #30434, opened 2020-03-22
docs a11y fixes

patrickhlauke:
some more fixes for the simpler pa11y errors referenced in recent https://github.com/twbs/bootstrap/pull/29315 discussion

note that some other outstanding errors (and even some of the tweaks here) are not necessarily hard failures (e.g. the `scope` attribute on `<th>` is not really mandatory/necessary, and the need for a submit button when the only form control present is a single text/search input is debatable) ... seems pa11y is a bit overzealous with its errors when they're not necessarily failures

### DIFF
--- a/site/content/docs/4.3/examples/navbars/index.html
+++ b/site/content/docs/4.3/examples/navbars/index.html
@@ -56,7 +56,7 @@ extra_css:
         </li>
       </ul>
       <form>
-        <input class="form-control" type="text" placeholder="Search">
+        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
       </form>
     </div>
   </div>
@@ -90,7 +90,7 @@ extra_css:
         </li>
       </ul>
       <form>
-        <input class="form-control" type="text" placeholder="Search">
+        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
       </form>
     </div>
   </div>
@@ -124,7 +124,7 @@ extra_css:
         </li>
       </ul>
       <form>
-        <input class="form-control" type="text" placeholder="Search">
+        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
       </form>
     </div>
   </div>
@@ -158,7 +158,7 @@ extra_css:
         </li>
       </ul>
       <form>
-        <input class="form-control" type="text" placeholder="Search">
+        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
       </form>
     </div>
   </div>
@@ -192,7 +192,7 @@ extra_css:
         </li>
       </ul>
       <form>
-        <input class="form-control" type="text" placeholder="Search">
+        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
       </form>
     </div>
   </div>
--- a/site/content/docs/4.3/forms/form-control.md
+++ b/site/content/docs/4.3/forms/form-control.md
@@ -24,9 +24,9 @@ toc: true
 Set heights using classes like `.form-control-lg` and `.form-control-sm`.
 
 {{< example >}}
-<input class="form-control form-control-lg" type="text" placeholder=".form-control-lg">
-<input class="form-control" type="text" placeholder="Default input">
-<input class="form-control form-control-sm" type="text" placeholder=".form-control-sm">
+<input class="form-control form-control-lg" type="text" placeholder=".form-control-lg" aria-label=".form-control-lg example">
+<input class="form-control" type="text" placeholder="Default input" aria-label="deafult input example">
+<input class="form-control form-control-sm" type="text" placeholder=".form-control-sm" aria-label=".form-control-sm example">
 {{< /example >}}
 
 ## Readonly
@@ -34,7 +34,7 @@ Set heights using classes like `.form-control-lg` and `.form-control-sm`.
 Add the `readonly` boolean attribute on an input to prevent modification of the input's value. Read-only inputs appear lighter (just like disabled inputs), but retain the standard cursor.
 
 {{< example >}}
-<input class="form-control" type="text" placeholder="Readonly input here..." readonly>
+<input class="form-control" type="text" placeholder="Readonly input here..." aria-label="readonly input example" readonly>
 {{< /example >}}
 
 ## Readonly plain text
--- a/site/content/docs/4.3/forms/form-control.md
+++ b/site/content/docs/4.3/forms/form-control.md
@@ -9,16 +9,14 @@ toc: true
 ## Example
 
 {{< example >}}
-<form>
-  <div class="mb-3">
-    <label for="exampleFormControlInput1">Email address</label>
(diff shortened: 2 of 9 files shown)


patrickhlauke:
it's very likely that i targeted the wrong branch etc for this...it's been a while since i ventured into this, and my whole setup may be a bit borked...if that's the case, sorry...

XhmikosR:
Thanks @patrickhlauke 🙂

The branch is for the v5 docs so we are OK for starters; if we want to backport the fixes we'll need to do it later.

That being said, it depends how/when you want to land the fixes. If you want to land this regardless of the pa11-ci branch, then I'll need to manually cherry pick the patches to see what's left/fixed. Otherwise we can merge this PR into that branch and you push to that branch later.

XhmikosR:
Oh, also, we can always ignore some errors/warnings as we deem it right.

patrickhlauke:
can we explicitly set some of the tests not to be run? thinking that the opinionated ones from pa11y (which aren't strict failures of WCAG, but more their opinion) like the lack of submit button in forms would be a good candidate. and the color contrast warnings for the theming page would be good to suppress.

otherwise, if we just ignore them in terms of "they throw errors, but we just let the test fail anyway", then of course we'll miss cases in future when other things break.

also, happy to do a separate manual PR for v4 docs. would that target v4-dev ?

XhmikosR:
>can we explicitly set some of the tests not to be run? thinking that the opinionated ones from pa11y (which aren't strict failures of WCAG, but more their opinion) like the lack of submit button in forms would be a good candidate. and the color contrast warnings for the theming page would be good to suppress.

pa11-ci is using pa11 which is using HTML_CodeSniffer by default. We can ignore pages, selectors and rules. I already ignore some like our docs search form missing button.

>also, happy to do a separate manual PR for v4 docs. would that target v4-dev ?

Let's wait about that, until we decide how to proceed with this PR first. Generally, I always prefer to try and cherry pick patches instead of manually backporting them. We'll see how easy that would be in this case.

So, first let me know if you want to land these fixes without the pa11-ci PR. In the meantime, I merged this branch into the pa11-ci branch so you can see the remaining issues

patrickhlauke:
i'm not able to find those two instances of this error

"Error: Duplicate id attribute value "constructor" found on the web page"

when i go to the deploy preview version of the site, to those docs pages, and view source/inspect, i can only find one single instance of `#constructor`. am i looking in the wrong places?

will also be diving into the "`<fieldset>` without `<legend>`" thing. suspect it's a false positive/overzealous rule again as well.

XhmikosR:
>i'm not able to find those two instances of this error
>"Error: Duplicate id attribute value "constructor" found on the web page"

Yeah, I saw those too, and they are false positives AFAICT.

Let me know after you are done how you want me to proceed with ignoring some stuff and merging this to which branch. I think I prefer landing this branch in master separately from the pa11-ci PR for starters.

ffoodd:
FWIW in Boosted I chose to run aXe instead of HTML_CodeSniffer, and to restrict to errors. Results might be more accurate?

patrickhlauke:
ok, confirmed with colleague @stevefaulkner that `<fieldset>` without `<legend>` not a hard failure. it's pointless (as treated exactly like a `<div>` by assistive tech), but not an outright failure per se (unless the grouped controls DO need an overarching label/legend to make sense) https://codepen.io/stevef/post/fielsdet-and-legend-semantics

patrickhlauke:
right, fixed one instance where the `<fieldset>` would benefit from an accessible name. the other error about "Error: Fieldset does not contain a legend element. All fieldsets should contain a legend element that describes a description of the field group." is, in my mind, opinionated (what counts is the fieldset's accessible name, not whether or not the legend element is used...and in the case of forms/layout it's a known problem we have with not being able to have `<legend>` as the immediate first child of the `<fieldset>`, but this in practice seems not to cause accname issues...so i'd suppress this type of error as long as the other one about fieldset accessible names is kept)

patrickhlauke:
axe is definitely less opinionated (though not perfect) compared to codesniffer

XhmikosR:
I don't mind switching engines, I just went with the default.

So,I suggest we review and merge this, then I rebase the other branch and
switch engines and see where we are at.

patrickhlauke:
> So,I suggest we review and merge this, then I rebase the other branch and
switch engines and see where we are at.

sounds good to me

patrickhlauke:
btw while working on the "adding submit buttons where appropriate" part, i noticed a major flaw with validation messages for file picker...but i'll file a separate issue about that

XhmikosR:
Cleaned up the patches, LGTM AFAICT.

XhmikosR (approved):

mdo:
Are these `<form>` elements just unnecessary, or do they cause some kind of issue? I think I'd prefer to see the `<form>` elements in there as an explicit signal to folks. Plus it _looks_ more semantic than seeing `<div>`s, which I suspect folks may call out in the future.

patrickhlauke:
> Are these <form> elements just unnecessary, or do they cause some kind of issue?

they do cause issues (mainly of usability) as there's no explicit submission mechanism (and it would rely on submitting on enter in a text field). it's acceptable enough for a lone search field, but on a form with more that one input (and certainly once you mix in anything else that doesn't allow keyboard entry, like just a series of checkboxes or radio buttons) it gets really impossible to submit.

> Plus it looks more semantic

it does suggest though that what you're seeing is a *complete* form. Which is not what we should be suggesting. Either have a fully formed form (which we do have a few of, e.g. in the validation pages), or don't use a start and end `<form>` tag. examples are just focused fragments that show individual components. i'd argue with anybody who moans about them not being wrapped in a `<form>`

ffoodd (approved):

ffoodd:
Just wondering about `aria-label`: I usually prefer to visually hide a `<label>` (or `<legend>` for `<feldset>`) since it would be available without CSS. Should we care about this in Bootstrap?

XhmikosR:
Rebased again @patrickhlauke don't just merge after a rebase 😛

I'll merge without squashing after tests pass here and see what we can cherry pick in v4-dev, otherwise we'll need to manually backport these.

patrickhlauke:
sorry, i think my whole local setup here is a bit messed up, which led to the merge

patrickhlauke:
> Just wondering about `aria-label`: I usually prefer to visually hide a `<label>` (or `<legend>` for `<feldset>`) since it would be available without CSS. Should we care about this in Bootstrap?

i think we mention in many places that there are various ways of achieving things (like using `aria-label` etc). i think adding visually hidden labels etc here would be more confusing than helpful. in the end, authors need to also use their own judgement on how some of these examples should be actually used in their own development...and their priorities/personal tastes will differ (i mean, if we want to be really strict, every form control should have an always visible label to pass WCAG 2.1 3.3.2, but again that's an aspect we don't necessarily enforce here for examples)

PR #30435, opened 2020-03-23
Add `.caption-top` for tables

### LINKED ISSUE #29879: Add option for treating `<caption>` in table as a table title above the table
Before opening:

I read https://github.com/twbs/bootstrap/issues/28499 and would like to request an additional configuration instead of a reversion.

I think it's pretty common to want to have a table's caption come before the table, see [Example 1](https://www.w3.org/WAI/tutorials/tables/caption-summary/#identifying-a-table-using-a-caption) from Web Accessibility Tutorials' Caption & Summary. 

It'd be nice to be able to apply a class to a caption and have it override default behavior like:

```css
    padding-top: .75rem;
    padding-bottom: .75rem;
    text-align: left;
    caption-side: top;
    /** insert additional properties to mimick a, say, h2
}
```

mdo:
- Move caption-side to `<table>` element in Reboot
- Add class for `.caption-top` (other options aren't supported anywhere it seems?)

Fixes #29879.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -355,6 +355,7 @@ svg {
 // Prevent double borders
 
 table {
+  caption-side: bottom;
   border-collapse: collapse;
 }
 
@@ -363,7 +364,6 @@ caption {
   padding-bottom: $table-cell-padding;
   color: $table-caption-color;
   text-align: left;
-  caption-side: bottom;
 }
 
 // 1. Matches default `<td>` alignment by inheriting `text-align`.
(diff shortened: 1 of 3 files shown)


ffoodd:
Other options are either deprecated or unsupported (depending on `writing-mode` logical property, whic we don't use) so you're right. Have a look at [`caption-side` page on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/caption-side).

ffoodd (approved):

MartijnCuppens (approved):

Issue #27121, opened 2018-08-23
Floating label stay at empty input position browser preload input 


munzirstudio:
Hi

When browser preload input (saved username & password) at floating label input, it seem the placeholder still visible:
![image](https://user-images.githubusercontent.com/4507016/44517274-daa34200-a6f9-11e8-84e5-0afd6d0daf34.png)

only after the input is focused/clicked the label move to it top position:
![image](https://user-images.githubusercontent.com/4507016/44517378-18a06600-a6fa-11e8-8cc8-9e5338077949.png)

Thanks

mdo:
This has been open for quite awhile, and I'm not a daily Chrome user, so I apologize for that. Is this still an issue for folks? Does #30449 make a difference at all?

eduardogoncalves:
Hi @mdo, I fixed this on #31948

eduardogoncalves:
@mdo I think this can be closed.

PR #30449, opened 2020-03-25
v5: Floating labels

### LINKED ISSUE #21312: Feature Request: Floating Labels
I want to add a feature request! I would like to see floating labels be incorporated into bootstrap 4. Please take a look at this dribble shot: https://dribbble.com/shots/1254439--GIF-Float-Label-Form-Interaction which shows this in action as well as some pros for it:

> User keeps context
> The main advantage of this pattern is that the user keeps the field’s context after they’ve focused and entered a value. This provides for a more accessible, less frustrating experience.
> 
> Clean and scannable by default
> This pattern allows for a clean inline label experience by default, and only becomes a little more cluttered once the user has filled things out.
> 
> Elegant
> It needs to be said: this pattern is sexy. You usually can’t say that about forms. It looks good and the animations are a nice subtle touch.
> 

Google uses floating labels in their material design: https://material.google.com/components/text-fields.html#text-fields-floating-labels and designers love them: https://dribbble.com/search?q=floating+label 

Here is a live example http://mrlopez.me/bootstrap-floating-labels/ someone else did with bootstrap.

mdo:
Fixes #21312.

---

- [x] Document more things?
- [x] Confirm browser support
- [x] Look into #27121 (see https://github.com/twbs/bootstrap/pull/30449#issuecomment-704859889)

Preview: <https://deploy-preview-30449--twbs-bootstrap.netlify.com/docs/5.0/forms/floating-labels/>


scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,59 @@
...
+    border: $input-border-width solid transparent;
+    @include border-radius(.25rem);
+    @include transition(all .1s ease-in-out);
+  }
+
+  > .form-control {
+    &::-webkit-input-placeholder  { color: transparent; }
+    &::-ms-input-placeholder      { color: transparent; }
+    &::-moz-placeholder           { color: transparent; }
+    &::placeholder                { color: transparent; }
```
MartijnCuppens: I think autoprefixer already prefixes these?

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,59 @@
...
+
+    &:not(:placeholder-shown) {
+      padding-top: $form-floating-input-padding-t;
```
MartijnCuppens: We shouldn't animate paddings, but use transforms instead

scss/forms/_floating-labels.scss
XhmikosR: Yup and makes me wonder why Stylelint doesn't error out...

micka-fdz:
Good news! Is it possible to add support for `<textarea>` and `.form-file` elements?

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+//
+// Fallback for classic Edge
+//
+
+@supports (-ms-ime-align: auto) {
+  .form-floating > label {
+    display: none;
+  }
+  .form-floating > .form-control::-ms-input-placeholder {
+    color: #777;
```
XhmikosR: Is the fallback supposed to be like that here, intact from the example?

mdo: One of the todos I have listed is to check browser support, so I'll know more when I get at this on my PC :).

MartijnCuppens:
It feels a bit odd to me that the label doesn't shrink when The input field is focused, it looks like you're going to type in text before the password text:
image

I would prefer:
image

Thoughts?

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+    top: 0;
+    left: 0;
+    display: block;
+    width: 100%;
+    margin-bottom: 0; // Override default `<label>` margin
+    pointer-events: none;
+    cursor: text; // Match the input under the label
+    border: $input-border-width solid transparent;
+    @include border-radius(.25rem);
+    @include transition(all .1s ease-in-out);
```
MartijnCuppens: I would explicitly set `opacity` and `transform` here, like we do with other elements (eg. buttons)

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  > .form-select {
+    padding-top: $form-floating-input-padding-t;
+    padding-bottom: $form-floating-input-padding-b;
+  }
+
+  > .form-control:not(:placeholder-shown),
+  > .form-select {
+    ~ label {
+      @include font-size($form-floating-label-font-size);
```
MartijnCuppens: Transitioning the font-size isn't optimal. We better use scale in transform to fix this (we'll probably need to adjust the transform origin too)

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+      opacity: .65;
+      transform: translate3d(0, -.75rem, 0);
```
MartijnCuppens: Using translate3d triggers implicit compositing, which might result in blurry labels. We better use `translateY` here

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  > label {
+    position: absolute;
+    top: 0;
+    left: 0;
+    display: block;
+    width: 100%;
+    margin-bottom: 0; // Override default `<label>` margin
+    pointer-events: none;
+    cursor: text; // Match the input under the label
```
MartijnCuppens: This is moot since we have `pointer-events: none;`

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+    padding: $form-floating-padding-y $form-floating-padding-x;
+  }
+
+  > label {
+    position: absolute;
+    top: 0;
+    left: 0;
+    display: block;
+    width: 100%;
+    margin-bottom: 0; // Override default `<label>` margin
```
MartijnCuppens: This is moot since we use `position: absolute`

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+  > .form-select {
+    height: $form-floating-height;
+    padding: $form-floating-padding-y $form-floating-padding-x;
+  }
+
+  > label {
+    position: absolute;
+    top: 0;
+    left: 0;
+    display: block;
```
MartijnCuppens: This is moot since we use `position: absolute`

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+    width: 100%;
```
MartijnCuppens: Not sure why we need this, but I think we can leave this out too

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+    margin-bottom: 0; // Override default `<label>` margin
+    pointer-events: none;
+    cursor: text; // Match the input under the label
+    border: $input-border-width solid transparent;
```
MartijnCuppens: Why do we have this transparent border?

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,65 @@
...
+    @include border-radius(.25rem);
```
MartijnCuppens: There is no visible background or border, what's the point of this than?

scss/forms/_form-select.scss
```diff
@@ -6,7 +6,7 @@
 .form-select {
   display: block;
   width: 100%;
-  height: $form-select-height;
+  min-height: $form-select-height;
```
MartijnCuppens: We can't do this, because FF adds 2px to the height in a pseudo element we can not control: https://searchfox.org/mozilla-central/rev/c7e8bc4996f979e5876b33afae3de3b1ab4f3ae1/layout/style/res/forms.css#339

alecpl:
1. The "placeholder" color is wrong. Should be grey-ish, now it uses input text color. Is this something possible to fix?
2. I think I would prefer the way it looks as proposed in https://github.com/twbs/bootstrap/issues/21312#issuecomment-605595742

scss/forms/_floating-labels.scss
mdo: Added a code comment—it affects the box model and shifts the text.

mdo:
Pushed some changes based on the code review and fixing some conflicts. Also made the change from @MartijnCuppens's suggestion and moved the label on `:focus` now, too. I dig!

Added textarea support too, @micka-fdz. Unsure about file inputs... those ones are more complicated. Could be an add-on in the future once this gets settled.

mdo:
Pushed some more updates to address @MartijnCuppens's comments. Wondering if there's some more I can do to make this more bulletproof. Happy to hear out any comments!

ffoodd:
@mdo Regarding browser support and accessibility, I think my PR against our example (#30966) should help:
1. Better fallback for Legacy Edge, avoiding to hide labels,
2. more precise browser support doc.

patrickhlauke:
once you/we settled on the definitive version, i can have a quick look accessibility-wise if needed / if you ping me and point me to it.

mdo:
With the latest changes from @ffoodd's PR, plus some more tweaks, we've got a cleaned up Legacy Edge setup.

Screen Shot 2020-09-23 at 4 34 41 PM

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,70 @@
...
+
+@supports (-ms-ime-align: auto) {
+  .form-floating {
+    display: flex;
+    flex-direction: column-reverse;
+  }
+  .form-floating > label {
+    position: static;
+    padding: 0;
+    margin-bottom: calc(#{$form-floating-padding-y} / 2);
```
XhmikosR, line 63: Are we supposed to use `calc` here or our function?

scss/_variables.scss
```diff
@@ -640,7 +640,7 @@ $input-bg:                              $white !default;
 $input-disabled-bg:                     $gray-200 !default;
 $input-disabled-border-color:           null !default;
 
-$input-color:                           $gray-700 !default;
+$input-color:                           $body-color !default;
```
XhmikosR, line 643: Just to confirm that this change is intended here?

site/content/docs/5.0/forms/overview.md
```diff
@@ -18,6 +18,8 @@ sections:
     description: Replace browser default range inputs with our custom version.
   - title: Input group
     description: Attach labels and buttons to your inputs for increased semantic value.
+  - title: Floating labels
+    description: Create beautifully simple forms labels that float over your input fields.
```
XhmikosR, line 22: Should "forms" be "form" here and in the new page description?

patrickhlauke: Yes it should be singular

mmrath:
I was looking at the examples from deployed site. I see that when the input is empty, placeholder is not shown. It just shows the label. It would be nice if the placeholder text can be shown with the label always floating on top, something like the screenshot below.

image

patrickhlauke:
@mmrath mixing both placeholder and floating label makes no sense, unless i'm misunderstanding what you're after... by design, floating labels initially appear visually as if they were in the place of the form input/placeholder text, and then move out of the way when the user wants to enter something.

sounds more like you're after a style for labels that makes them appear small/as part of the input, statically?

mmrath:
@patrickhlauke Thanks for your reply. I could be wrong or misunderstanding the feature but I'll try to explain what I am thinking.

With the code below I was expecting that `name@example.com` would appear as a placeholder and `Email address` as label.  

```
<div class="form-floating mb-3">
  <input type="email" class="form-control" id="floatingInput" placeholder="name@example.com">
  <label for="floatingInput">Email address</label>
</div>
```

patrickhlauke:
floating labels are labels that initially appear like an entry, but when clicked move to the side/up/become smaller 

floating labels animation

(from https://medium.com/simple-human/floating-labels-are-a-bad-idea-82edb64220f6)

patrickhlauke:
but now looking at even the first example on https://deploy-preview-30449--twbs-bootstrap.netlify.app/docs/5.0/forms/floating-labels/ with the "Email address" field, do you mean you'd expect the placeholder to show once the user clicks into the input, the floating label itself moves out of the way, but the user hasn't entered anything yet? in which case yeah, looking just at the code there that does indeed include a `placeholder`, that would probably be expected behavior - not sure if we're explicitly suppressing this for a reason of not...

mmrath:
I did a bit more digging, I think I was expecting it to work like "Infield Top-Aligned Labels" as per the article below. So as per the article current "floating labels" are probably implemented correctly in this PR. 
As you said currently the placeholder is totally unused (except for `:placeholder-shown` usage). I personally think 'Infield Top-Aligned Labels" would have been better, but I am not a UX person.

https://uxmovement.com/forms/infield-top-aligned-labels-floating-labels/

screenshot from the website
image

MartijnCuppens:
#27121 seems to be fixed:
image

mmrath:
@MartijnCuppens #27121 appears to be related to value of the input field not placeholder.

HamedFathi:
Guys, Is it possible to have a smaller version of floating inputs? I think the final result is so big!

I think [this way](https://pryley.github.io/float-labels.js) was so flexible to use any size of inputs. (sm, lg, default)

```html
<input class="form-control form-control-lg" type="text">
<input class="form-control" type="text">
<input class="form-control form-control-sm">
```

image

scss/_variables.scss
mdo: Yup, decided there's no good reason to lighten the color or detach it from the body.

eduardogoncalves:
@mdo It's working now
image

mdo:
@eduardogoncalves Yup, pushed a new fix to override that last commit with separated rules. Should be good now.

mdo:
This should be good to go now! Can I get a final review?

alecpl:
I think I don't like these big inputs (and with strange padding when they are empty). Something like in https://github.com/twbs/bootstrap/pull/30449#issuecomment-710769866 would be better.

micka-fdz:
I agree, fields are currently much too big.
I prefer this render : https://getbootstrap.com/docs/4.5/examples/floating-labels/

MartijnCuppens:
https://github.com/twbs/bootstrap/pull/30449#issuecomment-710769866 would require an additional background to fix the border and shadow overlap

image

This will look weird if the parent element has another background
image

Or if the background of the input is semi-transparent:
image


But I can agree the input looks a little to big. https://getbootstrap.com/docs/4.5/examples/floating-labels/ might indeed be a better look as @micka-fdz suggests.

alecpl:
Semi-transparent input is a rare-case that could be just unsupported by floating labels. The other label background issue I could live with if the padding-top was smaller (and maybe with rounded corners). Still I would prefer this over what's proposed in the PR.

alecpl:
@mdo, in firefox 82 on linux the label animation has some issue. When you focus the input there are two steps of the animation visible, first the label moves up, then it's font size drops to even smaller. Maybe it's a firefox's transform/scale bug. Looks like when the text is moving it is blured then it sharpens. I don't see this in Chrome.

scss/forms/_floating-labels.scss
```diff
@@ -0,0 +1,85 @@
...
+  > label {
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: 100%; // allow textareas
+    padding: $form-floating-padding-y $form-floating-padding-x;
+    pointer-events: none;
+    border: $input-border-width solid transparent; // Required for aligning label's text with the input as it affects inner box model
+    transform-origin: 0 0;
+    @include transition(opacity .1s ease-in-out, transform .1s ease-in-out);
```
ffoodd, line 21: Shouldn't this value be variable, as we do with `$input-transition`?
Apart from that, seems fine!

ffoodd (approved):

scss/forms/_floating-labels.scss
ffoodd: Trying this on both Firefox & Chromium on Ubuntu, I can see a little font rendering jump (more obvious on Firefox). I guess it can be related to the font-stack I updated recently—already mentionned a related bug in #31955 by @MartijnCuppens.

That's unrelated to the PR I guess, I'd need more testing on those using our old font-stack and compare other possibilities… Opening an issue to track this.

micka-fdz:
Hi @MartijnCuppen, what about my suggestion of inputs height? I think it’s a quick modification. Maybe can I help?

PR #30455, opened 2020-03-25
v4: Button checked state for back/refresh

icio:
Unlike any of the other browsers, Firefox will restore \<input>s' checked status after a page refresh. (This behaviour can be suppressed with autocomplete=off on the form or input.) You can see this in action on 4.4.1 by selecting the [last example radio](https://getbootstrap.com/docs/4.4/components/buttons/#checkbox-and-radio-buttons) then refreshing.

Other browsers and Firefox will restore the \<inputs>s' checked status after navigating away from a page then navigating back. There's #25122 documenting this, which we can resolve by updating the initial button state on window `pageshow` as-well as `load`.

Trying to fix all of those things lead me to this PR, where button toggles with label/input rely on change events to update state, and `.btn[data-toggle=button]` remains handled by click events.

First PR here so feedback on testing/style/anything gratefully received.

---

I made the mistake of branching off v4-dev. Apologies :relaxed: Hopefully #28463 lands and v5 drops the requirement for JavaScript for this control, otherwise I can rebase onto master.

If a CSS-only alternative proves difficult, the toggle control could be far simpler if only label\>input based markup is supported because we can then toggle the active class through the change event alone.

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -39,11 +39,12 @@ const Selector = {
 }
 
 const Event = {
-  CHANGE_DATA_API     : `change${EVENT_KEY}${DATA_API_KEY}`,
-  CLICK_DATA_API      : `click${EVENT_KEY}${DATA_API_KEY}`,
-  FOCUS_BLUR_DATA_API : `focus${EVENT_KEY}${DATA_API_KEY} ` +
-                        `blur${EVENT_KEY}${DATA_API_KEY}`,
-  LOAD_DATA_API       : `load${EVENT_KEY}${DATA_API_KEY}`
+  CHANGE_DATA_API        : `change${EVENT_KEY}${DATA_API_KEY}`,
+  CLICK_DATA_API         : `click${EVENT_KEY}${DATA_API_KEY}`,
+  FOCUS_BLUR_DATA_API    : `focus${EVENT_KEY}${DATA_API_KEY} ` +
+                           `blur${EVENT_KEY}${DATA_API_KEY}`,
+  LOAD_PAGESHOW_DATA_API : `load${EVENT_KEY}${DATA_API_KEY} ` +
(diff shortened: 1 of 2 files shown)


js/src/button.js
```diff
@@ -118,10 +88,33 @@ class Button {
 
   // Static
 
+  static _update(e) {
```
icio: I haven't seen this style elsewhere in the bootstrap code, but there were several code-paths with similar logic, and in the process of trying to unpick and debug it, I ended up here.

patrickhlauke:
this seems to behave oddly in IE11 (which for v4 we're still supporting). for radio buttons, it seems to apply the highlighting to all other radio buttons that change, and basically not doing it in the correct order/way (you can end up with one radio button appearing unselected, the other two selected, but it's actually the opposite).

IE11 showing odd mismatch between highlighted radio button and actually selected one


i.e. also does the "remember which checkbox/radio was checked" thing when you refresh, and there the correct highlighting is not applied either.

on reload, checked state and visual state in IE11 go out of sync

patrickhlauke:
it does, however, behave correctly in Firefox on reload. and otherwise this approach seems to work quite smoothly/nicely (at least from my quick checks) for checkboxes/radios/toggle states on buttons

patrickhlauke (changes requested):
behaves erratically in IE11

(noting that even currently there's some oddity happening in IE11 where things can go out of sync for the radio buttons at least, it seems)

mdo:
Closing as a won't fix in v4, and we've since abandoned the JS for this in v5 for a CSS-only approach. Thanks though!

icio:
> we've since abandoned the JS for this in v5 for a CSS-only approach.

That's great news. Keep up the good work, team.

PR #30460, opened 2020-03-25
Add Release Drafter Action

mdo:
Pulls in the Release Drafter setup from the Icons project that's been working pretty well. Need to confirm how we'd do v4 and v5 here (perhaps just limiting by labels?).


.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,14 @@
+name: Release notes
+
+on:
+  push:
+    branches:
+      - master
```
XhmikosR, line 6: I suppose we don't want to add v4-dev here on purpose?

mdo: Think so. Makes me think we'd want to setup a separate workflow just for that branch?

mdo:
Going to merge to try this out and then tweak as needed once I see releases being drafted.

PR #30464, opened 2020-03-26
Missing submit buttons in "Navbar" examples

ffoodd:
Tackling a failing test in pa11y CI, in #29315 

Preview: <https://deploy-preview-30464--twbs-bootstrap.netlify.com/docs/4.3/examples/navbars/>

### DIFF
--- a/site/content/docs/4.3/examples/navbars/index.html
+++ b/site/content/docs/4.3/examples/navbars/index.html
@@ -32,8 +32,9 @@ extra_css:
           </div>
         </li>
       </ul>
-      <form>
-        <input class="form-control" type="text" placeholder="Search" aria-label="Search">
+      <form class="d-flex">
+        <input class="form-control mr-2" type="search" placeholder="Search" aria-label="Search">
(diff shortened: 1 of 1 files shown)


patrickhlauke:
it'd probably be slicker if we could have a graphical submit button using the magnifying glass icon or something? not sure how easy that'd be now that we split out bootstrap icons though, admittedly.

otherwise these may feel a bit clunky, visually. not bad, but probably not the slickest of examples.

(and i think for forms with just an input this is not necessarily a big problem, as even on a touch device, users would be on the on-screen keyboard and able to hit enter/submit on that, so not a biggie i think not having a submit button per se there....it's more pa11y flagging something as an error that is arguably more a recommendation/warning)

ffoodd:
@patrickhlauke Would you recommend trynig to ignore this specific test on this specific page?

If it's doable, it definetely wortyh a try, I guess.

patrickhlauke:
i would say ignore the whole test. it's opinionated, not directly an accessibility issue per se. i would exclude it from the pa11y run altogether.

PR #30466, opened 2020-03-26
Prevent nested tables style leaks

### LINKED ISSUE #24529: Inconsistant .table-active color

### LINKED ISSUE #25675: Table row-variants does not respect font-color

### LINKED ISSUE #27800: Nested table inherits classes that cannot be reset
The documentation states that "All table styles are inherited in Bootstrap 4", which already feels like a horrible decision for me, because it's completly unintuitive. And for worse it even passes styling to native <table> tags that do not reference any bootstrap classes.

```
<table class="table-bordered table-striped">
	<table>
		...
	</table>
</table>
```

The above example would cause the inner table to be both bordered and striped, just like the containing table. Using table-borderless on the inner table it is possible to get rid of the border but I haven't found a way to prevent the passing of the striped styling.

Is this intended behaviour? It seems really buggy and user-unfriendly to force styling upon child elements which don't use bootstrap classes without even giving a possibility to neutralize the inherited styling. (Maybe there is a possibility to overwrite stylings like table-striped, but I haven't found any classes like "table-nostripe" that would overwrite the striped effect).

### LINKED ISSUE #27879: .table-dark is used in two conflicting contexts

### LINKED ISSUE #28160: Suggestion for .table-hover

### LINKED ISSUE #29244: Table child elements (thead, tbody, tfoot, tr, td, th) should inherit parent background color

### LINKED ISSUE #29290: Contextual table classes apply tr class instead of td class to individual cells
Windows 10.0.17763 Build 17763
Chrome Version 76.0.3809.100 (Official Build) (64-bit)
JSBin of the problem: https://output.jsbin.com/zoqonukaru/
Using Bootstrap 4.3.1

When using Contextual Classes (such as table-warning, table-success) on Bootstrap tables, the class on the `tr` element applies to the entire row as expected, but placing an additional class on one of the `td` elements does not apply the new color - it continues to have the color from the `tr` element. I would expect the `td` element's class and color to take precedence over the parent `td` element's class and color.

I'm not sure where the fix would go to be properly compiled from a `.scss` file into the dist `.css` file, but adding `tr > td ~ .table-danger` to the dist `.css` file seems to make the cell-specific coloring work as intended. This fix doesn't apply to single cell rows, but you wouldn't need to color a single cell differently from the row color if your row is a single cell in the first place, so it's hopefully a moot point.

The fix (in compiled css file, so only a little useful):
`bootstrap/dist/css/bootstrap.css` line 1681
```
.table-danger,
.table-danger > th,
.table-danger > td {
  background-color: #f5c6cb;
}
```
to
```
.table-danger,
.table-danger > th,
.table-danger > td,
tr > td ~ .table-danger {
  background-color: #f5c6cb;
}
```

MartijnCuppens:
Prevent nested table style leaks by using children combinators and CSS custom properties:

- Change in reboot: inherit border color and set border style to solid to make it easier to change borders. Border width is still 0, so no borders are visible by default.
- The only way to get the colors right is by only colorizing the `td` and `th` backgrounds, therefore we need to work with CSS custom properties.
- Stole the background image hack from @mdo in https://github.com/twbs/bootstrap/pull/30342
- All table variants are now stored in `$table-variants`. This way we have more control over colorizing tables. The dark table is also a variant now.
	- `$table-border-factor` will determine the contrast between the background and the borders
	- `$table-accent-bg-factor` will determine the contrast between the background and striped rows
	- `$table-hover-bg-factor` will determine the contrast between the background and hovered rows 
    - The color of the text will be determined by `contrast-color()`
- Simplified the docs a little
	- Stored the table content in a partial, which is easier to reuse
	- Only kept the whole html for the first example, stripped some html in the other examples (inspired by the modal docs change)

TODO:

- [x] Document BCs in migration guide
- [x] Update partial template (see https://github.com/twbs/bootstrap/pull/30466#pullrequestreview-382820120)
- [x] The docs might need some adjustments
- [x] Cleanup git if feedback Tweaks are applied
- [x] Have a look at what tickets can be closed by this

Preview: <https://deploy-preview-30466--twbs-bootstrap.netlify.com/docs/4.3/content/tables/>

---

- Fixes https://github.com/twbs/bootstrap/issues/29244, background issues are fixed
- Fixes #30251, fixes https://github.com/twbs/bootstrap/issues/27800, nested tables styles are reset
- Fixes https://github.com/twbs/bootstrap/issues/24529, `active` is now a state and can be applied to any table variant
- Fixes https://github.com/twbs/bootstrap/issues/27879, `table-dark` is a table variant now
- Fixes https://github.com/twbs/bootstrap/issues/28160, hoverable rows work on any table variant
- Fixes https://github.com/twbs/bootstrap/issues/25675, correct colors are applied to the table variants
- Fixes https://github.com/twbs/bootstrap/issues/29290, table variant classes can be added to table,thead,tbody,tfoot,tr,th and td.
- Closes #30342


scss/_tables.scss
```diff
@@ -1,31 +1,42 @@
...
+  > thead {
     vertical-align: bottom;
-    border-bottom-color: $table-head-border-color;
   }
 
-  tbody + tbody {
-    border-top: (2 * $table-border-width) solid $table-border-color;
+  > :not(caption) + * {
+    border-top-width: var(--table-children-border-width);
   }
```
ffoodd: Are those direct children selector necessary, since they apply to every `.table`? 

It always applies, even in nested `.table`s — except when nesting a `<table>` without `.table` in a `.table`. Is that your goal?

The same goes for `> * > * > *` since all values are set with custom properties: they can be easily overriden in variants by setting those custom properties directly (which is already what you're doing).

Am I missing something?

MartijnCuppens: Styles can be leaked to `<table>`s without `.table` class 😉

ffoodd: If we're trying to solve this case, shouldn't we show an example of a `.table` containing a standard `<table>`?

It looks odd to me that one could use our table styles for a parent table and not in its table children :D

MartijnCuppens: This seems to be a common case: https://codepen.io/MartijnCuppens/pen/wvaRaVP

This PR version: https://codepen.io/MartijnCuppens/pen/NWqeGWY

site/layouts/shortcodes/partial.html
```diff
@@ -1 +1 @@
-{{ partial (.Get 0) . }}
+{{- partial (.Get 0) . -}}
```
XhmikosR: Why this change? If it's really needed (and I have to confirm myself later), it should be made in a separate PR.

scss/_tables.scss
```diff
@@ -1,31 +1,42 @@
...
   vertical-align: $table-cell-vertical-align;
-  background-color: $table-bg; // Reset for nesting within parents with `background-color`.
+  border-color: $table-border-color;
 
-  th,
-  td {
-    padding: $table-cell-padding;
-    border-bottom: $table-border-width solid $table-border-color;
+  // th & td
+  > * > * > * {
```
XhmikosR: Isn't this a bit too much?

MartijnCuppens: It's the most efficient way to target `td`s and `th`s of the current table. 

Alternative is:
```css
.table > thead > tr > td,
.table > thead > tr > th,
.table > tbody > tr > td,
.table > tbody > tr > th,
.table > tfoot > tr > td,
.table > tfoot > tr > th {
  ...
}
```
which is:
- more code
- less performant (negligible though)

site/layouts/shortcodes/partial.html
MartijnCuppens: In this PR, I use the partial to include code in the codesnippets. Without these dashes, a newline is added to the example code snippet. Only adding the last dash can be enough I think, since there is no leading newline.

scss/_tables.scss
ffoodd: Feels weird to me, I'd recommend to move the modal outside. But if you experienced issues with this kind of things, it probably should be addressed in some way :+1:

site/layouts/shortcodes/partial.html
XhmikosR: Please move this change to a separate PR and I'll have a look :)

MartijnCuppens: See https://github.com/twbs/bootstrap/pull/30472#issue-394832461

scss/_tables.scss
tmorehouse: Perhaps this should be:

```scss
> * > * > td,
> * > * > th {
```

So that it doesn't affect content in the `<caption>` element (and possibly the `<colgroup>/<col>`)

MartijnCuppens: `colgroup/col` shouldn't be an issue. `> :not(caption) > * > *` could also hep us out here? (bummer this increases specificity)

tmorehouse: Yeah. `:not(caption)` might be simpler for  reducing the ruleset size

site/layouts/shortcodes/partial.html
MartijnCuppens: Made a custom shortcode, which seems to fix the issue and decrease the code in `tables.md` even more.

mdo:
This is pretty damned awesome, pretty powerful. Nicely done!

Only feedback on this right now is that I'm really partial to the modified table styles I implemented months back with the darker, single pixel table header border :). In addition, I think we'll need more detailed "how it works" and "how to customize" sections to explain the collision of CSS variables and Sass variables.

Screen Shot 2020-04-02 at 9 41 08 AM

MartijnCuppens:
Restored the table thead styles (also added the border between multiple `tbody`s or a `tfoot`)

Would be handy if I could use https://github.com/twbs/bootstrap/pull/30502 to document things.

MartijnCuppens:
@twbs/css-review, ready for review. @mdo, could you doublecheck the doc edits?

tmorehouse:
Should the custom CSS property vars be prefixed ( i.e. `--bs--table-bg`)?  Just in case users have set up their own vars for other purposes (i.e. not bootstrap table styles)?

MartijnCuppens:
@tmorehouse, I've opened https://github.com/twbs/bootstrap/issues/30558 to discuss this, we'll need to change our custom properties in `:root` too, which is unralated to this PR.

MartijnCuppens:
`bs` prefix added ✅

MartijnCuppens:
@twbs/css-review, this one is ready to review.

ffoodd (approved):

ffoodd:
LGTM

site/content/docs/4.3/content/tables.md
```diff
@@ -314,59 +339,77 @@ Add `.table-hover` to enable a hover state on table rows within a `<tbody>`.
...
   </tbody>
 </table>
-{{< /example >}}
+{{< /highlight >}}
 
-### Small tables
+## How nesting works
 
-Add `.table-sm` to make any `.table` more compact by cutting all cell `padding` in half.
+To prevent **any** style from leaking to nested tables we worked with the child combinator (`>`). Since we needed to target all the `td`s and `th`s in the `thead`, `tbody` and `tfoot`, our selector would look pretty long. Therefor we use the rather odd looking `.table > :not(caption) > * > * ` selector to target all `td`s and `th`s of the `.table` and not a nested table. Note: if you add `<tr>`s as direct children of a table, those `<tr>` will be wrapped in a `<tbody>` by default and therefor making the selector work.
```
mdo, line 402: Love this explanation overall!

```suggestion
To prevent _any_ styles from leaking to nested tables, we use the child combinator (`>`) selector in our CSS. Since we need to target all the `td`s and `th`s in the `thead`, `tbody`, and `tfoot`, our selector would look pretty long without it. As such, we use the rather odd looking `.table > :not(caption) > * > * ` selector to target all `td`s and `th`s of the `.table`, but none of any potential nested tables.

Note that if you add `<tr>`s as direct children of a table, those `<tr>` will be wrapped in a `<tbody>` by default, thus making our selectors work as intended.
```

mdo (commented):
Just some docs copy changes, otherwise looks great!

site/content/docs/4.3/content/tables.md
```diff
@@ -8,296 +8,321 @@ toc: true
...
     </tr>
   </tbody>
 </table>
-{{< /example >}}
+{{< /highlight >}}
 
-### Bordered
+## How do the variants and accented tables work?
+
+For the accented tables ([striped rows](#striped-rows), [hoverable rows](#hoverable-rows) and [active tables](#active-tables)), we used some techniques to make these effects work for all our [table variants](#variants):
```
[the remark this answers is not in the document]
MartijnCuppens: Didn't know this was a thing, thanks!

PR #30472, opened 2020-03-27
Partial issue

MartijnCuppens:
This PR is actually an issue, I just use this to illustrate the issue with the Netlify previews.

2 newlines are added in the examples, one from the partial template, and one from the line-2.html.

By adding a dash (`{{ partial (.Get 0) . -}}`) we can fix the issue in the partial template, but I'n not sure how we should handle the second newline.

See https://deploy-preview-30472--twbs-bootstrap.netlify.com/docs/4.3/components/carousel/#partial-example


MartijnCuppens:
Gonna close this since I'm using a different technique in https://github.com/twbs/bootstrap/pull/30466 now.

PR #30475, opened 2020-03-28
Switch to custom properties to control grid gutter widths

MartijnCuppens:
Not sure yet how we should continue with the grid, so I'm just gonna make some improvements step by step which are easier to review/merge (unlike https://github.com/twbs/bootstrap/pull/28517 😛). In this PR I've switched to CSS custom properties which makes it easier to control the gutter width and saves a lot of CSS.

I've also moved the loop to save some media query prints.

CSS diff see: https://www.diffchecker.com/XRJUJMzm

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -3,10 +3,13 @@
 // Generate semantic grid columns with these mixins.
 
 @mixin make-row($gutter: $grid-gutter-width) {
+  --grid-gutter-x: #{$gutter};
+  --grid-gutter-y: 0;
   display: flex;
   flex-wrap: wrap;
-  margin-right: -$gutter / 2;
-  margin-left: -$gutter / 2;
+  margin-top: calc(var(--grid-gutter-y) * -1); // stylelint-disable-line function-blacklist
+  margin-right: calc(var(--grid-gutter-x) / -2); // stylelint-disable-line function-blacklist
+  margin-left: calc(var(--grid-gutter-x) / -2); // stylelint-disable-line function-blacklist
 }
 
 @mixin make-col-ready($gutter: $grid-gutter-width) {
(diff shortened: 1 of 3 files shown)


ffoodd:
LGTM.

Since we're using more and more custom properties, we'll be using more and more `calc()` function: shouldn't we allow them in stylelint from now on?

Unrelated thing, there's a flexbug fix  [L68](https://github.com/twbs/bootstrap/blob/ecfe1cb5ee4e97b792602b156e9323113d541b9a/scss/mixins/_grid.scss#L68) which only targets IE10-11. I'll do a PR latter.

MartijnCuppens:
> Since we're using more and more custom properties, we'll be using more and more calc() function: shouldn't we allow them in stylelint from now on?

The reason we disallow them is because `calc(... + 0)` is invalid. Might be worth reviewing later on if we bump into this more than once.

ffoodd (approved):

scss/mixins/_grid.scss
```diff
@@ -3,10 +3,13 @@
 // Generate semantic grid columns with these mixins.
 
 @mixin make-row($gutter: $grid-gutter-width) {
+  --grid-gutter-x: #{$gutter};
+  --grid-gutter-y: 0;
```
ffoodd: As a sidenote, you could remove this declaration and use `var(--grid-gutter-y, 0)` in `margin-top` to shorten things a bit. Could be useful in some cases to reduce custom properties bloat.

Not sure if it's readable enough, though.

MartijnCuppens: We do need to reset this custom property, otherwise styles will leak in to nested rows.

ffoodd: OW nested things. I need to accomodate :D

PR #30476, opened 2020-03-28
Require `.form-label` classes on form labels

MartijnCuppens:
There's currently no clean way to theme form labels, simply because they have no class. By just adding styles to the `<label>` element, styles can leak to other elements like the checkbox labels.

I added a new `.form-label` class and some `null` variables which makes it easier to theme these labels and prevent style leaks.


ffoodd (approved):

MartijnCuppens:
Gonna wait for @mdo's approval before merging this.

mdo (approved):
I've debated this for the last two major versions, so I'm definitely inclined to make the switch. Styling all `<label>`s led to a few extra lines, so making that update feels like a promising move for others.

PR #30480, opened 2020-03-28
Use `box-shadow` mixin for form controls

zalog:
This PR fixes css property `box-shadow` become invalid.
That happends if one of `$input-box-shadow` or `$input-focus-box-shadow` is `none`.
Our `box-shadow` mixin verifies for `length($string) == 1` and it covers that case.

LE: Oh, this relates with #30394 

cc:// @mdo @ysds

### DIFF
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -29,10 +29,10 @@
     background-color: $input-focus-bg;
     border-color: $input-focus-border-color;
     outline: 0;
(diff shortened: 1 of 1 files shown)


ysds:
Does this remove the focus outline when `enable-shadow: false`?

zalog:
@ysds tnks for feedback! Of course you're right :)
You can take another look now.

ysds (approved):

PR #30481, opened 2020-03-30
Remove IE11 flexbugs workarounds

ffoodd:
Not sure about this one: IE11 only workarounds for flexbugs.

Keeping them would drastically improve the backward compatibility, however v5 officially drops IE11 support and we're still trying to decrease filesize… This one doesn't do much, so feel free to close or merge this :)

### DIFF
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -163,8 +163,7 @@
     // The child selector allows nested `.card` within `.card-group`
     // to display properly.
     > .card {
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
In general we only use shorthand properties when all properties are set in the property. If we keep doing things like this, we might end up with an inconsistent usage of the `flex` property (one, two or three values).

ffoodd:
Understood, thanks for clarifying :)

PR #30482, opened 2020-03-31
Cleanup responsive tables

MartijnCuppens:
In https://github.com/twbs/bootstrap/pull/25058, the `.table-reponsive` class was moved from the table to a wrapper div.

- `display: block;` and `width: 100%;` have become redundant since the wrapper is a div
- The double border fix isn't needed either since the table keeps its `display` value
- I've also simplified the Sass class concatenation.

https://deploy-preview-30482--twbs-bootstrap.netlify.com/docs/4.3/content/tables/#breakpoint-specific

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -174,23 +174,14 @@
 // Generate series of `.table-responsive-*` classes for configuring the screen
 // size of where your table will overflow.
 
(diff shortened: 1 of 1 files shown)


ffoodd (approved):
Nice catch.

Issue #30494, opened 2020-04-01
[Proposal] CSS variables Phase 1


yordis:
## Goal

Replace static SASS variables to the identical representation using CSS variables.

### Upside

- Easy to test
- Easy to code review

### Downside

- More iterations
- Require more collaboration

### What to avoid 

- Modify more than one file at the time
- Update any code related to the usage of mixin
- Modify any code related to SASS specific functions

## How to?

1. Communication your intention about working in a particular file
2. Duplicate in `_variables.scss` the SASS variable with the equivalent CSS variable in the `scss/_root.scss` file. Use `--b-` prefix for the variables names (subject to change the naming in the future).
3. Create a PR scoped to only one file, please tag this issue in the PR
4. Repeat.

#### scss

- [ ] _alert.scss
- [ ] _badge.scss
- [ ] _breadcrumb.scss
- [ ] _button-group.scss
- [ ] _buttons.scss
- [ ] _card.scss
- [ ] _carousel.scss
- [ ] _close.scss
- [ ] _containers.scss
- [ ] _dropdown.scss
- [ ] _forms.scss
- [ ] _functions.scss
- [ ] _grid.scss
- [ ] _helpers.scss
- [ ] _images.scss
- [ ] _list-group.scss
- [ ] _mixins.scss
- [ ] _modal.scss
- [ ] _nav.scss
- [ ] _navbar.scss
- [ ] _pagination.scss
- [ ] _popover.scss
- [ ] _progress.scss
- [ ] _reboot.scss
- [ ] _root.scss
- [ ] _spinners.scss
- [ ] _tables.scss
- [ ] _toasts.scss
- [ ] _tooltip.scss
- [ ] _transitions.scss
- [ ] _type.scss #30495
- [x] _utilities.scss
- [x] _variables.scss
- [x] bootstrap-grid.scss
- [x] bootstrap-reboot.scss
- [x] bootstrap-utilities.scss
- [x] bootstrap.scss

#### scss/forms

- [ ] _form-check.scss
- [ ] _form-control.scss
- [ ] _form-file.scss
- [ ] _form-range.scss
- [ ] _form-select.scss
- [ ] _input-group.scss
- [ ] _labels.scss
- [ ] _validation.scss

#### scss/helpers

- [ ] _background.scss
- [x] _clearfix.scss
- [ ] _colored-links.scss
- [ ] _embed.scss
- [ ] _position.scss
- [x] _screenreaders.scss
- [ ] _stretched-link.scss
[...]

PR #30495, opened 2020-04-01
replace static vars for _type.scss

yordis:
### Testing Notes

- Verify http://localhost:9001/docs/4.3/content/typography styles are being applied correctly

Related to #30494

### DIFF
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -1,4 +1,16 @@
 :root {
+  --b-gray-600: #6c757d;
+  --b-headings-line-height: 1.2;
+  --b-display1-weight: 300;
+  --b-display2-weight: 300;
+  --b-display3-weight: 300;
+  --b-display4-weight: 300;
+  --b-display-line-height: var(--b-headings-line-height);
+  --b-lead-font-weight: 300;
+  --b-list-inline-padding: .5rem;
+  --b-blockquote-small-color: var(--b-gray-600);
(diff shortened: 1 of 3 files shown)


ffoodd:
I don't think this is the way to go for CSS properties, but I'd like @MartijnCuppens and @mdo opinion. Let me explain myself!

In this PR, the generated output will be bigger thant the original one; and custom properties don't make any sense to me in that particular case.

Custom properties are *not* similar to Sass variables, they can't be a simple replacement. Using them like you did in here only helps users that don't want to compile code through Sass, and doesn't solve any issue. This results in a code bloat in generated output.

**As of now, I think the goal should be to decrease file size.** Thoughts, anyone?

XhmikosR:
> **As of now, I think the goal should be to decrease file size.** Thoughts, anyone?

Definitely! 💯

ffoodd:
@XhmikosR Are we able to run the Bundlesize action on PR like this one? Would definetely help :)

XhmikosR:
You can see the result in tests https://github.com/twbs/bootstrap/pull/30495/checks?check_run_id=550704973#step:10:9

MartijnCuppens:
> Custom properties are not similar to Sass variables, they can't be a simple replacement. Using them like you did in here only helps users that don't want to compile code through Sass, and doesn't solve any issue. This results in a code bloat in generated output.

Yup, I agree. Gonna close this.

PR #30498, opened 2020-04-02
wip: adds `$nav-link-*` vars

zalog:
This PR adds a new vars for `$nav-link-*`'s:
```
$nav-link-font-size:                null !default;
$nav-link-font-style:               null !default;
$nav-link-font-weight:              null !default;
$nav-link-color:                    $primary !default;
$nav-link-hover-color:              darken($primary, 15%) !default;
```

This PR depends on https://github.com/twbs/bootstrap/pull/30622

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7047,9 +7047,9 @@
       }
     },
     "karma": {
-      "version": "5.0.1",
-      "resolved": "https://registry.npmjs.org/karma/-/karma-5.0.1.tgz",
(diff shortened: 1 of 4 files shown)


zalog:
I think we also need to add hover and active states as well. What do you think?

MartijnCuppens:
> I think we also need to add hover and active states as well. What do you think?

Any suggestions theming wise?

zalog:
I'll come back here soon, this days.

zalog:
Removed wip and rebased.

ffoodd:
LGTM.

FWIW, I'd go with the active state too — even if the active state should not rely on color only (but this is another topic to tackle in Bootstrap and/or examples, I guess).

mdo:
Closing for #31035.

Issue #29422, opened 2019-09-19
Text and Non-text Contrast


lucalanca:
**Operating system and version**: All
**Browser and version**: All
**Report**: https://www.aditus.io/button-contrast-checker/getbootstrap-com-2019-09-12-at-15-02-11-139

Most buttons and links of the framework don't meet contrast criteria in all or some of their states.

## Examples

**Primary Button**
![Screenshot 2019-09-19 at 14 42 29](https://user-images.githubusercontent.com/389459/65249370-c6dc3200-daeb-11e9-8a62-154ab417075c.png)

**Success Button**
![Screenshot 2019-09-19 at 14 43 12](https://user-images.githubusercontent.com/389459/65249405-d78ca800-daeb-11e9-8552-eb46a233eef4.png)

Happy to help you guys in this regard, but I think it's mostly a design decision that needs to be made on the default colors. I want to take some time to see if I can investigate a set of variables that pass AA and AAA when it comes to contrast without changing too much of the design.

XhmikosR:
Thanks for this, we have already many issues in the past about it, so we are definitely aware of this.

I have already landed a PR for the docs for master, and for core I have #29198 but I don't like the changes except for the blue color and maybe the pink one.

Also, note I have a WIP PR to automatically check for accessibility issues on #29315.

lucalanca:
@XhmikosR your branch is a lot better. It still has a lot of issues but it's definetely better:

![Screenshot 2019-09-19 at 15 23 59](https://user-images.githubusercontent.com/389459/65252890-a1eabd80-daf1-11e9-9738-6d2f035f366e.png)

For example, comparing to the buttons I shared above:

**Primary Button**
![Screenshot 2019-09-19 at 15 24 23](https://user-images.githubusercontent.com/389459/65252953-baf36e80-daf1-11e9-88ab-e5c33552ba8d.png)

**Success Button**
![Screenshot 2019-09-19 at 15 24 17](https://user-images.githubusercontent.com/389459/65252945-b7f87e00-daf1-11e9-81ee-e3e3f5992496.png)


I don't know where is the best place to focus on contrast of buttons alone. This issue or the one of your PR.

XhmikosR:
I don't think we'll land the other colors changes. I mean, there's a huge difference for green.

But for blue and maybe pink, we can move forward in my PR.

XhmikosR:
Blue and pink should be fine now on master. I'm not sure about the other colors because they change a lot.

[...]

PR #30500, opened 2020-04-02
Switch to custom properties for colorizing buttons

MartijnCuppens:


scss/_buttons.scss
```diff
@@ -43,38 +57,82 @@
...
+    background-color: var(--btn-accent-color);
+    border-color: var(--btn-accent-color);
+  }
+
+  &.active,
+  &:active {
+    color: var(--btn-accent-contrast-color);
+    background-color: var(--btn-accent-color);
+    border-color: var(--btn-accent-color);
+  }
```
ffoodd, line 89: Those are the same, can't we have a single selector stack?

ffoodd:
Looks very promising, makes a lot of sense.

MartijnCuppens:
We need to fix https://github.com/twbs/bootstrap/issues/29422#issuecomment-570641382 first, since the focus states also need to change for inputs, otherwise we 'll have inconsistent focus styles.

scss/_buttons.scss
```diff
@@ -43,38 +57,82 @@
...
+}
```
yordis, line 90: If I am not mistaken, the approach would be something like,

```css
.btn {
  --btn-color: default here;
  --btn-background-color: default here;
  --btn-border-color: currentColor;

  color: var(--btn-color);
  background-color: var(--btn-background-color);
  border-color: var(--btn-border-color);
}

.btn-outline {
  --btn-color: change me;
  --btn-background-color: change me;
  --btn-border-color: change me;

  &:hover {
     --btn-color: change me;
     --btn-background-color: change me;
     --btn-border-color: change me;
  }

  &:focus,
  &.focus {
      --btn-color: default here;
      --btn-background-color: default here;
      --btn-border-color: currentColor;
  }

  &.active,
  &:active {
      --btn-color: change me;
      --btn-background-color: change me;
      --btn-border-color: change me;
  }
}
```

the idea is that you focus more on swapping the values than overriding styles.

At least that is how I understand it.

scss/_buttons.scss
```diff
@@ -43,38 +57,82 @@
...
   }
 }
 
+.btn-outline {
+  color: var(--btn-accent-color);
+  background-color: transparent;
+  border-color: currentColor;
+
+  &:hover {
+    background-color: var(--btn-accent-color);
```
ydmitry, line 73: If we put here `--btn-accent-color` for hover `background-color` and for `color` library user can't make these values different without styles overwrite or setting own selectors with variables values.

For example I want to set `color` as `red`, but `background-color` as `green`
I will need to write:
```
.btn-outline {
  --btn-accent-color: red;
  &:hover {
    --btn-accent-color: green;
  }
}
```
I think better do config just with set of variables without writing new nested selectors. What do you think?
And, btw, isn't there color will be the same as background color or not?

ydmitry: If we use approach similar to https://github.com/twbs/bootstrap/pull/31708 with placing in SCSS styles only SCSS variables, but in SCSS variables set CSS variables, we could do:

```
// Styles
.btn-outline {
  color: $btn-color;
  &:hover {
     background-color: $btn-hover-bg-color;
  }
}

// Default variables, will do the same as in PR
$btn-color: var(--btn-accent-color) !default;
$btn-hover-bg-color: var(--btn-accent-color) !default;

// In application can changed to e.g.:
$btn-color: var(--btn-color);
$btn-hover-bg-color: var(--btn-hover-bg-color);

// In application custom styles - setting variant without knowing selectors structure
.btn-custom {
  --btn-color: red;
  --btn-hover-color-bg: green;
}

File: scss/_buttons.scss [hunk 4]
```diff
@@ -43,38 +57,82 @@
...
   .btn-#{$color} {
-    @include button-variant($value, $value);
-  }
-}
+    $contrast-color: color-contrast($value);
 
-@each $color, $value in $theme-colors {
-  .btn-outline-#{$color} {
-    @include button-outline-variant($value);
+    --btn-accent-color: #{$value};
```
felixfbecker, line 100: I'm a bit confused by the fact that the variables are set here. The nice thing about CSS variables are that they cascade through the DOM hierarchy, so that devs can set these e.g. at the `:root` or some other ascendant of a button, and every button (and other element that was configured in the ancestor) below that ancestor will use the same theme - without having to use any descendant selectors for `.btn` etc.

But if the variable is set here on the `.btn` element, then it seems that wouldn't work, because it would always override what was set in an ancestor. So to make it work, define a theme for all descendant Bootstrap components at an ancestor element, it seems we'd have to use specific descendant selectors like `.btn` again to override the variables. That kind of defeats the purpose of the variables mostly.

How I imagine it, components like `.btn` would only _read_ variables, and they are set e.g. on `:root` or any ancestor by devs (and Bootstrap can provide defaults on those and integrate them there with the Sass variables).

ydmitry: I have the same expression, looks like CSS custom properties here used to refactor code in a cool way, not to increase customization capabilities and flexibility...

XhmikosR:
@MartijnCuppens can you rebase this please?

mdo:
Closing as stale and with hella conflicts 😅

PR #30502, opened 2020-04-02
Create `scss-docs` shortcode to get some snippets from the Scss source files

MartijnCuppens:
This should be a little bit easier that creating a custom regex per case.

Usage in `.scss`:
``` scss
// scss-docs-start awesome-code
.my-awesome-code {
  // The best CSS in the world
}
// scss-docs-end awesome-code
```

In the docs files:
``` twig
{{< scss-docs name="awesome-code" file="scss/_variables.scss" >}}
```

I don't think the positive lookbehind syntax works in Hugo, so I replaced the comments in the end of the file, there are probably better ways to fix this (bump @XhmikosR)

TODO:
- [x] Check all occurrences of code snippets
- [x] Strip `!default` flags by default

### DIFF
--- a/scss/_alert.scss
+++ b/scss/_alert.scss
@@ -40,8 +40,7 @@
 }
 
 
-// Alternate styles
-//
+// scss-docs-start alert-modifiers
 // Generate contextual modifier classes for colorizing the alert.
 
 @each $color, $value in $theme-colors {
@@ -49,3 +48,4 @@
     @include alert-variant(color-level($value, $alert-bg-level), color-level($value, $alert-border-level), color-level($value, $alert-color-level));
   }
 }
+// scss-docs-end alert-modifiers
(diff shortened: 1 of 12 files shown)


mdo:
Oh you brilliant son of a bucket, that's rad. This is fresh as hell.

ffoodd:
I love this, it'll make things so much easier!

MartijnCuppens:
Neither the positive lookbehind syntax nor grouping seem to be supported in Hugo regex, so the replacing the comments seems to be the way to go.

Ready for review!

XhmikosR:
It's in my backlog 🙂

I'll check what's supported but this seems to do the job anyway.

mdo (approved):
This looks lovely to me, so adding an approval for whatever it's worth :).

XhmikosR (approved):

PR #30508, opened 2020-04-03
Remove useless quotes in font-stacks

ffoodd:
* [Matthias Bynens wrote about "unquote font family names in CSS"](https://mathiasbynens.be/notes/unquoted-font-family) back in 2012;
* and went with [a neat tool to check wheter your font-family name requires quotes or not](https://mothereff.in/font-family);
* FWIW, [StyleLint rule to check quotes in family names](https://stylelint.io/user-guide/rules/font-family-name-quotes) is [set to `where-recommended` in our StyleLint config](https://github.com/twbs/stylelint-config-twbs-bootstrap/blob/master/css/index.js#L22) → [PR #55 opened](https://github.com/twbs/stylelint-config-twbs-bootstrap/pull/55)

### DIFF
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -8,8 +8,6 @@
     --#{$color}: #{$value};
   }
 
-  // Use `inspect` for lists so that quoted items keep the quotes.
-  // See https://github.com/sass/sass/issues/2383#issuecomment-336349172
-  --font-family-sans-serif: #{inspect($font-family-sans-serif)};
-  --font-family-monospace: #{inspect($font-family-monospace)};
+  --font-family-sans-serif: #{$font-family-sans-serif};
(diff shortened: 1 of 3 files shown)


XhmikosR:
Why doesn't Stylelint fail here, though?

MartijnCuppens:
@XhmikosR: https://github.com/twbs/bootstrap/pull/27771#issuecomment-444412153

XhmikosR:
Thanks, I forgot about it. Weird that this is still an issue.

ffoodd:
As we agreed in [#55 of stylelint-config](https://github.com/twbs/stylelint-config-twbs-bootstrap/pull/55), let's leave it as-is.

PR #30509, opened 2020-04-03
Use custom properties for font-stacks

ffoodd:
Just a POC for now, should work  as a basis but requires some proofreading and to exposed to @twbs/css-review wisdom and experience :D

- [x] check output's filesize to see if it's a good idea!
- [x] explore using different font-stacks for headings, inputs, etc.: started something in `_root.scss` but very coarse for now

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -441,7 +441,7 @@ $utilities: map-merge(
     "font-family": (
       property: font-family,
       class: font,
-      values: (monospace: $font-family-monospace)
+      values: (monospace: var(--font-family-monospace))
     ),
     "user-select": (
       property: user-select,
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -369,7 +369,7 @@ $embed-responsive-aspect-ratios: (
 // stylelint-disable value-keyword-case
 $font-family-sans-serif:      -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;
 $font-family-monospace:       SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace !default;
-$font-family-base:            $font-family-sans-serif !default;
+$font-family-base:            var(--font-family-sans-serif) !default;
 // stylelint-enable value-keyword-case
 
 // $font-size-root effects the value of `rem`, which is used for as well font sizes, paddings and margins
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -283,7 +283,7 @@ pre,
 code,
 kbd,
 samp {
-  font-family: $font-family-monospace;
+  font-family: var(--font-family-monospace);
   @include font-size(1em); // Correct the odd `em` font sizing in all browsers.
 }
 


scss/_root.scss
```diff
@@ -12,4 +12,21 @@
   // See https://github.com/sass/sass/issues/2383#issuecomment-336349172
   --font-family-sans-serif: #{inspect($font-family-sans-serif)};
   --font-family-monospace: #{inspect($font-family-monospace)};
+  --font-family-base: var(--font-family-sans-serif);
+
+  @if not $headings-font-family = null {
```
MartijnCuppens, line 17: Shouldn't this be `==`?

MartijnCuppens: Or just `!=` since we use not

MartijnCuppens:
Might be worth checking out if it's worth blacklisting the `$font-family-*` values so that the custom properties are always used (would only enable that rule in the config of this repo).

ffoodd:
@MartijnCuppens FWIW, if we blacklist `$font-family-*`, we'd have no way to prevent outputting `font-family` property if variable is `null`—am I wrong?

FYI I think it's ready but still need to review docs and migration, I guess.

ffoodd:
I didn't find any thing to change in the docs: this is not a BC from a user nor dev perspective, and our docs only mentions the native font-stack in Theming.

Let me know if you think about any reference to change about font stacks.

scss/_buttons.scss
```diff
@@ -4,7 +4,7 @@
 
 .btn {
   display: inline-block;
-  font-family: $btn-font-family;
+  font-family: if($btn-font-family != null, var(--btn-font-family), null);
```
MartijnCuppens, line 7: Why do we need this test?

scss/_root.scss
```diff
@@ -12,4 +12,25 @@
...
+    --input-font-family: #{inspect($input-font-family)};
+  }
+
+  @if $form-select-font-family != null {
+    --form-select-font-family: #{inspect($form-select-font-family)};
+  }
+
+  @if $form-file-font-family != null {
+    --form-file-font-family: #{inspect($form-file-font-family)};
+  }
```
MartijnCuppens, line 35: Instead of all the if test we could just use `#{}` for the custom properties, this way we wouldn't need inspect() either: https://codepen.io/MartijnCuppens/pen/LYpYdEN

scss/_buttons.scss
MartijnCuppens: ok, I get it, nvm

ffoodd: If we simply use `var()`, the property would be generated even if the Sass variable is `null` —thus not defining the custom property—which would defeat the whole point.

Am I wrong?

MartijnCuppens:
Not sure about adding custom properties for each component. I don't wan't `:root{}` to become a gigantic wrapper for every possible variable.

ffoodd:
Sure, but why?

We're just tackling font-family here, most of them are `null` in our standard build. And FWIW, `font-family` is probably the best (the only?) value to be set this way: very long values, used several times, and applied everywhere.

We already have colors in `:root` and we don't even use them.

Just to notice: Bundlewatch notifies a 150b gain in main CSS file. Didn't look at gzipped yet, but looks promising.

MartijnCuppens:
I don't see any benefit if you want to use for example `"Comic Sans MS", cursive, sans-serif` as button font family, then the following output is rendered:

```css
:root {
  --btn-font-family: "Comic Sans MS", cursive, sans-serif;
}
.btn {
  font-family: var(--btn-font-family);
}
```
instead of
```css
.btn {
  font-family: "Comic Sans MS", cursive, sans-serif;
}
```

What's the gain here?

ffoodd:
Those called only once, there's no gain in our standard build, you're right.

The only potential gain is reusability, which is over-engineered for us, I guess.

I'll try to remove font stacks with a single occurrence!

scss/_utilities.scss
```diff
@@ -441,7 +441,7 @@ $utilities: map-merge(
     "font-family": (
       property: font-family,
       class: font,
-      values: (monospace: $font-family-monospace)
+      values: (monospace: var(--font-family-monospace))
```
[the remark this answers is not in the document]
ffoodd: Wouldn't `font-sans-serif` be a better option, here? Or maybe use `family-base` as a value, to get `.font-family-base`?

scss/_root.scss
ffoodd: Wow, didn't know about this!

FWIW `if` statements have been dropped since they're all matching a called-only-once property, thus unhelpful as you mentionned in another comment.

Should I use this for `--font-family-sans-serif` and `--font-family-monospace`?

ffoodd:
@MartijnCuppens Two remaining questions here:
1. add a utility fonr `font-family-base`?
2. keep existing `inspect()` or use interpolation on custom properties?

scss/_variables.scss
```diff
@@ -369,7 +369,6 @@ $embed-responsive-aspect-ratios: (
 // stylelint-disable value-keyword-case
 $font-family-sans-serif:      -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;
 $font-family-monospace:       SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace !default;
-$font-family-base:            $font-family-sans-serif !default;
```
MartijnCuppens, line 372: This should be `$font-family-base: var(--font-family-sans-serif)`, which can be used in the `:root`.

scss/_utilities.scss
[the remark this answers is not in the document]
MartijnCuppens: Maybe leave it for now, not sure what's best yet

MartijnCuppens:
> add a utility font font-family-base?

Let's leave that for now, not sure about this yet.

> keep existing inspect() or use interpolation on custom properties?

Maybe we should use interpolation for all our custom properties, so that they can be nullified if needed. Something to look into later maybe?

scss/_variables.scss
ffoodd: You're right, much cleaner and less diff!

MartijnCuppens (approved):

PR #30517, opened 2020-04-06
fix: allow popper.js@1.16.1-lts

FezVrasta:
The docs should be updated as well, so that users are instructed to install `popper.js@1.16.1-lts`. This will allow your consumers to not get notified about the deprecation notice included in the `stable` release branch of Popper.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
   "dependencies": {},
   "peerDependencies": {
     "jquery": "1.9.1 - 3",
-    "popper.js": "^1.16.0"
+    "popper.js": "^1.16.0 || 1.16.1-lts"
   },
   "devDependencies": {
     "@babel/cli": "^7.10.3",
@@ -176,7 +176,7 @@
     "dependencies": {},
     "peerDependencies": {
       "jquery": "1.9.1 - 3",
-      "popper.js": "^1.16.0"
+      "popper.js": "^1.16.0 || 1.16.1-lts"
     }
   }
 }
--- a/site/docs/4.5/getting-started/webpack.md
+++ b/site/docs/4.5/getting-started/webpack.md
@@ -28,7 +28,7 @@ import 'bootstrap/js/dist/alert';
 
 Bootstrap depends on [jQuery](https://jquery.com/) and [Popper](https://popper.js.org/),
 which are specified in the `peerDependencies` property; this means that you will have to make sure to add both of them
-to your `package.json` using `npm install --save jquery popper.js`.
+to your `package.json` using `npm install --save jquery popper.js@1.16.1-lts`.
 
 ## Importing Styles
 


XhmikosR:
Thanks, but I'd rather tackle all the related changes in one PR. Are you able to do it yourself or should do you want me to tackle it later?

FezVrasta:
I can try to update the docs, may you confirm `v4-dev` is the correct target branch?

XhmikosR:
Yup, assuming in master we'll manage to update to popper.js 2.x 🙂

FezVrasta:
Done, please let me know if you think I missed something.

XhmikosR:
I think we should update the Nuget files too.

BTW with this tag, I wonder how things will be with the semver operators for people. Say you release a `1.16.1-lts`, this will manual updating, right?

FezVrasta:
I don't plan to release any additional version on the 1.x development branch, unless security vulnerabilities are discovered (which is very hard, considered the scope of the library).

In that case, I'll release a new stable release (1.16.2 - with deprecation warning) and a `1.16.2-lts` that you will need to explicitly whitelist like I did for this PR. I'm not aware of any way to allow to whitelist multiple pre-release tags spawning between different versions in a single command.

XhmikosR:
Maybe we could use the `lts` tag instead? This should work in theory assuming you always keep the lts tag present.

I played with it with <https://semver.npmjs.com/> and using `lts` seems to work.

FezVrasta:
So it would be `popper.js@lts`? How would you define a specific version range though?

XhmikosR:
AFAICT you don't. The tag will always point to the latest `lts` tag. Similarly to what the `latest` tag does.

But anyway, that's something you should decide, I'm just spitballing here 🙂

So, basically this PR just needs the NuGet files updated and the rest should be fine. I'll review it after that locally.

FezVrasta:
```diff
-       <dependency id="popper.js" version="[1.16.0,2)" />
+      <dependency id="popper.js" version="[1.16.1-lts,2)" />

```

Is this change correct? I have no idea what the `,2` refers to 😅

XhmikosR:
It means `< 2` AFAICT.

FezVrasta:
> AFAICT you don't. The tag will always point to the latest `lts` tag. Similarly to what the `latest` tag does.

The `1.16.1-lts` is already published on `@lts` so I think the command should work, in case a user runs `npm i popper.js@lts` though, the `package.json` will be populated with `^1.16.1.-lts`, which allows `1.16.2` (non-lts), which may lead to some confusion if a new release is ever published.

nuget/bootstrap.nuspec
```diff
@@ -18,7 +18,7 @@
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <dependencies>
       <dependency id="jQuery" version="[3.0.0,4)" />
-      <dependency id="popper.js" version="[1.16.0,2)" />
+      <dependency id="popper.js" version="[1.16.1-lts,2)" />
```
XhmikosR, line 21: Now that I think about it, maybe this allows for non-LTS versions.

@supergibbs what do you think?

FezVrasta: the NuGet package doesn't ship with a deprecation warning, so it doesn't really make any difference.

supergibbs: Adding a `-lts` to semver [designates it as pre-release](https://semver.org/#spec-item-9) so any platform that implements it correctly will think `1.16.1-lts` is before `1.16.1`. You can see that on [nuget.org](https://www.nuget.org/packages/popper.js/). This also [applies to npm](https://semver.npmjs.com/); installing `^1.16` would get you `1.16.1` not `1.16.1-lts`. It would be better to release a `1.16.2` and mention in the release notes it's LTS. You can also deprecate it on nuget.org to let users know it's no longer maintained.

https://docs.microsoft.com/en-us/nuget/nuget-org/deprecate-packages

FezVrasta: I don't plan to deprecate the NuGet package, I didn't want to deprecate the npm either, but it's the only way to warn users about a package rename.

The NuGet support for Popper ended with 1.x so there's not a new namespace to point the users to.

Personally I would keep the NuGet dependency at 1.16.1 and update only the npm dependency.

supergibbs: Ya, if there is no functional differences I’d just leave the dependency as 1.16.1

Another side note, LTS is often on a minor version, in this case 1.16, meaning no new features (aka 1.17) just security patches going forward. Then 1.16.2 would then be lts too if there was some critical reason to release it. 👍🏼

XhmikosR:
@FezVrasta sorry for the late response here.

So, this PR seems to be ready as a first patch, but I was thinking maybe we should update to 1.16.1-lts in our devDependencies in another PR? There haven't been too many code changes AFAICT, just a small cleanup.

Also, is the project name "Popper" or "Popper.js"? We seem to be using both on our docs and I'd like to streamline this.

XhmikosR:
BTW I reverted the _config.yml change because the hashes won't match, hence why my point above about updating to 1.16.1-lts in our devDependencies.

supergibbs:
Sorry, not sure how I missed this thread. For semver, `1.16.1-lts` means prerelease and if bootstrap depends on it, it’ll be also be considered prerelease (by Nuget). You can use a `+lts` instead. Meant for build number or meta data and is ignored. So `1.16.1+lts` would be the same as `1.16.1`. Not sure how that would work since there is a `1.16.1` already. From what I can tell, there isn't any significant differences in `1.16.1` vs `1.16.1-lts` so I vote updating Nuget to `1.16.1` or release a `1.16.2` if I missed something.

PR #30528, opened 2020-04-07
Add checkbox transition

### LINKED ISSUE #30526: Bootstrap v5 removes animations even when prefers-reduced-motion is unset.
Bootstrap v5 removes animations even when prefers-reduced-motion is unset. Tested on Windows 10 and Android 10 phone running the latest version of chrome.
See https://deploy-preview-29925--twbs-bootstrap.netlify.com/docs/5.0/forms/checks/

MartijnCuppens:
Closes https://github.com/twbs/bootstrap/issues/30526. Transitioning the `background-position` for the switch ain't ideal, but since it's such a small element, I don't expect any dramatic performance downsides.


mdo (commented):
Would we ever consider splitting the transition variables out to reduce repetition?

```scss
$form-check-transition-duration: .15s !default;
$form-check-transition-timing: ease-in-out !default;
$form-check-transition-property: background-color, background-position, opacity !default;
```

mdo (commented):
_Dupe_

MartijnCuppens:
> Would we ever consider splitting the transition variables out to reduce repetition?

Yup, that might be a lot clearer. I sticked to this way to be consistent with the buttons and inputs.

I think we better tackle this in another PR, and have a look at the benefits/disadvantages there.

ysds (approved):

mdo (approved):

Justsnoopy30:
Side affect: Now when a change occurs, like a checkbox being checked or a radio choice being changed, during the transition, the box/radio slightly moves down and to the right.

Issue #27163, opened 2018-08-31
Restyle close icon's focus style


raqibjs:
This button shouldn't be styled that way when focused (clicked) :(

![image](https://user-images.githubusercontent.com/11327451/44931830-ca880400-ad84-11e8-8b64-cb7b3ccc5578.png)

mdo:
If it's focused, it should have a focus style, no?

patrickhlauke:
until browsers have more consistent support for `:focus-visible` (and we also supplement it with a polyfill https://github.com/WICG/focus-visible), i'm afraid that yes there's no way around this. (on `:focus-visible`, see also https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/)

arraqib:
@mdo Yes of course. But shouldn't we make a better style on focus?

patrickhlauke:
PRs / suggestions welcome

TruptiM18:
Hi @patrickhlauke ,
I went through your blog https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/ .
Can we use the negation pseudo class approach that you have mentioned in the blog for solving this issue?
Thanks

patrickhlauke:
We could, but since support for `:focus-visible` is currently...bad https://caniuse.com/#search=focus-visible ... and we're not adding the polyfill at this point ... it would be rather pointless for now

flammea:
Hi, I can suggest 2 solutions.
1. Font-weight
![image](https://user-images.githubusercontent.com/27705542/56751244-e1e98680-6785-11e9-8c73-ef65127e06d7.png)

2. Underline
![image](https://user-images.githubusercontent.com/27705542/56751312-09d8ea00-6786-11e9-9cf5-89683a4df4b3.png)

mdo:
Unfortunately I don’t think a color change is enough for focus states. Neither is underline, which is synonymous with hover states. Perhaps we could at least revisit the `width` and `height` so it’s a perfect square, and then maybe make use of `border`/`box-shadow` like we use on inputs and buttons?

HERBOD7:
why you set outline none?

Demactus:
Hello, I would like to have a look at it together with a colleague from university. If its still open, updates and suggestion will follow

PR #30532, opened 2020-04-08
Misaligned close button in toasts examples

ffoodd:
https://deploy-preview-30532--twbs-bootstrap.netlify.com/docs/4.3/components/toasts/#basic


MartijnCuppens:
I prefer the current implementation:

Current:
image

PR:
image

ffoodd:
Oooh I think I understand what's going on… Might be font related, since I'm on Ubuntu. This is what it looks like on both Chrome and Firefox on Ubuntu:

Master:
Screenshot_2020-04-08 Toasts

This PR:
Screenshot_2020-04-08 Toasts(1)

Removing `.mb-1` may not be the answer, but this alignment should probably not depend on font / OS, am I wrong?

MartijnCuppens:
we should ditch the × in favour of an SVG which is easier to align and style (see https://github.com/twbs/bootstrap/issues/27163)

ffoodd:
Agreed, I need an override in Boosted for now!

I'm closing this.

PR #30533, opened 2020-04-08
Use correct hex color code

ffoodd:


ysds (approved):

MartijnCuppens (approved):
Haha, you found the easter egg. Should be `00` indeed, little typo.

PR #30534, opened 2020-04-08
Improve table markup in Layout / Containers

ffoodd:
1. Drop useless `.text-left` on table
2. Use `scope="col"` on `th`, and replace empty `th` with `td`
3. Use `th scope="row"` on class names' column.

This is all for accessibility purpose —since class names being headers makes sense here. [See WAI tutorials about tables](https://www.w3.org/WAI/tutorials/tables/two-headers/#table-with-header-cells-in-the-top-row-and-first-column).

@patrickhlauke Feel free to tell me if it's not helpful :) I made separate commits to ease reverting some.


mdo (approved):

XhmikosR:
@patrickhlauke LGTY?

XhmikosR:
Arguably, we could probably switch this table to Markdown? But that's for another PR.

XhmikosR:
@patrickhlauke I think this looks good, I'm going to merge it later today.

patrickhlauke:
> Arguably, we could probably switch this table to Markdown? But that's for another PR.

sorry coming in late, but just to touch on this point: I believe (unless I'm missing something obvious) that markdown tables still don't allow to define a row header. while not absolutely critical, it's nonetheless nice from a structural point of view to explicitly be able to say that the cells in the first column are the headers for their respective rows.

(though if I'm mistaken, and there IS a way to get markdown to explicitly mark row headers...I'd be super interested to hear it as I've recently switched to using markdown tables for some of my resources, and feel dirty not having row headers)

XhmikosR:
It's still not possible AFAICT.

But currently we seem to use a mix of HTML and Markdown tables across the codebase. We should streamline this stuff and make sure what we do is right. :)

Do note that we can pass extra attributes in Markdown, since kramdown is used in v4-dev, but for master I think we can't, so better not touch this.

PR #30541, opened 2020-04-08
Add v4-dev in bundlewatch config.

XhmikosR:
I'm not sure if this will fix the issue we have on v4-dev, but it's worth the try I guess


MartijnCuppens (approved):
Let's find out

PR #30547, opened 2020-04-10
Prevent markdown styling in scss-docs attributes

MartijnCuppens:
The underscores in the scss-docs sortcode triggered markdown styling. To prevent this from happening, we can better switch to backticks, which will prevent this from happening.

### DIFF
--- a/site/content/docs/4.3/forms/validation.md
+++ b/site/content/docs/4.3/forms/validation.md
@@ -372,8 +372,8 @@ Please note that we do not recommend customizing these values without also modif
 
 This is the Sass map from `_variables.scss`. Override this and recompile your Sass to generate different states:
 
-{{< scss-docs name="form-validation-states" file="scss/_variables.scss" >}}
+{{< scss-docs name=`form-validation-states` file=`scss/_variables.scss` >}}
 
 This is the loop from `forms/_validation.scss.scss`. Any modifications to the above Sass map will be reflected in your compiled CSS via this loop:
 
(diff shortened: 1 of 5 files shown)


XhmikosR:
Not quite sold on this. We use double quotes in all shortcodes.

MartijnCuppens:
@XhmikosR, any other suggestions to fix the issue?

XhmikosR:
Nope. I'd say we should just live with it. It's not that big of a deal. There are many cases GitHub chokes on stuff and that's expected and normal since we don't use proper Markdown everywhere. The Hugo (or Jekyll or whatever tool) code can cause this stuff.

MartijnCuppens:
It is pretty annoying in Atom though:
image
and I think lot's of developers are using Atom, like @mdo.

XhmikosR:
Like I said, we don't use Markdown only, so this is quite normal and it already happens in other places.

If we go this way, it's yet another thing we have to remember, and I personally favor simplicity over working around editors' issues.

This happens in all of our docs. And it will keep happening.

XhmikosR:
Also, VSCode works fine so I'd suggest that you report the issue to Atom or GitHub. If the underscore is in quotes, it shouldn't have been interpreted like it does in your screenshot.

Annotation 2020-04-10 095806

MartijnCuppens:
Opened https://github.com/github/markup/issues/1345, I hope it's in the right repo. This should be fixed upstream indeed.

PR #30552, opened 2020-04-11
proposal: simplify the border-radius mixin

iamnewton:
Just a proposal, but this would might simplify the various mixins that continue to grow in this file. Thoughts?

I realize this breaks some of the existing code within the framework, but if y'all like this, I'll fix all of that, but don't want to put in the effort if this is not on par with the project.

### DIFF
--- a/scss/mixins/_border-radius.scss
+++ b/scss/mixins/_border-radius.scss
@@ -15,24 +15,24 @@
 }
 
 @mixin border-radius($radius, $direction: null, $fallback-border-radius: false) {
-	$directions: (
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
I do like the simple syntax we now have with the existing mixins, so I would like to stick with the current implementation. @twbs/css-review, your thoughts?

mdo:
> I do like the simple syntax we now have with the existing mixins, so I would like to stick with the current implementation. @twbs/css-review, your thoughts?

Agreed :)

PR #30555, opened 2020-04-11
Use `box-shadow` mixin for `.form-select`

zalog:
This PR fixes css property box-shadow become invalid when `$enable-shadows: true;` and one of `$form-select-box-shadow` or `$form-select-focus-box-shadow` are `null`.


MartijnCuppens (approved):
Thanks!

MartijnCuppens:
This should be backported manually

PR #30557, opened 2020-04-11
Simplify checkbox & radio input

MartijnCuppens:
This PR simplifies the checkboxes a bit. A lot of properties are very unlikely to change, like `background-position` or `background-repeat`, so I think it's better to remove them from the variables file to keep things clear.

- Define background properties in `.form-check-input`
- Use `background-size: contain` to simplify background sizes (especially for those with linear gradients)
- Remove double `color-adjust`
- Adjust svgs to 20x20 box. This way the strokes aren't rescaled.

Also tested this with the gradients enabled.

Fixes https://github.com/twbs/bootstrap/pull/30528#issuecomment-612219215

Demo: https://deploy-preview-30557--twbs-bootstrap.netlify.com/docs/4.3/forms/checks/


MartijnCuppens:
Ping @twbs/css-review for reviwe

ffoodd (approved):
LGTM. Being curious again, is there a reason not to use `background` shorthand here? Seems pretty safe in that case, isn't it?

MartijnCuppens:
We try to live with the rule: Only use shorthand properties if all properties are covered. `background` is a shorthand of 8 background properties.

Another thing, someone might have configured something like `background-position: inherit`, in the shorthand, `inherit` doesn't work. 

I also once came across an issue where someone used custom properties, similar to the `inherit` issue.

ffoodd:
> Only use shorthand properties if all properties are covered.

That's a good one :)

mdo:
Missed this entirely, nice!

PR #30559, opened 2020-04-11
Update jQuery to v3.5.0.

### LINKED ISSUE #30553: Bootstrap v4.4.1 collapse doesn't work with jQuery v3.5.0
Hello together,

after upgrade to jQuery 3.5.0 I'm getting an exception 
`TypeError: Cannot convert object to primitive value`
while clicking to navbar-toggler button

It leads to the line
https://github.com/twbs/bootstrap/blob/dca1ab7d877bc4b664b43604657a2b5fbe2b4ecb/js/src/collapse.js#L346

Works fine with jQuery 3.4.1.

Any system and browser.

XhmikosR:
Preview: <https://deploy-preview-30559--twbs-bootstrap.netlify.com/>

Fixes #30553


XhmikosR:
I wonder if we should add the patch in master too even though it's not affected? It does make the intention clearer.

Johann-S:
We can make that change in v5, it make sense to me 👍

Johann-S (approved):

PascalPixel:
Did not check if we included it here, but could we also list jQuery and the specific version number as a dependency on our npm package?

DanFTRX:
@XhmikosR Is this patch going into the 3.x series as well?  I hit this bug running bootstrap 3.4.1 with jQuery 3.5.0

XhmikosR:
No. v3 is EOL. But either way, the jQuery developers have already merged a fix on their side, they just need to release a patch version.

DanFTRX:
@XhmikosR Yeah I just read that in the original ticket,  the comments were "off-topic" so not originally viewable.  Thanks for the response though!

peon501:
I use jquery v3.5.0 and Bootstrap v4.4.1 and it doesn't work with it.

JamesQMurphy:
jQuery 3.5.1 is out and it fixed my UI issues.

daraaleja:
> jQuery 3.5.1 está fuera y solucionó mis problemas de IU.

Hello, sorry to bother you, I'm new to this topic but since you updated to 3.5.1, could you tell me the steps? thanks

JamesQMurphy:
> > jQuery 3.5.1 está fuera y solucionó mis problemas de IU.
> 
> Hello, sorry to bother you, I'm new to this topic but since you updated to 3.5.1, could you tell me the steps? thanks

In your `package.json` file, look for a line that says `"jquery": "^3.5.0"` and change the version number there.

PascalPixel:
> Did not check if we included it here, but could we also list jQuery and the specific version number as a dependency on our npm package?

Unclear on why Bootstrap doesn't define its version in dependencies, but could you at least list the jQuery version as a `peerDependency`? That way there is at least a warning to use a certain version.

PR #30560, opened 2020-04-11
Add `karma-sabarivka-reporter` to ensure all source files under coverage

kopach:
This project is fantastic and it's unit test coverage on very high level. Therefore, I've decided to create this PR and improve it a little more.
By adding this simple plugin, developers/architects of the project can ensure no new functionality (as new source file) will be added to the project without testing it.

**Explanation**:
With `karma` and `istanbul` reporter, if someone simply forgets to create `feature.spec.js` file for `feature.js` file - coverage result won't include `feature.js` source file

### DIFF
--- a/js/tests/karma.conf.js
+++ b/js/tests/karma.conf.js
@@ -113,11 +113,18 @@ if (browserStack) {
     'karma-chrome-launcher',
     'karma-firefox-launcher',
     'karma-detect-browsers',
+    'karma-sabarivka-reporter',
     'karma-coverage-istanbul-reporter'
   )
-  reporters.push('coverage-istanbul')
+  reporters.push(
+    'sabarivka',
+    'coverage-istanbul'
+  )
   conf.customLaunchers = customLaunchers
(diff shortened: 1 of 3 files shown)


XhmikosR:
TBH I don't see any big difference in our case.

kopach:
@XhmikosR Yes, you're totally right. Currently - no difference at all. Anyway, reason for this is to automate one more thing and save time in the future, e.g. during code review.

Johann-S:
I'm a bit mismatched here, I understand the need of this karma plugin, but unit test is a requirement for our JS code, so if there is a lack of unit test we'll see it during code review because we review unit tests too

kopach:
@Johann-S, OK, got it. Feel free to close or merge this PR. If you think, you don't need this change – fine with me. In this repo, as I see, it's very little chance to get such tests regression. This repo is tested very well. So, leaving this for you to decide what to do with this PR. Thanks.

XhmikosR:
Thanks for the PR but we are gonna pass for now. :)

PR #30561, opened 2020-04-11
v5: Update system font stack

### LINKED ISSUE #25604: Update system font stack
Chrome will soon deprecate it's `BlinkMacSystemFont` system font alias for `system-ui`, but there are issues with that, too. Need to dig in more and prep something for that.

##

X-ref: https://github.com/sindresorhus/modern-normalize/issues/18

mdo:
**Demo:** https://codepen.io/emdeoh/pen/oNjgOPy.
**caniuse.com data:** https://caniuse.com/#feat=font-family-system-ui

Definitely needs more testing here, but latest Chrome and Safari support it, Firefox still needs `-apple-system`, and legacy Edge/IE (not that we care) require explicit fonts later on.

---

Fixes #25604.


MartijnCuppens:
Shouldn't we drop `BlinkMacSystemFont` here?

mdo:
> Shouldn't we drop `BlinkMacSystemFont` here?

Yup! You're right, thanks for the callout. Verified our browser support list and compared, it's safe to remove.

MartijnCuppens (approved):

BryceAMcDaniel:
I'm likely out of my league here, but researching updating my font-stack I can across issues with system-ui as below.

https://github.com/twbs/bootstrap/pull/22377
Does this still apply?

mdo:
I think we’re in the clear but I don’t know for sure. We’ll find out in testing :).

fireattack:
Just a headup: #22377 (i.e. https://infinnie.github.io/blog/2017/systemui.html) is still an issue on Chinese Windows.

quiple:
The `system-ui` still issues problems on Korean as well as Chinese Windows. #6139

Issue #23709, opened 2017-08-28
Cursor pointer Utility 


lucasdavidferrero:
Hello, It will be nice if someone could add a class that let you point the cursor.
example:

```
.c-pointer {
 cursor: pointer;
}
```

andresgalante:
What does `c-pointer` means? is it a utility class?

lucasdavidferrero:
It means that the cursor will be pointer (cursor: pointer). Another example could be the following:

```
.c-wait {
cursor: wait;
}
```
or

```
.c-help {
 cursor: help;
 }
```

Those are utilities classes. I missed the dot in my first comment.

MartijnCuppens:
I like that idea. Maybe we can add all [cursor-values](https://www.w3schools.com/cssref/playit.asp?filename=playcss_cursor&preval=pointer)? And maybe add a [global option](https://getbootstrap.com/docs/4.0/getting-started/options/) to enable/disable these classes like the grid classes?

patrickhlauke:
there's a danger that we'd need to add utility classes for every single CSS property at that rate, though. is it really necessary to do this, or can authors not do their own minor CSS additions?

andresgalante:
I agree with @patrickhlauke, I don't like the idea. If we add a single utility for each CSS property it'd be like writing inline styles.

MartijnCuppens:
@patrickhlauke and @andresgalante, your comments make sense, these classes may indeed be more appropriate in custom stylesheets.

rbonomo:
I don't see a problem with adding to the Bootstrap utilities. Why have utility classes for border-radius and not cursor? Do you have a prejudice against the cursor property? I do not agree with the above comments

RavenHursT:
Judging by 👍's it would appear most people agree w/ you @rbonomo (myself included) as opposed to the project maintainers that chimed in here. 🤷‍♂️

Ismaestro:
> What does `c-pointer` means? is it a utility class?

omg, seriously?

RavenHursT:
There's a whole slew of CSS properties that Bootstrap has enumerated the values for and provided utility classes for, another being `.d-{value}`: https://getbootstrap.com/docs/4.1/utilities/display/

It's great to see someone has reopened this. 💯

patrickhlauke:
"There's a whole slew of CSS properties..." ... so let's strive to aim all the other ones too?

RavenHursT:
[...]

PR #30562, opened 2020-04-11
v4: Add interaction utils

mdo:
This might be overstepping things for our next v4 release, but I noticed how highly requested the `.cursor-pointer` utility was in #23709 and couldn't help but stub out a PR to address it. I've taken the v5 approach of an interactions page and included this new util, and the `user-select` one from v5. Decided not to include `pointer-events` utils because of browser compatibility.

I'd really like to ship small enhancements to v4 for folks with highly requested features like this, especially if it helps bridge the gap between v4 and v5. My biggest question is how we version/release this though.

Do we squeeze this into a v4.4.2 despite of it being a new feature and thus requiring a semver minor bump? Or do we upgrade v4.4.2 to v4.5?


XhmikosR:
Squashed the patches.

TBH I'd prefer it if we didn't make it a new minor release. On the other hand, semver...

Let's wait for the others to chime in.

MartijnCuppens:
> I'd prefer it if we didn't make it a new minor release

I agree

mdo:
Revisiting this, if @twbs/css-review is into it, I'm inclined to sneak this into v4.4.2 to help close the gap between v4 and v5. Really think this will be some solid good will for our users :).

XhmikosR:
We never followed semver 100% and this, at least, is not a breaking change, so I think we can add it.

MartijnCuppens:
I'm gonna stick with what I said in https://github.com/twbs/bootstrap/pull/30563#issuecomment-613005377. I don't want people to abuse this property.

What I am willing to consider is adding
```
[role="button"] {
  cursor: pointer;
}
```
to reboot since this will be required anyway.

XhmikosR:
That's a valid concern of course, but on the other hand, one can buy a gun legally (in some countries); if they kill someone the store cannot be held liable for any murders. 😇

But I definitely see your point. Maybe if we added a warning callout it'd be better?

ffoodd:
I'd be on both sides on this one: 
- I agree with @MartijnCuppens about `cursor`s and don't think it should be a utility at all. I already know tons of people who'd just add a `div` with `.btn` things and `.cursor-pointer`—and it'll probably be bad. `[role="button"]` seems more accurate.
- `user-select` on the other hand seems handy, I don't see any weird use cases (but I may be a bit naive on this).

mdo:
Updated to reflect the team's feedback! Thanks y'all.

ffoodd:
Shouldn't `[role="button"]` sit in reboot, as suggested by @MartijnCuppens? I think it shouldn't be considered as a utility since it needs accessibility and scripting considerations to ne used the right way.

Maybe even provide a sentence with explanation or link?

scss/utilities/_interactions.scss
```diff
@@ -0,0 +1,9 @@
+// stylelint-disable declaration-no-important
+
+@each $value in $user-selects {
+  .user-select-#{$value} { user-select: $value !important; }
+}
+
+[role="button"] {
+  cursor: pointer;
+}
```
ffoodd, line 9: I think this should sit in `_reboot.scss`, and not being encouraged.

site/docs/4.4/utilities/interactions.md
```diff
@@ -0,0 +1,30 @@
...
+## Cursors
+
+Change the cursor from a default value to `pointer` with our built-in attribute utility, `role="button"`. Add this attribute to an element to convey semantics to your visitors and add an explicit pointer cursor to help indicate elements are interactive.
+
+{% capture example %}
+<span role="button">Non-button element button</span>
+{% endcapture %}
+{% include example.html content=example %}
+
+Customize the available classes by changing the `$cursors` Sass list in `_variables.scss`.
```
ffoodd, line 30: Since this part only mentions `[role="button"]`, I'd be in favor of dropping it totally.

Using this `role` has multiple prerequisite, like ensuring the interaction layer respects the `button` pattern: activation with <kbd>Space</kbd> and <kbd>Enter</kbd>, at least. This requires JavaScript and has very strong accessibility concerns: I wouldn't recommend to use this attribute to change the `cursor`.

Adding this `cursor` thing to our reboot seems good to prevent people from adding it manually… But this is more a bad practice than anything, and a `button` element should be used instead.We cannot recommend this.

BTW the last sentence is outdated anyway, isn't it?

ffoodd (changes requested):

mdo:
Think I got the feedback applied @MartijnCuppens and @ffoodd. Let me know with any suggested changes or review comments if there's something I can clarify more in the docs <3.

ffoodd (approved):
LGTM. We'll probably add some warning about the button role usage, but I think it's more @patrickhlauke's business here.

MartijnCuppens (approved):
:shipit:

XhmikosR:
FYI the utilities/interactions.scss wasn't included in utilities.scss. I fixed this in my v4-dev PR #30768.

PR #30563, opened 2020-04-11
v5: Add cursor util

### LINKED ISSUE #23709: Cursor pointer Utility

mdo:
v5 version of #30562. Fixes #23709.

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -443,6 +443,11 @@ $utilities: map-merge(
       class: font,
       values: (monospace: var(--font-family-monospace))
     ),
+    "cursor": (
+      property: cursor,
+      class: cursor,
+      values: pointer
+    ),
     "user-select": (
       property: user-select,
       values: all auto none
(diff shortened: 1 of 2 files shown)


MartijnCuppens:
Gonna stick with my opinion in https://github.com/twbs/bootstrap/issues/23709#issuecomment-571290182. A lot of people want this, but no one seems to be able to define a use case for the `cursor-pointer` utility.

I'm afraid developers will just abuse this utility and add this class to elements which aren't focusable with keyboard navigation.

mdo:
> A lot of people want this, but no one seems to be able to define a use case for the `cursor-pointer` utility.

Good point, just commented again on that issue.

ffoodd:
I agree with @MartijnCuppens here. Cursors are usually good natively, and a very precious indicator of browser and OS interaction layer.

With these utilities, it'd be too easy to mess things up. If something is wrong with our cursors, those styles should belong to reboot.

Pointer events and user-select can be really helpful since they affect behaviour, but cursors can be really misleading.

mdo:
Coming back to this after shipping the interactions and `role="button"` in v4.5, thinking this needs to be updated to include that instead, right?

mdo:
Replaced by #31040.

PR #30548, opened 2020-04-10
Ensure  to pass 4.5:1 contrast ratio everywhere

ffoodd:
Sounds like dependabot one, but isn't :)

Paving the way in #30468 — and to keep going with #29315 : my goal is to make this happen after tackling every obstacles.

## To solve before the bump

Those should probably adressed separately, consider this as a roadmap to enlightment :D 

- [x] #30989 → [`.btn-success:hover` case mentionned in #30468 (comment)](https://github.com/twbs/bootstrap/pull/30468#issuecomment-605419628);
  - #30972 is maybe the wrong direction too—there mght be no need to change colors at all.
  - #31276 did it :tada: 
- [x] #30622 → `color-level()`, `lighten()` and `darken()` seem misused: either on `:hover` for buttons (always darken) or to define `color` and `background-color` with the same level, no matter what the base color is (eg. `.alert-light` seems wrong) — **there's probably room to improvement by using `tint()` and `shade()` here**;
  - Relates to #30487 
- [x] #30550 → `.text-info`, `.text-black-50` and `.text-white-50` can't make it to 4.5:1 — however, **using `$gray-900` instead of `$gray-800` for `$dark` would solve this** (and several other dark-ish things with insufficient contrasts);
- [ ] A few components use **light grays as background** (mostly in cards and navbars) without accounting for it in their color usage, resulting in insufficient contrasts **when used with `.text-muted` or `.disabled` things**;
- [ ] Some `.btn-outline-*` and `.link-*` examples **should be shown with dark background**:
  1. #30044 will help to associate a contrast color for the docs part, through the data layer
  2. #30551 was a wrong direction but helped to define a visual consideration for examples with multiple backgrounds
  3. **TODO** will need another PR after #30044 to tackle this, I guess.

Merging this would be a kind of consecration :tada: 

Fixes #25126

PR #30564, opened 2020-04-11
v5: Unify spacing across components

### LINKED ISSUE #25057: Inconsistent component default/spacer padding

mdo:
- Updates horizontal padding across alerts, cards, dropdowns, list groups, and modals to be our default `$spacer` of `1rem`
- Reassigns some variables to improve ease of customization
- ~~Restyles the breadcrumb a smidge to make the background lighter for improved contrast and a little less vertical padding~~

Fixes #25057 for v5.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -232,8 +232,8 @@ $enable-important-utilities:  true !default;
 $spacer: 1rem !default;
 $spacers: (
   0: 0,
-  1: $spacer * .25,
-  2: $spacer * .5,
+  1: $spacer / 4,
+  2: $spacer / 2,
   3: $spacer,
   4: $spacer * 1.5,
   5: $spacer * 3,
@@ -916,8 +916,8 @@ $dropdown-link-active-bg:           $component-active-bg !default;
 
 $dropdown-link-disabled-color:      $gray-600 !default;
 
-$dropdown-item-padding-y:           .25rem !default;
-$dropdown-item-padding-x:           1.5rem !default;
+$dropdown-item-padding-y:           $spacer / 4 !default;
+$dropdown-item-padding-x:           $spacer !default;
 
 $dropdown-header-color:             $gray-600 !default;
 $dropdown-header-padding:           $dropdown-padding-y $dropdown-item-padding-x !default;
@@ -957,22 +957,22 @@ $pagination-disabled-border-color:  $gray-300 !default;
 
 // Cards
 
-$card-spacer-y:                     1.25rem !default;
-$card-spacer-x:                     1.25rem !default;
-$card-title-spacer-y:               .75rem !default;
+$card-spacer-y:                     $spacer !default;
+$card-spacer-x:                     $spacer !default;
+$card-title-spacer-y:               $spacer / 2 !default;
 $card-border-width:                 $border-width !default;
 $card-border-radius:                $border-radius !default;
 $card-border-color:                 rgba($black, .125) !default;
 $card-inner-border-radius:          subtract($card-border-radius, $card-border-width) !default;
-$card-cap-padding-y:                .75rem !default;
-$card-cap-padding-x:                1.25rem !default;
+$card-cap-padding-y:                $card-spacer-y / 2 !default;
+$card-cap-padding-x:                $card-spacer-x !default;
 $card-cap-bg:                       rgba($black, .03) !default;
 $card-cap-color:                    null !default;
 $card-height:                       null !default;
 $card-color:                        null !default;
 $card-bg:                           $white !default;
 
-$card-img-overlay-padding:          1.25rem !default;
+$card-img-overlay-padding:          $spacer !default;
 
 $card-group-margin:                 $grid-gutter-width / 2 !default;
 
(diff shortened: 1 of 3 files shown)


MartijnCuppens:
The spacer changes LGTM.

> Restyles the breadcrumb a smidge to make the background lighter for improved contrast and a little less vertical padding

Should be in another PR, or at least in a separate commit. Contrast ratio still doesn't meet 4.5 (4.26), but it's an improvement.

mdo:
> > Restyles the breadcrumb a smidge to make the background lighter for improved contrast and a little less vertical padding
> 
> Should be in another PR, or at least in a separate commit. Contrast ratio still doesn't meet 4.5 (4.26), but it's an improvement.

Sounds good, I can re-split this shortly. I recall requests for removing the background and padding entirely on the breadcrumb—that might satisfy the color contrast issue then. Thoughts?

MartijnCuppens:
> Sounds good, I can re-split this shortly. I recall requests for removing the background and padding entirely on the breadcrumb—that might satisfy the color contrast issue then. Thoughts?

I'm ok with that. tbh, I regularly need to remove the background and paddings. The background, padding and border radius can be achieved with utilities anyway

scss/_variables.scss
```diff
@@ -1078,8 +1078,8 @@ $modal-scale-transform:             scale(1.02) !default;
 //
 // Define alert colors, border radius, and padding.
 
-$alert-padding-y:                   .75rem !default;
-$alert-padding-x:                   1.25rem !default;
+$alert-padding-y:                   $spacer * .75 !default;
```
ffoodd, line 1081: Why keeping an exotic value here? `.5rem` or `1rem` looks good too, don't they?
This would be the last value out of the spacing scale.

ffoodd:
The versions dropdown's current item's `background` is a bit stuck in `.5rem` :)

And for breadcrumbs, I strongly agree with @MartijnCuppens (and will probably try to get off every light gray background as of #30548).

Apart from this, I love spacing consistency!

scss/_variables.scss
mdo: Vertical padding I care less about syncing up, but will look at this again to see how they look when cut down to `.5` or up to `1`.

mdo: Done! Went with `1rem`. Thanks!

MartijnCuppens:
The left padding added by the `ol` should be removed from the breadcrumbs.

I also realised there is no `.75rem` padding utility, so maybe we should keep the paddings configurable (0 by default) and `null`-ify the once that are removed now. 

Also the the breadcrumb looks so "naked" now: https://deploy-preview-30564--twbs-bootstrap.netlify.com/docs/4.3/components/breadcrumb/

MartijnCuppens:
Maybe we should cover the breadcrumb changes in another PR, @mdo?

mdo:
Removed breadcrumb restyling, updated the commits, and fixed the docs versions dropdown. Ready for review again!

scss/_variables.scss
```diff
@@ -916,8 +916,8 @@ $dropdown-link-active-bg:           $component-active-bg !default;
 
 $dropdown-link-disabled-color:      $gray-600 !default;
 
-$dropdown-item-padding-y:           .25rem !default;
-$dropdown-item-padding-x:           1.5rem !default;
+$dropdown-item-padding-y:           $spacer / 4 !default;
```
ffoodd, line 919: Any reason not to use `map-get($spacers, 1)` here — and eventually for each `$spacer / 2` , `*.25`, `*.5` etc.?

MartijnCuppens: If the developer uses a different `$spacers` map where there is no `1` key, this won't work.

site/content/docs/4.3/migration.md
```diff
@@ -125,6 +125,8 @@ Changes to Reboot, typography, tables, and more.
 
 ## Components
 
+- Unified `padding` values for alerts, breadcrumbs, cards, dropdowns, list groups, modals, popovers, and tooltips to be based on our `$spacer` variable.
```
MartijnCuppens, line 128:
```suggestion
- Unified `padding` values for alerts, breadcrumbs, cards, dropdowns, list groups, modals, popovers, and tooltips to be based on our `$spacer` variable. [See #30564](https://github.com/twbs/bootstrap/pull/30564).
```

MartijnCuppens (approved):
LGTM. Wonderful improvement, @mdo!

scss/_variables.scss
ffoodd: Oh right! Thanks :)

ffoodd (approved):

ffoodd:
The only thing to question is having both `$spacer / 2` and `$spacer * .5`: we may normalize this :)

MartijnCuppens:
@ffoodd, good call, fixed that in the latest rebase

PR #30565, opened 2020-04-12
adds back `.form-text`

### LINKED ISSUE #30477: Bring back `.form-text`
In the notorious form updates PR (https://github.com/twbs/bootstrap/pull/28450), `.form-text` was dropped. I don't think that was the right thing to do, having a class which we can use to keep form descriptions consistent would help. We could also add some `null` variables (font size, font weight, font style, color) to make this element more more configurable.

Thoughts @twbs/css-review?

zalog:
This PR adds back `.form-text` css class.

- deploy [preview](https://deploy-preview-30565--twbs-bootstrap.netlify.com/docs/4.3/forms/overview/)
- closes https://github.com/twbs/bootstrap/issues/30477

### DIFF
--- a/scss/forms/_form-text.scss
+++ b/scss/forms/_form-text.scss
@@ -5,4 +5,5 @@
 .form-text {
   display: block;
   margin-top: $form-text-margin-top;
+  color: $form-text-color;
 }
--- a/scss/forms/_form-text.scss
+++ b/scss/forms/_form-text.scss
@@ -1,0 +1,8 @@
+//
+// Form text
+//
+
+.form-text {
+  display: block;
+  margin-top: $form-text-margin-top;
+}
--- a/scss/forms/_form-text.scss
+++ b/scss/forms/_form-text.scss
@@ -5,5 +5,6 @@
 .form-text {
   display: block;
   margin-top: $form-text-margin-top;
+  @include font-size($form-text-font-size);
   color: $form-text-color;
 }
--- a/scss/forms/_form-text.scss
+++ b/scss/forms/_form-text.scss
@@ -6,5 +6,6 @@
   display: block;
   margin-top: $form-text-margin-top;
   @include font-size($form-text-font-size);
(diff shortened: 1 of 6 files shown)


zalog:
@MartijnCuppens I would like to add a `$form-text-color: $text-muted !default;` too, what do you think?

Also, what should we do about docs? I think we should remove references about _`.help-block` in v3_ that comes from an older version.

MartijnCuppens:
> @MartijnCuppens I would like to add a $form-text-color: $text-muted !default; too, what do you think?

Yup! A font size variable may also be appropriate, maybe a `null` font-style variable.

> Also, what should we do about docs?

Use `.form-text` where needed.

> I think we should remove references about .help-block in v3, text that comes from an older version docs.

Yup.

zalog:
Tnks for feedback!

Updated & done. What do you think?

scss/forms/_form-text.scss
```diff
@@ -0,0 +1,10 @@
+//
+// Form text
+//
+
+.form-text {
+  display: block;
+  margin-top: $form-text-margin-top;
+  font-size: $form-font-size;
```
MartijnCuppens, line 8: We should use the font-size mixin here:
```suggestion
  @include font-size($form-font-size);
```

I thought we disallowed `font-size` properties at some point, gotta check since when we didn't blacklist this property anymore.

zalog: Oh yeah!
I'm not yet updated, but getting there :D
Just a sec.

zalog: Done. Also rebased and squash with the relevant commit.

MartijnCuppens:
Could you add a `font-style` (default: `null`)?

zalog:
Yep, just a sec.
Also I saw a glitch there on a var naming.

zalog:
Done

mdo (approved):

zalog:
Rebased

PR #30568, opened 2020-04-13
Redo display headings

### LINKED ISSUE #26515: More display-* headings - .display-5

mdo:
- Convert to Sass map
- Consolidate font-weights into single option (matching line-height for them)
- Add two new sizes and create a new scale
- Update migration docs

Fixes #26515 and closes #28453.

Demo: https://deploy-preview-30568--twbs-bootstrap.netlify.com/docs/4.3/content/typography/#display-headings


MartijnCuppens:
Maybe we should add the `$display-font-sizes` map to the docs with `scss-docs`?

MartijnCuppens (approved):

PR #30571, opened 2020-04-13
v5: Add .fs-* utilities for font-size and rename font-weight/-style utilities

### LINKED ISSUE #25832: Type scale map and type utility classes. (Font)

mdo:
- Adds new font-sizes Sass map
- Generates six new classes for setting only font-size
- Updates docs to mention this, including a scss-docs reference

It occurs to me this might need some `rfs` magic, but I'm unsure how to add that right now to the utilities file. Also, should these be responsive so folks can do `.fs-md-3 .fs-lg-4`? I'm thinking not given obvious file bloat.

Also makes me think `.font-weight-*` utilities should be shortened to `.fw-*`. `.font-style` is obviously trickier... could remap to `.text-*` or `.font-*`?

Fixes #25832.

https://deploy-preview-30571--twbs-bootstrap.netlify.app/docs/5.0/utilities/text/#font-size

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -395,7 +395,7 @@ $utilities: map-merge(
     ),
     "font-style": (
       property: font-style,
-      class: f,
+      class: fst,
       values: italic normal
     ),
     "font-weight": (
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -387,6 +387,11 @@ $utilities: map-merge(
       values: $spacers
     ),
     // Text
+    "font-size": (
+      property: font-size,
+      class: fs,
+      values: $font-sizes
+    ),
     "font-weight": (
       property: font-weight,
       values: (
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -399,7 +399,7 @@ $utilities: map-merge(
       values: italic normal
     ),
     "font-weight": (
-      property: fw,
+      property: font-weight,
       values: (
         light: $font-weight-light,
         lighter: $font-weight-lighter,
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -388,12 +388,18 @@ $utilities: map-merge(
     ),
     // Text
     "font-size": (
+      rfs: true,
       property: font-size,
       class: fs,
       values: $font-sizes
     ),
+    "font-style": (
+      property: font-style,
+      class: f,
+      values: italic normal
+    ),
     "font-weight": (
-      property: font-weight,
+      property: fw,
       values: (
         light: $font-weight-light,
         lighter: $font-weight-lighter,
@@ -466,11 +472,6 @@ $utilities: map-merge(
     "text-decoration": (
       property: text-decoration,
       values: none underline line-through
-    ),
-    "font-style": (
-      property: font-style,
-      class: font,
-      values: italic normal
     ),
     "word-wrap": (
       property: word-wrap word-break,
--- a/site/content/docs/5.0/migration.md
+++ b/site/content/docs/5.0/migration.md
@@ -25,6 +25,13 @@ toc: true
 
 - The longstanding [Missing border radius on input group with validation feedback bug](https://github.com/twbs/bootstrap/issues/25110) is finally fixed by adding an additional `.has-validation` class to input groups with validation.
 
+### Utilities
+
+- **Text utilities:**
+  - Added `.fs-*` utilities for `font-size` utilities (with RFS enabled). These use the same scale as HTML's default headings (1-6, large to small), and can be modified via Sass map.
+  - Renamed `.font-weight-*` utilities as `.fw-*` for brevity and consistency.
+  - Renamed `.font-style-*` utilities as `.fst-*` for brevity and consistency.
+
 ## v5.0.0-alpha2
 
 ### Sass
@@ -138,6 +145,9 @@ toc: true
   - Some great examples have been added to the docs to show these off.
 - [#31484](https://github.com/twbs/bootstrap/pull/31484): Added new [`border-width` utility]({{< docsref "/utilities/borders#border-width" >}}).
 - [#31473](https://github.com/twbs/bootstrap/pull/31473): The `.d-none` utility was moved in our CSS to give it more weight over other display utilities.
+   Renamed `.text-monospace` to `.font-monospace`.
+- Removed `.text-hide` as it's an antiquated method for hiding text that shouldn't be used anymore.
+- New `line-height` utilities: `.lh-1`, `.lh-sm`, `.lh-base` and `.lh-lg`. See [here]({{< docsref "/utilities/text#line-height" >}}).
 
 ---
 
@@ -323,15 +333,6 @@ Badges were overhauled to better differentiate themselves from buttons and to be
 ### Grid
 
 - Decreased the number of responsive order utilities per breakpoint. The highest order utility with a number now is `.order-5` instead of `.order-12`. [See #28874](https://github.com/twbs/bootstrap/pull/28874).
-
-### Fonts and text
-
-- Added `.fs-*` utilities for `font-size`.
-- Renamed `.font-weight-*` utilities as `.fw-*` for brevity.
-- Renamed `.font-style-*` utilities as `.fst-*` for brevity.
-- Renamed `.text-monospace` to `.font-monospace`.
-- New `line-height` utilities: `.lh-1`, `.lh-sm`, `.lh-base` and `.lh-lg`. See [here]({{< docsref "/utilities/text#line-height" >}}).
-- Removed `.text-hide` as it's an antiquated method for hiding text that shouldn't be used anymore.
 
 ### Misc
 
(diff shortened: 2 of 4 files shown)


MartijnCuppens:
> It occurs to me this might need some `rfs` magic, but I'm unsure how to add that right now to the utilities file. 

Just add `rfs: true` to the utility map, and we should be fine.

> Also, should these be responsive so folks can do `.fs-md-3 .fs-lg-4`? I'm thinking not given obvious file bloat.

Since RFS already rescales the font sizes, I don't think there's any need for this.

> Also makes me think .font-weight-* utilities should be shortened to .fw-*. .font-style is obviously trickier... could remap to .text-* or .font-*?

Yup, `fw` seems appropriate, `text-*` might be confusing, especially for `font-weight: normal`.

---

I think it can be quite confusing that the margin of `m-2` is larger than `m-1`, but `fs-2` is smaller than `fs-1`. Also, if you want to add font sizes larger than `fs-1`, you'll need to add `fs-0`, which doesn't really make sense.

It might be better to work with `sm/md/lg` modifiers like this:

```scss
$font-sizes: (
  "xs": .75rem,
  "sm": .875rem,
  "md": 1rem,
  "lg": 1.25rem,
  "xl": 1.5rem,
  "xxl": 2rem
) !default;
```

scss/_utilities.scss
```diff
@@ -358,8 +358,19 @@ $utilities: map-merge(
...
     // Text
+    "font-size": (
+      rfs: true,
+      property: font-size,
+      class: fs,
+      values: $font-sizes
+    ),
+    "font-style": (
+      property: font-style,
+      class: f,
```
MartijnCuppens, line 369: I don't think `f-normal` is clear enough, maybe use the `fst` prefix (or something else?)

mdo: Yeah, playing with this for sure still. Tailwind does `.italic`/`.no-italic`, the latter of which makes me laugh a little bit haha. Wanted to keep it somewhat familiar and `.fst` does fit nice enough. Any other ideas?

MartijnCuppens: Not atm. `no-italic` feels very "unlike any other class".

mdo:
> It might be better to work with `sm/md/lg` modifiers like this:

I've been avoiding new classes that mix with our breakpoints notation to see if we can clear up "scales". I chose this approach since it maps 1:1 with our `font-size`s on headings. I know we did it with `line-height`, but maybe we change that? Hmm...

scss/_utilities.scss
```diff
@@ -358,8 +358,19 @@ $utilities: map-merge(
...
+      values: $font-sizes
+    ),
+    "font-style": (
+      property: font-style,
+      class: fst,
+      values: italic normal
+    ),
     "font-weight": (
-      property: font-weight,
+      property: fw,
```
ffoodd, line 373: Shouldn't `fw` be the class, and `property` stick to `font-weight`?

ffoodd:
Naming is pretty hard here, my two cents: 
1. if map keys are 1-6, why keep `$h*-font-size` variable at all: they could be replaced by a loop in our reboot.
2. if not, couldn't we handle a single map with all our font-sizes, from small to `display-1`—but I don't have any clue of an accurate naming for now -_-

We could stick with numbers (like spacers) but it wouldn't match headings hierarchy anymore: but should it?

Thinking out loud for now, but there's more to be done I guess.

mdo:
Going to hold on this for now with that last comment.

I'm wondering if we can build a single type scale system (GitHub/Primer does `h000`, `h00`, and `h0`) to go with all our type sizes and then nix the display classes. Unsure if it makes sense to change more classes or just add on.

mdo:
Decided against the above. Let's leave the display classes as-is. v6 can tackle a single set of font-sizes if we still want it.

Just rebased and resolved the conflicts here. This should be ready for review now!

site/content/docs/5.0/migration.md
```diff
@@ -246,12 +246,19 @@ Badges were overhauled to better differentiate themselves from buttons and to be
...
 ### Utilities
 
-- Renamed `.text-monospace` to `.font-monospace`
+### Grid
 - Decreased the number of responsive order utilities per breakpoint. The highest order utility with a number now is `.order-5` instead of `.order-12`. [See #28874](https://github.com/twbs/bootstrap/pull/28874).
+
+### Fonts and text
+- Added `.fs-*` utilities for `font-size`.
+- Renamed `.font-weight-*` utilities as `.fw-*` for brevity.
+- Renamed `.font-style-*` utilities as `.f-*` for brevity.
```
ffoodd, line 255: Your map sets `fst` as a class name for `font-style`.

MartijnCuppens: Fixed that

MartijnCuppens (approved):

XhmikosR:
@mdo this needs to move the migration section to alpha3 :)

site/content/docs/5.0/migration.md
```diff
@@ -7,6 +7,17 @@ aliases: "/migration/"
...
+## v5.0.0-alpha3
+
+### Fonts and text
+
+- Added `.fs-*` utilities for `font-size`.
+- Renamed `.font-weight-*` utilities as `.fw-*` for brevity.
+- Renamed `.font-style-*` utilities as `.fst-*` for brevity.
+- Renamed `.text-monospace` to `.font-monospace`.
+- New `line-height` utilities: `.lh-1`, `.lh-sm`, `.lh-base` and `.lh-lg`. See [here]({{< docsref "/utilities/text#line-height" >}}).
+- Removed `.text-hide` as it's an antiquated method for hiding text that shouldn't be used anymore.
```
XhmikosR, line 19: I think these changes belong to alpha2?

site/content/docs/5.0/utilities/text.md
```diff
@@ -63,18 +63,35 @@ Transform text in components with text capitalization classes.
...
-<p class="font-weight-lighter">Lighter weight text (relative to the parent element).</p>
-<p class="font-italic">Italic text.</p>
-<p class="font-normal">Text without font style</p>
+<p class="fw-bold">Bold text.</p>
+<p class="fw-bolder">Bolder weight text (relative to the parent element).</p>
+<p class="fw-normal">Normal weight text.</p>
+<p class="fw-light">Light weight text.</p>
+<p class="fw-lighter">Lighter weight text (relative to the parent element).</p>
+<p class="fst-italic">Italic text.</p>
+<p class="fst-normal">Text without font style</p>
```
XhmikosR, line 94: Maybe we should reword this? It does have a `font-style`, it's just set to `normal`.

stormitornado:
I propose to name new classes as shortcuts in emmet:
fz - font-size
fs - font-style
fw - font-weight

site/content/docs/5.0/migration.md
```diff
@@ -21,10 +21,22 @@ toc: true
 
 - Introduce `$enable-smooth-scroll`, which applies `scroll-behavior: smooth` globally—except for users asking for reduced motion through `prefers-reduced-motion` media query. [See #31877](https://github.com/twbs/bootstrap/pull/31877)
 
+<<<<<<< HEAD
```
MartijnCuppens, line 24: Oh no

Issue #26596, opened 2018-05-25
Use custom properties (CSS variables) in bootstrap


tobi-or-not-tobi:
I like sass variables and they're great during buildtime. But projects that need a microfrontend architecture (i.e. webcomponents) tend not to rebuild components. They rather reuse existing components and provide the style context to it. 

Using CSS variables will allow to provide runtime configuration for styling, such as colors. CSS variables can work in conjunction of SASS. 
I'm happy with a few the first couple of CSS variables in bootstrap, but as [already mentioned](https://github.com/twbs/bootstrap/issues/23349#issuecomment-346813716) they're only exposed and not used by bootstrap itself. This means, if a project likes to override these CSS variables at runtime, there's no effect with the bootstrap components themself. 

Is there anything against going into the CSS variable direction? [Browser support is really great](https://caniuse.com/#feat=css-variables) and the fallback option will keep non-supported browsers happy. 

I understand this is a big change though. It woudl work like this:

Intialize CSS vars. We use use double-double fallback to ensure we have a fallack and do not need to repeately add the fallback everywhere. 
```
:root {
    --primary: var(--primary-color, #{$primary});
}
```

Then for each and every component that uses the primary color, we could use the --primary variable instead. For example on a button, it would go like this (just a few lines that i'm using in my sandbox project, so its not so relevant to bootstrap sources):
```
.btn {
    &-primary {
        background-color: var(--primary);
        border-color: var(--primary);
        &:hover,
        &:not(:disabled):not(.disabled):active {
            background-color: var(--primary-darken);
            border-color: var(--primary-darken);
        }
    }
}
```

This would allow to change the color at runtime.

tobi-or-not-tobi:
Understood that support for IE10/11 still is important. What about introducing a fallback. We could have a mixin that would generate both the fallback as well as the custom property. 

```
@mixin customProperty($name, $value) {
  
    // fallback for older browser who do not support custom properties / css variables
    #{$name}: #{$value};

    // create a css variable (this might override the variable on any of the ancestor elements)
    --#{$name}: #{$value};

    // use the css variable on the property
    #{$name}: var(--#{$name}, #{$value});
}
```

Alternatively we could use postcss but I'm personally not so much in favor of that since it doesn't allow to use those properties outside the root scope.

MartijnCuppens:
Could you provide a demo of how you would use that mixin?

tobi-or-not-tobi:
Hi @MartijnCuppens,
[...]

PR #30575, opened 2020-04-13
Documenting CSS vars usage in upcoming Customize docs section

mdo:
WIP, but trying to capture some of the work @MartijnCuppens and @ffoodd are doing across multiple PRs. Fellas, please weigh in here on anything else we should be striving to mention. Some thoughts from me thus far mean we could include guidelines for how we decide what gets CSS vars, how to customize and utilize them, how we generate them with Sass, etc.

---

X-ref: #26596

### DIFF
--- a/site/content/docs/4.3/customize/css-variables.md
+++ b/site/content/docs/4.3/customize/css-variables.md
@@ -6,9 +6,9 @@ group: customize
 toc: true
 ---
 
-Bootstrap includes around two dozen [CSS custom properties (variables)](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) in its compiled CSS. These provide easy access to commonly used values like our theme colors and primary font stacks when working in your browser's Inspector, a code sandbox, or general prototyping.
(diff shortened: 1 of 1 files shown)


ffoodd (approved):

ffoodd:
LGTM. @MartijnCuppens?

MartijnCuppens:
> support dynamic restyling and extending of Bootstrap components without recompiling Sass

I'm not sure about this. You can't just convert all Sass variables into custom properties.

For example, the buttons. They have `:hover`, `:focus` & `:active` states. These states use colors that are generated by Sass and if you change for example `var(--primary)`, those colors won't change. Likewise, contrast colors won't automatically adjust.

As long as we don't have native CSS color functions, this idea won't work.

mdo:
I'll fine-tune that copy and make sure I'm not implying something we cannot technically do.

mdo:
Pulled back on the customization/extension bits... I think this is okay? Doesn't overpromise and clarifies some other pieces. Also removed buttons from the last line there.

MartijnCuppens (approved):
Beautiful.

PR #30582, opened 2020-04-14
Delete appearance from `button.close`

k-utsumi:
ref.
https://github.com/twbs/bootstrap/blob/master/scss/_reboot.scss#L453

Added log
https://github.com/twbs/bootstrap/commit/473e5714f756d9be81e9321d451bb05431dd28ee#diff-de295610e4369f044d82fd2bcf812a8f

### DIFF
--- a/scss/_close.scss
+++ b/scss/_close.scss
@@ -33,5 +33,4 @@ button.close {
   padding: 0;
   background-color: transparent;
   border: 0;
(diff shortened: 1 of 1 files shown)


MartijnCuppens (approved):
We don't need that indeed, `.btn`s do not have an appearance either.

PR #30585, opened 2020-04-14
Disable negative margins by default

MartijnCuppens:
Negative margins were introduced to create/hack responsive gutter widths in BS 4.2. But since we now have the responsive gutter, those classes are now less applicable. Also, since we now have an `xxl` breakpoint, our file size has increased quite a lot, since there are lots of negative margins.

Negative margins can be useful, but I don't think we should enable them by default, therefore using a `$enable-` variable feels like a good idea.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -214,7 +214,7 @@ $enable-grid-classes:                         true !default;
 $enable-pointer-cursor-for-buttons:           true !default;
 $enable-rfs:                                  true !default;
 $enable-validation-icons:                     true !default;
-$enable-negative-margins-utilities:           false !default;
+$enable-negative-margins:                     false !default;
 $enable-deprecation-messages:                 true !default;
 $enable-important-utilities:                  true !default;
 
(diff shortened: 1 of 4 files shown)


scss/_variables.scss
```diff
@@ -214,6 +214,7 @@ $enable-grid-classes:                         true !default;
 $enable-pointer-cursor-for-buttons:           true !default;
 $enable-rfs:                                  true !default;
 $enable-validation-icons:                     true !default;
+$enable-negative-margins-utilities:           false !default;
```
mdo, line 217: Is `$enable-negative-margins` enough? Trying to cut down the length of these as for some reason it bothers the hell outta me. 😬

MartijnCuppens: It's less clear, but should be fine, especially since we document these variables

mdo (approved):

XhmikosR:
@mdo @MartijnCuppens maybe we should reconsider this patch? I mean, I know it saves us a lot from the dist files, but it's something that we have in v4 and it seems like a good thing to have by default.

Assuming we'll proceed with using CSS vars more, the dist size will drop anyway.

MartijnCuppens:
@XhmikosR, opened the discussion here: https://github.com/twbs/bootstrap/discussions/30649

PR #30587, opened 2020-04-15
Remove .btn-block for utilities

### LINKED ISSUE #30223: Add support for block buttons based on specific screen widths, eg. btn-xs-block

mdo:
Alternate fix for #30376, which closes #30223.

### DIFF
--- a/site/content/docs/4.3/migration.md
+++ b/site/content/docs/4.3/migration.md
@@ -129,6 +129,10 @@ Badges were overhauled to better differentiate themselves from buttons and to be
 - **Todo:** Removed `.badge-pill` for the `.rounded-pill` utility class
 - **Todo:** Removed badge's hover and focus styles for `a.badge` and `button.badge`.
 
+### Buttons
+
+- Removed `.btn-block` as it duplicated functionality of utilities, but in a more limited way. With a utilities based approach, "block buttons" are now easily made responsive.
+
 ### Cards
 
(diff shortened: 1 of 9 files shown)


site/content/docs/4.3/migration.md
```diff
@@ -129,6 +129,10 @@ Badges were overhauled to better differentiate themselves from buttons and to be
 - **Todo:** Removed `.badge-pill` for the `.rounded-pill` utility class
 - **Todo:** Removed badge's hover and focus styles for `a.badge` and `button.badge`.
 
+### Buttons
+
+- Removed `.btn-block` as it duplicated functionality of utilities, but in a more limited way. With a utilities based approach, "block buttons" are now easily made responsive.
```
XhmikosR, line 134: I think we should mention the exact replacement classes to make things easier for people 🙂

MartijnCuppens:
I don't think we need the `.d-block` class, only `.w-100` should be ok. We can even ditch the `mb-2`, since the inline blocks already have some spacing.

mdo:
Might not go this route after all actually... requires more utility classes to _undo_ the responsive blocks (assuming block on mobile, not block on say medium). 🤔

mdo:
Closing per my last comment.

PR #30588, opened 2020-04-15
Fix for sliding checkmark

Sojaner:
If the $form-check-input-width variable is changed and therefore the size of the checkbox or the radio button is different, the transition for the background-position will make the checkmark or the radio circle to slide a bit while appearing on-checked.
This change will fix the position to the center to prevent the undesirable sliding effect.

### DIFF
--- a/scss/forms/_form-check.scss
+++ b/scss/forms/_form-check.scss
@@ -20,6 +20,7 @@
   margin-top: ($line-height-base - $form-check-input-width) / 2; // line-height minus check height
   vertical-align: top;
   background-color: $form-check-input-bg;
+  background-position: $form-check-input-checked-bg-position;
   background-repeat: no-repeat;
   background-position: center;
   background-size: contain;


XhmikosR:
https://github.com/twbs/bootstrap/blob/master/.github/CONTRIBUTING.md#checking-coding-style

mdo:
You've duplicated the existing property and it's unclear to me the bug you're running into without a live demo.

MartijnCuppens:
This won't fix this issue in any case, you're adding a property to a wrong selector.

Sojaner:
I'm trying to figure out what's going on. This is the master branch and I have cloned the master branch also, but I can see on the document pages that it's actually the v5.0, so this might be invalid for the actual current version, but it's a bug in the version 5.0 @MartijnCuppens @mdo
image

Sojaner:
For technical difficulties, I couldn't push from my local, that's why I tried to replicate the fix from Github interface itself and that's where it got the problem from.

PR #30591, opened 2020-04-15
Add `bs-` prefixes to custom properties

### LINKED ISSUE #30558: Prefix custom properties with `bs-`
Something @tmorehouse mentioned in https://github.com/twbs/bootstrap/pull/30466#issuecomment-612177063, we could prefix our custom properties with `bs-`.

Thoughts @twbs/css-review?

MartijnCuppens:
Closes https://github.com/twbs/bootstrap/issues/30558.

Also shortened the `font-family` custom properties a bit, since we also ditch the `family` in the class names. Also did this for the gutter variables, just to save some bytes.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -283,7 +283,7 @@ pre,
 code,
 kbd,
 samp {
-  font-family: var(--font-family-monospace);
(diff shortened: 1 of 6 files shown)


ffoodd (approved):
Well done :)

PR #30596, opened 2020-04-15
Remove negative margins leftovers

MartijnCuppens:
... and fix the header height.


XhmikosR:
It might still be worth reverting #30585, though. We should discuss this further.

XhmikosR (approved):

MartijnCuppens:
> It might still be worth reverting #30585, though. We should discuss this further.

It would be easy to restore them if needed. I don't expect these were widely used since we only used them in responsive gutter example.

XhmikosR:
@mdo WDYT?

mdo (approved):
I say :shipit:!

PR #30597, opened 2020-04-15
Replace event.which + remove event.delegateTarget + fix crash

tkrotoff:
Following https://github.com/twbs/bootstrap/issues/19991#issuecomment-606050071

3 commits

## Replace event.which

[KeyboardEvent.which](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) and [MouseEvent.which](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which) are deprecated/non standard.

It's a relic from the jQuery time.

Replaced with [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) and [MouseEvent.button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button).

## Remove event.delegateTarget inside event-handler.js

event.delegateTarget is only used by tooltip, replacing it with event.target works fine, thus no more fixEvent().

## Fix crash in dropdown: "Uncaught TypeError: Cannot read property 'focus' of undefined"

While working on this, I've noticed a crash when opening the dropdown and pressing ArrowUp for the first time. Crash already present here: https://twbs-bootstrap.netlify.com/docs/4.3/components/dropdowns/

https://github.com/twbs/bootstrap/blob/f0abe26b98ade4d6a2d0b234758a94ea33b85388/js/src/dropdown.js#L482

`items.indexOf(event.target) || 0` gives `-1 || 0` which returns... `-1`... See explanations ["Why are JavaScript negative numbers not always true or false?"](https://stackoverflow.com/q/3619797)

## Testing

All tests passing, manually tested (dropdown, tooltip, carousel, tooltip) under macOS with Chrome 81, Firefox 75 and Safari 13.1.

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -33,8 +33,8 @@ const DATA_KEY = 'bs.carousel'
 const EVENT_KEY = `.${DATA_KEY}`
 const DATA_API_KEY = '.data-api'
 
-const ARROW_LEFT_KEYCODE = 37 // KeyboardEvent.which value for left arrow key
-const ARROW_RIGHT_KEYCODE = 39 // KeyboardEvent.which value for right arrow key
+const ARROW_LEFT_KEY = 'ArrowLeft'
+const ARROW_RIGHT_KEY = 'ArrowRight'
 const TOUCHEVENT_COMPAT_WAIT = 500 // Time for mouse compat events to fire after touch
 const SWIPE_THRESHOLD = 40
 
@@ -342,12 +342,12 @@ class Carousel {
       return
     }
 
-    switch (event.which) {
-      case ARROW_LEFT_KEYCODE:
+    switch (event.key) {
+      case ARROW_LEFT_KEY:
         event.preventDefault()
         this.prev()
         break
-      case ARROW_RIGHT_KEYCODE:
+      case ARROW_RIGHT_KEY:
         event.preventDefault()
         this.next()
         break
--- a/js/src/dom/event-handler.js
+++ b/js/src/dom/event-handler.js
@@ -17,7 +17,6 @@ import { defaultPreventedPreservedOnDispatch } from './polyfill'
 const $ = getjQuery()
 const namespaceRegex = /[^.]*(?=\..*)\.|.*/
 const stripNameRegex = /\..*/
-const keyEventRegex = /^key/
 const stripUidRegex = /::\d+$/
 const eventRegistry = {} // Events storage
 let uidEvent = 1
@@ -94,11 +93,6 @@ function getEvent(element) {
 }
 
 function fixEvent(event, element) {
-  // Add which for key events
-  if (event.which === null && keyEventRegex.test(event.type)) {
-    event.which = event.charCode === null ? event.keyCode : event.charCode
-  }
-
   event.delegateTarget = element
 }
 
--- a/js/src/dom/event-handler.js
+++ b/js/src/dom/event-handler.js
@@ -92,13 +92,8 @@ function getEvent(element) {
   return eventRegistry[uid]
 }
 
-function fixEvent(event, element) {
-  event.delegateTarget = element
-}
-
 function bootstrapHandler(element, fn) {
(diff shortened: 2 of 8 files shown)


XhmikosR:
Can you please split the bugfix and make a new PR with that patch only?

As for this PR, I'm not so sure if we should keep using features that aren't available in most browsers. I mean, sure, we officially don't support some browsers, but with using ES3 features until v4, things worked decently in older browsers.

Just thinking out loud, but I think we can just continue with using more modern features.

* https://caniuse.com/#feat=keyboardevent-key
* https://caniuse.com/#feat=mdn-api_mouseevent_button

tkrotoff:
> features that aren't available in most browsers
> - https://caniuse.com/#feat=keyboardevent-key
> - https://caniuse.com/#feat=mdn-api_mouseevent_button

That is not true.

Here a demo: https://codepen.io/tkrotoff/pen/RwWrjer (switch to "debug" or "full" view on mobile).

I've tested all browsers under Android 8.1 on a real smartphone

## Android Chrome

Chrome


## Android Firefox

Firefox


## Android Opera

Opera


## Android Opera Mini

Opera Mini


## Android UC Browser

UC Browser


## Android QQ Browser

(downloaded from https://browser.qq.com/ not available on Google Play Store)

QQ Browser



As you can see event.key and event.button are always present. It happens that on mobile everything (keyCode, charCode, which...) is trash: https://stackoverflow.com/q/36753548

As you can see, all the alternative browsers are based on Chromium...

The only browser that do not respect the standard is IE ("Esc" instead of "Escape", "Up" instead of "ArrowUp"...):

(ES5 demo: https://codepen.io/tkrotoff/pen/xxwZYVd)

## Windows 7 IE 10

Windows 7 IE 10

## Windows 10 IE 11

Windows 10 IE 11

## Old Edge (v18 on Windows 10 here) does fine

Windows 10 Edge 18

tkrotoff:
> Can you please split the bugfix and make a new PR with that patch only?

That would be painful: I would have to rewrite the fix for event.which and then change it back to event.key (and you would need to review it 2 times). This PR is only 3 commits.

XhmikosR:
Not sure why it's painful since it's just 2 lines to change.

XhmikosR:
Anyway, LGTM. I'll wait for another approval before merging.

XhmikosR (approved):

Johann-S (approved):
Seems a good change to me 👍

nlemoine:
Salut @Johann-S ! 

I think you might reconsider the removal of `event.delegateTarget`. 

Providing an event delegation function without providing the `delegateTarget` seems awkward to me. This means you have to look by yourself for the `delegateTarget`. 

Consider the following example using the latest version of BS: https://codepen.io/hellonico/pen/JjYwMPa?editors=1011

The `event.target` will never be the `a` tag which is what we're probably looking for. You can still use CSS to avoid this by setting `pointer-events: none` on the `svg` element but I don't think it's a good solution. 

What do you think?

nlemoine:
Another example demonstrating the need for `event.delegateTarget`: https://codepen.io/hellonico/pen/PoPXEJz?editors=1111

(hover the icon)

Tooltip won't work if the trigger has nested elements.

XhmikosR:
> BS now provides modules such as `SelectorEngine`, `EventHandler`, etc. that could be used without any BS components (modal, carousel, etc.).

You are not supposed to use these, they are for private use only.

As for the rest, please wait for @Johann-S to have a look.

nlemoine:
Thanks for the prompt reply!

> You are not supposed to use these, they are for private use only.

Private or not, these modules are exported. They can be used and actually behave just like any [other library](https://www.npmjs.com/package/delegated-events) (a quite popular event delegation lib for example). 

However, event they are not supposed to be used directly. There's still an issue about `delegateTarget` 😉

tkrotoff:
Re-introduce delegateEvent here: #30928

PR #30598, opened 2020-04-15
feat: removes `.form-text` display

### LINKED ISSUE #30577: `.form-text` follow-up
Follow up on https://github.com/twbs/bootstrap/pull/30565

- We use `<small>` in our docs, which might not always be correct, the font size is not smaller per se.
- We still use `text-muted` in the horizontal form here: https://twbs-bootstrap.netlify.com/docs/4.3/forms/overview/#form-text
- We should share properties between `.form-text`, `.valid-feedback` and `.invalid-feedback`

I think we better remove `display: block` from `.form-text` and replace the `<small>`s with `<div>`s and use a `<span>` in the horizontal form.

zalog:
This PR:
- removes `display: block;` from `.form-text`
- uses div / block-level element or span / inline-level where needed with `.form-text`
- shares `$form-text-font-size` and `$form-text-font-style` with `$form-feedback-*`

Deploy [preview](https://deploy-preview-30598--twbs-bootstrap.netlify.app/docs/4.3/forms/overview/).
Closes #30577

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -570,7 +570,7 @@ $btn-transition:              color .15s ease-in-out, background-color .15s ease
 // Forms
 
 $form-text-margin-top:                  .25rem !default;
-$form-text-font-size:                   null !default;
+$form-text-font-size:                   $small-font-size !default;
 $form-text-font-style:                  null !default;
 $form-text-color:                       $text-muted !default;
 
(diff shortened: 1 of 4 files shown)


MartijnCuppens (approved):

MartijnCuppens:
Thanks Z!

PR #30605, opened 2020-04-16
grid: checks for `$grid-columns > 0`

zalog:
This PR adds:
- a check for $grid-columns > 0
- permits us to use to assign `$grid-columns: 0;` and removing all `.col-[number]` when needed

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -85,17 +85,19 @@
         @include make-col-auto();
       }
 
-      @for $i from 1 through $columns {
-        .col#{$infix}-#{$i} {
-          @include make-col($i, $columns);
+      @if $columns > 0 {
+        @for $i from 1 through $columns {
+          .col#{$infix}-#{$i} {
+            @include make-col($i, $columns);
+          }
         }
-      }
 
-      // `$columns - 1` because offsetting by the width of an entire row isn't possible
-      @for $i from 0 through ($columns - 1) {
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
We already have the `$enable-grid-classes` variable, which covers this.

mdo:
Yeah, and I favor the explicit option of `$enable-grid-classes` vs something implicit based on a variable's value. Seems safe to close?

zalog:
Sorry @mdo, @MartijnCuppens and I, have some slack dm's on this subject and I forgot to log a follow up here.

Here is a use case for a grid version, a minimal one:

$enable-grid-classes: true; // option as is default
$grid-columns: 0; // covered by this PR
$grid-row-columns: 0; // covered by https://github.com/twbs/bootstrap/pull/30606
$grid-breakpoints: (); // covered by https://github.com/twbs/bootstrap/pull/30609

So after this, our grid will look like:
.row {...}
.row > * {...}
.col {...}
.row-cols-auto > * {...}
.col-auto {...}

So no .col-* and .offset-*. This config, but with some $grid-breakpoints of course :) can be a valid choice if we don't need cols number and we just want equals one or auto.

What do you think?

I think that it's a small change for a big opportunity of improving css filesize when our users doesn't need the whole grid :)

MartijnCuppens:
It's an edge case, but since it only adds a single `if()` test and prevents our codebase from crashing if `$grid-columns: 0;`, this is worth the change.

MartijnCuppens (approved):

PR #30611, opened 2020-04-17
alert border & position fix

AnnaLisik:


MartijnCuppens:
What does this fix?

AnnaLisik:
> What does this fix?

it isnt fix; its enhacement.

ffoodd:
And what does it enhances?

- `.alert-header` would be forced to be `.alert` direct child, why?
- `position: absolute` makes `.alert-header` out of flow, meaning your alert size does not account for it anymore and could possibly overflow—moreover your heading is now above additionnal content…
- why on earth applying `border-radius` on header? it doesn't have `background` nor `border` so it seems useless.

This seems irrelevant. Do you have further explanations?

MartijnCuppens:
This doesn't make sense, closing this.

Issue #28662, opened 2019-04-16
Include import statements in all SCSS files.


JoshMcCullough:
Shouldn't each SCSS file stand on its own? E.g. if you look at `_variables.scss`, it references things in `_functions.scss`, etc. I believe these local imports should be at the top of the file otherwise the consumer needs to import all of the referenced SCSS files in the proper order.

If I am consuming some Bootstrap SCSS, and I need to access the variable `$spacer`, I would write my SCSS as follows:

```scss
@import "~bootstrap/_variables";

.my-class {
  margin-top: $spacer * 2;
}
```

But this will fail to compile with the message: 

>argument `$color` of `darken($color, $amount)` must be a color

But I'm not doing _anything_ with color here. This is because of the line `$link-hover-color:          darken($link-color, 15%) !default;` in `_variables.scss` -- that line references `$link-color` which in turn references the `theme-color` function in `_functions.scss`.

In order to get my SCSS to compile, I need to dig into `_variables.scss` and determine which files it needs to import, and then do the same task again for each of those files, and add them all as imports to my SCSS:

```scss
@import '~bootstrap/_functions';
@import '~bootstrap/_mixins';
@import '~bootstrap/_variables';

.my-class {
  margin-top: $spacer * 2;
}
```

IMO, `_variables.scss` should look like this at the top...

```scss
@import '_functions';
@import '_mixins';
```

...and all of the other SCSS files should import what they need as well. In the end, I should be able to pick any single SCSS file and compile it on its own.

chrisbo:
With no disrespect to the Bootstrap team, I agree with you that the dependencies and order are as clear as mud.

MartijnCuppens:
There are some variables in the variables file that rely on functions defined in `_functions.scss` but I think it would be possible to ditch them.

We 'll need to find a solution for things like this though:
https://github.com/twbs/bootstrap/blob/091aa1e9fd411440c3ff3c95042a5aa1502e6978/scss/_variables.scss#L199-L200

JoshMcCullough:
I'm not suggesting they be removed. All that needs to happen is that each file should include the imports it needs. I should be able to compile `_variables.scss` on its own, because it should import what it needs from the other SCSS files. I suggest leaving _everything else_ untouched, just add imports.

Let me know if I'm overlooking something here and there's a reason it's this way!

MartijnCuppens:
[...]

Issue #28553, opened 2019-03-24
Reorganize Scss folder structure


MartijnCuppens:
Our current `scss` file and folder architecture is a bit unclear. We have a `_variables.scss` file on the same level as the components. There's also a `_mixins.scss` file which includes the mixins in another folder. There might be some room for reorganisation.

We could come up with our own folder structure but it might be interesting to have a look at existing solutions. Maybe [ITCSS](https://www.xfive.co/blog/itcss-scalable-maintainable-css-architecture/) could help us out.

ITCSS has been discussed already in https://github.com/twbs/bootstrap/issues/18233 but this was also about splitting our variables file, I wouldn't do that for now. Just reorganising the files into folders would be a great start.

Feel free to spread your thoughts on this here

/CC @twbs/css-review

mdo:
With the forms redux, I already want to move all that into a new subfolder. What other changes would you be interested in making?

MartijnCuppens:
I guess a folder structure like this would be the result in the end:

```text
bootstrap/
└── scss/
    ├── settings/
    │   └── variables.scss
    ├── tools/
    │   ├── functions.scss
    │   └── mixins
    │       └── ... All mixin files here
    ├── generic/
    │   ├── reboot.scss
    │   └── root.scss
    ├── elements/
    │   └── type.scss
    ├── objects/
    │   ├── images.scss
    │   ├── code.scss
    │   ├── grid.scss
    │   └── transitions.scss
    ├── components/
    │   ├── tables.scss
    │   ├── forms.scss
    │   ├── buttons.scss
    │   ├── dropdown.scss
    │   ├── button-group.scss
    │   ├── input-group.scss
    │   ├── custom-forms.scss
    │   ├── nav.scss
    │   ├── navbar.scss
    │   ├── card.scss
    │   ├── breadcrumb.scss
    │   ├── pagination.scss
    │   ├── badge.scss
    │   ├── alert.scss
    │   ├── progress.scss
    │   ├── list-group.scss
    │   ├── close.scss
    │   ├── toasts.scss
    │   ├── modal.scss
    │   ├── tooltip.scss
    │   ├── popover.scss
    │   ├── carousel.scss
    │   └── spinners.scss
    ├── utilities/
    │   └── utilities.scss
    ├── bootstrap.scss
    ├── bootstrap-grid.scss
    └── bootstrap-reboot.scss
```

In each folder some files could be grouped into folders if needed (like the form stuff).

[...]

PR #30632, opened 2020-04-22
[Idea] ✨ Add TOC to variables.scss

k-utsumi:
to make the structure easier to understand and organize♻️

---

I often look at [variables file](https://github.com/twbs/bootstrap/blob/v4.4.1/scss/_variables.scss).
That file is hard to understand what exists, so I suggest to include TOC.

I usually do like that.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -3,7 +3,46 @@
 // Variables should follow the `$component-state-property-size` formula for
 // consistent naming. Ex: $nav-link-disabled-color and $modal-content-box-shadow-xs.
 
-// Color system
+// TOC
+//
+// - Color system
+// - Options
+//
+// - Spacing
+// - Body
+// - Links
+// - Paragraphs
+//
+// - Grid
+// - Components
+// - Typography
+//
+// - Tables
+// - Buttons + Forms
+// - Z-index master list
+// - Navs
+// - Jumbotron
+// - Cards
+// - Tooltips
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Not sure what the benefit of a TOC is if it doesn't provide any anchor links? The find function in IDE's are doing a great job imo?

ffoodd:
@MartijnCuppens I think the main concern here is not to go to a specific section, but to know they exists.

FWIW I'm not in favor of this (yet) since it could conflict with #28553 and #28662 — and at least doesn't seem to go the same way.

MartijnCuppens:
In that case, we should focus on using `scss-docs` more, which makes it easier to notice without having to dive into the variables file.

Gonna close this since I think this implementation won't be the solution.

k-utsumi:
I understood it is not the solution.
This is for a messy CSS project in my job.

I want to make the TOC simpler and easier to find information like this.
https://github.com/k-utsumi/bootstrap/blob/bd0f58ec0e941c041fe413ad115c0791ee6f1341/scss/_variables.scss

I think it will help organize the order and grouping of codes.

I'll watch those issues. thanks!

k-utsumi:
I found a similar change 💡
https://github.com/twbs/bootstrap/commit/5310bb4c0f1d8516edccd416a976ed903016e406#diff-df1f74faf207a474f1d78c86393221c8R1-R10

PR #30639, opened 2020-04-23
Easier disabled state customization for button variants

ffoodd:
We can seamlessly ease the `disabled` state customization in the `button-variant()` mixin by adding dedicated arguments, which default values are the previously used ones.

I may have to mention this somewhere in the docs, if you're OK with the feature :)


mdo (approved):
Let's document this in our Migration guide (and explain it's new in alpha 2 somehow, I can work on that). Perhaps we also do some Sass references in the buttons docs?

ffoodd:
I added a mention in the migration page, not sure about the wording but at least there's something, now :)

Not sure about Sass reference in the buttons docs either, looks like a meta feature for all components—as requested several times, isn't it?

XhmikosR:
Waiting to merge #31122 first so that we place all alpha2 changes in the new section.

mdo (approved):

PR #30653, opened 2020-04-27
Drop closest from SelectorEngine

MartijnCuppens:
`SelectorEngine.closest()` just calls the native `.closest()`, which beats the point of having this function.

Not sure on which project board this should be added, @XhmikosR


js/src/button.js
```diff
@@ -143,10 +139,7 @@ class Button {
 EventHandler.on(document, EVENT_CLICK_DATA_API, SELECTOR_DATA_TOGGLE_CARROT, event => {
   event.preventDefault()
 
-  let button = event.target
-  if (!button.classList.contains(CLASS_NAME_BUTTON)) {
```
XhmikosR, line 147: Is this intentional?

XhmikosR:
Yeah, I guess it's a leftover from the IE removal PR.

js/src/button.js
MartijnCuppens: Yes, `.closest()` returns the element itself if it contains the class (see https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

XhmikosR:
@Johann-S ping

Johann-S (approved):
LGTM 👍

PR #30656, opened 2020-04-27
Add Action: Label PRs

mdo:
Adds an Action to automatically label PRs based on what files they modify. I could go either way on this, because we'd likely also want to automatically add `v5` and projects perhaps. Thoughts?


skmedix:
**Note:** This action does not support `push` event

PR #30677, opened 2020-04-29
Move codeql.yml one dir up

XhmikosR:
TODO:

- [x] Fix the permission error

PR #30678, opened 2020-04-29
Move codeql.yml workflow to top level and specify javascript

greysteil:
Follow up to https://github.com/twbs/bootstrap/pull/30676. That PR created the GitHub Action workflow for CodeQL in the wrong place (🤦) - this moves it so it will actually run.


XhmikosR:
Hey, @greysteil I've made #30677 for this :)

I was just waiting to get rid of this error before merging:

>##[error]The repo on which this action is running is not opted-in to CodeQL code scanning.

greysteil:
Nice! Let me get that error fixed for you...

PR #30680, opened 2020-04-29
Use `safeHTMLAttr` in the `integrity` attributes.

XhmikosR:
I just noticed this after testing Microsoft Edge (old Edge) and I was getting this error although the CSS did load:

```
SEC7136: [Integrity] The origin 'https://twbs-bootstrap.netlify.app' failed an integrity check for a style resource at 'https://twbs-bootstrap.netlify.app/docs/4.3/dist/css/bootstrap.min.css'.
```

It seems Hugo escapes those. Now the problem is these two:

1. The masthead-followup.html becomes more and more unreadable IMO. But at least it does work properly.
2. There are 2 cases we print integrity hashes in Markdown with shortcodes where I don't know how to fix the issue:
  * https://github.com/twbs/bootstrap/blob/master/site/content/docs/4.3/getting-started/download.md
  * https://github.com/twbs/bootstrap/blob/master/site/content/docs/4.3/getting-started/introduction.md

@regisphilibert: maybe you have any suggestions how to fix this? 🙂

Preview: <https://deploy-preview-30680--twbs-bootstrap.netlify.app/>

### DIFF
--- a/site/layouts/_default/examples.html
+++ b/site/layouts/_default/examples.html
@@ -44,13 +44,13 @@
 
     {{ if ne .Page.Params.include_js false -}}
       {{- if eq hugo.Environment "production" -}}
-        <script src="/docs/{{ .Site.Params.docs_version }}/dist/js/bootstrap.bundle.min.js" integrity="{{ .Site.Params.cdn.js_bundle_hash }}" crossorigin="anonymous"></script>
+        <script src="/docs/{{ .Site.Params.docs_version }}/dist/js/bootstrap.bundle.min.js" {{ printf "integrity=%q" .Site.Params.cdn.js_bundle_hash | safeHTMLAttr }} crossorigin="anonymous"></script>
       {{- else -}}
         <script src="/docs/{{ .Site.Params.docs_version }}/dist/js/bootstrap.bundle.js"></script>
       {{- end }}
 
       {{ range .Page.Params.extra_js -}}
-        <script{{ with .async }} async{{ end }} src="{{ .src }}"{{ with .integrity }} integrity="{{ . }}" crossorigin="anonymous"{{ end }}></script>
+        <script{{ with .async }} async{{ end }} src="{{ .src }}"{{ with .integrity }} {{ printf "integrity=%q" . | safeHTMLAttr }} crossorigin="anonymous"{{ end }}></script>
       {{- end -}}
     {{- end }}
   </body>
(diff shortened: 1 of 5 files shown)


regisphilibert:
@XhmikosR I can take a look, but could you point me the shortcodes in questions, which line on the page? What you're expecting and what is the current result?

Thanks a ton.

XhmikosR:
It's just the `highlight` shortcode. Unfortunately I cannot link directly to the lines because GitHub renders those pages as HTML. Here's an example:


```
{{< highlight html >}}
<link rel="stylesheet" href="{{< param "cdn.css" >}}" integrity="{{< param "cdn.css_hash" >}}" crossorigin="anonymous">
<script src="{{< param "cdn.js" >}}" integrity="{{< param "cdn.js_hash" >}}" crossorigin="anonymous"></script>
{{< /highlight >}}
```

Thanks for taking a look!

XhmikosR:
Thanks for the reply, @regisphilibert!

So, your suggestion is to make 2 different shortcodes? Because we have similar but not identical code in content pages.

In https://github.com/twbs/bootstrap/blob/master/site/content/docs/4.3/getting-started/download.md:

```markdown
{{< highlight html >}}
<link rel="stylesheet" href="{{< param "cdn.css" >}}" integrity="{{< param "cdn.css_hash" >}}" crossorigin="anonymous">
<script src="{{< param "cdn.js" >}}" integrity="{{< param "cdn.js_hash" >}}" crossorigin="anonymous"></script>
{{< /highlight >}}

If you're using our compiled JavaScript, don't forget to include Popper.js, via a CDN preferably, before our JS.

{{< highlight html >}}
<script src="{{< param "cdn.popper" >}}" integrity="{{< param "cdn.popper_hash" >}}" crossorigin="anonymous"></script>
{{< /highlight >}}
```

And in https://raw.githubusercontent.com/twbs/bootstrap/master/site/content/docs/4.3/getting-started/introduction.md we have 3 different snippets.

Do you think you could make a PR against this PR's branch with your suggestion please? :)

BTW maybe I should report this upstream; I'm just not 100% sure it's a bug or not.

regisphilibert:
I changed my suggestion :) I think it's just about rewriting the param shortcode with a simple test. For now I'm looking for the `hash` substring in the param keys:

You can just drop the following in your shortcode dir (/layouts/shortcodes/param.html):
```
{{- $name := (.Get 0) -}}
{{- with $name -}}
{{- $value := $.Page.Param . -}}
{{- if in $name "_hash" -}}
  {{- $value = $value | safeHTML -}}
{{- end -}}
{{- with $value }}{{ . }}{{ else }}{{ errorf "Param %q not found: %s" $name $.Position }}{{ end -}}
{{- else }}{{ errorf "Missing param key: %s" $.Position }}{{ end -}}
```

XhmikosR:
Wow, simple and works perfectly, thank you!

I used you as the author for the second patch, I hope you don't mind. BTW I don't suppose we could reformat the shortcode to get rid of the one-liners? Nitpicking, I'm just used to multi-line statements 🙂

MartijnCuppens (approved):

PR #30685, opened 2020-04-29
`<pre>`: Disable auto-hiding scrollbar in legacy Edge

MartijnCuppens:
Disable auto-hiding scrollbar in legacy Edge to avoid overlap, making it impossible to interact with the content.
image

Ref. https://github.com/twbs/bootstrap/commit/f3fc973dd93d1bb69183404d9985157284a3be7f#diff-7f9eec82312835f6155c8de839bf0e49

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -290,6 +290,8 @@ samp {
 // 1. Remove browser default top margin
 // 2. Reset browser default of `1em` to use `rem`s
 // 3. Don't allow content to break outside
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

XhmikosR:
FYI @patrickhlauke

XhmikosR:
Also, this affects IE 11 and 10. This patch fixes the issue and restores the behavior.

PR #30686, opened 2020-04-29
[type=radio] should always be rounded

ffoodd:
They're currently squared if `$enable-rounded` is set to false—which is IMHO bad for usability, and quite confusing for the end users.


MartijnCuppens:
We could just use the property here instead.

ffoodd:
Indeed, it'll be more robust if `border-radius()` mixin evolves in some way.

MartijnCuppens (approved):

PR #30689, opened 2020-04-29
Server side and client side is misplaced

sagarvd01:
Fixes #30688

In the page [/docs/4.1/components/forms/](https://getbootstrap.com/docs/4.1/components/forms/), there is a section named **Server side** and it's content is as follows

>We recommend using client side validation, but in case you require server side, you can indicate invalid and valid form fields with .is-invalid and .is-valid. Note that .invalid-feedback is also supported with these classes.

I believe the words server side and client side is misplaced here because the recommended method is server side validation (because client side validation can easily bypassed) and as mentioned in the paragraph, we can't access the class names from server side.


XhmikosR:
4.1 are ancient docs BTW https://getbootstrap.com/docs/4.4/components/forms/#server-side

mdo:
The docs are correct as-is.

> Server side
> We recommend using client side validation, but in case you require server side

The recommendation is to use native browser or custom styles, as shown before these in the docs before this section. The opening line there is simply a transition.

sagarvd01:
Ah ok. Thanks for the clarification.

PR #30691, opened 2020-04-29
Replace double tilde with `Math.floor`

XhmikosR:
Nowadays, browser engines have improved a lot and `Math.floor` is the same, if not faster.
Also, clarity is better than micro-optimizations in this case.


js/src/util/index.js
```diff
@@ -26,7 +26,7 @@ const toType = obj => {
 
 const getUID = prefix => {
   do {
-    prefix += ~~(Math.random() * MAX_UID) // "~~" acts like a faster Math.floor() here
+    prefix += Math.floor((Math.random() * MAX_UID))
```
alecpl, line 29: We don't need that extra brackets, do we?

XhmikosR: True, that's why the PR is a draft also 😇

XhmikosR:
IE and old Edge do benefit from double tilde. Not that it matters a lot but anyway, I removed the change from 4.4.x.

mdo (approved):

PR #30703, opened 2020-05-01
Remove unneeded twitter:title and twitter:description from docs

coliff:
These only need to be defined if they are different to the Opengraph title/descriptions. As they are identical they aren't needed.
https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started#twitter-cards-and-open-graph
Removing them slightly reduces HTML file size / improves page load time.
I didn't remove/touch the twitter:image tag as it looks like that behaves differently to the og:image one.


XhmikosR:
I know they are redundant, but the claim about improving page load time is misleading; it's only a few bytes less which after compression it won't make any difference.

I'd rather keep things as is for now.

mdo:
Closing per last comment. Happy to revisit though if we think @XhmikosR and team approve.

PR #30707, opened 2020-05-02
Add position top, right, bottom, left _utilities.scss

marco-van-zomeren:
Add position top, right, bottom, left utilities


mdo:
This is a no from me. No need for all of these in Bootstrap; the API is there for this to be customized by you.

PR #30708, opened 2020-05-03
docs: add back to top button

XhmikosR:
Just a first draft.

Preview: <https://deploy-preview-30708--twbs-bootstrap.netlify.app/>

### DIFF
--- a/site/assets/scss/_smooth_scrolling.scss
+++ b/site/assets/scss/_smooth_scrolling.scss
@@ -1,0 +1,7 @@
+html {
+  scroll-behavior: smooth;
+
+  @media (prefers-reduced-motion: reduce) {
+    scroll-behavior: auto;
+  }
+}
--- a/site/assets/scss/docs.scss
+++ b/site/assets/scss/docs.scss
@@ -47,6 +47,7 @@
 @import "colors";
 @import "clipboard-js";
 @import "placeholder-img";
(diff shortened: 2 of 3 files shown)


mdo:
I'm still 👎 on adding a back to top link. Haven't had one in years in Bootstrap and it's been awfully quiet on the requests for it. Page Up, Cmd+Up, tap status bar, etc all get you back to the top depending on your device.

XhmikosR:
I'm not tied to adding it, but my rationale is that it's handy at certain situations.

Not everyone is using a keyboard and we do have some really long pages.

Regardless, I don't mind dropping this if everyone thinks it's not useful.

patrickhlauke:
i'll note that for keyboard users, back to top links can still be useful, as using `Home` or `PgUp` etc will only move the visual scroll position, but not the actual focus (though there are other cumbersome workarounds like doing `ALT+d` to focus the browser's address bar and then tab forward)

ffoodd:
FWIW, Boosted has a JS plugin to handle a "Back to top" link apparition when a full viewport height has been scrolled: it's on my roadmap anyway. 

I'd be pleased to open a dedicated PR when working on this: you'll be able to play around and discuss it furthermore.

XhmikosR:
Closing for now, we can always revisit if needed later.

PR #30701, opened 2020-04-30
v4: Add Release Drafter

mdo:

PR #30711, opened 2020-05-03
Revert "v4: Add Release Drafter"

XhmikosR:
Reverts twbs/bootstrap#30701

>@mdo this doesn't seem to work for v4?
>
>EDIT: Maybe we need the config for v4-dev added in master too? I had the same issue with dependabot. If so, let me know and I'll revert the v4-dev patch later.

TODO:

- [x] remove #30701 from the v4.4.2 project if this gets merged


mdo:
Yeah let's try putting it in `master`. Seems Actions don't get picked up at all in a non-default branch.

XhmikosR:
@mdo please approve then 🙂

MartijnCuppens (approved):

PR #30712, opened 2020-05-03
Update LICENSE grammar

sandip2224:


patrickhlauke:
that's in the original MIT license, which we shouldn't really edit on a whim. https://opensource.org/licenses/MIT

PR #30720, opened 2020-05-04
Update karma conf and karma-browserstack-launcher

XhmikosR:
* use the recommended karma-browserstack-launcher settings
* update karma-browserstack-launcher to the latest version

https://github.com/browserstack/karma-browserstack-example/blob/540e6ffd457ec6f0c59627b1d310a0c3109c27b0/karma.conf.js#L85-L89


XhmikosR:
If this breaks BrowserStack again, we can revert it later. I really hope this fixes the BrowserStack issues once and for all, though.

MartijnCuppens (approved):

Issue #30684, opened 2020-04-29
Revisit our BrowserStack browsers


XhmikosR:
Currently, we test the latest versions of all the browsers we support.

While that's fine, it only covers half of the picture. Also, in the case of Edge, we are no longer testing on the old Edge which is still supported.

This issue is present on both master and v4-dev. I think we better test both the minimum browser versions and the latest ones. It will make tests slower, but after we split the actions, this should be a lot more bearable.

/CC @Johann-S

XhmikosR:
I think we can close this for now; I've made some recent updates to the BrowserStack browsers we test on.

PR #30725, opened 2020-05-04
BrowserStack: test Edge 16 too

XhmikosR:
Refs #30684 

I'll tackle the iOS/macOS browsers in another PR. So far this patch seems to only add ~5 seconds, so I think it's fine.


XhmikosR:
Arguably we could probably stop testing on latest Edge since it's Chromium-based.

MartijnCuppens:
> Arguably we could probably stop testing on latest Edge since it's Chromium-based.

Indeed

XhmikosR:
I say let's just merge this and we revisit this in another PR along with the other BrowserStack version changes.

Johann-S (approved):

PR #30740, opened 2020-05-05
karma: stop excluding polyfill.js from istanbul

XhmikosR:


Johann-S:
I don't understand why do we need to remove that 🤔 we still have some polyfills

XhmikosR:
It seems the custom coverage thresholds are no longer needed.

Johann-S:
oh yeah saw that on Coveralls, how is it possible ? 🤔

XhmikosR:
We have removed some polyfills.

XhmikosR:
@Johann-S LGTY? The reason the file was excluded was the inline ignore comment. Arguably, the whole file could probably go away later see #32017

Johann-S:
yep LGTM 👍 and indeed we should check if we still need those polyfills

Johann-S (approved):

PR #30742, opened 2020-05-05
Update to `@rollup/plugin-babel`

### LINKED ISSUE #30683: Update `rollup-plugin-babel` to `@rollup/plugin-babel`
For master at first and we backport it to v4-dev too.

/CC @Johann-S

XhmikosR:
Fixes #30683

Changelog: https://github.com/rollup/plugins/blob/master/packages/babel/CHANGELOG.md#500

> Removed `externalHelpers` & `runtimeHelpers` options. There is now a single `babelHelpers` option which can take one of `'bundled'`, `'inline'`, `'runtime'` and `'external'` as a value. The default is `'bundled'` which matches 4.x behavior, but it is recommended to configure this option explicitly.

So, apart from changing the import, I also had to adapt to the `babelHelpers` helper option.

Also, I removed the `plugins` option in  `js/tests/karma.conf.js` since we already have this in our `.babelrc.js` in root, but we need to confirm if it's being discovered automatically.

TODO:

- [x] Verify all builds are the same
- [ ] Verify that `.babelrc.js` is discovered and used by karma
- [ ] Backport this in v4-dev


XhmikosR:
Not sure who can review this, it does seem to work

>     * Verify that `.babelrc.js` is discovered and used by karma

This one I'm not sure about, but the dist builds are the same.

XhmikosR:
@mdo I need to proceed someway here...

mdo (approved):

PR #30752, opened 2020-05-06
Fix storing multiple data keys

MisatoTremor:
The current implementation does not support storing multiple sets of data (the `key` parameter is mostly useless), which is fixed by this PR.

As a sidenote: The data storage indices for the keys are stored on the element attribute `key`. HTMLElement does not use this attribute, yet, and to my knowledge no subtype does. But it is a very generic name which might be prone to collide in the future, maybe a more bootstrap specific attribute would be better.


Johann-S:
Hi @MisatoTremor,

Thanks for your work 👍 

Currently we have no use case where multiple data keys would be necessary, so it's not a good time for that feature.

> But it is a very generic name which might be prone to collide in the future, maybe a more bootstrap specific attribute would be better.

I totally agree with you on that point, and I would gladly accept a PR which target that 👍

MisatoTremor:
I see your point.
My thinking was, that as part of a public framework it's likely that this API will be used by others in their projects, moreso when this version will be released. And that it would be better to have this relatively simple fix applied, than having a "broken" version "out in the wild".
Also if there ever will be a use case for this, will the one adding that feature it recognize this bug or just stumble over it, wondering why theirs is not working?

Johann-S:
It's not part of our public API, it's a internal tool we use for our plugins.
You can find our what we expose here: https://github.com/twbs/bootstrap/blob/master/js/index.esm.js

MisatoTremor:
Sorry for my bad wording, I did not mean that Data is part of the public API but an API that is part of a publicly available codebase.

My point is it can be easily copied and used anywhere else and as beeing part of a popular site or framework, someone doing so might assume it should be working correctly. StackOverflow recently experienced what that can lead to. So this "fix" is more about not letting something with known bugs "lying around".

Again, sorry, I don't how to express this in a better way.

Johann-S:
Hi @MisatoTremor 

I renamed `key` to `bsKey`, can you rebase your branch ?

Johann-S (approved):
LGTM 👍 good work @MisatoTremor 👍

mdo:
Should this land in alpha 3? Do we need to add anything to the migration guide?

Johann-S:
@mdo it should land as it please you it's ready, and there are no changes to made in the migration guide since it wasn't exist in v4

XhmikosR:
My only concern is that we don't need this ourselves.

Johann-S:
Yep AFAIK we won't enable multi plugins on the same element

XhmikosR:
@rohit2sharma95 should we close this since we plan to move with #32180?

rohit2sharma95:
Yes, this should be closed then. :+1: 
(Or we can wait for #32180 to be merged and then require changes here)

XhmikosR:
Like I mentioned in another issue, we shouldn't sneak in the multiple plugins change in the same PR. But this one will be outdated if we are going to go with #32180.

Let's leave this open until we decide how to proceed with #32180.

GeoSot:
It will be closed for now, as bs doesn't support multiple plugins on the same element

@MisatoTremor It is a brilliant idea, for a future update. Do not hesitate to propose any new ideas

MisatoTremor:
@GeoSot Is this support planned for a specific update (i.e. v6) / any issue or PR I could follow for this?

GeoSot:
As I know, it is not, but if you believe it is important to be supported, you can open an issue with code examples.
I want to remind you, that almost all components listen on events and attributes, and an implementation like this, may cause many other problems, pare example, which js component will be responsible to listen an event and which not

PR #30755, opened 2020-05-06
Removed role="document" from the modal dialog

### LINKED ISSUE #30687: Remove role="document" from the dialog and add aria-modal="true"
### Current Code
    <div class="modal" tabindex="-1" role="dialog">
    <div class="modal-dialog" role="document">

### Solution code
    <div class="modal" tabindex="-1" role="dialog" aria-modal="true">
     <div class="modal-dialog" >

## Rational
Document role is only necessary if there role="application" on a parent that overrides screen reader short cut keys.
The ```role="dialog"``` doesn't do that, so it's fine without it ... see aria-practices example
https://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html

```aria-modal="true"``` ensure the screen reader user cannot arrow out of the modal inadvertently and replaces the old technique (not used in bootstrap) of aria-hidden on all background content.

rohit2sharma95:
Closes #30687 

Updated the document for the modal component.

### DIFF
--- a/js/tests/visual/modal.html
+++ b/js/tests/visual/modal.html
@@ -35,7 +35,7 @@
       <h1>Modal <small>Bootstrap Visual Test</small></h1>
 
       <div class="modal fade" id="myModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
-        <div class="modal-dialog" role="document">
+        <div class="modal-dialog">
           <div class="modal-content">
             <div class="modal-header">
               <h4 class="modal-title" id="myModalLabel">Modal title</h4>
@@ -122,7 +122,7 @@
       </div>
 
       <div class="modal fade" id="firefoxModal" tabindex="-1" role="dialog" aria-labelledby="firefoxModalLabel" aria-hidden="true">
-        <div class="modal-dialog" role="document">
+        <div class="modal-dialog">
           <div class="modal-content">
             <div class="modal-header">
               <h4 class="modal-title" id="firefoxModalLabel">Firefox Bug Test</h4>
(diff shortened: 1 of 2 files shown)


XhmikosR:
Not sure if this applies to v4 too @patrickhlauke

ffoodd:
I'm not sure `aria-modal` should be shown in the examples markup: our scripts adds it when showing modal and removes it when hiding it. @patrickhlauke @Johann-S ?

rohit2sharma95:
> I'm not sure `aria-modal` should be shown in the examples markup: our scripts adds it when showing modal and removes it when hiding it.

@ffoodd is right though. @patrickhlauke and @Johann-S?

patrickhlauke:
indeed, i forgot our scripts already handle `aria-modal="true"`. so this really is just about removing the redundant `role="document"`.

and yes should be done for v4 as well.

patrickhlauke (changes requested):
leave out all the additions of `aria-modal="true"`. just keep the removal of `role="document"` from the tests, code examples, and the explanatory paragraph of text

rohit2sharma95:
Okay @patrickhlauke! Let me do it.

XhmikosR:
@patrickhlauke please check again so that I backport this and include it in the upcoming v4.4.2 🙂

ffoodd (approved):
LGTM

patrickhlauke (approved):

patrickhlauke:
@XhmikosR looks good to me. possibly in future we could even consider injecting the `role="dialog"` via scripting (the same way `aria-modal="true"` is being injected. one less thing for authors to worry about.

XhmikosR:
Sounds good to me! Just make an issue about it so that we don't forget 🙂

PR #30757, opened 2020-05-07
Update punctuation

fulldecent:
Switch to n-dash for a range


mdo:
> Switch to n-dash for a range

😻

mdo (approved):

PR #30759, opened 2020-05-07
Improve build/zip-examples.js

XhmikosR:
* remove favicons meta tags
* copy the two brand images we use in the examples

`sh.sed` has the same limitations as the `sed` tool, hence this solution. Not perfect but does the job.

https://github.com/shelljs/shelljs#sedoptions-search_regex-replacement-file--file-

>Also, like unix sed, ShellJS sed runs replacements on each line from the input file (split by '\n') separately, so search_regexes that span more than one line (or include '\n') will not match anything and nothing will be replaced.

### DIFF
--- a/build/zip-examples.js
+++ b/build/zip-examples.js
@@ -27,15 +27,12 @@ sh.cd(path.join(__dirname, '..'))
 
 // remove any previously created folder with the same name
 sh.rm('-rf', folderName)
+// create any folders so that `cp` works
 sh.mkdir('-p', folderName)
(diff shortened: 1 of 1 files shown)


mdo (approved):
Tested locally

XhmikosR:
I added one more patch which combines the dist and the assets folder, let me know what you guys think.

XhmikosR:
@MartijnCuppens friendly ping

MartijnCuppens (approved):
Tested locally, LGTM

PR #30762, opened 2020-05-07
Update _custom-forms.scss

arthurshlain:
Typo in box-shadow attribute, global variable used instead of local.

Found this because i cannot set checkbox shadow to `none` on `focus` state.
Checkbox shadow dissapear only when i turn off all shadow, but it's not true way to solve it.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -37,7 +37,7 @@
   &:focus ~ .custom-control-label::before {
     // the mixin is not used here to make sure there is feedback
     @if $enable-shadows {
-      box-shadow: $input-box-shadow, $input-focus-box-shadow;
+      box-shadow: $input-box-shadow, $custom-control-indicator-focus-box-shadow;
     } @else {
       box-shadow: $custom-control-indicator-focus-box-shadow;
     }


XhmikosR:
@twbs/css-review is this still valid?

ffoodd:
I guess so, and it makes sense since `$custom-control-indicator-focus-box-shadow` is set to match `$input-focus-box-shadow` ATM.

However this should not be used to remove `box-shadow`: as mentionned in the comment above, `// the mixin is not used here to make sure there is feedback`. If `$custom-control-indicator-focus-box-shadow` is `null` or `none`, there would be no focus indicator anymore…

To clarify, this fix is good, but shouldn't be used as requested. Not sure about what to do with this though. @MartijnCuppens @mdo any idea?

XhmikosR:
Isn't this present in master too? If so please add the needed labels.

ffoodd:
`master` only sets a single `box-shadow`, with a specific variable, so it can be nullified at any time.

mdo (approved):
Yeah, this is valid still—slating for v4.5.3!

mdo:
Looks like the remote is no longer there—tests look a little wonky here. Replacing with #31693.

PR #30768, opened 2020-05-09
V4 dev updates and backports

XhmikosR:
TODO:

- [ ] Simplify/improve the newly added scrollspy test cases
- [ ] Tweak the newly added validation workaround examples if needed
- [ ] Do we need to port the workaround to master too?
- [ ] Review the wording changes
- [ ] **Make a final rebase before merge**

Preview:

* <https://deploy-preview-30768--twbs-bootstrap.netlify.app/>
* <https://deploy-preview-30768--twbs-bootstrap.netlify.app/docs/4.4/components/forms/#input-group-validation-workaround>

PR #30772, opened 2020-05-10
Prevent scrollbar replacement on non-integer width

### LINKED ISSUE #29681: Bad scrollbar replacement on non-integer width windows

kremerd:
Fixes #29681, Equals #30017, but also includes a test case.

### DIFF
--- a/js/tests/unit/modal.spec.js
+++ b/js/tests/unit/modal.spec.js
@@ -649,6 +649,38 @@ describe('Modal', () => {
       modal.show()
     })
 
+    it('should not adjust the inline body padding when it does not overflow, even on a scaled display', done => {
+      fixtureEl.innerHTML = '<div class="modal"><div class="modal-dialog"></div></div>'
+
+      const modalEl = fixtureEl.querySelector('.modal')
+      const modal = new Modal(modalEl)
+      const originalPadding = window.getComputedStyle(document.body).paddingRight
+
+      // Remove body margins as would be done by Bootstrap css
+      document.body.style.margin = '0'
+
+      // Hide scrollbars to prevent the body overflowing
+      document.body.style.overflow = 'hidden'
+
+      // Simulate a discrepancy between exact, i.e. floating point body width, and rounded body width
+      // as it can occur when zooming or scaling the display to something else than 100%
+      document.documentElement.style.paddingRight = '.48px'
+
+      modalEl.addEventListener('shown.bs.modal', () => {
+        const currentPadding = window.getComputedStyle(document.body).paddingRight
+
+        expect(currentPadding).toEqual(originalPadding, 'body padding should not be adjusted')
+
+        // Restore overridden css
+        document.body.style.removeProperty('margin')
+        document.body.style.removeProperty('overflow')
+        document.documentElement.style.paddingRight = '16px'
+        done()
+      })
+
+      modal.show()
+    })
+
     it('should enforce focus', done => {
       fixtureEl.innerHTML = '<div class="modal"><div class="modal-dialog"></div></div>'
 
--- a/js/src/modal.js
+++ b/js/src/modal.js
@@ -443,7 +443,7 @@ class Modal {
 
   _checkScrollbar() {
     const rect = document.body.getBoundingClientRect()
-    this._isBodyOverflowing = rect.left + rect.right < window.innerWidth
+    this._isBodyOverflowing = Math.round(rect.left + rect.right) < window.innerWidth
     this._scrollbarWidth = this._getScrollbarWidth()
   }
 


js/tests/unit/modal.spec.js
```diff
@@ -11,8 +11,11 @@ describe('Modal', () => {
   beforeAll(() => {
     fixtureEl = getFixture()
 
-    // Enable the scrollbar measurer
-    const css = '.modal-scrollbar-measure { position: absolute; top: -9999px; width: 50px; height: 50px; overflow: scroll; }'
+    // Enable the scrollbar measurer and reset body margin
+    const css = `
+      .modal-scrollbar-measure { position: absolute; top: -9999px; width: 50px; height: 50px; overflow: scroll; }
+      body { margin: 0; }
```
kremerd, line 17: While not strictly necessary for the test to work, I believe it makes sense to include this style definition here: The values of `rect.left` and `rect.right` in the `_checkScrollbar` method depend on this margin, and when not set explicitly, a browser default will be used (e.g. `8px` in Chrome) instead of the [value actually set by Bootstrap](https://github.com/twbs/bootstrap/blob/master/scss/_reboot.scss#L41).

XhmikosR:
Thanks @kremerd! I rebased this to include the original patch and authorship, so please don't overwrite it.

BTW for the v4-dev backport the test will have to be adapted since we use QUnit there. But that's after this is merged.

XhmikosR:
Tested this on BrowserStack too, seems to be fine.

@Johann-S please review ASAP so that we backport it to #30768 and include it in 4.4.2 🙂

js/tests/unit/modal.spec.js
```diff
@@ -11,8 +11,8 @@ describe('Modal', () => {
   beforeAll(() => {
     fixtureEl = getFixture()
 
-    // Enable the scrollbar measurer
-    const css = '.modal-scrollbar-measure { position: absolute; top: -9999px; width: 50px; height: 50px; overflow: scroll; }'
+    // Enable the scrollbar measurer and reset body margin
```
XhmikosR, line 14: @kremerd my only concern is that this should be set only for the newly added test. What do you think?

kremerd: @XhmikosR Yes, you're right. I originally thought it might make sense for all scrollbar related tests, but that is neither necessary, nor within the scope of this PR. I pushed a patch to restrict the setting.

Since I'm still new to the PR process: Was it right to publish this as a new commit, or should I rather have amended my previous one and force pushed?

XhmikosR: @kremerd it's OK, ideally should have been squashed, but I squashed it myself. Thanks!

XhmikosR (approved):

XhmikosR:
@kremerd thanks! Now the backport to v4-dev is left. I'll try to do it so that we include the patch in the upcoming v4.4.2, but if you have time feel free to backport it to v4-dev and ping me.

XhmikosR:
All good, I cherry picked it in #30768, so it'll be in v4.4.2. Thanks again @kremerd!

PR #30778, opened 2020-05-11
v5: Utilities CSS cleanup, new border and bg utils, renames rounded to radius

### LINKED ISSUE #29989: Confusing classes `rounded-sm` and `rounded-lg`

mdo:
- Rearranges utilities CSS to better reflect our property order
- Renames `rounded` utils to `radius`, closes #30048 and closes #29989.
- Adds new `border-gray-500` and `border-gray-700` class
- Adds new `bg-gray` class
- Adds some docs to the border utils page to show how I intend these new utils to be combined

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -4,20 +4,13 @@ $utilities: () !default;
 // stylelint-disable-next-line scss/dollar-variable-default
 $utilities: map-merge(
   (
-    "align": (
-      property: vertical-align,
-      class: align,
-      values: baseline top middle bottom text-bottom text-top
+    // Positioning
+    "position": (
+      property: position,
+      values: static relative absolute fixed sticky
     ),
-    "float": (
-      responsive: true,
-      property: float,
-      values: left right none
-    ),
-    "overflow": (
-      property: overflow,
-      values: auto hidden,
-    ),
+
+    // Display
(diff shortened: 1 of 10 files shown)


alecpl:
I kinda prefer the long suffixes (-left, -right, -top, -bottom). I see from where the short version comes from, but.... `radius-1` (one) and `radius-l` (left) looks almost the same, which might be confusing.

I propose to use `rl`, `rr`, `rb`, `rt`. Also, adding `x` and `y`variants and all 6 size specifiers would make sense. Also, make the size specified required (so no `radius`, but `r-X`). It would unify radius tool with spacing utilities.

alecpl:
After writing above comment I realized that most utilities do not use abbreviations. So, it would make more sense if spacing utilities would use whole words (padding, margin, top, left, etc.) too. Rather than changing `radius` to `r`

MartijnCuppens:
Feels weird to have numbers in the border utilities, but not in the bg colors. Not sure how we could improve that though.

mdo:
> Feels weird to have numbers in the border utilities, but not in the bg colors. Not sure how we could improve that though.

Agreed... will think on it more.

mdo:
Closing as it's full of conflicts.

PR #30781, opened 2020-05-12
Reboot's `th` updates: Inherit `font-weight: bold` that comes from user agent stylesheets.

zalog:
Inherits `font-weight: bold` that comes from user agent stylesheets.


MartijnCuppens:
I like the default bold font weight which makes a visual difference between the table cells. We could `null`ify this variable though

zalog:
Hmm, yep but I think the default needs to be consistent through all browsers.
Done adding a var :)

ffoodd:
Agreed, [every browser applies `font-weight: bold;`](https://browserdefaultstyles.com/) (search for `th`) so there is a kind of consensus here.

I'm not sure about the interest of allowing this through a variable, TBH: it's pretty easy to either apply a utility or to override those styles. We'd be adding code for something that hasn't been requested.

zalog:
Yep.

I think if bs users needs to style th, they should do this on the html element, and not class. Seems "natural" I think :)

ffoodd:
Using a Sass variables is not easier than editing `_reboot.scss`—and as outlined previously, there has never be any request for this, as far as I can search in issues.

Furthermore, if one uses this to unset `bold`ness on `th`, how are they supposed to make them distinguishable from other cells?Because they should.

zalog:
"Using a Sass variables is not easier than editing _reboot.scss" actually it is and the thing is, I think we don't want to force our users to edit any of our bootstrap core code.

"how are they supposed to make them distinguishable", yep. I think the possibilities are endless here: a simple border under `thead > * > *`, or a background can do that. There are a lot of use cases here. Right now, I'm working on a table that has on thead font size bigger and no other differences between tbody's :)

Imho, we should give our user this option, of not leaving browsers manage this and not trough `.table` class :)

ffoodd:
It's a bit clearer, thanks for your patience :)

I'm still not sure about this, since it could too easily give undistinguishable `th`. Not strictly against but not really for either :')

@MartijnCuppens feel free to approve or discuss!

mdo (approved):
I'm in favor of this since the font-weight stays bold. We could easily backport this to v4 as well I'd think. Thoughts @MartijnCuppens?

MartijnCuppens (approved):

MartijnCuppens (approved):

XhmikosR:
I marked this for backport, but you guys decide if it's considered a breaking changes. The next v4 release is supposed to be a patch release.

MartijnCuppens:
It's not a BC, this PR just adds a `null` variable

XhmikosR:
I will backport it if it applies clean then.

XhmikosR:
This doesn't apply clean to v4-dev, feel free to backport it manually @MartijnCuppens

PR #30809, opened 2020-05-13
v4: Update docs callout for responsive SVG images

### LINKED ISSUE #25242: IE11: Figure with .img-fluid is not resized
The fluid image in a figure container is not resized at all making it overflow the content area.

On MacOS with Safari 11.0.2 it looks fine:
macos-safari

On Win 7 with Internet Explorer 11 .0.9600.18015 it looks like this:
bildschirmfoto 2018-01-09 um 14 36 47

Does seem to be somewhat related to `display: inline-block` on `figure` as deactivating that property (falling back to 'display: block' makes it look fine on IE11:
win7-ie11-display

There have been related issues (#20970, #21360, #21886, #21888) describing this behavior for older versions of Firefox or which are closed to inactivity. There doesn't seem to be a fix for IE11 (which is a supported version I suppose).

I created a [minimal example on JS Bin](http://jsbin.com/yewehaz/1/edit?html,output).

mdo:
Turns out this applies to IE11 as well as IE10, so this clarifies the docs callout.

Closes #25242.

### DIFF
--- a/site/docs/4.5/content/images.md
+++ b/site/docs/4.5/content/images.md
@@ -19,9 +19,9 @@ Images in Bootstrap are made responsive with `.img-fluid`. `max-width: 100%;` an
 {% endhighlight %}
 
 {% capture callout %}
-##### SVG images and IE 10
+##### SVG images and Internet Explorer
 
-In Internet Explorer 10, SVG images with `.img-fluid` are disproportionately sized. To fix this, add `width: 100% \9;` where necessary. This fix improperly sizes other image formats, so Bootstrap doesn't apply it automatically.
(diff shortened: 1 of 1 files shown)


ffoodd:
Didn't know of this section.

For SVGs, couldn't our reboot use something like `img[src$=".svg"]` to apply `width`?

AFAIC this PR looks good 🙂

ffoodd (approved):

mdo:
> For SVGs, couldn't our reboot use something like `img[src$=".svg"]` to apply `width`?

If I recall correctly, this could have unintended consequences for perfectly sized SVGs. Opt-in has seemed like the best approach here, and we don't ned this in v5 at all, so I'm okay leaving nearly as-is.

PR #30810, opened 2020-05-13
v4: Add z-index to .custom-check to fix custom checks in CSS columns

### LINKED ISSUE #29607: IE 10/11/Edge custom control indicators in a parent element with column-count do not render on the last column
- Operating system and version (Windows, macOS, Android, iOS, Win10 Mobile)
**Windows 10**
Reproduced on **IE Edge, IE 11 and IE 10**
- [Reduced test case](https://css-tricks.com/reduced-test-cases/) and suggested fix using [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/)

**Screenshot of issue:**
ie-custom-control

**Codepen reproduction:**
https://codepen.io/mattywong/full/NWWajYJ

Seems like the fix is to create a new stacking context on custom-control to get it rendering correctly (ideally z-index, though transform and opacity will also create a new stacking context).

mdo:
For some reason, custom checks in CSS columns need a new stacking order to maintain visibility. This PR adds `z-index: 1` to trigger that new stacking order. Since v5's forms are being redone entirely, and we're dropping IE10/11 there, I think this is totally fine to ship with v4 only. Will need to check v5's forms it in Edge though.

Fixes #29607.

### DIFF
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -9,6 +9,7 @@
 
 .custom-control {
   position: relative;
+  z-index: 1;
   display: block;
   min-height: $font-size-base * $line-height-base;
   padding-left: $custom-control-gutter + $custom-control-indicator-size;


ffoodd:
FYI CSS columns trigger a new containing block, just as `transform` and `perspective`. 

I mentionned the spec in another issue this week (don't remember which) with `transform: scale(1);` as a workaround. Something about stretched links in tables, I think.

Don't know if it needs to be mentionned in the docs, but at least in the PR 😉

ffoodd (approved):

XhmikosR:
@MartijnCuppens LGTY, too?

MartijnCuppens (approved):

dmudro:
@mdo in my app, adding `z:index: 1` to custom control caused a visual bug. imagine a complex form with a dropdown already using `z:index: 1` for its overlay in expanded state (menu); since this the custom control comes after the dropdown in the form, it overlays the expanded menu.

so I traced the source of the issue and found this changed shipped in v4.5.1.

bumping up the z-index on the other component was an obvious fix in my case but touching z-index is always tricky. I wonder how many affected apps are out there and I if there was an alternative solution for solving the original bug #29607.

XhmikosR:
Please make a new issue if you think this still applies.

Issue #29919, opened 2019-12-25
Investigate adding markdownlint


XhmikosR:
Not sure how many FPs we'll get.

mdo:
After fixing bullet inconsistencies...

```
README.md:196 MD045/no-alt-text Images should have alternate text (alt text)
README.md:197 MD045/no-alt-text Images should have alternate text (alt text)
README.md:198 MD045/no-alt-text Images should have alternate text (alt text)
README.md:199 MD045/no-alt-text Images should have alternate text (alt text)
README.md:200 MD045/no-alt-text Images should have alternate text (alt text)
README.md:201 MD045/no-alt-text Images should have alternate text (alt text)
README.md:202 MD045/no-alt-text Images should have alternate text (alt text)
README.md:203 MD045/no-alt-text Images should have alternate text (alt text)
README.md:204 MD045/no-alt-text Images should have alternate text (alt text)
README.md:205 MD045/no-alt-text Images should have alternate text (alt text)
site/content/docs/5.0/about/license.md:12:1 MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
site/content/docs/5.0/components/button-group.md:22 MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h3; Actual: h5]
site/content/docs/5.0/components/button-group.md:22 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above] [Context: "##### Ensure correct `role` and provide a label"]
site/content/docs/5.0/components/buttons.md:108 MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h3; Actual: h5]
site/content/docs/5.0/components/buttons.md:108 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above] [Context: "##### Link functionality caveat"]
site/content/docs/5.0/components/dropdowns.md:805 MD001/heading-increment/header-increment Heading levels should only increment by one level at a time [Expected: h4; Actual: h5]
site/content/docs/5.0/components/dropdowns.md:805 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Above] [Context: "##### `data-toggle="dropdown"` still required"]
[...]

PR #30812, opened 2020-05-14
Markdownlint fixes

mdo:
Addresses some Markdown linting found from testing out `markdownlint` in #29919.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,16 +193,16 @@ Thanks to [BrowserStack](https://www.browserstack.com/) for providing the infras
 
 Support this project by becoming a sponsor. Your logo will show up here with a link to your website. [[Become a sponsor](https://opencollective.com/bootstrap#sponsor)]
 
-[![](https://opencollective.com/bootstrap/sponsor/0/avatar.svg)](https://opencollective.com/bootstrap/sponsor/0/website)
-[![](https://opencollective.com/bootstrap/sponsor/1/avatar.svg)](https://opencollective.com/bootstrap/sponsor/1/website)
-[![](https://opencollective.com/bootstrap/sponsor/2/avatar.svg)](https://opencollective.com/bootstrap/sponsor/2/website)
-[![](https://opencollective.com/bootstrap/sponsor/3/avatar.svg)](https://opencollective.com/bootstrap/sponsor/3/website)
-[![](https://opencollective.com/bootstrap/sponsor/4/avatar.svg)](https://opencollective.com/bootstrap/sponsor/4/website)
-[![](https://opencollective.com/bootstrap/sponsor/5/avatar.svg)](https://opencollective.com/bootstrap/sponsor/5/website)
-[![](https://opencollective.com/bootstrap/sponsor/6/avatar.svg)](https://opencollective.com/bootstrap/sponsor/6/website)
-[![](https://opencollective.com/bootstrap/sponsor/7/avatar.svg)](https://opencollective.com/bootstrap/sponsor/7/website)
-[![](https://opencollective.com/bootstrap/sponsor/8/avatar.svg)](https://opencollective.com/bootstrap/sponsor/8/website)
-[![](https://opencollective.com/bootstrap/sponsor/9/avatar.svg)](https://opencollective.com/bootstrap/sponsor/9/website)
+[![OC sponsor](https://opencollective.com/bootstrap/sponsor/0/avatar.svg)](https://opencollective.com/bootstrap/sponsor/0/website)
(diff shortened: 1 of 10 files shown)


site/content/docs/5.0/components/card.md
```diff
@@ -17,7 +17,7 @@ Cards are built with as little markup and styles as possible, but still manage t
 Below is an example of a basic card with mixed content and a fixed width. Cards have no fixed width to start, so they'll naturally fill the full width of its parent element. This is easily customized with our various [sizing options](#sizing).
 
 {{< example >}}
-<div class="card" style="width: 18rem;">
+<div class="card col-md-6 col-lg-5">
```
XhmikosR, line 20: These seem unrelated to the patch?

XhmikosR:
@mdo did you use my branch https://github.com/twbs/bootstrap/tree/master-xmr-markdownlint? Because there are false positives Markdownlint cannot handle hence why I haven't made a PR to actually add Markdownlint.

site/content/docs/5.0/components/card.md
mdo: Oh damn, yes, will fix.

mdo:
I didn't use your branch—I use markdownlint at GitHub and just ran the CLI against our docs to clear out the easy wins. There are tons of false positives due to our docs tags and what not.

README.md
```diff
@@ -193,16 +193,16 @@ Thanks to [BrowserStack](https://www.browserstack.com/) for providing the infras
...
-[![](https://opencollective.com/bootstrap/sponsor/2/avatar.svg)](https://opencollective.com/bootstrap/sponsor/2/website)
-[![](https://opencollective.com/bootstrap/sponsor/3/avatar.svg)](https://opencollective.com/bootstrap/sponsor/3/website)
-[![](https://opencollective.com/bootstrap/sponsor/4/avatar.svg)](https://opencollective.com/bootstrap/sponsor/4/website)
-[![](https://opencollective.com/bootstrap/sponsor/5/avatar.svg)](https://opencollective.com/bootstrap/sponsor/5/website)
-[![](https://opencollective.com/bootstrap/sponsor/6/avatar.svg)](https://opencollective.com/bootstrap/sponsor/6/website)
-[![](https://opencollective.com/bootstrap/sponsor/7/avatar.svg)](https://opencollective.com/bootstrap/sponsor/7/website)
-[![](https://opencollective.com/bootstrap/sponsor/8/avatar.svg)](https://opencollective.com/bootstrap/sponsor/8/website)
-[![](https://opencollective.com/bootstrap/sponsor/9/avatar.svg)](https://opencollective.com/bootstrap/sponsor/9/website)
+[![OC sponsor](https://opencollective.com/bootstrap/sponsor/0/avatar.svg)](https://opencollective.com/bootstrap/sponsor/0/website)
```
XhmikosR, line 196: I'm not sure if "OC Sponsor" is better than nothing. @patrickhlauke

XhmikosR:
> There are tons of false positives due to our docs tags and what not.

I have disabled some rules in my branch https://github.com/twbs/bootstrap/runs/589127421#step:9:235

mdo:
To clarify, this isn't meant to add a script or Action anywhere—it's just cleaning up some Markdown :). I'll defer to another PR for implementing it and running it as a check for us everywhere.

XhmikosR:
>To clarify, this isn't meant to add a script or Action anywhere

I know, just trying to get the best of the two branches :)

So, you prefer a `-` for ul style?

That being said, let's wait to hear about the `alt` change from @patrickhlauke. Personally I just ignored these specific errors in README.md in my branch and allowed them everywhere else.

README.md
XhmikosR: Let's revert this too @mdo since the patch is only about the dashes for `ul`s.

I will adapt my branch to use dashes too later

XhmikosR:
@patrickhlauke friendly ping :)

README.md
patrickhlauke: so i take it there's no way for us to somehow get that pulled in via third party or something and have it provided with more accurate alt etc? (like a validation badge or similar that actually gets generated separately and dynamically pulled in)

if not, then...yeah, better than nothing i guess

XhmikosR: Unfortunately, I'm not aware of a way. But perhaps we should use `OC sponsor 1` etc?

patrickhlauke: i guess you could. that would still not provide the equivalent of what sponsor it is, but i guess it would at least indicate it's not always the same image. least worst fix, probably

XhmikosR (approved):

PR #28314, opened 2019-02-19
Remove the now unneeded `shrink-to-fit` directive.

XhmikosR:
Redo #27818.

>`shrink-to-fit=no` is not needed anymore - Apple removed the need for it from iOS9.3 onwards

>See https://www.scottohara.me/blog/2018/12/11/shrink-to-fit.html and https://github.com/h5bp/html5-boilerplate/issues/2102

Do we plan to also bump the minimum iOS version @mdo?

Fixes #27861

PR #30827, opened 2020-05-14
🔥 Remove now unnecessary `shrink-to-fit` viewport directive for v4

k-utsumi:
I found out after researching `shrink-to-fit`.

ref.
https://getbootstrap.com/docs/4.5/getting-started/introduction/#responsive-meta-tag
https://github.com/h5bp/html5-boilerplate/pull/2103

---

Redo #28314.

### DIFF
--- a/js/tests/index.html
+++ b/js/tests/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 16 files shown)


XhmikosR:
The browsers that need this are still supported in v4.

k-utsumi:
So that's it. Thank you!

Issue #30531, opened 2020-04-08
Dismissable popover color conflicts with a:not[href])


ffoodd:
Next to [#29140](https://github.com/twbs/bootstrap/pull/29140), we now have a conflicting style in dismissible popovers since it has no `[href]` and its styles rely on a single class specificity:

https://twbs-bootstrap.netlify.com/docs/4.3/components/popovers/#dismiss-on-next-click

I'm not quite sure about the solution for this:
- re-introduce `:not([tabindex])`, which would reintroduce what @ysds wanted to remove;
- use `!important` on `.btn-*` variants (ugh)?
- force color by using `.text-white` in that case?
- anything else? Maybe using colored links or something?

And moreover, is there any other case of links without target styled as buttons?

mdo:
This button is lacking an `href` and should likely just be converted to a `<button>` element. That'd fix the issue I think?

ffoodd:
@mdo Indeed! Noted this quickly while theming things around, but should definetely be a button!

ffoodd:
It's been fixed in v5 and backport in progress to v4.

PR #30830, opened 2020-05-14
Prevent conflicts with components with classes

### LINKED ISSUE #30726: Since v4.4.0, dropdown-item with "active" class without href attribute have black text on hover
Windows 10: Chrome, Firefox, IE, Edge

Bootstrap v4.4.1 test case: https://codepen.io/caseyjhol/pen/ExVoYGg
1. Click button
2. Hover over active item
3. Text is black

Bootstrap v4.3.1 test case: https://codepen.io/caseyjhol/pen/bGVabzZ
1. Click button
2. Hover over active item
3. Text is white

Caused by commit 970c41755a2bdd8f3a4f5dbee039ca7c2dc8b053 (#29140). Obviously a button element could be used instead, but it would potentially require changing a lot of markup, and any existing `a` styling would also have to be applied to `button`. An empty `href` tag could also be used, but that's technically invalid HTML. Perhaps an exception for links with the dropdown-item class could be added?

MartijnCuppens:
Fixes #30726

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -268,7 +268,7 @@ a {
 // causes specificity issues in many other styles that are too complex to fix.
 // See https://github.com/twbs/bootstrap/issues/19402
 
-a:not([href]) {
+a:not([href]):not([class]) {
   &,
   &:hover {
     color: inherit;


mdo (approved):
Love it.

ffoodd:
Also fixes #30531 BTW, when backported tp v4.

PR #30831, opened 2020-05-14
v5: Make navbar active links consistent

### LINKED ISSUE #30652: navbar-nav / nav "active" class inconsistency
Hello!
To make an item active, on a navbar we have to add "active" class to the "nav-item", which I found normal.

But on tabs/pills we have to add the "active" class to the nav-link (not on the nav-item!), which I found strange.

Can I ask what is the logic behind this? Logically, the TAB (nav-item) should be the one selected (active), not the link within (nav-link).

I propose to **ALWAYS** add the active class to the nav-item, as it makes css styling easier. With the current design basically I can't style the selected nav-item, only the nav-link.

mdo:
Two sets of changes:

- Updates the navbar docs to put `.active` on `.nav-link` instead of `.nav-item` like the rest of our navs. Fixes #30652.

- Removes two selectors, `.active > .nav-link` and `.nav-link.show`, because as far as I can tell, those are duplicative and unused, respectively. `.active` just needs to be on the `.nav-link`, and `.show` is always on the parent `.nav-item` because it toggles the `.dropdown-menu`.

Am I missing something with those two selectors? Clicking around on dropdowns in the nav shows no issues thus far.

~~Regarding consistency, I also just noticed that `.btn`s get `.btn.show` for the dropdown-menu though. Do we consolidate and make sure `.show` is always on the `.btn` and `.nav-link` instead of parents?~~ Punted to #30846.


XhmikosR:
> Regarding consistency, I also just noticed that `.btn`s get `.btn.show` for the dropdown-menu though. Do we consolidate and make sure `.show` is always on the `.btn` and `.nav-link` instead of parents?

That part is in the dropdown.js plugin. Better make a new issue if this PR is merged, because the dropdown code is big and we need to be careful.

ffoodd (approved):
LGTM. Don't know why it's here: did you try to go back in Git history for this part?

MartijnCuppens (approved):

PR #30836, opened 2020-05-15
V4 dev backports and updates

XhmikosR:
Preview: <https://deploy-preview-30836--twbs-bootstrap.netlify.app/>


XhmikosR:
@twbs/css-review I pushed a temp patch after dist. Can you check if the IE/Edge workarounds are still fine?

XhmikosR:
So, can someone from @twbs/css-review check my comment above please? https://github.com/twbs/bootstrap/pull/30836#issuecomment-629967644

The "Temp commit is here: https://github.com/twbs/bootstrap/pull/30836/commits/593a63c2c7a28e64b511788ce99214922c67fcc0

ffoodd:

For the `.text-break` class, I think `overflow-wrap` should be `word-wrap`. Am I right, @mdo?

XhmikosR:
>I'd be in favor of removing it (but should probably be another PR, souldn't it?).

Yup, thanks for having a look.

> For the `.text-break` class, I think `overflow-wrap` should be `word-wrap`. Am I right, @mdo?

I just reverted #30074 here.

MartijnCuppens (approved):

PR #30838, opened 2020-05-15
Explain the "dispose" method appropriately

### LINKED ISSUE #27957: Replace use of "destroys" jargon with an appropriate explanation.
default

> I looked at the source code and didn't see what the code did.I am a little confused about the function of this code. Can you give an example in the official manual?

rohit2sharma95:
Closes #27957

Updated the document for the alert component.

In my opinion, it should be done for all the components that have `dispose` method.


rohit2sharma95:
In my opinion, it should be done for all the components that have `dispose` method.

site/content/docs/5.0/components/alerts.md
```diff
@@ -128,7 +128,7 @@ This makes an alert listen for click events on descendant elements which have th
         <code>dispose</code>
       </td>
       <td>
-        Destroys an element's alert.
+        Destroys an element's alert. (Removes stored data on the DOM element and unset the alert's attributes.)
```
XhmikosR, line 131: `unsets`. There's also an unneeded period I believe inside the parens.

rohit2sharma95: @XhmikosR Updated.

site/content/docs/5.0/components/alerts.md
```diff
@@ -1,3 +1,4 @@
+
```
XhmikosR, line 1: This newline is redundant

rohit2sharma95: @XhmikosR Didn't notice that white space. 🤭
Updated again.

site/content/docs/5.0/components/alerts.md
```diff
@@ -128,7 +128,7 @@ This makes an alert listen for click events on descendant elements which have th
         <code>dispose</code>
       </td>
       <td>
-        Destroys an element's alert.
+        Destroys an element's alert. (Removes stored data on the DOM element and unsets the alert's attributes)
```
Johann-S, line 131: We do not unsets the alert's attribute, we just remove the stored data on the DOM element that's all. https://github.com/twbs/bootstrap/blob/master/js/src/alert.js#L78-L81

rohit2sharma95: @Johann-S Updated.

XhmikosR:
That's not how you rebase :P

rohit2sharma95:
My bad @XhmikosR. I actually merged a pull request to update my branch.

Johann-S (approved):

XhmikosR:
@rohit2sharma95 can you please update all the other instance too?

rohit2sharma95:
Sure @XhmikosR 
Let me do it 😊

mdo:
What exactly does `removes stored data on the DOM element` mean? Is there data we're tracking or storing somewhere? My impression is that the entire element was removed, not that stored data was removed from the element.

Johann-S:
Yes we store data in the DOM element so the `dispose` method remove that

Johann-S (changes requested):
The comment for our dispose cannot be the same because on some of our components we remove listeners too

mdo:
If you let me know which one does what in a list here, I can happily update. I'm already riffing on this PR to fix the inconsistency around our methods and options documentation (going all in on the tables instead of a mix of tables and text).

mdo:
@Johann-S Friendly ping—can you list out what is removed for each component? <3

rohit2sharma95:
@mdo These are my findings after reading the code
@Johann-S Please correct me where I am wrong :)

When a constructor is called for any component, it stores the data (the actual instance being created from that constructor) on a private variable (not on the DOM element in real). The association between the private variable and DOM element is done by a key and id, those are registered on the DOM Element.
This data is stored so that the user can get the actual instance of the component by the DOM element and it is being used in the jQueryInterface as well.

So when calling the `dispose` method for any component's instance:
- **Alert:**
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null
- **Button:**
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null
- **Carousel:**
  - Removes event listeners from the DOM Element
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null
- **Collapse:**
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null
- **Dropdowns:**
  - Removes event listeners from the DOM Element
  - Removes this instance from the DOM Element
  - Destroys popper's instance if created for the dropdown
  - Sets the instance's private properties to null
- **Modals:**
  - Removes event listeners from the DOM Elements (from window and document as well)
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null
- **Tooltips and popovers:**
  - Removes event listeners from the DOM Elements
  - Removes this instance from the DOM Element
  - Destroys popper's instance if created for the tooltip or popover
  - Sets the instance's private properties to null
- **Scrollspy:**
  - Removes event listeners from the DOM Element
  - Removes this instance from the DOM Element
  - Sets the instance's private properties to null

**Removes this instance from the DOM Element** means removing this instance from the private variable where it was stored and removing the key and id from the actual DOM element.

Johann-S:
it looks good to me @rohit2sharma95 👍

Johann-S:
Nothing to me, we're ready to merge 👍

Johann-S (approved):

rohit2sharma95:
These commits are there in PR #31780 though

XhmikosR:
Not sure what you mean?

rohit2sharma95:
Sorry, @XhmikosR I mentioned the wrong PR in the comment above. I mean the changes suggested in this PR are kept in the new PR #31337 created by @mdo that is still in progress.

XhmikosR:
True, let's close this one then.

XhmikosR:
Actually wait, I'd rather merge this one and rebase #31337.

PR #30848, opened 2020-05-18
Update iOS app header

mdo:
Hoping to try to get rid of the black bar on my iPhone when saving our docs to the iOS home screen. See https://twitter.com/mdo/status/1262208859608723456 and the replies.

Will open a separate PR for v4 since our docs are much different.


mdo:
Making this change requires modifying the `body`'s `background-color`, so not going to go for it just yet. I could add wrappers, but that just leads to more changes. Maybe there's a better solution, but haven't found it yet.

PR #30866, opened 2020-05-19
Apply utilities in docs styles

ffoodd:
Use our utilities as much as possible—when they're strictly equivalent to the custom styles previously in place.

Preview: <https://deploy-preview-30866--twbs-bootstrap.netlify.app/>


MartijnCuppens (approved):
LGTM!

XhmikosR:
I'll check closely later/tomorrow. I'm not against it, but maybe we should keep some stuff like the `bd-example`. There's a sweet line between using utils and overdoing it.

ffoodd:
Agreed! I kept some cases that introduce a bit of logic (with context, or when utilities wouldn't make sense from a markup only perspective) but I might still have gone too far.

For `.bd-example`, I only applied utilities on the container itself, since what's inside should be as clean as possible for examples to be understood.

Let me know if some things should better be reverted 🙂

FWIW, there're tons of `font-weight: 600` in our docs. It might worth to add a utility on the doc side to handle this (but that seemed unrelated at first glance, may bé another PR).

XhmikosR:
Sorry for the late reply!

I like it minus the examples and btn-clipboard changes. 🙂

The reason is that we include these multiple times per page. What do you guys think?

ffoodd:
Those results in quite verbose markup, I agree.

XhmikosR (changes requested):
Let's remove the examples + btn-clipboard and it should be good to go

ffoodd:
Done :)

XhmikosR (approved):

XhmikosR:
I need an approval from @mdo just to be safe :)

mdo (approved):

Issue #29875, opened 2019-12-18
Changing focus outline for tabindex="-1" has undesired side effects


artopaavola:
[Removing indiscriminate outline suppression for tabindex="-1" elements](https://github.com/twbs/bootstrap/pull/28437/) in version 4.4.0 causes undesired side effects in vanilla js behaviour.

An example: When element.scrollIntoView() is used with focus() there will now be an outline around the element which has been scrolled into view. In case you are wondering, the focus must be programmatically moved to the element so that keyboard focus will also follow the scrollIntoView.

Tested on macOS with Chrome 79, Safari 13.04 and Firefox 71.0

- Focus outline is visible but shouldn't be [CodePen: ScrollIntoView with focus in Bootstrap 4.4.1](https://codepen.io/artopaavola/pen/oNgBBEP)
- Focus outline is not visible which is correct [CodePen: ScrollIntoView with focus in Bootstrap 4.3.1](https://codepen.io/artopaavola/pen/mdyRRaZ)

danielnixon:
`scrollIntoView()` is unrelated to what's happening here. It's the call to `focus()` that is moving focus, which is triggering the focus outline styles.

> the focus must be programmatically moved to the element so that keyboard focus will also follow the scrollIntoView.

Yep exactly. And focus must be visible:

* https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html
* https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F78
* http://www.outlinenone.com/

If you absolutely _must_ remove focus outlines (but why?), consider using the [:focus-visible polyfill](https://github.com/WICG/focus-visible) so focus outlines are only hidden from mouse users, _not_ keyboard users.

artopaavola:
You are right, `scrollIntoView()` is unrelated for the focus styles but I thought the example would make more sense with the scrolling behaviour.

It's clear that focus must be visible for keyboard actionable elements, but I disagree that the focus outline must be visible in my example on the heading where the focus is set programmatically. Are you sure, that WCAG means that focus styles must also be visible on non-keyboard actionable elements? I haven't found anything on that in WCAG. For example, the links to WCAG specs that you provided are only talking about focus styles on keyboard actionable elements.

danielnixon:
> People with attention limitations, short term memory limitations, or limitations in executive processes benefit by being able to discover where the focus is located.

[...]

Issue #26802, opened 2018-07-04
Custom form control box shadow missing on focus when shadows are enabled


jdanil:
When shadows are enabled in bootstrap, focus box shadows are missing for certain form control input states.

For example...
```
$enable-shadows: true;

// _variables.scss
$enable-shadows: false !default;
$custom-checkbox-indicator-indeterminate-box-shadow: none !default;

// mixins/_box-shadow.scss
@mixin box-shadow($shadow...) {
  @if $enable-shadows {
    box-shadow: $shadow;
  }
}

// _custom-forms.scss
.custom-checkbox {
  .custom-control-input:indeterminate ~ .custom-control-label {
    &::before {
      @include box-shadow($custom-checkbox-indicator-indeterminate-box-shadow);
    }
  }
}
```

Testing in [SassMeister](https://www.sassmeister.com/), this evaluates to...
```
.custom-checkbox .custom-control-input:indeterminate ~ .custom-control-label::before {
  box-shadow: none;
}
```

Since `.custom-checkbox .custom-control-input:indeterminate ~ .custom-control-label::before` is more specific, it overrides the focus box shadow in `.custom-control-input:focus ~ .custom-control-label::before`.

This affects `indeterminate` checkboxes and `active` inputs. `checked` appears to be okay, as the `focus` styling is declared after it.

Can I suggest that the default values of the following variables be updated from `none` to `null`, so that they don't override the focus styling unless they are set explicitly?
`$custom-control-indicator-checked-box-shadow`
`$custom-control-indicator-active-box-shadow`
`$custom-control-indicator-indeterminate-box-shadow`
Also the `focus` styling for custom control input should be declared after `active`, and the specificity of the rule should be increased so that it applies to `indeterminate` checkboxes if the box shadows are modified by consumers.

Happy to contribute if you're OK with the proposed solution.

mdo:
Can you verify this is an issue in our latest v5 work from `master`? We've combined our normal checks and custom checks into a single set of form controls.

ffoodd:
This does not apply to v5, since only `:focus` gets a `box-shadow`.

As a matter of fact, I think this should be fixed by defining (and applying) a preferred pseudo-class order. `:focus` should come after `:indeterminate`, in that case.

[Remember "LoVe HAte"?](https://css-tricks.com/snippets/css/link-pseudo-classes-in-order/)—which was later extended to "LoVe Fuck HAte" to include `:focus`?

For form controls, I'd go with:
[...]

PR #30872, opened 2020-05-20
Use `:focus-visible` polyfill

### LINKED ISSUE #27163: Restyle close icon's focus style

ffoodd:
This is a very rough first try to include and use [WICG's `:focus-visible` polyfill](https://github.com/WICG/focus-visible) in Bootstrap. Let me add some background.

## What

[`:focus-visible` belong to CSS Selectors level 4 specification](https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo). As per the spec:

> The `:focus-visible` pseudo-class applies while an element matches the `:focus` pseudo-class and the user agent determines via heuristics that the focus should be made evident on the element.

Thus it's a very helpful selector to massively enhance accessibility. I suggest you read ["`:focus-visible` and backwards compatibility"](https://developer.paciellogroup.com/blog/2018/03/focus-visible-and-backwards-compatibility/) by @patrickhlauke for more explanation and even more background.

## Why

First of all: it's an easy way to safely drop the (sometimes) ugly default `outline`s when activating things using a mouse. This PR, as-is, fixes #27163 and may help with #29875.

For years, Bootstrap has been using `box-shadow` for focus states, which can lead to some confusion since there's also a `$enable-shadows` variable that may cause some focus style to not render appropriately. See #26802 for an example.

But what's more promising here is that it'd allow us to **make focus styles stronger** and more obvious.

It would allow us to increase contrasts on `:focus`, which had been requested a few times (#23329 and #29422).

And I'm quite sure there's probably a few cases with our components where focus could be better. :)

## How

There, I need some help from @twbs/js-review :D

I added the dependency, but this definetly require some JS love to be included in the best possible way. For now, I roughly imported the file form the `node_modules`—which can't obviously stay this way.

On the CSS side:
1. I added a global reset for default `outline`s: you can preview this in our docs navbar for example, which got the default `outline` until now.
2. This is using [the `[data-focus-visible-added]` attribute instead of the `.focus-visible` class since it's environment-proof](https://github.com/WICG/focus-visible/issues/179).
3. And we'll be able to enhance focus everywhere, but that's another story :) FYI in Boosted we chose to use `2px` wide solid `outline`s, and a `transition` on `outline-offset` (which may vary, depending on the component). You can use keyboard to navigate through [Boosted website](https://boosted.orange.com/) to preview this.

## Sidenote

@patrickhlauke mentionned in his post in 2018 the potential use of `@supports`—but it couldn't test for selector support at that time. Well, [it can now](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports#selector) and [the support for, say, `@supports selector(:focus-visible)`](https://caniuse.com/#feat=mdn-css_at-rules_supports_selector) is pretty much the same than [support for `:focus-visible`](https://caniuse.com/#search=focus-visible).

So a future-proof, JS-free way to go might be using `@supports`—but it'd be much, much more verbose on the CSS side.

### DIFF
--- a/js/index.esm.js
+++ b/js/index.esm.js
@@ -16,7 +16,7 @@ import ScrollSpy from './src/scrollspy'
 import Tab from './src/tab'
 import Toast from './src/toast'
 import Tooltip from './src/tooltip'
-import '../node_modules/focus-visible/dist/focus-visible.js'
+import 'focus-visible'
 
 export {
   Alert,
--- a/js/index.esm.js
+++ b/js/index.esm.js
@@ -16,6 +16,7 @@ import ScrollSpy from './src/scrollspy'
 import Tab from './src/tab'
 import Toast from './src/toast'
 import Tooltip from './src/tooltip'
+import '../node_modules/focus-visible/dist/focus-visible.js'
 
 export {
   Alert,
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -34,27 +34,27 @@
     },
     {
       "path": "./dist/js/bootstrap.bundle.js",
-      "maxSize": "51 kB"
+      "maxSize": "53.2 kB"
     },
     {
       "path": "./dist/js/bootstrap.bundle.min.js",
-      "maxSize": "23 kB"
+      "maxSize": "23.1 kB"
     },
     {
       "path": "./dist/js/bootstrap.esm.js",
-      "maxSize": "28 kB"
+      "maxSize": "30.2 kB"
     },
     {
       "path": "./dist/js/bootstrap.esm.min.js",
-      "maxSize": "19 kB"
+      "maxSize": "19.3 kB"
     },
     {
       "path": "./dist/js/bootstrap.js",
-      "maxSize": "29 kB"
+      "maxSize": "30.8 kB"
     },
     {
       "path": "./dist/js/bootstrap.min.js",
-      "maxSize": "16.5 kB"
+      "maxSize": "16.6 kB"
     }
   ],
   "ci": {
--- a/js/index.umd.js
+++ b/js/index.umd.js
@@ -16,7 +16,7 @@ import ScrollSpy from './src/scrollspy'
 import Tab from './src/tab'
 import Toast from './src/toast'
 import Tooltip from './src/tooltip'
-import '../node_modules/focus-visible/dist/focus-visible.js'
+import 'focus-visible'
 
 export default {
   Alert,
--- a/js/index.umd.js
+++ b/js/index.umd.js
@@ -16,6 +16,7 @@ import ScrollSpy from './src/scrollspy'
 import Tab from './src/tab'
 import Toast from './src/toast'
 import Tooltip from './src/tooltip'
+import '../node_modules/focus-visible/dist/focus-visible.js'
 
 export default {
   Alert,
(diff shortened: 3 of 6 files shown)


ffoodd:
I don't fix the Bundlewatch threshold yet, to make the diff obvious: `+767B` for `bootstrap.min.js`, `+3.18KB` before minification.

XhmikosR:
I'm still against adding this. It's extra dependency.

ffoodd:
Indeed.

However it'll allow Bootstrap to make a (big) step further on the accessibility side. Maybe using the polyfill isn't the best option as of now, but I wanted to introduce at least a POC for this concern.

I mentionned an alternative already using **only future-proof CSS** (props to @patrickhlauke):
```css
button:focus { /* some exciting button focus styles */ }

@supports selector(:focus-visible) {
    button:focus { /* undo all the above focused button styles */ }
    button:focus-visible { /* and then reapply the styles here instead */ }
}
```
but this would bloat our styles and wouldn't really work for now—except for Firefox users.

We could also **implement our own polyfill** for this: we already have a polyfill file, and we might not need every case covered by WICG's polyfill (it supports IE11, also supports Shadow DOM, etc.).

I'm not fluent enough on the JavaScript side to determine what can be done, but from both an accessibility and a CSS perspective, I think it's the safer way to go.

As I mentionned previously, this is only a POC and I'd like to get as much advices as possible to see what can be done :)

PigeardSylvain:
Ok it is an extra dependency but it is usefull on all websites.

js/index.esm.js
```diff
@@ -16,6 +16,7 @@ import ScrollSpy from './src/scrollspy'
 import Tab from './src/tab'
 import Toast from './src/toast'
 import Tooltip from './src/tooltip'
+import '../node_modules/focus-visible/dist/focus-visible.js'
```
XhmikosR, line 19: That's not how it should be imported BTW :P

ffoodd: That's pretty much what I meant:
> I added the dependency, but this definetly require some JS love to be included in the best possible way. For now, I roughly imported the file form the `node_modules`—which can't obviously stay this way.

:D

patrickhlauke:
as this shows as closed rather than merged, a silly question...is this now in v5 or not?

just asking because i've long been a proponent of adding this (with the polyfill, with an eye towards future proofing), and now that Chrome shipped it ... it's probably a good time to do it.

ffoodd:
It's not in v5, this was the latest attempt but adding a polyfill seemed an extra effort that we couldn't afford ATM.

FWIW I merged it in Boosted v5 (and it was already in v4 since 4.4.0).

patrickhlauke:
> an extra effort that we couldn't afford ATM

meanwhile we're adding utility classes for every possible CSS property and value... oh well

ffoodd:
I agree, but I think that what was the main argument is to add a dependency: since we dropped jQuery, only Popper remains.

I suggested the polyfill and was the only one in favor of this ATM. Nevertheless, the PR could be restored pretty easily if you dare  since as I said, works fine in Boosted currently.

ffoodd:
@patrickhlauke Just to be sure, are you in favor of using the WICG polyfill, or "simply" the CSS enhancement based on `@supports()`? Maybe if I try a PR without the polyfill, the team would be less relunctant.

patrickhlauke:
could try the `@supports` route, though not sure how verbose that would get. at least it would future-proof things.

PR #30883, opened 2020-05-22
Adding padding not in line with values setting to zero

ghost:
This causes unwanted padding on using the ol, ul for other use cases than lists.

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -168,12 +168,6 @@ address {
 
 
 // Lists
-
-ol,
-ul {
-  padding-left: 2rem;
-}
-
 ol,
 ul,
 dl {


ffoodd:
That's pretty much [what `.list-unstyled` is for](https://twbs-bootstrap.netlify.app/docs/5.0/content/typography/#unstyled). You can also use padding utilities (in that case `.pl-0`) te remove padding.

I don't think default padding for list should be removed.

marco-van-zomeren:
> That's pretty much [what `.list-unstyled` is for](https://twbs-bootstrap.netlify.app/docs/5.0/content/typography/#unstyled). You can also use padding utilities (in that case `.pl-0`) te remove padding.
> 
> I don't think default padding for list should be removed.

Thanks for your feedback. It just feels wrong that it is placed in the reboot file. I would suggest that these settings are removed. Or in any case, move to "type" settings or somewhere else so when a lists is within a <p> it gets the padding, or by a class to give lists some "default" styling.

ffoodd:
FYI, we're going the opposite way in v5: `_type.scss` only concerns added classes. What's in `_reboot.scss` are styles applied to HTML elements.

As a matter of fact, this rule currently only normalizes the value, but removing it won't change anything to your issue since browsers apply an equivalent default styling. You may want to check on [browserdefaultstyles.com](https://browserdefaultstyles.com/).

We may use [AndyBell's reset](https://hankchizljaw.com/wrote/a-modern-css-reset/):
```css
ul[class],
ol[class] {
  padding: 0;
}
```
but it would probably lead to regressions and unpredictable behaviour, since it'd remove `padding` if we use a utility class to only change, say, the list's color.

So I'd recommend you either override this or you simply use the provided utilities.

PR #30897, opened 2020-05-26
Prevent skip links from overlapping header

### LINKED ISSUE #30491: Review skip links theming
In https://github.com/twbs/bootstrap/pull/30414 we discovered the skip links do not look that well if backported to `v4`:
image

We could also improve the theming in v5, in some viewports, it also covers a critical part of the navbar
image

CC @patrickhlauke @ffoodd

ffoodd:
Here it is, after #30073 discussions: I got back the kind of banner that appeared previously, only aligning skip-links to the left.

As a bonus (feel free to discuss it, of course), I enhanced progressively the whole thing by making all skip links appear together when `:focus` is within the container.

And to stay consistent with #30866 I used utilities, seems legit in that case.

Preview: <https://deploy-preview-30897--twbs-bootstrap.netlify.app/>

Fixes #30491

### DIFF
--- a/site/assets/scss/_skippy.scss
+++ b/site/assets/scss/_skippy.scss
@@ -1,13 +1,20 @@
+// stylelint-disable declaration-no-important
+
 .skippy {
-  position: fixed;
-  top: .5rem;
-  left: .5rem;
-  z-index: $zindex-fixed;
-  padding: .5rem;
-  color: $white;
   background-color: $bd-purple;
 
-  &:hover {
+  a {
     color: $white;
+  }
+
+  &:focus-within a {
+    position: static !important;
+    width: auto !important;
+    height: auto !important;
+    padding: $spacer / 2 !important;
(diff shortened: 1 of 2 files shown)


XhmikosR:
Will this solution work for the supported browsers in v4-dev too? Because IE is supported in v4-dev.

ffoodd:
Yes it will, it'll simply work as it did previously: displaying links one by one, at the same place.

IE11 and other unsupporting browsers will simply ignore the `: focus-within` rule.

XhmikosR:
OK, cool, then I'll wait for the other to have a look and if this lands, I'm going to backport it to v4-dev.

mdo (approved):
Looks great!

XhmikosR:
@ffoodd can you backport this and the previous skip links PR if it applies to v4-dev? We have it in the v4-dev-next project.

PR #30909, opened 2020-05-26
Add Parcel Bundler doc

eldk:
Hello,

Not sure if it helps but proposed.

Tested and used with V5 alpha.

Thanks,

Eric

Preview: <https://deploy-preview-30909--twbs-bootstrap.netlify.app/docs/5.0/getting-started/parcel/>


site/content/docs/5.0/getting-started/parcel.md
```diff
@@ -0,0 +1,143 @@
...
+// Helpers
+@import "../node_modules/bootstrap/scss/helpers";
+
+// Utilities
+@import "../node_modules/bootstrap/scss/utilities/api";
+{{< /highlight >}}
+
+## Build app
+
+### Create App  HTML
```
XhmikosR, line 98: Double space here

site/content/docs/5.0/getting-started/parcel.md
```diff
@@ -0,0 +1,143 @@
...
+Include `src/index.js`
+
+{{< highlight html >}}
+<!-- index.html -->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+  </head>
+  <body>
+      <!-- Here 👇 -->
```
XhmikosR, line 109: Use 2-space indentation

site/content/docs/5.0/getting-started/parcel.md
```diff
@@ -0,0 +1,143 @@
...
+{{< /highlight >}}
+
+### Edit `package.json`
+
+Add dev and build scripts.
+
+{{< highlight json >}}
+...
+"scripts": {
+    "dev": "parcel ./src/index.html",
```
XhmikosR, line 122: Use 2-space indentation

site/content/docs/5.0/getting-started/parcel.md
```diff
@@ -0,0 +1,143 @@
...
+    "prebuild": "rm -rf build",
+    "build": "parcel build --public-url ./ ./src/index.html --experimental-scope-hoisting  --out-dir build"
+  },
+...
+{{< /highlight >}}
+
+### Run dev script
+
+App is accessible at http://127.0.0.1:1234
```
XhmikosR, line 131: Wrap the link in code

XhmikosR (dismissed):

XhmikosR:
That being said, I'm not sure if this gives any value? I mean, it's the same procedure as webpack.

eldk:
Hello,
> That being said, I'm not sure if this gives any value? I mean, it's the same procedure as webpack.
Not at all.

Simply use this `'import 'bootstrap' don't work. I also needed to add `as bootstrap` to avoid building errors and have bootstrap js available (I don't remember the case).

You should check with : 

```
if (typeof bootstrap !== 'undefined' || bootstrap !== null) {
    console.log("Popover version : " + bootstrap.Popover.VERSION);
} else {
    console.log("index.js : Bootstrap undefined");
}
```

Thanks,

Eric

XhmikosR:
@mdo WDYT?

Johann-S:
I'm not a big fan of doing a separate page for all the bundlers 🤔

mdo:
I don't mind having separate pages for key tools and more from the ecosystem. If anything it boosts our visibility and usability across multiple communities and gives folks a solid starting point inside our own docs.

mdo:
Pushed a bunch of updates to simplify and standardize:

- Removed some headings, sentence cased some other headings
- Removed the CSS snippets and defer folks to the extensive customize section w/ details on Sass
- Combined and clarified JS import options
- Buttoned up some copy

XhmikosR:
@mdo my only problem is listing this and parcel in Getting Started. We should probably add a parent submenu later.

Branch rebased.

site/content/docs/5.0/getting-started/parcel.md
```diff
@@ -0,0 +1,99 @@
...
+{{< /highlight >}}
+
+### Edit `package.json`
+
+Add dev and build scripts in your `package.json` file.
+
+{{< highlight json >}}
+"scripts": {
+  "dev": "parcel ./src/index.html",
+  "prebuild": "rm -rf build",
```
XhmikosR, line 80: I don't really like `rm -rf` because it doesn't work on Windows.

mdo:
Yeah good point. And while we’re thinking about it, what else could we do for guides like this? Are there more languages and tools?

site/content/docs/5.0/getting-started/parcel.md
eldk: See there https://github.com/parcel-bundler/parcel/issues/1234

XhmikosR: Yes, I'm aware of `shx`, but you don't use it here nor there is a comment that this won't work on vanilla Windows cmd :)

XhmikosR:
@mdo not sure but maybe we should land this after the Windows issue is sorted out, and we add a new parent page later.

XhmikosR:
@eldk care to take care of the `rm` issue?

eldk:
Hello @XhmikosR ,

I have no windows device, so I can't test it. 

I think this should be used : 

`"prebuld": "rmdir /Q /S \"build\"" `

https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/rmdir
https://stackoverflow.com/questions/50696920/configuring-package-json-on-windows

Maybe should someone test it to confirm it's OK ?

Thanks,

Eric

XhmikosR:
@eldk now the docs will only work on Windows 😛 Maybe just go with `npx rimraf build`?

eldk:
> But I think with the npx solution we can make this cross-platform.

That's ok for me.

mdo:
@XhmikosR Do we need this in the beta, or can it wait until v5 stable or v5.1.0?

XhmikosR:
It can wait, but it's ready I think. We could probably create a new subgroup named Bundlers and move Webpack and Parcel and whatever new shows up in the future, or leave it as is for now.

mdo (approved):

Issue #30843, opened 2020-05-17
Atmosphere package for Bootstrap is no longer compatible with Meteor 1.10 release  


sarahmattar:
Meteor 1.10.x has some breaking changes for the version of Bootstrap that is hosted on [www.atmospherejs.com](https://atmospherejs.com/twbs/bootstrap) - it will not allow the package to install because the version of Bootstrap in the package wants to downgrade the version of jQuery that Meteor 1.10 uses to 1.11.1 from 3.0.0. 

For those who are attempting to install Bootstrap for use with Meteor via the [package management system, it is no longer able to be installed. Here is the output from the console when trying to install:

```
 => Errors while adding packages:
While selecting package versions:
error: Potentially incompatible change required to top-level dependency: jquery 1.11.11, was 3.0.0.
Constraints on package "jquery":
* jquery@1.11.9 || 3.0.0 <- blaze 2.3.4 <- blaze-html-templates 1.1.2
* jquery@1.0.1 <- twbs:bootstrap 3.3.6
To allow potentially incompatible changes to top-level dependencies, you must pass --allow-incompatible-update on the command line.
```

This package is long overdue for an update, as it was last packaged with 3.3.7 in November of 2015, but as the official producers of the package, it would be nice to get it upgraded to the latest bootstrap version.

XhmikosR:
Feel free to submit a PR to fix it. We don't use that package manager.

@mdo that's why I insist that we don't support package managers we don't control nor test... Something that isn't tested is bound to break.

sarahmattar:
Okay, I'll look into what needs to be done to package the most recent version of Bootstrap for Atmosphere's package management and then put it in. 

To be fair, this is likely more of a factor of Meteor being bought by Tiny and pushing out a few sizeable releases within the last 6 months after being pretty sparse with the releases in the late 2010's. Atmosphere also stagnated in use and adoption as most developers went to NPM, so it would have been easy to lose track of, even if someone else on the Bootstrap development team packaged it for Atmosphere back in 2015.

ffoodd:
@XhmikosR Just to know, couldn't we add some action on release event or something to test some package managers installation?

sarahmattar:
[...]

PR #30914, opened 2020-05-28
Updates to package.js for Meteor

sarahmattar:
Add the minified versions of Bootstrap's CSS and JS bundle files, which includes Popper.js.
Add source maps for CSS and JS files, as this is giving a warning in the browser console.
Add jQuery as a required dependency from the Meteor core package.

Resolves issue #30843


XhmikosR:
Please format the file back like it was. These changes are not needed.

sarahmattar:
`sarahmattar@Sarahs-MacBook-Pro Desktop % cd bootstrap-test
sarahmattar@Sarahs-MacBook-Pro bootstrap-test % meteor add twbs:bootstrap
 => Errors while adding packages:             
                                              
While selecting package versions:
error: Potentially incompatible change required to top-level dependency: jquery 1.11.11, was 3.0.0.
Constraints on package "jquery":
* jquery@1.11.9 || 3.0.0 <- blaze 2.3.4 <- blaze-html-templates 1.1.2
* jquery@1.0.1 <- twbs:bootstrap 3.3.6

To allow potentially incompatible changes to top-level dependencies, you must pass --allow-incompatible-update on the command line.

sarahmattar@Sarahs-MacBook-Pro bootstrap-test % 

`

This is what happens when I try to create a Meteor project and add Bootstrap to it. This is the crux of the [issue](https://github.com/twbs/bootstrap/issues/30843), and what the file changes are to fix. Reverting the package.js file is not solving the problem, so I am confused as to why you deem the changes to be not needed?

XhmikosR:
The formatting changes are not needed.

XhmikosR:
Alright, I see the problem. master does not need jQuery. The v4-dev package.js has jQuery.

https://github.com/twbs/bootstrap/blob/v4-dev/package.js

Any changes should be made against the v4-dev branch.

XhmikosR:
...unless, there's a way to handle both cases? I mean, does meteor only use the main branch package.js? Because if it does, we need to find a way to handle both cases. The upcoming v5.0 doesn't need jQuery, but all 4.x versions do.

sarahmattar:
Ahhh, Bootstrap 5 Alpha does not require jQuery.

Meteor only uses the `package.js` file to assess what it needs, and as far as I know it can make conditional use of resources based on whether previous versions of Meteor require them, but not previous versions of Bootstrap. As long as there is a `package.js` file in `v4-dev` branch, and the `v4-dev` branch gets published to Atmosphere, then it should take without issue. 

It sounds like publishing the `master` branch would skip Bootstrap 4 altogether and put it up to 5 alpha, so I could go and do a PR against `v4-dev` instead of `master` and you could push that one up to Atmosphere. Does that sound like an optimal solution? 

I also see what you mean about the formatting now in your newer commits - single quotes, no trailing commas. Sorry about that. Prettier must've pulled one over on me on the first save.

package.js
```diff
@@ -2,15 +2,20 @@
...
   git: 'https://github.com/twbs/bootstrap.git'
 });
 
 Package.onUse(function (api) {
   api.versionsFrom('METEOR@1.0');
-  api.addFiles([
-    'dist/css/bootstrap.css',
-    'dist/js/bootstrap.js'
-  ], 'client');
+  api.addFiles(
```
sarahmattar, line 13: I removed the `api.use('jquery');` from the version of `package.js` on the `master`, since v5.0.0-alpha1 does not use jQuery.

sarahmattar:
I've updated the`package.js` file on the `master` branch to remove the line `api.use('jquery');`, as v-5.0.0-alpha1 does not need jQuery as a dependency. I'll start a second PR with the `v4-dev` branch to update it for Atmosphere.

XhmikosR:
Thanks but please try not to change the file's style :)

sarahmattar:
Sorry, I tried to keep it the same: I put an unstaged `.prettierrc` file in place to keep the trailing commas out, single quotes in place, and 2 spaces instead of tabs. Before I put in a second PR for the v4-dev branch, is there something you can recommend so it doesn't happen again?

XhmikosR:
Nope, that should be all :)

That being said, does changing the v4-dev branch package.js even make sense? If Meteor is using the default branch's package.js then it sounds moot. But I'm not sure how else we should do it.

Do note that none of the core team is using Meteor.

sarahmattar:
It would make sense to do the `v4-dev` branch as well because it would add those updated filepaths to the `package.js` file. The Meteor package would then have the bundled, minified JS file, the minified CSS file, and the source maps for Bootstrap 4. 

After that, to effect any change on Meteor's end, the updates we made would need to be [published](https://guide.meteor.com/writing-atmosphere-packages.html#publishing-atmosphere) to Atmosphere. I'm not sure who may have done the original publishing and maintaining - @mdo perhaps? If this is something you'd want me to do and maintain, I'd be up for it, but I think I may have to be added as a collaborator on the [Atmosphere](https://atmospherejs.com/twbs/bootstrap) package. :)

PR #30922, opened 2020-05-29
Disabled checked states wrong background

### LINKED ISSUE #29809: disabled custom checkbox with gradients is not grayed out in checked state
If the custom checkbox is in checked state and if gradients are enabled it has a same look whether it is enabled or disabled. 

Can't reproduce a codepen example, because I can't enabled gradients without compiling the scss.

I think the solution might be adding:
```
background-image:none;
```
to scss/_custom-forms.scss line 138

Tested with:
* bootstrap 4.4.1
* linux
* chromium  78.0.3904.108

ffoodd:
Fixes #29809 for v4.

V5 doesn't need this since `:disabled` simply decreases `opacity` for disabled fields, whatever their `background` is.


mdo (approved):

XhmikosR:
@ffoodd it seems this results in duplicate background properties:


```diff
 .custom-checkbox .custom-control-input:disabled:checked ~ .custom-control-label::before {
   background-color: rgba(0, 123, 255, 0.5);
+  background-color: rgba(0, 123, 255, 0.5);
 }
```

PR #30923, opened 2020-05-29
Ease .btn alignment by switching to inline-flex

### LINKED ISSUE #30837: Wrong vertical align in button links
I saw that the button links doesn't obey to `vertical-align` style. It keep being to top, not in the middle. [Here](https://jsfiddle.net/ebj9L6vx/) is visible this problem.

ffoodd:
Fixes #30837 

Simple PR to try this out. Should probably be backported to v4 if merged.

### DIFF
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -12,6 +12,7 @@
   text-align: center;
   text-decoration: if($link-decoration == none, null, none);
   white-space: $btn-white-space;
+  vertical-align: middle;
   cursor: if($enable-button-pointers, pointer, null);
   user-select: none;
   background-color: transparent;
--- a/scss/_buttons.scss
+++ b/scss/_buttons.scss
@@ -3,7 +3,8 @@
 //
 
 .btn {
-  display: inline-block;
+  display: inline-flex;
+  align-items: center;
   font-family: $btn-font-family;
   font-weight: $btn-font-weight;
   line-height: $btn-line-height;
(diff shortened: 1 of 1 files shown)


scss/_buttons.scss
```diff
@@ -3,15 +3,15 @@
...
+  display: inline-flex;
+  align-items: center;
   font-family: $btn-font-family;
   font-weight: $btn-font-weight;
   line-height: $btn-line-height;
   color: $body-color;
   text-align: center;
   text-decoration: if($link-decoration == none, null, none);
   white-space: $btn-white-space;
-  vertical-align: middle;
```
MartijnCuppens, line 14: `vertical-align` applies to aligning the button to its parent, not aligning its content.

MartijnCuppens:
> Should probably be backported to v4 if merged.

This will probably cause lot's of breaking changes. for example see https://codepen.io/MartijnCuppens/pen/LYpwjXy

ffoodd:
So does it make sense, though?

I think easier buttons alignment is very handy since it's a very common use case—however I'm not sure about the side-effects compared to v4.

mdo:
Inclined to hold on this for the side effects and what not. Let's keep things closer to v4 where we can and see what paradigms we can really break down and reinvent in v6 :).

mdo:
Going to close this out (would be changing to `main` branch anyway post alpha 1) for now. Can always dig it back up and restore if we need to. <3

PR #30927, opened 2020-05-29
Close button consistency across components

### LINKED ISSUE #27163: Restyle close icon's focus style

### LINKED ISSUE #30917: Bootstrap 4.5 or browser(s) bug ? modal close button with frame ?
go to Bootstrap page :

https://getbootstrap.com/docs/4.5/components/modal/
https://getbootstrap.com/docs/4.0/components/modal/

Then click on sample modal dismiss button (right upper corner). See the frame around ? 
**Wasn't there two days ago.**

it is in 
Chome 83.0.4103.61 Win7 64, 
new Edge based on chromium engine, 
latest Firefox 75.

When set css element from
pointer-events:auto
to
pointer-events: none

frame gone, but loose some visual behaviour.

regards.

@XhmikosR, next time check before You close an issue. I Know what ama writing. This problem is new.

ffoodd:
Fixes #30917, fixes #27163 

Set `width` & `height` on `.close` to:
- have a consistent `outline` on focus,
- make close button consistent across components: dropping `padding`s makes all close button the same size, and less dependent about its context.

I'm not convinced about the new `$alert-close-offset` variables but didn't find a cleaner way yet.

Might be valuable to backport to v4, too.


MartijnCuppens:
Thoughts on changing to a em based width/height? Now you need to adjust the width and height if the font size is changed?

> Might be valuable to backport to v4, too.

BC

mdo:
Thoughts on switching to an SVG `x` for more control over sizing?

MartijnCuppens:
> Thoughts on switching to an SVG x for more control over sizing?

Definitely the best option. The cross renders differently on all OS's anyway.

ffoodd:
Very nice ideas! I'm getting on this.

FWIW, my first attempt will be to use SVG-in-CSS (as we do for now everywhere) but would definitely be a case for #30052 too.

ffoodd:
A case for inline SVG: I don't know how to color dissmissible alerts' close button for now. Sticking to CSS, there're only two ways:
- switching to `mask-image` in that case, but it won't work on Legacy Edge—could easily fallback to `background-image` thanks to `@support()`, though
- using `filter`, but it'd be pretty hard to not getting hacky this way.

I'd be in favor of using inline SVG right away, since it'd simply replace the previous markup-based icon and would allow to drop a few more styles.

Any thought?

mdo:
Couple ideas come to mind for embedded SVGs:

- For an `x` on light backgrounds, `#000` and `opacity` can do the trick. We do a bit of this today basically.
- We could apply `filter: brightness()` via modifier class (e.g., `.close-light`) to handle `x`s on dark backgrounds.

Or we could do an HTML-embedded `x` SVG and provide all sorts of options for customization and more.

My gut says that if we do all embedded SVGs everywhere, we do that here. But, since we don't have that decision implemented just yet, I'm thinking we proceed with embedded SVG and provide a modifier class if it's warranted.

(Worth thinking about that changing to HTML-based SVGs really affects upgradeability from v4 to v5. Maybe that comes in v6? Maybe I'm over thinking it...)

ffoodd:
@mdo @MartijnCuppens I managed to inherit the color in some way, here's the trick: using our SVG icon as a `clipPath` through CSS `clip-path(url#path)`.

A few explanations:
1. since `clip-path`, well, clips, we need to set the icon in a pseudo-element to prevent the button from being clipped.
2. to ease button resizing, I needed to change the coordinate system to use `objectBoudingBox` (to basically make path relatives to the object, not absolute): 
   - [Sara Soueidan has a great write-up about `clipPath` in SVG, and more specifically the `clipPathUnits` section](https://www.sarasoueidan.com/blog/css-svg-clipping/#clippathunits)
   - I used [an online tool by @yoksel](https://yoksel.github.io/relative-clip-path/) to convert paths
3. the `clipPath` ID is needed has an anchor at the end of our `url()` function, and must not be escaped: 
   - that required to only set `data:…` in the `$close-icon` variable,
   - and use interpolation around `escape-svg()` to be used in a raw `url()` function,
   - thanks to [@iamvdo and his comprehensive test cases for SVG clips and masks (#C6, specifically)](https://lab.iamvdo.me/css-svg-masks/#testC6) which made me understand this before it drove me crazy.
4. As a matter of fact, using[ `clip-path: url()` is the most supported value](https://caniuse.com/#feat=mdn-css_properties_clip-path) for `clip-path`: it should work down to IE10 and Edge Legacy (not tested yet).
5. Then to use custom colors, it as simple as defining `background: currentColor` on the clipped pseudo-element, and `color: inherit` on `.alert-dismissible .close` :innocent: 

I still need to check through BrowserStack, but I find it promising.

And we may simply drop it when we'll start using inlined SVGs everywhere. 

Any thought?

ffoodd:
Meh, I'll have a look tommorrow since it seems to only work in Firefox at the moment. I probably missed something :/

MartijnCuppens:
@ffoodd, masking is supported in the latest legacy edge, this approach might be way less complicated: https://codepen.io/MartijnCuppens/pen/YzwQqrW

ffoodd:
@MartijnCuppens Just tried with `mask-image`, better but not there yet: it won't work with Edge 16 & 17, and in Edge 18 it's pretty buggy with the `mask` not applying on `:focus` (seems related to `button:focus { outline: 5px auto -webkit-focus-ring-color; }` but don't know why…). I keep the commit for now but will drop it, I guess.

Moreover using both `mask-image` or `clip-path` on the `button`, we're missing the focus style( since it's out of the painted area). May explain why Edge drops the `mask` BTW, to ensure `:focus` visibility.

So for now, both tricks have browser support limitations, and would probably require another element to get both icon and focus right (like a `span`).

Given that, the only way to go with our current browser support is an inline SVG.

I think I'll drop all icon relative things in this PR (to only set size consistently).

Let me know if you find another way around, but the icn would probably be adressed in another PR.

ffoodd:
I revamped this PR (dropped SVG related code and comments) so it's only about sizing now.

mdo:
Closing for #31641.

PR #30930, opened 2020-05-29
Create custom 404 page

techboyg5:
The current 404 page is the default GitHub Pages one. I prefer one which does not use the GitHub branding, and this is a simple one.

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>404 Not Found</title>
+        <style>
+            body {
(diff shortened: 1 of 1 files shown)


mdo:
If we're going to do a custom 404 page, we should do one with Bootstrap's navigation and what not. A new one-off page doesn't change much IMO. Thanks, though!

techboyg5:
> 
> 
> If we're going to do a custom 404 page, we should do one with Bootstrap's navigation and what not. A new one-off page doesn't change much IMO. Thanks, though!

OK. All that I want is a custom 404 page that is not the default one.

techboyg5:
Can we re-consider this?

PR #30935, opened 2020-05-31
width of container

akhildhanesh:
width of container in 1200px was increased from 1140px to 1200px


mdo:
We don't modify the dist files, and there won't be any changes to the container sizes at this time.

PR #30936, opened 2020-05-31
Add role="dialog" in modals via JavaScript

### LINKED ISSUE #30773: Add role="dialog" in the modal via script
As [commented](https://github.com/twbs/bootstrap/pull/30755#issuecomment-626583855) by @patrickhlauke in #30755 

> We could even consider injecting the role="dialog" via scripting (the same way aria-modal="true" is being injected).

`role="dialog"` should be removed from the HTML markup and should be added in the `.modal` via script.

rohit2sharma95:
Closes #30773
Updated documentation and unit tests as well, for the modal component


patrickhlauke (approved):
LGTM

XhmikosR:
@patrickhlauke should we backport it?

patrickhlauke:
we could backport it yes. it's more of a refinement/nice to have - saves authors having to manually add `role="dialog"` themselves, but even if they did/have it in their current code, it will still work

XhmikosR:
Ok, thanks! I'll think about backporting it, but it does seem this is a non-breaking change, i.e. if the attribute already exists, it will be updated.

site/content/docs/5.0/components/modal.md
```diff
@@ -577,7 +577,7 @@ If the height of a modal changes while it is open, you should call `myModal.hand
 
 ### Accessibility
 
-Be sure to add `role="dialog"` and `aria-labelledby="..."`, referencing the modal title, to `.modal`. Additionally, you may give a description of your modal dialog with `aria-describedby` on `.modal`.
+Be sure to add `aria-labelledby="..."`, referencing the modal title, to `.modal`. Additionally, you may give a description of your modal dialog with `aria-describedby` on `.modal`.
```
XhmikosR, line 580: @rohit2sharma95 can you reword this so that we mention that we already add `role="dialog" ourselves?

rohit2sharma95: > Be sure to add `aria-labelledby="..."`, referencing the modal title, to `.modal`. You don't need to add `role="dialog"`, we already add it ourselves via script. Additionally, you may give a description of your modal dialog with `aria-describedby` on `.modal`.

Should I change it as above  @XhmikosR?

XhmikosR: > Be sure to add `aria-labelledby="..."`, referencing the modal title, to `.modal`. Additionally, you may give a description of your modal dialog with `aria-describedby` on `.modal`. Note that you don't need to add `role="dialog"` since we already add it via JavaScript.

rohit2sharma95: @XhmikosR 
Updated. Thanks for your suggestion.

XhmikosR:
Now that I think about it, maybe we should not remove the role dialog on hide? It seems redundant.

/CC @Johann-S @patrickhlauke

patrickhlauke:
doesn't really make a difference, since it's `aria-hidden` on hide. personally i like the idea that the script cleans up after itself, rather than modify the markup on first run and then leaving attributes hanging around that it injected...but it's probably more a philosophical matter and wouldn't be opposed to just leaving it there either.

XhmikosR:
Yeah, it may be better. I was just thinking that it's also better the lest stuff we do with the DOM, the better, that's all :)

PR #30940, opened 2020-06-02
Prevent `.row` from shrinking in flex container

### LINKED ISSUE #30840: row width is not correct
in bootstrap version 4.5:

    <nav class="navbar navbar-expand-md navbar-light">
      <button class="navbar-toggler mr-auto mb-2 bg-light" data-toggle="collapse" data-target="#myNavbar">
        <span class="navbar-toggler-icon"></span>
      </button>
      <div class="collapse navbar-collapse" id="myNavbar">
        <div class="container-fluid border border-success">
          <div class="row border border-danger">
            
            <div class="col-lg-3 bg-info">
				123
            </div>
            

            
            <div class="col-lg-9 bg-warning">
				456
            </div>
			this row must be full width but it isn't.
            
          </div>
        </div>
      </div>
    </nav>

- Operating system and version : Windows 10 17763.1217
- Browser and version: Chrome: Version 81.0.4044.138 (Official Build) (64-bit)

in Bootstrap version 4.1.1 it works fine.

ffoodd:
Fixes #30840 

Working around this issue, this seems the simplest and cleanest way to prevent row from shrinking.

Didn't notice any drawback yet, since it'll only be applied if .row's parent is a flex container.

FYI it'll still break if .row's parent is `display: flex` but doesn't allow wrapping. Don't know if it should be adressed at all, though.

@mdo @MartijnCuppens you know rows and history better than I do, so feel free to drop this if it looks wrong.


mdo (approved):

MartijnCuppens:
Why do we set the `flex-basis` to 100% and than use `flex-grow` to grow anyway? just `flex-grow` should do the trick?

ffoodd:
`flex-grow: 1` does the job, indeed.

I always use the `flex` shorthand [as recommend by the specs](https://www.w3.org/TR/css-flexbox-1/#flex-basis-property). FWIW, after reading [`flex` common patterns](https://www.w3.org/TR/css-flexbox-1/#flex-common), `flex: 1;` would do the job, too.

Do we stick with `flex-grow: 1;`?

MartijnCuppens:
> Do we stick with flex-grow: 1;?

Yeah we should. `100%` doesn't make any sense since the width is actually more  due to the negative margins.

MartijnCuppens:
Scratch that, a `flex-basis` of `100%` might be useful if you're adding multiple `.row`s in a flex container.

MartijnCuppens (approved):

PR #30941, opened 2020-06-02
Apply colors on :disabled, [readonly] .form-controls

### LINKED ISSUE #30890: Safari and Chrome on Mac and iOS didn't apply color for disabled and readonly inputs
Safari and Chrome on Mac and iOS didn't apply color for readonly and disabled inputs, they use internal `-webkit-text-fill-color` property instead of `color`, so you can't customize text color for disabled/readonly inputs.

Because we already have styles for fixing browser inconsistencies for form-colntrols like
* placeholder color in Firefox: https://github.com/twbs/bootstrap/blob/master/scss/forms/_form-control.scss#L43 
* and iOS disabled/readonly color: https://github.com/twbs/bootstrap/blob/master/scss/forms/_form-control.scss#L56 

it seems legit to add correspondent styles to fix the same issues for Chrome/Safari on MacOS/iOS.

It could be fixed by:
```
.form-control {
  ...
  &:disabled,
  &[readonly] {
    -webkit-text-fill-color: currentColor;
  }
}
```

Screenshots of issue:
MacOS disabled form-controls
MacOS readonly form-controls

ffoodd:
Fixes #30890

A sidenote here: this property is extremely powerful and should not be used anywhere else, as [it'll override `color` no matter the specificity nor the source order](https://codepen.io/ffoodd/pen/JjGjPjK).

I'm not sure it's a good thing to introduce it after all, `:disabled` and `[readonly]` inputs aren't bad in gray—but it can be quite confusing for anyone custimizing forms.


mdo:
@ffoodd See my comment at https://github.com/twbs/bootstrap/issues/30890#issuecomment-643558793. It's been a minute, but I don't think we need something super powerful here?

ffoodd:
Indeed, I agree. This property still makes me tink. Let's close this.

PR #30957, opened 2020-06-03
nullify custom forms states' box-shadow

### LINKED ISSUE #26802: Custom form control box shadow missing on focus when shadows are enabled

ffoodd:
Fixes #26802

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -546,7 +546,7 @@ $custom-control-label-disabled-color:           $gray-600 !default;
 $custom-control-indicator-checked-color:        $component-active-color !default;
 $custom-control-indicator-checked-bg:           $component-active-bg !default;
 $custom-control-indicator-checked-disabled-bg:  rgba(theme-color("primary"), .5) !default;
-$custom-control-indicator-checked-box-shadow:   none !default;
+$custom-control-indicator-checked-box-shadow:   null !default;
 $custom-control-indicator-checked-border-color: $custom-control-indicator-checked-bg !default;
 
 $custom-control-indicator-focus-box-shadow:     $input-focus-box-shadow !default;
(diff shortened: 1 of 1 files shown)


mdo (approved):

mkamensky:
For some reason, after this change, I get syntax errors with `sassc-rails`, because of empty `shadow-box` declarations

PR #30959, opened 2020-06-03
Toasts in IE11

### LINKED ISSUE #28341: Toasts and Internet Explorer 11
Having an interesting issue with IE11 and the new Toasts in 4.2.1.
I am using the functionality to centre a toast on the screen (horizontally at least) as found in this page: 
https://getbootstrap.com/docs/4.2/components/toasts/#placement

With IE11, the toast that appears is just over 120px wide, where as in Chrome the toast is 348px wide.

OS: Windows 10 1809 & Windows Server 2012 R2
Browser: IE11
JSBin: https://jsbin.com/suyuguzidu/1/edit?html,js

When testing with JSBin, I found if you use the output pane, it shows correctly in IE11, but if you click the little arrow to pop it out into its own tab, it exhibits the problem.

I found if I set a 'style="width: 350px"' on the div with a class of 'toast' it shows correctly, and is an OK work around for now, but it's not ideal, mostly because if I was to change the scss to have a bigger max-width I would then have to manually change any html file with a toast on it.

ffoodd:
Fixes #28341 

I don't really know what happens here, I tried several known bugs and fixes without any luck. So here's my solution:

1. `flex-basis` seems better than fixing `width` since it has no impact on non-flexible toasts
2. our centered examples couldn't work anyway on IE11 because of [flexbugs #3](https://github.com/philipwalton/flexbugs#flexbug-3)
3. v4-only because IE11.


mdo:
Wondering if we could do `width: 350px; max-width: 100%;`? I believe I used the `max-width` to account for narrow mobile devices.

ffoodd:
All toasts would be 350px wide (or 100%), whereas for now it's fluid under 350px.

Our toasts examples currently don't reach 350px, so they'd become wider.

I think it'd be a good idea to normalize toast size, however it could come back as a bug.

`flex-basis` shouldn't have any side-effect and won't change current behaviour.

Should at least be done on v5 though.

What do you think for v4? Want to give `width` a try?

mdo:
Oh word, yeah that makes sense then lol. Let's skip the `width` thing for now.

site/docs/4.5/components/toasts.md
```diff
@@ -167,7 +167,7 @@ You can also get fancy with flexbox utilities to align toasts horizontally and/o
 
 {% capture example %}
 <!-- Flexbox container for aligning the toasts -->
-<div aria-live="polite" aria-atomic="true" class="d-flex justify-content-center align-items-center" style="min-height: 200px;">
+<div aria-live="polite" aria-atomic="true" class="d-flex justify-content-center align-items-center" style="height: 200px;">
```
mdo, line 170: Why change the `min-height` to `height`?

ffoodd: This is the 2. in the description, should have be clearer: flexbugs #3 concerns IE11 that doesn't justify content correctly when the flex container only has a `min-height`.

The centered example was top-aligned in IE11. Since I don't think it has any bad consequence, I simply switched to `height` 🙂

scss/_toasts.scss
```diff
@@ -1,4 +1,5 @@
 .toast {
+  flex-basis: $toast-max-width; // prevents from shrinking in IE11, when in a flex container https://github.com/twbs/bootstrap/issues/28341
```
XhmikosR, line 2: Nitpicking can you move the comment above the property so that it's not so long? :)

mdo (approved):

axzx:
This change broke the appearance of toasts on Firefox.
Now, instead of a single toast matching the height of the alert I have in the center, a 350x350px square is displayed.

obraz

XhmikosR:
@axzx please make a new issue. Commenting in old issues is moot.

PR #30965, opened 2020-06-04
Floating labels examples in IE

### LINKED ISSUE #30964: Revisit floating example for IE/Edge

_Originally posted by @ffoodd in https://github.com/twbs/bootstrap/pull/30836#issuecomment-638711684_

ffoodd:
**IE11**

While checking #30836, I came across floating labels in IE11 that could just work fine since [it supports a vendor prefixed version of `::placeholder-shown`](https://caniuse.com/#feat=css-placeholder-shown) and Autoprefixer just does it job…

Preview: https://deploy-preview-30965--twbs-bootstrap.netlify.app/docs/4.5/examples/floating-labels/

**Edge**

Moreover the Edge workaround uses `display: none` to hide labels, which is a pitty from an accessibility perspective.

However Edge also concerns v5, so I'll target `master` for this and it should be backported late (Xref: #30966).


ffoodd:
Well, I don't get it but Autoprefixer doesn't seem to apply here. Don't know why it did in #30836.

XhmikosR:
@ffoodd it's normal, it's because in my v4 backport and updates PR I updated autoprefixer.

Let's get that merged first :)

mdo (approved):

PR #30966, opened 2020-06-04
Floating labels improvements

### LINKED ISSUE #30964: Revisit floating example for IE/Edge

ffoodd:
Next to #30965 — for v5 concerns.

1. Edge fallback can be accessible by not making labels `display: none` and simply switching to flex reversed layout…
2. Refactored things a bit since we have plenty of vendor prefixes in this file, whilst none of our supported browser require it.

Preview: https://deploy-preview-30966--twbs-bootstrap.netlify.app/docs/5.0/examples/floating-labels/

### DIFF
--- a/site/content/docs/5.0/examples/floating-labels/floating-labels.css
+++ b/site/content/docs/5.0/examples/floating-labels/floating-labels.css
@@ -23,13 +23,13 @@ body {
   margin-bottom: 1rem;
 }
 
-.form-label-group > input,
-.form-label-group > label {
+.form-label-group input,
+.form-label-group label {
   height: 3.125rem;
   padding: .75rem;
 }
 
-.form-label-group > label {
+.form-label-group label {
   position: absolute;
   top: 0;
   left: 0;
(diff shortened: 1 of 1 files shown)


mdo (dismissed):

mdo:
Testing this on Netlify in IE11 isn't super clear to me... that's likely because I'm looking at v5's code that doesn't support IE at all anymore, but the example doesn't work there anymore. Signed off on this earlier, but going to retract and punt to alpha 2 to confirm a few things first <3.

XhmikosR:
Could be because the branch was behind, have another try @mdo. Note that this PR is for Edge not IE since IE is not supported on v5.

site/content/docs/5.0/examples/floating-labels/floating-labels.css
```diff
@@ -43,18 +43,6 @@ body {
   transition: all .1s ease-in-out;
 }
 
-.form-label-group input::-webkit-input-placeholder {
```
XhmikosR, line 46: Better put these back. The prefixes are automatically added based on our browserslist config.

XhmikosR (changes requested):

XhmikosR (approved):

PR #30912, opened 2020-05-27
Fixed Navbar documentation

psjishnu:
Fix #30880: .nav-item is not required on .nav-link inside .navbar-nav

Issue #30880, opened 2020-05-22
.nav-item is not required on .nav-link inside .navbar-nav


rohit2sharma95:
When list-based approach is not used for the navbar, class `.nav-item` is not required on `.nav-link` inside `.navbar-nav`. Link for the [docs](https://getbootstrap.com/docs/4.5/components/navbar/#nav).

See [this comment](https://github.com/twbs/bootstrap/issues/28423#issuecomment-630286599) in #28423

XhmikosR:
Not sure what you mean by this issue? Is it a documentation issue?

rohit2sharma95:
Yup @XhmikosR. I mean docs should be updated and `.nav-item` class should be removed from `.nav-link`. (Only when the list-based approach is not used for the navbar)

XhmikosR:
Fixed in #30973.

Issue #28117, opened 2019-01-24
Documentation - nav-justified


retan:
The documentation for `.nav` components has bug.
The first example for `.nav-justified` isn't working at all.
https://getbootstrap.com/docs/4.2/components/navs/#fill-and-justify

![image](https://user-images.githubusercontent.com/3779428/51700128-a3212480-200e-11e9-84b4-559073cc8d22.png)

My guess is that the example should use the `ul>li.nav-item` structure.

XhmikosR:
Yeah something is wrong either with the docs or the CSS.

/CC @MartijnCuppens

XhmikosR:
I believe it's a docs issue though, since by design this needs `.nav-item` so that flexbox works.

PR #30973, opened 2020-06-05
Remove .nav-item from .nav-link to make it consistent

### LINKED ISSUE #28423: Non ul or ol based .nav-link should require .nav-item class

rohit2sharma95:
Remove the class `.nav-item` from `.nav-link` inside the non `ul` or `ol` based navs.
This makes the consistency for `.nav-item` (This class will not be required on the `.nav-link`).
`.nav-item` was only required when you use `.nav-fill` or `.nav-justified` on the `.nav`. In other cases, it was redundant. And the purpose of `.nav-item` on `.nav-link`s can be achieved via flexbox utilities as well (Mentioned in the docs also).

Closes #28423
Related #30880, #28117
(This PR covers #30912 as well)


mdo (approved):

site/content/docs/5.0/components/navs.md
```diff
@@ -187,14 +187,14 @@ Force your `.nav`'s contents to extend the full available width one of two modif
 </ul>
 {{< /example >}}
 
-When using a `<nav>`-based navigation, be sure to include `.nav-item` on the anchors.
+When using a `<nav>`-based navigation.
```
XhmikosR, line 190: This seems off now

rohit2sharma95: Unable to understand @XhmikosR 😊
Can you please define?

XhmikosR: The sentence seems weird with that part removed :) Maybe it should be reworded?

rohit2sharma95: Or you can even use a `<nav>`-based navigation.
@XhmikosR?

mdo, line 190:
```suggestion
When using a `<nav>`-based navigation, you can safely omit `.nav-item` as only `.nav-link` is required for styling `<a>` elements.
```

PR #30979, opened 2020-06-08
`min-width: 0` on .col has too much side-effects

### LINKED ISSUE #30852: Default Col min-width: 0 affects responsive behavior
After upgrading from Bootstrap 4.4 to 4.5 I noticed cards within columns which used to respond responsively when the screen size is made smaller, now become stacked (i.e. overlap).

It appears to be related to a CSS change in 4.5 where the col min-width defaults to 0. https://github.com/twbs/bootstrap/pull/30049

Effect can be viewed in this example by simply changing the CSS bootstrap setting between 4.4.1 (responsive) and 4.5 (stacked) https://codepen.io/alonergan/pen/BaoGqwO 

OS: Win 10
Browser: Chrome Version 81.0.4044.138 (Official Build) (64-bit)

ffoodd:
Closes #30852

For some reason I cannot revert the merged commit. So let's simply drop it.

Flexible items are flexible, so in order to fix what #30049 tried to fix: either use fixed width on `pre`, or let it wrap by using:

```css
.col > pre {
  white-space: pre-wrap;
}
```


mdo (approved):

MartijnCuppens (approved):

poebrand:
@ffoodd ,

According to https://github.com/twbs/bootstrap/releases , the last 4.x release was May 12th. However, this fix was put in place on June 9th.

Is it possible to get another 4.x release so this fix can go out? We just ran into the problem as well.

PR #30981, opened 2020-06-08
Fix undefined custom properties in reboot

MartijnCuppens:
Custom properties like
https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap-reboot.css#L16

are not available in reboot.


ffoodd (approved):

ffoodd:
Nice catch!

PR #30985, opened 2020-06-09
V4 dev backports and updates

XhmikosR:
I could use some help updating `@rollup/plugin-node-resolve`, not a blocker but we should do it eventually (same on master).

Preview: <https://deploy-preview-30985--twbs-bootstrap.netlify.app/>

Before merging:

- [x] Adapt v4-dev branch protection rules


XhmikosR:
@mdo should I point the license headers to v4-dev too?

ffoodd (approved):

PR #30989, opened 2020-06-09
Ensure to increase contrasts on buttons & colored links hover/active states

ffoodd:
Alongside #30972 — and in #30548 roadmap — this ensures that hover and active states have higher contrasts than the default state:

1. for colored links it makes obvious that we have to find a clean way to add darkish background behind some of them.
2. the warning button's shadow feels a bit weird, not quite sure about what to do with it for now.

Preview: [Buttons](https://deploy-preview-30989--twbs-bootstrap.netlify.app/docs/5.0/components/buttons/) — [Colored links](https://deploy-preview-30989--twbs-bootstrap.netlify.app/docs/5.0/helpers/colored-links/)

Since this is only visual, should we consider to backport this to v4?


patrickhlauke:
will this also apply to the toggle buttons https://v5.getbootstrap.com/docs/5.0/forms/checks/#toggle-buttons ? in particular, i find the outline styles ones confusing when hovering with the mouse, as at that point you can't tell at all whether the toggle is on or off https://v5.getbootstrap.com/docs/5.0/forms/checks/#outlined-styles - likely, we'd need a separate hover style for non-toggled buttons (probably one that is still outlined, rather than filled, so that the toggle state is visually clear on hover)

ffoodd:
@patrickhlauke This PR bascially only choose between `darken()` and `lighten()` for background depending on the foreground color (if it's light, the background darkens; it lightens otherwise).

So it'll apply to toggle buttons, but won't improve what you're talking about, I guess.

patrickhlauke:
no worries, i'll file a separate issue then :)

patrickhlauke:
posted the separate issue here #31149

mdo (approved):

XhmikosR:
This was merged wrong it seems.

Issue #30849, opened 2020-05-18
Cannot use javascript MouseEvent on checkbox button in 4.5


mrlife:
After upgrading from 4.4.1 to 4.5, the checkbox button no longer allows an onclick event on the checkbox. Reverting back to 4.4.1 fixes it. I believe the issue is from the `event.preventDefault()` (line 163) added in this pull request: https://github.com/twbs/bootstrap/pull/30388.

Here is a repro comparison with checkbox button `test1` and regular checkbox `test2`.

4.4.1 (works): https://www.codeply.com/p/9DhKAgjsCP
4.5: https://www.codeply.com/p/v1lNDIWenF

OS: macOS 10.15.4
Browsers: Chrome 81.0.4044.138, Safari 13.1

Lausselloic:
I've tried to avoid preventing the event onclick onto the input, as a workarround you could catch the onclick event onto the label instead of the input waiting for the next release

PR #30992, opened 2020-06-09
avoid preventing input event onclick

Lausselloic:
instead of stopping event if onclick is triggered on input, call toggle method only if its not on checkbox inside a label

fix https://github.com/twbs/bootstrap/issues/30849 following regression introduced by https://github.com/twbs/bootstrap/pull/30388

### DIFF
--- a/js/src/button.js
+++ b/js/src/button.js
@@ -153,10 +153,9 @@ $(document)
         return
       }
 
-      if (initialButton.tagName === 'LABEL' && inputBtn && inputBtn.type === 'checkbox') {
-        event.preventDefault() // work around event sent to label and input
+      if (initialButton.tagName !== 'LABEL' || inputBtn && inputBtn.type !== 'checkbox') {
+        Button._jQueryInterface.call($(button), 'toggle')
       }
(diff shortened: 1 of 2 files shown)


XhmikosR:
@Lausselloic thanks! Would it be possible to add a test case? Because this wasn't caught there.

Lausselloic:
I don't make the initial PR to the master branch, so I need to PR the initial fix and that one to master

XhmikosR:
OK, if it applies to master too yeah. But we need a test case for both either way to make sure this doesn't happen again :)

Lausselloic:
unit test added, need to port patrick fix (https://github.com/twbs/bootstrap/commit/24abed13369469c174acbec5e5939e252b29936f) and my PR's to master

XhmikosR:
Are you sure the unit test covers the opposite too?

Thanks for the patches, and waiting for the master PR later :)

Lausselloic:
no it just test if event is propagate to the input wrong unit test title. I could add one to ensure event is also emit on label (I will update my PR)

XhmikosR (approved):

PR #30996, opened 2020-06-09
[Suggestion] ✨ Adjust container and add xs size

k-utsumi:
Try and Suggestion PR for below issue for v5.
https://github.com/twbs/bootstrap/pull/30969

v4 `.container, .container-[sm ~ xl]` sample
https://jsfiddle.net/p2s1f9zn/

I don't understand the difference in behavior,
so I tried changing to a container class with different max-width.

Also, add the xs width to be used on the login screen, etc.

### DIFF
--- a/scss/_containers.scss
+++ b/scss/_containers.scss
@@ -4,18 +4,19 @@
 
 @if $enable-grid-classes {
   // Single container class with breakpoint max-widths
-  .container {
-    @include make-container();
-    @include make-container-max-widths();
-  }
-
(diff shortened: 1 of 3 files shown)


mdo:
No plans to change these or add another breakpoint at `xs`—we consider the default classes our "xs" tier.

Here's the container classes in effect. Basically they are 100% wide until a breakpoint.

Screen Shot 2020-06-09 at 5 29 31 PM

PR #30997, opened 2020-06-10
Update Actions

XhmikosR:
* skip BrowserStack if we are not on the main repo
* remove the action filename from cache


XhmikosR:
I'm still not sure about fork PRs, we'll see how this will go. This is an issue we are having since the switch to Actions.

mdo (approved):

PR #30998, opened 2020-06-10
Examples: a few fixes

ffoodd:
- Fix Headings hierarchy in Examples list: easy to fix, no drawback
- Drop redundant alternative on screenshots: block links were labelled like "Album screenshot Album" which is annoying
- Pricing example layed on `.card-deck`, which has been dropped in v5.

The first two points could be backported to v4, not quite sure about its usefullness however since those are pretty low priorities— @patrickhlauke, any opinion?


mdo (approved):

XhmikosR:
Not sure about the `alt` change, the rest look good to me though. @patrickhlauke WDYT?

mdo:
Merged today to clear out the v5 project as much as possible. Sorry for not waiting, but if there's more, happy to help tackle it in another PR <3.

Issue #30924, opened 2020-05-29
Switch button firing change event 3 times in BS 4.5.0


dkwebtec:
I have tried this issue under version 4.5.0 and the method seems working but the change event is fired 3 times on each button click

[https://codepen.io/DK568/pen/MWYxWMO](https://codepen.io/DK568/pen/MWYxWMO)

try to click the button

_Originally posted by @dkwebtec in https://github.com/twbs/bootstrap/issues/30077#issuecomment-636015174_

XhmikosR:
@Lausselloic is this the same issue as #30849?

Lausselloic:
it's not the same issue but yes event is propagate to childrens inside the label. I could give a try to handle that

mdo:
#31000 targets `v4-dev`, but what about this change in v5?

XhmikosR:
@Lausselloic can you confirm please that this issue does not affect v5? I believe we removed most of the Button JS so this doesn't seem to apply to v5.

Lausselloic:
Sorry for the late reply, that doesn't affect v5 at time, there's a need to backport @patrickhlauke fixes (https://github.com/twbs/bootstrap/pull/28952 and https://github.com/twbs/bootstrap/pull/28834) those accessibility fixes are not present's in v5 I will give a try to port it

XhmikosR:
Fixed by #31000, thanks @Lausselloic!

PR #31000, opened 2020-06-10
buttons plugin : avoid multiple change event trigger

Lausselloic:
Fixes https://github.com/twbs/bootstrap/issues/30924

- add unit test to count how many event are thrown when widget contains multiple tags inside label
- add a parameter to toggle, if click event is provided onto an input then don't trigger another change event already thrown by the browser
- simplify the case where toggle interface is called click provide from input itself OR it's a button without label. If label is present, then browser propagate clic event from childrens thru label and then cause multiple calls to toggle
- the test assume that `.btn` class is always set onto the label if there's one, otherwise need to update this plugin and look for label arround the input

Test with keyboard, mouse and js clic call


mdo:
@Johann-S Any chance you could review this PR this week? Thanks!

js/src/button.js
```diff
@@ -86,7 +86,9 @@ class Button {
             input.checked = !this._element.classList.contains(CLASS_NAME_ACTIVE)
           }
 
-          $(input).trigger('change')
+          if (!avoidTriggerChange) {
+            $(input).trigger('change')
```
Johann-S, line 90: I'm not a big fan of this parameter (`avoidTriggerChange`) I would prefer a class attribute.

For exemple here, when we trigger the change event and if it's an input element, we would set a class attribute to avoid to trigger too much events.

What do you think @Lausselloic ?

mdo:
Friendly bump to @Lausselloic :).

js/src/button.js
Lausselloic: Why not, I could add a `CLASS_NAME_CHANGING` to the button element instead of using a boolean parameter

Lausselloic: @Johann-S I let you check the update if it fit what you think

Lausselloic: Ok my bad too, I've updated the PR and create a class attribute shouldAvoidTriggerChange updated data value thru _jQueryInterface I think that's the idea

Johann-S (approved):
LGTM thanks @Lausselloic 👍

PR #31004, opened 2020-06-10
fix some missing prefix

Lausselloic:
- add missing `bs-` prefix when using Manipulator
- fix wrong data-content attribute
- fix data-original-title in reboot.scss


Lausselloic:
@XhmikosR i've fix data passed thru javascript and update the documentation.
Still have a question around tooltip content config option present only in unit test and not in documentation. Maybe an old param? Otherwise I think the rest is okay

XhmikosR:
I'm going to merge it and rebase my branch and see where we are left. Thanks for the PR!

PR #31025, opened 2020-06-14
Feature z-indez level - utils

MauricioHernanCabrera:
added z-index level

```scss
"z-index": (
      property: z-index,
      values: (
        0: 0,
        1: 1,
        2: 2,
        3: 3,
        4: 4,
        5: 5,
      ),
)
```

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -497,6 +497,17 @@ $utilities: map-merge(
         visible: visible,
         invisible: hidden,
       )
+    ),
+    "z-index": (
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Thanks @MauricioHernanCabrera, but this needs some more documentation. Could you please add that?

mdo:
I feel like it'd be more appropriate to use the z-index scale we have for our components instead of 0-5. This could help standardize the layers we use in our components and what not.

PR #31027, opened 2020-06-14
added shapes like right arrow left arrow and X Close button

xtianmarucco:
added shapes like right arrow left arrow and X Close button using clip-path tool to create basic forms with css.


mdo:
This is interesting, but given we have Bootstrap Icons, we'll pass on this. Thanks though!

xtianmarucco:
thanx Mark, I'll keep thinking in some new ideas for features

PR #31029, opened 2020-06-14
Add .rotate-* utilities

MauricioHernanCabrera:
Added

```scss
"rotate": (
      property: transform,
      class: rotate,
      values: (
        0: rotate(0deg),
        45: rotate(45deg),
        90: rotate(90deg),
        135: rotate(135deg),
        180: rotate(180deg),
        225: rotate(225deg),
        270: rotate(270deg),
        315: rotate(315deg),
      )
)
```

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -497,6 +497,20 @@ $utilities: map-merge(
         visible: visible,
         invisible: hidden,
       )
+    ),
+    "rotate": (
+      property: transform,
+      class: rotate,
+      values: (
+        0: rotate(0deg),
+        45: rotate(45deg),
+        90: rotate(90deg),
+        135: rotate(135deg),
+        180: rotate(180deg),
+        225: rotate(225deg),
+        270: rotate(270deg),
+        315: rotate(315deg),
+      )
     )
   ),
   $utilities


scss/_utilities.scss
```diff
@@ -497,6 +497,20 @@ $utilities: map-merge(
         visible: visible,
         invisible: hidden,
       )
+    ),
+    "rotate": (
+      property: transform,
+      class: rotate,
+      values: (
+        0: rotate(0deg),
```
XhmikosR, line 505: Can't this just be `0`?

MauricioHernanCabrera: yeah, you right

MartijnCuppens: Even just `none`

MartijnCuppens:
This one also needs some docs.

mdo:
Going to leave this open for feedback before we think about merging. From my own experience, rotations aren't as common as many other `property-value` pairs.

MauricioHernanCabrera:
Okay thanks, Where can I find any example of that?

MartijnCuppens:
> Going to leave this open for feedback before we think about merging. From my own experience, rotations aren't as common as many other property-value pairs.

Accordions usually have them, our own docs sidebar have them too. I have also used them for dropdowns lately.

Not sure for the 45deg (and other "diagonals") transforms yet here.

mdo:
@MartijnCuppens I can dig that. What are we thinking for values then? 0, 90, 180, 270?

MartijnCuppens:
@mdo, Yeah, that might have been a better way to explain my thoughts 😄

ffoodd:
Just a sidenote here: we should mention in the docs that it'll prevent to add other `transform` functions alongside `rotate()`.

I'm very used to define several `transform` functions altogether, so I'm pretty sure this will get back to us as an issue if this PR gets merged.

mdo:
Indeed, @ffoodd. This PR will need some documentation to go with these new utilities and we'll need to make a mention of conflicting transforms.

mdo:
Punting on these for now. Thanks though!

PR #31035, opened 2020-06-15
Add null variables for nav-link

### LINKED ISSUE #30847: No scss variable to change color for a .nav-item inside a .nav
- Windows 10
- Chrome  81.0.4044.138
- Bootstrap 4.7.0

After fiddeling around with a vertical nav, i found out there is no scss variable to change the color of these. There is `$link-color` but that changes all the links. I expected there to be a `$nav-color` or something. A nav is different than a navbar, and also different than just normal links. So should there be a scss variable to change the color, hover etc for an a href inside `.nav-item` thats inside `.nav`

mdo:
Replaces #30498 by adding four new null default variables for .nav-link. Doesn't carry over font-style from the original PR though since that's rarely used, at least by default Bootstrap. Nullifies all values from that PR, too, since we count on some basic inheritance here and don't need color by default.

Fixes #30847.


mdo:
Do we backport? Thinking maybe we could actually.

mdo:
Nixing the backport for now assuming we're doing v4.5.x and not a v4.6.x release.

PR #30779, opened 2020-05-11
v5: Port new homepage changes

mdo:
Could use some refinement on the sizing and spacing IMO. Will sit on it for a couple days maybe. Feedback welcome though!

Preview: https://deploy-preview-30779--twbs-bootstrap.netlify.app/

PR #31042, opened 2020-06-16
v5: Update favicons

mdo:
Will require a rebase before opening as full PR after #30779 is merged.


mdo:
I'll update our brand page too during our second alpha. Fine to leave as-is until then as tweaks will likely happen here and there.

XhmikosR:
Some quick notes:

1. The new icons need to have the exact same dimensions like before. For example android-chrome-192x192 is 256x256 etc
2. favicon.ico is fine now regarding its size, but it looks bad, pixelized

What's the original SVG that was used? I still don't think we should reinvent the wheel, so I'd like to generate them using favicon generator and just exclude any favicons we might not need.

mdo:
Can't approve my own PR, but looks great!

XhmikosR (approved):

PR #31045, opened 2020-06-16
Fix a broken link and a redirect.

XhmikosR:


mdo (approved):
Looks good, but that link will conflict with the `master` > `main` rename.

PR #31047, opened 2020-06-16
README: remove `.svg` from shields.io badges

XhmikosR:


XhmikosR:
BTW it seems the gzip size badges are broken since the master -> main change, but from their docs it seems we do it right. https://github.com/ngryman/badge-size#filepath

Johann-S (approved):

PR #31050, opened 2020-06-16
Rename "master" to "main".

XhmikosR:


mdo (approved):

build/generate-sri.js
```diff
@@ -7,7 +7,7 @@
  *
  * Copyright 2017-2020 The Bootstrap Authors
  * Copyright 2017-2020 Twitter, Inc.
- * Licensed under MIT (https://github.com/twbs/bootstrap/blob/master/LICENSE)
```
MartijnCuppens, line 10: Will this mean all existing licence links to master will be broken?

XhmikosR:
Unfortunately, yeah. GitHub doesn't do redirects.

build/generate-sri.js
mdo: Thinking this is fine since the license name is there (MIT) and the copyright holders are there. We can also fix this in v4.5.1 for v4 and at least end on the right note.

patrickhlauke:
Lovely work

PR #31067, opened 2020-06-17
data: reword storage key in DOM

Johann-S:
I was migrating one of my project to BS 5 and React use `key` as attribute a lot, so there was a conflict with us.
Which is why I renamed `key` --> `bsKey`

### DIFF
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -16,22 +16,22 @@ const mapData = (() => {
   let id = 1
   return {
     set(element, key, data) {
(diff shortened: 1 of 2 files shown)


XhmikosR:
I like it, but I think it's a good chance to prefix all our data attributes. I have a WIP branch so it would be nice if we could finish that too.

Johann-S:
it's not a data attribute here, it's an object attribute

XhmikosR:
I know but still I could use help with that.

XhmikosR (approved):

PR #31071, opened 2020-06-17
Docs: clarify rationale for button group on links

patrickhlauke:
- clarify that putting links in a button group is, in essence, an alternative to using the navigation component (and cross-reference it)
- remove the non-sequitur about using `.active`...that's irrelevant here, as it's already mentioned in the docs page for `.btn`. otherwise, this would then also need the whole extra explanation about conveying meaning with `aria-current` etc, which is already covered in ample detail in the navs page


MartijnCuppens (approved):

XhmikosR:
@patrickhlauke this needs to be backported manually, I don't even know if we have those snippets in v4-dev.

XhmikosR:
This doesn't seem to apply to v4-dev. If you still think it does, re-add it in the v4-next project and make a PR against v4-dev.

patrickhlauke:
ah you're right, could have sworn v4 also had a link button group example, but i must have dreamt it

Issue #31057, opened 2020-06-16
Beta/Stable v5 tasks


mdo:
Tracking for before we go stable with v5—we'll need to update `robots.txt`.

TODO:

- [x] Redirect `v5.getbootstrap.com` pages or add a callout at the top of each page
- [ ] Make sure there are no page overrides on v4-dev docs; we should be careful after we move v5 docs to the main site and when we release the next 4.x version
- [ ] Unmark RTL as experimental?

Beta 1:

- [x] ~~Revert 723cd8d12bdd48f389234b5fc40c3b1789807b4d (not blindly)~~ - Done in #32255
- [x] ~~Revert #32099~~ - Done in #32255
- [ ] ~~Remove unneeded docs redirects? We may skip this~~

Stable:

- [ ] Revert fc74a0d91589c120bf8b9a3913fbceb016733c61 (when we switch to the `latest` npm tag)
- [ ] Revert #32194

mdo:
Adding to both v5-stable and v5-beta in case we do what we did with v4 and put the beta docs on the main site.

XhmikosR:
For reference https://github.com/twbs/bootstrap/commit/723cd8d12bdd48f389234b5fc40c3b1789807b4d#diff-d7f6c1ca03bd330a561f8e674d577ee8

We'll need to revert all those changes when we are ready to move the v5 docs to the main repo.

Curious-Ujjwal:
I want to highlight the common issue I am facing. 
The recent version of Bootstrap CSS as specified in this [link](<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.3/css/bootstrap.min.css" integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">) is either behaving badly or not working properly.

Please pay some attention to this.

Curious-Ujjwal:
HI
I want to highlight the issue I am facing with the current CSS link on the Bootstrap website, which is as given below:
<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.3/css/bootstrap.min.css" integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">

[...]

PR #31075, opened 2020-06-17
Updated the command to install using yarn

mattonit:
Currently, the download page for v5-alpha is showing the command to install Bootstrap via Yarn, but this will install 4.5.0 version while the other ways are showing how to install BS 5.0-alpha. I'm assuming the yarn one was overlooked and you guys wanted to show us how to install BS 5.0-alpha. This can be simply fixed with this PR


XhmikosR:
Thanks, I need to check if I missed other places too.

mdo:
@XhmikosR Maybe also get an issue going for stable/beta when we need to update these again to drop the `@next` etc?

XhmikosR:
@mdo I'm already tracking this in https://github.com/twbs/bootstrap/issues/31057#issuecomment-645196734. I will rename the issue later and list any other patches that we'll need to revert

XhmikosR (approved):

XhmikosR (approved):

PR #31085, opened 2020-06-17
input-file: create a plugin

### LINKED ISSUE #30265: `bs-custom-file-input` doesn't work anymore since the form updates
https://twbs-bootstrap.netlify.com/docs/5.0/forms/file/ broke since the class names aren't identical anymore.

Also, imo the plugin should be part of Bootstrap by default, since the file input's don't work without it. Also this should work out of the box without initializing it, this will be way more flexible if content is added to the DOM async.

Johann-S:
Demo: https://deploy-preview-31085--twbs-bootstrap.netlify.app/docs/5.0/forms/file/

Edit:
- some unit tests are disabled for Edge because it opens a real file chooser and made them failed

Closes #30265


site/content/docs/5.0/forms/file.md
```diff
@@ -6,16 +6,12 @@ group: forms
 toc: true
 ---
 
-{{< callout info >}}
-The recommended plugin to animate custom file inputs is [bs-custom-file-input](https://www.npmjs.com/package/bs-custom-file-input); it's what we use here in our docs.
-{{< /callout >}}
-
 ## Default
 
 The file input is the most gnarly of the bunch and requires additional JavaScript if you'd like to hook them up with functional *Choose file...* and selected file name text.
```
XhmikosR, line 11: This might need to be reworded since it won't require extra JS, but our plugin :)

Johann-S: if you have any ideas do not hesitate 😅

XhmikosR: Also, not sure if this should be listed as a component or not (probably not)

@mdo please have a look and let us know :)

js/src/file-input.js
```diff
@@ -0,0 +1,166 @@
...
+const SELECTOR_DATA_TOGGLE = '[data-toggle="file-input"]'
+const SELECTOR_FILE_INPUT = '.form-file-input'
+const SELECTOR_FILE_INPUT_LABEL = '.form-file-text'
+const SELECTOR_FORM = 'form'
+
+const EVENT_CLICK_DATA_API = `click${EVENT_KEY}${DATA_API_KEY}`
+const EVENT_CHANGE_DATA_API = `change${EVENT_KEY}${DATA_API_KEY}`
+const EVENT_RESET_DATA_API = `reset${EVENT_KEY}${DATA_API_KEY}`
+
+const fileApi = Boolean(window.File)
```
XhmikosR, line 34: Shouldn't this be in capitals since it's a constant? Also, is the variable name right? It seems like we are feature detecting.  Just wondering :)

Johann-S: yep it's a feature detection, I have to check that on Can I Use because nowadays File API should be supported everywhere

Johann-S: image
finally FileList API isn't supported by Opera Mini so we should keep the feature detection

mdo:
Any idea when this would be ready for review?

XhmikosR:
@mdo this is ready but please check out the comments above.

site/content/docs/5.0/forms/file.md
mdo, line 11:
```suggestion
Unlike our selects, checkboxes, and radios, the file input is much more difficult to customize. This means we've had to add additional HTML around the `<input type="file">` to create our the custom design. As such, it also requires additional JavaScript to make the custom form work with the input.

As of Bootstrap 5, this JavaScript is included in a new official plugin. Folks using Bootstrap 4 will need to make use of an [external plugin](https://www.npmjs.com/package/bs-custom-file-input).
```

site/content/docs/5.0/forms/file.md
```diff
@@ -50,24 +46,114 @@ Longer placeholder text is truncated and an ellipsis is added when there's not e
...
+<div class="form-file" data-toggle="file-input">
+  <input type="file" class="form-file-input" id="customFileData">
+  <label class="form-file-label" for="customFileData">
+    <span class="form-file-text">Choose file...</span>
+    <span class="form-file-button">Browse</span>
+  </label>
+</div>
+{{< /highlight >}}
+
+### Via JavaScript:
```
mdo, line 122:
```suggestion
### Via JavaScript
```

site/content/docs/5.0/forms/file.md
```diff
@@ -50,24 +48,114 @@ Longer placeholder text is truncated and an ellipsis is added when there's not e
...
+    <input type="file" class="form-file-input" id="customFileForm">
+    <label class="form-file-label" for="customFileForm">
+      <span class="form-file-text">Choose file...</span>
+      <span class="form-file-button">Browse</span>
+    </label>
+  </div>
+</form>
+<button id="btnResetFormFileInput" class="btn btn-primary mt-3">
+  Reset form
+</button>
```
mdo, line 81: Could this be moved inside the form with `type="reset"`?

Johann-S: yep the place doesn't really matter

site/content/docs/5.0/forms/file.md
```diff
@@ -50,24 +46,97 @@ Longer placeholder text is truncated and an ellipsis is added when there's not e
...
 </div>
 {{< /example >}}
+
+## JavaScript behavior
+
+### Usage
+
+#### Via data attributes
+
+Activate a file input without writing JavaScript. Set `data-toggle="file-input"` on a `.form-file` element.
```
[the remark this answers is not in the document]
Johann-S: maybe something like: `data-enable` ?

Johann-S:
Any reason @mdo it shouldn't be in Alpha 2 ? There are comments which are waiting for you here, and after those fixes it should be good to go

site/content/docs/5.0/forms/file.md
XhmikosR: This is still pending AFAICT @Johann-S :)

Johann-S: I wasn't sure if @mdo would like to tackle that or not

js/src/file-input.js
```diff
@@ -0,0 +1,163 @@
...
+  // Getters
+
+  static get VERSION() {
+    return VERSION
+  }
+
+  // Public
+
+  dispose() {
+    [window, this._element].forEach(htmlElement => EventHandler.off(htmlElement, EVENT_KEY))
```
XhmikosR, line 65: Might be better to remove the forEach call here since we only have 2 calls?

MartijnCuppens:
FWIW, next week (October 20th), Firefox 82 will ship. This version will support the [::file-selector-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button) pseudo element. Webkit has been supporting `::-webkit-file-upload-button` for a while.

This means we style file inputs without any JS, see https://codepen.io/MartijnCuppens/pen/VwjvXpe?editors=1100

XhmikosR:
@MartijnCuppens our target browsers aren't just Chromium latest, though.

MartijnCuppens:
@XhmikosR, `::-webkit-file-upload-button` has been around for a while, it's supported in Chrome 1 and Safari 3. Firefox tends to update pretty fast, so by the time BS5 releases, this will also work for (nearly) all FF users.

mdo:
Closing for #31955.

PR #31087, opened 2020-06-18
README.md: use unpkg.com in badgesize badges

XhmikosR:
Pros: reflects the master branch's sizes without us having to do a dist commit
Cons: it seems the sizes are off compared to bundlewatch. Either I'm missing something or the compression levels the 2 services use differ.

I'll experiment a bit more and if we go with this approach I'll backport it.


XhmikosR:
Closing since unpkg needs a tag, so it beats the point.

PR #31090, opened 2020-06-18
Docs: rename form checks page to "Checks / radios"

patrickhlauke:
it's not immediately obvious that "checks" includes information about both checkboxes and radio buttons. while "checks / radios" is also still not perfect (as honestly even I didn't quite grok that "checks" meant checkboxes), it should make it slightly more obvious nonetheless

Preview: https://deploy-preview-31090--twbs-bootstrap.netlify.app/docs/5.0/forms/checks-radios/


patrickhlauke:
(btw not sure if there's anything more that needs to be done for such a fundamental restructure...forgive my naive approach here)

MartijnCuppens:
I'm also in favor of renaming checks to checkboxes (and also `.form-check` to `form-checkbox`). Maybe we should also rename `.form-check[type="radio"]` to `.form-radio` (could be covered in another PR).

Btw, the link in the sidemenu is broken in this PR

patrickhlauke:
> Btw, the link in the sidemenu is broken in this PR

I wasn't sure how to resolve this in terms of having it shorter in the sidebar ("Checks / radios") but still have the title longer on the page itself ("Checks and radios"). I tried to work out how the "Browsers & devices"/"Browsers and devices" page does it, but to no avail.

I think the other changes (actually changing the classnames etc) may be a bit more fundamental...in this PR I'm really just after some clarity for the docs themselves, as it's unintuitive that "Checks" covers both checkboxes and radio buttons (though yes, they still use the same classname)

mdo:
I'd like to leave the class name as-is to keep v4 and v5 close. Every rename means more work for folks to upgrade. I'm okay to change the name of the page, though.

patrickhlauke:
any clue on how to end up with having the link in the sidebar say "Checks / radios", the URL be checks-radios, the title in full be "Checks and radios"? (similar to what "Browsers & devices" does)

am i missing some key location/file that helps with this mapping/connection?

MartijnCuppens:
@mdo
> I'd like to leave the class name as-is to keep v4 and v5 close.

In `v4`, those controls were named `.custom-checkbox` and `.custom-radio`. The change to `.form-checkbox` and `.form-radio` would be easier I guess.

@patrickhlauke 
> am i missing some key location/file that helps with this mapping/connection?

Hugo's [urlize](https://gohugo.io/functions/urlize/) function is used as "mapping"

https://github.com/twbs/bootstrap/blob/b84c9ed4894c7ff7824dca0c0e97c6e353eb8590/site/layouts/partials/docs-sidebar.html#L8

patrickhlauke:
while not a fan, i've fixed it by using `&` rather than `/` in sidebar.yml

patrickhlauke:
any clue why the two checks for Docs keep failing for this PR?

ffoodd:
@patrickhlauke There's a link somewhere pointing to `/form/checks`, resulting in a 404 error. That's why it rails, currently.

patrickhlauke:
@ffoodd aha, it helps when i look at the actual details for the "docs (pull_request)" (as the "docs (push)" one was just cryptic). found it...

site/content/docs/5.0/forms/checks-radios.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: docs
-title: Checks
+title: Checks and radios
```
XhmikosR, line 3: Needs a redirect

patrickhlauke: is redirect something i can do / add to the PR itself? or something that needs to be done separately?

XhmikosR: You should do it here because that's where you change the URL. Search the content files for `aliases`.

patrickhlauke: I meant "here" as in in the repo (compared to defining the redirects somewhere else in some server configuration file outside of the repo or on some web-based admin interface). But looking at `aliases` now, I think I get it. made an attempt, seems to work ok locally.

XhmikosR (approved):

PR #31096, opened 2020-06-18
modal: fix scrollable modal-body

### LINKED ISSUE #31084: modal dialog scrollable isn't working in v5 alpha, need some css improvement
in v5 alpha `.modal-dialog-scrollable` not adding scrolling to modal body when content goes too long.

[Modal Documentation](https://v5.getbootstrap.com/docs/5.0/components/modal/#scrolling-long-content)

When i check for the issue. I found some property missing

missing-css

[CodePen](https://codepen.io/S-Jaydip/pen/GRoNxXp)

Hiws:
Implements the suggested fix to make `.modal-body` scrollable when the content would otherwise overflow the viewport.

Closes: #31084

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -56,9 +56,10 @@
 }
 
 .modal-dialog-scrollable {
-  max-height: subtract(100%, $modal-dialog-margin * 2);
+  display: flex;
 
   .modal-content {
+    max-height: subtract(100vh, $modal-dialog-margin * 2);
     overflow: hidden;
   }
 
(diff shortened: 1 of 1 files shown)


ysds:
I'll see you later.

BTW, it is intentional to use 100% instead of 100vh. `100vh` doesn't avoid the address/toolbar of mobile browsers.

Hiws:
> BTW, it is intentional to use `100%` instead of `100vh`. `100vh` doesn't avoid the address/toolbar of mobile browsers.

Oh, i had no idea.
Thanks for the tip :)

PR #31098, opened 2020-06-18
Docs: change sidebar collapse controls to actual button elements

patrickhlauke:
apart from the visual styling, there's probably no good reason why these shouldn't be `<button>` elements, semantically

the buttons still look a shade too button-like, despite using `.btn-light`. the last remnant of button styling can probably be suppressed further, *or* this makes the case for expanding the core button styles to have one that looks completely non-button and non-link like (for cases where an additional style/hint was given already, or it's clear from context that something's an actionable button or link)

Preview: https://deploy-preview-31098--twbs-bootstrap.netlify.app/docs/5.0/getting-started/introduction/

### DIFF
--- a/site/assets/scss/_sidebar.scss
+++ b/site/assets/scss/_sidebar.scss
@@ -32,6 +32,7 @@
   a {
     padding: .1875rem .5rem;
     margin-top: .125rem;
+    margin-left: 1.25rem;
     color: rgba($black, .65);
     text-decoration: if($link-decoration == none, null, none);
 
@@ -41,28 +42,40 @@
       text-decoration: if($link-hover-decoration == underline, none, null);
       background-color: rgba($bd-purple-bright, .1);
     }
+  }
 
-    // Indent if there's no submenu
-    &:only-child {
-      margin-left: 1.25rem;
+  .btn {
+    // Custom styles (as we don't have a completely neutral button style)
+    padding: .25rem .5rem;
+    font-weight: 600;
+    color: rgba($black, .65);
+    background-color: none;
+    border: 0;
+
+    &:hover,
+    &:focus {
+      color: rgba($black, .85);
+      background-color: rgba($bd-purple-bright, .1);
+    }
+
+    &:focus {
+      box-shadow: 0 0 0 1px rgba($bd-purple-bright, .7);
     }
 
     // Add chevron if there's a submenu
-    &:not(:only-child) {
+    &::before {
+      width: 1.25em;
+      line-height: 0; // Align in the middle
+      content: escape-svg($sidebar-collapse-icon);
+      @include transition(transform .35s ease);
+      transform-origin: .5em 50%;
+    }
+
+    &[aria-expanded="true"] {
+      color: rgba($black, .85);
+
       &::before {
-        width: 1.25em;
-        line-height: 0; // Align in the middle
-        content: escape-svg($sidebar-collapse-icon);
-        @include transition(transform .35s ease);
-        transform-origin: .5em 50%;
-      }
-
-      &:not(.collapsed) {
-        color: rgba($black, .85);
-
-        &::before {
-          transform: rotate(90deg);
-        }
+        transform: rotate(90deg);
       }
     }
   }
--- a/site/layouts/partials/docs-sidebar.html
+++ b/site/layouts/partials/docs-sidebar.html
@@ -16,9 +16,9 @@
     {{- $is_active_group := eq $.Page.Params.group $group_slug }}
 
     <li class="mb-1{{ if $is_active_group }} active{{ end }}">
-      <a class="d-inline-flex align-items-center rounded{{ if not $is_active_group }} collapsed{{ end }}" data-bs-toggle="collapse" href="#{{ $group_slug }}-collapse" role="button" aria-expanded="{{ $is_active_group }}"{{ if $is_active_group }} aria-current="true"{{ end }}>
+      <button class="btn d-inline-flex align-items-center rounded{{ if not $is_active_group }} collapsed{{ end }}" data-bs-toggle="collapse" data-bs-target="#{{ $group_slug }}-collapse" aria-expanded="{{ $is_active_group }}"{{ if $is_active_group }} aria-current="true"{{ end }}>
         {{ $group.title }}
-      </a>
+      </button>
 
       {{- if $group.pages }}
       <div class="collapse{{ if $is_active_group }} show{{ end }}" id="{{ $group_slug }}-collapse">


patrickhlauke:
treat this as a proof of concept for now, probably...something for a bit more discussion/tweaking. but conceptually, it would be good if we led with a good example and avoided unnecessary uses of `<a href="..." role="button">` pseudo-buttons where real ones will work just as well.

site/assets/scss/_sidebar.scss
```diff
@@ -26,28 +27,28 @@
       text-decoration: if($link-hover-decoration == underline, none, null);
       background-color: rgba($bd-purple-bright, .1);
     }
+  }
 
-    // Indent if there's no submenu
-    &:only-child {
-      margin-left: 1.25rem;
-    }
+  button {
```
XhmikosR, line 32: I'd target the element with a class, personally.

XhmikosR:
As long as it's displayed properly I don't have any objection. `btn-light` doesn't look like it was before.

site/assets/scss/_sidebar.scss
patrickhlauke: done

patrickhlauke:
> As long as it's displayed properly I don't have any objection. `btn-light` doesn't look like it was before.

as said, we don't have a button class that looks just like the custom look that the docs currently use for those links. so we either hack extra styles in just for that (like we do for the links), or take the opportunity to add a new standard button style (that makes buttons look just like normal text, nothing else).

going for option 1 for now

site/assets/scss/_sidebar.scss
```diff
@@ -26,28 +27,36 @@
       text-decoration: if($link-hover-decoration == underline, none, null);
       background-color: rgba($bd-purple-bright, .1);
     }
+  }
+
+  .btn {
+    font-weight: 600;
+    color: rgba($black, .65);
+    // Custom styles (as we don't have a completely neutral button style)
+    background: none;
```
XhmikosR, line 36: should be `background-color`

patrickhlauke:
> ... *or* this makes the case for expanding the core button styles to have one that looks completely non-button and non-link like (for cases where an additional style/hint was given already, or it's clear from context that something's an actionable button or link)

thinking about this aspect some more...isn't it weird that for our own use case here with the docs, we end up having to craft our own custom button style potentially? this feels like a fairly clear indication that we could probably benefit from having something that's almost completely unstyled, maybe even with a standardised arrow/chevron icon that can be changed, as part of bootstrap's core styles themselves, no? if we can't even eat our own dogfood on its own to build the docs the way we want them? @twbs/css-review ?

patrickhlauke:
re-opening up the conversation here: @mdo @twbs/css-review etc ... how would we feel about creating a button style that is, essentially, completely clean, which can then more easily be repurposed for things like our own navigation rather than having to hack custom styles?

MartijnCuppens:
> How would we feel about creating a button style that is, essentially, completely clean, which can then more easily be repurposed for things like our own navigation rather than having to hack custom styles?

I'm ok with that. Something like `btn-reset`, or do you have any other suggestions?

XhmikosR:
Agreed, it's something we could definitely use ourselves. I'm not sure if you suggest adding it in core, but it should help us in our docs for sure.

patrickhlauke:
sorry, slammed with work so hadn't had time to give this more attention ...

>  not sure if you suggest adding it in core, but it should help us in our docs for sure

I think it makes sense to add it to core. our docs site should be dog-fooding bootstrap, and if *we* find the need for a particular component/style, then it's a good indicator that other devs will as well, so makes sense to add it to core

ffoodd:
Agreed, but how would we prevent this `.btn-reset` from being used inapropriately—and should we?

BTW, isn't what `.btn-link` is for, or so?

patrickhlauke:
>  how would we prevent this .btn-reset from being used inapropriately

we can't. we can only suggest authors be very careful and consider when this is/isn't appropriate to do / confusing to users. like anything else. if they're determined to abuse it, they will either way.

> BTW, isn't what .btn-link is for, or so?

that has an underline, so no - people want something that acts as a button but is otherwise "clean" (like we do for the actual side navbar, since we went out of our way to make custom styles for it)

XhmikosR:
Maybe we should just add the component in our docs only for starters?

mdo:
Agreed on docs only component for now. Let's set the styles up that we need, then see how to extrapolate later into Bootstrap's core if needed.

@patrickhlauke Want me to have a go at that, or you okay rebasing and handling?

patrickhlauke:
@mdo just rebased/solved the conflict... ready for review if you think it's a goer

patrickhlauke:
ah, scratch that, i see i had left it with `.btn-light` but then kicked around the idea of a completely reset button, which i've not done yet...

XhmikosR:
I still believe we need a basic button component for our doc. There are more places we could use it IIRC.

patrickhlauke:
before/after comparison here - almost identical (very subtle alignment difference with the chevrons, but i didn't think it warranted further messing)

> before-after

only noticeable difference now is focus behavior - as buttons have been given an explicit `:focus` style in general, they don't behave the same way as links, where browsers use heuristics to determine whether to show a focus outline or not by default (as is currently the case with all the sidebar links). i override the `:focus` style here to give it a slightly visually more pleasing box shadow, but this will now also show on mouse click.

this is where `:focus-visible` would be handy to use, because we do want/need focus styles for keyboard users (it would look real bad for our main navigation not to have that and then claim that BS can be used to make accessible sites), but ideally would be good to not show them on mouse click/touchscreen tap.

remind me, where were we with the idea of using the focus-visible polyfill? i've used it successfully in different apps/sites so far and it's nice...it would solve quite a few long-standing niggles we have

patrickhlauke:
> I still believe we need a basic button component for our doc. There are more places we could use it IIRC.

agree, would be my ideal preference too ... otherwise it's odd bootstrap's actual documentation site has to resort to custom hacked styles for things, rather than being able to use bootstrap's own built-ins

site/assets/scss/_sidebar.scss
```diff
@@ -41,28 +42,41 @@
...
     }
+  }
+
+  .btn {
+    // Custom styles (as we don't have a completely neutral button style)
+    padding: .25rem .5rem;
+    font-weight: 600;
+    color: rgba($black, .65);
+    background-color: none;
+    border: none;
```
mdo, line 53: Smidge shorter:

```suggestion
    border: 0;
```

site/assets/scss/_sidebar.scss
```diff
@@ -41,28 +42,41 @@
...
+    &:focus {
+      color: rgba($black, .85);
+      background-color: rgba($bd-purple-bright, .1);
+    }
 
-    // Indent if there's no submenu
-    &:only-child {
-      margin-left: 1.25rem;
+    &:focus {
+      box-shadow: 0 0 0 1px rgba($black, .45);
```
mdo, line 62: Can this be themed like the rest? The black looks off to me.

```suggestion
      box-shadow: 0 0 0 1px rgba($bd-purple-bright, .25);
```

mdo (commented):
Looking great overall!

site/assets/scss/_sidebar.scss
patrickhlauke: if this leads to a sufficient contrast of at least 3:1 for the outline, sure. let me check (as it's all variables within variables, it's difficult to ascertain without actually running it)

patrickhlauke: ok, `.25` is insufficient to reach 3:1 ratio. `.7` (technically, `.69`) was the cleanest one that got us just over the 3:1 mark.

patrickhlauke: (and yes, as mentioned previously elsewhere, our button focus styles for `.btn-primary` etc are well below the 3:1 ratio as well, something that we'll need to look into. and again, even more reason for something related to `:focus-visible` to allow us strong styles for keyboard users and light-touch ones for mouse/touch users)

patrickhlauke:
with new purple-themed focus box shadow (at `.7` opacity to get sufficient contrast ratio)

> themed-focus-style

patrickhlauke:
Anything more left to do on this?

XhmikosR:
We need a review from @twbs/css-review and ideally move the new CSS to a new docs component, instead of nesting this inside sidebar, so that we reuse it if needed :)

mdo (approved):

PR #31108, opened 2020-06-19
Update .rounded utilities

### LINKED ISSUE #29989: Confusing classes `rounded-sm` and `rounded-lg`

mdo:
- Adds .rounded-1, .rounded-2, and .rounded-3
- Drops .rounded-sm and .rounded-lg because they conflict with breakpoint nomenclature
- Rearrange rounded docs

Closes #30048, fixes #29989, closes #30778.

---

My hope with this PR is to balance keeping the primary `.rounded` class while reducing confusion around class names and adding additional control. To that end, I'm thinking of keeping `.rounded` even though it's the same as `.rounded-1` as of right now. This then drops `.rounded-sm` and `.rounded-lg` for something else.

That last change feels maybe problematic though since our large and small inputs and buttons use those variables.

Which of the referenced PRs are the best approaches? I feel like I could go any direction...

/cc @twbs/css-review

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -339,6 +339,7 @@ $border-radius:               .25rem !default;
 $border-radius-sm:            .2rem !default;
 $border-radius-lg:            .3rem !default;
 
+$rounded-sizes:               map-remove($spacers, 4, 5) !default;
 $rounded-pill:                50rem !default;
 
 $box-shadow:                  0 .5rem 1rem rgba($black, .15) !default;
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -461,13 +461,13 @@ $utilities: map-merge(
     "rounded": (
       property: border-radius,
       class: rounded,
-      values: (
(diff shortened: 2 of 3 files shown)


scss/_variables.scss
```diff
@@ -339,6 +339,7 @@ $border-radius:               .25rem !default;
 $border-radius-sm:            .2rem !default;
 $border-radius-lg:            .3rem !default;
 
+$rounded-sizes:               map-remove($spacers, 4, 5) !default;
```
MartijnCuppens, line 342: I think we better redelare the sizes here, would be weird if the spacers map is changed to :
```scss
$spacers: (
  0: 0,
  1: $spacer / 4,
  2: $spacer / 2,
  3: $spacer,
  4: $spacer * 1.5,
  5: $spacer * 3,
  6: $spacer * 4.5,
) !default;
```

mdo:
Replaced by #31687

PR #31109, opened 2020-06-19
toasts: make default duration 5s

### LINKED ISSUE #28185: Change the default toast duration
The default toast duration is 500 milliseconds.

I would argue that this duration is too short which leads to the toast often disappearing before user can fully read the toast message.

I argue in favor of raising the default toast duration.

Android also have the concept of toasts and provides two constants. `SHORT_DELAY` at 2000 ms and `LONG_DELAY` at 3500 ms.

rohit2sharma95:
Closes #28185


XhmikosR:
Unfortunately, this can't land in v4, but only in v5 because it's a BC. Other than that, we should definitely change the default duration. 2s might be a lot, we should discuss it.

/CC @Johann-S @mdo

XhmikosR:
After thinking about it, I think 1s might be a better default.

Johann-S:
I chose to set the default `duration` at `500ms` because I wanted our users to think about what's a good duration for them and change it accordingly

rohit2sharma95:
> I wanted our users to think about what's a good duration for them and change it accordingly

@Johann-S Yes, users can change the default duration. But the default duration should be enough for the general toast so that the users don't need to change the default whenever they use the toast.

mdo:
Agreed on this being longer. Will play with the timing and would appreciate others doing the same before we decide and merge.

Johann-S (approved):

XhmikosR:
So, do we all agree to 2s? Should it be something else?

patrickhlauke:
i'd tend more towards 5s to be honest (would give sufficient time by default for assistive tech to announce a short toast notification without it disappearing mid-reading)

ysds:
Personally I prefer 4-5 seconds.

Also, although a little old, this discussion may be helpful https://github.com/jackbsteinberg/std-toast/issues/37

mdo:
Let's go with 5s. @rohit2sharma95 Can you make the update, please?

rohit2sharma95:
Updated @mdo

XhmikosR:
We should add this in Migration too.

bardiharborow (approved):

vanillajonathan:
Android have constants for a long  and a short duration, they are 4 seconds for the short toast and 7 seconds for the long toast.

These _might_ be well thought out (or not?) durations backed by UX research.

https://github.com/aosp-mirror/platform_frameworks_base/blob/android10-release/core/java/android/widget/Toast.java#L384
> ```java
> static final long SHORT_DURATION_TIMEOUT = 4000;
> static final long LONG_DURATION_TIMEOUT = 7000;
> ```

PR #31111, opened 2020-06-19
Update docs example on how to optimize JavaScript bundle

pascalpepe:
The lean JavaScript example in v5 docs has not been updated from v4.

This PR suggests the two following changes in the Lean JavaScript code snippet:

* Remove the import of `js/dist/util.js`, which does not exist anymore.
* Change the comment about Popper.js to also mention dropdowns instead of just tooltips and popovers.

https://deploy-preview-31111--twbs-bootstrap.netlify.app/docs/5.0/customize/optimize/

### DIFF
--- a/site/content/docs/5.0/customize/optimize.md
+++ b/site/content/docs/5.0/customize/optimize.md
@@ -24,20 +24,31 @@ For instance, assuming you're using your own JavaScript bundler like Webpack or 
 ```js
 // Import just what we need
 
-// import "bootstrap/js/dist/alert";
-// import "bootstrap/js/dist/button";
-// import "bootstrap/js/dist/carousel";
-// import "bootstrap/js/dist/collapse";
-// import "bootstrap/js/dist/dropdown";
-import "bootstrap/js/dist/modal";
-// import "bootstrap/js/dist/popover";
-// import "bootstrap/js/dist/scrollspy";
-// import "bootstrap/js/dist/tab";
-// import "bootstrap/js/dist/toast";
-// import "bootstrap/js/dist/tooltip";
+// import 'bootstrap/js/dist/alert';
+// import 'bootstrap/js/dist/button';
+// import 'bootstrap/js/dist/carousel';
+// import 'bootstrap/js/dist/collapse';
+// import 'bootstrap/js/dist/dropdown';
+import 'bootstrap/js/dist/modal';
+// import 'bootstrap/js/dist/popover';
+// import 'bootstrap/js/dist/scrollspy';
+// import 'bootstrap/js/dist/tab';
+// import 'bootstrap/js/dist/toast';
+// import 'bootstrap/js/dist/tooltip';
 ```
 
 This way, you're not including any JavaScript you don't intend to use for components like buttons, carousels, and tooltips. If you're importing dropdowns, tooltips or popovers, be sure to list the Popper.js dependency in your `package.json` file.
+
+{{< callout info >}}
+### Exports
+
+Files in `bootstrap/js/dist` use the **default export**, so if you want to use one of them you have to do the following:
+
+{{< highlight js >}}
+import Modal from 'bootstrap/js/dist/modal'
+const modal = new Modal(document.getElementById('myModal'))
+{{< /highlight >}}
+{{< /callout >}}
 
 ## Autoprefixer .browserslistrc
 
--- a/site/content/docs/5.0/customize/optimize.md
+++ b/site/content/docs/5.0/customize/optimize.md
@@ -40,14 +40,15 @@ import 'bootstrap/js/dist/modal';
 This way, you're not including any JavaScript you don't intend to use for components like buttons, carousels, and tooltips. If you're importing dropdowns, tooltips or popovers, be sure to list the Popper.js dependency in your `package.json` file.
 
 {{< callout info >}}
-### Exports
+### Default Exports
 
 Files in `bootstrap/js/dist` use the **default export**, so if you want to use one of them you have to do the following:
 
-{{< highlight js >}}
+```js
 import Modal from 'bootstrap/js/dist/modal'
+
 const modal = new Modal(document.getElementById('myModal'))
-{{< /highlight >}}
+```
 {{< /callout >}}
 
 ## Autoprefixer .browserslistrc
--- a/site/content/docs/5.0/customize/optimize.md
+++ b/site/content/docs/5.0/customize/optimize.md
@@ -24,14 +24,20 @@ For instance, assuming you're using your own JavaScript bundler like Webpack or 
 ```js
 // Import just what we need
 
-// If you're importing tooltips or popovers, be sure to include the Popper.js dependency
-// import "../../node_modules/popper.js/dist/popper.min.js";
-
-import "../../node_modules/bootstrap/js/dist/util.js";
-import "../../node_modules/bootstrap/js/dist/modal.js";
+// import "bootstrap/js/dist/alert";
+// import "bootstrap/js/dist/button";
+// import "bootstrap/js/dist/carousel";
+// import "bootstrap/js/dist/collapse";
+// import "bootstrap/js/dist/dropdown";
+import "bootstrap/js/dist/modal";
+// import "bootstrap/js/dist/popover";
+// import "bootstrap/js/dist/scrollspy";
+// import "bootstrap/js/dist/tab";
+// import "bootstrap/js/dist/toast";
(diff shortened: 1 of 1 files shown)


site/content/docs/5.0/customize/optimize.md
```diff
@@ -24,10 +24,9 @@ For instance, assuming you're using your own JavaScript bundler like Webpack or
 ```js
 // Import just what we need
 
-// If you're importing tooltips or popovers, be sure to include the Popper.js dependency
+// If you're importing dropdowns, tooltips or popovers, be sure to include the Popper.js dependency
 // import "../../node_modules/popper.js/dist/popper.min.js";
 
-import "../../node_modules/bootstrap/js/dist/util.js";
 import "../../node_modules/bootstrap/js/dist/modal.js";
```
Johann-S, line 30: it seems are assuming folks use a bundles so the path should be:
```js
import 'bootstrap/js/dist/modal'

Comment (pascalpepe) [reply site/content/docs/5.0/customize/optimize.md:30]: You're right, at least that's how it should be for Webpack and Parcel. But the documentation for this example also mentions Rollup and I don't know about it, so I assumed the code was written like this for Rollup.

I'll try to get a bundle with Rollup to see how it goes.

Comment (pascalpepe) [reply site/content/docs/5.0/customize/optimize.md:30]: You're right, at least that's how it should be for Webpack and Parcel. But the documentation for this example also mentions Rollup and I don't know about it, so I assumed the code was written like this for Rollup.

I'll try to get a bundle with Rollup to see how it goes.

**Comment by XhmikosR** [comment]
Yeah, I think this should be cleaned up. See #31111

**Comment by pascalpepe** [comment]
Alright, I just created [this repo](https://github.com/pascalpepe/bootstrap-js-bundling) containing minimal working examples with Parcel, Rollup and webpack. 

In all three examples, the bundle contains only the collapse and dropdown plugins (plus Popper.js, required for dropdowns). There's also a template with a navbar that uses both plugins.

Regardless of the bundler, the imports are written like this:

```js
import 'bootstrap/js/dist/collapse';
import 'bootstrap/js/dist/dropdown';
```

And actually there is no need to explicitly import Popper.js in the entry file because all bundlers will automatically include it if needed. It simply needs to be installed with npm in the same way as Bootstrap.

If this sounds ok for you, I can rebase and update this PR with cleaned imports and the correct comment about Popper.

mdo:
@pascalpepe Holler if you need any more feedback here before opening up for reviews :).

Johann-S:
LGTM @pascalpepe  👍

pascalpepe:
Thanks for reviews ! Here's the updated PR:

- The .js extension has been removed from all imports as it is not needed.
- To mirror the section about SCSS customization that's just above this one, I updated the JS example like this:
  * The example now lists all available JS components.
  * Only the modal component is not commented out (to mirror the comments above: "If you’re not using a component, comment it out or delete it entirely").
  * The comment about Popper.js was previously in the code snippet. I moved it at the end of section.

Ready for review :)

mdo (dismissed):

Johann-S:
Btw I think we should mention we use export default for those files, which mean if you want to create a new modal you have to do:
```js
import Modal from 'bootstrap/js/dist/modal'

const modal = new Modal(document.getElementById('myModal'))

**Comment by mdo** [comment]
Any interest in the last comment @pascalpepe?

**Comment by XhmikosR** [comment]
@Johann-S can you add the changes you mentioned above?

**Comment by Johann-S** [comment]
@XhmikosR I'm unable to add a JS example in a `callout`, do you know how to do that?

File: site/content/docs/5.0/customize/optimize.md [hunk 2]
```diff
@@ -24,21 +24,33 @@ For instance, assuming you're using your own JavaScript bundler like Webpack or
...
 This way, you're not including any JavaScript you don't intend to use for components like buttons, carousels, and tooltips. If you're importing dropdowns, tooltips or popovers, be sure to list the Popper.js dependency in your `package.json` file.
 
+{{< callout info >}}
+### Exports
+
+Files in `bootstrap/js/dist` use **default export**, so if you want to use one of them you have to do the following:
+
+{{< highlight js >}}
+import Modal from 'bootstrap/js/dist/modal'
+
```
XhmikosR, line 49: Removing this newline should fix the issue, but I'm not sure for the long-term fix so that we can use line breaks here.

XhmikosR: To clarify, the issue must be because we use `markdownify` in the callout shortcode:

https://github.com/twbs/bootstrap/blob/c5966de27395a407f9a3d20d0eb2ff8e8fb7b564/site/layouts/shortcodes/callout.html#L9

But removing `markdownify` means we won't be able to use Markdown in the callouts, which isn't something we want. After having a closer look, I think we are calling the callout shortcode wrong. We should use `{{% callout ... %}}` which parses the content as Markdown and remove our `markdownify` usage.

I'll need to have a closer look to see if this affects other places and I'll make a PR.

XhmikosR: All right, I can't make it work with `{{% foo %}}` either; the new line is causing an HTML error.

```html
div class="bd-callout bd-callout-warning">
<h5 id="ensure-correct-role-and-provide-a-label">Ensure correct <code>role</code> and provide a label</h5>
<p>
  In order for assistive technologies (such as screen readers) to convey that a series of buttons is grouped, an appropriate <code>role</code> attribute needs to be provided. For button groups, this would be
  <code>role=&quot;group&quot;</code>, while toolbars should have a <code>role=&quot;toolbar&quot;</code>.
</p>
<p>
  In addition, groups and toolbars should be given an explicit label, as most assistive technologies will otherwise not announce them, despite the presence of the correct role attribute. In the examples provided here, we use
  <code>aria-label</code>, but alternatives such as <code>aria-labelledby</code> can also be used.
</p>
<div class="highlight">
  <pre class="chroma"><code class="language-js" data-lang="js"><span class="kr">import</span> <span class="nx">Modal</span> <span class="nx">from</span> <span class="s1">&#39;bootstrap/js/dist/modal&#39;</span>
<!-- this is wrong and caused by the newline -->
<p>
<span class="kr">const</span> <span class="nx">modal</span> <span class="o">=</span> <span class="k">new</span> <span class="nx">Modal</span><span class="p">(</span><span class="nb">document</span><span class="p">.</span><span class="nx">getElementById</span><span class="p">(</span><span class="s1">'myModal'</span><span class="p">))</span>
</code></pre>
<!-- these 2 are wrong -->
</div></p>
</div>

```

I still think our callout usage is wrong, so I'll try to look into that, at least.

Johann-S: I tried the line break but it doesn't work 😟

XhmikosR: I still see the line break present. This should work:

```
{{< highlight js >}}
import Modal from 'bootstrap/js/dist/modal'
const modal = new Modal(document.getElementById('myModal'))
{{< /highlight >}}
```

Johann-S: ok now it works, thanks! 👍

XhmikosR: Actually please try this without the `\`:

```
`\`\`js
import Modal from 'bootstrap/js/dist/modal'

const modal = new Modal(document.getElementById('myModal'))
\`\`\`
```

It works for me, and it's a better solution. I will add it in my TODO to clean this situation up later :)

XhmikosR: OK, NVM the triple backticks works but I need to remove `markdownify`. So let's leave this snippet as is until I have a look later.

XhmikosR:
For some reason, the Netlify preview is broken; it throws a `ReferenceError`:

```
 Uncaught ReferenceError: ClipboardJS is not defined
    <anonymous> https://deploy-preview-31111--twbs-bootstrap.netlify.app/docs/5.0/assets/js/docs.min.js:38
    <anonymous> https://deploy-preview-31111--twbs-bootstrap.netlify.app/docs/5.0/assets/js/docs.min.js:52
```

XhmikosR:
Let's wait until we proceed with #31806 and #31802 because docs JS is broken with the current patch as is.

XhmikosR:
So, it seems the error comes from ClipboardJS due to the "Exports" heading. Should probably be reported upstream, but in the meantime I renamed the heading to "Default Exports".

If someone could report it, that would help.

XhmikosR (approved):

Issue #31099, opened 2020-06-18
Dynamic Tabs, where href starts with number, not working


PackeTsar:
Example: https://jsfiddle.net/wcq4L2fr/1/

When the nav-link href starts with a number, or is only a number (`<a class="nav-link" data-toggle="tab" href="#1">`), it will not expose the referenced contents (`<div class="tab-pane fade" id="1">`).

If the nav-link href is simply alpha, or begins with an alpha character, it works fine.

PackeTsar:
This may be a duplicate of #1956, but it doesn't seem the issue was resolved there

patrickhlauke:
this falls under the same umbrella as this https://github.com/twbs/bootstrap/issues/31005

short version: HTML5 now allows ids to start with a number. but then when you use them in your CSS, or in JS when using things like `querySelector` (which under the hood BS uses), you'd need to do extra work to escape these id references first. BS currently doesn't do that (and hints at this with https://getbootstrap.com/docs/4.5/getting-started/javascript/#selectors), but admittedly this could be made clearer / BS could try and auto-escape references in its functions.

XhmikosR:
I'm not sure what we can do in the JS side to allow this? Personally, I always follow the specs in all my projects so I never had the need for such thing.

XhmikosR:
You are welcome to look into it. As long as we follow the specs, and we can fix this on our side without lots of code, I don't see why not adding support for this.

Johann-S:
Being able to use the native: `CSS.escape` would be the best option, but not really supported currently

ffoodd:
@Johann-S Just checked [on MDN](https://developer.mozilla.org/fr/docs/Web/API/CSS/escape) and [CanIUse](https://caniuse.com/mdn-api_css_escape), support is pretty good. Only missing Legacy Edge, for v5…

Also [Mathias Bynens wrote a lightweight polyfill](https://github.com/mathiasbynens/cssesc) (only 3.43kb unminified): might it be used?

XhmikosR:
I guess this is one of the reasons we should consider regarding dropping Legacy Edge support.

mdo:
Let's drop it then :). We've been teetering on the decision for awhile now, I think Alpha 3 is the time to do it.

XhmikosR:
Now with Legacy Edge dropped, any PRs to switch to CSS.escape are welcome!

alpadev:
[...]

PR #31112, opened 2020-06-19
Escape selectors to support HTML5 ID spec 

Hiws:
This is a *rough* draft, of a possible simple solution to support HTML5 ID's in combination with the various JS plugins.

This allows the user to use non-alpha characters as the first character in the ID, which is currently not possible.

Basic example of what will be possible:
```html
<a class="nav-link" data-toggle="tab" href="#1-tab">Tab 1</a>

<div class="tab-pane fade" id="1-tab">
  Content
</div>
```

I've tested `Modals`, `Tabs`, `Scrollspy`, `Carousel` and `Collapse` locally via the docs, and it appears to function correctly, but will definitely need more testing to ensure it works across supported browsers and environments.
My knowledge of all the ways the plugins above can be used is limited, which is why I'd love some input whether the current draft makes sense.

The selector is escaped using the built-in `CSS.escape`, function. However, I don't *think* this is supported in Legacy Edge. So a different solution might be needed here.

Alternatives to native `CSS.escape`:

- https://github.com/mathiasbynens/cssesc - Standalone function (with more features?)
- https://github.com/mathiasbynens/CSS.escape - A direct polyfill
- Write a standalone function specific to Bootstrap

### DIFF
--- a/js/src/util/index.js
+++ b/js/src/util/index.js
@@ -41,6 +41,15 @@ const getSelector = element => {
     selector = hrefAttr && hrefAttr !== '#' ? hrefAttr.trim() : null
   }
 
+  if (selector) {
+    // Can we assume the first char is always a #(id) or .(class)?
+    const baseSeletor = selector.slice(0, 1)
+    // Do we need to polyfill CSS.escape for Legacy Edge?
+    const escapedSelector = CSS.escape(selector.slice(1).trim())
+
+    selector = baseSeletor + escapedSelector
+  }
+
   return selector
 }
 


Johann-S:
simple curiosity, do you know if it would work with selector containing several ID, something like: `#test, #test2` ?

Hiws:
@Johann-S 
Are you asking whether `CSS.escape('#test, #test2')` would work?
If so, then not to my knowledge. Unless you striped out `test` and `test2` (without the `#`) and escaped them separately and placing it back.

Johann-S:
I was asking about your changes.
It seems a lot of work for just a letter before an ID 🤔 BTW it was the case in v4 too

Hiws:
Well, it isn't really limited to a letter, could be anything (from my understanding of the HTML5 spec).
```html
<div id="#👌-id">
```

I forgot to link it to the issue (#31099) that caused me to look into it.
Wanted to give it a try, to see if it was possible to do without having to add a *ton* of extra work, which it does for the cases i (limited) testing of the mentioned plugins.

But yeah, if more work needs to be done to make it work properly, like with `#foo, #far`. 
Then it's probably not worth handling this edge case, since the workaround is rather simple (start `id` with an alpha character)

PR #31114, opened 2020-06-19
Docs: forms accessibility cleanup

patrickhlauke:
- expands on what's needed for "faked" controls inside a disabled `<fieldset>`
- centralises the accessibility advice about always providing an accessible name, and mentions some of the possible variants; includes strong suggestion not to rely on placeholder and invisible labels as beyond AT users it can be bad for other user groups
- updated warning about validation, which (due to #28414), is actually worse than the (now improved) browser default validation; expanded the example for server-side to explicitly tie error messages to form fields using `aria-describedby`

this should at least partially be backported (manually, since v4 doesn't have the overview page)


XhmikosR:
@patrickhlauke this needs a rebase :)

patrickhlauke:
rebased @XhmikosR

site/content/docs/5.0/forms/validation.md
```diff
@@ -183,43 +185,43 @@ We recommend using client-side validation, but in case you require server-side v
     <label for="validationServerUsername" class="form-label">Username</label>
     <div class="input-group">
       <span class="input-group-text" id="inputGroupPrepend3">@</span>
-      <input type="text" class="form-control is-invalid" id="validationServerUsername" aria-describedby="inputGroupPrepend3" required>
-      <div class="invalid-feedback">
+      <input type="text" class="form-control is-invalid" id="validationServerUsername" aria-describedby="inputGroupPrepend3 validationServerUsernameFeedback" required>
```
MartijnCuppens, line 188: Not sure `@` is a proper description, shouldn't we move the id of the prefix to the label?

patrickhlauke: is is a proper additional description in that it makes it obvious what sighted users would perceive. keep in mind that this description is in addition to the preceding `<label>` that also still acts on the `<input>`, so it's announced as something like "Username, @, input"

XhmikosR:
@patrickhlauke this needs another rebase. Do note that for v4 these will need to be manually backported so please rethink if they need to be backported and if so make a PR against the v4-dev branch or push to my v4-dev-xmr branch.

patrickhlauke:
ok so after being told to do a rebase, i tried...and have no idea why now all of a sudden i get all those other extraneous commits in here. sigh.

XhmikosR:
Yeah, that's not a rebase :P A rebase should have only your patches.

Anyway, I'll try to fix it tomorrow myself.

patrickhlauke:
i think for some reason my local env didn't have the latest main, and was refusing to grab it. had to drop it and refetch it. odd. anyway, i THINK i got it more correctly this time around, let's see...

patrickhlauke:
manual v4 partial backport here #31234

XhmikosR (approved):
LGTM, I'd like a review from @mdo and we squash and merge

mdo:
@patrickhlauke Gave this a read through and made some edits. Let me know if anything looks off, then we can merge. Thanks!

mdo (approved):

patrickhlauke:
yup, all looking good @mdo, thanks. merging.

PR #31115, opened 2020-06-19
fix tooltip hide method when already hidden

### LINKED ISSUE #31110: Modals that include tooltip/popover throws error on close
A modal with a tooltip/popover throws an error on close, if the tooltip/popover wasn't shown while the modal was open.
`Uncaught TypeError: Cannot read property 'destroy' of null`

Steps to reproduce:

1. Go to [modal w/ tooltip example](https://v5.getbootstrap.com/docs/5.0/components/modal/#tooltips-and-popovers)
2. Open dev console (i used chrome on windows)
3. Open the modal, and close it again.
4. Notice 3 errors in the console.
5. Open the modal again, hover the one of the links to show a tooltip.
6. Close the modal again.
7. Notice only 2 errors are displayed in the console now, since one tooltip was shown.

The error isn't limited to the docs.
[Codepen example](https://codepen.io/Hiws/pen/qBbRjvg)

Hiws:
Fixes the error displayed when closing a modal with a tooltip that was never shown.

Closes: #31110


XhmikosR (approved):

XhmikosR:
I wonder if we have more cases like this one?

js/src/tooltip.js
```diff
@@ -340,7 +340,9 @@ class Tooltip {
       this._cleanTipClass()
       this.element.removeAttribute('aria-describedby')
       EventHandler.trigger(this.element, this.constructor.Event.HIDDEN)
-      this._popper.destroy()
+      if (this._popper !== null) {
```
Johann-S, line 343: I would add that check at the beginning of the hide function, something like:
```js
if (!this._popper) {
  return
}

**Review by Johann-S — CHANGES_REQUESTED** [review]

File: js/src/tooltip.js (hunk 1 above)
Comment (Hiws) [reply js/src/tooltip.js:343]: I wasn't really sure if the code before the destroy was still relevant, but i can move it if you think that's better.

Comment (Johann-S) [reply js/src/tooltip.js:343]: A tooltip cannot be shown without Popper.js

Comment (XhmikosR) [reply js/src/tooltip.js:343]: True, good catch and suggestion 👍

**Comment by Johann-S** [comment]
I'm more concerned about listening to a modal hiding in our tooltip component, for me we shouldn't do that @XhmikosR

**Comment by XhmikosR** [comment]
> I'm more concerned about listening to a modal hiding in our tooltip component, for me we shouldn't do that @XhmikosR

Also, true. Does this code come from v4? You rewrote most of v5's JS code :)

**Comment by Johann-S** [comment]
Yep it comes from our v4

**Comment by Hiws** [comment]
> I'm more concerned about listening to a modal hiding in our tooltip component, for me we shouldn't do that @XhmikosR

It makes sense why it's there (if that's what you're questioning), since it's a bit awkward that if you open a popover/tooltip and then close the modal. 
The popover/tooltip will be open the next time you open the modal.

**Comment by Johann-S** [comment]
That's the reason yes, but is it really a good thing ? IMO it should be handled by our users to delete their tooltip if needed

**Comment by Hiws** [comment]
Perhaps. 
I guess the current implementation limits it so they can't keep it open. Even if they wanted to.
While if the code was removed, it would keep it open, while still allowing the users to specifically close them if they wish.

**Comment by Johann-S** [comment]
Anyway you fix is good @Hiws but can you add a unit test please ?

**Comment by Hiws** [comment]
> Anyway you fix is good @Hiws but can you add a unit test please ?
@Johann-S 

Tried writing this test for 2 hours, and it looks like it's working now.
However, I've never really written unit tests, so not sure if it's correct or not.

**Comment by XhmikosR** [comment]
I think the test should be in modal not in tooltip :)

**Comment by Hiws** [comment]
> I think the test should be in modal not in tooltip :)

Ii feel like this case isn't specifically tied to modals, but the tooltip's `hide` method being run without the `show` method ever being ran.
Which is why i placed it under the tooltip tests and not modal.

**Comment by XhmikosR** [comment]
The error your patch fixes is in modal.js so that's where it should be tested that we don't throw.

**Comment by XhmikosR** [comment]
NVM, you are right, I need to sleep.

**Comment by Hiws** [comment]
> NVM, you are right, I need to sleep.

I'll take part of the blame for having a bad title 😉

File: js/tests/unit/tooltip.spec.js [hunk 2]
```diff
@@ -717,6 +717,20 @@ describe('Tooltip', () => {
...
     })
+
+    it('should not throw error running hide if popper hasn\'t been shown', () => {
+      fixtureEl.innerHTML = '<div></div>'
+
+      const div = fixtureEl.querySelector('div')
+      const tooltip = new Tooltip(div)
+
+      try {
+        tooltip.hide()
```
Johann-S, line 728: you should call it twice and except nothing

Johann-S:
just my last comment and we are good to gere here @Hiws thanks for your work 👍

polRk:
Before merge, or alpha 2, use on child component `style="pointer-events: none;" `

```
<button class="btn btn-light" data-toggle="tooltip" data-placement="top" title="Tooltip on top">
    <svg width="18" height="18" style="pointer-events: none;" ><use xlink:href="#heart" /></svg>
</button>
```

mdo:
Sorry for being out of the loop, what's the status of this one?

mdo:
Pulling from Alpha 2 given unclear status.

Johann-S:
We are waiting for @Hiws to do a last change and rebase his branch

Hiws:
Sorry for the delay.

I'm not quite sure what you mean with `you should call it twice and except nothing`.

Johann-S:
You're right it was good like that thanks 👌

Johann-S (approved):

PR #31117, opened 2020-06-19
Release Drafter tweaks

mdo:
- [x] Don't jump to next major version
- [ ] Look into the v4-dev and main issues (why does publishing a v4.x release reset for `main` branch) - Upstream issue, a fix doesn't seem to exist yet
- [ ] Anything else?
- [ ] Exclude dependencies? It adds a lot of noise
- [ ] Sorting doesn't seem to be by PR number?
- [ ] Add Examples section


mdo:
Closing as stale. Would love if anyone has ideas on generating release changelogs to pick this up and see how we can do two branches for two major versions without interruption.

PR #31119, opened 2020-06-19
v5: Adjust docs sidebar and navbar

mdo:
Some tweaks to ideally improve navigating across our documentation on narrow devices. This PR removes the navbar scroll that I had implemented and replaces it with our classic collapsing navbar, but with a slight twist. I've added column classes to stack the nav across two columns on mobile devices so that the opened nav isn't too tall.

Still need to adjust the menu toggles (missing some styles, tap targets could be larger, etc), but hoping to get some initial feedback from the team on overall direction. Also wondering if I should do more to condense the subnav (search, version picker, and sidebar toggle) on mobile views.

Any thoughts, feedback, or examples of other patterns would be appreciated!

/cc @twbs/css-review 

Preview: https://deploy-preview-31119--twbs-bootstrap.netlify.app/

### DIFF
--- a/site/assets/scss/_masthead.scss
+++ b/site/assets/scss/_masthead.scss
@@ -2,10 +2,6 @@
   padding: 3rem 0;
   background: linear-gradient(165deg, lighten($bd-purple-light, 16%) 50%, $white 50%);
 
-  @include media-breakpoint-up(sm) {
-    padding: 5rem 0;
-  }
-
   h1 {
     @include font-size(4rem);
     line-height: 1;
--- a/site/assets/scss/_navbar.scss
+++ b/site/assets/scss/_navbar.scss
@@ -1,22 +1,21 @@
 .bd-navbar {
-  padding: .625rem 0;
+  padding: .75rem 0;
   background-color: $bd-purple-bright;
 
-  @include media-breakpoint-down(lg) {
-    .navbar-nav-scroll {
-      width: 100%;
+  .navbar-toggler {
+    padding: 0;
+    border: 0;
 
-      .navbar-nav {
-        margin: -.5rem 0;
-        overflow-x: auto;
-        white-space: nowrap;
-        -webkit-overflow-scrolling: touch;
-      }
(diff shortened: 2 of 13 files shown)


patrickhlauke:
like this, particularly since small screen doesn't necessarily always mean "mobile user with touch" (it could, for instance, still be a mouse user on desktop, but with high magnification/zoom). and the horizontal scrolling without scrollbar is unusable for these users, currently.

alecpl:
I'd like search/version/toggle in a single line. Maybe the version selector could be visually integrated with the search input (and does not need the "Bootstrap" label, only version number). Maybe the icon should be different than the main menu icon (folder tree, hierarchy?).

btw, on the home page (not docs, the main page) I propose to make `.masthead-followup-icon` in the same line with the following `h2`. And smaller bottom padding/margin on `#content`.

mdo:
Rebased to bring this up to speed with the latest `main` changes, added @alecpl's suggestion for subnav layout all on one line with a cheeky `d-none` trick, and tweaked the homepage masthead `padding` (more to come on this later I think).

Ready for review!

ffoodd:
I like it!

I agree with @alecpl about the two icons, there's so much lines stacking here… Not sure about a solution though, I just noticed that felt strange.

mdo:
Yeah the stacked bars is a little odd. How about this? Just pushed a commit for it—icon flips between the expand/collapse icons from Bootstrap Icons.

Screen Shot 2020-09-16 at 11 20 20 AM

ffoodd:
@mdo Nice, IMHO much better!

mdo:
Rebased and conflicts resolved.

Ready for review!

MartijnCuppens (approved):
Neat improvement!

ffoodd (approved):

XhmikosR (approved):

mdo:
Thanks for the reviews fam! This will be a nice little addition to Alpha 2 :D.

PR #31122, opened 2020-06-20
Update migration.md

XhmikosR:
Draft because I want to update a couple more things like the supported browsers.

- [x] Update supported browsers section to reflect the current `main` branch
- [ ] Tackle the `Todo` in migration

Preview: https://deploy-preview-31122--twbs-bootstrap.netlify.app/docs/5.0/migration/


jipexu:
hi
What means NN - MM
Dropped support for Firefox NN - MM

Thanks

XhmikosR:
> 
> 
> hi
> What means NN - MM
> Dropped support for Firefox NN - MM
> 
> Thanks

NN means we forgot to change these, hence this PR :) 

I plan to add more info in migration page, but I'll probably need some help @mdo about the `Todo` in this file.

site/content/docs/5.0/migration.md
```diff
@@ -51,22 +68,22 @@ Changes to our source Sass files and compiled CSS.
...
 - Dropped jQuery dependency and rewrote plugins to be in regular JavaScript.
 - Removed underscore from public static methods like `_getInstance()` → `getInstance()`.
 
-## Color system
+### Color system
 
 We've updated the color system that powers Bootstrap to improve color contrast and provide a much more extensive set of colors.
 
-- Updated blue and pink base colors (`-500`) to ensure AA contrast.
+- Updated blue and pink base colors (`-500`) to ensure WCAG 2 AA contrast.
```
XhmikosR, line 82: @patrickhlauke should this be 2.1?

patrickhlauke: yes would make sense to change to 2.1

patrickhlauke: (technically it's not incorrect, as 2.1 only extends 2.0 so the contrast requirement is unchanged/inherited in 2.1, but still)

XhmikosR:
Can I get an approval please? We can leave the TODO for another time, as long as we do it before beta/stable

ffoodd (approved):

PR #31125, opened 2020-06-20
BrowserStack: stop testing on latest Edge.

XhmikosR:
It's Chromium-based so it seems redundant.

Split from #30986

If the BrowserStack issues are fixed, I plan to test both the minimum and latest versions of the browsers we support. Since Actions are split, the slowdown shouldn't be huge, but it remains to see how much it is.


0532302779:
https://github.com/users/0532302779/emails/108216487/confirm_verification/d5ee6314feb69cf3885cd9d8cb8792bbde8ff3ae?utm_campaign=github-email-verification&utm_content=html&utm_medium=email&utm_source=verification-email

0532302779:
And mention moon tacos

Johann-S (approved):

PR #31130, opened 2020-06-21
Add code to method/property tables

patrickhlauke:
initial step to making tables for methods/properties consistent (before we possibly switch to full markdown tables)

Preview: https://deploy-preview-31130--twbs-bootstrap.netlify.app/


XhmikosR:
It seems we are also inconsistent with tables in some components. For example, https://deploy-preview-31130--twbs-bootstrap.netlify.app/docs/5.0/components/tooltips/#methods is using a table but https://deploy-preview-31130--twbs-bootstrap.netlify.app/docs/5.0/components/carousel/#methods doesn't.

The no table approach has the advantage that one can link directly to a method. Either way, we should streamline this.

patrickhlauke:
can we merge this so far and circle back to the consistency later? or do it all here? if the latter, we'd need some consensus on what to do...

XhmikosR:
Made #31189 to track the issue.

XhmikosR (approved):

PR #31133, opened 2020-06-21
Docs: screen reader helper/visibility utility tweaks

patrickhlauke:
- rename "Screen readers" helper to "Visually hidden" (in future, may consider actually renaming the classes as well? `.visually-hidden` / `.visually-hidden-focusable` or similar?)
- reword description for visibility utility, emphasise/call out the fact that `.invisible` hides stuff from AT

### DIFF
--- a/scss/_helpers.scss
+++ b/scss/_helpers.scss
@@ -2,6 +2,6 @@
 @import "helpers/colored-links";
 @import "helpers/embed";
 @import "helpers/position";
-@import "helpers/screenreaders";
+@import "helpers/visually-hidden";
 @import "helpers/stretched-link";
 @import "helpers/text-truncation";
--- a/scss/helpers/_visually-hidden.scss
+++ b/scss/helpers/_visually-hidden.scss
@@ -1,5 +1,5 @@
 //
-// Screenreaders
+// Visually hidden
 //
 
 .visually-hidden,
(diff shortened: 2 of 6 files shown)


patrickhlauke:
meta-question: what's the cut-off between something being a "helper" and something being a "utility"?

patrickhlauke:
Related, the proposed PR to actually rename the classes/mixins as well #31139

XhmikosR:
Note that we should focus to v5 and leave v4 in maintenance mode. We don't need to backport everything. Also, this needs to be handled carefully in v4 with a redirect.

patrickhlauke:
ok, there's probably little point in backporting just this bit. focusing on v5 makes sense.

patrickhlauke:
@XhmikosR should this also have an `aliases`?

XhmikosR:
> 
> 
> @XhmikosR should this also have an `aliases`?

Doesn't hurt but maybe we don't need to, since it's still an alpha. I'll add a note to revisit the aliases later.

mdo:
> meta-question: what's the cut-off between something being a "helper" and something being a "utility"?

Generally speaking, utilities are single property-value pairings and helpers are shortcuts or snippets.

patrickhlauke:
any chance of a review please? ;)

XhmikosR (approved):
LGTM but I want @mdo to also review

ffoodd (approved):
LGTM

PR #31135, opened 2020-06-21
Docs: move color utility callouts to start of page

patrickhlauke:
Hierarchically/structurally, in the position they are currently at, the two callouts seem like they "belong" just to the "background color" section. Moving them to the start makes it clearer that those two callouts relate to everything in the page (both "Color" and "Background color" classes).

### DIFF
--- a/site/content/docs/5.0/utilities/colors.md
+++ b/site/content/docs/5.0/utilities/colors.md
@@ -6,6 +6,16 @@ group: utilities
 toc: true
 ---
 
+{{< callout info >}}
+#### Dealing with specificity
+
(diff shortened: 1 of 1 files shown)


XhmikosR (approved):

XhmikosR:
@patrickhlauke Just FYI, all these doc backports need to be made manually due to v4-dev using Jekyll.

XhmikosR:
Also, please use the proper labels/projects.

patrickhlauke:
> Also, please use the proper labels/projects.

is this documented anywhere?

XhmikosR:
No. Just add the projects if something needs to be backported like we do for all PRs.

PR #31137, opened 2020-06-21
Docs: (subjectively) nicer looking accordion example

patrickhlauke:
the use of `.btn-link` currently just looks a bit odd, subjectively (as in v5 we now force the underline to always be there, while in v4 at least it was only on hover/focus).
`.btn-light` seems more fitting


patrickhlauke:
Current look in v5:

> accordion-current

After change:

> accordion-light

ysds:
I suggest more additional classes :)
* Add a `p-0` to the `.card-header`
* Add a `p-3` and `rounded-0` to the button

Before:
image


After:
image

patrickhlauke:
the example with more classes does look good. wonder though if it will obscure the point of the example?

mdo:
There's a PR to make a dedicated accordion component at #30042. Maybe we explore that?

XhmikosR:
I think this PR with the suggestions @ysds made is good to land way before #30042

patrickhlauke:
Added extra classes as suggested by @ysds

patrickhlauke:
Also squeaked in a tiny clarification in the accessibility note there about what bootstrap doesn't support (the *optional* keyboard interactions)

patrickhlauke:
any more thoughts on this?

mdo (approved):

PR #31160, opened 2020-06-24
Update getSelectorFromElement to handle invalid selectors

janklimo:
Bootstrap v5 introduced a regression wherein an unhandled `DOMException` would be thrown if initialising a dropdown element on an anchor tag with a standard URL in the `href` tag.

This is a realistic scenario when the UI shows dropdown on hover but still allows the user to click the link.

This PR adds error handling for this scenario and a spec to cover the change, essentially reimplementing how this is handled in v4:

https://github.com/twbs/bootstrap/blob/4f44a2afd0acbf8a2d199c6fffe619b91dee1c99/js/src/util.js#L88-L92

### DIFF
--- a/js/src/util/index.js
+++ b/js/src/util/index.js
@@ -48,7 +48,11 @@ const getSelectorFromElement = element => {
   const selector = getSelector(element)
 
   if (selector) {
-    return document.querySelector(selector) ? selector : null
+    try {
+      return document.querySelector(selector) ? selector : null
+    } catch (_) {
+      return null
+    }
   }
 
   return null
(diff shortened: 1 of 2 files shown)


XhmikosR:
I think this was intentional, /CC @Johann-S.

ffoodd:
Dropdown should not open on hover and make another action on click, that'd be an accessibility failure since a keyboard user wouldn't be able to access whatever appears on hover.

You need to use the Split button to implement such cases (basically a link beside a dropdown, grouped together).

BTW our dropdown don't open on hover.

Johann-S:
Yes it was intentional, we shouldn't silent errors, because it'll help our users to understand why there is something wrong.

BTW @XhmikosR I don't know if you remember that, but we talk about stopping using `href` to target element and instead only use `data-target` and I think it's a good opportunity to do that, it'll avoid this kind of issues

XhmikosR:
Closing for the aforementioned reason.

@Johann-S we should investigate that in #31185

Issue #30844, opened 2020-05-18
Toast background filter blur rendering issue in Safari


Katrienvh:
Rendering issue toast in safari only visible when there's an image behind the toast.

The background blur is broken at the left side of the toast. When you don't use a header, it's also broken at the top. 
![image](https://user-images.githubusercontent.com/12270833/82203834-22252500-9904-11ea-8244-3b761882d714.png)

I also found out, that when you remove the box-shadow of the toast, the issues disappears. As a box shadow for a toast is a must in my eyes, this is not a solution.

I made a codepen to as example: https://codepen.io/katrienvh/pen/JjYebmy

- Operating system and version: macOS Catalina version 10.15.4
- Browser and version: Safari version 13.1

mdo:
Yeah, I've been noticing some issues in Safari with this CSS in a few locations. Will look into it further to see if there's a fix, and if not, I'll likely just remove it.

PR #31161, opened 2020-06-24
Fix bad backdrop-filter rendering on Safari

Gwe-n:
This PR is a proposed fix for the bug https://github.com/twbs/bootstrap/issues/30844

In order to have the same behavior on the different browsers we could imagine positioning the backdrop-filter and box-shadow effect on pseudo-elements. 

Before : 

Capture d’écran 2020-06-24 à 16 23 12

After : 
Capture d’écran 2020-06-24 à 16 23 28

### DIFF
--- a/scss/_toasts.scss
+++ b/scss/_toasts.scss
@@ -1,4 +1,5 @@
 .toast {
+  position: relative;
   max-width: $toast-max-width;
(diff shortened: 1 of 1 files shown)


mdo:
Closing for #31165.

PR #31176, opened 2020-06-25
V4 dev backports and updates

XhmikosR:
Notes:

1. I can't seem to be able to find an Edge version older than 15 on BrowserStack, so I went with 15 which is the oldest one I see
2. The #30326 backport needs the test backported too. /CC @ysds for this


ysds:
@XhmikosR Can I push the unit test to the `v4-dev-xmr` branch directly?

XhmikosR:
Of course :)

ysds:
> The #30326 backport needs the test backported too. /CC @ysds for this

Done :)

MartijnCuppens (approved):

PR #31180, opened 2020-06-25
fix failure to hide tooltip if already destroyed

jwilson8767:
This one is a super simple bugfix, let me know if there's anything else I need to do (rebuild / move to different branch / etc).

Cheers!


XhmikosR:
We already have #31115 which is pending review, isn't this about the same issue?

XhmikosR:
I'm going to close this because #31115 is older and also has a test case, but thanks for the patch!

jwilson8767:
Didn't see that one! Thanks for the quick review!

PR #31185, opened 2020-06-26
use only data target to query elements in our plugin

Johann-S:
With this PR we'll rely only on `data-target` to find targeted elements. The only exception is our Scrollspy which use a lot of link, it'll still be possible to use selectors in links for this plugin.

Related to: #31160

Preview: https://deploy-preview-31185--twbs-bootstrap.netlify.app/


site/content/docs/5.0/components/collapse.md
```diff
@@ -42,12 +42,12 @@ You can use a link with the `href` attribute, or a button with the `data-target`
 
 ## Multiple targets
 
-A `<button>` or `<a>` can show and hide multiple elements by referencing them with a selector in its `href` or `data-target` attribute.
-Multiple `<button>` or `<a>` can show and hide an element if they each reference it with their `href` or `data-target` attribute
+A `<button>` or `<a>` can show and hide multiple elements by referencing them with a selector in its `data-target` attribute.
+Multiple `<button>` or `<a>` can show and hide an element if they each reference it with their `data-target` attribute
```
patrickhlauke, line 46: as earlier it says "any element", maybe here it should also keep it more general than just `<button> or <a>`

patrickhlauke:
this made me remember a bugbear that i kept meaning to post #31186

site/content/docs/5.0/components/collapse.md
Johann-S: I agree 👌

Johann-S:
@patrickhlauke I think it'll fix your issue too 🤔

patrickhlauke:
> @patrickhlauke I think it'll fix your issue too 🤔

yeah i meant it more as "once this lands, we should also do this" and then avoid unnecessarily using `<a href="#" data-target="..." role="button">` when a real button is the preferred/logical/semantic choice. while mentioning that links can be used, we should set a good example of using the more correct element as trigger for in-page actions

XhmikosR:
The "problem" with swapping `a` tags with `button`s is that it requires CSS changes in our components, but that's for another place to discuss.

Do note that we should probably try to keep the breaking changes to a minimum to make things easier for people to upgrade. Or perhaps, we should make them all at once :P

I'd definitely like to hear @mdo's opinion before proceeding with this. Have `a href="#"` all over the place does not seem good either, but if it's an intermediate solution I guess we could live with it for a while.

js/tests/unit/util/index.spec.js
```diff
@@ -35,36 +35,6 @@ describe('Util', () => {
       expect(Util.getSelectorFromElement(testEl)).toEqual('.target')
     })
 
-    it('should get selector from href if no data-target set', () => {
```
XhmikosR, line 38: Shouldn't we move these tests to scrollspy?

Johann-S: good catch 👍

patrickhlauke:
> The "problem" with swapping `a` tags with `button`s is that it requires CSS changes in our components, but that's for another place to discuss.

yup, certainly for carousel, possibly other places like tabs (navs with js). let's discuss in #31186 

> Do note that we should probably try to keep the breaking changes to a minimum to make things easier for people to upgrade. Or perhaps, we should make them all at once :P

Was trying to get some of these breaking ones in while 5 is still in alpha...otherwise it'll be another 2-3 years before we can change anything again ;)

XhmikosR:
>Was trying to get some of these breaking ones in while 5 is still in alpha...otherwise it'll be another 2-3 years before we can change anything again ;)

That's not true since we have moved to a faster release cycle. :)

mdo:
Any status update here?

Johann-S:
@mdo it seems we are waiting for your decision

septatrix:
Wouldn't this be a degradation for JS-less users?

mdo:
Closing as stale.

Issue #31088, opened 2020-06-18
popovers: do not dismiss on next click inside


inwardmovement:
I think [popovers dismissed on next click](https://getbootstrap.com/docs/4.5/components/popovers/#dismiss-on-next-click) should not dismiss when the click is inside the popover, to allow selecting text, etc.

patrickhlauke:
I'd agree with this, yes

PR #31191, opened 2020-06-28
popovers: do not dismiss on next click inside

RobustProgram:
Added code to tooltip.js in response to the [issue](https://github.com/twbs/bootstrap/issues/31088)


XhmikosR:
You are missing the src files and the dist files are not supposed to be committed.

RobustProgram:
Hi @XhmikosR I just modified the tooltip.js in the src folder. So, I am not sure what you mean by missing the src files.

I will remove the dist files from the commits.

Thanks!

PR #31196, opened 2020-06-28
Fixed typo in docs v5 (Utilities/API) page

inf1nite-lo0p:
Seems like there is a typo in : https://v5.getbootstrap.com/docs/5.0/utilities/api/
```
$utilities: (
  "opacity": (
    property: opacity,
    values: (
      0: 0,
      25: .25,
      50: .5, // This should be 75: .75
      100: 1,
    )
  )
 );
```
Since the generated class is : 
```
.opacity-75 {
  opacity: .75;
}
```


XhmikosR:
@MohammadxAli please add the 2 cases @MartijnCuppens suggested :)

mdo (approved):

PR #30980, opened 2020-06-08
Add RTL support

ffoodd:
Fixes #30918, fixes #24807, fixes #27783, fixes #31845

[Preview](https://deploy-preview-30980--twbs-bootstrap.netlify.app/) — [Examples](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/) — [cheatsheet](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/cheatsheet/) — [RTL cheatsheet](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/cheatsheet-rtl/)

## Beta 2

- [ ] Move cheatsheet from examples to main docs
- [ ] Allow main docs to be passed a `direction` in the front matter and have that automatically toggle LTR to RTL CSS inclusion
- [ ] _The above items should resolve most of this, but..._ De-dupe the RTL CSS once moved
- [ ] Convert page content to Markdown
- [ ] And https://github.com/twbs/bootstrap/pull/30980#issuecomment-737096962

### Cheatsheet

- [x] use `button`s in cheatsheet side navigation, as it's been done in our main docs recently
- [x] Fix the sticky components title covering the page content
- [x] Fix floating labels example (remove row)
- [x] Make the header a little less huge
- [ ] check and update translations in examples and cheatsheet

### Docs

- [x] check again I didn't break @mdo's changes to the docs
- [x] ensure there's no missing occurrences of our old-fashioned classes (will take care of that when we'll be about to merge)
- [x] Fix alignment utilities breakage https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/utilities/position/
- [x] Search docs for `left` and `right` in case we missed something

## Done

- [x] **add** RTLCSS and dedicated scripts

<details>
<summary>Test cases</summary>

  - [x] convert some example templates to `dir="rtl"`—those on which direction makes a difference:
    - [x]  [Album](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/album-rtl/)
    - [x] [Checkout](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/checkout-rtl/)
    - [x] [Dashboard](https://deploy-preview-30980--twbs-bootstrap.netlify.app/docs/5.0/examples/dashboard-rtl/)
[...]

PR #31210, opened 2020-06-30
RTL: refactor using logical names

ffoodd:
Mostly a POC for now, here's a first try on using logical names for physical properties.

It's meant to:
1. keep naming consistent between RTL and LTR (eg. prevent `.float-right { float: left; }` in RTL)
2. ease the switch to custom properties one day, since naming would already be logical.

There's still things to consider here, I'm not done with SCSS yet (missing components, carousel, dropdown, popovers or tooltips for the most noticeable) and it'd require to apply those changes in our JS as well as in our docs (and to be heavily documented in the migration page).

@twbs/css-review is this a way I should follow right away?

This is highly questionnable, being consistent requires to rename for example `caret-down` to `caret-block-end`, or to use new abreviations for our utilities (`.mt` becomes `.mbs` for `block-start`, .etc). I tried to be consistent but there's an real amount of work here.

## To do

- [x] update **migration** guide
- [x] update **docs** everywhere
- [x] update docs **theming**

### Questions

What to do with:

- [x] ~~**RFS** mixins?~~ → **to tackle in RFS directly**
- [x] ~~**Dropdowns, popovers and tooltips**, relying on Popper?~~ → **JS related stuff, not only renaming…**
- [x] **Carousels** class names (eg `.carousel-item-left`) → **Rename**.
- [x] `_position.scss` defines fixed and sticky helpers for `top` and `bottom`. They're not impacted by RTL but will require refactor to use `inset` values and subvalues (eg. `inset-block-start`) when switching to logical properties. Should we rename classes to be consistent with the whole PR? → **Do nothing**.


ffoodd:
@twbs/css-review I think I'm done with renaming here: please step in to review the CSS part before I continue with the documentation part. There are a few questions in the PR description, BTW :)

scss/_button-group.scss
```diff
@@ -76,12 +76,12 @@
   padding-left: $btn-padding-x * .75;
 
   &::after,
-  .dropup &::after,
-  .dropright &::after {
+  .drop-bs &::after,
+  .drop-ie &::after {
```
mdo, line 80: These abbreviations are a little cryptic—I took `bs` and `ie` as Bootstrap and IE. 😅 Haven't reviewed everything here, still thinking on the rest.

mdo:
One more comment while I'm here—I _think_ our `top`/`bottom` classes can stay as-is and `left`/`right` can change to `start`/`end`. This keeps half our classes intact and reinforces a shared language across horizontal and vertical alignment.

Vertical: top, middle, bottom
Horizontal: start, center, end

Does this get us too far away from logical properties? Is it a weird bastardization? I want to keep these class names approachable.

ffoodd:
@mdo yeah I've been thinking the same thing, considering the amount of changes here. The target is RTL only for v5, so logical names should only be a first step for this release. 

I'll make a step back, and topics like dropdowns, tooltips and popovers will get simplified. However that'd come to us when we'll switch to logical properties: that might require to work on Popper.js at some point…

Consider this first commit as a v6 preview or something—I'll trash it while rebasing :+1:

scss/_button-group.scss
ffoodd: Indeed—it'll be dropped for now, but we probably have to get used to it in a (not so) far future since `*-block-start` and `*-inline-end` are the new `*-top` and `*-right` :D

ffoodd:
@mdo I think it's ready now. To sum up:
1. only `left` and `right` things were renamed — `.ms` and `.me` for `margin-left` and `margin-right` being the most shocking changes to me (but way less than my previous attempt)
2. docs have been revamped, I guess wording and phrasing could be better — just let me know :)
3. a new RTL section is introduced in the migration guide, above the alpha2 since I don't think it'd be ready for alpha 2,
4. Popper.s related questions should better be solved in main PR before rename anything here (I think there'll be more JS related topics than CSS ones)
4. RFS is a completely different topic, I'd have to open a PR against RFS directly.

Do not forget that it can be merged in my main RTL branch and reworked later :D

ffoodd:
Thinking about RFS mixins, wouldn't it be better to simply add a few aliases in Bootstrap instead of introducing a BC in RFS?

Something like:
```scss
@mixin padding-start($value) {
  @include padding-left($value);
}
```
We'd use our aliases in Bootstrap core, without touching to RFS — that could still have a branch or next release using logical names, maybe logical properties too?

Any thought, @mdo & @MartijnCuppens?

NB: TBH I don't think this is mandatory to tackle when coming to RTL in v5 since RTL output is post-processed based on dist files. This is mostly to be more convenient and accurate for our RTL users (which I hope will be more of them in v5!).

MartijnCuppens:
Not sure why and where we need additional mixins? We could use `@include rfs($padding-start, padding-left);`

ffoodd:
I didn't search for usage of those mixins, that might not be needed at all. However since they're in our sources, they can be used by our Sass users.

Still thinking out loud. Since we'll write CSS being LTR-first, that might not be usefu.

Not sure yet 🤔

ffoodd:
@MartijnCuppens You're right, I think there's nothing needed RFS side: we don't use its direction-related mixins in our codebase. So until the feature is asked for in RFS, I guess we're safe :+1: 

So the latest pain point in #30980 concerns Popper.js.

This specific PR is definetely ready to review :)

FaridAghili:
I've used Google's Flutter for android development which supports RTL by default.
They use start and end instead of left and right as their standard without changing top and bottom to anything else.

PR #31212, opened 2020-06-30
Minor JS consistency changes

XhmikosR:


js/src/tooltip.js
```diff
@@ -676,12 +676,11 @@ class Tooltip {
   _getConfig(config) {
     const dataAttributes = Manipulator.getDataAttributes(this.element)
 
-    Object.keys(dataAttributes)
-      .forEach(dataAttr => {
-        if (DISALLOWED_ATTRIBUTES.indexOf(dataAttr) !== -1) {
-          delete dataAttributes[dataAttr]
-        }
-      })
+    Object.keys(dataAttributes).forEach(dataAttr => {
```
Johann-S, line 679: I would keep the indent because we chain calls here

XhmikosR: There's not another instance we break after `Object.keys` in the whole codebase...

Johann-S: so maybe we should apply that to the other calls of `Object.keys`, shorter line is always better for the reading

XhmikosR: I think it's fine here.

XhmikosR: FYI:

```
Search "Object.keys" (11 hits in 7 files)
  C:\Users\xmr\Desktop\bootstrap\build\build-plugins.js (1 hit)
	Line 180:     await Promise.all(Object.keys(bsPlugins).map(plugin => build(plugin)))
  C:\Users\xmr\Desktop\bootstrap\js\src\dom\event-handler.js (5 hits)
	Line 129:   const uidEventList = Object.keys(events)
	Line 212:   Object.keys(storeElementEvent).forEach(handlerKey => {
	Line 251:       Object.keys(events).forEach(elementEvent => {
	Line 257:     Object.keys(storeElementEvent).forEach(keyHandlers => {
	Line 304:       Object.keys(args).forEach(key => {
  C:\Users\xmr\Desktop\bootstrap\js\src\dom\manipulator.js (1 hit)
	Line 50:     Object.keys(attributes).forEach(key => {
  C:\Users\xmr\Desktop\bootstrap\js\src\tooltip.js (1 hit)
	Line 679:     Object.keys(dataAttributes)
  C:\Users\xmr\Desktop\bootstrap\js\src\util\index.js (1 hit)
	Line 113:   Object.keys(configTypes).forEach(property => {
  C:\Users\xmr\Desktop\bootstrap\js\src\util\sanitizer.js (1 hit)
	Line 103:   const allowlistKeys = Object.keys(allowList)
  C:\Users\xmr\Desktop\bootstrap\js\tests\browsers.js (1 hit)
	Line 63: const browsersKeys = Object.keys(browsers)
```

So, it's basically only this instance and also we don't enforce it, which TBH is better since this is really just a trivial thing :)

bardiharborow (approved):

PR #31231, opened 2020-07-03
Use collapse plugin for the "Show components requiring JavaScript" dropdown

techboyg5:
The "Show components requiring JavaScript" dropdown menu currently uses `<details>` and `<summary>`. It can easily be converted to the collapse plugin to be consistent with the sidebar.


MartijnCuppens:
Thanks for the suggestion, but we rather stick with the details/summary option since it's more semantic for this specific case.

techboyg5:
> Thanks for the suggestion, but we rather stick with the details/summary option since it's more semantic for this specific case.

It looks mostly the same however and works with older browsers, unlike details/summary.

PR #31233, opened 2020-07-03
Add Hugo Pipes logic for local and vendored scripts.

regisphilibert:
@XhmikosR 

I stored the Hugo pipe logic in a returning partial so we can:
1. not concern ourselves with whitespace and line breaks when working logics
2. use partialCached on the logic
3. compontentize.

Also: 
- When missing mounts, I'm using warnings instead of errors, but I think you wanted errors so the build is stopped. In this case we can replace `warnf` by `errorf` through the code.
- For now the `docsearch.min.js` file is not published at the desired permalink (including version) but will after Hugo .74 introduces the new `resources.Put` method (see comment)
- I'm not 100% sure the order in which files should be bundled is correct, but I don't have any errors in console.
- I used my own [conventions](https://regisphilibert.com/blog/2019/12/hugo-partial-series-part-2-functions-with-returning-partials/#some-conventions) for file naming and commenting, feel free to adjust to the team's own.
- I left alone the asset loaded from CDN (`bootstrap.bundle.min.js`), but once you are ready to let Hugo handle it we'll just have to include in the GetJSAssets partial.

Preview: <https://deploy-preview-29283--twbs-bootstrap.netlify.app/>

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -40,10 +40,12 @@ module:
       target:           static/apple-touch-icon.png
     - source:           site/static/docs/5.0/assets/img/favicons/favicon.ico
       target:           static/favicon.ico
-    # docsearch is referenced in a `script` tag so we just mount it in the static folder
+    
+    # the following are processed via Hugo pipes so we need them in the assets folder
+    ## the following will be printed with Hugo pipes
     - source:           node_modules/docsearch.js/dist/cdn/docsearch.min.js
-      target:           static/docs/5.0/assets/js/vendor/docsearch.min.js
-    # the following are concatenated via Hugo pipes with docs.js so we need them in the assets folder
+      target:           assets/js/vendor/docsearch.min.js
+    ## The following vendor files will be bundled together on top of local js files as /docs.js
     - source:           node_modules/anchor-js/anchor.min.js
       target:           assets/js/vendor/anchor.min.js
     - source:           node_modules/bs-custom-file-input/dist/bs-custom-file-input.min.js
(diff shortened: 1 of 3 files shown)


XhmikosR:
First of all, sorry for the late response; the weird times we are living in have hit me hard :/

I rebased my branch and yours, made a few typo fixes and renamed the partial.

> * When missing mounts, I'm using warnings instead of errors, but I think you wanted errors so the build is stopped. In this case we can replace `warnf` by `errorf` through the code.

Yup, that's the whole idea otherwise I think we could simplify this a lot.

> * For now the `docsearch.min.js` file is not published at the desired permalink (including version) but will after Hugo .74 introduces the new `resources.Put` method (see comment)

We are using a fairly recent Hugo build so feel free to adapt the code :)

> * I'm not 100% sure the order in which files should be bundled is correct, but I don't have any errors in console.

As long as the vendor files come first, it should be fine indeed

> * I left alone the asset loaded from CDN (`bootstrap.bundle.min.js`), but once you are ready to let Hugo handle it we'll just have to include in the GetJSAssets partial.

I think that's for another time since we currently just include the dist files (please correct me if I misunderstood something).

---

That being said, how do you feel of this approach in general? It seems it doesn't scale totally in the sense that we have 2 places we have to handle the vendor files; one is config.yml and the other one is the new partial. What's the alternative solution? Dropping this solution and just go with my patch only and in case a vendor file is missing we'll just get a JS runtime error?

Thanks again for the patch, I really appreciate helping us, especially now that we are moving towards v5 stale :)

XhmikosR:
I'm going to merge this to my branch for now and we can tweak it later hopefully :)

PR #31240, opened 2020-07-05
Single line example for inline forms

### LINKED ISSUE #31230: Docs: inline forms example wraps to two lines in new v5 layout
due to the narrower central content column, https://v5.getbootstrap.com/docs/5.0/forms/layout/#inline-forms wraps to two lines even at large 1920x1080 browser window size. while it's good that it shows that it adapts, maybe we should provide a slightly smaller example that's at least on one line at the large size (e.g. drop one of the fields?)

> inline form example at large browser window size with submit button on second line

rohit2sharma95:
Closes #31230

### DIFF
--- a/site/content/docs/5.0/forms/layout.md
+++ b/site/content/docs/5.0/forms/layout.md
@@ -302,11 +302,6 @@ Be sure to always include a `<label>` with each form control, even if you need t
 {{< example >}}
 <form class="row row-cols-md-auto g-3 align-items-center">
   <div class="col-12">
-    <label class="visually-hidden" for="inlineFormInputName">Name</label>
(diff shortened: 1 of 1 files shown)


MartijnCuppens (approved):

XhmikosR:
Maybe we should switch to `row-cols-lg-auto`? BTW it seems the other examples above suffer too in some breakpoints between md and lg

rohit2sharma95:
Should I update PR then? @XhmikosR

XhmikosR:
Not sure yet, maybe better wait until someone else chimes in too :)

mdo (approved):

PR #31242, opened 2020-07-05
Use transparent for background gradient

techboyg5:
`transparent` should be used for the background gradient.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -226,7 +226,7 @@ $enable-important-utilities:  true !default;
 //
 // The gradient which is added to components if `$enable-gradients` is `true`
 // This gradient is also added to elements with `.bg-gradient`
-$gradient: linear-gradient(180deg, rgba($white, .15), rgba($white, 0)) !default;
+$gradient: linear-gradient(180deg, rgba($white, .15), transparent) !default;
 
 // Spacing
 //


ffoodd:
`transparent` is not the same as `rgba(255, 255, 255, 0)` since it is an alias of `rgba(0, 0, 0, 0)` (as per [the spec](https://www.w3.org/TR/css-color-3/#transparent).

It means the gradient would now go from white to black, it could be noticeable. I don't think there's a point here.

Any thought, @twbs/css-review ?

mdo:
Agreed on leaving as-is. Thanks!

techboyg5:
> `transparent` is not the same as `rgba(255, 255, 255, 0)` since it is an alias of `rgba(0, 0, 0, 0)` (as per [the spec](https://www.w3.org/TR/css-color-3/#transparent).
> 
> It means the gradient would now go from white to black, it could be noticeable. I don't think there's a point here.
> 
> Any thought, @twbs/css-review ?

No, it's still fully transparent

ffoodd:
Well, that's why I said "could". FYI, here's [a CSS-Tricks psot on the topic](https://css-tricks.com/thing-know-gradients-transparent-black/)—mostly a Safari bug.

techboyg5:
> 
> 
> Well, that's why I said "could". FYI, here's [a CSS-Tricks psot on the topic](https://css-tricks.com/thing-know-gradients-transparent-black/)—mostly a Safari bug.

OK, thanks!

PR #31243, opened 2020-07-06
Remove period in list item link for consistency.

tixic:
To remain consistent with the other list items in the `Several quick start options are available` section, I removed the period in the first list item link.


patrickhlauke:
good catch, thanks

patrickhlauke (approved):

PR #31245, opened 2020-07-06
Remove Browser support link/note from Floating Labels Example

coliff:
Now that Placeholder shown is widely supported and this example works with all browsers which Bootstrap 5 target, this link can safely be removed.
REF: https://caniuse.com/#feat=css-placeholder-shown


ffoodd:
It doesn't, [Bootstrap 5 will support Edge >= 16](https://github.com/twbs/bootstrap/blob/main/.browserslistrc).

coliff:
ah, I missed that. thanks for pointing that out :-)

coliff:
Maybe the message should be changed, to something like 'View browser support'?

PR #31252, opened 2020-07-07
Removed vertical-align from form-select class

mikayelgr:
vertical-align: middle Property is ignored due to the display. With 'display: block', vertical-align should not be used.

### DIFF
--- a/dist/css/bootstrap.css
+++ b/dist/css/bootstrap.css
@@ -4,9726 +4,9725 @@
  * Copyright 2011-2020 Twitter, Inc.
  * Licensed under MIT (https://github.com/twbs/bootstrap/blob/main/LICENSE)
  */
-:root {
-  --bs-blue: #0d6efd;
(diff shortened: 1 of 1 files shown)


alecpl:
You should not touch dist files.

MichaelGrigoryan25:
I have just changed a one line. Everything else has stayed as it was.

alecpl:
Did you see your change? It is not even possible to see what have you changed.

Dist files are generated from other files, so you have to find the proper file with your line and do a change there.

MichaelGrigoryan25:
> 
> 
> Did you see your change? It is not even possible to see what have you changed.
> 
> Dist files are generated from other files, so you have to find the proper file with your line and do a change there.

Yeah I see the commit now. Lemme see how can I fix this

MichaelGrigoryan25:
I'll just create a new PR

PR #31258, opened 2020-07-08
Update FUNDING.yml

AayushSinha107:


XhmikosR:
Thanks but this doesn't really help anything

patrickhlauke:
i see what you're trying to get at (as it's a single platform), but then you'd need to change the whole sentence, and if in future there's another platform we add, we'd have to change the whole sentence again...

PR #31259, opened 2020-07-08
README.md: show brotli size too

XhmikosR:
This was something @MartijnCuppens suggested some time ago on Slack.

My only concern is that gzip is the defacto compression used. We could probably keep both sizes in README.md as a middle solution?

Also, not sure if we should backport this or not.

/CC @mdo @ffoodd @MartijnCuppens @Johann-S


MartijnCuppens:
> My only concern is that gzip is the defacto compression used. We could probably keep both sizes in README.md as a middle solution?

Sounds good

> Also, not sure if we should backport this or not.

BS4 doesn't support Brotli, so backporting doesn't feel right.

XhmikosR:
> BS4 doesn't support Brotli, so backporting doesn't feel right.

That's unrelated 😛

XhmikosR:
Another solution would be to maybe add a separate table with the dist files and their sizes?

MartijnCuppens:
> BS4 doesn't support Brotli, so backporting doesn't feel right.
> That's unrelated 😛

I mean IE, woops.

XhmikosR:
> 
> 
> > BS4 doesn't support Brotli, so backporting doesn't feel right.
> > That's unrelated 😛
> 
> I mean IE, woops.

Yeah, but we don't care about that here. It's a useful metric and servers do fall back to gzip.

XhmikosR:
So, to summarize, we can check for both gzip and brotli in bundlewatch and show both in README.md.

@mdo WDYT?

mdo:
Sounds reasonable to me!

XhmikosR:
How about just showing both in README.md for starters? What we use in bundlewatch shouldn't matter a lot.

mdo (approved):

PR #31272, opened 2020-07-10
Update version.md to include v5

Laim:
Version.md on v5.getbootstrap.com is only referring to v1 > v4 when it should be referring to v1 > v5.

**Preview**: https://deploy-preview-31272--twbs-bootstrap.netlify.app/docs/versions/


mdo:
Cherry-picked into #31277.

PR #31274, opened 2020-07-10
Removed vertical-align:middle from both minified and standard CSS files

mikayelgr:
---Reason---
Property is ignored due to the display. With 'display: block', vertical-align should not be used.


mdo:
Please revisit the contributing guidelines for our project—we don't directly modified our `dist` files.

PR #31276, opened 2020-07-10
v5: Tweak some colors

mdo:
Attempts to make green and cyan more accessible. Here's current/old and the new from this PR:

old

new

You'll also note I was playing with some padding and border-radius here. Might still share that soon.


patrickhlauke:
could also go the other way with the green, with black text and something like #43CB62 ?

i like the cyan

MartijnCuppens (approved):
Looks neat!

XhmikosR:
I think we should mention the change in migration and assuming everyone is fine with the new green it should be good to go :)

ffoodd (approved):
♥️

PR #31277, opened 2020-07-10
v5: Versions cleanup

mdo:
I'd like to be able to sort this page in the reverse order, showing v5 first. I still struggle with Hugo's syntax and couldn't first out if I needed to use `sort` here or `reverse`, or where to even place it. I give up on it.

Otherwise, this cleans up some spacing and pulls in the chnage from #31272.


XhmikosR:
You mean show v5, v4 etc from left to right?

XhmikosR (approved):

PR #31280, opened 2020-07-11
Extra position utilities

jaumesala:
## Introduction
I've been using bootstrap since v1 and I've always wished for this kind of utilities to be present on the framework. 
Given that there are utilities for the *position* property, it seems logic to have utilities for the *top*, *left*, *bottom* and *right* propertires. I've also added some utilities to center the elements using *translations*.

### Arrange elements

Arrange elements easily with the edge positioning utilities. The format is `{property}-{position}`.

Where *property* is one of:

- `top` - for the vertical `top` position
- `left` - for the horizontal `left` position
- `bottom` - for the vertical `bottom` position
- `right` - for the horizontal `right` position

Where *position* is one of:

- `0` - for `0` edge position
- `50` - for `50%` edge position
- `100` - for `100%` edge position

Screenshot 2020-07-11 at 18 39 05

```html
<div class="position-relative">
  <div class="position-absolute left-0 top-0"></div>
  <div class="position-absolute right-0 top-0"></div>
  <div class="position-absolute left-50 top-50"></div>
  <div class="position-absolute right-50 bottom-50"></div>
  <div class="position-absolute right-0 bottom-0"></div>
  <div class="position-absolute left-0 bottom-0"></div>
</div>
```

### Center elements

In addition, you can also center the elements with transform utilities. The format is `tt-{x-axis}{y-axis}`.

Where *x-axis* is one of:

- `x` - for `50%` horizontal translation
- `nx` - for `-50%` horizontal translation

Where *y-axis* is one of:

- `y` - for `50%` vertical translation
- `ny` - for `-50%` vertical translation

Screenshot 2020-07-11 at 18 39 14

```html
<div class="position-relative">
  <div class="position-absolute left-0 top-0 tt-nxny"></div>
  <div class="position-absolute left-50 top-0 tt-nxny"></div>
  <div class="position-absolute right-0 top-0 tt-xny"></div>
  <div class="position-absolute right-0 top-50 tt-xny"></div>
  <div class="position-absolute left-0 bottom-0 tt-nxy"></div>
  <div class="position-absolute left-50 bottom-0 tt-nxy"></div>
  <div class="position-absolute right-0 bottom-0 tt-xy"></div>
  <div class="position-absolute left-0 top-50 tt-nxny"></div>
  <div class="position-absolute left-50 top-50 tt-nxny"></div>
</div>
```

I've used the format `tt-{x-axis}{y-axis}` to denote the *transform*: **translateX** **translateY**  action, but I'm open to a new prefix...

### Examples

Here are some basic real case usages to demonstrate these utilities.

Screenshot 2020-07-11 at 18 56 50

```html
<button type="button" class="btn btn-primary position-relative">
  Mails <span class="badge rounded-pill bg-secondary position-absolute top right tt-xny">+99</span>
  <span class="visually-hidden">unread messages</span>
</button>
```

Screenshot 2020-07-11 at 18 57 00

```html
<button type="button" class="btn btn-primary position-relative">
  Mails <span class="badge border border-light rounded-circle bg-danger position-absolute top-0 right-0 tt-xny p-1">
  <span class="visually-hidden">unread messages</span>
  </span>
</button>
```


scss/_utilities.scss
```diff
@@ -39,6 +39,66 @@ $utilities: map-merge(
...
+      property: right,
+      values: (
+        0: 0,
+        50: 50%,
+        100: 100%
+      )
+    ),
+    "transform-translate-x-y": (
+      property: transform,
+      class: tt-xy,
```
mdo, line 76: Brainstorming here...

- `translate-x-y`?
- `translate-middle`?
- `translate-center`?

scss/_utilities.scss
```diff
@@ -39,6 +39,66 @@ $utilities: map-merge(
...
+      values: (
+        null: (translateX(50%) translateY(50%))
+      )
+    ),
+    "transform-translate-x-negative-y": (
```
mdo, line 81: Seems less likely? Maybe drop>

scss/_utilities.scss
```diff
@@ -39,6 +39,66 @@ $utilities: map-merge(
       property: position,
       values: static relative absolute fixed sticky
     ),
+    "top": (
```
mdo, line 42: Would we want to be more prescriptive with these utilities? Instead of all four directions, we do all four _corners_?

For example, `.top-right` vs `.top-100.right-0`?

Rest of those could include `.top-center`, `.top-left`, `.bottom-left`, `.bottom-center`, and `.bottom-right`. Those would fit the use case shown in the PR here, but I don't want to be too limiting for folks.

mdo (commented):
Some questions and thinking out loud. About to be out for 1.5 weeks, so won't be able to respond for a bit, but I'm loving this exploration and use case!

scss/_utilities.scss
```diff
@@ -39,6 +39,66 @@ $utilities: map-merge(
...
+        0: 0,
+        50: 50%,
+        100: 100%
+      )
+    ),
+    "right": (
+      property: right,
+      values: (
+        0: 0,
+        50: 50%,
```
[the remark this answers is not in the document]
jaumesala: While coding I thought about it as well, but many use cases were coming to my mind. For instance:

A card with a centered avatar on the top
Screenshot 2020-08-12 at 19 57 59

Chevron on a selected menu item
Screenshot 2020-08-12 at 19 53 03

The anchor of a popup
Screenshot 2020-08-12 at 19 55 32

scss/_utilities.scss
jaumesala: Yup, we need to work on the naming 😅

scss/_utilities.scss
jaumesala: Maybe, I couldn't find another way to center an "outer" positioned bottom/right element.

scss/_utilities.scss
jaumesala: Maybe, I've explored two possible options to simplify the utilities.

First of all, let's say we decide to use the following classes:

- ```.top-left```
- ```.top-center```
- ```.top-right```
- ```.center-left```
- ```.center-center```
- ```.center-right```
- ```.bottom-left```
- ```.bottom-center```
- ```.bottom-right```

## Using left and top allways

All utilities should use left and top to position the child, so ```.bottom-right``` would be: 
```css
.bottom-right {
  left: 100%;
  top: 100%;
}
```

#### Pros
- We only need one class to center them. Let's call it ```.translate-center```
```css
.translate-centered{
  transform: translateX(-50%) translateY(-50%);
}
```

#### Cons
- ```.bottom-right``` is not **bottom: 0** **right: 0** but **top: 100%** **left: 100%**

- The element cannot be absolute positioned *inside* the parent. This is not possible:
Screenshot 2020-08-12 at 19 56 24

## Use class selectors to apply the correct props

For each utility, use the named props, so:
```css
.top-left {
  top:0; 
  left:0;
}
.bottom-right {
  bottom: 0; 
  right:0;
}
.center-right {
  top: 50%; 
  right: 0;
}
```

Then, using class selectors we can define the translation for each option:

```css
.top-left.translate-center {
    transform: translateX(-50%) translateY(-50%);
}
.bottom-right.translate-center {
    transform: translateX(50%) translateY(50%);
}
.center-right.translate-center {
    transform: translateX(50%) translateY(-50%);
}
```

### Pros
- Utility classes are named with the prop names (easy to understand)
- We only need one class to center them. ```.translate-center```

### Cons
- Requires more css selectors
- The element cannot be absolute positioned "outside" the parent
Screenshot 2020-08-12 at 19 56 16


Let's open the debate 🙂...

MartijnCuppens:
Thoughts on using:
- `.t-0`, `.t-50` & `.t-100`
- `.r-0`, `.r-50` & `.r-100`
- `.b-0`, `.b-50` & `.b-100`
- `.l-0`, `.l-50` & `.l-100`
- `.translate-middle` (which applies `transform: translate(-50%, -50%)`)

So instead of classes like `.top-left` we have `.t-0.l-0`. We should be able to create all examples @jaumesala mentioned.

jaumesala:
Sounds good to me!

I wonder if we should reserve ```t-*```, ```r-*```, ```b-*```, ```l-*```  for other utilities?

Let's see what others say. I'll update the PR once we agreed on the implementation.

mdo:
The directional names are short enough, so I don't know that we need to abbreviate all those (e.g., `.top-0.left-0` instead of `.t-0.l-0`). The former feels more legible to me.

scss/_utilities.scss
jaumesala: I've updated the utility name to `translate-middle` as discussed.

scss/_utilities.scss
jaumesala: Removed. Only one utility class to absolute center an element. `.translate-middle`

jaumesala:
PR updated!

- Utility classes with full directional names (e.g. `.top-0.left-0`) 
- Dropped all center utilities in favor of only one `translate-middle` class.
- Documentation updated with new classes.

mdo (approved):
Love it! Let's get @MartijnCuppens's opinion, too, though :).

mdo (changes requested):
@jaumesala Can you add a default Sass map that folks can customize and add to for these new values?

In `_variables.scss`, add:

```scss
$position-values: (
  0: 0,
  50: 50%,
  100: 100%
) !default;
```

And then use the maps in the `_utilities.scss` like so:

```scss
"bottom": (
  property: bottom,
  values: $position-values
),
```

mdo (approved):
Looks good!

mdo:
At some point I'd like us to also add those real life examples from the opening comment here. I think that would help explain and demonstrate the usefulness of these.

jaumesala:
Sure, I may have some spare time tomorrow.
@mdo which ones do you prefer?
- Mail button with label
- Mail button with red dot indicator
- Card with centered avatar
- ...

mdo:
I wouldn't mind all three if that's alright by you! Definitely the button ones, those are honestly what sold me on this PR when I first looked at it :).

jaumesala:
Ok I've added those real life button examples. 
Screenshot 2020-09-20 at 11 49 57
I've also done some brainstorming to think of another example that mixes multiple components. I've ended up doing this:
Screenshot 2020-09-20 at 11 49 51
As far I can recall, those are the most common "bootstrap-how to's" questions I've always read ;-)
Let me know what you think.

martinbean:
These utilities look amazing. Thank you, @jaumesala! 🎉

PR #31285, opened 2020-07-12
Add side modals

### LINKED ISSUE #24718: Drawer, sidebar, offcanvas nav

### LINKED ISSUE #27033: Feature request: side Modals

mdo:
This PR adds two new modifier classes to our `.modal` component, `.modal-left` and `.modal-right`. It uses these two classes to create offcanvas drawers for navigation, content, and more. I've included two slightly different examples for now: a nav component and a list group component. I'm thinking a new example (perhaps a sidebar shopping cart or additional dashboard layout) could help show more of this off, too.

Still needs more work, but wanted to capture the idea from comments in #29017.

**Source:** https://bootstrap-menu.com/detail-modal.html by @vosidiy

Fixes #27033, fixes #24718.

---

**Demo:** https://deploy-preview-31285--twbs-bootstrap.netlify.app/docs/5.0/components/modal/#sidebar-modals

Screen Shot 2020-07-12 at 10 50 47 AM

---

This could also probably be backported—thoughts?

TODO:

- [ ] Allow customizing side modals transition?
- [ ] Nuke the off canvas example or adapt it to use this solution?
- [ ] Fix the right padding issue


alecpl:
Bug: For me on Firefox the right-side modal has `padding-right: 16px` (for page scrollbar).

These look like a common mobile menus, so it's probably good to have them. However, I once build a multi-level menu system based on popovers, which are imo better suited for that than modals. On desktop it looked like normal popover, but on phones it was styled to look like your `.modal-right` (with some javascript to add backdrop layer, etc.).

polRk:
SCROL BUG on iOS.

michael-roth:
I found the following differences compared to Offcanvas component (#29017), I guess there should only be one in the future?

Maybe this is helpful for your decision.

**Bottom**
- no implementation of bottom Modal yet

**Backdrop**
- backdrop is animated in Modal component but not yet in Offcanvas
- disabling backdrop for Modal works with data-backdrop="false" but body is not scrollable nor otherwise interactive
  - Offcanvas has option: data-body="scroll" and the page can be used as if the Offcanvas were not open

**Layout**
- Modal has its own header and footer

**JS**
- Offcanvas does not have a dispose method

**Nesting**
```
Bootstrap only supports one modal window at a time. 
Nested modals aren’t supported as we believe them to be poor user experiences.
```
- Only one Offcanvas can be opened at the same time as well but that is an open point in https://github.com/twbs/bootstrap/pull/29017#issuecomment-551128164 (4.) I think

mdo:
With some updates, this is ready for review. I've fixed the right side modal glitch (transition was choppy, solution was separate `.modal-right.show` and `.modal-right.fade` declarations) and added examples of a slide-in and a fade.

Scroll bug on iOS is an existing bug with modals.

Firefox, not seeing an issue on latest for macOS.

mdo:
Thinking I’d like to try to backport this to v4.

XhmikosR:
Pushed a few updates, I will squash/clean up the patches later. One thing I notice is that when closing the sidebar, it's closed immediately; the transition seems off a bit.

ffoodd (dismissed):
LGTM

XhmikosR:
I still don't think this is ready due to my comment above :/

ffoodd:
You're right, I've been to quick on trying this.

XhmikosR:
BTW there is a white thingy at the bottom on mobile. Can also be seen when long pressing in the overlay area

mdo:
@MartijnCuppens @ffoodd Should we rename these classes to `.modal-start` and `.modal-end` ahead of the RTL feature shipping?

ffoodd:
I guess so, if we all agree with logical class names in RTL PR.

mrlife:
@mdo Do you think backport will make the cut for 4.6?

MartijnCuppens:
Just checked this and it looks like the padding issue (https://github.com/twbs/bootstrap/pull/31285#issuecomment-657991512) is still present.

HamedFathi:
Please consider the `Bottom` side too. It helps us to simulate `Material BottomSheet` too.
To be honest, [Offcanvas as component](https://github.com/twbs/bootstrap/pull/29017) was complete and amazing. In contrast, this issue is much more incomplete.

XhmikosR:
I don't know if #29017 is a breaking change, it probably is due to at least the `z-index` change. Otherwise, I'm in between this myself.

#29017 feels like a cleaner solution, we are not mixing up the components. I will try to rebase #29017 in the next days; it probably won't be too easy because we have landed a few breaking changes already, but it's worth a try.

HamedFathi:
@XhmikosR 

> #29017 feels like a cleaner solution, we are not mixing up the components.

We are on the same page. #29017 really is a perfect and reliable component.

rohit2sharma95:
Found padding-right in the right modal here: https://deploy-preview-31285--twbs-bootstrap.netlify.app/docs/5.0/components/modal/#sidebar-modals

That creates unnecessary space on the right side.

space-in-right-modal

**Google Chrome** - Version 87.0.4280.66 (Official Build) (64-bit)
**Ubuntu** - 20.10

ghiscoding:
@rohit2sharma95 I see this too, it is actually caused by the scrollbar width and mine is thicker and so my margin is worst than yours. I'm not sure if this is intended, probably not.

Gaslan:
When the right modal opened `modal-open` class is added to body and body's overflow is set to `hidden`. So there is no need to add right-padding to modal because scrollbar is not seen when modal is opened.

vanillajonathan:
The classes are currently named `.modal-left` and `.modal-right`. Should they be named that, or should they be named `.modal-start` and `.modal-end` instead? #32330 #30980

XhmikosR:
Closing this one on favor of #29017.

PR #31297, opened 2020-07-14
V4: switch to xo

XhmikosR:
This makes sure we use the same linting as we do in main. It shouldn't affect the consumers of the JS code AFAICT.

Better review without whitespace changes: https://github.com/twbs/bootstrap/pull/31297/files?w=1


Johann-S (approved):

202006233011:
thanks

PR #31323, opened 2020-07-18
Make width utilities responsive

Luke-zhang-04:
- feat: add `responsive: true` to width utilities

There are many use cases where the width utility needs to be responsive
Responsive sizing can be added with the class w-[breakpoint]-[size]
e.g w-sm-100


mdo:
This is covered by grid classes in v5. We removed their grid styles beyond the `width` so you can freely use them in other situations.

narenkram:
@mdo I need such classes inside a grid, Like when I don't want a element to use it's full space inside a grid. Not having this making it difficult designing responsive for all screens. Take a look at attached images, please add this, 

image
image

PR #31336, opened 2020-07-20
Adding text-ellipsis as an utility

### LINKED ISSUE #31332: '.form-group row' and multi language
---
- Operating system and version: macOS.
- Browser and version: Google Chrome 84.0.4147.89 (64 bits) and Firefox 78.0.2 (64 bits).
- [Reduced test case](https://codepen.io/julien-deramond/pen/MWKZybb)
- [Related issue in Boosted](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/issues/444)
---

In a `.form-group row`, the `<label>`s (most of the time) will be on the left part of the form.

Considering a multi language website, we can't control what text will be in those `<label>`s.

So we can be confronted to cases where the `<label>`s contain words very long in some languages; and we must add something like a `.text-break` class to handle it:

```
.text-break {
  word-wrap: break-word;
}
```

Because it's everywhere in our website, it seems that we have 2 choices:

- Change the global rule of `.form-group row` to automatically add the equivalent of the `.text-break` content.
- Add a `.text-break` everywhere manually.

Wouldn't be possible to automatically add the equivalent of `.text-break` for the `.form-group row`s directly in Bootstrap or is it dumb?

The same thing is happening with `.form-control-plaintext`. Because we're not in control of what the user will enter, why not using automatically something like a `.text-ellipsis`:

```
.text-ellipsis {
  overflow: hidden;
  text-overflow: ellipsis;
}
```

Moreover, it worked in Bootstrap 3 with `.form-control.static` (probably because it wasn't a readonly `<input>`).

[>>> Reduced test case trying to show all of it](https://codepen.io/julien-deramond/pen/MWKZybb)

gauravbehere:
Fixes: #31332

This PR adds ```text-overflow: ellpisis``` as util.

# Demo
image

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -444,6 +444,11 @@ $utilities: map-merge(
       class: text,
       values: (break: break-word)
     ),
+    "text-ellipsis": (
+      property: text-overflow,
+      class: text,
+      values: ellipsis
+    ),
     "font-family": (
       property: font-family,
       class: font,
--- a/site/content/docs/5.0/utilities/text.md
+++ b/site/content/docs/5.0/utilities/text.md
@@ -51,6 +51,14 @@ Prevent long strings of text from breaking your components' layout by using `.te
 <p class="text-break">mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm</p>
 {{< /example >}}
 
+## Text Ellipsis
+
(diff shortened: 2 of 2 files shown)


ffoodd:
Please stop opening and closing issues, you can work on a single PR instead of spamming a public repos where thousands get notified.

gauravbehere:
@ffoodd Yep, my bad. Realised that I had a different username logged in github CLI. Gotcha!

florianlacreuse:
See https://v5.getbootstrap.com/docs/5.0/helpers/text-truncation/

gauravbehere:
@florianlacreuse I think `text-truncate` should suffice the need of the bug #31332

mdo:
Closing per previous comments.

PR #31337, opened 2020-07-20
JS docs formatting tweaks

### LINKED ISSUE #31189: Be consistent with API tables
It seems we are also inconsistent with tables in some components. For example, https://deploy-preview-31130--twbs-bootstrap.netlify.app/docs/5.0/components/tooltips/#methods is using a table but https://deploy-preview-31130--twbs-bootstrap.netlify.app/docs/5.0/components/carousel/#methods doesn't.

The no table approach has the advantage that one can link directly to a method. Either way, we should streamline this.

_Originally posted by @XhmikosR in https://github.com/twbs/bootstrap/pull/31130#issuecomment-647108644_

mdo:
This PR revisits the options, methods, and events sections for each component's JS section. Builds off @rohit2sharma95's suggested changes for the dispose method and takes a more holistic approach to the sections.

- All sections are now Markdown-powered tables
- All tables are using the `bs-table` short code to add our `.table-*` classes as needed
- Does a little bit of wordsmithing, but could use more editing overall

I'm waiting on @Johann-S or someone else to tell me what happens with dispose before calling content final here, but wanted to open the PR for when we're ready.

TODO:

- [x] Convert the remaining HTML tables to Markdown
- [x] Verify that the CSS still works and is adapted
- [x] When we are done converting all tables, see if the `markdown` shortcode is still used

Preview: https://deploy-preview-31337--twbs-bootstrap.netlify.app/

Fixes #31189

### DIFF
--- a/site/assets/scss/_content.scss
+++ b/site/assets/scss/_content.scss
@@ -31,15 +31,12 @@
   }
 
   // Override Bootstrap defaults
-  > .table {
-    max-width: 100%;
+  > .table,
+  > .table-responsive .table {
     margin-bottom: 1.5rem;
     @include font-size(.875rem);
 
     @include media-breakpoint-down(lg) {
-      display: block;
-      overflow-x: auto;
-
       &.table-bordered {
         border: 0;
       }
@@ -61,6 +58,7 @@
     }
 
     // Prevent breaking of code
+    // stylelint-disable-next-line selector-max-compound-selectors
     td:first-child > code {
       white-space: nowrap;
     }
--- a/site/content/docs/5.1/components/alerts.md
+++ b/site/content/docs/5.1/components/alerts.md
@@ -209,53 +209,29 @@ See the [triggers](#triggers) section for more details.
 
 ### Methods
 
-<table class="table">
-  <thead>
-    <tr>
-      <th>Method</th>
-      <th>Description</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <td>
-        <code>close</code>
-      </td>
-      <td>
-        Closes an alert by removing it from the DOM. If the <code>.fade</code> and <code>.show</code> classes are present on the element, the alert will fade out before it is removed.
-      </td>
-    </tr>
-    <tr>
-      <td>
-        <code>dispose</code>
-      </td>
-      <td>
-        Destroys an element's alert. (Removes stored data on the DOM element)
-      </td>
-    </tr>
-    <tr>
-      <td>
-        <code>getInstance</code>
-      </td>
-      <td>
-        Static method which allows you to get the alert instance associated to a DOM element, you can use it like this: <code>bootstrap.Alert.getInstance(alert)</code>
-      </td>
-    </tr>
-    <tr>
-      <td>
-        <code>getOrCreateInstance</code>
-      </td>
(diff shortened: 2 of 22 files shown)


XhmikosR:
We should make the table responsive BTW

mdo:
> And maybe use normal font-size? Not sure what the rationale behind this choice was.

I knocked down the font-size to ideally help with readability. Bump it up in the inspector and see how it feels for you. For me it really helped with reading a _table_ to downsize the font vs the body text. These tables are super content dense as well, and my secondary goal is to show more table contents to make it easier to scan the whole thing.

Johann-S:
@rohit2sharma95 summarized everything here: https://github.com/twbs/bootstrap/pull/30838#issuecomment-661284952
so you can use it @mdo

XhmikosR:
It just feels weird not having the tables responsive though. `font-size` might be objective :)

mdo:
> It just feels weird not having the tables responsive though.

Oh sorry, oversight on my part. I'll revisit and update.

XhmikosR:
After we merge #31803, I can switch to responsive tables in this branch, assuming we no longer use raw HTML tables.

XhmikosR:
@mdo I rebased this and switched to responsive tables too. Needs a thorough review just in case I missed something in the rebase (we should also check the whole docs again for more instances).

Otherwise, I'd say we can proceed with this.

XhmikosR:
@ffoodd what's your take on this? Mostly the CSS stuff.

I say we land this as is and we migrate any missing components later, otherwise we'll have this hanging forever.

ffoodd:
I think it needs some tweaks, e.g. [the content table](https://deploy-preview-31337--twbs-bootstrap.netlify.app/docs/5.1/getting-started/contents/) doesn't overflow properly and makes body overflow.

Not sure what happens for now.

GeoSot:
> I think it needs some tweaks, e.g. [the content table](https://deploy-preview-31337--twbs-bootstrap.netlify.app/docs/5.1/getting-started/contents/) doesn't overflow properly and makes body overflow.

Wasn't wrapped with `.table-responsive`. 

Added some more, and wrapped these that use html customization inside.

site/assets/scss/_content.scss
```diff
@@ -31,15 +31,12 @@
   }
 
   // Override Bootstrap defaults
-  > .table {
-    max-width: 100%;
+  > .table,
+  > .table-responsive .table {
     margin-bottom: 1.5rem;
     @include font-size(.875rem);
```
XhmikosR, line 37: @ffoodd @mdo: perhaps we could increase this a bit? I feel it's too small right now.

mdo: Goes from 16px default font size down to only 14px. I think it works really well for the type of content here.

mdo:
Updated this with some new copy changes to clean up `<br>` usage in a few JS tables and redo the containers and contents tables in Markdown. There are now no more responsive tables in our docs _not_ using the `table` shortcode.

GeoSot (approved):

ffoodd (approved):
Nice work! This'll be really helpful 🚀

XhmikosR:
Let's land this and revisit any remaining instances later :)

PR #31343, opened 2020-07-21
Adding overflow visible and scroll as utilities

### LINKED ISSUE #31302: Dropdown not working inside toast
Dropdowns do not work when inside toasts.

gauravbehere:
- Adding overflow visible & scroll as a utility.

# Demo
image

# How does this fix the issue?
image

Preview: https://deploy-preview-31343--twbs-bootstrap.netlify.app/docs/5.0/utilities/overflow/

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -16,7 +16,7 @@ $utilities: map-merge(
     ),
     "overflow": (
       property: overflow,
-      values: auto hidden,
+      values: auto hidden visible scroll,
     ),
     "display": (
       responsive: true,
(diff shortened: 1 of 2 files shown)


techboyg5:
How about adding `overflow-scroll` as well? That completes the overflow family!

mdo (approved):
Approved for Alpha 3.

XhmikosR:
Maybe we should add margin is some viewports in `bd-example.d-md-flex`? Also, it seems the other examples don't have the margin utilities.

I'd say it needs some tweaking before merging @mdo.

Issue #31378, opened 2020-07-30
Rename "Navs" in documentation to "Navs and tabs" or create separate "Dynamic tabs" document


patrickhlauke:
When a dev is looking specifically for "tabs" (as in dynamic tabs/tab panels), it's not obvious in the documentation that these are hiding under "Navs". Wonder if v5 gives us the opportunity to rename this page in the docs.

Admittedly, this may seem odd to veteran Bootstrap devs, since the term "tabs" isn't used to exclusively mean those actual dynamically changing tabs/tabpanel, but as a visual style, same as "pills".  A possible alternative/way around this could be to split out the whole "JavaScript behavior" part into its own separate documentation page for "Dynamic tabs"?

mdo:
Given our search is easily accessible and I haven't heard much feedback on this in the past, I feel like we can leave it as-is. Down to let this sit though and gather some feedback.

patrickhlauke:
anecdotally...even i sometimes have to think twice about "where was the tabs stuff again?"

other frameworks (just grabbing some random ones) have separate pages for tabs:

* https://materializecss.com/tabs.html
* https://jqueryui.com/tabs/
* https://get.foundation/sites/docs/tabs.html
* https://getuikit.com/docs/tab
* https://semantic-ui.com/modules/tab.html

ffoodd:
Someone at Orange asked me "where are the tabs, again" yesterday. So at least making them more obvious could help 🙂

mdo:
Happy to see this renamed. Maybe we tackle in Alpha 3?

PRO-BROS:
Hi!

you guys can check mine nav in my website :- [See Here](https://prhocoding.com/)

PR #31379, opened 2020-07-30
Renamed "Navs" in documentation to "Navs and tabs"

psjishnu:
Fixed issue #31378


patrickhlauke:
this will depend on whether we decide to rename, or split out. and it will need extra work like setting up aliases (possibly) and further changes in other files. thanks though

PR #31382, opened 2020-07-31
Update copy around breakpoint sizing

### LINKED ISSUE #31288: v5 breakpoints divisible by 12
I see this line in the v5 (and v4) docs:
`Each breakpoint size was chosen to be a multiple of 12 and to be representative of a subset of common device sizes and viewport dimension`

I have run into fractional pixel issues myself when stuff isn't divisible by 12 (column count), so I get the reasoning.

However, I am noticing that the new xxl breakpoint is 1400px, NOT divisible by 12. Just figured the explanation should be updated, or the default breakpoint should be changed to something like 1440px

mdo:
Fixes #31288

### DIFF
--- a/site/content/docs/5.0/layout/breakpoints.md
+++ b/site/content/docs/5.0/layout/breakpoints.md
@@ -61,7 +61,7 @@ Bootstrap includes six default breakpoints, sometimes referred to as _grid tiers
   </tbody>
 </table>
 
-Each breakpoint size was chosen to be a multiple of 12 and to be representative of a subset of common device sizes and viewport dimensions. They don't specifically target every use case or device, but the provide ranges provide a strong and consistent foundation to build on for any nearly device.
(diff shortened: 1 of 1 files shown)


site/content/docs/5.0/layout/breakpoints.md
```diff
@@ -61,7 +61,7 @@ Bootstrap includes six default breakpoints, sometimes referred to as _grid tiers
   </tbody>
 </table>
 
-Each breakpoint size was chosen to be a multiple of 12 and to be representative of a subset of common device sizes and viewport dimensions. They don't specifically target every use case or device, but the provide ranges provide a strong and consistent foundation to build on for any nearly device.
+Each breakpoint size was chosen to comfortably hold containers whose widths are multiples of 12. They're also aimed at be representative of a subset of common device sizes and viewport dimensions. They don't specifically target every use case or device, but the provide ranges provide a strong and consistent foundation to build on for any nearly device.
```
alecpl, line 64: "but the provide ranges provide" -> "but they provide"
"for any nearly device" -> "for nearly any device"

XhmikosR (approved):

Issue #31269, opened 2020-07-09
Changed font-size-base distorts radio and check boxes (v5 alpha1)


iMattPro:
MacOS High Sierra
Safari/Chrome

I changed the SCSS variable `$font-size-base` to be `0.875rem`

With this new font size base, form radio and check boxes are distorted, they are stretched horizontally.

![Screen Shot 2020-07-09 at 1 09 02 PM](https://user-images.githubusercontent.com/303711/87085856-6f779180-c1e5-11ea-9753-f0efc462a5de.png)
![Screen Shot 2020-07-09 at 1 09 10 PM](https://user-images.githubusercontent.com/303711/87085860-71415500-c1e5-11ea-82cd-dcda19d9fb3d.png)

https://codepen.io/matt_codez/pen/zYrjQWG

alecpl:
Firefox and Chrome shows the size is 17.5 x 17.5. However, the dot/check inside is indeed not in the middle in Chrome. No issue in Firefox for me.

iMattPro:
I see this problem in all my browsers:
![image](https://user-images.githubusercontent.com/303711/87185497-b5446080-c29e-11ea-8121-9c0a5db6a763.jpg)

mdo:
Might need to revisit the sizing of checks and radios. `1.25em` leads to some rough math and uneven pixels when paired with `.875rem` root font size. I recall some feedback about them feeling a bit too large previously... @twbs/css-review thoughts on shrinking things?

mdo:
Closed by #31383.

PR #31383, opened 2020-07-31
Adjust sizing of checks and radios

mdo:
Moves from 1.25em to 1em to better support custom font-size-base values. Most common would probably be .875rem (14px) and current sizing garbles that into 1.09-ish, thus causing the rendering issues in #31269.

https://deploy-preview-31383--twbs-bootstrap.netlify.app/docs/5.0/forms/checks-radios/


MartijnCuppens (approved):
I like

PR #31384, opened 2020-07-31
Add chat

techboyg5:
I think that adding a chat feature to the docs pages would be a good idea. This pull request adds [Gitter](https://gitter.im/twbs/bootstrap) to such pages, allowing this to happen.

I think that this would be a good idea because GitHub issues are mainly for feature requests or bug reports, not really for general talk.

### DIFF
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -11,7 +11,7 @@
   </div>
 </footer>
 <!-- Begin chat by Gitter -->
-<button class="chat-button btn btn-secondary" style="position:fixed;bottom:15px;right:15px">Open Chat</button>
+<button class="chat-button btn btn-secondary" type="button" style="position:fixed;bottom:15px;right:15px;z-index:999">Open Chat</button>
 <script>
   ((window.gitter = {}).chat = {}).options = {
     room: 'twbs/bootstrap',
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -11,7 +11,7 @@
   </div>
 </footer>
 <!-- Begin chat by Gitter -->
-<button class="chat-button btn btn-secondary" type="button" style="position:fixed;bottom:15px;right:15px;z-index:999">Open Chat</button>
+<button class="chat-button btn btn-secondary" type="button">Open Chat</button>
 <script>
   ((window.gitter = {}).chat = {}).options = {
     room: 'twbs/bootstrap',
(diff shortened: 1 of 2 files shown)


ffoodd:
There's [a Slack](https://bootstrap-slack.herokuapp.com/), GitHub discussions, and Stack Overflow.

Is there any team member in this Gitter, already? I didn't know about it :/

techboyg5:
> There's [a Slack](https://bootstrap-slack.herokuapp.com/), GitHub discussions, and Stack Overflow.
> 
> Is there any team member in this Gitter, already? I didn't know about it :/

This allows people to chat without leaving the docs, making it easier to reference information on the docs. Feel free to use another system instead :)

patrickhlauke:
Personally would find that pointless, but hey...

techboyg5:
> Personally would find that pointless, but hey...



> Personally would find that pointless, but hey...

GitHub discussions and Stack Overflow have no embedding feature.

ffoodd:
Using browser's tabs isn't that bad, is it?

Your PR also:
1. use a `button` without `type` attribute and inline styles — on Bootstrap docs, that's pretty messy;
2. callls a 7k script (one third of Bootstrap minified bundle!)
3. which probably adds an `iframe` on which we'd have zero control, thus adding both accessibility and security issues on a pretty well tailored site.

I'm pretty against adding this too, but hey… Maybe someone will find good arguments to work on this?

techboyg5:
> Using browser's tabs isn't that bad, is it?
> 
> Your PR also:
> 
> 1. use a `button` without `type` attribute and inline styles — on Bootstrap docs, that's pretty messy;
> 2. callls a 7k script (one third of Bootstrap minified bundle!)
> 3. which probably adds an `iframe` on which we'd have zero control, thus adding both accessibility and security issues on a pretty well tailored site.
> 
> I'm pretty against adding this too, but hey… Maybe someone will find good arguments to work on this?

I fixed some of those.

XhmikosR:
Thanks but we are going to pass on this.

techboyg5:
> Thanks but we are going to pass on this.

After I made my changes, are there any other problems or issues?

XhmikosR:
It makes no point adding gitter when none of the team members are present. And I definitely don't like adding 3rd-part scripts to our docs unless we really have to.

techboyg5:
> It makes no point adding gitter when none of the team members are present. And I definitely don't like adding 3rd-party scripts to our docs unless we really have to.

OK, thank you! I won't disturb you anymore about adding chat.

Issue #31391, opened 2020-08-01
a:not([href]) styles too strong/override .btn-* color definition


techboyg5:
The button text at https://getbootstrap.com/docs/4.5/components/popovers/#dismiss-on-next-click should be white instead of black.

patrickhlauke:
https://codepen.io/patrickhlauke/pen/gOPVNOW

techboyg5:
> https://codepen.io/patrickhlauke/pen/gOPVNOW

Dismiss on next click only works with `<a>` elements. Read the docs please.

patrickhlauke:
> Read the docs please.

lose your attitude please. i've just rewritten your issue title to actually hone in on the problem, and added that as a demonstration.

ffoodd:
@patrickhlauke @techboyg5 The popover example works fine now, it's been fixed in #30830 (and this issue duplicates #30531).

We assume that in Bootstrap, links without targets imlement one of our components—thus should have a `class`.

PR #31399, opened 2020-08-02
Docs: Use proper <button> instead of faked <a> element for popover example

patrickhlauke:
sidesteps #31391 but on principle should really use `<button>` here


patrickhlauke:
only browser that currently seems to have a problem with this, from testing, is Safari. investigating further what its beef is...

patrickhlauke:
ok, root of the problem seems to be the fact that Safari doesn't actually focus elements like `<button>` when you click them with the mouse. which is ... not ideal https://twitter.com/patrick_h_lauke/status/1289988990146211840

and Firefox/Mac copied that silly behavior too. good good. considering adding some basic additional behavior that also sets `focus()` when clicking these controls...

patrickhlauke:
will likely keep this PR parked, do a separate PR for a potential cross-browser fix (or, well, a patch for Safari and Firefox on Mac), and hopefully this can then get merged when that fix is merged

XhmikosR:
@patrickhlauke do we close this?

patrickhlauke:
yeah as i get the feeling Apple are in no rush to change their behaviour in Safari (and Firefox will just copy what Safari does on macOS for "platform consistency"), and since I doubt we want to hack away at adding extra scripting that explicitly focuses the button after a click ... i think i'll just abandon this. annoying.

PR #31438, opened 2020-08-05
Restore and deprecate make-container-max-widths mixin

### LINKED ISSUE #31436: 4.5.1 Breaking change, removed Mixing make-container-max-widths
### Version 4.5.1 made a breaking change removing a mixing 

The mixin is called: `make-container-max-widths`

ERROR LOG:

>>ERROR in ./src/styles.scss (./node_modules/css-loader/dist/cjs.js??ref--13-1!./node_modules/postcss-loader/src??embedded!./node_modules/sass-loader/dist/cjs.js??ref--13-3!./src/styles.scss)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: no mixin named make-container-max-widths
        on line 13 of node_modules/@my-components/styles/grid/_grid-container.scss
        from line 1 of C:\~\src\styles.scss
  @include make-container-max-widths();
 >>  -----------^

### OS/environment:
Windows, Angular 9, SCSS/SASS, bootstrap, node-sass

How to reproduce:

`.your-container {
    @include make-container-max-widths();
}`

### stackblitz example
https://stackblitz.com/edit/angular-ivy-2t6deq?file=src%2Fapp%2Fapp.component.scss

### LINKED ISSUE #31440: Deprecate make-container-max-widths mixin
After #31438 merges, let's add a deprecation notice to this mixin for v4.5.3.

mdo:
We removed this to de-dupe our grid containers in our compiled CSS (see #30969) and I wasn't thinking when reviewing that it'd be a breaking change. We're sticking with not using the mixin, but adding the mixin back for others to use.

We may need to add a deprecation notice to it as well.

Fixes #31436, fixes #31440

### DIFF
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -25,6 +25,7 @@
       max-width: $container-max-width;
     }
   }
+  @include deprecate("The `make-container-max-widths` mixin", "v4.5.2", "v5", $ignore-warning);
 }
 
 @mixin make-col-ready($gutter: $grid-gutter-width) {
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -25,7 +25,7 @@
       max-width: $container-max-width;
     }
   }
-  @include deprecate("The `make-container-max-widths` mixin", "v4.5.2", "v5", $ignore-warning);
+  @include deprecate("The `make-container-max-widths` mixin", "v4.5.2", "v5");
 }
 
 @mixin make-col-ready($gutter: $grid-gutter-width) {
--- a/scss/mixins/_grid.scss
+++ b/scss/mixins/_grid.scss
@@ -18,6 +18,15 @@
   margin-left: -$gutter / 2;
 }
 
+// For each breakpoint, define the maximum width of the container in a media query
+@mixin make-container-max-widths($max-widths: $container-max-widths, $breakpoints: $grid-breakpoints) {
+  @each $breakpoint, $container-max-width in $max-widths {
+    @include media-breakpoint-up($breakpoint, $breakpoints) {
+      max-width: $container-max-width;
+    }
+  }
+}
+
 @mixin make-col-ready($gutter: $grid-gutter-width) {
   position: relative;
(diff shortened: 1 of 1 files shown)


grumbaut:
This is currently a blocker on a rather urgent project. Any chance it will be merged soon? If not, I'll just require 4.5.0 exactly for now.

mdo:
Hoping to do a v4.5.2 tomorrow to patch this. Doing some digging and testing other things out. Lock in at v4.5.0 exactly for now to keep yourself unblocked if you can.

XhmikosR:
@mdo Do we need the mixin in main, though? We could only target v4-dev where this was a breaking change, right?

mdo:
Yup, we can retarget this one.

XhmikosR:
Let's add the deprecation warning too so that we are covered since the mixin is removed from v5. @twbs/css-review in case you can help us here to move faster.

scss/mixins/_grid.scss
```diff
@@ -18,6 +18,16 @@
...
 }
 
+// For each breakpoint, define the maximum width of the container in a media query
+@mixin make-container-max-widths($max-widths: $container-max-widths, $breakpoints: $grid-breakpoints) {
+  @each $breakpoint, $container-max-width in $max-widths {
+    @include media-breakpoint-up($breakpoint, $breakpoints) {
+      max-width: $container-max-width;
+    }
+  }
+  @include deprecate("The `make-container-max-widths` mixin", "v4.5.2", "v5", $ignore-warning);
```
MartijnCuppens, line 28: `$ignore-warning` is undefined, thus can be removed:

```suggestion
  @include deprecate("The `make-container-max-widths` mixin", "v4.5.2", "v5");
```

XhmikosR: True, bad CP.

MartijnCuppens (approved):

nh2:
Is there any documentation or an upgrade note what one should use instead of `make-container-max-widths`?

rchulyada:
would be good if there are list of depreciated and its replacement. Really annoying docs :(

PR #31443, opened 2020-08-06
Support for non-list based structure of the nav used with tabs plugin

### LINKED ISSUE #31166: [BS4.5] Button toolbar not working with Navs
Maybe this is not a real bug report but a feature request.

I'm trying to use [Bootstrap Navs with JavaScript behaviour](https://getbootstrap.com/docs/4.5/components/navs/#javascript-behavior) but instead of the tabs and pill I would like to have button toolbar.

Now with a simple toolbar this works:

```
  <div class="nav btn-toolbar" role="tablist">
    <a class="nav-link btn btn-secondary active" role="button" data-toggle="tab" href="#home-tab">Home</a>
    <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#profile-tab">Profile</a>
    <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#contact-tab">Contact</a>
  </div>
  
  <div class="tab-content">
    <div class="tab-pane fade show active" id="home-tab" role="tabpanel">
      Home...
    </div>
    <div class="tab-pane fade" id="profile-tab" role="tabpanel">
      Profile...
    </div>
    <div class="tab-pane fade" id="contact-tab" role="tabpanel">
      Contact...
    </div>
  </div>
```

I'm applying the `nav` and `nav-link` classes and the tab behaviour kicks in.

Now I wanted to go a but further with a grouped button toolbar like:

```
  <div class="nav btn-toolbar mb-3" role="tablist">
    <div class="btn-group mr-2" role="group">
      <a class="nav-link btn btn-secondary active" role="button" data-toggle="tab" href="#home-tab">Home</a>
      <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#profile-tab">Profile</a>
    </div>
    <div class="btn-group" role="group">
      <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#contact-tab">Contact</a>
    </div>
  </div>
```

The first time I navigate to each tab (using the toolbar buttons) works, but as soon as I want to navigate to a tab the second time it doesn't work.

The reason it does show the tab again is that the navlink anchor in the toolbar was still `.active`.  The `.active` class was never removed as it is only looking for children under the `.nav` div:

https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tab.js#L137-L140

```
  _activate(element, container, callback) {
    const activeElements = container && (container.nodeName === 'UL' || container.nodeName === 'OL')
      ? $(container).find(SELECTOR_ACTIVE_UL)
      : $(container).children(SELECTOR_ACTIVE)

    ..

    if (active && isTransitioning) {
      const transitionDuration = Util.getTransitionDurationFromElement(active)

      $(active)
        .removeClass(CLASS_NAME_SHOW)
        .one(Util.TRANSITION_END, complete)
        .emulateTransitionEnd(transitionDuration)
    } else {
      complete()
    }

```

Now I wonder if it would as simple to change:

`$(container).children(SELECTOR_ACTIVE)`

to:

`$(container).find(SELECTOR_ACTIVE)`

I think it would be nice to be able to navigate tabbed tabbed content with other Bootstrap controls like this.

Bug reports must include:

- Operating system and version: macOS 10.15.5
- Browser and version: Chrome 83.0.4103.116
- Test case: https://jsfiddle.net/marceloverdijk/5eznyvua/

### LINKED ISSUE #31392: v5 - tabs required classes
https://v5.getbootstrap.com/docs/5.0/components/list-group/
https://v5.getbootstrap.com/docs/5.0/components/navs/

https://github.com/twbs/bootstrap/blob/main/js/src/tab.js#L45

Why must the parent of the HTML element of the tab contain'.nav, .list-group'? If the parent of the HTML format does not contain one of these two classes, an error will be reported.

Moreover, in the document description, there is no key description, and one of these two categories must be included.

rohit2sharma95:
Find the element with `.active` class directly to support the non-list based structure (Button toolbar) of the nav used with the tabs plugin.

Closes #31166 and also closes #31392


XhmikosR:
@rohit2sharma95 feel free to resubmit the PR if it's still valid. Don't take it that we ignore you, it's just that there are periods of time we might not be very responsive.

Regardless, I appreciate taking the time to submit a PR, ❤

XhmikosR:
@mdo I'm not sure if we have any limitations coming from the CSS selectors and assuming tests pass we should be OK.

@rohit2sharma95 can you add a test case with the new accepted markup too, which would previously fail?

rohit2sharma95:
Thanks, @XhmikosR :+1:
I will add the required test cases.

rohit2sharma95:
@XhmikosR and @Johann-S 
While I was writing the test case I realize that the markup in this issue #31166 is rare. If they are using `.btn-toolbar` then I think the `.nav` class should not be required. They are using:

```html
<div class="nav btn-toolbar mb-3" role="tablist">
  <div class="btn-group mr-2" role="group">
    <a class="nav-link btn btn-secondary active" role="button" data-toggle="tab" href="#home-tab">Home</a>
    <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#profile-tab">Profile</a>
  </div>
  <div class="btn-group" role="group">
    <a class="nav-link btn btn-secondary" role="button" data-toggle="tab" href="#contact-tab">Contact</a>
  </div>
</div>
```

It is not mentioned in the docs but you need to put either `.nav` or `.list-group` class on the parent element to make the tabs working. And in my opinion, if bootstrap is going to support the non-list based structure for the navs then the requirement of the `.nav` or `.list-group` should be omitted because the above markup can be written as below and these both classes are not really required :thinking: 

```html
<div class="btn-toolbar mb-3" role="tablist">
  <div class="btn-group mr-2" role="group">
    <a class="btn btn-secondary active" role="button" data-toggle="tab" href="#home-tab">Home</a>
    <a class="btn btn-secondary" role="button" data-toggle="tab" href="#profile-tab">Profile</a>
  </div>
  <div class="btn-group" role="group">
    <a class="btn btn-secondary" role="button" data-toggle="tab" href="#contact-tab">Contact</a>
  </div>
</div>
```

There is already an issue about it #31392 
It is just my opinion :smile:

Johann-S:
@rohit2sharma95 

If you can submit a patch which required no Bootstrap classes do not hesitate it would be a good move 👌

js/src/tab.js
```diff
@@ -43,9 +43,8 @@ const CLASS_NAME_FADE = 'fade'
 const CLASS_NAME_SHOW = 'show'
 
 const SELECTOR_DROPDOWN = '.dropdown'
-const SELECTOR_NAV_LIST_GROUP = '.nav, .list-group'
+const SELECTOR_NAV_PARENT = '.nav, .list-group, .btn-toolbar'
 const SELECTOR_ACTIVE = '.active'
-const SELECTOR_ACTIVE_UL = ':scope > li > .active'
 const SELECTOR_DATA_TOGGLE = '[data-toggle="tab"], [data-toggle="pill"], [data-toggle="list"]'
```
rohit2sharma95, line 48: I think there may be some unique selector here instead of these three 🤔
Ref: #31976
@XhmikosR @Johann-S

js/src/tab.js
```diff
@@ -43,9 +43,8 @@ const CLASS_NAME_FADE = 'fade'
 const CLASS_NAME_SHOW = 'show'
 
 const SELECTOR_DROPDOWN = '.dropdown'
-const SELECTOR_NAV_LIST_GROUP = '.nav, .list-group'
+const SELECTOR_NAV_PARENT = '.nav, .list-group, .btn-toolbar'
 const SELECTOR_ACTIVE = '.active'
```
rohit2sharma95, line 47: I added one more class `.btn-toolbar` here to give the support of button toolbar in navs. But I was thinking that there may be some unique selector here as well that selects the parent of the navs in every case. Like we can use `data-toggle="something"`. I am not sure but can you please suggest @XhmikosR, if it is a good idea. 🙂

patrickhlauke:
would be good to check this again, as it's been lingering in limbo for a while. @twbs/js-review

mdo:
Adding to v5.2.0—would be rad to get some fresh eyes on this again :).

mdo:
Closing as stale, and also other markup is supported here already (per our docs right now).

PR #31455, opened 2020-08-09
Update current_ruby_version to 5.0.0.alpha1

glebm:


XhmikosR:
@glebm that would be an issue because we use the same format everywhere else. Can't you publish it with a hyphen? Otherwise I'll need to remember to bump it manually which increases greatly the chances of making a mistake.

glebm:
The hyphen is not a valid character for a rubygem version IIRC

XhmikosR:
Well, that's too bad. We always used a hyphen https://www.npmjs.com/package/bootstrap?activeTab=versions

I'll think about it, but worst case I guess we bump it separately in case I miss it.

XhmikosR (approved):

PR #31478, opened 2020-08-14
Improve disabled state close button

dejakob:
Add 
`not-allowed: cursor`
and
`user-select none`
to the close button in order to give the user feedback it is disabled


patrickhlauke:
we abandoned the `not-allowed` cursor quite a few versions back because it's non-standard and, subjectively, very disruptively ugly on windows.

mdo:
@patrickhlauke Thoughts on the `user-select: none` addition?

patrickhlauke:
the `user-select:none` part is good...sorry, should have mentioned that at the time

mdo:
Closing for ##31641.

PR #31480, opened 2020-08-14
Focus state input group

dejakob:
IMO on an input-group, the focus state should cover the entire group

Before
Screenshot 2020-08-14 at 19 57 19

After
Screenshot 2020-08-14 at 19 55 56


mdo:
Going to leave this as-is because input groups can have multiple form fields.

PR #31484, opened 2020-08-16
Add border width utility

### LINKED ISSUE #31299: Provide way to specify size to border
Currently we have **border**, border-primary, border-success etc all of them seem to only have a 1 pixel thick border with no ability to make the border thicker/wider. The border with 1 pixel gets lost when you have padding, margins or shadow.

Would it be possible to add a **border-size** attribute? 

We can either have 
- pixel count specified like border-1, border-2, border-3 (a finite list of choices) OR 
- named sizes like we have in **shadow**  such as border-none border border-sm, border-lg

MauricioHernanCabrera:
I added border-width utility. it was a request https://github.com/twbs/bootstrap/issues/31299

Example
```html
<div class="border border-primary border-width-5">Hello world</div>

<div class="border border-primary border-width-2">Hello world</div>
```

Fixes #31299

Demo: https://deploy-preview-31484--twbs-bootstrap.netlify.app/docs/5.0/utilities/borders/#border-width


MauricioHernanCabrera:
I have added the border-width utility and a variable for handle the sizes

MartijnCuppens:
I think we can leave the `width` out here (eg. `border-2`). Also, this needs documentation.

MauricioHernanCabrera:
> I think we can leave the `width` out here (eg. `border-2`). Also, this needs documentation.

Okay, What do you mean by documentation?

MartijnCuppens:
> Okay, What do you mean by documentation?

Explain how these utilities need to be used on our documentation site. You'll probably need to edit this file:

https://github.com/twbs/bootstrap/blob/main/site/content/docs/5.0/utilities/borders.md

MauricioHernanCabrera:
> > Okay, What do you mean by documentation?
> 
> Explain how these utilities need to be used on our documentation site. You'll probably need to edit this file:
> 
> https://github.com/twbs/bootstrap/blob/main/site/content/docs/5.0/utilities/borders.md

Thanksss. I am going to work on the docs

MauricioHernanCabrera:
@MartijnCuppens can I do some more por bootstrap? I want to contribute more to this project

site/content/docs/5.0/utilities/borders.md
```diff
@@ -19,7 +19,7 @@ Use border utilities to add or remove an element's borders. Choose from all bord
 <span class="border-bottom"></span>
 <span class="border-left"></span>
 {{< /example >}}
-
+
```
MartijnCuppens, line 22: This space should be removed.

```suggestion
```

MauricioHernanCabrera: okasss

MauricioHernanCabrera: done!

calvinkarpenko:
Would it not be better to have these in rem units rather than px units?

MartijnCuppens:
> Would it not be better to have these in rem units rather than px units?

We don't do that for borders. Browsers will always render them in px, so if the default browser font size is changed, the borders could be blurred because they are for example 1.25px.

@MauricioHernanCabrera, could you rebase this branch, it contains lots of unrelated changes?

MauricioHernanCabrera:
> > Would it not be better to have these in rem units rather than px units?
> 
> We don't do that for borders. Browsers will always render them in px, so if the default browser font size is changed, the borders could be blurred because they are for example 1.25px.
> 
> @MauricioHernanCabrera, could you rebase this branch, it contains lots of unrelated changes?

done!

MartijnCuppens (dismissed):
LGTM. ping @twbs/css-review for second opinion

XhmikosR:
@mdo WDYT?

mdo:
Can we get a mention on the `migration.md` page? Otherwise LGTM!

mdo (approved):

XhmikosR:
I'm not sure the example is right, we should probably add margin right in the examples?

MauricioHernanCabrera:
> I'm not sure the example is right, we should probably add margin right in the examples?

okasss, I'm going to work on that

site/content/docs/5.0/utilities/borders.md
```diff
@@ -43,6 +43,16 @@ Change the border color using utilities built on our theme colors.
 <span class="border border-white"></span>
 {{< /example >}}
 
+### Border-width
+
+{{< example >}}
```
MartijnCuppens, line 48: @MauricioHernanCabrera, you missed this class here (this used to be ok):
```suggestion
{{< example class="bd-example-border-utils" >}}
```

XhmikosR: Ah, yeah I accidentally removed that yesterday

PR #31485, opened 2020-08-16
Utility - Width fit, min and max

### LINKED ISSUE #31289: Feature Request - New Width Classes
`width: fit-content` -> `w-fit` would be great

Other options:

`width: min-content` -> `w-min`
`width: max-content` -> `w-max`

Could add the same for height, but not sure it's as useful.

MauricioHernanCabrera:
I have added a new classes of width property. Fixes #31289

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -88,7 +88,10 @@ $utilities: map-merge(
         50: 50%,
         75: 75%,
         100: 100%,
-        auto: auto
+        auto: auto,
+        fit: fit-content,
+        min: min-content,
+        max: max-content,
       )
     ),
     "max-width": (


ffoodd:
Not quite sure about this, even if it looks great:
1. naming using `fit`, `min` and `max` can be misunderstood: I'd suggest to use `fit-content`, `min-content` and `max-content`;
2. if we add them to `width`, shouldn't we do the same for `min-width` and `max-width`.
3. what would be the use cases? The related issue doesn't provide any.

ffoodd (changes requested):
This should at least appear in the docs, alongside existing `width` utilities.

XhmikosR:
Closing in favor of #31943

PR #31487, opened 2020-08-16
utility - inset and position sizes variable

### LINKED ISSUE #31015: css location support absolute and fixed
Hello,

I see there is currently a class called "position-absolute" but there are no classes like: "top: 0", "left: 0", "right: 0" ... I think there should be support classes Support for "position-absolute" and "position-fixed"

For example:
.top-0
.top-25
.top-50

The same goes for bottom, left, and right

Thanks

MauricioHernanCabrera:
I have added inset utility and position sizes variable. 

Demo https://deploy-preview-31487--twbs-bootstrap.netlify.app/docs/5.0/utilities/inset/
Fixes https://github.com/twbs/bootstrap/issues/31015

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -501,33 +501,33 @@ $utilities: map-merge(
     "inset": (
       property: top bottom right left,
       class: inset,
-      values: (0: 0, auto: auto)
+      values: (0: 0)
     ),
     "inset-y": (
       property: top bottom,
       class: inset-y,
-      values: (0: 0, auto: auto)
+      values: (0: 0)
     ),
     "inset-x": (
       property: right left,
       class: inset-x,
-      values: (0: 0, auto: auto)
+      values: (0: 0)
     ),
     "top": (
       property: top,
-      values: map-merge($position-sizes, (0: 0, auto: auto))
+      values: map-merge($position-sizes, (0: 0))
     ),
     "bottom": (
       property: bottom,
-      values: map-merge($position-sizes, (0: 0, auto: auto))
+      values: map-merge($position-sizes, (0: 0))
     ),
     "right": (
       property: right,
-      values: map-merge($position-sizes, (0: 0, auto: auto))
+      values: map-merge($position-sizes, (0: 0))
     ),
     "left": (
       property: left,
-      values: map-merge($position-sizes, (0: 0, auto: auto))
+      values: map-merge($position-sizes, (0: 0))
     ),
   ),
   $utilities
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -497,7 +497,38 @@ $utilities: map-merge(
         visible: visible,
         invisible: hidden,
       )
-    )
+    ),
+    "inset": (
+      property: top bottom right left,
+      class: inset,
+      values: (0: 0, auto: auto)
+    ),
+    "inset-y": (
+      property: top bottom,
+      class: inset-y,
+      values: (0: 0, auto: auto)
+    ),
+    "inset-x": (
+      property: right left,
+      class: inset-x,
+      values: (0: 0, auto: auto)
+    ),
+    "top": (
+      property: top,
+      values: map-merge($position-sizes, (0: 0, auto: auto))
+    ),
+    "bottom": (
+      property: bottom,
+      values: map-merge($position-sizes, (0: 0, auto: auto))
+    ),
+    "right": (
+      property: right,
+      values: map-merge($position-sizes, (0: 0, auto: auto))
+    ),
+    "left": (
+      property: left,
+      values: map-merge($position-sizes, (0: 0, auto: auto))
+    ),
   ),
   $utilities
 );
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -1267,3 +1267,7 @@ $kbd-color:                         $white !default;
 $kbd-bg:                            $gray-900 !default;
 
 $pre-color:                         null !default;
+
+// Position Sizes
+
+$position-sizes: () !default;
--- a/site/assets/scss/_component-examples.scss
+++ b/site/assets/scss/_component-examples.scss
@@ -217,6 +217,42 @@
   }
 }
 
+.bd-example-inset-utils {
+  display: flex;
+  flex-wrap: wrap;
+
+  span {
+    position: absolute;
+    display: block;
+  }
+
+  div {
+    position: relative;
+    width: 64px;
+    height: 64px;
+    margin-right: 12px;
+    margin-bottom: 12px;
+
+    &:nth-child(1) span,
+    &:nth-child(3) span {
+      height: 32px;
+    }
+
+    &:nth-child(2) span,
+    &:nth-child(4) span {
+      width: 32px;
+    }
+
+    &:nth-child(6) span,
+    &:nth-child(7) span,
+    &:nth-child(8) span,
+    &:nth-child(9) span {
+      width: 32px;
+      height: 32px;
+    }
+  }
+}
+
 //
 // Code snippets
 //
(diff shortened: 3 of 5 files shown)


ffoodd:
Nicely done, thanks :) 
I'm not sure about the `auto` value, since it's mostly used to reset another value (thus rarelyusable as a utility in markup, I guess).

I'd like other feedbacks before approving, but it looks great!

MauricioHernanCabrera:
> Nicely done, thanks :)
> I'm not sure about the `auto` value, since it's mostly used to reset another value (thus rarelyusable as a utility in markup, I guess).
> 
> I'd like other feedbacks before approving, but it looks great!

thanksss :)

scss/_utilities.scss
```diff
@@ -497,7 +497,38 @@ $utilities: map-merge(
         visible: visible,
         invisible: hidden,
       )
-    )
+    ),
+    "inset": (
+      property: top bottom right left,
+      class: inset,
+      values: (0: 0, auto: auto)
```
ffoodd, line 504: Would you mind dropping `auto`? As I said, I think it's useless as a utility.

scss/_variables.scss
```diff
@@ -1267,3 +1267,7 @@ $kbd-color:                         $white !default;
 $kbd-bg:                            $gray-900 !default;
 
 $pre-color:                         null !default;
+
+// Position Sizes
+
+$position-sizes: () !default;
```
ffoodd, line 1273: We'll need documentation for those new utilities and this new Sass map, too.

ffoodd (changes requested):

scss/_variables.scss
MauricioHernanCabrera: I'm going to work on that

site/assets/scss/_component-examples.scss
```diff
@@ -217,6 +217,42 @@
...
+  display: flex;
+  flex-wrap: wrap;
+
+  span {
+    position: absolute;
+    display: block;
+  }
+
+  div {
+    position: relative;
```
ffoodd, line 230: It'd be better to use our `position` utilities in those examples, since they're likely to be used alongside `inset` ones.

site/content/docs/5.0/utilities/inset.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Inset
+
+Use inset utilities to position elements.
+
+{{< example class="bd-example-inset-utils">}}
+<div class="bg-light border">
+  <span class="inset-x-0 top-0 bg-primary"></span>
+</div>
+
```
ffoodd, line 17: Drop line breaks here, try to keep things short and concise for examples :)

site/assets/scss/_component-examples.scss
```diff
@@ -217,6 +217,42 @@
...
+    &:nth-child(4) span {
+      width: 32px;
+    }
+
+    &:nth-child(6) span,
+    &:nth-child(7) span,
+    &:nth-child(8) span,
+    &:nth-child(9) span {
+      width: 32px;
+      height: 32px;
```
ffoodd, line 251: Use `rem` for those examples, it'd be more responsive if a user changes his base font size in his browser settings.

ffoodd (changes requested):

site/assets/scss/_component-examples.scss
MauricioHernanCabrera: done!

site/assets/scss/_component-examples.scss
MauricioHernanCabrera: done!

ffoodd: You need to remove the declaration here, since you used the utilities.

site/assets/scss/_component-examples.scss
ffoodd: Did you push these changes?

ffoodd (changes requested):

site/assets/scss/_component-examples.scss
MauricioHernanCabrera: Sorry my bad. I forget send my changes

site/assets/scss/_component-examples.scss
MauricioHernanCabrera: Sorry my bad. I forget send my changes

MauricioHernanCabrera: done!

site/content/docs/5.0/utilities/inset.md
```diff
@@ -0,0 +1,41 @@
...
+description: Use inset utilities to quickly style position of an element.
+group: utilities
+toc: true
+---
+
+## Inset
+
+Use inset utilities to position elements. You can combine it with position and display utilities
+
+{{< example class="bd-example-inset-utils">}}
```
ffoodd, line 13: Not mandatory at all but by using `justify-content-between` utility here you might drop you `margin`s on `div`. :)

site/assets/scss/_component-examples.scss
```diff
@@ -221,35 +221,16 @@
...
-    &:nth-child(6) span,
-    &:nth-child(7) span,
-    &:nth-child(8) span,
-    &:nth-child(9) span {
-      width: 32px;
-      height: 32px;
-    }
+  span {
+    min-width: 2rem;
+    min-height: 2rem;
```
ffoodd, line 233: Very nice move!

MauricioHernanCabrera: thankss @ffoodd!

site/content/docs/5.0/utilities/inset.md
MauricioHernanCabrera: Okasss, I can do that

MartijnCuppens:
- In general I'm a fan of using utilities as much as possible, but since we need to explain only the inset classes, the other utilities might be a little overwhelming:
image
An alternative approach could be moving a part of the styles to the CSS side, just to make it a little clearer (see [border documentation](https://deploy-preview-31487--twbs-bootstrap.netlify.app/docs/5.0/utilities/borders/#additive)):
image

- Should we abbreviate the `top-*` (etc) classes to `t-0`, like we do with our spacing classes?

- I'm not sure if the inset naming is clear enough, I guess lot's of people will search for top/right/bottom/left classes. Since those terms do not even appear in titles on this page, they probably won't find them. Thoughts on this?

ffoodd:
@MartijnCuppens since those utilities won't do anything without `position`, I think we should at least keep `position` utilities here.

Maybe should those be documented in the `position` utility page, instead of having their own?

For discoverability, indeed the sides should at least appear in the page description  and the content, I guess.

mdo:
`inset` support is virtually nonexistent, right? https://caniuse.com/?search=inset

ffoodd:
Indeed but it's just used as an alias for multiple properties utilities (eg. `top` and `bottom` are `inset-y`).

This is consistent with what I introduce in #30980 with "logical names".

mdo:
I get the logical properties thing, but is it of use to us or anyone else at this point? Would RTL be using those aliases? Feels premature to include right now.

jaumesala:
Maybe related to #31280

ffoodd:
@jaumesala It's addressing the same issue, indeed. 

Comparing the two, #31280 feels more natural to me (mostly for the docs part, included in position utilities). However this one introduces `$position-sizes` map to fine tune positions—which I find useful.

@mdo regarding naming / RTL, `left` and `right` will be renamed as `start` and `end` but there's no need for `inset`. 

BTW I'd be in favor of declaring each needed directions instead of using extra utilities for corners or axis.

mdo:
I'm liking the docs and approach in #31280 more, but we'll need to update the directions to be logical ones instead. Thoughts on closing this PR for that one @ffoodd?

ffoodd:
I agree. Does the Map for values make any sense and might be added?

mdo:
Thanks for the work here, but we'll be going with #31280.

PR #31489, opened 2020-08-18
Add classes to remove rounding from single side

blueyetisoftware:
This PR adds styles that remove the rounding from a single side of an element that may already be round. They are similar to rounded-0 and are the corollary to their rounded-SIDE siblings.

### DIFF
--- a/dist/css/bootstrap-utilities.css
+++ b/dist/css/bootstrap-utilities.css
@@ -1097,6 +1097,26 @@
   border-top-left-radius: 0.25rem !important;
 }
 
+.rounded-top-0 {
+  border-top-left-radius: 0 !important;
+  border-top-right-radius: 0 !important;
+}
+
+.rounded-right-0 {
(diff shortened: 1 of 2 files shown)


phinatic:
I wasn't certain of how the placeholders work in the markdown. If the radius is 0 by default, then the additional styling will not have a visible impact. The original style must be rounded to start, but I didn't want to add the rounded style into the class and confuse readers.

phinatic:
I also didn't see anywhere that the *-0 properties were configured in the scss files, so I added the in the utils css

ffoodd:
Do not commit `dist` files. Please read [our contribution guidelines](https://github.com/twbs/bootstrap/blob/main/.github/CONTRIBUTING.md#pull-requests) and  update your branch consequently.

mdo:
Closing as stale and as incorrect. Unclear to me if we actually need these anyway tbh.

PR #31500, opened 2020-08-18
Use next breakpoint when targetting xs only

### LINKED ISSUE #31199: media-breakpoint-only(xs) not working
While migrating to v5 alpha1 I've experienced `@media` query for `media-breakpoint-only(xs)` is not generated any more.
I can work around the issue by specifying other breakpoint mixings, but I think it is good to open an issue for this use case.

Have a nice week
midzer

ffoodd:
Fixes #31199

[I made a Sassmeister to try this](https://www.sassmeister.com/gist/3372695771ff2a0d5fa708dac470e414). Simply change `$upper` to `$name` on L327 to reproduce the issue.


MartijnCuppens:
Maybe `$next` is a better name than `$upper`?

MartijnCuppens (approved):

mdo (approved):

PR #31502, opened 2020-08-18
Responsively align last line of text using utils

blueyetisoftware:
For themes that modify the alignment of the last line of text, the responsive utils do not update the alignment of the last line of text. This keeps the responsiveness consistent with the rest of the text block.

### DIFF
--- a/scss/_utilities.scss
+++ b/scss/_utilities.scss
@@ -376,7 +376,7 @@ $utilities: map-merge(
     ),
     "text-align": (
       responsive: true,
-      property: text-align,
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
This is not supported in Safari: https://caniuse.com/#feat=css-text-align-last

phinatic:
@MartijnCuppens That is true, but why close the issue? It is CSS3 compliant. If it is being used, this makes bootstrap handle responsiveness correctly. For browsers that don't handle it, there doesn't appear to be a negative effect.

PR #31503, opened 2020-08-19
Typo Mistake

JiGo2282:
Typo Mistake Fixed on 

https://getbootstrap.com/docs/4.5/layout/grid/


patrickhlauke:
i believe it's intentional - see also all the other cases in https://getbootstrap.com/docs/4.5/layout/grid/#alignment

the "One of" is not trying to number them (a la "1 of 3", "2 of 3", "3 of 3"), but saying "this is one of the three columns"

PR #31518, opened 2020-08-21
#25110 Missing border radius on input group with validation feedback

dejakob:
## Issue reported

[See: #25110](https://github.com/twbs/bootstrap/issues/25110)

34426369-e70a858c-ec35-11e7-86f4-f178276e1f6e

## Cause of issue

There is a selector in `_input-group.scss` that sets border-radius to 0 on all elements within an input-group that are not a dropdown or the last element:

```scss
.input-group {
  > :not(:last-child):not(.dropdown-toggle):not(.dropdown-menu)
```

## Proposed solution

I added an extra pseudo-selector to check if there is no `.invalid-feedback` element coming after it, before overriding the border-radius to 0.

Screenshot 2020-08-21 at 14 11 48


ffoodd:
Thanks for submmitting a PR, and documenting it clearly!

However [`:has()` has litterally no support at all…](https://caniuse.com/#search=%3Ahas) so your PR only invalidates the selector stack: look at `.input-group-text`, it get its right angles rounded.

As [per comment](https://github.com/twbs/bootstrap/issues/25110#issuecomment-586565165), the current pattern does not allow to solve this: v5 would probably need to handle this by changing the HTML pattern, [as stated in a comment too](https://github.com/twbs/bootstrap/issues/25110#issuecomment-628865021).

I'm closing this as a won't fix for now, but feel free to open another PR with a new pattern suggestion. This is one of our painpoint in v4, and we hope to find a way for v5 :)

Issue #31515, opened 2020-08-20
Add a default border radius value to all border radius mixins


jeremyvii:
It would be convenient if the `$border-radius` variable was the default value for all border radius mixins, like it is for [`border-radius()`](https://github.com/twbs/bootstrap/blob/main/scss/mixins/_border-radius.scss#L17). That way you could include mixins, such as `border-top-radius()`, without providing a parameter. 

I believe this would be an easy change and would be happy to open a PR. However, I wanted to make sure there wasn't a reason why there isn't already a default parameter for these mixins.

Before proposal
``` scss
.item {
  @include border-top-radius($border-radius);
}
```

After proposal
``` scss
.item {
  @include border-top-radius();
}
```

jeremyvii:
Closing since #31571 has been merged.

PR #31571, opened 2020-09-01
Add default parameters to each border radius mixin

jeremyvii:
This addresses my feature request (#31515) by adding a default parameter to each border radius mixin. The default value used is the `$border-radius` SCSS variable.


ffoodd (approved):
LGTM, very handy indeed.

mdo:
Can you add a note to the Migration guide about the mixin now having a default value?

mdo (approved):
Thanks!

PR #31572, opened 2020-09-01
Fix docs incorrect green and cyan hex codes

Deckluhm:
https://v5.getbootstrap.com/docs/5.0/customize/color

### DIFF
--- a/site/data/colors.yml
+++ b/site/data/colors.yml
@@ -13,11 +13,11 @@
 - name: yellow
   hex: "#ffc107"
 - name: green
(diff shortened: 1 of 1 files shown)


Deckluhm:
I just noticed https://github.com/twbs/bootstrap/blob/main/site/data/theme-colors.yml is also referencing the wrong `success` hex code (`#28a745` instead of `#198754`) but the hex part of this file doesn't seem to be used anywhere?

Should I open another PR to fix the `success` hex code or to remove all hex codes in theme-colors.yml?

Issue #31538, opened 2020-08-26
v5: Improve customization by allowing manually setting calculated variables in styles


ydmitry:
Hello!

We can have cases when we need to change specific style property, but we can't do this because this property value set up in Bootstrap library as calculated formula. The idea is to create separate variables for these cases and precalculate it in default values of these variables.

The same was done in this issue and PR, but only for button component:
https://github.com/twbs/bootstrap/issues/28899
https://github.com/twbs/bootstrap/pull/29444

The idea is to follow this practice in all components.

### Example

Before:

```
@mixin list-group-item-variant($state, $background, $color) {
  .list-group-item-#{$state} {
    // ...
    background-color: darken($background, 5%);
    // ...
  }
}
```

After:

```
@mixin list-group-item-variant($state, $background, $color, $background-hover: darken($background, 5%)) {
  .list-group-item-#{$state} {
    // ...
    background-color: $background-hover;
    // ...
  }
}
```

The same can be done not only for colors, but for calculated unit values:

Before:
```
.card-subtitle {
  margin-top: -$card-spacer-y / 2;
}
```

After:
```
// In variables:
$card-subtitle-spacer: -$card-spacer-y / 2 !default;

// In styles:
.card-subtitle {
  margin-top: $card-subtitle-spacer;
}
```

### Benefits

- Allows to customize more properties and if you don't need - just keep default calculation of the value
- Allows to transform it to CSS variable usage by setting e.g. `$background-hover: var(--bs-background-hover);` (if we have `background-color: darken($background, 5%);` we can't do this)

### Drawbacks

- More variables to support
- Longer migration guide (probably) 

What do you thik about this? Should I do a PR?

MauricioHernanCabrera:
I can do that!

ydmitry:
Thank you :) . What about to do it as a rule for all scss styles? The original feature request to have this as styleguide rule and apply it everywhere, not only for list group and card subtitle

ffoodd:
@ydmitry That sounds great indeed. It may require to think our variables organization again—and probably to document each components' variables too, as already requested several times…

Any thoughts, @mdo?

MauricioHernanCabrera:
> Thank you :) . What about to do it as a rule for all scss styles? The original feature request to have this as styleguide rule and apply it everywhere, not only for list group and card subtitle

if you want, I can work on that

mdo:
Been trying to find the best way to review these PRs, and I think overall most of the changes feel over-engineered.

[...]

PR #31574, opened 2020-09-01
Customization: card subtitle and list group

MauricioHernanCabrera:
https://github.com/twbs/bootstrap/issues/31538

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -39,7 +39,7 @@
   // Prevent double borders when buttons are next to each other
   > .btn:not(:first-child),
   > .btn-group:not(:first-child) {
-    margin-left: -$btn-border-width;
+    margin-left: $btn-group-not-first-child-margin-left;
   }
 
   // Reset rounded corners
@@ -125,7 +125,7 @@
 
   > .btn:not(:first-child),
   > .btn-group:not(:first-child) {
-    margin-top: -$btn-border-width;
+    margin-top: $btn-group-vertical-not-first-child-margin-top;
   }
 
   // Reset rounded corners
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -72,8 +72,8 @@
 //
 
 .dropdown-toggle-split {
-  padding-right: $btn-padding-x * .75;
-  padding-left: $btn-padding-x * .75;
+  padding-right: $dropdown-toggle-split-padding-right;
+  padding-left: $dropdown-toggle-split-padding-left;
 
   &::after,
   .dropup &::after,
@@ -87,13 +87,13 @@
 }
 
 .btn-sm + .dropdown-toggle-split {
-  padding-right: $btn-padding-x-sm * .75;
-  padding-left: $btn-padding-x-sm * .75;
+  padding-right: $dropdown-toggle-split-btn-sm-padding-right;
+  padding-left: $dropdown-toggle-split-btn-sm-padding-left;
 }
 
 .btn-lg + .dropdown-toggle-split {
-  padding-right: $btn-padding-x-lg * .75;
-  padding-left: $btn-padding-x-lg * .75;
+  padding-right: $dropdown-toggle-split-btn-lg-padding-right;
+  padding-left: $dropdown-toggle-split-btn-lg-padding-left;
 }
 
 
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -236,7 +236,7 @@
 
     > .card-header {
       @include border-radius(0);
-      margin-bottom: -$card-border-width;
+      margin-bottom: $accordion-card-header-margin-bottom;
     }
   }
 }
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -55,7 +55,7 @@
 }
 
 .card-subtitle {
-  margin-top: -$card-title-spacer-y / 2;
+  margin-top: $card-subtitle-spacer;
   margin-bottom: 0;
 }
 
--- a/scss/_card.scss
+++ b/scss/_card.scss
@@ -120,8 +120,8 @@
 }
 
 .card-header-pills {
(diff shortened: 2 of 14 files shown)


ydmitry:
Should this have docs to be updated or migration guide?

MauricioHernanCabrera:
> Here are several values calculated in styles https://github.com/twbs/bootstrap/blob/main/scss/_card.scss#L109

Okaay. Im going to work on that

ydmitry:
@MauricioHernanCabrera need to handle somehow places like this:

https://github.com/twbs/bootstrap/blob/main/scss/_alert.scss 

```
@each $color, $value in $theme-colors {
  .alert-#{$color} {
    @include alert-variant(color-level($value, $alert-bg-level), color-level($value, $alert-border-level), color-level($value, $alert-color-level));
  }
}
```

E.g. it can be:

```
// Variables file
$alert-theme-colors: ();
@each $color, $value in $theme-colors {
  $alert-theme-colors: map-merge(
     $color: (
        background: color-level($value, $alert-bg-level),
        border: color-level($value, $alert-border-level),
        color: color-level($value, $alert-color-level),
     )
  );

// ...

// Styles file

@each $color, $value in $alert-theme-colors {
  .alert-#{$color} {
    @include alert-variant(map-get($value, background), map-get($value, border), map-get($value, color));
  }
}
```

MauricioHernanCabrera:
> @MauricioHernanCabrera need to handle somehow places like this:
> 
> https://github.com/twbs/bootstrap/blob/main/scss/_alert.scss
> 
> ```
> @each $color, $value in $theme-colors {
>   .alert-#{$color} {
>     @include alert-variant(color-level($value, $alert-bg-level), color-level($value, $alert-border-level), color-level($value, $alert-color-level));
>   }
> }
> ```
> 
> E.g. it can be:
> 
> ```
> // Variables file
> $alert-theme-colors: ();
> @each $color, $value in $theme-colors {
>   $alert-theme-colors: map-merge(
>     $color: (
>       background: color-level($value, $alert-bg-level),
>       border: color-level($value, $alert-border-level),
>       color: color-level($value, $alert-color-level),
>     )
>   );
> }
> 
> // ...
> 
> // Styles file
> 
> @each $color, $value in $alert-theme-colors {
>   .alert-#{$color} {
>     @include alert-variant(map-get($value, background), map-get($value, border), map-get($value, color));
>   }
> }
> ```

okasss, I'm going to work on that

scss/_button-group.scss
```diff
@@ -125,7 +125,7 @@
 
   > .btn:not(:first-child),
   > .btn-group:not(:first-child) {
-    margin-top: -$btn-border-width;
+    margin-top: $btn-group-vertical-not-first-child-margin-top;
```
ydmitry, line 128: For this case need to consider what to do, because this is more technical solution to keep margins with border and align everything, so probably it will be always `-$btn-border-width`, so solution here can be

`margin-top: calc($btn-border-width * -1)` or `margin-top: $btn-margin-top-align-by-border-width`

ydmitry:
@MauricioHernanCabrera I think need to rename PR or split by components

scss/_button-group.scss
```diff
@@ -39,7 +39,7 @@
   // Prevent double borders when buttons are next to each other
   > .btn:not(:first-child),
   > .btn-group:not(:first-child) {
-    margin-left: -$btn-border-width;
+    margin-left: $btn-group-not-first-child-margin-left;
```
mdo, line 42: This would never change really, so it doesn't make sense to have a unique variable for it.

scss/_button-group.scss
```diff
@@ -72,8 +72,8 @@
 //
 
 .dropdown-toggle-split {
-  padding-right: $btn-padding-x * .75;
-  padding-left: $btn-padding-x * .75;
+  padding-right: $dropdown-toggle-split-padding-right;
```
mdo, line 75: These values should likely always be the same. Let's stick with one `-x` in the variable name instead of two variable with `-right` and `-left`.

scss/_card.scss
```diff
@@ -106,9 +106,9 @@
 //
 
 .card-header-tabs {
-  margin-right: -$card-cap-padding-x / 2;
-  margin-bottom: -$card-cap-padding-y;
-  margin-left: -$card-cap-padding-x / 2;
+  margin-right: $card-header-tabs-margin-right;
```
mdo, line 109: `-x` and `-y`, not `-right` etc.

scss/_card.scss
```diff
@@ -120,8 +120,8 @@
 }
 
 .card-header-pills {
-  margin-right: -$card-cap-padding-x / 2;
-  margin-left: -$card-cap-padding-x / 2;
+  margin-right: $card-header-pills-margin-right;
+  margin-left: $card-header-pills-margin-left;
```
mdo, line 124: Same here—`-x` and `-y`, not `-right` etc.

scss/_carousel.scss
```diff
@@ -185,9 +185,9 @@
 
 .carousel-caption {
   position: absolute;
-  right: (100% - $carousel-caption-width) / 2;
+  right: $carousel-caption-right;
```
mdo, line 188: Same here too—`-x` and `-y`, not `-right` etc.

mdo (approved):
Half reviewed this, but stopped after thinking about the fact that this PR should be split up. I'd also encourage you to ask about our approach to naming before tackling larger efforts like this to avoid wasting any time :).

mdo (changes requested):
Meant to hit request changes, sorry!

scss/_card.scss
ydmitry: @mdo isn't `y`  for `top` / `bottom` values ? 
according to naming rules what if we need to create variable for `right` and `left` separately? We can't use `x` for both...

mdo:

PR #31578, opened 2020-09-02
Scrollspy keyboard access in docs

ffoodd:
Mention using `tabindex="0"` on scrollable areas and apply it in our examples, to ensure keyboard access [as stated in axe-core](https://dequeuniversity.com/rules/axe/3.5/scrollable-region-focusable) and [explained by Heydon Pickering in its inclusive components (eg. "Content Slider")](https://inclusive-components.design/a-content-slider/#keyboardsupport).

### DIFF
--- a/site/content/docs/5.0/components/scrollspy.md
+++ b/site/content/docs/5.0/components/scrollspy.md
@@ -12,10 +12,15 @@ Scrollspy has a few requirements to function properly:
 
 - It must be used on a Bootstrap [nav component]({{< docsref "/components/navs" >}}) or [list group]({{< docsref "/components/list-group" >}}).
 - Scrollspy requires `position: relative;` on the element you're spying on, usually the `<body>`.
-- When spying on elements other than the `<body>`, be sure to have a `height` set and `overflow-y: scroll;` applied.
 - Anchors (`<a>`) are required and must point to an element with that `id`.
 
 When successfully implemented, your nav or list group will update accordingly, moving the `.active` class from one item to the next based on their associated targets.
+
+{{< callout >}}
+### Scrollable containers and keyboard access
+
+If you're making a scrollable container (other than the `<body>`), be sure to have a `height` set and `overflow-y: scroll;` applied to it—alongside a `tabindex="0"` to ensure keyboard access.
+{{< /callout >}}
 
 ## Example in navbar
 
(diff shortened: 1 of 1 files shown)


site/content/docs/5.0/components/scrollspy.md
```diff
@@ -12,7 +12,7 @@ Scrollspy has a few requirements to function properly:
 
 - It must be used on a Bootstrap [nav component]({{< docsref "/components/navs" >}}) or [list group]({{< docsref "/components/list-group" >}}).
 - Scrollspy requires `position: relative;` on the element you're spying on, usually the `<body>`.
-- When spying on elements other than the `<body>`, be sure to have a `height` set and `overflow-y: scroll;` applied.
+- When spying on elements other than the `<body>`, be sure to have a `height` set and `overflow-y: scroll;` applied—alongside a `tabindex="0"` to ensure keyboard access.
```
patrickhlauke, line 15: i'd even consider splitting out the entire bullet into a separate callout, giving a bit more context about "If you're making a scrollable container..."

ffoodd: That made sense to me to add it on the same sentence than the one specifying how to make a scrollable container.
I'll try another bullet, let me know what you think :)

patrickhlauke: no sorry i meant the entire bullet (not just your additional point) there could be ripped out into a callout (as it's not necessarily about scrollspy per se, but more about "make sure you do these things when making a scroller")

ffoodd: Oh ok, I misunderstood!

Let me try this, then. That makes sense.

XhmikosR:
@patrickhlauke LGTY?

site/content/docs/5.0/components/scrollspy.md
```diff
@@ -12,11 +12,16 @@ Scrollspy has a few requirements to function properly:
 
 - It must be used on a Bootstrap [nav component]({{< docsref "/components/navs" >}}) or [list group]({{< docsref "/components/list-group" >}}).
 - Scrollspy requires `position: relative;` on the element you're spying on, usually the `<body>`.
-- When spying on elements other than the `<body>`, be sure to have a `height` set and `overflow-y: scroll;` applied.
 - Anchors (`<a>`) are required and must point to an element with that `id`.
 
 When successfully implemented, your nav or list group will update accordingly, moving the `.active` class from one item to the next based on their associated targets.
 
+{{< callout >}}
+### Spying on elements other than the `<body>`
```
patrickhlauke, line 20: I'd say this is more a quick tip on "Scrollable containers" - not specific to spying

patrickhlauke (changes requested):
Apart from the tiny change in heading, it looks good

site/content/docs/5.0/components/scrollspy.md
ffoodd: My understanding is:
1. if you're not spying `body`, you'll have a scrollable container,
2. if we mention "scrollable containers",  `body` is concerned.

I'm not sure how I should reword this, I may be misunderstanding both words and concepts here.

patrickhlauke:
my point is: this isn't about spying per se. the tip is in general about how to make a proper div or whatever that works as a scrollable container and is keyboard accessible. the "spying" bit is incidental.

ffoodd:
@patrickhlauke I reworded things a bit. Is that better?

patrickhlauke (approved):

patrickhlauke:
:shipit:

PR #31580, opened 2020-09-02
Wrap home content in <main> & use <section>

ffoodd:
Followups content should probably be in a landmark to improve discoverability for ATs; thus I wrapped them in `main`, and also used `section`s too since that's basically what this element is for.

This one is not critical at all, opinions are welcome.

BTW: our current template for docs also have its intro content (`h1` and lead) outside any landmark—thus not discoverable by users navigating with landmarks. It'd require to revamp the whole grid layout to change this: should I suggest a fix?


mdo (approved):
I think this all makes sense to me!

XhmikosR:
How about the other layouts? We aren't using layouts hierarchy yet.

ffoodd:
@XhmikosR that what I mention at the end of my description: it needs to redo our grid layout. I think I'm going to tackle this as a separate PR.

PR #31583, opened 2020-09-02
docs(Modal): Add xxl fullscreen example

kyletsang:
Docs was missing a modal fullscreen xxl example, so added it here.


mdo (approved):
Approved for Alpha 3.

PR #31607, opened 2020-09-06
Use Babel and ES6 in docs JS files

### LINKED ISSUE #30818: docs v5: switch to ES6
[Hugo v0.70.0](https://github.com/gohugoio/hugo/releases/tag/v0.70.0) introduced the ability to use [Babel](https://gohugo.io/hugo-pipes/babel/), which means we can finally use ES6 everywhere in the codebase for v5.

There's still a bug that needs to be fixed, but should only affect Windows: <https://github.com/gohugoio/hugo/issues/7251>

NotWoods:
Fixes #30818

Updates the `assets/js` path inside of `site` so that the code is passed through Babel, using the root Babel RC file.

Additionally adds a ESLint config in the assets folder to enable a higher ecma version, but only for doc files.

---

I didn't pass `extra_js` scripts through Babel, since the equivalent `extra_css` scripts in the Examples folder don't get processed. I can add this if desired.

Preview: https://deploy-preview-31607--twbs-bootstrap.netlify.app/

TODO:

- [x] Revisit if we need babel after all
- [x] `Array.from` instead of `Array.prototype.slice.call`
- [x] Look into inheriting from root .eslintrc.json

### DIFF
--- a/site/assets/js/application.js
+++ b/site/assets/js/application.js
@@ -35,7 +35,7 @@
         toastPlacement.dataset.originalClass = toastPlacement.className
       }
 
-      toastPlacement.className = toastPlacement.dataset.originalClass + ' ' + this.value
+      toastPlacement.className = `${toastPlacement.dataset.originalClass} ${this.value}`
     })
   }
 
@@ -63,7 +63,7 @@
 
   function alert(message, type) {
     const wrapper = document.createElement('div')
-    wrapper.innerHTML = '<div class="alert alert-' + type + ' alert-dismissible" role="alert">' + message + '<button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button></div>'
+    wrapper.innerHTML = `<div class="alert alert-${type} alert-dismissible" role="alert">${message}<button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button></div>`
 
     alertPlaceholder.append(wrapper)
   }
@@ -112,7 +112,7 @@
       const modalTitle = exampleModal.querySelector('.modal-title')
       const modalBodyInput = exampleModal.querySelector('.modal-body input')
 
-      modalTitle.textContent = 'New message to ' + recipient
+      modalTitle.textContent = `New message to ${recipient}`
       modalBodyInput.value = recipient
     })
   }
@@ -165,7 +165,7 @@
 
   clipboard.on('error', event => {
     const modifierKey = /mac/i.test(navigator.userAgent) ? '\u2318' : 'Ctrl-'
-    const fallbackMsg = 'Press ' + modifierKey + 'C to copy'
+    const fallbackMsg = `Press ${modifierKey}C to copy`
     const tooltipBtn = bootstrap.Tooltip.getInstance(event.trigger)
 
     tooltipBtn.setContent({ '.tooltip-inner': fallbackMsg })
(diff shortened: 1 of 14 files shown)


XhmikosR:
Thanks for the PR! I'm going to have a closer look in the next days, but
from a quick look I notice you don't use imports. Also please don't change
any listing configuration files.

NotWoods:
I didn't realize adding imports was one of the goals, I'll take another look and see how Hugo's bundler works. I didn't mean to change the root eslint file but the subfile is nessecary.

XhmikosR (changes requested):
babel should only run for our files and not all files

site/assets/js/application.js
```diff
@@ -39,85 +39,79 @@
...
-      // Update the modal's content.
-      var modalTitle = exampleModal.querySelector('.modal-title')
-      var modalBodyInput = exampleModal.querySelector('.modal-body input')
-
-      modalTitle.textContent = 'New message to ' + recipient
-      modalBodyInput.value = recipient
-    })
-  }
+  const exampleModal = document.getElementById('exampleModal')
+  exampleModal?.addEventListener('show.bs.modal', event => {
```
XhmikosR, line 67: Can you revert the optional chaining changes and revert the .eslintrc.json change? We don't use optional chaining anywhere else right now and I'd like to be consistent.

XhmikosR:
@regisphilibert: how does this implementation look good to you? Is there any other way to achieve the same result without the temp file?

Also, thanks @NotWoods for sticking with us, every thing to improve things is welcome :)

site/assets/js/application.js
NotWoods: If I remove the eslint file, the linter will complain about any ES6 syntax (i.e. `const`). It's set up to only affect files in the docs folder.

I'll revert the optional chains.

XhmikosR: NVM about the comment regarding the ESLint changes. Your changes are good since we have an upper eslintrc.json in site folder because the examples still use ES5.

XhmikosR:
@Johann-S what do you think about this?

My concerns are:

1. it slows down the build a lot, but I suppose it's bearable
2. we don't use imports

I wonder if we could use esbuild later which should be a lot faster.

Johann-S:
It seems ok to me, but yes we can import our source code that's possible too, but it'll produce a bigger js file and we won't use the Bootstrap CDN

XhmikosR:
We don't use a CDN for our docs :)

mdo:
Still worth finishing off @XhmikosR?

XhmikosR:
I'll need to rebase the branch since it has conflicts.

That being said, it's the wise thing to make things consistent with code. The only downside I see is that this slows down the build, hence why I've left the PR hanging.

But I'll take another stab at it in the next days.

GeoSot:
I think we can use `NodeList.prototype.forEach()` in place of `Array.from` 

https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#browser_compatibility

GeoSot (approved):

GeoSot:
**/CC** @XhmikosR any opinion over this? 
>I think we can use NodeList.prototype.forEach() in place of Array.from

I think it's the last thing in order to merge

mdo:
@GeoSot Want to wait for that change, or merge this and make it in another PR?

Issue #28479, opened 2019-03-13
Hugo TODO


XhmikosR:
1. **make sure our folder structure is the proper one for what we want to achieve. Also see if we can make folder nesting level smaller, maybe by using permalinks or something. For content and static.**
2. can we generate SRI hashes for existent files on the fly? For bootstrap.min.css, bootstrap.*.js,  popper.js/1.14.7/umd/popper.min.js. This would simplify things as we could remove the release-sri script
3. get image width/height with Hugo when possible (examples images, homepage)
4. ~~maybe use Hugo Extended version for docs' CSS?~~ This is now done
5. bep's suggestion:
    > Re.
    > ```
    > {{< partial "callout-warning-color-assistive-technologies.md" >}}
    > ```
    > The partial above uses the `partial` func to load, parse and execute templates. This has its limitations and is possibly ineffective.
    > I would recommend that you store your callouts (and other content files you want to include) in a so-called headless bundle -- and load them via `.GetPage`. This way you can use markdown + shortcodes etc.
6. bep's suggestion:
    > Ref. `{{ block "header" . }}` in your `single.html` template.
    > Blocks only make sense inside base template(s) (i.e. `baseof.html`), which I cannot see.  I would highly recommend that you restructure your templates to use them. It removes some repetition.
7. Fix https://github.com/twbs/bootstrap/blob/6335fbe224f28d2ce65c628756a84113ea65643f/site/layouts/_default/docs.html#L27
8. Add a body class like `layout-{{layout}} page-{{kind}}-{{title}}{{ with .Params.body_class }} {{ . }}{{ end }}`

XhmikosR:
@bep: can you give me an example how to get the SRI hashes https://github.com/twbs/bootstrap/blob/master/config.yml#L49?

The files do exist locally, we just use the cdn link for the doc snippets.

Also, is there any way to get a sha384 hash instead of sha256? I know that 256 should be enough but that's what  we've been using from day 1 and also we suggest on srihash.org, BootstrapCDN etc.

Thanks :)

bep:
@XhmikosR have you looked in the documentation? Note that I'm very busy at the moment (also with Hugo), so I would appreciate if you could take general Hugo questions either with 1) the docs or 2) the forum

It is sha256 or sha512 (or md5).

XhmikosR:
All right, I'll try again based on the docs alone.

Have you thought about exposing sha384 too, though? I can make an issue upstream if you want.

[...]

PR #31608, opened 2020-09-06
Add baseof.html layout to docs

NotWoods:
For #28479

Adds a base layout used by most other layouts. Also adds a base "redirect" layout to avoid some duplication.

---

Biggest sticking point here is `body_override`. 

Seems like hugo doesn't like having an attribute defined like this:

```
<body {{ block "body_attribute" . }}{{ end }}>
```

Since this is used just for the class attribute, you could add a class override like this:

```
<body class="{{ block "body_class" . }}{{ end }}">
```

But then you have `<body class="">` on most pages. The issue indicates that you may want to add a body class for most pages anyways, so if this is a preferable approach let me know.

Preview: <https://deploy-preview-31608--twbs-bootstrap.netlify.app/>

TODO:

- [x] Test release-zip-examples just in case


XhmikosR:
@NotWoods Thanks for the PR. Can you rebase this please?

@regisphilibert could you also have a quick look please?

XhmikosR:
> But then you have `<body class="">` on most pages. The issue indicates that you may want to add a body class for most pages anyways, so if this is a preferable approach let me know.

Indeed, I'd like to always have a body class, but maybe we should tackle that later. My idea was to has a body class automatically generated based on the layout used and page's file path (with slashes and the extension removed).

I think your patch is a good start and can land as is for now :)

XhmikosR:
BTW it seems the example.html is left out.

NotWoods:
I changed it around to use a block for the class, and that can be expanded upon later to generate a class name from the page path. `example.html` is intentionally left out because the layout is unique and doesn't share much with the other pages.

XhmikosR:
> I changed it around to use a block for the class, and that can be expanded upon later to generate a class name from the page path. `example.html` is intentionally left out because the layout is unique and doesn't share much with the other pages.

I'll revert the change since it doesn't make sense to have this unless it's ready.

About examples, I still think we could reuse some of the partials, but I'll leave this to explore later. Like scripts.html and try to generate the placeholder inline CSS from Sass instead of manually hardcoding its output.

XhmikosR (approved):

XhmikosR:
I'll do a little more testing later and merge this. Thanks for your contribution(s)!

XhmikosR:
@NotWoods can you apply this against my v4-dev-xmr-hugo branch too please? The only difference is the meta viewport tag and the fact that we don't include footer.html in the docs layout.

NotWoods:
#32192

PR #31643, opened 2020-09-14
Add a state field to create utility pseudo-classes

### LINKED ISSUE #31639: [feature][v5] Hover-* class Utilities
Hello

I would like to suggest the hover utilities. The utility builds on existing utilities.

Example:

```html
<p class="hover-success">On hover, the text will be the color "success"</p>
<div class="hover-shadow-sm">On hover there will be a shadow from the "shadow-sm" utility</div>
```

Developers often want to use custom hovers in their blocks. Instead of using ready-made utilities, they write their own custom code. For faster development, it is good to have a standard set of hovers.

The `$enable-hover-utilites` variable will be used to enable hover utilities.

When enabled, classes for hover utilities are created. These utilities are properties that do not affect the box-model.

I suggest including this in Bootstrap 5.

Psixodelik:
Fixes #31639 

Added state option for utilities. This will generate custom pseudo-classes when generating utilities.

### SCSS

```scss
$utilities: (
  "shadow": (
    property: box-shadow,
    state: hover focus,
    class: shadow,
    values: (
      null: $box-shadow,
      sm: $box-shadow-sm,
      lg: $box-shadow-lg,
      none: none,
    )
  )
);
```

### CSS output

```css
.shadow-hover:hover {
  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
}

.shadow-focus:focus {
  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
}

.shadow-sm-hover:hover {
  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
}

.shadow-sm-focus:focus {
  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
}

.shadow-lg-hover:hover {
  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
}

.shadow-lg-focus:focus {
  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
}

.shadow-none-hover:hover {
  box-shadow: none !important;
}

.shadow-none-focus:focus {
  box-shadow: none !important;
}
```

### DIFF
--- a/scss/mixins/_utilities.scss
+++ b/scss/mixins/_utilities.scss
@@ -20,6 +20,9 @@
     $property-class: if(map-has-key($utility, class), map-get($utility, class), nth($properties, 1));
     $property-class: if($property-class == null, "", $property-class);
 
+    // State params to generate pseudo-classes
+    $state: if(map-has-key($utility, state), map-get($utility, state), ());
+
     $infix: if($property-class == "" and str-slice($infix, 1, 1) == "-", str-slice($infix, 2), $infix);
 
     // Don't prefix if value key is null (eg. with shadow class)
@@ -42,6 +45,14 @@
       .#{$property-class + $infix + $property-class-modifier} {
         @each $property in $properties {
           #{$property}: $value if($enable-important-utilities, !important, null);
+        }
+      }
+
+      @each $pseudo in $state {
+        .#{$property-class + $infix + $property-class-modifier}-#{$pseudo}:#{$pseudo} {
+          @each $property in $properties {
+            #{$property}: $value if($enable-important-utilities, !important, null);
+          }
         }
       }
     }
--- a/site/content/docs/5.0/utilities/api.md
+++ b/site/content/docs/5.0/utilities/api.md
@@ -264,3 +264,59 @@ $utilities: map-merge(
   )
 );
 ```
+
+## Adding pseudo-classes to utilities
+
+With the `state` option, pseudo-class utilities can be generated:
+
+```scss
+$utilities: (
+  "shadow": (
+    property: box-shadow,
+    state: hover focus,
+    class: shadow,
+    values: (
+      null: $box-shadow,
+      sm: $box-shadow-sm,
+      lg: $box-shadow-lg,
+      none: none,
+    )
+  )
+);
+```
+
+Output:
+
+```css
+.shadow-hover:hover {
+  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
+}
+
+.shadow-focus:focus {
+  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
+}
+
+.shadow-sm-hover:hover {
+  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
+}
+
+.shadow-sm-focus:focus {
+  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
+}
+
+.shadow-lg-hover:hover {
+  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
+}
+
+.shadow-lg-focus:focus {
+  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
+}
+
+.shadow-none-hover:hover {
+  box-shadow: none !important;
+}
+
+.shadow-none-focus:focus {
+  box-shadow: none !important;
+}
+```
--- a/site/content/docs/5.0/utilities/api.md
+++ b/site/content/docs/5.0/utilities/api.md
@@ -17,6 +17,7 @@ The `$utilities` map contains all our utilities and is later merged with your cu
 | `property` | **Required** | Name of the property, this can be a string or an array of strings (e.g., horizontal paddings or margins). |
 | `values` | **Required** | List of values, or a map if you don't want the class name to be the same as the value. If `null` is used as map key, it isn't compiled. |
 | `class` | Optional | Variable for the class name if you don't want it to be the same as the property. In case you don't provide the `class` key and `property` key is an array of strings, the class name will be the first element of the `property` array. |
+| `state` | Optional | List of pseudo-class variants like `:hover` or `:focus` to generate for the utility. No default value. |
 | `responsive` | Optional | Boolean indicating if responsive classes need to be generated. `false` by default. |
 | `rfs` | Optional | Boolean to enable fluid rescaling. Have a look at the [RFS]({{< docsref "/getting-started/rfs" >}}) page to find out how this works. `false` by default. |
 | `print` | Optional | Boolean indicating if print classes need to be generated. `false` by default. |
@@ -79,6 +80,39 @@ Output:
 .o-50 { opacity: .5; }
 .o-75 { opacity: .75; }
 .o-100 { opacity: 1; }
+```
+
+## States
+
+Use the `state` option to generate pseudo-class variations. Example pseudo-classes are `:hover` and `:focus`. When a list of states are provided, classnames are created for that pseudo-class. For example, to change opacity on hover, add `state: hover` and you'll get `.opacity-hover:hover` in your compiled CSS.
+
+Need multiple pseudo-classes? Use a space-separated list of states: `state: hover focus`.
+
+```scss
+$utilities: (
+  "opacity": (
+    property: opacity,
+    class: opacity,
+    state: hover,
+    values: (
+      0: 0,
+      25: .25,
+      50: .5,
+      75: .75,
+      100: 1,
+    )
+  )
+);
+```
+
+Output:
+
+```css
+.opacity-0-hover:hover { opacity: 0; }
+.opacity-25-hover:hover { opacity: .25; }
+.opacity-50-hover:hover { opacity: .5; }
+.opacity-75-hover:hover { opacity: .75; }
+.opacity-100-hover:hover { opacity: 1; }
 ```
 
 ### Responsive utilities
@@ -263,60 +297,4 @@ $utilities: map-merge(
     "width": null
   )
 );
-```
-
-## Adding pseudo-classes to utilities
-
-With the `state` option, pseudo-class utilities can be generated:
-
-```scss
-$utilities: (
-  "shadow": (
-    property: box-shadow,
-    state: hover focus,
-    class: shadow,
-    values: (
-      null: $box-shadow,
-      sm: $box-shadow-sm,
-      lg: $box-shadow-lg,
-      none: none,
-    )
-  )
-);
-```
-
-Output:
-
-```css
-.shadow-hover:hover {
-  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
-}
-
-.shadow-focus:focus {
-  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
-}
-
-.shadow-sm-hover:hover {
-  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
-}
-
-.shadow-sm-focus:focus {
-  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;
-}
-
-.shadow-lg-hover:hover {
-  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
-}
-
-.shadow-lg-focus:focus {
-  box-shadow: 0 1rem 3rem rgba(0, 0, 0, 0.175) !important;
-}
-
-.shadow-none-hover:hover {
-  box-shadow: none !important;
-}
-
-.shadow-none-focus:focus {
-  box-shadow: none !important;
-}
 ```


scss/_utilities.scss
```diff
@@ -27,6 +27,7 @@ $utilities: map-merge(
     ),
     "shadow": (
       property: box-shadow,
+      hover: $enable-hover-utilites,
```
ydmitry, line 30: In this case I will generate `hover` together with simple state. What if I need to generate only `hover` utilities? And why only `hover`? Maybe I need `focus`? 
I would say declaration should be via additional property which will generate only for this property:

`state: "hover"`

full declaration:

```
 "shadow-hover": (
    property: box-shadow,
    state: hover,
    class: shadow,
    values: (...)
)
```

Psixodelik: Yes, at first I did it (you can see the old pr), but then I thought that it might be a complication. We need some kind of default state that describes classes that will not affect the box-model

Psixodelik: In fact, this is how it works now. By removing the hover field, you disable class generation for a specific property

patrickhlauke:
personally, not a fan of bloating the CSS with these sorts of pervasive utility classes

Psixodelik:
> personally, not a fan of bloating the CSS with these sorts of pervasive utility classes

Hi

I proceed from the fact that I have often seen and used such classes when creating sites. This is common since most designers like to use hover effects. For example, on the card component

scss/_utilities.scss
ydmitry: Is this about this PR https://github.com/twbs/bootstrap/pull/31640/files ? I see it's almost the same, just $enable-hover-utilites added

ffoodd:
Not sure about this either. It has very low value without transition, and this would massively increase the file size when enabled.

@twbs/css-review this (and other PR adding utilities out there) makes me wonder if we shouldn't publish our utility API as an external project —just like RFS— to allow anyone to easily build their own atomic CSS framework, and keep Bootstrap utilities in finer control.

Any opinions?

MartijnCuppens:
I'm ok with this PR since it's not disabled by default and could be useful in some cases.

> and this would massively increase the file size when enabled

When enabling `$enable-hover-utilites`, it's only added to some properties which don't have responsive variants, so I guess the difference won't be that huge.

I would personally not enable it on the rounded utilities, I can't think of any use cases for that. Also, this needs documentation.

> publish our utility API as an external project

I'm cool with that. The only downside is that we currently copy paste updates since we do not want to rely on package managers like npm. This could potentially be a PITA in the long run.

Psixodelik:
**MartijnCuppens**, thanks for the comment

>  would personally not enable it on the rounded utilities, I can't think of any use cases for that.

Yes you are right. Submitted a new commit

> Also, this needs documentation.

Is this worth adding to the theming section of the documentation?

MartijnCuppens:
@Psixodelik,

I would like to have some input from @mdo before you continue with this. We haven't really decided (or even discussed) what functionalities we're going to add to our new utility API. We could for example also add focus, dark mode utilities or even RTL utilities.

@mdo, what do you think?

Psixodelik:
@MartijnCuppens

Ok :). I added to the documentation for an example. Thanks

Psixodelik:
@MartijnCuppens 

If I come in handy in development, I am always ready to help

ffoodd:
FWIW my RTL branch extends the API a bit, to prevent some utilities from being generated in RTL output (`.text-break` mostly). Only used a boolean property: `rtl: false` that should be set explicitly.

site/content/docs/5.0/utilities/api.md
```diff
@@ -291,3 +291,39 @@ $utilities: (
   "float": null,
 );
 ```
+
+## Adding hover utilities
```
mdo, line 295:
```suggestion
## Hover utilities
```

site/content/docs/5.0/customize/options.md
```diff
@@ -26,4 +26,5 @@ You can find and customize these variables for key global options in Bootstrap's
 | `$enable-negative-margins`     | `true` or `false` (default)        | Enables the generation of [negative margin utilities]({{< docsref "/utilities/spacing#negative-margin" >}}). |
 | `$enable-deprecation-messages` | `true` or `false` (default)        | Set to `true` to show warnings when using any of the deprecated mixins and functions that are planned to be removed in `v5`. |
 | `$enable-important-utilities`  | `true` (default) or `false`        | Enables the `!important` suffix in utility classes. |
+| `$enable-hover-utilites`       | `true` or `false` (default)        | Enables the `:hover` pseudo-classes utility. |
```
mdo, line 29:
```suggestion
| `$enable-hover-utilities`       | `true` or `false` (default)        | Enables the `:hover` pseudo-classes utilities. |
```

scss/_variables.scss
```diff
@@ -221,6 +221,7 @@ $enable-validation-icons:     true !default;
 $enable-negative-margins:     false !default;
 $enable-deprecation-messages: true !default;
 $enable-important-utilities:  true !default;
+$enable-hover-utilites:       false !default;
```
mdo, line 224:
```suggestion
$enable-hover-utilities:       false !default;
```

scss/_utilities.scss
```diff
@@ -437,6 +442,7 @@ $utilities: map-merge(
     ),
     "text-decoration": (
       property: text-decoration,
+      hover: $enable-hover-utilites,
```
mdo, line 445:
```suggestion
      hover: $enable-hover-utilities,
```

scss/_utilities.scss
```diff
@@ -424,6 +428,7 @@ $utilities: map-merge(
     ),
     "gradient": (
       property: background-image,
+      hover: $enable-hover-utilites,
```
mdo, line 431:
```suggestion
      hover: $enable-hover-utilities,
```

scss/_utilities.scss
```diff
@@ -412,6 +415,7 @@ $utilities: map-merge(
     ),
     "background-color": (
       property: background-color,
+      hover: $enable-hover-utilites,
```
mdo, line 418:
```suggestion
      hover: $enable-hover-utilities,
```

scss/_utilities.scss
```diff
@@ -387,6 +389,7 @@ $utilities: map-merge(
     ),
     "color": (
       property: color,
+      hover: $enable-hover-utilites,
```
mdo, line 392:
```suggestion
      hover: $enable-hover-utilities,
```

scss/_utilities.scss
```diff
@@ -76,6 +77,7 @@ $utilities: map-merge(
     ),
     "border-color": (
       property: border-color,
+      hover: $enable-hover-utilites,
```
mdo, line 80:
```suggestion
      hover: $enable-hover-utilities,
```

scss/_utilities.scss
mdo, line 30:
```suggestion
      hover: $enable-hover-utilities,
```

site/content/docs/5.0/utilities/api.md
```diff
@@ -291,3 +291,39 @@ $utilities: (
...
+
+To add hover utility actions, you must add `hover: $enable-hover-utilites` to the desired utility:
```
mdo, line 297:
```suggestion
To add hover utilities, add `hover: $enable-hover-utilities` to the desired utility. This is globally set to `false` by default.
```

mdo (commented):
This could be pretty cool. Originally expected to be against this, but given it's opt-in, that seems like a powerful update for folks that want it. Leaving suggestions here that I'll batch merge after submitting this review to fix the typos.

Do we need any other controls here? Do we need a global option at all? Feels like we could skip that entirely and just have the option there for the API.

Psixodelik:
@mdo 

Thanks for fix the typos ❤️

Psixodelik:
@mdo Hi, Any news on this PR?

scss/_utilities.scss
ydmitry: Seems usage of `state` / `variant` instead of `hover: true` is more intuitive and flexible according to this issue: https://github.com/twbs/bootstrap/issues/31856

mdo:
@Psixodelik Would you be willing to revamp this to support multiple states that users could specify in a list? For example, `hover, focus, active` for `:hover`, `:focus`, and `:active`? We've had some discussion in other issues where it could be nice to enable folks to extend any utility into various pseudo-classes.

Psixodelik:
@mdo Sure. Today I will refactor the code and add the ability to specify various pseudo-classes

scss/_utilities.scss
Psixodelik: Yes, that's a good idea. thanks

Psixodelik:
@mdo Sorry, I forgot about your commit and made a force push. But everything changed there :D

scss/mixins/_utilities.scss
```diff
@@ -44,6 +47,16 @@
           #{$property}: $value if($enable-important-utilities, !important, null);
         }
       }
+
+      @if $state {
```
MartijnCuppens, line 51: Can we avoid the if test here if we use `()` as fallback in L24?

Psixodelik: This is a great idea. Thanks. I Updated PR

scss/mixins/_utilities.scss
```diff
@@ -44,6 +47,14 @@
           #{$property}: $value if($enable-important-utilities, !important, null);
         }
       }
+
+      @each $pseudo in $state {
+        .#{$property-class + $infix + $property-class-modifier}-#{$pseudo}:#{$pseudo} {
```
mdo, line 52: I think we'd want to keep the class the same, and add the pseudo states to it. Thoughts?

```suggestion
        .#{$property-class + $infix + $property-class-modifier}:#{$pseudo} {
```

site/content/docs/5.0/utilities/api.md
```diff
@@ -291,3 +291,59 @@ $utilities: (
...
+  )
+);
+```
+
+Output:
+
+```css
+.shadow-hover:hover {
+  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
+}
```
mdo, line 320: Per my above comment, I think just `.shadow:hover` makes the most sense. Here we'd be allowing folks to opt-in to any state(s) for any classes.

```suggestion
.shadow:hover {
  box-shadow: 0 0.5rem 1rem rgba(0, 0, 0, 0.15) !important;
}
```

Psixodelik: А как в таком случае задать hover? Классы shadow и псевдокласс shadow:hover никак не будут отличаться. Это не приведёт ни к какому эффекту. Например: блок с shadow-sm при hover нужно изменить на shadow-lg

```html
<div class="shadow-sm"></div>
```

```css
.shadow-sm:hover {
  /* ??? He will repeat his own styles */
}
```

Psixodelik: And it should look like this:

```html
<div class="shadow-sm shadow-lg-hover"></div>
```

or 

```html
<div class="shadow-sm shadow-hover-lg"></div>
```

MartijnCuppens: > Per my above comment, I think just `.shadow:hover` makes the most sense. Here we'd be allowing folks to opt-in to any state(s) for any classes.

This confuses me a bit @mdo. This means adding the `shadow` class would only add a shadow on hover?

mdo: Hmm might have been a brain fart from me—for some reason I was thinking of these as adding to the existing utilities, but that's not right as y'all have mentioned. So, my apologies and carry on!

Psixodelik: I would suggest generating with this pattern

```
.#{$pseudo}-#{$property-class + $infix + $property-class-modifier}:#{$pseudo}
```

This will separate the utility and its states. The user will be able to work faster with pseudo-classes.

For instance:

```
.hover-shadow-sm
.hover-text-primary
.focus-border-info
```

etc

mdo: Problem with that is it breaks our conventions elsewhere. I think we'd have to do the state at the end. Thoughts @MartijnCuppens and @ffoodd?

Psixodelik: @mdo Then I propose to leave it as it was. Classes .shadow-sm-hover etc

site/content/docs/5.0/utilities/api.md
```diff
@@ -291,3 +291,59 @@ $utilities: (
   "float": null,
 );
 ```
+
+## Adding pseudo-classes to utilities
+
+With the `state` option, pseudo-classes utility classes can be generated:
```
MartijnCuppens, line 297: `pseudo-classes utility classes` sounds weird. Maybe use `utility pseudo-classes` or `pseudo-class utilities`?

Psixodelik: «pseudo-class utilities» sounds good. Thanks

Psixodelik:
@MartijnCuppens anything to fix here?

Psixodelik:
@mdo Mark, maybe add another transition field to the API? This will make hover and focus less sharp.

It might look like this:

```scss
$utilities: (
  "shadow": (
    property: box-shadow,
    state: hover focus,
    transition: 0.7s,
    class: shadow,
    values: (
      null: $box-shadow,
      sm: $box-shadow-sm,
      lg: $box-shadow-lg,
      none: none,
    )
  )
);
```

**CSS output:**

```css
.shadow-sm {
  box-shadow: 0 0.125rem 0.25rem rgba(0, 0, 0, 0.075) !important;

  transition: 0.7s;
}
```

mdo:
I don't think we need to add transitions. Utilities are inherently meant to be single property-value pairs.

devhoussam:
Does it support all **[Pseudo-Classes]**(https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes)

I think that in **Utilities** map should make some simple changes something like this : 

remove the default **responsive** option and add it to **states** option to be like this : 

```
$utilities: () !default;
$utilities: map-merge(
  (
	"text-decoration": (
		states: ('responsive', 'hover', 'focus'),
		property: text-decoration,
		class: text,
		values: none
	),
  ),
  $utilities
);
```

mdo:
Pushed some docs updates here to match the rest of the API page. Care to review @twbs/css-review?

MartijnCuppens (approved):
Seems to work fine! I've rebased this PR and squashed some commits :shipit:

PR #31649, opened 2020-09-14
Match container gutters with row gutters

### LINKED ISSUE #31642: Container x-padding (1rem + 1rem) larger than row negative margins (-1.5rem)
I'm a little confused why the padding on containers,

```
  padding-right: 1rem;
  padding-left: 1rem;
```

is 2rem in total, but the row negative margins are -1.5rem in total
```
.row {
  --bs-gutter-x: 1.5rem;
  /* snip */
  margin-right: calc(var(--bs-gutter-x) / -2);
  margin-left: calc(var(--bs-gutter-x) / -2);
}
```

This results in 0.25rem of padding on the left and right of the containers, and makes edge-to-edge a challenge. While it doesn't matter much in the examples because they use a white background, it makes edge-to-edge backgrounds a problem. Is this intentional or a bug? Because my interpretation of the documentation is that the negative margins on the rows should exactly counteract the padding on containers. When I set --bs-gutter-x to 2rem, the problem goes away.

Thanks for all the work on Bootstrap v5! I am very excited about the stable release.

mdo:
Grid container, row, and column padding/margin should always match. I missed this in my reviews and testing, unless there's uh something else I missed. 😅

- Replaces make-container mixin's padding-x param with gutter to match naming
- Changes value of container padding variable to match grid gutter width variable
- Uses local CSS variable for container padding

Fixes #31642

/cc @MartijnCuppens directly since the original PR at #29146 changed this

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -325,7 +325,7 @@ $gutters: $spacers !default;
 
 // Container padding
 
-$container-padding-x: 1rem !default;
+$container-padding-x: $grid-gutter-width !default;
 
 
 // Components
--- a/scss/mixins/_container.scss
+++ b/scss/mixins/_container.scss
@@ -4,8 +4,8 @@
   --bs-gutter-x: #{$gutter};
 
   width: 100%;
-  padding-right: var(--bs-gutter-x);
-  padding-left: var(--bs-gutter-x);
+  padding-right: calc(var(--bs-gutter-x) / 2); // stylelint-disable-line function-blacklist
+  padding-left: calc(var(--bs-gutter-x) / 2); // stylelint-disable-line function-blacklist
   margin-right: auto;
   margin-left: auto;
 }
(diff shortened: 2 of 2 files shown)


MartijnCuppens:
I changed that in #28517 after @ysds' comment https://github.com/twbs/bootstrap/pull/28517#pullrequestreview-359350157

The paddings for containers or navbars are (imo) indeed a little too much in this PR, eg.
https://deploy-preview-31649--twbs-bootstrap.netlify.app/docs/5.0/examples/navbars/

mdo:
Oh yeah, duh, I forgot to divid it by 2 to match the `--grid-gutter-width` being divided by 2.

mdo:
Is this okay to get into an Alpha 2 in the next week or so, @twbs/css-review?

MartijnCuppens (approved):
Fine by me

PR #31667, opened 2020-09-16
Overlay form controls, selects, and files with icons or spinners

### LINKED ISSUE #28600: Spinner inside input

mdo:
This is a draft of a new set of options for our `.form-control`, `.form-select`, and `.form-file` to add icons and spinners. I've included support for sizing, too. Had to rearrange some things though, including how we're calculating the input height variables (they weren't right, now they are), but otherwise this was relatively straightforward.

Questions, ideas, and todos:

- Do we need to update this to support icons on the left and right? If so, do we use `start` and `end` over right and left in the names? This is mentioned here: https://github.com/twbs/bootstrap/pull/29107#issuecomment-575000676.
- Do we split the Sass here, perhaps with a mixin, across each file?
- Or maybe we put this as a separate page altogether?
- Any other variations we need to support? Perhaps something on focus to change the icon color?
- Do we implement this in v4 as well? Would be relatively straightforward I think, but would be a minor bump.
- Anything else?

Replaces #29107. Fixed #28600.

Screen Shot 2020-09-15 at 10 22 49 PM

Screen Shot 2020-09-15 at 10 22 55 PM

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -659,12 +659,15 @@ $input-height-inner:                    add($input-line-height * 1em, $input-pad
 $input-height-inner-half:               add($input-line-height * .5em, $input-padding-y) !default;
 $input-height-inner-quarter:            add($input-line-height * .25em, $input-padding-y / 2) !default;
 
-$input-height:                          add($input-line-height * 1em, add($input-padding-y * 2, $input-height-border, false)) !default;
-$input-height-sm:                       add($input-line-height * 1em, add($input-padding-y-sm * 2, $input-height-border, false)) !default;
-$input-height-lg:                       add($input-line-height * 1em, add($input-padding-y-lg * 2, $input-height-border, false)) !default;
+$input-height:                          add($input-font-size * $line-height-base, add($input-padding-y * 2, $input-height-border, false)) !default;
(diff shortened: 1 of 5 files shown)


ffoodd:
My two cents:
1. Not in favor of left and right icons: having a decorative icon far away from content feels useless.I n your examples, it kind of enhances the placeholder—so it follows reading direction.
  - If it finally lands, it'll need to use `start` and `end` since we'll very likely land RTL at some point.
2. A mixin might help indeed, but is not mandatory to me: this is currently self-explainatory (is that a word?).
3. A dedicated page seems a good think to me, just like layout and validation.

BTW, there's a relation between this PR and validation icons, isn't it? Should they share a few variables or something? That'd make sense.

alecpl:
Icons on the left side are quite common on logon forms. However, if I'm thinking about the loading icon, I see it on the right side, as it often would belong to a validation process.

AgentSmith0:
Is it also possible to place the overlay on the right? And is it also possible to use a link or button as overlay? Would be very nice to make a custom clear button or similar.

PR #31673, opened 2020-09-16
Dropdown: fix variable always evaluating to true

XhmikosR:


mdo:
Is there an attached issue that this fixes?

XhmikosR:
Nope, they should be identical. If is checked is falsy then it is redundant
checking for truthy values in the other branch. I just need a confirmation
:)

Johann-S (approved):
LGTM 👍

PR #31676, opened 2020-09-16
Make `.list-group-item-action` in `.list-group-item` fill their conta…

goulvench:
List-groups can either contain `<li class="list-group-item">` or links, but not both. 
The proposed change allows using either text or links in a `<ul class="list-group">`, by moving padding from the containing `<li class="list-group-item">` to the contained `<a class="list-group-item-action">` when it is the only direct child.

This allows the following HTML to display nicely:

```html
<ul class="list-group">
   <li class="list-group-item">Some text here</li>
   <li class="list-group-item">
      <a href="#" class="list-group-item-action">A link that spans the entire `<li>`</a>
   </li>
</ul>
```

I'm using this in a project, and thought it might serve others as well. 

Any improvement to make? For instance, I'm wondering if the `:only-child` selector covers all likely use cases.

As an aside, this would help with issue #21755


scss/_list-group.scss
```diff
@@ -83,6 +83,14 @@
       border-top-width: $list-group-border-width;
     }
   }
+  
+  // Make single `.list-group-item-action`s in `.list-group-item` fill their container
+  .list-group-item-action:only-child {
```
mdo, line 88: It seems unlikely that this will always be an only child. Hesitant to try to solve this problem when the mentioned issue is also already closed.

mdo (changes requested):

scss/_list-group.scss
Melearn867: Me Me go

Melearn867: Me Me go

Melearn867:
C

ffoodd:
@goulvench What about using our `.stretched-link` utility to accomplish this? [Here's a codepen demonstrating your example with and without `.stretched-link`](https://codepen.io/ffoodd/pen/mdPGNRJ).

Or you facing any constraint preventing you from using utilities here?

scss/_list-group.scss
ffoodd: @mdo Based on the description, I guess it's meant to only targets the "only direct child" case.

goulvench:
@ffoodd Using `.stretched-link` the background color is smaller than the clickable area, unless as in the Codepen the `<a>` is a direct child of `<ul>` (i.e. not wrapped in an `<li>`), which is invalid markup.

scss/_list-group.scss
goulvench: @mdo Indeed, the `:only-child` selector might cause issues, with invisible content mostly (`.sr-only`, `<span hidden>`, or `<input type="hidden">` come to mind). It would be acceptable to remove `:only-child` from the selector. Since `.list-group-item-action` do not display properly when inside `.list-group-item` I don't think they have been used, which means little chance of breaking backwards-compatibility.
More complex markup would still require using flexbox magic, as explained later in the docs.

As it stands, I believe that mixing list-items with plain text, complex layout and full-line links is quite a frequent use case. This 4-line hack-free addition would allow mixing links and text in list-groups, like in dropdowns.

If you or other maintainers believe this isn't worth it, just close the PR and I'll move on with absolutely no hard feelings. Otherwise I can rebase, lint, and include an example in the documentation, and wrap it all up with a nice red ribbon. 🎁 😉

ffoodd:
Aw you're right, didn't notice the difference.

ffoodd:
@mdo friendly ping, I find this pretty good :)

goulvench:
_On a personal note, sorry for not acting on this earlier, I've been down with the covid a few weeks, things piled up quite a bit during then._

I'm adding a note in the docs to this pull request, and testing with links, buttons, theme colors, and active/hover/disabled states, to make it work predictably. I'll update the PR when things look good enough.

goulvench:
Looks good to me now. I've checked active and disabled links, buttons, theme colors, and border-radius. 
I've included a new section in the docs to explain that links buttons and plain text can now be mixed together. 

@ffoodd, @mdo can you look at this and tell me if there's anything else to do to make this release-ready?

ffoodd (approved):
LGTM :+1: 

Deploy preview: https://deploy-preview-31676--twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#using-actionable-items-in-lists

mdo:
Crazy this has been around for two years! Revisiting it, I'm still not convinced this much duplication (from the `.list-group-item` and `.list-group-item-action`) is a good move. Instead, I'm wondering if we can find a path where we do a `<li>` with no class, and then an anchor within it with `.list-group-item` and `.list-group-item-action`. It'd require some changes for borders, but I think it'd be cleaner with less magic around negative margins and the like, too.

Going to close this out as stale, but if anyone wants to take a pass at this with fresh eyes, I'm happy to review.

PR #31679, opened 2020-09-16
Don't wrap the code snippets in our docs

mdo:
Bringing back no wrapping in our code snippets. This helps keep pages and doc snippets a little shorter on our longer pages. Found myself hitting wrapped lines with only a few characters on a new line.

**Only affects our docs.**

Preview: <https://deploy-preview-31679--twbs-bootstrap.netlify.app/>


XhmikosR:
I must be blind, but I don't see any difference?

patrickhlauke:
have to admit i'm not seeing it either, unless i'm not looking at the right place? e.g. comparing the second code block on, say, https://deploy-preview-31679--twbs-bootstrap.netlify.app/docs/5.0/components/button-group/#button-toolbar vs https://v5.getbootstrap.com/docs/5.0/components/button-group/#button-toolbar i see no difference (both bring up horizontal scrollbar?)

mdo:
This does have an affect. See the difference in this page:

- https://deploy-preview-31679--twbs-bootstrap.netlify.app/docs/5.0/getting-started/introduction/
- https://v5.getbootstrap.com/docs/5.0/getting-started/introduction/

Difference is most clear under the JS sections.

Before:

Screen Shot 2020-10-21 at 9 08 59 AM

After:

Screen Shot 2020-10-21 at 9 08 51 AM

XhmikosR:
It still looks the same for me with Firefox 82.0 and Chrome 86 on Windows :/ What is your window dimensions and browser?

ffoodd:
Same here, no change on both Firefox and Chromium on Ubuntu.

patrickhlauke:
this seems to be a Safari/macOS issue only (even Chrome/macOS currently already does the scrollbar on the live v5 docs). but can confirm now that I can see what @mdo described above, and this change indeed forces Safari to behave as intended.

XhmikosR:
OK, that makes sense since I don't own any Apple devices myself 😛

So, assuming this does not affect any other browsers and fixes the issue on Safari, it should be good to go, right?

patrickhlauke:
> So, assuming this does not affect any other browsers and fixes the issue on Safari, it should be good to go, right?

yup, i'd say so

XhmikosR (approved):

MartijnCuppens (approved):

PR #31680, opened 2020-09-16
Add stale Action for old issues

mdo:
It seems a tad silly to keep old issues around that no one has touched in years. Opening this up for debate to keep things leaner here, but I'd like to see about closing issues more than two years old that have seen no new activity.

Should we go longer? Is one week's notice enough before fully closing?

/cc @twbs/team

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,6 +10,6 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: 'This issue is stale because it has been open two years with no activity. Remove the stale label or leave a comment to keep it open, or this will be closed in 7 days.'
+        stale-issue-message: 'This issue is stale because it has been open for two years with no activity. Remove the stale label or leave a comment to keep it open, or this will be automaticaly closed in 14 days.'
         days-before-stale: 730
-        days-before-close: 7
+        days-before-close: 14
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,15 @@
...
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue is stale because it has been open two years with no activity. Remove the stale label or leave a comment to keep it open, or this will be closed in 7 days.'
```
coliff, line 13: looks like it's missing a word here:

> it has been open two years

should be:

> it has been open for two years

ffoodd:
One week notice might be short: this summer was very quiet on our side and I'm pretty sure there was a few weeks where no one would have seen notifications.

Two years old looks good to me :+1:

XhmikosR:
Personally I give it 2 weeks after I add the awaiting reply label. 1 week is too soon, and whatever we choose we might need to adapt it later anyway.

That being said, I'm not 100% sure this will help us unless I see it in action. For example, in many repos the bot just keeps closing valid issues and one has to make a new comment. This leads to extra noise, notifications etc IMHO.

mdo:
Unsure how to limit to issues only right now. Still looking through documentation for it: https://github.com/actions/stale/blob/main/action.yml.

XhmikosR:
I think we need some labels so that the bot doesn't bother with issues having them. Like `stale-ignore` or something?

We should look into https://github.com/actions/stale/blob/main/action.yml and adapt the config here.

mohe2015:
I like it that you considered the problem of issues marked as stale although they are not. But considering that you "only" have 325 issues open currently I think manual checking may be more effective. You could still close them as "stale" in the sense that it probably won't be implemented if there isn't major interest but there probably are quite a few issues that are already fixed or are extremely old browser bugs. But as you probably have a better overview of your issues I also may be wrong.

PR #31681, opened 2020-09-17
Try horizontal collapse, but not working right now

mdo:
Trying to implement https://www.codeply.com/go/UiteY92ke8/bootstrap-4-horizontal-collapse, but hitting my head against a wall for too long. Hoping someone on @twbs/css-review can point out the error of my ways.

Preview: https://deploy-preview-31681--twbs-bootstrap.netlify.app/docs/5.0/components/collapse/#horizontal


ffoodd:
Isn't some JavaScript involved here? I see you only changed the Sass part.

mdo:
> Isn't some JavaScript involved here? I see you only changed the Sass part.

Most of the guides I've seen online don't change any JS, just CSS. Not quite sure if I'm missing some changed JS in that linked example, but you could very well be right.

XhmikosR:
The demo don't seem to use any JS apart from the core classes. Not sure if we have changed those in v5, since the demo is using v4.1.3. I also tried the demo with alpha1 and seems to work fine, so I'd say something is missing in the CSS or markup?

XhmikosR:
BTW this whole thing is transitioning properties that are not performant, like width and height.

Here is a Codepen with v5.0.0-alpha2: https://codepen.io/XhmikosR/pen/wvGLgwv

Flarescape:
At the moment, the collapsible in the docs is no longer animated with the latest 4.1.x. 
https://getbootstrap.com/docs/4.1/components/collapse/
This was working in 4.0.

mdo:
I give up lol

alpadev:
> I give up lol

Nooo.. Don't!😯Let's collaborate then.

I think you don't need that visibility stuff. There are two things that are involved.

First the collapse module checks for a [`width`](https://github.com/twbs/bootstrap/blob/main/js/src/collapse.js#L284) class on the collapse element. 🤯

Second _magic_ thing is there is some hardcoded width in that demo. Not sure if it's possible to make this responsive without JS.

Also not sure why that demo also got that visibility stuff. My guess is that you can't transition `display` but `visibility`. For what reason? I don't know. If I remove that visibility stuff the demo works the same as for me.

This leaves us with something like that: https://codepen.io/-alpa-/pen/yLgObPM

PR #31684, opened 2020-09-17
Add Sass variable for CSS variable prefix

mdo:
Suggestion from @8nix at https://github.com/twbs/bootstrap/issues/26596#issuecomment-685204702.

Question though: should we include the trailing `-` in the prefix variable? Thinking we either do this, or detect if it's `nil` and remove the dash. Leaning towards the former and requiring that we include it.

Thoughts @twbs/css-review?

### DIFF
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -1,16 +1,16 @@
 :root {
   // Custom variable values only support SassScript inside `#{}`.
   @each $color, $value in $colors {
-    --bs-#{$color}: #{$value};
+    --#{$variable-prefix}#{$color}: #{$value};
   }
 
   @each $color, $value in $theme-colors {
-    --bs-#{$color}: #{$value};
+    --#{$variable-prefix}#{$color}: #{$value};
   }
 
   // Use `inspect` for lists so that quoted items keep the quotes.
   // See https://github.com/sass/sass/issues/2383#issuecomment-336349172
-  --bs-font-sans-serif: #{inspect($font-family-sans-serif)};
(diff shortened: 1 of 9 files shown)


ffoodd:
We may invert your suggestion too, adding the dash if a string is set—[here's a SassMeister](https://www.sassmeister.com/gist/75ed7c96bbd6fe7938dc1e2201fa1b4f?token=d7026bcfccd00705419abebba03c8e8d375a1b2f&scope=gist,read:user).

Not sure if it's better or not, just another possible way :)

MartijnCuppens:
> Question though: should we include the trailing - in the prefix variable? Thinking we either do this, or detect if it's nil and remove the dash. Leaning towards the former and requiring that we include it.

I'm ok with the current implementation. We'll need the prefix in other files like https://github.com/twbs/bootstrap/blob/main/scss/_tables.scss too, not sure if there are any other files (read: too lazy to look it up 😄)

mdo:
Good point @MartijnCuppens, might undo the last change here and keep it with the `-` in there given the prefix is used throughout the project. Will keep thinking! (Returning to draft mode for now)

MartijnCuppens (approved):
LGTM

XhmikosR:
Just a reminder (mostly for myself since I usually go on a merging spree 😁) some PRs target a next release like alpha3.

XhmikosR:
@mdo this needs a rebase

M055277702:
> @mdo this needs a rebase

PR #31685, opened 2020-09-17
WIP: Adding CSS variables to Pagination

mdo:
Trying out a simpler component to start the conversation around adding CSS variables for more components. Is this the right approach overall? Something else that's more pared back? How far do we take things?

Cross posting with #30500 given adding CSS vars to our buttons presents some more fun and challenging concepts to work around as we build this out.

/cc @twbs/css-review


yordis:
@mdo this will hoist the value of those variables behind that class, as long as that is what you want.

mdo:
> @mdo this will hoist the value of those variables behind that class, no!? as long as that is what you want.

I think so? I don't think component specific variables need to be root level like our global/base settings around fonts, colors, spacing, default border, etc.

yordis:
Agree, but then I think we must target the proper cascading, which I kind of like it, instead of being at the root.

ffoodd:
FWIW, the current step would only increase our file size. Pagination might not be a valuable component to massively move to custom props, but might benefit from using our globals (like colors et all).

I'll get onto this at some point, since some components would really benefit from custom props (eg. navbar themes).

mdo:
> FWIW, the current step would only increase our file size. Pagination might not be a valuable component to massively move to custom props, but might benefit from using our globals (like colors et all).
> 
> I'll get onto this at some point, since some components would really benefit from custom props (eg. navbar themes).

Yeah, that hit me while I was working on this last night. Definitely blows out file size, so optimizing around theming and what not is probably most important.

yordis:
I am not sure if file size should be a priority at the moment, CSS Variables enabled a paradigm shift, and until we get used to it, we will do things as we did before (like using classes to overwrite properties instead of variables values)

An interesting situation.

ffoodd:
File size is one of our main concern, since each byte we deliver is downloaded billions of times.

I agree with the paradigm shift, however that doesn't mean we should use custom props everywhere.

KISS, DRY and the end user first seems a better axiome to me 🙂

yordis:
Absolutely bundle size matter, my point is more about where to target the energy.

> however that doesn't mean we should use custom props everywhere.

Making the decision of where people will start disagreeing with the implementation. Maybe just start with colors so that way it is easier to swap themes, and try to avoid any other concern for now.

PR #31686, opened 2020-09-17
v5: Add disabled examples for .form-control, .form-select, and .form-range

mdo:
Forgot to include these somehow. Checkboxes, radios, and file pages already have examples. ~~Range input might need to be documented as well.~~


ffoodd (approved):
LGTM.

Wouldn't it be more accurate to add a new page dedicated to states?

mdo:
> Wouldn't it be more accurate to add a new page dedicated to states?

I think each type of form should have it's own page, but we could do some round up ones of some kind. Might explore that in the beta phase as we collect feedback.

PR #31687, opened 2020-09-17
Updated .rounded utilities

### LINKED ISSUE #29989: Confusing classes `rounded-sm` and `rounded-lg`

### LINKED ISSUE #31784: Expansion of the 'rounded' classes
As it stands we have classes which allow us to add rounding to 2 sides (via the top/bottom/left/right classes), 3 sides (by combining the 2 sided classes) and 4 sides (either through combining the 2 sided classes or just "rounded"). We can also control, to some degree, the size of the rounding (large or small).

I'm proposing an expansion of these classes to add the following options:

- Rounding on 1 corner only
Currently, this is only possible (to my knowledge) by direct CSS

- Size increments for rounding more akin to the margin/padding sizes (0-5)
This is more of a nicety and I'm aware that the small and large classes match the input controls and buttons.

- Responsive rounding (so adding the intermediate qualifier -md or -lg etc)
This is probably the trickiest to add, I don't know a whole lot about how the media queries for these are setup but I feel it would be useful to allow rounding on larger screens, where there is space to have things alongside each other, but revert to full-width stripes on smaller displays.

I realise that this is really 3 requests and that it's no small undertaking but I feel that each tweak has its own merit.

mdo:
#30048 is out of date, and I'm second guessing renaming the utilities. @ffoodd and @MartijnCuppens help me out—should we rename these classes all from `rounded` to `radius`? Should we even both changing the size scale to a numbered scale?

Fixes #29989.
Closes #30048.
Closes #31784 (classes can now be expanded and made responsive with utilities API).


ffoodd:
About size scale, I'd go for number. `-sm` et all are heavily used through utilities for responsiveness, and I think this should be considered as a convention.

Using them in components (eg buttons) is another context, another convention. We may have to document our namings somewhere, would help a lot to spot inconsistencies and name newcomers.

ffoodd:
And I'd be in favor of radius, but that doesn't seem much important IMHO.

But naming conventions (↑) could help in that sort of case (eg. follow CSS properties / HTML elements to ease discoverability).

mdo:
I think leaving it as rounded for now is fine. We can ask folks in our next release and tackle with the beta IMO. Otherwise I think this is good to go!

ffoodd (approved):

MartijnCuppens (approved):

site/content/docs/5.0/utilities/borders.md
```diff
@@ -65,15 +65,15 @@ Add classes to an element to easily round its corners.
...
 {{< /example >}}
 
+### Sizes
 
-## Sizes
+Use the scaling classes for larger or smaller rounded corners. Sizes range from `0` to `3`, and can be configured by modifying the utilities API.
 
-Use `.rounded-lg` or `.rounded-sm` for larger or smaller border-radius.
-
-{{< example class="bd-example-rounded-utils" >}}
```
XhmikosR, line 76: @mdo Just to confirm, we don't need the bd-example-rounded-utils class here?

pricop:
+1 for the -sm -md -lg. Let the numbers available for us if we want to implement rounded classes based on a percentage.

harnishdesign:
We are use radius size (`rounded-3`) scaling classes with any one of `rounded-top`, `rounded-start`, `rounded-end`, `rounded-bottom`. So, does not override border radius size scale as like `rounded-3`. it does show default rounded size.

**Does not work properly in below situation.**
Start rounded corner does in default size and end rounded corner size scale as `rounded-3`. but, We think, Here, It should Start rounded corner size scale as `rounded-3` and should end rounded corner size is 0(none).

``

Issue #31546, opened 2020-08-28
v5: <sup> <sub> inconsistent styling


misterybodon:
**Context**
* Booststrap Version: `5.0.0-alpha1/css/bootstrap.min.css`
* Mozilla FIrefox 79.0, either locally or codepen:

**CODE**
```
<ul class='breadcrumb'>
  <li class='breadcrumb-item'>
    <sub>0</sub>H
    <sup>0</sup>
  </li>
</ul>
```

* Minimal working example: https://codepen.io/misterybodon/pen/gOrWqVN

**Output**
![screen](https://user-images.githubusercontent.com/64955604/91525555-ad498a00-e8f9-11ea-9e5d-441ed90ef5ae.png)

mznet:
sub tag should be located at lower of a character and sup tag should be located at upper. like below the image.
![image](https://user-images.githubusercontent.com/9622730/91699028-a008f580-ebae-11ea-9431-a74436243275.png)

This problem can be occurred when the parent element has `display:flex` property.

mznet:
@misterybodon The inconsistency could be fixed if `align-items-baseline` class is added to the parent tag which has `display: flex`

Rachitatrey:
<sub>0</sub><var>H<sup>0</sup>
this may work you should try this

Rachitatrey:
actually i am new to it ...
so what did you ask actually about the code that symbolise H or something else ...
if no please correct me

mdo:
See #31692.

PR #29745, opened 2019-11-27
display: flex on breadcrumb-item

719media:
Fixes #29689

display: flex on breadcrumb-item

PR #31692, opened 2020-09-17
Change breadcrumb-item display to block

### LINKED ISSUE #29689: breadcrumb spacing still an issue
Breadcrumb spacing with whitespace is still an issue. Please either reopen #17473 or track here.

Thanks

mdo:
Fixes #29689 again, which was fixed in #29745 by @719media, but this introduced #31546. Undoing that last PR by setting this to `block` instead of `inline-block` (original value) or `flex` (current value).

Demo: https://codepen.io/emdeoh/pen/XWdPPxG

### DIFF
--- a/scss/_breadcrumb.scss
+++ b/scss/_breadcrumb.scss
@@ -10,7 +10,7 @@
 }
 
 .breadcrumb-item {
-  display: flex;
+  display: block; // inline-block adds white space from some HTML formatting, flex has some side effects
 
   // The separator between breadcrumbs (by default, a forward-slash: "/")
   + .breadcrumb-item {


mdo:
Slating for Alpha 3 and v4.6 (not immediate v4.5.3).

XhmikosR:
@mdo it seems the 4.6.0 project doesn't have any columns, so we might miss any projects you added?

MartijnCuppens:
`display: block` will be moot, since the breadcrumb item is a child of a flex container. Also, this PR will reintroduce #29689, made this codepen to illustrate this:

https://codepen.io/MartijnCuppens/pen/PoNyzwx

It seems a good old `float` can help us out here, opened https://github.com/twbs/bootstrap/pull/31700 to illustate this.

PR #31695, opened 2020-09-18
[SCSS] Reuse $dark variable on $table-dark-bg

andrerom:
Instead of Setting the same grey tone, reuse variable value to simplify theming.

Note: The thead-light color does not reflect `light` color too though.


andrerom:
Note: The thead-light color does not reflect `light` color too though.

XhmikosR:
Not sure if this even applies to main or if it's valid @twbs/css-review ?

MartijnCuppens:
This was already changed in v5:
https://github.com/twbs/bootstrap/blob/5bd7aa365cc268ec52fbdf4e139242a0c9769939/scss/_variables.scss#L545

In v4, this would be a BC?

XhmikosR:
Probably, yeah.

MartijnCuppens:
Closing this than.

PR #31696, opened 2020-09-18
ensure `hidePrevented.bs.modal` can be prevented

### LINKED ISSUE #31615: Bug in modal.js: defaultPrevented called on jQuery.Event instead of isDefaultPrevented()
Tried to prevent the `hidePrevented.bs.modal` event using `e.preventDefault()`, but it won't work since the code checks for `defaultPrevented` instead of `isDefaultPrevented()`. 

The bug on this line: https://github.com/twbs/bootstrap/blob/v4.5.2/js/src/modal.js#L237

All the other events are properly using `isDefaultPrevented()` see for example this line: https://github.com/twbs/bootstrap/blob/v4.5.2/js/src/modal.js#L117

Here's a simple fix: https://github.com/twbs/bootstrap/compare/v4.5.2...joakimriedel:joakimriedel/defaultPrevented?expand=1

Patch for 4.5.2 ?

joakimriedel:
Fixes #31615


js/tests/unit/modal.js
```diff
@@ -976,4 +976,42 @@ $(function () {
         backdrop: 'static'
       })
   })
+
+  QUnit.test('should get modal-static class when clicking outside of modal-content if backdrop = static', function (assert) {
```
Johann-S, line 980: this test already exist here: https://github.com/joakimriedel/bootstrap/blob/joakimriedel/defaultPrevented/js/tests/unit/modal.js#L892

You can improve it by adding your `expect`: `assert.ok($modal.hasClass('modal-static'), 'has modal-static class')`

js/tests/unit/modal.js
```diff
@@ -976,4 +976,42 @@ $(function () {
...
+      })
+  })
+
+  QUnit.test('should not get modal-static class when clicking outside of modal-content if backdrop = static and event is prevented', function (assert) {
+    assert.expect(1)
+    var done = assert.async()
+    var $modal = $('<div class="modal" data-backdrop="static"><div class="modal-dialog" style="transition-duration: 20ms;"/></div>').appendTo('#qunit-fixture')
+
+    $modal.on('hidePrevented.bs.modal', function (e) {
+      e.preventDefault()
```
Johann-S, line 1003: Adding an assert here you should be good to ensure we pass here

Johann-S (changes requested):
Thanks for your work @joakimriedel 👍 

Just a few modifications and it'll be good to go

js/tests/unit/modal.js
joakimriedel: OK, this is done. It required a modification of the old test, adding the transition-duration style. Is that OK?

js/tests/unit/modal.js
joakimriedel: Not sure what you mean by that?

Johann-S: sorry it's due to my bad english 😅 
I think you should add an assert here, to be sure this listener is called

js/tests/unit/modal.js
Johann-S: Hmm my bad, I didn't saw the `transition-duration` so you were right it should be better to split those tests

joakimriedel: OK, reverted.

js/tests/unit/modal.js
joakimriedel: Haha ok got it. Done!

joakimriedel:
Btw @Johann-S I did not commit the dist folder, assumed that is done somehow by CI?

Johann-S:
yep we build the `dist` folder before each release 👌

Johann-S (approved):

PR #31702, opened 2020-09-18
Add keyboard shortcut to focus search field

mdo:
Picks up #29834 and adds a little tag to the search input to show you can use the `/` key to focus on it. I'm not sure what bugs @MartijnCuppens was seeing based on his comment there, but this seems to work great for me locally.

Screen Shot 2020-09-18 at 4 18 48 PM

Preview: <https://deploy-preview-31702--twbs-bootstrap.netlify.app/docs/5.0/getting-started/introduction/>


site/assets/js/search.js
```diff
@@ -12,6 +12,14 @@
...
   var siteDocsVersion = inputElement.getAttribute('data-docs-version')
 
+  document.addEventListener('keydown', function (event) {
+    if (event.key === '/') {
+      event.preventDefault()
+
+      inputElement.focus()
+    }
+  })
+
```
chuckrincon, line 22: Actually this code is not necessary anymore, because algolia uses [`autocompleteOptions`](https://docsearch.algolia.com/docs/behavior/#autocompleteoptions), now we can set the keycode value of `/` to handle this behaviour using the `keyboardShortcuts` option.

When I opened #29834 I forgot to check this in the `docsearch` documentation

patrickhlauke:
I would advise against this, as it would fail https://www.w3.org/TR/WCAG21/#character-key-shortcuts

site/assets/js/search.js
chuckrincon: Adding this options to the `window.docsearch` should be do the work:
```js
window.docsearch({
  […],
  autocompleteOptions: {
    keyboardShortcuts: [111]
  }
})
```
Where `111` is the keycode value of `/` key

mdo:
Oh boy, didn't know there were guidelines for that. Wonder if the built-in Algolia feature mentioned helps with that at all. Also wonder if GitHub does anything like this for their shortcuts.

patrickhlauke:
Using an additional modifier key like `ALT + /` or `CTRL + /` would pass the WCAG SC and may be worth considering.

MartijnCuppens:
> I'm not sure what bugs @MartijnCuppens was seeing based on his comment there

If an input or textarea is focused for example on the [form controls page](https://deploy-preview-31702--twbs-bootstrap.netlify.app/docs/5.0/forms/form-control/#example), the focus jumps to the search bar if `/` is typed.

We could make use of the native [accesskey](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey) attribute, but I think it won't be clear enough which keyboard combo needs to be pressed.

patrickhlauke:
`accesskey` has historically seen very little use because of issues of discoverability, and because it varied between browsers which modifier key you'd use to trigger them (so sites shied away from providing explicit hints like `ALT + /` or similar, because that may not match the user's browser/platform).

and yes, if you just register an event listener on the page as a whole, it will kick in when it shouldn't (when user's focus is actually on an input, textarea, select, etc), so would generally need extra checks in the code about "if any of these types of element are currently the `document.activeElement`, just bail out and do nothing)"

mdo:
Pushed an update here to modify the shortcut from `/` to `Ctrl /`. I couldn't figure out how to do multiple keys in Algolia's autocomplete options, so I've stuck with the approach I started with. Thoughts?

patrickhlauke:
> I couldn't figure out how to do multiple keys in Algolia's autocomplete options

if you mean the actual "once the search is focused, pressing normal keys triggers autocomplete suggestions", then that's kosher per the WCAG criterion, as you're not triggering actual functionality. in short, this now looks good :shipit:

patrickhlauke (approved):

XhmikosR:
Please give me some time to look into this and get back to you

XhmikosR:
Not sure what's left here, it seems to work fine like @patrickhlauke said?

XhmikosR (approved):

einfallstoll:
Nice this even works with `Ctrl + Shift + 7` (if `/` is created by typing `Shift + 7`).

PR #31708, opened 2020-09-19
feat(table): variables strategy change - SCSS and CSS variables

ydmitry:
PoC for suggestion from this issue https://github.com/twbs/bootstrap/issues/26596#issuecomment-694214372 

It adds some rules for variables usage:
1. In SCSS styles files (e.g. `_table.scss`) we use only SCSS variables in style properties
   - It makes SCSS style more clean, so that we don't have mix of CSS variables and SCSS variables together here
   - Simplier to work with this file as we know that we need to work with SCSS variables only
   - Allows to remove properties from CSS output together with selectors by setting `null` to variables which can reduce output CSS file size
   - Depend on application requirements library user can select which to use SCSS variables or CSS variables
   - "Use var() in Sass variables and not directly in components" (c) ffoodd https://github.com/twbs/bootstrap/issues/26596#issuecomment-695957040 
2. In SCSS variables file (`_variables.scss`) we can set CSS variables as a value, e.g. `$x: var(--x);`
   - This creates flexibility and provides example how to create variants by font size or border width, was asked in this comment https://github.com/twbs/bootstrap/issues/26596#issuecomment-607125580
   - We can set not only CSS variables, but formulas of CSS variable so for easier customization it will be enough to operate with one variable
   - Possible here to override on application side:
     * specific value, e.g. `$table-bg: $red;`
     * reference to another CSS variable `$table-bg: var(--primary-color);`
     * set to null `$table-bg: null;` to remove property together with CSS variable declaration `$table-bg-var: null;`
     * set to custom expression of CSS other variables `$table: hsl(var(--color),var(--l));`
3. We use CSS custom properties together with `$enable-custom-properties` variable in SCSS styles file (e.g. `_table.scss`)
   - It makes backport to IE11 easier by replacing SCSS variables, maybe without all color variants, but they can be created on application side using good SCSS styles file (e.g. `_table.scss`)
   - Easier to reason about what exactly CSS variable will change
4. We generate variants (modifications) using CSS variables with modification classname


scss/_tables.scss
```diff
@@ -25,8 +26,7 @@
   // stylelint-disable-next-line selector-max-universal
   > :not(caption) > * > * {
     padding: $table-cell-padding-y $table-cell-padding-x;
-    background-color: var(--bs-table-bg);
-    background-image: linear-gradient(var(--bs-table-accent-bg), var(--bs-table-accent-bg));
+    background-color: $table-bg;
```
ydmitry, line 29: In this file we use only SCSS variables in style properties values, every - without expressions, just variables.

scss/_tables.scss
```diff
@@ -3,14 +3,15 @@
...
-  --bs-table-hover-color: #{$table-hover-color};
-  --bs-table-hover-bg: #{$table-hover-bg};
+  --bs-table-bg: transparent;
+  --bs-table-color: #{$body-color};
+  --bs-table-border-color: #{$border-color};
+  --bs-table-striped-bg: #{rgba($black, $table-striped-bg-factor)};
+  --bs-table-striped-color: #{$table-color};
+  --bs-table-active-bg: #{rgba($black, $table-active-bg-factor)};
+  --bs-table-active-color: #{$table-color};
+  --bs-table-hover-bg: #{rgba($black, $table-hover-bg-factor)};
```
ydmitry, line 13: We can move this to `mixins/_table-variants.scss` file and enable opt-out from declaring these values

scss/_variables.scss
```diff
@@ -492,26 +492,26 @@ $table-cell-padding-x-sm:     .25rem !default;
 
 $table-cell-vertical-align:   top !default;
 
-$table-color:                 $body-color !default;
-$table-bg:                    transparent !default;
+$table-color:                 var(--bs-table-color) !default;
+$table-bg:                    var(--bs-table-bg) !default;
```
ydmitry, line 496: In SCSS variables list we set reference to a CSS variable, possible here to override on application side:
* specific value, e.g. `$table-bg: $red;`
* reference to another CSS variable `$table-bg: var(--primary-color);`
* set to null `$table-bg: null;`
* set to custom expression of CSS other variables `$table: hsl(var(--color),var(--l));`

scss/_variables.scss
```diff
@@ -492,26 +492,26 @@ $table-cell-padding-x-sm:     .25rem !default;
...
 
 $table-th-font-weight:        null !default;
```
ydmitry, line 498: This can be set a CSS variable on application side by users, so that they can create own variants.

scss/_tables.scss
```diff
@@ -99,8 +99,10 @@
 
 .table-striped {
   > tbody > tr:nth-of-type(#{$table-striped-order}) {
-    --bs-table-accent-bg: var(--bs-table-striped-bg);
-    color: var(--bs-table-striped-color);
+    color: $table-striped-color;
+    > * {
+      background-image: linear-gradient($table-striped-bg, $table-striped-bg);
+    }
```
ydmitry, line 105: We avoid here usage of custom properties. It is only possible during variants / modifiers generation. So that we can backport to IE11 here (except variants)

Also it is easier to reason about:
We had `--bs-table-accent-bg: var(--bs-table-striped-bg);` not clear what it changes, it changes accent color ? Need to Ctrl + F what exactly it will change and it will be easy to make a mistake when we will have more variables and more selectors

As a downside - yes, more code

ydmitry:
@ffoodd  Added a new commit after your comment suggestion https://github.com/twbs/bootstrap/issues/26596#issuecomment-695957040 

It also adds now variables `$table-bg-var` together with `$table-bg` to see all values in one place.
We can go to non-CSS-vars mode for specific properties by setting e.g.:
```
$table-bg-var: null; // You don't need CSS variable declaration now
$table-bg: red;
```

MartijnCuppens:
Hi @ydmitry,

Thanks for your work, but I don't think we'll be able to do anything with this.

- I'm not sure what the benefit of this is.
- There is no need for a `$enable-custom-properties` variable. We are already using custom properties in v5.
- If you want IE support, use Bootstrap 4. We don't have any intention to make v5 work on IE (we won't be able to do some fancy stuff)

ydmitry:
> Hi @ydmitry,
> 
> Thanks for your work, but I don't think we'll be able to do anything with this.
> 
> * I'm not sure what the benefit of this is.
> * There is no need for a `$enable-custom-properties` variable. We are already using custom properties in v5.
> * If you want IE support, use Bootstrap 4. We don't have any intention to make v5 work on IE (we won't be able to do some fancy stuff)
> 
> Edit: I just read [#26596 (comment)](https://github.com/twbs/bootstrap/issues/26596#issuecomment-695957040), so I just reopened this.

Yes, there are two suggestions actually:
1. Use only SCSS variables in SCSS files to give more flexibility in terms of customization and CSS variables usage
2. Provide backport for basic functionality for IE11, not full support of all variants, maybe with changing some SCSS variables.

ydmitry:
Added change for initial values for variables, now it uses CSS variables fallbacks:
```scss
$table-color:                 var(--bs-table-color, $body-color) !default;
```
Instead of assigning initial value via custom property:
```scss
--bs-table-color: #{$table-color-var};
```

ffoodd:
FWIW, I don't think`$enable-custom-properties` has any sense: by simply using Sass variables in components and using `var()` in Sass variables—as you suggested—it only requires to change those Sass variables to drop custom properties.

And custom properties declarations won't hurt at all, so there's no reason to output them conditionnally. This is already the case in v4, BTW.

I think it's very smart to set `var()` in Sass variables since it allows one to completely drop them, but no need for more complexity.

If @twbs/css-review find this interesting, we migth even add a stylelint rule to prevent `var()` usage outside `_variables.scss`—but even if I find it interesting, that'd be a paradigm shift that may bother us later…

So still not sure about this. Any thoughts welcome :)

mdo:

ydmitry:
@mdo hello
I would say this PR is more related to CSS variables, please check @ffoodd comments (it's about having flexibility if I want to add new CSS variables on app side or switch to plain value from CSS variable or null it)

ydmitry:
@ffoodd  well, users who removes some `var()` from variables should understand what they do it can be bad idea to replace `var()` with simple value in some cases, but can be very useful if I want to use CSS expression with CSS variables instead of single CSS variable. So, in terms of variants generation case this change will be still useful. It is just won't be recommended to change any variable to any value

It can be rare cases, but if application will need it - if no variable or some flexibility , they will have to overwrite Bootstrap styles or create a fork

ffoodd:
Indeed, then having us suporting this is almost useless, isn't it?

ydmitry:
@ffoodd well, when no need to overwrite styles on application side this is much better experience, they will need to support updates from Bootstrap. Multiply this on many themes in some cases and get huge cost for Bootstrap customization. So, better to have ability to change what needed with with variables

ydmitry:
@ffoodd if we check main thread about CSS variables we can notice that many people want CSS variables to switch whole themes at runtime. Current usage of CSS variables in Bootstrap more focused on refactor components to have less code and to generate variants. If we follow rule from this PR they can achieve "runtime theme switching" with corresponding change of SCSS variables.

Issue #29853, opened 2019-12-14
Switch from Sass imports to Sass Module System


inwardmovement:
Just putting this here, maybe for v5.

[Sass Module System](http://sass.logdown.com/posts/7858341-the-module-system-is-launched)

MartijnCuppens:
AFAICT, the module system only works in Dart Sass, not in ruby Sass or Lib Sass. Since the majority of build systems won't support this yet, I guess it won't be a good idea to switch now.

mocsy:
What are the 'majority of build systems' you mention?
I had no problems with webpack, parcel nor gulp using the [sass](https://www.npmjs.com/package/sass) package.

Maybe the right approach would be to migrate from node-sass to sass first.
The module system comes as a free bonus with the [migrator](https://sass-lang.com/documentation/cli/migrator) anyway.

That might clear these warning from my build logs as another added bonus:

```
WARNING: You probably don't mean to use the color value indigo in interpolation here.
It may end up represented as indigo, which will likely produce invalid CSS.
Always quote color names when using them as strings or map keys (for example, "indigo").
If you really want to use the color value here, use '"" + $color'.

  ╷
4 │     --#{$color}: #{$value};
  │         ^^^^^^
  ╵
    node_modules/bootstrap/scss/_root.scss 4:9       @import
    node_modules/bootstrap/scss/bootstrap.scss 11:9  @import
```

MartijnCuppens:
@mocsy, `node-sass` is used more to compile sass, see https://npmcharts.com/compare/sass,node-sass?interval=30

Other than that, we have a list of other compilers using LibSass: https://sass-lang.com/libsass

Even if we switch to `sass` instead of `node-sass` (which we might do at some point), we still need to support `node-sass` as long as it is not deprecated.

glen-84:
LibSass: https://github.com/sass/libsass/issues/2807
Ruby Sass: EOL

zackdotcomputer:
Just hopping in here to +1 this, especially since [SASS used Bootstrap as an example](https://sass-lang.com/blog/the-module-system-is-launched) for how the new package system could be used.

jquense:
[...]

PR #31728, opened 2020-09-23
Switch to dart-sass

XhmikosR:
This is something we've been discussing internally for some time now.

Things to note:

1. dart-sass's precision is hardcoded to 10; I dropped it down to 6 for the minified files, but we might want to reconsider this and just go with the default? I need to check how much it adds, though
2. dart-sass orders grouped selectors differently; I have not found a way to work around this and perhaps we should ask upstream. Do note that the minified files keep the old behavior since clean-css sorts the grouped selectors alphabetically
3. dart-sass converts escaped UTF-8 characters which in turn means we need a `@charset "UTF-8";`. Not the end of the world, but it is something to note. Also, note that there's a `--no-charset` flag, but then this could be dangerous since we do have Unicode characters, so we better not use it
4. dart-sass removes quotes from attribute selectors; not a big deal since they are already removed in the minified files
5. dart-sass adds leading zeros; again, not a big deal since they are already removed in the minified files
6. other minor output differences like whitespaces

@twbs/css-review for discussion

----

1. Check the bundlewatch size differences
2. Live preview: <https://deploy-preview-31728--twbs-bootstrap.netlify.app/>
3. Test that sourcemaps work properly; I see some changes in soucemaps' `sources` property

Refs #29853


dist/css/bootstrap.css
```diff
@@ -2320,42 +2468,35 @@ progress {
...
-
 .form-file-lg .form-file-text,
 .form-file-lg .form-file-button {
   padding: 0.5rem 1rem;
 }
 
 .form-range {
   width: 100%;
-  height: 1.4rem;
+  height: calc(1rem + 0.5em);
```
MartijnCuppens, line 2478: Caused by https://github.com/twbs/bootstrap/commit/1af79bbc5267428af9fcb4a8fc9e6ab0b30328f4. We might need to look into the other effects of the unit change here (not related to this PR btw)

XhmikosR: Yeah, that's why I decided to split the dist commits so that we can spot the dart-sass only changes.

MartijnCuppens: Opened https://github.com/twbs/bootstrap/pull/31746 to fix the unit

mdo (approved):
Pre-approving for when this comes up for review.

XhmikosR:
Let's land it as is and we can always revisit the clean-css options later. BTW we could also ask upstream for the output discrepancies and if they can do something about it, like with sorting grouped selectors.

PR #31729, opened 2020-09-23
Input group addon sass customization options

MartijnCuppens:
Needed these options today. This PR just adds some customization options in Sass.


ffoodd (approved):
LGTM

PR #31740, opened 2020-09-24
Stub out a `.nav-scroll` component for horizontal scrolling navs

mdo:
This is a WIP that takes some of the nav styles we've previously used in our docs navbar and the Offcanvas navigation, and puts some enhancements into it for wider use. I'm sure there's something problematic about this, but feels like it could be a nice little addition to our navs.

Thoughts @twbs/css-review?

Screen Shot 2020-09-24 at 9 27 45 AM

Preview: <https://deploy-preview-31740--twbs-bootstrap.netlify.app/docs/5.0/components/navs/#nav-scroll>


XhmikosR:
How is this supposed to work with mouse devices? I see no horizontal scrollbar?

mdo:
> How is this supposed to work with mouse devices? I see no horizontal scrollbar?

I've hidden it entirely on purpose. You can see something similar on the Offcanvas example added in v4.

XhmikosR:
I don't think this is a good change. The component is unusable on desktop
without touch support... Maybe Patrick can chime in in case I'm missing
something :)

PR #31741, opened 2020-09-24
Change to "pointing hand" cursor on hover

Thor-x86:
Improves user experience (UX) by adding "pointing hand" cursor on hover. So it looks more clickable even though it functioning properly.

### DIFF
--- a/scss/forms/_form-check.scss
+++ b/scss/forms/_form-check.scss
@@ -19,6 +19,7 @@
 .form-check-input {
   width: $form-check-input-width;
   height: $form-check-input-width;
+  cursor: pointer;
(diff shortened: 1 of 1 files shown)


Thor-x86:
The lint on my machine shows no problem. However, it always failing on lint check at GitHub. Did I missed something on lint configuration?

patrickhlauke:
this has been debated back and forth forever. https://github.com/twbs/bootstrap/issues?q=is%3Aissue+cursor+pointer

there's now a utlity class for those who want it, but not something we're going to do by default I'm afraid https://github.com/twbs/bootstrap/pull/30562

PR #31742, opened 2020-09-24
Fix bundlewatch size

mdo:


mdo:
Merging right away without review knowing full well this number could change, and to fix `main`. WIth the Alpha 3 plans, we'll be adding a bit more here, hence why I went up by `.5` instead of something less.

PR #31744, opened 2020-09-24
WIP: Add dependencies to our Sass files

mdo:


MartijnCuppens:
`_variables.scss` depends on `_functions.scss`, so these two should be switched

celestehorgan:
not what I envisioned when I opened an issue, but this would satisfy my needs as a user. Thank you!

XhmikosR:
@mdo I suppose we can close this? We could probably do something like you did on main in a new PR.

mdo:
Closing as stale.

Issue #31402, opened 2020-08-02
Include jQuery examples as well


igll:
It would be great to have all examples of plain JavaScript code with **their jQuery equivalent**. E.g. at https://v5.getbootstrap.com/docs/5.0/components/navs/#via-javascript

At least at https://v5.getbootstrap.com/docs/5.0/getting-started/javascript/#still-want-to-use-jquery-its-possible
showing in parallel the jQuery code with the equivalent plain JavaScript code.

JFYI: I was not able to get `$('[data-toggle="tooltip"]').tooltip()"` working because I don't see its connection with the code at https://v5.getbootstrap.com/docs/5.0/components/tooltips/#example-enable-tooltips-everywhere

divya661:
@XhmikosR  @igll I would like to work on it. Can I be assigned to this task?

juniorRubyist:
I have an example for the instance in question at https://v5.getbootstrap.com/docs/5.0/components/navs/#via-javascript, so I'll make a PR.

XhmikosR:
After discussing this with @mdo on Slack, we'd prefer to have a separate section in docs with a couple of examples and link to it from each component. We believe it's cleaner this way than adding a jQuery example in each component.

So, perhaps a new partial linking to the javascript examples.

sdkdeepa:
Hello! I would like to contribute to this as it is still open. Please let me know

GeoSot:
Based on this documentation section https://getbootstrap.com/docs/5.2/getting-started/javascript/#optionally-using-jquery, I suppose we can close this issue

/cc @mdo

PR #31751, opened 2020-09-25
Example of using jQuery with tab navs

josephgeis:
Contributes to #31402. Gives example of using jQuery for activating tabbable tabs.


site/content/docs/5.0/components/navs.md
```diff
@@ -513,6 +513,18 @@ You can activate a tab or pill navigation without writing any JavaScript by simp
 </div>
 {{< /highlight >}}
 
+### Via jQuery
```
XhmikosR, line 516: Not sure if this should be here and not an `h4` after JavaScript, since jQuery belongs to JavaScript. /CC @mdo

juniorRubyist: That makes a little more sense. I just thought that "Via JavaScript" implies pure JS.

XhmikosR: Yeah, let's move it below JavaScript. jQuery is JavaScript too.

mdo:
Should this be unique to the Navs? Or should we be including guidance elsewhere for an overall primer on using jQuery w/ Bootstrap?

XhmikosR:
Yup, agreed. We need maybe a new partial/callout and include it or we just add it in the JavaScript page and link to it from each component.

XhmikosR:
Closing since we need to add a common section and reference it from each component. Regardless, thanks for your contribution!

PR #31754, opened 2020-09-26
Tweak versions page columns

XhmikosR:
Not sure if we want something like this and if we should backport it. I haven't added a project but feel free to change it.

Preview: https://deploy-preview-31754--twbs-bootstrap.netlify.app/docs/versions/


mdo:
> Not sure if we want something like this

Definitely want it :)

mdo (approved):

PR #31755, opened 2020-09-26
Add a `gh` shortcode

XhmikosR:
WIP; I might end up changing the shortcode to accept named parameters too.


ffoodd:
Very nice idea, it'll help a lot!

mdo:
Very into it :)

XhmikosR:
All right, I'll update the docs to use the shortcode as is, and we can always tweak it later :)

XhmikosR:
After working on this a bit, I'm not sure having one shortcode is better. I'll revisit this, maybe by separating the shortcodes or going back to the original implementation, which has its limitations, but it's shorter to type.

mdo:
Removing from v5.0.0 for now to help prioritize.

XhmikosR:
Going to drop this. If anyone wants to take this up, feel free to CC me :)

PR #31756, opened 2020-09-26
Add Popover variables for derived properties

ydmitry:
Part of solution for https://github.com/twbs/bootstrap/issues/31538 for Popover


mdo:

PR #31761, opened 2020-09-27
Move homepage-hero.svg to an img tag

XhmikosR:
While testing our main branch (<https://twbs-bootstrap.netlify.app>) on Firefox for Android, I noticed that when the docs menu is opening/closing, the transition is very sluggish. Chrome has better performance, but it's still a little sluggish. Now, the device I use is not a low end I believe, Samsung Galaxy S9, and I also tested out a Galaxy S20, both with Firefox and Chrome.

I believe this fixes the issue since it reduces layout reflows. As a bonus, the image will be cached. I used an empty `alt` attribute since I believe the image is decorative only, but @patrickhlauke let me know if something else is needed.

Note that we didn't see the issue before because we were using a horizontal navbar approach.

@mdo I added this in the alpha2 project because its effect is pretty big.

Preview: <https://deploy-preview-31761--twbs-bootstrap.netlify.app/>


XhmikosR:
I made 2 recordings but apparently the issue is not so visible there so the following is a little moot. The problem does exist in real life, though:

| Main | This branch |
| ------ | ----- |
| 1-main | 2-new |

mdo (approved):
Awesome work, and great PR presenting the difference! <3

mdo:
We might need to revert or at least revisit this—the homepage image is all blurry in Safari (mobile and desktop). I didn't notice it until just tonight.

XhmikosR:
I don't have a machine with Safari, but can you try locally with just reverting the optimize patch? Also, can you try with the original dimensions in the `img` tag?

mdo:
Tried dimensions changes, but no difference. Opening the new image in a new tab, there's no issue. Seems to be because SVGs in an img tag render the SVG as a bitmap vs actually vector.

XhmikosR:
Sounds like a browser bug for sure... I don't mind reverting this temporarily, but keep in mind that the previous solution is too CPU hungry and low to mid devices will have a pretty poor experience.

XhmikosR:
* https://stackoverflow.com/questions/43546754/svg-icons-in-safari-are-blurred
* https://stackoverflow.com/questions/24439705/svgs-are-displaying-blurry-in-safari
* https://github.com/airbnb/lottie-web/issues/1976

Someone should really report this bug. That being said, please try and see if we can work around it (I can't test it myself since I don't own a device with Safari), otherwise let me know and I will revert it.

PR #31768, opened 2020-09-27
Update LICENSE

warrenharper28:


mdo:
Without any description or context, we’ll be skipping this. Thanks!

PR #31748, opened 2020-09-25
Release v5.0.0-alpha2.

XhmikosR:
Preview: <https://deploy-preview-31748--twbs-bootstrap.netlify.app/>

PR #31770, opened 2020-09-28
Updated version in ESM and UMD files

mattonit:
On #31748 @XhmikosR hasn't updated the version in js/index.esm.js and js/index.umd.js files so I did it myself. If this was intentional, please close this PR


XhmikosR:
Thanks but the PR isn't ready yet.

PR #31774, opened 2020-09-28
Replace ellipsis Unicode symbol with three dots.

XhmikosR:


ffoodd (approved):

martinbean:
What was the reason for this?

XhmikosR:
Consistency and I personally see no reason to have the UTF-8 character in those places.

mateusfccp:
I personally see this as a regression, as the ellipsis unicode char has the exact semantic of that what we are used to.

> Ellipses can indicate an unfinished thought, a leading statement, a slight pause, an echoing voice, or a nervous or awkward silence. Aposiopesis is the use of an ellipsis to trail off into silence—for example: "But I thought he was …" When placed at the beginning or end of a sentence, the ellipsis can also inspire a feeling of melancholy or longing.

Three consecutive dots have no defined semantics.

martinbean:
@mateusfccp Yeah, agreed. It seems bizarre to go from the typographical correct glyph, to the “dumb” equivalent.

benabraham:
On console output it‘s understandable.

In documentation it‘s plain wrong.

ffoodd:
Code example should probably not use the ellipsis character, should they?
While I mostly agree for the typographic concern in the docs, there're plenty of "three dots" and a single ellipsis used appropriately. You may still open a PR for better typography in our odcs, but that's definetely not a priority ATM (while consistency is).

patrickhlauke:
purely selfishly...on windows, I find it a pain to type "proper" ellipses. FWIW

benabraham:
@ffoodd definitely not the most important problem to spend time on.

Thank you for your work.

vanillajonathan:
I think … (ellipsis) ought to preferred over ... (three period characters) when appropriate just as – (en dash) and — (em dash) ought to preferred over - (hyphen) when appropriate.

martinbean:
I don’t know about you chaps, but my files are saved with UTF-8 encoding in my text editor, and my stock terminal uses UTF-8 encoding. So I’m not sure why people are hesitant to use appropriate UTF-8-encoded glyphs for typographic elements?

I guess I’ll be in the minority when I say I use “smart” apostrophes in my code comments as well, now 🙂

ffoodd:
No one hesitates here: 

1. we have other topics to cover, which are much more valuable for Bootstrap as a product than appropriate typograpy in our docs, obviously;
2. as Patrick mentions, using appropriate typographic glyphs isn't trivial on Windows.

We're still opened to PR, if one have time for this 🙂

PR #31786, opened 2020-09-28
Add gradient buttons

techboyg5:
This PR references gradient buttons in the docs.

[Demo](https://deploy-preview-31786--twbs-bootstrap.netlify.app/docs/5.0/components/buttons/#gradient-buttons)


mdo:
Buttons have their own gradients when `$enable-gradients` is enabled. Using a single color background utility isn't the best option since hover, focus, and active states would be affected.

techboyg5:
> 
> 
> Buttons have their own gradients when `$enable-gradients` is enabled. Using a single color background utility isn't the best option since hover, focus, and active states would be affected.

But most people just use the dist css, with no option for gradients.

PR #31727, opened 2020-09-23
Fix text-break in IE and Edge Legacy

ihorzenich:
Fix #29319 again.
Correcting Backport #30932 to v4

`word-break` didn't support `break-word` value in IE and Edge Legacy 
`overflow-wrap` is absent in IE/Edge Legacy (it's modern replace for `word-wrap`)
So we need to add back `word-wrap: break-word` for IE and Edge Legacy.

PR #29714, opened 2019-11-22
Keep themed appearance for checkboxes when page is printed

MartijnCuppens:
Before: https://twbs-bootstrap.netlify.com/docs/4.3/forms/checks/

image

After: https://deploy-preview-29714--twbs-bootstrap.netlify.com/docs/4.3/forms/checks/

image

Works in Chrome, FF & Chromium Edge. The themed checkboxes don't apply to EdgeHTML & IE in `v5`, these will have the native fallback.

PR #31793, opened 2020-09-29
Misc v4 backports

mdo:
- #30781: Reboot's th updates
- #29714: Keep custom check, radio, and switch theme when printing
- #31754: Improve versions page rendering (also reversed the order while I was here)
- #30932: `.text-break` changes to drop `overflow-wrap` and use `word-wrap` once again (closes #31727)

### DIFF
--- a/site/docs/versions.html
+++ b/site/docs/versions.html
@@ -5,8 +5,9 @@ description: An appendix of hosted documentation for nearly every release of Boo
 ---
 
 <div class="row">
-{% for release in site.data.docs-versions %}
-  <div class="col-md">
+{% assign releases = site.data.docs-versions | reverse %}
+{% for release in releases %}
+  <div class="col-md-6 col-lg-4 col-xl mb-4">
     <h2>{{ release.group }}</h2>
     <p>{{ release.description }}</p>
     {% assign versions = release.versions | reverse %}
--- a/scss/_custom-forms.scss
+++ b/scss/_custom-forms.scss
@@ -13,6 +13,7 @@
   display: block;
   min-height: $font-size-base * $line-height-base;
   padding-left: $custom-control-gutter + $custom-control-indicator-size;
+  color-adjust: exact; // Keep themed appearance for print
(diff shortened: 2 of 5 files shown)


site/docs/versions.html
```diff
@@ -5,8 +5,9 @@
 ---
 
 <div class="row">
-{% for release in site.data.docs-versions %}
-  <div class="col-md">
+{% assign releases = site.data.docs-versions | reverse %}
```
XhmikosR, line 8: This change is different than #31754

XhmikosR: NVM, I saw the the OP. maybe change the commit comment to reflect this?

mdo: PR reflects it, that good enough? Can also edit it when squashing and merging.

XhmikosR: The referenced PR doesn't include the reverse change, hence why I suggest to edit the commit message. Also, please do not squash when merging; these patches should be separate commits since they refer to different changes :)

mdo:
Updated @XhmikosR, should be good to go now!

XhmikosR (approved):
LGTM but I'd like another approval from @twbs/css-review for the Sass changes :)

ihorzenich:
Please use comments for `.text-break` from this commit: https://github.com/twbs/bootstrap/pull/31727/commits/e501897af6ae0383b0b4aade8c76468d0ca6fe00 for right correct explanation what are they doing.

XhmikosR:
They are too verbose and not really needed.

ihorzenich:
> They are too verbose and not really needed.
But then at least remove current `// IE & < Edge 18` comment because it confuse (see: https://github.com/twbs/bootstrap/commit/0783f638491c320df02aefb2167ead158712e80f#r42942080)

XhmikosR:
Friendly ping @twbs/css-review

ffoodd (approved):
LGTM

PR #31798, opened 2020-09-30
Add some words in security reporting

Rakesh-kumar-yadav:


mdo:
Thanks but this is clearer and more succinct as is.

PR #31802, opened 2020-10-01
Improve callout shortcode.

XhmikosR:
Remove the `markdownify` call, and instead rely on Hugo's proper syntax; `{{% callout %}}` when we want to the content to be rendered.

This allows for stuff like:

````
{{% callout info %}}
##### I'm an info callout!

````
```css
.foo {
  color: #fff;
}
```
````

{{< example >}}
<div class="alert alert-warning alert-dismissible fade show" role="alert">
  <strong>Holy guacamole!</strong> You should check in on some of those fields below.
  <button type="button" class="btn-close" data-dismiss="alert" aria-label="Close"></button>
</div>
{{< /example >}}
{{% /callout %}}
````

---

Preview: <https://deploy-preview-31802--twbs-bootstrap.netlify.app/>

### DIFF
--- a/site/content/docs/5.0/components/alerts.md
+++ b/site/content/docs/5.0/components/alerts.md
@@ -22,9 +22,9 @@ Alerts are available for any length of text, as well as an optional close button
 {{< /alerts.inline >}}
 {{< /example >}}
 
-{{< callout info >}}
+{{% callout info %}}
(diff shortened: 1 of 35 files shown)


XhmikosR:
@Johann-S @MartijnCuppens @mdo please try this; it seems to be the proper way AFAICT :)

MartijnCuppens (approved):

PR #31806, opened 2020-10-01
docs: switch to fenced codeblocks

XhmikosR:
Opening to hear your thoughts, but along with #31802 this might be better since most people should already be familiar with this syntax. As a plus, this works on GitHub too.


patrickhlauke:
i'd be in favor of this, assuming the highlighting then still works

mdo:
Nice, definitely into it.

XhmikosR:
This was done with a regex, and there were a few cases with one liners that broke and I had to fix manually, but apart from that everything else should be good to go.

patrickhlauke (approved):

PR #31810, opened 2020-10-01
removed unnecessary square brackets

krishnadevz:
I removed unnecessary square brackets in the becoming a backer line


therealklanni:
The brackets are there for a reason, it's not a typo/mistake.

PR #31813, opened 2020-10-01
Add optional CSS grid

### LINKED ISSUE #19729: W3C CSS Grids
Hey,

I was just introduce to this W3C document for implementing grids natively in CSS and was a bit surprised that there doesn't seem to be anyone from this community as a collaborator for this document.

https://www.w3.org/TR/css-grid-1/

I know that this might compete somewhat with this project, but if it's going to be a recommendation, I would prefer that it gets input from a community that has been doing grids for years.

### LINKED ISSUE #23057: Bootstrap css grid : Replace flexbox with CSS Grid Layout?
Any plans to use css grid over flexbox for the bootstrap grid?

The support seems pretty good : 
http://caniuse.com/#feat=css-grid

### LINKED ISSUE #24887: Ideas for BootStrap v5: Use CSS Grid Layout
Hello,

Bootstrap 5 may be rely on the new standard CSS Grid layout which allow to divide available space for layout into columns and rows:

https://caniuse.com/#feat=css-grid

mdo:
Stubbing out here for fun since #31812 as opened. I’ve done some experimenting elsewhere for this, and I’m not 100% convinced we’d want to drop the existing flex box grid for a CSS grid implementation. However, I think it’s important to at least get something out there so we can react to it and try things out.

This is heavily WIP, so no need for full on code review yet, but I’m open to ideas for how we can accomplish something cool here!

Intention here is to create an optional CSS grid that can be enabled in place of the default CSS grid classes. So, to switch from the default grid to CSS grid:

```scss
$enable-grid-classes: false;
$enable-cssgrid: true;
```

From there, we generate a `.grid` instead of a `.row`. The classes within the `.grid` are essentially the same by name, but that’s it. We’d use `column` properties as needed to accomplish this.

Fixes #19729, fixes #23057, fixes #24887, addresses #31812.

Previous demo from your's truly: https://mdo.github.io/bootstrap-css-grid-layout/

/cc @twbs/css-review

---

**Preview:** https://deploy-preview-31813--twbs-bootstrap.netlify.app/docs/5.0/layout/css-grid/

### DIFF
--- a/scss/_grid.scss
+++ b/scss/_grid.scss
@@ -12,6 +12,40 @@
   }
 }
 
+@mixin make-cssgrid($columns: $grid-columns, $breakpoints: $grid-breakpoints) {
+  @each $breakpoint in map-keys($breakpoints) {
+    $infix: breakpoint-infix($breakpoint, $breakpoints);
+
+    @include media-breakpoint-up($breakpoint, $breakpoints) {
+      @if $columns > 0 {
+        @for $i from 1 through $columns {
+          .g-col#{$infix}-#{$i} {
+            grid-column: auto / span $i;
+          }
+        }
+
+        // `$columns - 1` because offsetting by the width of an entire row isn't possible
+        @for $i from 0 through ($columns - 1) {
+          .g-start#{$infix}-#{$i} {
+            grid-column-start: $i;
+          }
+        }
+      }
+    }
+  }
+}
+
+@if $enable-cssgrid {
+  .grid {
+    display: grid;
+    grid-template-rows: repeat(var(--bs-rows, 1), 1fr);
+    grid-template-columns: repeat(var(--bs-columns, #{$grid-columns}), 1fr);
+    gap: var(--bs-gap, #{$grid-gutter-width});
+
+    @include make-cssgrid();
+  }
+}
+
 
 // Columns
 //
--- a/scss/_grid.scss
+++ b/scss/_grid.scss
@@ -38,9 +38,9 @@
 @if $enable-cssgrid {
   .grid {
     display: grid;
-    grid-template-rows: repeat(var(--bs-rows, 1), 1fr);
-    grid-template-columns: repeat(var(--bs-columns, #{$grid-columns}), 1fr);
-    gap: var(--bs-gap, #{$grid-gutter-width});
+    grid-template-rows: repeat(var(--#{$variable-prefix}rows, 1), 1fr);
+    grid-template-columns: repeat(var(--#{$variable-prefix}columns, #{$grid-columns}), 1fr);
+    gap: var(--#{$variable-prefix}gap, #{$grid-gutter-width});
 
     @include make-cssgrid();
   }
--- a/site/content/docs/5.0/layout/css-grid.md
+++ b/site/content/docs/5.0/layout/css-grid.md
@@ -1,0 +1,260 @@
+---
+layout: docs
+title: CSS Grid
+description: Learn how to enable, use, and customize our alternate layout system built on CSS Grid with examples and code snippets.
+group: layout
+toc: true
+---
+
+## Overview
+
+Bootstrap's default grid system represents the culmination of over a decade of CSS layout techniques, tried and tested by millions of people. But, it was also created without so many of the modern CSS features and techniques that we're seeing in browsers—features like the new CSS Grid.
+
+With Bootstrap 5, we've added the option to enable a separate grid system that's built on CSS Grid, but with a Bootstrap twist. You still get classes you can apply on a whim to build responsive layouts, but with a different approach under the hood.
+
+{{< callout info >}}
+**Heads up!** Since our CSS Grid system isn't enabled by default, we've included it in our documentation CSS. Keep reading to see how to enable it yourself.
+{{< /callout >}}
+
+## How it works
+
+- CSS Grid is opt-in. Disable the default grid system by setting `$enable-grid-classes: false` and enable the CSS Grid by setting `$enable-cssgrid: true`. Then, recompile your Sass.
+- Replace instances of `.row` with `.grid`. The `.grid` class sets `display: grid` and creates a `grid-template` that you build on with your HTML.
+- The number of columns and the width of the gutters are set via CSS variables on the `.grid`, which means you can customize those values on the fly: `--columns` and `--gap`.
+
+## Key differences
+
+Compared to the default grid system:
+
+- Flex utilities won't affect grid columns. Note that columns can still be flex containers.
+- There's no `padding` on columns as gutters function more like margins.
+- Unlike `.row`s, `.grid`s have no negative margins.
+- Margin utilities cannot be used to change the grid gutters. See the [customizing section](#customizing).
+- Grid gutters are applied horizontally and vertically by default. See the [customizing section](#customizing).
+- Inline and custom styles should be viewed as replacements for modifier classes (e.g., `style="--columns: 3;"` vs `class="row-cols-3"`).
+- Nesting works similarly, but may require you to reset your column counts on each instance of a nested `.grid`. See the [nesting section](#nesting) for details.
+
+## Examples
+
+### Three columns
+
+Three equal-width columns across all viewports and devices can be created by using the `.g-col-4` classes. Add [responsive classes](#responsive) to change the layout by viewport size.
+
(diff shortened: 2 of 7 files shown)


site/content/docs/5.0/layout/css-grid.md
```diff
@@ -0,0 +1,276 @@
...
+status: experimental
+group: layout
+toc: true
+---
+
+{{< callout info >}}
+- How to generate or create dynamic col classes above default of 12? Sass only?
+- Subgrid?
+- Shorthand for 2x2, 3x3, etc grids? .grid-2x12 (2 rows 12 cols)?
+- Wish we could do `cols="6"` and `[cols] { --columns: attr(cols number); }` but it's not supported.
```
ffoodd, line 14: Note sure it has any sense for us but I solved in chaarts (not about grid but, well, charts) by using `style="--columns: 6;"` for example, which may sounds funny but is really great to work with (and exactly what the `style` attribute is useful for, IMHO).

MartijnCuppens:
Not sure if it's worth it, but could we make use of custom properties for the flex grid columns, so that we could reuse them for the grid grid? This would make it easy to switch between the grids when needed.

Something like: 

```scss
.row > * {
  width: calc(var(--col) / var(--columns) * 100%);
}

.grid > * {
  grid-column: auto / span var(--col);
}

// And for the columns:
.col-6 {
  --col: 6;
}
```
\+ a fix for `col-auto` and `col` classes

ffoodd:
@MartijnCuppens I love the idea! Very clever use of custom properties.

mdo:
@MartijnCuppens Oh very cool idea. Downside would be columns wouldn't have a set width when used _outside_ rows, which was meant as a big improvement coming from v4. Thoughts?

ffoodd:
@mdo May then worth considering adding `width` utilities, since this was very confusing for most users to use `.col-*` to define a width, wasn't it?

site/content/docs/5.0/layout/css-grid.md
```diff
@@ -0,0 +1,271 @@
+---
+layout: docs
+title: CSS Grid
+description: Learn how to enable, use, and customize our alternate layout system built on CSS Grid with examples and code snippets.
+status: experimental
```
XhmikosR, line 5: This might be redundant?

scss/_grid.scss
```diff
@@ -12,6 +12,46 @@
...
+      }
+    }
+  }
+}
+
+@if $enable-cssgrid {
+  .grid {
+    --rows: 1;
+    --columns: #{$grid-columns};
+    --gap: #{$grid-gutter-width};
```
ffoodd, line 44: We could set those as fallback values, instead of explicitly declaring them.

ffoodd (approved):
Looks great!

ffoodd:
Sidenote here: we may introduce `contents` value for our `display` utilities since it'd help flatten the markup in some advanced cases:
- [Hidde de Vries about "More accessible markup with `display: contents`"](https://hiddedevries.nl/en/blog/2018-04-21-more-accessible-markup-with-display-contents)
- [Ire Aderinokun's "How `display: contents` works"](https://bitsofco.de/how-display-contents-works/)
However it should be restricted to non-semantic elements at the moment because of WebKit bugs (under discussion), so just leaving this here as a note for now :)

scss/_grid.scss
```diff
@@ -12,6 +12,46 @@
   }
 }
 
+$enable-cssgrid: true !default;
```
MartijnCuppens, line 15: Any reason why this is declared here?

scss/_grid.scss
MartijnCuppens: Indeed. Also, these should be prefixed.

Chetan11-dev:
Now currently we have been specifying column width through a child(col-6) what we can do is specify it through parent like so (g-3 see the problem below)

Problem: Consider an application where we want to have 
  4 card on large
  2 card on medium 
  1 card on small devices
  grid
We can easily build the above layout using bootstrap grid by flexbox(One of the best things that happen to the Web).
```
    <div class="row">
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
    </div>
``` 
But I propose an alternative grid system that is base on-grid and is easier to think build and refactor. (See Source Code) 
```
    <div style="display: grid;" class=" g-lg-4 g-md-2 g-1 ">
        <div class="card"></div>
        <div class="card"></div>
        <div class="card"></div>
        <div class="card"></div>
    </div>
```
Both codes achieve the same result but which one is easier to refactor and easy to think about.
I suggest you go through the interactive application to visually understand what I mean.

Benefits 
Easily refactorable (No more going to each dev to change the class!!! Leading to faster feedback cycle)
Easier for beginners to understand (No more division with 12!!! )
Lesser Code through use of Grid 

Final Thoughts
Bootstrap could provide us with Grid System which will really be a much-needed act.

# Demo Source
Live Link: https://bootstrap-grid.vercel.app/
```
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">

    <link rel="icon" type="image/png" sizes="32x32" href="./images/favicon-32x32.png">
    <link href="https://fonts.googleapis.com/css2?family=Poppins:wght@600&display=swap" rel="stylesheet">
    <link href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&display=swap" rel="stylesheet">
    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.15.2/css/all.css"
        integrity="sha384-vSIIfh2YWi9wW0r9iZe7RJPrKwp6bG+s9QZMoITbCckVJqGCCRhc+ccxNcdpHuYu" crossorigin="anonymous">
    <title>Frontend Mentor | Huddle landing page with alternating feature blocks</title>
    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css">

    <style>
        .card {
            height: 250px;
        }

        .g-1 {
            grid-template-columns: 1fr;
        }


        @media (min-width: 768px) {


            .g-md-2 {
                grid-template-columns: 1fr 1fr;
            }


        }

        @media (min-width: 992px) {

            .g-lg-4 {
                grid-template-columns: 1fr 1fr 1fr 1fr;
            }
        }
    </style>

</head>

<body>

    <h1 class="text-center">Flexbox</h1>
    <div class="row">
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
        <div class="card  col-lg-3 col-md-6 col-12"></div>
    </div>

    <h1 class="text-center">Grid</h1>
    
    <div style="display: grid;" class=" g-lg-4 g-md-2 g-1 ">
        <div class="card"></div>
        <div class="card"></div>
        <div class="card"></div>
        <div class="card"></div>
    </div>

</body>

</html>

```

ffoodd:
@Chetan11-dev Well, this is really interesting however if I'm getting your point, the exact same feature already exists in our current gird, using [`.row-cols-*` utilities](https://getbootstrap.com/docs/5.0/layout/grid/#row-columns).

Using `grid` has the real benefit of using less code, I'd say. Worth a try :)

mdo:
@ffoodd Want to give this another review now that I've updated a few things? Biggest differences since last major iteration focus on improved documentation, grid cols/gaps/rows are inherited for nesting, and the CSS Grid is included into our docs vs generated as part of the dist CSS.

ffoodd (approved):
I love it! Can't wait to use this :smiley:

ffoodd:
Back to the discussion, could still be paired to `width` utilities and `contents` value for `display`; but it's so easy now with the API that I'd say we wait for request. :)

site/data/sidebar.yml
```diff
@@ -34,6 +34,7 @@
     - title: Gutters
     - title: Utilities
     - title: Z-index
+    - title: CSS Grid
```
XhmikosR, line 37: Wouldn't it make more sense to follow the alphabetical order here?

tatansdev:
i dont know

PR #31815, opened 2020-10-02
updated build tools.md file

krishnadevz:
added lite-server local setup for the bootstrap project


Johann-S:
I don't see any interest to add that, thanks but we'll pass on that 👍

krishnadevz:
@Johann-S it is a good insight to the newbies for running their project on a local server using lite server any final thoughts

XhmikosR:
We already have everything that's needed set up.

PR #31818, opened 2020-10-02
Check for data-interval on the first slide of carousel

### LINKED ISSUE #31716: Carousel data-interval is always 5000 on first item
Hi all, I implemented a carousel using bootstrap carousel component. Everything were working fine until I tried to delay the first item with more than 5 seconds. It looks like bootstrap is ignoring `data-interval="10000"` attribute to the element. I checked it in the documentation and it's not working on the demo there as well. First item is delayed only 5 seconds, instead of 10seconds. Every other item is respecting the data-interval attribute. Can you suggest some workaround or am I missing something?

Link to the documentation: https://getbootstrap.com/docs/4.3/components/carousel/#individual-carousel-item-interval
MacOS - Chrome

mitchellfyi:
When starting a cycle for a carousel, it only checks for a default interval, and not an interval defined on the slide element via data props. This adds a check in before creating the interval to move to the next slide.

Fixes https://github.com/twbs/bootstrap/issues/31716


js/src/carousel.js
```diff
@@ -181,6 +181,16 @@ class Carousel {
...
     if (this._config && this._config.interval && !this._isPaused) {
+      this._activeElement = SelectorEngine.findOne(SELECTOR_ACTIVE_ITEM, this._element)
+
+      if (this._activeElement) {
+        const activeElementInterval = parseInt(this._activeElement.getAttribute('data-interval'), 10)
+
+        if (activeElementInterval) {
+          this._config.interval = activeElementInterval
+        }
+      }
```
Johann-S, line 192: I think we should create a separate private method to do that because earlier we do : 

```js
      const nextElementInterval = parseInt(nextElement.getAttribute('data-interval'), 10)
      if (nextElementInterval) {
        this._config.defaultInterval = this._config.defaultInterval || this._config.interval
        this._config.interval = nextElementInterval
      } else {
        this._config.interval = this._config.defaultInterval || this._config.interval
      }
```

here : https://github.com/mitchellbryson/bootstrap/blob/carousel-data-bug/js/src/carousel.js#L476-L482

So it would be better if we can do something like that : 

```js
this._updateInterval(this._activeElement);
```

which will update correctly the interval

Johann-S (changes requested):
Hi @mitchellbryson,

Thanks for your work 👍 

A few changes left:
- Adding a unit test
- Creating a private method to update the interval according to the next/current element

js/src/carousel.js
mitchellbryson: Thanks. Is it worth patching V4 as well?

mitchellbryson:
Let me know if you want the patch for V4 too.. https://github.com/twbs/bootstrap/pull/31820

XhmikosR:
We handle backports ourselves usually by cherry picking. And I'd prefer if we did the same here too. Only the tests need to be adapted, but we'll see after this lands.

mitchellbryson:
> We handle backports ourselves usually by cherry picking. And I'd prefer if we did the same here too. Only the tests need to be adapted, but we'll see after this lands.

Ok. FYI this change includes using SelectorEngine, V4 doesn't.

XhmikosR:
Hmm, true. I guess your v4-dev PR is fine then; I can change the commit message when merging that later.

Johann-S (dismissed):

XhmikosR:
@mitchellbryson thanks for the PRs! Pretty solid contributions. :)

One last thing before merging: can you please update (or make a new ones) https://codepen.io/Johann-S/pen/zYqeNNO?editors=1010 to print the duration of the first slider too?

mitchellbryson:
@XhmikosR thanks! 

Sure, here you go: https://codepen.io/mitchellbryson/pen/poyXXMd

XhmikosR:
Maybe I'm missing something but I'm getting inconsistent/wrong results. Isn't the second slider's duration supposed to be ~2s? I'm getting ~10s.

https://codepen.io/XhmikosR/pen/yLOdmgE?editors=1111

mitchellbryson:
@XhmikosR  Yep same here, it's not working with transitions. Looking into this now.

mitchellbryson:
The last commit fixes the issue with transitions by setting _activeElement to nextElement at the end of slide() but not after the transition is complete - I'm not familiar enough to know if this is going to be a problem or not so let me know.

A better approach might be to track nextElement on the instance too, but that would require more of a rewrite and I wanted to keep this PR simple until told otherwise.

XhmikosR:
I checked again with https://codepen.io/XhmikosR/pen/yLOdmgE?editors=1011 and it seems to work right :)

XhmikosR:
@mitchellbryson I don't suppose you can update the tests so the previous regression is caught? If I didn't try with the Codepen, the patch would have been wrong.

mitchellbryson:
@XhmikosR yep, I think that's a very good idea :)

mitchellbryson:
_slide() no longer deals with updating the _config.interval, it only updates the _activeElement (just like it does with updating the active indicators through _setActiveIndicatorElement()) and now only cycle() updates the _config.interval, based on the current _activeElement.

_updateInterval() now uses an already set _activeElement, or it finds one (first run).

Updates to the tests reflect that next() is no longer dealing with intervals, only cycle() is.

js/src/carousel.js
```diff
@@ -409,6 +411,21 @@ class Carousel {
     }
   }
 
+  _updateInterval() {
+    const element = this._activeElement || SelectorEngine.findOne(SELECTOR_ACTIVE_ITEM, this._element)
+
+    if (element) {
```
Johann-S, line 417: just a small change, but I prefer if we can return asap so this check should be:

```js
if (!element) {
  return
}

**Comment by Johann-S** [comment]
@XhmikosR thinks LGTM here 👍

**Comment by XhmikosR** [comment]
I'd still like tests to include the previous failed attempt which wasn't caught :/

**Comment by mitchellbryson** [comment]
@XhmikosR is https://github.com/twbs/bootstrap/pull/31818/files#diff-5eae769087ca4f58ffb64ddef2d4b0e1R877 not sufficient? I added the 2nd cycle() call (which was missing before).

**Comment by XhmikosR** [comment]
@mitchellbryson oh I didn't see that patch, sorry. I think it should be covered now, indeed. @Johann-S please confirm that this covers all cases.

**Comment by mitchellbryson** [comment]
I updated the pen to cover to show the 2nd bugfix working too: https://codepen.io/mitchellbryson/pen/ZEWgBEN?editors=1011

**Comment by XhmikosR** [comment]
I can confirm that tests fail propery on main BTW :)

```
Chrome Headless 85.0.4183.121 (Windows 10) Carousel cycle should get interval from data attribute on the active item element FAILED
        Error: Expected 1814 to equal 7.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:898:41 <- js/tests/unit/carousel.spec.js:16886:43)
            at <Jasmine>
        Error: Expected 1814 to equal 9385.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:903:41 <- js/tests/unit/carousel.spec.js:16889:43)
            at <Jasmine>
Chrome Headless 85.0.4183.121 (Windows 10) Carousel cycle should get interval from data attribute on the active item element FAILED
        Error: Expected 1814 to equal 7.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:898:41 <- js/tests/unit/carousel.spec.js:16886:43)
            at <Jasmine>
        Error: Expected 1814 to equal 9385.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:903:41 <- js/tests/unit/carousel.spec.js:16889:43)
            at <Jasmine>
........................................
Chrome Headless 85.0.4183.121 (Windows 10) Carousel next should update the active element to the next item before sliding FAILED
        Error: Expected null to equal <div id="secondItem" class="carousel-item carousel-item-next carousel-item-left">...</div>.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:656:39 <- js/tests/unit/carousel.spec.js:16779:41)
            at <Jasmine>
Chrome Headless 85.0.4183.121 (Windows 10) Carousel next should update the active element to the next item before sliding FAILED
        Error: Expected null to equal <div id="secondItem" class="carousel-item carousel-item-next carousel-item-left">...</div>.
            at <Jasmine>
            at UserContext.<anonymous> (js/tests/unit/js/tests/unit/carousel.spec.js:656:39 <- js/tests/unit/carousel.spec.js:16779:41)
            at <Jasmine>
...
Chrome Headless 85.0.4183.121 (Windows 10): Executed 450 of 450 (2 FAILED) (5.366 secs / 4.968 secs)
TOTAL: 2 FAILED, 448 SUCCESS
```

XhmikosR:
FYI @mitchellbryson I just wait for another review; unfortunately it seems we are a little stuck (lack of JS-involved people). I'd like to get this patch into alpha3 and #31820 assuming you reopen it and adapt it after this one is merged.

Johann-S (approved):

PR #31827, opened 2020-10-02
Add namespace "bs" in the data attributes

### LINKED ISSUE #29479: Add a namespace for data-* attribute?
Hello,

As of V5 is coming, maybe it's also time to add a namespace on `data` attribute to prevent conflict with others library?

Something like for events `data-bs-toggle`?

rohit2sharma95:
Closes #29479

Preview: <https://deploy-preview-31827--twbs-bootstrap.netlify.app/>


js/src/carousel.js
```diff
@@ -366,7 +366,7 @@ class Carousel {
     const activeIndex = this._getItemIndex(activeElement)
     const lastItemIndex = this._items.length - 1
     const isGoingToWrap = (isPrevDirection && activeIndex === 0) ||
-                            (isNextDirection && activeIndex === lastItemIndex)
```
XhmikosR, line 369: Can you revert any whitespace changes that unrelated to the PR so that the diff is as clean as possible?

js/src/dom/manipulator.js
```diff
@@ -47,9 +47,13 @@ const Manipulator = {
       ...element.dataset
     }
 
-    Object.keys(attributes).forEach(key => {
-      attributes[key] = normalizeData(attributes[key])
-    })
+    Object.keys(attributes)
+      .filter(attributes => attributes.includes('bs'))
```
XhmikosR, line 51: I'm not 100% sure includes is the safest solution here. Aren't the attributes expected to always being with `bs`?

site/content/docs/5.0/examples/dashboard/index.html
```diff
@@ -31,37 +31,37 @@
         <ul class="nav flex-column">
           <li class="nav-item">
             <a class="nav-link active" aria-current="page" href="#">
-              <span data-feather="home"></span>
+              <span data-bs-feather="home"></span>
```
XhmikosR, line 34: All these data-bs-feather seem wrong :)

site/content/docs/5.0/examples/masonry/index.html
```diff
@@ -17,11 +17,11 @@ <h1>Bootstrap and Masonry</h1>
 &lt;script src=&quot;https://cdnjs.cloudflare.com/ajax/libs/masonry/4.2.2/masonry.pkgd.min.js&quot; integrity=&quot;sha384-GNFwBvfVxBkLMJpYMOABq3c+d3KnQxudP/mGPkzpZSTYykLBNsZEnG2D9G/X/+7D&quot; crossorigin=&quot;anonymous&quot; async&gt;&lt;/script&gt;
   </code></pre>
 
-  <p>By adding <code>data-masonry='{"percentPosition": true }'</code> to the <code>.row</code> wrapper, we can combine the powers of Bootstrap's responsive grid and Masonry's positioning.</p>
+  <p>By adding <code>data-bs-masonry='{"percentPosition": true }'</code> to the <code>.row</code> wrapper, we can combine the powers of Bootstrap's responsive grid and Masonry's positioning.</p>
```
XhmikosR, line 20: Same here, should be reverted.

XhmikosR (changes requested):

XhmikosR:
@rohit2sharma95 thanks for the PR! I left a few comments so that we get this moving and I'll drop my branch since this seems cleaner. I will need to discuss with the rest of @twbs/team but I think we could include this change in alpha3.

site/content/docs/5.0/components/dropdowns.md
```diff
@@ -848,14 +848,14 @@ var dropdownList = dropdownElementList.map(function (dropdownToggleEl) {
 {{< /highlight >}}
 
 {{< callout info >}}
-##### `data-toggle="dropdown"` still required
+##### `data-bs-toggle="dropdown"` still required
 
-Regardless of whether you call your dropdown via JavaScript or instead use the data-api, `data-toggle="dropdown"` is always required to be present on the dropdown's trigger element.
+Regardless of whether you call your dropdown via JavaScript or instead use the.data-api, `data-bs-toggle="dropdown"` is always required to be present on the dropdown's trigger element.
```
XhmikosR, line 853: The period here is also wrong :)

site/content/docs/5.0/components/tooltips.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: docs
 title: Tooltips
-description: Documentation and examples for adding custom Bootstrap tooltips with CSS and JavaScript using CSS3 for animations and data-attributes for local title storage.
+description: Documentation and examples for adding custom Bootstrap tooltips with CSS and JavaScript using CSS3 for animations and data-bs-attributes for local title storage.
```
XhmikosR, line 4: Not sure about this change, /CC @mdo

site/content/docs/5.0/getting-started/javascript.md
```diff
@@ -208,12 +208,12 @@ var myDefaultAllowList = bootstrap.Tooltip.Default.allowList
...
 
-// To allow td elements and data-option attributes on td elements
-myDefaultAllowList.td = ['data-option']
+// To allow td elements and data-bs-option attributes on td elements
+myDefaultAllowList.td = ['data-bs-option']
 
 // You can push your custom regex to validate your attributes.
 // Be careful about your regular expressions being too lax
-var myCustomRegex = /^data-my-app-[\w-]+/
+var myCustomRegex = /^data-bs-my-app-[\w-]+/
```
XhmikosR, line 216: Perhaps this should be reverted since it seems app specific?

XhmikosR (changes requested):

js/src/dom/manipulator.js
```diff
@@ -47,9 +47,13 @@ const Manipulator = {
...
+      .forEach(key => {
+        let pureKey = key.replace('bs', '')
```
XhmikosR, line 53: Same here, if we know `key` starts with `'bs'` it might be better to do `replace(/^bs/, '')` so that reduce the chances of modifying something that isn't ours.

rohit2sharma95:
Thanks @XhmikosR 
I will do the required changes :heart:

js/src/carousel.js
rohit2sharma95: Done :+1:

js/src/dom/manipulator.js
rohit2sharma95: Now using `startsWith` instead :)

site/content/docs/5.0/examples/dashboard/index.html
rohit2sharma95: Done :+1:

site/content/docs/5.0/components/dropdowns.md
rohit2sharma95: Updated :+1:

js/src/dom/manipulator.js
rohit2sharma95: Done :+1:

js/src/dom/manipulator.js
XhmikosR: I'd go with a regex for now since we don't use `startsWith`/`endsWith` yet. I tried to switch to it in https://github.com/twbs/bootstrap/compare/master-xmr-js-modern-api but I'm hesitant. We'll tackle that later all at once.

js/src/dom/manipulator.js
```diff
@@ -47,9 +47,13 @@ const Manipulator = {
       ...element.dataset
     }
 
-    Object.keys(attributes).forEach(key => {
-      attributes[key] = normalizeData(attributes[key])
-    })
+    Object.keys(attributes)
+      .filter(attributes => attributes.startsWith('bs'))
```
XhmikosR, line 51: BTW this should be `attribute` probably?

site/layouts/partials/docs-navbar.html
```diff
@@ -4,9 +4,9 @@
       {{ partial "icons/bootstrap-white-fill.svg" (dict "class" "d-block my-1" "width" "40" "height" "32") }}
     </a>
 
-    <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#bdNavbar" aria-controls="bdNavbar" aria-expanded="false" aria-label="Toggle navigation">
+    <button class="navbar-toggler" type="button" data-bs-toggle="collapse" data-bs-target="#bdNavbar" aria-controls="bdNavbar" aria-expanded="false" aria-label="Toggle navigation">
       <svg class="bi" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
-        <path fill-rule="evenodd" d="M2.5 11.5A.5.5 0 0 1 3 11h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5zm0-4A.5.5 0 0 1 3 7h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5zm0-4A.5.5 0 0 1 3 3h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5z"/>
+        <path fill-rule="evenodd" d="M2.5 11.5A.5.5 0 0 1 3 11h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5zm0-4A.5.5 0 0 1 3 7h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5zm0-4A.5.5 0 0 1 3 3h10a.5.5 0 0 1 0 1H3a.5.5 0 0 1-.5-.5z" />
```
XhmikosR, line 9: Please revert this change also

js/src/dom/manipulator.js
rohit2sharma95: Yes :smile: updated

js/src/dom/manipulator.js
XhmikosR: Actually, I only see one more case from a quick look that we could use `startsWith`, so leave it as is for now and we'll see later :)

js/src/dom/manipulator.js
```diff
@@ -47,9 +47,13 @@ const Manipulator = {
       ...element.dataset
     }
 
-    Object.keys(attributes).forEach(key => {
-      attributes[key] = normalizeData(attributes[key])
-    })
+    Object.keys(attributes)
+      .filter(key => key.startsWith('bs'))
+      .forEach(key => {
+        let pureKey = key.replace(/^bs/, '')
```
Johann-S, line 53: since the logic changed here, you should update the according unit tests and add new ones

Johann-S (changes requested):
it LGTM, adding and Bootstrap namespace is a good idea 👍 

it lacks a bit of unit tests but thanks for the work 👍

js/src/dom/manipulator.js
rohit2sharma95: @Johann-S 
Updated 👍

Johann-S: I would prefer to just have `bsToggle` in that case

js/tests/unit/dom/manipulator.spec.js
```diff
@@ -60,14 +60,30 @@ describe('Manipulator', () => {
...
       const div = fixtureEl.querySelector('div')
 
       expect(Manipulator.getDataAttributes(div)).toEqual({
-        test: 'js',
-        test2: 'js2'
+        bsToggle: 'tabs',
+        bsTarget: '#element',
+        another: 'value',
+        toggle: 'tabs',
+        target: '#element'
```
Johann-S, line 73: sorry if I wasn't clear but for me should just expect that here:
```js
{
  bsToggle: 'tabs',
  bsTarget: '#element',	        
  another: 'value',
}
```

any thoughts here @XhmikosR ?

rohit2sharma95: @Johann-S **bs** is removed from the attributes because it is not present in the configuration of the plugins.
For example:
https://github.com/twbs/bootstrap/blob/e0b8fcdf899aa1c25fe2ddf050452f7451ed0cdd/js/src/carousel.js#L42-L49
Ref: https://github.com/twbs/bootstrap/commit/1232ed355a7ee9a201f5d7c14ffe4b19ede41e98#commitcomment-40771099

Johann-S: Ok understood 👌

Johann-S (approved):

site/layouts/partials/docs-subnav.html
```diff
@@ -1,12 +1,12 @@
 <nav class="bd-subnavbar py-2" aria-label="Secondary navigation">
   <div class="container-xxl d-flex align-items-md-center">
     <form class="bd-search position-relative mr-auto">
-      <input type="search" class="form-control" id="search-input" placeholder="Search docs..." aria-label="Search docs for..." autocomplete="off" data-docs-version="{{ .Site.Params.docs_version }}">
+      <input type="search" class="form-control" id="search-input" placeholder="Search docs..." aria-label="Search docs for..." autocomplete="off" data-bs-docs-version="{{ .Site.Params.docs_version }}">
```
XhmikosR, line 4: Note that this is not a core thing and we should probably rename to `data-bd-version` or leave it like it was. @mdo thoughts?

XhmikosR:
How do you all feel if we added code to handle the old case too but clearly state in the migration notes that this will be removed in the next version instead of v5?

Just thinking out loud, trying to see what's better for our users and make the v5 upgrade smooth.

All in all, this branch is almost ready, I just want @mdo's thoughts in a few things.

rohit2sharma95:
> How do you all feel if we added code to handle the old case too but clearly state in the migration notes that this will be removed in the next version instead of v5?

For now, all the configurations for the plugins can be passed with either `data-bs-` or `data-`. For example `data-bs-toggle="5000"` and `data-interval="5000"` both can be used and attribute with bs keyword has more priority.

https://github.com/rohit2sharma95/bootstrap/blob/3566f465078104bda744ff7cff8a8b8839b93cb6/js/tests/unit/dom/manipulator.spec.js#L77-L88

Breaking changes are just for selectors like `data-toggle` is now `data-bs-toggle`.

XhmikosR:
Branch rebased and fixed a few missed instances

> Breaking changes are just for selectors like `data-toggle` is now `data-bs-toggle`.

Can you please clarify this in the migration note I added?

rohit2sharma95:
Okay @XhmikosR :+1:

rohit2sharma95:
@XhmikosR These are the attributes that are being used either as selectors or in the `getAttribute` method in js files. So these attributes must contain **bs** namespace in between:

**data-bs-toggle
data-bs-target
data-bs-dismiss
data-bs-slide
data-bs-slide-to
data-bs-ride
data-bs-spy
data-bs-interval
data-bs-parent
data-bs-content**

All other configurations to the javascript plugins can be passed without using **bs** namespace as well. Now I think that you guys should consider it as a completely breaking change because these are many selectors.

site/content/docs/5.0/migration.md
```diff
@@ -9,6 +9,8 @@ toc: true
 
 ## v5.0.0-alpha3
 
+- **BREAKING**: all JavaScript plugins now use `data-bs-foo` attributes instead of `data-foo`.
```
XhmikosR, line 12: Reminder: we need to reword/explain a bit more @mdo

XhmikosR:
So, this makes me think if we should only accept prefixed attributes @rohit2sharma95. I mean, it's a breaking change already, is there any gain in accepting some attributes unprefixed?

/CC @mdo @Johann-S for thoughts too

rohit2sharma95:
Agree @XhmikosR 
Updates are here: https://github.com/rohit2sharma95/bootstrap/pull/13

js/src/dom/manipulator.js
XhmikosR: @rohit2sharma95 one last request: can you add a couple of test cases with `BS` and `bs` not in the start of the data key name please?

XhmikosR: Oh, NVM I see you have already covered these cases too :)

rohit2sharma95: Yes, these are covered already @XhmikosR. I have not checked for uppercase namespace though.

XhmikosR: I'm not sure if it makes sense after all, I mean, are those even case sensitive?

rohit2sharma95: To filter the namespaced attributes `startsWith` method is used that is already case sensitive. So I think there is no need to check for the uppercase namespace. And I did not check it before because I was thinking that people always write HTML attributes in the small case. 🙂

XhmikosR (approved):

rohit2sharma95:
@XhmikosR This PR has been merged already though. And you have made data attributes to be bootstrap specific.
https://github.com/twbs/bootstrap/blob/60c79b28aa52bfbb1736bfed07666fea4eef172f/js/tests/unit/dom/manipulator.spec.js#L63
I was thinking that you guys should also consider making the remaining attribute methods (In manipulators) to be bootstrap specific. I mean these changes: https://github.com/twbs/bootstrap/compare/main...rohit2sharma95:bs-data-cons

So whenever dealing with data attributes, it should be done only with bootstrap specific data attributes automatically. It will also make manipulator methods consistent. 🙂

XhmikosR:
Please open a new PR :)

PR #31830, opened 2020-10-03
Simplify .form-check style

jonathanhefner:
This allows `.form-check-input` to be used with fewer classes when not wrapped in `.form-check`, and it prevents the space between input and label from scaling uncomfortably big at larger font sizes.

**Before**

simplify-form-check-style-before

**After**

simplify-form-check-style-after

---

I admit I am unsure of the purpose of the `float: left`.  I evaluated this change with all of the "Checks and radios" examples in the docs, and they all looked correct.  Is there a use case I am missing, or is its presence historical?

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -672,8 +672,8 @@ $input-transition:                      border-color .15s ease-in-out, box-shado
 
 
 $form-check-input-width:                  1em !default;
+$form-check-input-margin-right:           .2rem !default;
 $form-check-min-height:                   $font-size-base * $line-height-base !default;
-$form-check-padding-left:                 $form-check-input-width + .5em !default;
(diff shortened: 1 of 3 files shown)


scss/_variables.scss
```diff
@@ -672,8 +672,8 @@ $input-transition:                      border-color .15s ease-in-out, box-shado
 
 
 $form-check-input-width:                  1em !default;
+$form-check-input-margin-right:           .2rem !default;
```
ydmitry, line 675: `*-margin-x` or `*-margin-r` by convention if I'm not wrong

mdo:
This removes the hanging checkbox styling, so we'll have to pass on it. If you want to keep the checkbox and text closer together, we can look at modifying the existing styles in another way.

PR #31834, opened 2020-10-05
tooltip/popover: add a `customClass` option

### LINKED ISSUE #19415: Simplify the customization of popover

remeika:
**Please squash before merging**

Fixes #19415

Add extra classes to Tooltips and Popovers using the `additionalClasses` option, using [`$.addClass()`](https://api.jquery.com/addClass/) behind the scenes. 

Details:

- Additional classes are evaluated and added every time the `show()` command is run. This should be idempotent, and allows the class list to be updated every time the component is shown.
- New classes are added after the component template is hydrated in the DOM; if classes are added to a custom template and using `additionalClasses`, both will show up. 

Questions I still have:

- I targeted this release at the 4.X branch. I am not sure if I did the right thing there, rather than basing my work off `main`.
- It looks like the file size of `bootstrap.bundle.js` has been locked in place: adding any additional code causes the build pipeline to fail. I'm wondering what course of action an individual contributor should take in this situation: Is the 4.X branch feature-locked? Should I try to delete some code elsewhere to try to squeeze under the limit? Should I just raise the limit?

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -41,6 +41,7 @@ const DefaultType = {
   container: '(string|element|boolean)',
   fallbackPlacement: '(string|array)',
   boundary: '(string|element)',
+  customClass: '(string|function)',
   sanitize: 'boolean',
   sanitizeFn: '(null|function)',
   whiteList: 'object',
@@ -70,6 +71,7 @@ const Default = {
   container: false,
   fallbackPlacement: 'flip',
   boundary: 'scrollParent',
+  customClass: '',
   sanitize: true,
   sanitizeFn: null,
   whiteList: DefaultWhitelist,
@@ -284,6 +286,7 @@ class Tooltip {
       this._popper = new Popper(this.element, tip, this._getPopperConfig(attachment))
 
       $(tip).addClass(CLASS_NAME_SHOW)
+      $(tip).addClass(this._getCustomClass())
 
       // If this is a touch-enabled device we add extra
       // empty mouseover listeners to the body's immediate children;
@@ -729,6 +732,10 @@ class Tooltip {
     this.hide()
     this.show()
     this.config.animation = initConfigAnimation
+  }
+
+  _getCustomClass() {
+    return this.element.getAttribute('data-custom-class') || this.config.customClass
   }
 
   // Static
--- a/js/tests/unit/popover.js
+++ b/js/tests/unit/popover.js
@@ -61,6 +61,19 @@ $(function () {
       .bootstrapPopover('show')
   })
 
+  QUnit.test('should render popover element with additional classes', function (assert) {
+    assert.expect(2)
+    var done = assert.async()
+    $('<a href="#" title="mdo" data-content="https://twitter.com/mdo" data-custom-class="a b">@mdo</a>')
+      .appendTo('#qunit-fixture')
+      .on('shown.bs.popover', function () {
+        assert.strictEqual($('.popover').hasClass('popover fade bs-popover-right show'), true, 'has default classes')
+        assert.strictEqual($('.popover').hasClass('a b'), true, 'has custom classes')
+        done()
+      })
+      .bootstrapPopover('show')
+  })
+
   QUnit.test('should store popover instance in popover data object', function (assert) {
     assert.expect(1)
     var $popover = $('<a href="#" title="mdo" data-content="https://twitter.com/mdo">@mdo</a>').bootstrapPopover()
--- a/js/tests/unit/tooltip.js
+++ b/js/tests/unit/tooltip.js
@@ -1283,4 +1283,54 @@ $(function () {
 
     assert.strictEqual(popperConfig.placement, 'left')
   })
+
+  QUnit.test('additional classes can be applied via data attribute', function (assert) {
+    assert.expect(2)
+
+    $('<a href="#" rel="tooltip" data-trigger="click" title="Another tooltip" data-custom-class="a b"/>')
+      .appendTo('#qunit-fixture')
+      .bootstrapTooltip()
+      .bootstrapTooltip('show')
+
+    var tooltip = $('.tooltip')
+
+    assert.strictEqual(tooltip.hasClass('a b'), true)
+    assert.strictEqual(tooltip.hasClass('tooltip fade bs-tooltip-top show'), true)
+  })
+
+  QUnit.test('additional classes can be applied via config string', function (assert) {
+    assert.expect(2)
+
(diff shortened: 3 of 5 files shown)


XhmikosR:
Thanks for the PR!

That being said, we always target the main branch first and then we backport anything if deemed necessary. That being said, since the tests differ, we'd need a v4-dev patch anyway.

But we'd need a patch for the main branch anyway. About bundlewatch, leave that to us, we'll update the file before merging.

js/src/tooltip.js
```diff
@@ -435,6 +442,10 @@ class Tooltip {
     return title
   }
 
+  getAdditionalClasses() {
```
XhmikosR, line 445: TBH I don't know if we need the separate function. And if we did, it should begin with an underscore since it's private, and moved in the private block

js/src/tooltip.js
```diff
@@ -41,6 +41,7 @@ const DefaultType = {
   container: '(string|element|boolean)',
   fallbackPlacement: '(string|array)',
   boundary: '(string|element)',
+  additionalClasses: '(string|function)',
```
Johann-S, line 44: I would prefer a parameter called : `customClass` which would allow an array of string or a function it would be better if someone wants to add a lot of classes

js/src/tooltip.js
Johann-S: yep it should be a private method and start with an `_`
btw it should return an array of classes and with we would be able to do just one call to `addClass` it should be better in terms of perfs

js/tests/unit/popover.js
```diff
@@ -61,6 +61,18 @@ $(function () {
       .bootstrapPopover('show')
   })
 
+  QUnit.test('should render popover element with additional classes', function (assert) {
+    assert.expect(1)
+    var done = assert.async()
+    $('<a href="#" title="mdo" data-content="https://twitter.com/mdo" data-additional-classes="a b">@mdo</a>')
+      .appendTo('#qunit-fixture')
+      .on('shown.bs.popover', function () {
+        assert.strictEqual($('.popover').hasClass('a'), true)
```
Johann-S, line 70: Can you assert the default classes are present too ?

js/tests/unit/tooltip.js
```diff
@@ -1283,4 +1283,50 @@ $(function () {
...
+    assert.expect(2)
+
+    $('<a href="#" rel="tooltip" data-trigger="click" title="Another tooltip" data-additional-classes="a b"/>')
+      .appendTo('#qunit-fixture')
+      .bootstrapTooltip()
+      .bootstrapTooltip('show')
+
+    var tooltip = $('.tooltip')
+
+    assert.strictEqual(tooltip.hasClass('a'), true)
```
Johann-S, line 1297: same here (checking if default classes are present)

js/tests/unit/tooltip.js
```diff
@@ -1283,4 +1283,50 @@ $(function () {
...
+    assert.strictEqual(tooltip.hasClass('b'), true)
+  })
+
+  QUnit.test('additional classes can be applied via function', function (assert) {
+    assert.expect(2)
+
+    $('<a href="#" rel="tooltip" data-trigger="click" title="Another tooltip" />')
+      .appendTo('#qunit-fixture')
+      .bootstrapTooltip({
+        additionalClasses: () => 'a b'
```
Johann-S, line 1323: we run on old browsers in v4 and our unit tests doesn't use Babel, so now arrow function please

Johann-S (changes requested):
Nice work @remeika! 👍 

Some changes to do but not so much

js/tests/unit/tooltip.js
XhmikosR: I'm going to try to enforce this in linting, I thought we already did.

js/src/tooltip.js
remeika: @Johann-S  As this PR stands now, the additionalClasses property is a logic-less wrapper around [`jQuery.addClass()`](https://api.jquery.com/addClass/#addClass-classNames). As of jquery 3.3, addClass() does accept an array of classes. So my initial thinking was that in Bootstrap 4.X, we guarantee that a string or function are valid inputs (although passing in an array will work in environments with a modern version of jQuery), and then in Bootstrap 5 we could rely on a more modern version of jQuery. But I just checked `package.json` on the `main` branch, and jQuery is no longer listed as a peer dependency. Will Bootstrap 5 not depend on jQuery? If so I will definitely change my strategy.

js/src/tooltip.js
remeika: Changed name to `_getCustomClass()`. See below for discussion about using built-in array processing in `$.addClass`.

js/tests/unit/popover.js
remeika: 👍 done.

cankitm:
> But we'd need a patch for the main branch anyway. About bundlewatch, leave that to us, we'll update the file before merging.

@XhmikosR @Johann-S  Can I pick the patch for main branch?

js/src/tooltip.js
Johann-S: in v4 we can stay like you did, just the renaming (additionalClasses -> customClass) but since v5 is intended to work without jQuery I would prefer an array and a function

XhmikosR:
My only concern is that people might ask for the ability add custom classes in all components. So, shouldn't we have a helper function and add this functionality in all components?

Johann-S:
yup you're right it will be asked for sure.

But currently our code architecture doesn't allow to have one helper shared across our components because we don't use inheritance

XhmikosR:
Maybe we should wait until this can be applied to all components? I feel like it's sort of incomplete if we don't add this for all components

Johann-S:
it's a point of view, or we can add it when folks ask for it 🤷

XhmikosR:
@remeika can you make a PR for main too please?

XhmikosR:
@Johann-S LGTY?

@rohit2sharma95 if we merge this, can you make a PR against main later please? 🙂

rohit2sharma95:
Okay sure 👍  @XhmikosR

Johann-S (approved):

PR #31835, opened 2020-10-05
Move `href` attribute first

XhmikosR:
Move `href` attribute first

@mdo not sure what the order should be? If it's supposed to be the other way around let me know and I'll update the branch.


mdo (commented):
I like including `href` first so it's clearer earlier in the line what's being imported, if that makes sense. I could go either way though since these aren't regularly updated or anything.

mdo (approved):

PR #30550, opened 2020-04-10
Bump `$dark` to `$gray-900`

ffoodd:
On the road to #30548 , trying to increase contrasts wherever `$dark` is used.

https://deploy-preview-30550--twbs-bootstrap.netlify.com/docs/5.0/utilities/colors/

FWIW, I really prefer this.

PR #31839, opened 2020-10-05
docs: fix contrast color against `.bg-info`

### LINKED ISSUE #31749: Info badge colour
This is the first time I've reported something so I'm not sure if I'm doing it in the right place or not.

I note that the new info badge doesn't provide sufficient colour contrast per AA guidelines in the V5 Alpha 2.

https://v5.getbootstrap.com/docs/5.0/components/badge/

The contrast is 1.96 apparently when used with white text as seen in the docs. I use it with black text myself, which does provide the contrast required. 
So the docs either need changing or the colour needs changing :)

ffoodd:
Fixes #31749 

Preview: <https://deploy-preview-31839--twbs-bootstrap.netlify.app/>

### DIFF
--- a/site/content/docs/5.0/components/badge.md
+++ b/site/content/docs/5.0/components/badge.md
@@ -45,7 +45,7 @@ Use our background utility classes to quickly change the appearance of a badge. 
 {{< example >}}
 {{< badge.inline >}}
 {{- range (index $.Site.Data "theme-colors") }}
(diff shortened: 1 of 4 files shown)


XhmikosR:
Should we backport this? I'm not sure if we have backported similar patches.

XhmikosR (approved):

ffoodd:
#31276 isn't backported, neither are #30989 and #30550 — so I guess not, even though we could pretty easily.

XhmikosR:
Better leave it as is then, because we might be missing more patches :)

Issue #31823, opened 2020-10-02
Remove js/dist/index.js and js/dist/index.js.map since they seem unused?


XhmikosR:
/CC @Johann-S the thing we were discussing the other day

XhmikosR:
/CC @Johann-S the thing we were discussing the other day

Johann-S:
we should do the same in v5 too

XhmikosR:
I'm mostly concerned about v4 because it really seems unused and the map file is definitely redundant and not auto-generated. Just keep in mind we don't want any accidental breaking changes in v4, though :)

XhmikosR:
Fixed by #32001

PR #31840, opened 2020-10-05
V4 dev backports and updates

XhmikosR:
Refs #31823


XhmikosR:
This should restore the ES5 enforcement in tests @Johann-S

XhmikosR:
I left js/dist.index.js for the time being, although it does seem unused. We should revisit later if we need it and remove it if not.

mdo:
> I left js/dist.index.js for the time being, although it does seem unused. We should revisit later if we need it and remove it if not.

I see this is covered in #31823, so maybe we slate for v4.6.0? Do we need to warn folks at all?

mdo (approved):

XhmikosR:
> I see this is covered in #31823, so maybe we slate for v4.6.0? Do we need to warn folks at all?

I intentionally kept index.js in v4-dev, since I can't be 100% sure if this will break something. But the sourcemap was outdated and moot for sure, hence why I decided to remove only that here :)

PR #31841, opened 2020-10-05
spacing problem between dropdowns on dropdown docs

mznet:
image

I've found that a spacing problem between dropdowns on docs. This problem is caused by left margin even though line breaks. To fix problem, I removed left margin between dropdowns on xs size and applied the margin on over sm size.

You can check the issue on [Sizing](https://v5.getbootstrap.com/docs/5.0/components/dropdowns/#sizing) section of dropdown page. Check it on xs size screen. And this issue is happened to v4 as well as v5.

Preview: https://deploy-preview-31841--twbs-bootstrap.netlify.app/docs/5.0/components/dropdowns/#sizing

### DIFF
--- a/site/content/docs/5.0/components/dropdowns.md
+++ b/site/content/docs/5.0/components/dropdowns.md
@@ -260,7 +260,7 @@ Button dropdowns work with buttons of all sizes, including default and split dro
         <li><a class="dropdown-item" href="#">Separated link</a></li>
       </ul>
     </div><!-- /btn-group -->
-    <div class="btn-group ml-2">
+    <div class="btn-group ml-sm-2">
       <button type="button" class="btn btn-lg btn-secondary">Large split button</button>
       <button type="button" class="btn btn-lg btn-secondary dropdown-toggle dropdown-toggle-split" data-toggle="dropdown" aria-expanded="false">
         <span class="visually-hidden">Toggle Dropdown</span>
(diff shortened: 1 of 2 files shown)


site/content/docs/5.0/components/dropdowns.md
```diff
@@ -260,7 +260,7 @@ Button dropdowns work with buttons of all sizes, including default and split dro
         <li><a class="dropdown-item" href="#">Separated link</a></li>
       </ul>
     </div><!-- /btn-group -->
-    <div class="btn-group ml-2">
+    <div class="btn-group ml-sm-2">
```
mdo, line 263: This now no longer matches L290, where we still use `ml-2`. I'd suggest deleting both `ml-2` instances instead.

mdo (dismissed):

site/content/docs/5.0/components/dropdowns.md
mznet: @mdo set `L263` and `L290` the same with deleting both `ml-2`.
I'd adjusted margin of `> .btn-toolbar > .btn-group,` in the previous commit and I realized that it affected button group docs. So, I additionally adjusted margin of `> .btn-toolbar > .input-group` to set proper margin in button group docs

MartijnCuppens (approved):
LGTM, thanks @mznet!

mdo:
Replacing this with #31861—while reviewing, I realized we could split those examples and avoid unnecessary changes to our components CSS (which negatively impact all many other `btn-group`s in other snippets).

Issue #31832, opened 2020-10-04
[V4 + V5] Add Separator For Same-Colored Button Groups


einfallstoll:
I think it's incredibly hard to visually distinguish between regular buttons with a caret (which open a dropdown) and buttons in a button group with a separate button for the dropdown.

Regular buttons with a caret that opens a dropdown:
<img width="641" alt="image" src="https://user-images.githubusercontent.com/619048/95012491-45453c80-0639-11eb-93fd-f47049735282.png">

Button groups with a separate caret button that opens a dropdown:
<img width="717" alt="image" src="https://user-images.githubusercontent.com/619048/95012479-31013f80-0639-11eb-8249-8f71e6ffbbaf.png">

To see the difference between the buttons, you either have to get a feeling for margins so you can guess the bigger margin means a separate button, or you have to hover the button:
<img width="145" alt="image" src="https://user-images.githubusercontent.com/619048/95012528-98b78a80-0639-11eb-9998-1792efd7fc13.png">

I think this not only applies to button groups for dropdowns but also to regular button groups. I also got critical responses for not visually separate buttons:
<img width="243" alt="image" src="https://user-images.githubusercontent.com/619048/95012555-d7e5db80-0639-11eb-8db5-0654ec419fea.png">

The same applies to V5 as well.

I'm quite sure there was a separator in the past, which got removed at some point. So I request to re-add the separator again, as it makes it easier to distinguish between same-colored button groups as well as making it more accessible for visually impaired people.

zokizuan:
yeah...I need those

FarkasJanos:
Hi guys! I would work on it

einfallstoll:
This is how I solved it for the moment:
```
.btn-group {
  @each $color, $value in $theme-colors {
    .btn-#{$color} + .btn-#{$color} {
      border-left: $btn-border-width solid darken($value, 10%);
    }
  }
}
```

This will add a border between all buttons of the same color within a button group, but none if they have a different color.

PR #31853, opened 2020-10-08
Added separator for button groups

ARKEOLOGIST:
Worked on a small fix for #31832, reverting some code from branch v3-dev to ensure borders in button groups.

### DIFF
--- a/scss/_button-group.scss
+++ b/scss/_button-group.scss
@@ -108,6 +108,21 @@
   }
 }
 
+// Reposition the caret
+.btn .caret {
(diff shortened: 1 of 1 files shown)


mdo:
There's no `.caret` in the main branch.

cankitm:
Issue was there because we are using same color for button background and button border.
Updated the color of border to make it different.

Please See: Used the same color we use for hover in btn-primary.

If approved, will do the same for all buttons. @XhmikosR

PR #31858, opened 2020-10-09
Btn : Add different border color for btn-primary #31832

cankitm:
Regrading #31832
Issue was there because we are using same color for button background and button border.
Updated the color of border to make it different.

Please See: Used the same color we use for hover in btn-primary.

If approved, will do the same for all buttons.

Made changes as required. @XhmikosR


MartijnCuppens:
This isn't a generic solution and uses a hardcoded border color

PR #31859, opened 2020-10-09
Add a "skeletons" component

### LINKED ISSUE #31665: Skeleton loading screens
Hi,

I would love to see something like a skeleton loading screen for specific elements (paragraphs, buttons, images, headings etc).
The reason why I want this is because its very handy to show the user that there is something loading, without placing spinners.

I'm aware of other solutions, but it would be nice to see it included in Bootstrap.

Examples:
[Example CSS tricks](https://css-tricks.com/building-skeleton-screens-css-custom-properties/)
[Example Youtube](https://miro.medium.com/max/700/0*ABjKedHjIe8El9RJ.png)

Hope my request is clearly and understandable.

Anyway, thanks in advance!
Stan.

### LINKED ISSUE #31909: Content loading placeholder
I think it is interesting if bootstrap natively supports content loading placeholder components.
Giving to developer standard content placeholders to using when content is loading.

image

Another examples

image

jaumesala:
## Introduction

Nowadays, webapps use different techniques to improve the UX of the user when some data needs to be loaded. One of those trending techniques is the use of css "skeletons". No matter what javascript tool we use, we should facilitate and encourage developers to use them. Now that the framework is evolving and other components like the `spinner`has been added. I've thought on give a shot and explore the `skeletons` option.

skeleton example

## How it works

Skeletons are composed of `.skeleton-line`s. Each of them needs one or more `.skeleton-dash`es.

```html
<span class="skeleton-line">
  <span class="skeleton-dash"></span>
</span>
```

By default all `.skeleton-dash`es together will take the full with of the `.skeleton-line`. Still, you can customize them with custom css widths or even using the utility classes.

Screenshot 2020-10-09 at 18 07 52Screenshot 2020-10-09 at 18 08 03

The way it works is using a hidden empty space to preserve the size of the parent HTML element. This way we can mimic the original element proportion.

Screenshot 2020-10-09 at 18 08 11

For more information on how it works, check the documentation.

Waiting for feedback.

Fixes #31665

Preview: <https://deploy-preview-31859--twbs-bootstrap.netlify.app/docs/5.0/components/placeholders/>

TODO:

- [x] fix the empty heading errors properly
- [x] a11y
- [x] scss vars in docs like the other components?
- [x] fix examples
- [x] use a `range` in the colors docs?
- [ ] maybe add a button to toggle animation?


scss/_skeletons.scss
```diff
@@ -0,0 +1,61 @@
...
+    height: 1em;
+    background-color: $gray-400;
+
+    + .skeleton-dash {
+      margin-left: 4px;
+    }
+  }
+
+  // Sizing
+  &.skeleton-xs {
```
mdo, line 27: Because we use a base-modifier pairing in our classes, you don't ned to chain these selectors. You can un-nest and omit the `&`.

scss/_skeletons.scss
```diff
@@ -0,0 +1,61 @@
...
+  // the font size of the parent
+  &::before {
+    width: 0;
+    content: "\00A0";
+  }
+
+  > .skeleton-dash {
+    width: 100%;
+    height: 1em;
+    background-color: $gray-400;
```
mdo, line 19: Maybe we use `currentColor` or something to allow for easy customization of colors with our text utilities? Wondering if that allows folks to insert these into other examples more readily? I'm not 100% since I've not used these myself, so feel free to discard.

```suggestion
    background-color: currentColor;
```

mdo (commented):
This is awesome! Just a few questions and stuff in a quick review :D.

patrickhlauke:
to make any kind of sense to assistive tech users, at the very least these will need to have some indication like `aria-busy="true"` which is then removed when they are actually populated with content, and you'll likely want to have the actual content inside the container `aria-hidden="true"` so AT users that wander into it don't accidentally heard jibberish.

scss/_skeletons.scss
jaumesala: fixed

scss/_skeletons.scss
```diff
@@ -0,0 +1,61 @@
...
+  width: 100%;
+
+  // We use a &nbsp; to preserve
+  // the font size of the parent
+  &::before {
+    width: 0;
+    content: "\00A0";
+  }
+
+  > .skeleton-dash {
```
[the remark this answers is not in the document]
jaumesala: nah, it's just that I used `> span` before adding the `.skeleton-dash` class. Fixed.

scss/_skeletons.scss
jaumesala: I've updated the color to `currentColor`. I used a grayish color to "decolorize" because to my eyes it was too blackish...

I've tried with different approaches here:

First with `gray-400` 
Screenshot 2020-10-10 at 10 27 18

New one with `currentColor`
Screenshot 2020-10-10 at 10 28 20

Alternative, use `currentColor + opacity`
Screenshot 2020-10-10 at 10 28 57

The best would be to use something like `lighten(currentColor, 50%)` but this is not possible 😅.
Any other idea is welcome, otherwise `currentColor` is a good option. Developers can always use the color utilities so...

scss/_skeletons.scss
```diff
@@ -0,0 +1,61 @@
...
+
+  &.skeleton-sm {
+    > .skeleton-dash {
+      height: .8em;
+    }
+  }
+
+  &.skeleton-lg {
+    > .skeleton-dash {
+      height: 1.2em;
```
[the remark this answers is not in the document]
jaumesala: hmmm, I used `1em` as the base height to inherit tha parent text/line height. What I tried to represent is something between the cap-height and the x-height of the typography with the `skeleton-dash` class.

Screenshot 2020-10-10 at 11 37 32

Feedback/help? 😋

scss/bootstrap.scss
```diff
@@ -41,6 +41,7 @@
 @import "popover";
 @import "carousel";
 @import "spinners";
+@import "skeletons";
```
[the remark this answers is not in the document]
jaumesala: I don't like it either, I used that name because it seems to be the known name for this css technique.
The `placeholder-*` word is an option, too long maybe?

`replace-*`
`mock-*`
`fake-*`
`skl-*`
`ph-*`
`marker-*`
`dash-*`
...

I let you decide 😋

jaumesala:
@patrickhlauke I've added the `aria-busy="true"` attribute and a hint on how to use it with `aria-hidden="true"`. Take a look and see if it is correct 51d89f00a8c2e5f9db3eb3d5cea3aa4d0447867b

site/content/docs/5.0/components/skeletons.md
```diff
@@ -0,0 +1,157 @@
...
+layout: docs
+title: Skeletons
+description: Create loading skeletons for your components or pages with Bootstrap skeleton, built entirely with HTML, CSS, and no JavaScript.
+group: components
+toc: true
+---
+
+## About
+
+Bootstrap “skeletons” can be used to enhance the experience of your application. They’re built only with HTML and CSS, meaning you don’t need any JavaScript to create them. You will, however, need some custom JavaScript to toggle their visibility. Their appearance, color, and sizing can be easily customized with our amazing utility classes.
```
XhmikosR, line 11: Please do not use curly double/single quotes. This is handled automatically for us.

site/content/docs/5.0/components/skeletons.md
```diff
@@ -0,0 +1,157 @@
...
+
+Skeletons are composed of `.skeleton-line`s. Each of them needs one or more `.skeleton-dash`es.
+
+{{< example >}}
+<span class="skeleton-line">
+  <span class="skeleton-dash"></span>
+</span>
+{{< /example >}}
+
+{{< callout info >}}
```
XhmikosR, line 72: Let's replace any callouts with the updated #31802

site/content/docs/5.0/components/skeletons.md
```diff
@@ -0,0 +1,157 @@
...
+  <span class="skeleton-dash bg-warning"></span>
+  <span class="skeleton-dash bg-info"></span>
+  <span class="skeleton-dash bg-light"></span>
+  <span class="skeleton-dash bg-dark"></span>
+  <span class="skeleton-dash bg-white"></span>
+  <span class="skeleton-dash bg-transparent"></span>
+</span>
+{{< /example >}}
+
+### Sizing
```
XhmikosR, line 118: Line break please

site/content/docs/5.0/components/skeletons.md
```diff
@@ -0,0 +1,147 @@
...
+
+{{% callout info %}}
+Even if you use size modifiers, the space between lines is not affected, only the thickness of the dash. This is in purpose to help you customize the weights of the different elements of the skeleton composition.
+{{% /callout %}}
+
+{{< example >}}
+<div class="d-flex justify-content-around">
+  <div style="width: 45%;">
+    <h3>
+      <span class="skeleton-line"><span class="skeleton-dash"></span>&#8232;</span>
```
XhmikosR, line 123: Not a huge fan of one nested one-liners. If it doesn't affect the display, I'd say better nest the elements for better readability.

site/content/docs/5.0/components/skeletons.md
```diff
@@ -35,7 +35,7 @@ In this example we have built a fancy HTML card using Bootstrap components. Then
     {{< placeholder width="100%" height="180" class="card-img-top" text="false" >}}
     <div class="card-body">
       <h5 class="card-title">
-        <span class="skeleton-line skeleton-glow w-50"><span class="skeleton-dash bg-secondary"></span></span>&nbsp;
+        <span class="skeleton-line skeleton-glow w-50"><span class="skeleton-dash bg-secondary"></span>&#8232;</span>
```
XhmikosR: Also, I'm not sure this is the proper solution here @patrickhlauke

scss/_skeletons.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+// The animation effect
+@keyframes skeleton-glow {
+  0%,
+  100% {
+    opacity: 1;
+  }
+  50% {
```
ffoodd, line 57: No need to define `0%` and `100%` here since `opacity: 1` is the initial value. See an example: https://codepen.io/ffoodd/pen/qBNdMLY

scss/_skeletons.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+.skeleton-dash {
+  width: 100%;
+  height: 1em;
+  background-color: currentColor;
+
+  + .skeleton-dash {
+    margin-left: 4px;
```
ffoodd, line 28: `4px` feels wrong, you'd better use one of our spacers (and probably define a variable for this).

scss/_skeletons.scss
```diff
@@ -0,0 +1,60 @@
+// .skeleton-line      - the skeleton container
+// .skeleton-dash      - the "text" replacement
```
ffoodd, line 2: This is already well explained in the docs, no need to mention this in the sources :)

scss/_skeletons.scss
```diff
@@ -0,0 +1,60 @@
...
+  display: flex;
+  align-items: center;
+  width: 100%;
+
+  // We use a &nbsp; to preserve
+  // the font size of the parent
+  &::before {
+    width: 0;
+    content: "\00A0";
+  }
```
ffoodd, line 14: Even though I truly like this trick, wouldn't a grid-based layout be easier to use for anyone using Bootstrap?
Coouldn't we use this alongside `.row` and `.col`? Still thinking about it, but I like this way :)

scss/bootstrap.scss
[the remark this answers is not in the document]
ffoodd: `placeholder` sounds better to me, it was known as "content placeholder" for a long time (eg in 2014, https://cloudcannon.com/deconstructions/2014/11/15/facebook-content-placeholder-deconstruction.html).

scss/_skeletons.scss
```diff
@@ -0,0 +1,60 @@
...
+  > .skeleton-dash {
+    height: .8em;
+  }
+}
+
+.skeleton-lg {
+  > .skeleton-dash {
+    height: 1.2em;
+  }
+}
```
ffoodd, line 49: Since only `.skeleton-dash` gets modified, why don't we apply modifiers on them instead of their parent?

By the way, I think they're not necessary as-is: setting `1em` as `height` on `.skeleton-dash` allows to use `font-size` on the parent, eg: 
```scss
.skeleton-lg {
  font-size: 1.2em;
}
```

This would lighten sources and selectors a bit, plus solve one of @mdo's comment ;)

ffoodd (dismissed):
Apart from my CSS-part comments, did you ever think of using SVG for this? Using some `<rect>`s to lay up the pattern would be quite easy too, and it'd be easier to convey alternate text.

Not quite sure about this, but that's [how I handled it years ago](https://dabblet.com/gist/466b6a47db1b4ac14ecb). I guess using raw markup may be easier to use but still want to consider alternate ways :)

Nice feature idea and overall work though, thanks!

site/content/docs/5.0/components/skeletons.md
jaumesala: Should I inline the h3 with the skeleton classes then? Is that what you are saying? 🤔

jackmu95:
We at BootstrapVue recently introduced sketeton components.
These include text, images, buttons, inputs and even tables.

It would be awesome to have official support for all of these components in Bootstrap v5.
Our implementation may be a good starting point and we could adopt to the official markup in the future.

https://bootstrap-vue.org/docs/components/skeleton

site/content/docs/5.0/components/skeletons.md
jaumesala: It was a hotfix to pass the checks! I didn't know exactly what to do...

site/content/docs/5.0/components/skeletons.md
XhmikosR:
```html
<span class="skeleton-line">
  <span class="skeleton-dash"></span>
  &#8232;
</span>
```

But I don't think the `&#8232;` solution is right. I wait for @patrickhlauke's feedback about that.

ffoodd:
The `cursor: wait;` declaration is a good idea. The overall implementation is pretty much the same as this PR, from a fast and superficial review, isn't it?

scss/_skeletons.scss
jaumesala: It was bugging me too. I forgot about the spacers. Here's a quick test. I'd say `$spacer * .5` looks the best? Agree with the variable too. Let me know what you think.

`margin-left: $spacer;`
Screenshot 2020-10-12 at 11 27 26
`margin-left: $spacer * .5;`
Screenshot 2020-10-12 at 11 27 48
`margin-left: $spacer * .25;`
Screenshot 2020-10-12 at 11 28 07

ffoodd: I'd be for quarter myself based on those screenshots, but will need to play a bit with it to make a decision. BTW:
1. we prefer to use division to divide; see [our spacers](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L235).
2. and I guess a variable would be helpful for this `margin`, something like `$skeleton-spacing-x`.

scss/_skeletons.scss
jaumesala: I've just pushed the change, but then I realized it's breaking the "line-height" trick. 
Should I revert the commit or forget about the "trick" ?

With the new commit
Screenshot 2020-10-12 at 11 59 32

Before the commit
Screenshot 2020-10-12 at 12 02 36

scss/_skeletons.scss
jaumesala: Ok, I've changed the margin to `$spacer / 4` to play with it. 
I hold out the variable until we settle down the implementation, is it ok?

jaumesala:
> We at BootstrapVue recently introduced sketeton components.
> These include text, images, buttons, inputs and even tables.
> 
> It would be awesome to have official support for all of these components in Bootstrap v5.
> Our implementation may be a good starting point and we could adopt to the official markup in the future.
> 
> https://bootstrap-vue.org/docs/components/skeleton

@jackmu95 I didn't know of the existence of this (component)! 
But looking at your examples, I also had in mind to add a skeleton for images, `skeleton-figure` for instance, not sure if it is needed though... 

Also, I was plaing with `mask-image` + `linear gradient` to create another effect like your *wave* effect, but for the first PR I stopped prototyping it. 

IMO I'd say those two effects "glow" and "wave" are a must for the Bootstrap version. 😋

XhmikosR:
@jaumesala:

```
git fetch upstream
git rebase -i upstream/main
```

Your last push has the wrong patches. If you don't how to fix it, let me know so that I fix it, but please don't override the changes again.

jaumesala:
I apologize @XhmikosR! Please correct it. I pulled the changes this morning thought it was ok. 
If you can explain what I've done wrong so I don't make the same mistake again I'll appreciate it. 🙇‍♂️

XhmikosR:
NP, these things happen :)

So, basically, you merged instead of a rebase after I had already rebased the branch. So, now, you need to do this:

```
git fetch --all
git reset --hard origin/v5-skeletons-component
```

Assuming `origin` is your repo and `upstream` the upstream repo and you are already in the local `v5-skeletons-component` branch. I won't rebase anything else to make things easier, but keep in mind you can always rebase your branch to clean up/re-organize  your patches.

site/content/docs/5.0/components/skeletons.md
```diff
@@ -0,0 +1,147 @@
...
+
+The size of the `.skeleton-dash`es are calculated according to the typography style of the parent element. To customize them you can use classes like `.skeleton-lg`, `.skeleton-sm` or `.skeleton-xs`.
+
+{{% callout info %}}
+Even if you use size modifiers, the space between lines is not affected, only the thickness of the dash. This is in purpose to help you customize the weights of the different elements of the skeleton composition.
+{{% /callout %}}
+
+{{< example >}}
+<div class="d-flex justify-content-around">
+  <div style="width: 45%;">
```
XhmikosR, line 121: Would it make more sense if we used the grid here?

shahoob:
Just to tell you:
https://bootstrap-vue.org/docs/components/skeleton
Bootstrap Vue has it already implemented.

MartijnCuppens:
The markup can be simplified quite a bit if we're using `display: inline-block` for the skeleton/placeholders. Made a codepen here, what do you think? https://codepen.io/MartijnCuppens/pen/ExyNKYq?editors=1100

jaumesala:
@MartijnCuppens I like this implementation. In fact I think it is a clever trick because it uses Bootstrap's current grid system. 
Let's wait some more feedback before rewriting it?

Besides I was thinking about moving the *animation class* to a parent element so we have more control over the component's "content". For instance if we want to create an effect with `mask-image: linear-gradient(...)` on all placeholders or any other trick. I would like to know some opionons.

MartijnCuppens:
> Let's wait some more feedback before rewriting it?

Yup. @twbs/css-review, thoughts on https://github.com/twbs/bootstrap/pull/31859#issuecomment-713819368?

> Besides I was thinking about moving the animation class to a parent element so we have more control over the component's "content". For instance if we want to create an effect with `mask-image: linear-gradient(...)` on all placeholders or any other trick. I would like to know some opinions.

Good idea, but I think this opens the door for "Should we introduce some mask/animation helper classes that can be used for multiple components". To somewhat scope this PR I wanna focus on introducing this placeholder component that works out of the box for now, so we can ship something.

jaumesala:
> > Besides I was thinking about moving the animation class to a parent element so we have more control over the component's "content". For instance if we want to create an effect with `mask-image: linear-gradient(...)` on all placeholders or any other trick. I would like to know some opinions.
> 
> Good idea, but I think this opens the door for "Should we introduce some mask/animation helper classes that can be used for multiple components". To somewhat scope this PR I wanna focus on introducing this placeholder component that works out of the box for now, so we can ship something.

Agree! I do not mean about implementing any new feature just change this:
```scss
.placeholder {
  &.placeholder-glow { animation: glow; ... }
}
```
To something like this:
```scss
.placeholder-glow {
   .placeholder { animation: glow; ... }
}
```

ffoodd:
Concerning @MartijnCuppens suggestion, I'm for: much cleaner markup and easier to understand.

@jaumesala is the nesting required for `animation`? You may simply define `.placeholder-glow { animation: …; }`, this would allow to use it either on a placeholder directly or on any parent. Maybe need to dig a bit more, but nesting shouldn't be a concern.

jaumesala:
> @jaumesala is the nesting required for `animation`? You may simply define `.placeholder-glow { animation: …; }`, this would allow to use it either on a placeholder directly or on any parent. Maybe need to dig a bit more, but nesting shouldn't be a concern.

Not really, at least for the *glow* effect, but maybe for another type of effect... It also prevents writing the animation class  for each `.placeholder` class. Small details...

I'll have limited time to spend on the PR on the following days. Let me know if we should move on with the new implementation.

mdo:
I'm down for the approach in @MartijnCuppens's comment! The only thing I see that I'm unsure how to get around is the `&nbsp;` in things like the button. I'm braindead right now, but otherwise I like the approach!

MartijnCuppens:
> The only thing I see that I'm unsure how to get around is the `&nbsp;` in things like the button.

We could add this maybe? 
```scss
.placeholder.btn {
  min-height: add($btn-line-height * 1em, add($btn-padding-y * 2, $btn-border-width * 2, false));
}

.placeholder.btn-sm {
  min-height: add($btn-line-height * 1em, add($btn-padding-y-sm * 2, $btn-border-width-sm * 2, false));
}

.placeholder.btn-lg {
  min-height: add($btn-line-height * 1em, add($btn-padding-y-lg * 2, $btn-border-width-lg * 2, false));
}
```

jaumesala:
I've refactored the code of the PR.

- renamed *skeletons* to *placeholders*
- refacotres css classes according to @MartijnCuppens approach
- updated docs
- added ```placeholder-wave``` animation

patrickhlauke:
ok, having now found time to test https://deploy-preview-31859--twbs-bootstrap.netlify.app/docs/5.0/components/placeholders/ ... it seems browser/AT support for how/when to announce `aria-busy` is a bit flakey. it works reasonably well when the attribute is directly on an otherwise interactive control or widget, but not so much when it's on a container element like a `<div>`. now, some browser/AT combos treat `aria-busy` on something as a signal not to even announce the presence of the thing, while others will still let the user navigate through things (e.g. announce the heading and link in the placeholder card), without giving any indication at all that it's inside a busy container...

so, authors will need to basically adapt their approach and test stuff. i'll mull over how this can be conveyed somehow in that callout note, and may tweak the example card case a tiny bit...leaning towards making the whole thing `aria-hidden` in that case (noting that when you have BOTH `aria-hidden` and `aria-busy`, the latter will never be seen/perceived by AT anyway as the thing is hidden, so having both would be redundant).

mdo:
@jaumesala Thanks for everything here! We're close to our next release and want to leave this for the release after that. Still have every intention of shipping this then, so hold tight while we keep the train moving :).

jaumesala:
Any update on this? Let me know if I any work should be done on the PR. 👌

ffoodd:
Some more insights regarding accessibility: [Adrian Roselli's "More accessible skeletons"](https://adrianroselli.com/2020/11/more-accessible-skeletons.html). We may also share some thoughts on when to & how to use, eg. [Tim Kadlec's "Effective skeleton screens"](https://timkadlec.com/remembers/2020-11-02-skeleton-screens/) is a must-read :)

mdo:
I think I'm still good on this, just waiting for v5.0 or v5.1 to add the new feature. Gotta focus on cleaning up and improving the beta work :).

MartijnCuppens:
Little update on the codepen example, I've now added
```scss
  &.btn::before {
    content: '';
    display: inline-block;
  }
```
instead of the `&nbsp;` trick.

I prefer my `1em` approach over the `.placeholder-xs`, `placeholder-sm`, `.placeholder-lg` used in this PR since the `1em` adjusts to all font sizes (`h1,h2,h3,h4,h5,h6`) and the implementation is just way simpler.

I'm not a fan of the waves since the glow isn't in sync when used over different elements (because of the angle and the possible difference in element width). I also noticed my laptop fan was going going nuts when I enabled the waves (which makes sense, since animating masks can be quite heavy), so I rather drop that for now.

https://codepen.io/MartijnCuppens/pen/ExyNKYq?editors=1100

jaumesala:
First of all, happy new year to everybody. I wish a year full of releases and contributions for the bootstrap team! 🥳
Regarding this PR, I know we are waiting for a stable release before adding this feature. 
Still, If you want me to help in anything; try new changes, discuss some code, etc. Let me know it. 🙂

StanMenten99:
Any update on this feature?

mdo:
Revisiting this to check up on some other PRs and I noticed there's now a small todo list in the opening comment. @jaumesala @XhmikosR can y'all confirm what needs doing and see if we need any other updates here? Maybe we can merge this sooner than later.

XhmikosR:
@mdo we need a proper solution for the whitespace.

@patrickhlauke can you check the first two TODOs please? I can take care of the rest.

jaumesala:
I may have some free time this weekend. Let me know if I can help with anything.

patrickhlauke:
i should be able to spend some time this weekend on this as well, yes

patrickhlauke:
is netlify updating its preview or is it just not running due to bundlewatch failure?

patrickhlauke:
assuming https://deploy-preview-31859--twbs-bootstrap.netlify.app/docs/5.0/components/placeholders/ is still the latest version:

* out of all browser/AT combinations, it seems only JAWS (with Chrome/Firefox) actually respects `aria-busy="true"` and effectively skips any content marked as busy. NVDA and VoiceOver (with various browsers) still read out the content without any indication that it's "busy".
* focusable elements inside placeholder containers also still receive keyboard focus (with or without assistive technologies running)

I'd say two things at least should happen:

- in addition to `aria-busy="true"`, there should also be `aria-hidden="true"` on the placeholder container. arguably, this makes the `aria-busy="true"` redundant as it hides the element that is set to be busy, but until browser/AT support for this is improved, that attribute is pointless anyway (but makes sense to still keep there, regardless, as at least it shows the intent). arguably - though more involved - you could even say the `aria-busy="true"` should be on the placeholder container, and then any direct *child* element should be given an explicit `aria-hidden="true"`. and this should then programmatically be removed when the placeholder becomes active/live (and the `aria-busy-"true"` is removed or set to `"false"`)
- any focusable element inside the placeholder should be given an explicit `tabindex="-1"` to prevent it being keyboard focusable (and again, this should be removed when the element becomes active)

mdo:
@jaumesala Hey there! It's been awhile, but we're finally shipping v5.0.0 and are ready to look at our next set of features for v5.1.0. Would you be willing to jump back in here and help us ship this? If not we can try to tackle ourselves :). Thanks!

jaumesala:
Hey @mdo! Yeah, glad to help you guys! I've reread the last comments and commited some changes regarding "a11y" and SCSS. I've also cleaned up some examples in the documentation. Let me know what you think. Btw, It looks like Bundlewatch is still complainging about the size.

patrickhlauke:
circling back to this, as `aria-busy` support is still poor...i think the simplest and most sensible approach overall is simply to use `aria-hidden="true"` on any container/element that's in a skeleton/busy state (as there's no point doing both `aria-busy="true"` and `aria-hidden="true"` on the same element, since even *if* it was used in a browser/AT combo that supported busy, it would not be announced as it's hidden).

note also that any focusable elements need to be explicitly made non-focusable with `tabindex="-1"` or (where supported, e.g. on form controls/buttons) `disabled`).

jaumesala:
Ok @patrickhlauke! Just to be sure I understand your comment. Using the example as a reference, it is enough to `add aria-hidden="true"` on the the direct descendants elements (see below)? Or, should we should event add it on the `h5`, `span`, `p`, `a`, too?

```html
<div class="card" aria-busy="true" style="width: 18rem;">
    <img aria-hidden="true" src="..." class="card-img-top" alt="...">
    <div aria-hidden="true" class="card-body">
      <h5 class="card-title placeholder-glow">
        <span class="placeholder col-6"></span>&#8232;
      </h5>
      <p class="card-text placeholder-glow">
        <span class="placeholder col-7"></span>
        <span class="placeholder col-4"></span>
        <span class="placeholder col-4"></span>
        <span class="placeholder col-6"></span>
        <span class="placeholder col-8"></span>
      </p>
      <a href="#" tabindex="-1" class="btn btn-primary disabled placeholder col-6">
        &nbsp;<!-- needed to give the element some height -->
      </a>
    </div>
  </div>
```

patrickhlauke:
i think currently, seeing that AT with one exception don't actually do anything useful/announce anything to users, we can probably completely forego `aria-busy` for generic containers, particularly if they don't have any additional text (like a "Loading...") or similar that gives at least some information to AT users that something is supposed to be there.

in the case of the card, just hiding the entire card with `aria-hidden="true"` and neutralising the link with `tabindex="-1"` is enough 

```html
<div class="card" aria-hidden="true" style="width: 18rem;">
    <img src="..." class="card-img-top" alt="...">
    <div class="card-body">
      <h5 class="card-title placeholder-glow">
        <span class="placeholder col-6"></span>&#8232;
      </h5>
      <p class="card-text placeholder-glow">
        <span class="placeholder col-7"></span>
        <span class="placeholder col-4"></span>
        <span class="placeholder col-4"></span>
        <span class="placeholder col-6"></span>
        <span class="placeholder col-8"></span>
      </p>
      <a href="#" tabindex="-1" class="btn btn-primary disabled placeholder col-6">
        &nbsp;<!-- needed to give the element some height -->
      </a>
    </div>
  </div>
```

it gets a bit trickier/nuanced if it's a very specific element with a role and accessible name that is currently being updated ... for instance, a link on its own that is being updated seems to be announced as "busy" - for instance, the example given of the busy link is indeed announced as "busy, link" by Chrome/NVDA. but i think that's more an edge case. i'd be ok with the whole link just being hidden.

long story short: just hiding the element that is set as a skeleton is probably sufficient at this stage. It gets more case-specific depending how authors will actually use the placeholder styles, how they plan to update things, etc. Might be worth just focusing on the styling here, showing a basic example with just `aria-hidden`, and then hinting at the fact that more may need to be done for AT users. I can have a think about the latter part...

XhmikosR:
Does anybody have any ideas how we could skip the `&nbsp;` part? It should simplify our markup.

jaumesala:
I've updated the code with @patrickhlauke recommendations and rewritten the callout information regarding the loading implementation. Take a look, it might need better words. 😋

mdo:
Pushed some changes and squashed everything into a single commit here. The main change was to add some pseudo-class styles for `.btn::before` that negates the need for `&nbsp;` characters in the HTML. Going to review docs once more and then ask for a re-review.

mdo (approved):

XhmikosR:
Pinging @patrickhlauke for another look at a11y and @ffoodd for the CSS changes :)

XhmikosR (approved):

PR #31863, opened 2020-10-10
Add separator to following same colored buttons in button group

FarkasJanos:
#31832


mdo:
We shouldn't be adding custom separators. If folks want borders visible at all times, we should make that happen instead of applying patches based on the button's placement in another component.

einfallstoll:
@mdo What do you mean by borders? If you have a look at the [issue](https://github.com/twbs/bootstrap/issues/31832) you see how it's visually difficult to differentiate between single and grouped buttons because of the missing separator. IMHO this is a design flaw in Bootstrap and should be visually separated by default (if there's more than one button glued together).

patrickhlauke:
FWIW @mdo I agree that button groups look rather...odd/broken without some form of visible separation between the individual buttons. and i'd welcome some change here to address that - and doing it as part of button group (where that problem occurs since they're all butted together) seems most relevant?

> image

PR #31871, opened 2020-10-12
Revert "Improve callout shortcode."

XhmikosR:
Reverts twbs/bootstrap#31802

Unfortunately, this results in the callout headings being included in the ToC. I'm not sure if there is any way to exclude specific headings apart from maybe using CSS to hide them, so maybe reverting this might be a better solution.


XhmikosR:
@regisphilibert: if you have any suggestions please let me know :)

Ideally, we should use `{{% foo %}}` since we do want the content to be processed by the Markdown engine.

mdo (approved):

Issue #24889, opened 2017-11-27
Feature request: smooth scroll behaviour


MartijnCuppens:
Setting `scroll-behaviour` to `smooth` will enable smooth scrolling when scrolling happens due to navigation or CSSOM scrolling APIs. In my opinion it is clearer if a user scrolls to the section where he needs to be instead of the _instant scoll_.

The scroll behaviour can be set on `:root` or on every element (not inherited by default). Maybe make a variable, so this behaviour can easily be turned off if needed?

Currently only Chrome and Firefox support this feature, but it's more like an enhancement so I don't think this is an issue.

Demo: https://codepen.io/MartijnCuppens/pen/bYjEby?editors=1100
More info: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior
Support: https://caniuse.com/#feat=css-scroll-behavior

mdo:
Thanks, but we'll pass on this for now.

PR #31877, opened 2020-10-13
Apply smooth-scroll when `prefers-reduced-motion: no-preference`

ffoodd:
After watching [@mmatuzo's talk "Writing even more CSS with accessibility in mind"](https://www.matuzo.at/blog/writing-even-more-css-with-accessibility-in-mind-progressive-enhancement/), I wanted to suggest optionnal smooth scroll.

I saw #24889 but it's kinda outdated, and with the support of `prefers-reduced-motion` it's more robust now.

Very simple and lightweight PR so feel free to discard it if you think it's still not valuable.


patrickhlauke (approved):

patrickhlauke:
i'd be in favour of this

mdo (approved):
I can dig!

mdo:
We'll want to mention this in the Migration guide, and perhaps add something else in our docs about this.

ffoodd:
Added a mention in our accessibility page, and introduced a reboot section in the migration guide for alpha3.

site/content/docs/5.0/getting-started/accessibility.md
```diff
@@ -47,6 +47,8 @@ For visually hidden interactive controls, such as traditional "skip" links, use
 
 Bootstrap includes support for the [`prefers-reduced-motion` media feature](https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion). In browsers/environments that allow the user to specify their preference for reduced motion, most CSS transition effects in Bootstrap (for instance, when a modal dialog is opened or closed, or the sliding animation in carousels) will be disabled.
 
+Alternatively, the `scroll-behavior` property enables smooth scroll if the user did not ask for reduced motion.
```
patrickhlauke, line 50: this, in isolation, makes no sense. alternatively to what? i think this needs a bit more of an explanation.

ffoodd: Alternatively to disabling `transitions`s, in my mind :D 
Apart from the beginning of the sentence, using maybe something like "Bootstrap also use…" which would clarify the subject a bit, I'm not quite sure of what to explain more. I'll try something :)

scss/_reboot.scss
```diff
@@ -27,6 +27,10 @@
 
 :root {
   font-size: $font-size-root;
+
+  @media (prefers-reduced-motion: no-preference) {
+    scroll-behavior: smooth;
```
MartijnCuppens, line 32: Maybe we should make this configurable (Sass variable)?

site/content/docs/5.0/getting-started/accessibility.md
```diff
@@ -47,7 +47,7 @@ For visually hidden interactive controls, such as traditional "skip" links, use
 
 Bootstrap includes support for the [`prefers-reduced-motion` media feature](https://drafts.csswg.org/mediaqueries-5/#prefers-reduced-motion). In browsers/environments that allow the user to specify their preference for reduced motion, most CSS transition effects in Bootstrap (for instance, when a modal dialog is opened or closed, or the sliding animation in carousels) will be disabled.
 
-Alternatively, the `scroll-behavior` property enables smooth scroll if the user did not ask for reduced motion.
+Bootstrap also enables smooth scroll if the user did not ask for reduced motion, using the `scroll-behavior` property.
```
patrickhlauke, line 50: Perhaps a bit verbose, but I'd go for something along the lines of:

```
On browsers that support the `prefers-reduced-motion`, and where the user has *not* explicitly signaled that they'd prefer reduced motion (i.e. where `prefers-reduced-motion: no-preference`), Bootstrap enables smooth scrolling using the `scroll-behavior` property. 
```

scss/_reboot.scss
ffoodd: Neat idea; I introduced `$enable-smooth-scroll` (`true` as a default) to ease override.

site/content/docs/5.0/getting-started/accessibility.md
ffoodd: Very clear, thanks!

XhmikosR (approved):

andrehsu:
On single-page applications, the effect is really jarring when using client-side navigation, which scrolls to the top of a page after page load. Is there a way to disable this effect? Using
```
html {
    scroll-behavior: auto;
}
```
didn't seem to work.

ffoodd:
@andrehsu Please open an issue instead of reviving a closes pull request.

panoply:
> On single-page applications, the effect is really jarring when using client-side navigation

@andrehsu indeed it is and it should not be the default. 

When using `scroll-behavior: smooth` then`window.scrollTo` will not be respected. It's problematic, especially considering modern web is leveraging client side rendered frameworks. Smooth scroll behaves identical to issues which arise when setting `body, html {  width: 100%; height: 100%; }`. If (for example) you are swiping back and forth between history in web applications that restore scroll positions manually then issues arise. The interesting thing here is that (in chrome at-least) `scroll-behavior: smooth` will work fine when navigating back and forth by buttons, ie: users clicks back/forward button but the moment a user swipes back/forward is where the issues arise and previously referenced scroll positions or scroll positions being applied onload and invoked via `window.scrollTo` will not work. 

The solution or workaround I've found is to explicitly set this to `auto` on the `:root`selector and also asserting an `!important`on the selector. 

```
:root {
  scroll-behavior: auto !important;
}
```

Also, I'm not going to create a new issue for this because reviving an old one is how those come across this defect will be able to better find context regarding the bug, it is also linked in documentation.

PR #31886, opened 2020-10-13
Fixed: Undefined mixin 'deprecate' when importing 'bootstrap-grid-scss'

masserra:
Following @mdo 's fix #31438 that adds a **deprecate** notice, I've noticed that this also breaks the compilation when solely importing `bootstrap-grid.scss`.

### DIFF
--- a/scss/bootstrap-grid.scss
+++ b/scss/bootstrap-grid.scss
@@ -16,12 +16,10 @@ html {
   box-sizing: inherit;
 }
 
-// Deprecate
-@import "mixins/deprecate";
-
 @import "functions";
 @import "variables";
 
+@import "mixins/deprecate";
 @import "mixins/breakpoints";
 @import "mixins/grid-framework";
 @import "mixins/grid";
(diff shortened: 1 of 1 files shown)


scss/bootstrap-grid.scss
```diff
@@ -16,6 +16,9 @@ html {
   box-sizing: inherit;
 }
 
+// Deprecate
+@import "mixins/deprecate";
```
mdo, line 20: Can you move this down to the other mixins?

masserra: Sure, I had it there initially but then I went to check where you guys have put it on the [nixon's file](https://github.com/twbs/bootstrap/blob/a716fb03f965dc0846df479e14388b1b4b93d7ce/scss/_mixins.scss#L9) and so it at the top. Should I move it down there too?

mdo: The new spot for it is right, thanks! Just wanted it at the top of the mixins section :).

mdo (approved):

PR #31934, opened 2020-10-19
Update carousel.md

manonthemat:
Clarify that carousel's interval is set in milliseconds.


site/content/docs/5.0/components/carousel.md
```diff
@@ -286,7 +286,7 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td><code>interval</code></td>
       <td>number</td>
       <td><code>5000</code></td>
-      <td>The amount of time to delay between automatically cycling an item. If false, carousel will not automatically cycle.</td>
+      <td>The amount of time in ms to delay between automatically cycling an item. If false, carousel will not automatically cycle.</td>
```
XhmikosR, line 289: @mdo or `in milliseconds`, your call. We should also backport this.

mdo, line 289:
```suggestion
      <td>The amount of time in milliseconds to delay between automatically cycling an item. If false, carousel will not automatically cycle.</td>
```

XhmikosR:
@mdo I think we need to change all instances like in popovers and toasts too, so that we are consistent.

@manonthemat can you take care of it please?

site/content/docs/5.0/components/carousel.md
patrickhlauke: counter-counter proposal: "the amount of delay (in ms) before cycling an item"

manonthemat:
> @manonthemat can you take care of it please?

Will do.

XhmikosR:
@manonthemat can you also change the other instances in docs?

XhmikosR:

*This is an automated reply*

PR #31948, opened 2020-10-21
Examples/Floating-labels: fix bad behavior with autofill

eduardogoncalves:
If the browser has already credentials on the autocomplete, the layout of <input> will breaks.

image


site/content/docs/5.0/examples/floating-labels/floating-labels.css
```diff
@@ -76,7 +76,8 @@ body {
   color: #777;
 }
 
-.form-label-group input:not(:placeholder-shown) ~ label {
+.form-label-group input:not(:placeholder-shown) ~ label,
+.form-label-group input:-webkit-autofill ~ label {
```
ffoodd, line 80: Since [`:-webkit-autofill` isn't supported](https://caniuse.com/?search=%3A-webkit-autofill) in all [our supported browsers](https://github.com/twbs/bootstrap/blob/main/.browserslistrc) (Firefox and Edge 16 <> 79), this selector invalidates the whole declaration block for some of them.

There's [an internal pseudo-class in Firefox (`:-moz-auto-fill`)](https://bugzilla.mozilla.org/show_bug.cgi?id=1355438) but it doesn't seem to be available for authors as of now.

So in order to apply this as a progressive enhancement, we need to dedupe those selectors (and duplicate the rules).

ffoodd (changes requested):

mdo:
Is this nullified by #30449

eduardogoncalves:
Hello @mdo, this issue is still present int #30449. The issue appears when browser autofill the field.

image

site/content/docs/5.0/examples/floating-labels/floating-labels.css
XhmikosR: Does the current autoprefixer version support this property? I haven't tried it yet.

MartijnCuppens:
@eduardogoncalves, what version of Chrome is that? I can't reproduce this on v86 (mac & windows)

eduardogoncalves:
Hi @MartijnCuppens it's  version 86.0.4240.111 running on Windows 10 Pro 2004 19041.572

image

site/content/docs/5.0/examples/floating-labels/floating-labels.css
eduardogoncalves: Hi @XhmikosR, where can I learn on how to check this info?

eduardogoncalves:
I made a video about to demonstrate it, because if you click in any part of the screen, it fixes the issue. But if you just scroll the page, you can see the issue.

https://youtu.be/FUoBUuhYJDY

patrickhlauke:
are you running some form of autofill extension or similar? can't replicate the behavior, at least not on the https://deploy-preview-30449--twbs-bootstrap.netlify.app/docs/5.0/forms/floating-labels/ page

eduardogoncalves:
@patrickhlauke I'm using chrome's default credential manager.
image


To make it work, I typed some text into email and password input, and asked the browser to remember that info.
image

patrickhlauke:
ah, hadn't tried explicitly saving the data once typed in (i usually rely on chrome saving credentials on an actual submit). ok, i can see the issue now. it does look like a bug in Chrome

patrickhlauke:
indeed, just tried a rough minimal test case, and chrome has a bug there where it doesn't evaluate the `:placeholder-shown` until the first interaction/click on the page (maybe for security reasons?)

https://codepen.io/patrickhlauke/pen/gOMgLxR / https://cdpn.io/patrickhlauke/debug/gOMgLxR

eduardogoncalves:
Adding style rules to `input:-webkit-autofill` seems to prevent this from happen.

site/content/docs/5.0/examples/floating-labels/floating-labels.css
mdo: https://github.com/postcss/autoprefixer/issues/626 is outdated by a few years, unsure if there's anything else for it.

patrickhlauke: note that Firefox doesn't have this bug in the first place, so autoprefixing may not be an issue here (for now anyway, unless it turns out Chrome are doing this on purpose as a security measure, and Firefox later copies the behavior)

mdo:
This will need to be retargeted to v4-dev I think, and then applied to the floating labels PR at #30449. Any interest in handling that @eduardogoncalves?

eduardogoncalves:
@mdo, ok, I can do it. Do you mean edit `_floating-labels.scss ` in `v5-floating-labels` branch?
https://github.com/twbs/bootstrap/blob/v5-floating-labels/scss/forms/_floating-labels.scss

mdo:
@eduardogoncalves Hah, yeah, testing it out for v5 in that PR. Would you mind doing this PR for v4-dev though? We can fix this issue for v4 users that way. You might be able to retarget this PR to `v4-dev`, but unsure if there are any major differences between the example in v5 vs v4.

eduardogoncalves:
Too many conflicts while retargeting to v4-dev...

eduardogoncalves:
close in favor of #31979

PR #31953, opened 2020-10-22
Fix input group border radii

### LINKED ISSUE #25110: Missing border radius on input group with validation feedback

MartijnCuppens:
Fixes https://github.com/twbs/bootstrap/issues/25110 by requiring a `.has-validation` class on input groups with validation (tooltips/feedback).

Closes https://github.com/twbs/bootstrap/pull/31666

Preview:
https://deploy-preview-31953--twbs-bootstrap.netlify.app/docs/5.0/forms/input-group/
https://deploy-preview-31953--twbs-bootstrap.netlify.app/docs/5.0/forms/validation/#server-side
https://deploy-preview-31953--twbs-bootstrap.netlify.app/docs/5.0/forms/validation/#tooltips

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -128,9 +128,11 @@
 
 // stylelint-disable-next-line no-duplicate-selectors
 .input-group {
-  > :not(:last-child):not(.dropdown-toggle):not(.dropdown-menu),
-  > .dropdown-toggle:nth-last-child(n + 3) {
-    @include border-right-radius(0);
+  &:not(.has-validation) {
+    > :not(:last-child):not(.dropdown-toggle):not(.dropdown-menu),
+    > .dropdown-toggle:nth-last-child(n + 3) {
+      @include border-right-radius(0);
+    }
   }
 
   > :not(:first-child):not(.dropdown-menu) {
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -135,7 +135,12 @@
     }
   }
 
-  > :not(:first-child):not(.dropdown-menu) {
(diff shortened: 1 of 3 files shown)


mdo (approved):
Damn, what a doozy of a PR. Props for a fairly elegant solution here.

mdo:
This could be backported, yeah? Technically not a breaking change if we added to v4.6.0 since we're adding a class that fixes a longstanding bug.

MartijnCuppens:
Hmm, this is still not the solution we wanted, there are still some issues like these:
image

I think we'll need a JS solution so that we can move the validation messages out of the `input-group`.

MartijnCuppens:
Ok, nevermind, I think I nailed it now 😄

> This could be backported, yeah? Technically not a breaking change if we added to v4.6.0 since we're adding a class that fixes a longstanding bug.

Indeed, we're adding a class to fix the issue. That's not a breaking change since it was already broken 😛

ffoodd:
LGTM! Shouldn't this be added to the migration guide? Or backporting it to v4 means this is not needed?

ffoodd (approved):

mdo:
> LGTM! Shouldn't this be added to the migration guide? Or backporting it to v4 means this is not needed?

I think it'd be helpful to mention in Migration because people _want_ this to be a bugfix or breaking change.

MartijnCuppens:
Updated the migration guide and opened #31962 to backport this to v4

PR #31955, opened 2020-10-22
Drop custom file upload plugin in favor of CSS solution

### LINKED ISSUE #30265: `bs-custom-file-input` doesn't work anymore since the form updates

MartijnCuppens:
CSS only alternative solution for https://github.com/twbs/bootstrap/pull/31085.
Advantages:
- No js dependencies
- No need to add translated button labels
- Labels are automatically adjusted when with `<input type="file" multiple>` 

TODO:
- [x] ~Wait until legacy Edge is dropped~ This isn't really necessary, since legacy Edge already has a decent fallback   
- [x] Extend migration guide
- [x] Check if the alpha version is correct
- [x] Fix border radii on iOS

https://deploy-preview-31955--twbs-bootstrap.netlify.app/docs/5.0/forms/form-control/#file-input
https://deploy-preview-31955--twbs-bootstrap.netlify.app/docs/5.0/forms/validation/#supported-elements

Closes #30265
Closes #31085


site/content/docs/5.0/migration.md
```diff
@@ -250,7 +250,7 @@ Changes to Reboot, typography, tables, and more.
 - Updated file input component with the same overall design, but improved HTML.
   - Refactored `.form-file` markup to resolve some visual bugs while allowing translation and button text changes via HTML instead of CSS.
   - Dropped native `.form-control-file` and `.form-control-range` components entirely.
-  - Renamed `.custom-file` to `.form-file` (including variables).
+  - Renamed `.custom-file` to `.form-file` (including variables). **Watch out:** `.form-file` was later on dropped in `v5.0.0-alpha4`, now you can use `.form-control`.
```
mdo, line 253:
```suggestion
  - Renamed `.custom-file` to `.form-file` (including variables). **Watch out:** `.form-file` was later on dropped in `v5.0.0-alpha3`, now you can use `.form-control`.
```

mdo:
Thoughts on adding `cursor: pointer` here? I know it's not required, but compared to `.form-control` on textual inputs, where the cursor changes to indicate interactivity, there's no change here.

patrickhlauke:
> Thoughts on adding `cursor: pointer` here? I know it's not required, but compared to `.form-control` on textual inputs, where the cursor changes to indicate interactivity, there's no change here.

this old chestnut again ;)

as we do it on buttons by default, sure. also worth adding the `<input type="color">` ?

MartijnCuppens:
> Thoughts on adding `cursor: pointer` here?

Fine by me. Only on the button or the whole input?

site/content/docs/5.0/migration.md
MartijnCuppens: Yeah, I didn't know how much prio this had. We should also wait for legacy Edge support to be dropped, so I used alpha4. I'll add this as a TODO.

patrickhlauke:
everything that's clickable, so the whole thing i'd say

patrickhlauke:
also, good work on this btw. looks/works nicely

MartijnCuppens:
Ok, should we add hover states? If yes, what kind of hoverstate? Shading the button background color maybe?

MartijnCuppens:
I've added a subtle hover state

patrickhlauke (approved):

mdo:
@MartijnCuppens this is amazing. Love the little hover state.

alecpl:
There's some padding issue in Firefox 82 on Ubuntu.
input-bug

MartijnCuppens:
@alecpl, 
Nice catch
That seems to be caused by these 5px:
https://searchfox.org/mozilla-central/rev/0b7007a23bc16c857f894140e12f307bfeef2fdd/layout/style/res/forms.css#494

I've now added a `-moz-margin-end` property to compensate these 5px.

scss/forms/_form-control.scss
```diff
@@ -23,6 +23,14 @@
   @include box-shadow($input-box-shadow);
   @include transition($input-transition);
 
+  &[type="file"] {
```
patrickhlauke, line 26: while not related to this, mind if we slip in `&[type="color"]` here as well?

patrickhlauke: otherwise, to keep it clean, happy to wait for this to be merged and then adding it in a new PR that's targeted

alecpl:
@MartijnCuppens The fix is not working. And I don't see how a "horizontal" margin-end could fix a "vertical" issue.

alecpl:
Maybe I wasn't clear about the issue description. I'm talking about the space (1 or 2px) between the button and bottom border.

MartijnCuppens:
@alecpl, the issue I fixed was the additional padding between the button and the input text, see this:
image

I can't reproduce the issue you have over there on Mac or Windows. It's also rather small (I didn't spot it until you mentioned what was wrong), so I'm not sure this is a real deal breaker. Thoughts @twbs/team?

ffoodd:
I can see this on Ubuntu, only on Firefox. Feels weird, can't find anything yet. Seems related to `line-height` but I can't spot any reason for now…

alecpl:
This is related to font-family. If I remove "font-family:inherit" from the input the font will change to "Noto Sans" and the issue disappears.

MartijnCuppens:
Hmm, I can't really debug this myself here, if anyone can come up with a fix, please share 🙂

MartijnCuppens:
I zoomed in a bit and detected the error. Here, it was caused by the `min-height` rounding. Do you guys still have this issue when you remove `min-height` from the file input?

ffoodd:
@MartijnCuppens Doesn't seem to change anything, sadly. But maybe changing it in the devtools is not the best way?

alecpl:
I've found a fix: `.form-control::file-selector-button { font-family: inherit; }`. It can be also `font-family: unset`.

mdo (approved):
Love this PR! :D

mdo:
Also going to cross reference that #31993 fixes the issues @alecpl mentions here.

alecpl:
@mdo, #31993 does not fix my issue, need the `font-family:inherit` fix mentioned in my previous comment.

A few more words about the issue. Looks like firefox does not inherit font-family on the file input button from the input. It needs to be told to do so explicitely. Without the fix the computed font-family is "Noto Sans" on my computer. This font's line-height is 1px higher than what I get when I add the inherit rule.

ffoodd:
Very nice finding @alecpl, I confirm it solves the issue. 

We should find some reference to document this behaviour, either a bugzilla link or something… The [MDN page for `::file-selector-button` does not mention this](https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button) but made me notice the need for `::-webkit-file-upload-button` version too: does Autoprefixer handles this one?

- [Current spec discussion](https://github.com/w3c/csswg-drafts/issues/5049)
- **(!)** [Bugzilla updated 23 days ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1635675)
- My findings: I guess it relates to [how the font is set](https://hg.mozilla.org/mozilla-central/rev/ff6c38ffa425#l7.104): see how the pseudo element gets `-moz-button` font-family directly?) however it should apply to every [Firefox > 82 out there, which rolled out only a few days ago](https://www.mozilla.org/en-US/firefox/82.0/releasenotes/). I'll cross-ref a link to bugzilla later.

XhmikosR:
We should ask upstream if it's not handled. And I'll try to move forward
with the postcss update in the next days.

ffoodd:
FWIW the [bug I filled against Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1673819) got a very clear explanation (mainly, this pseudo-element should look like a button like it always did, not like an input).

I guess this "fix" @alecpl found could move to our rebot, and maybe suggested to Normalize and friends.

MartijnCuppens:
> We should find some reference to document this behaviour

It's shadow DOM, so all properties are reset to the default behaviour. So basically we should keep in mind these things need to re-reset (font size seems to be inherit by default though): 
https://github.com/twbs/bootstrap/blob/3e2f9ab237e0e338fbe497213ad2353e59fbee9c/scss/_reboot.scss#L431-L434

#31993 seems to be unrelated, for the issue @alecpl reported, but does seem to fix issues with errors in rounding when zoomed in/out which resulted in the same issue visually.

Thanks for the fix @alecpl, this was indeed needed, not only for FF, but also for Chrome (see https://codepen.io/MartijnCuppens/pen/yLJpmvE?editors=1100)

MartijnCuppens:
Hmm yeah, maybe we should indeed move this to reboot, I'll fix that

MartijnCuppens:
Ok, fixed the font-family (and all other font related properties) inheritance and also reported the 5px thing with Firefox, so we might be able to remove that dirty fix we now have in the future (https://bugzilla.mozilla.org/show_bug.cgi?id=1673895).

ffoodd:
> It's shadow DOM, so all properties are reset to the default behaviour.

Not really, as you may see in Firefox sources I linked earlier: https://hg.mozilla.org/mozilla-central/rev/ff6c38ffa425#l7.104 — only a few properties are set (and for example `::placeholder` does not require anything to inherit the font).

But that was fun to search, and a very nice PR! Great job @MartijnCuppens :100:

ffoodd (approved):
LGTM!

mdo (approved):
:shipit: when ready!

MartijnCuppens:
Shipped 🚢

I'll keep an eye on https://bugzilla.mozilla.org/show_bug.cgi?id=1673895 if FF is planning to change how they set their spacing

Johann-S:
That's amazing @MartijnCuppens nice work 👍

PR #31961, opened 2020-10-24
Remove IE entirely, add Bootstrap version badge

techboyg5:
* Remove IE entirely
* Add Bootstrap version badge


XhmikosR:
Thanks but the there's a pending PR for the browserslist and the second one we'll pass.

PR #31964, opened 2020-10-25
Remove background-clip from .btn-close

### LINKED ISSUE #31816: Close icon is clipped
Something I noticed on a non-retina screen, the close icon is clipped since we switched to `.btn-close`:
image

mdo:
Hopefully fixes #31816.


patrickhlauke:
yes, seems to fix it for me at least in windows/chrome (oddly, the clipping seemed to happen only happens when i have the browser maximised, so probably something to do with the rounding/aliasing and the browser getting confused - but this fix solves it even for maximised anyway)

patrickhlauke (approved):

PR #31965, opened 2020-10-25
Rename Navs to Navs and Tabs

### LINKED ISSUE #31378: Rename "Navs" in documentation to "Navs and tabs" or create separate "Dynamic tabs" document

mdo:
Fixes #31378

/cc @patrickhlauke


patrickhlauke:
hah, had that issue of mine on my perennial "need to really do a PR when i find a minute..." pile. thanks, will review now.

patrickhlauke:
build fails. i believe you also need to change the `title:` in the renamed navs-tabs.md file

patrickhlauke:
i believe it also needs changing in the overview.md file for components (see the changes I had to make back when i renamed checks) https://github.com/twbs/bootstrap/commit/490acf60ca149e88e669f2b3f157dd326dad9573

patrickhlauke (approved):

site/content/docs/5.0/components/navs-tabs.md
```diff
@@ -1,8 +1,9 @@
 ---
 layout: docs
-title: Navs
+title: Navs and tabs
 description: Documentation and examples for how to use Bootstrap's included navigation components.
 group: components
+aliases: "/docs/5.0/components/navs/"
```
XhmikosR, line 6: I don't think we need this alias since v5 is a major version. I need to remember to drop any redundant redirects when we hist stable.

patrickhlauke: might still be nice to keep around in case people come from older versions and just hacked the url with `5.0`? not precious about it either way though

XhmikosR: We always followed this pattern when we released major/minor versions since the URLs change. I'd rather keep things clean because we can easily end up with hundreds of redirects.

PR #31970, opened 2020-10-26
fix body offsetTop value issue

gbujanski:
Regarding #31569 &  #31950

Just like Mikołaj wrote: "scrollSpy reads the wrong window offsetTop and marks the wrong link as active".  This should fix it.


XhmikosR:
Thanks for the PR, but we need a test case which would fail on main and pass here to make sure the patch is right :)

xuyihan-china:
bootstrap's scrollSpy's API document is difficult to use ,please improve

js/src/dom/manipulator.js
```diff
@@ -62,8 +62,8 @@ const Manipulator = {
     const rect = element.getBoundingClientRect()
 
     return {
-      top: rect.top + document.body.scrollTop,
-      left: rect.left + document.body.scrollLeft
+      top: rect.top + (document.body.scrollTop || document.documentElement.scrollTop),
```
emilio, line 65: I think you really want `document.scrollingElement.scrollTop`.

XhmikosR:
@gbujanski can you add the needed test and also check out @emilio's comment?

XhmikosR:

*This is an automated reply*

PR #31972, opened 2020-10-26
Drop Legacy Edge support. 

### LINKED ISSUE #32017: v5: Drop unused polyfills
It seems that https://github.com/twbs/bootstrap/blob/main/js/src/dom/polyfill.js is redundant based on the currently supported browsers. We should remove any unneeded polyfills. Any help is welcome. 🙂

XhmikosR:
Needs thorough review in case something is wrong, for example some of the vendor prefixes.

~~Also, it seems we can drop more code like the Polyfills, but I need some help with that.~~

Preview: <https://deploy-preview-31972--twbs-bootstrap.netlify.app/>

Closes #32017 along the way


XhmikosR:
Split from #30986. I think we are going to leave any further JS code cleanup for later.

@twbs/css-review please make sure I didn't remove something that's still needed 🙂

mdo (approved):
Looks great for CSS! Stoked to drop so much for even the form range component. Crazy how much those pseudo elements add.

js/tests/visual/tooltip.html
```diff
@@ -12,8 +12,6 @@
         height: 50px;
         border: 1px solid;
         margin-left: 50px;
-        -webkit-transform: rotate(270deg);
```
XhmikosR, line 15: I'm going to split this to a separate PR since it's not related to this patch

site/content/docs/5.0/migration.md
```diff
@@ -9,6 +9,10 @@ toc: true
 
 ## v5.0.0-alpha3
 
+### Browser support
+
+- Dropped support for Microsoft Legacy Edge. See [here](#browser-support-1) for the previous browser support changes.
```
XhmikosR, line 14: Not sure if there is any other way to do this so that it's foolproof; now it's fragile and will break if we add another `Browser support` heading.

MartijnCuppens:
Still spotted some Edge references in the md files (`browsers-devices.md` for example). CSS wise it looks ok.

XhmikosR:
I intentionally kept those, should I remove them?

MartijnCuppens:
Yeah, this seems to be incorrect now:
https://github.com/twbs/bootstrap/blob/d2c13e0f3187ed00f62fb3905cc34b1ffd7d3010/site/content/docs/5.0/getting-started/browsers-devices.md#L11

Just spotted this could also be removed:
https://github.com/twbs/bootstrap/blob/d2c13e0f3187ed00f62fb3905cc34b1ffd7d3010/scss/forms/_floating-labels.scss#L66-L85

XhmikosR:
All right, I'll update the docs tomorrow too.

XhmikosR:
PR updated. The only thing I'm not sure about is this:

>This approach works in the new Microsoft Edge built on Chromium and gracefully degrades on older versions.

ffoodd:
Sidenote here about our philosophy: shouldn't we use progressive enhancement as a basis, to ensure nothing's broken on unsupported browsers? I'm thinking of our new floating labels—as far as I remember they were barely unusable no Edge before I added a few things— but there might be more cases.

Could be as easy as wrapping our code in `@supports()` rules.

mdo:
Good question @ffoodd. I think our approach needs to be more about embracing the future, and less about making the old stuff work. At least with our major version releases. If folks need to support older browsers, they should continue using v4. We should codify that maybe in these docs and update that snippet that @XhmikosR highlighted.

XhmikosR:
I think this is ready for a final review. I made #32057 for the polyfills, but I could also move that change here.

I assume we all stick with the decision to drop Legacy Edge support, right?

Unfortunately, there's no sweet spot between size/redundant code and progressive enhancement AFAICT; we either keep some dead code or forget about progressive enhancement.

MartijnCuppens (approved):

Johann-S (approved):
LGTM 👍 happy to see those polyfills gone

XhmikosR:
Merged the polyfills removal to this PR so that we have everything in one place.

XhmikosR:
I think the only thing left is this: https://github.com/twbs/bootstrap/pull/31972#issuecomment-719347820

> 
> 
> PR updated. The only thing I'm not sure about is this:
> 
> > This approach works in the new Microsoft Edge built on Chromium and gracefully degrades on older versions.

mdo:
> This approach works in the new Microsoft Edge built on Chromium and gracefully degrades on older versions.

I'd say remove that sentence entirely @XhmikosR.

rohit2sharma95 (approved):

PR #31978, opened 2020-10-26
Run sass-migrator with forward flag

mdo:
Just testing things out.


adamkudrna:
@mdo Hello! I've just migrated two projects to Sass modules (a website and a [React UI](https://github.com/react-ui-org/react-ui/pull/199) library) and I'm open to help with the migration here.

By the way, it will be a **breaking change** for SCSS consumers of Bootstrap. I haven't come across any Bootstrap's plans with Sass modules so please forgive me if there is one. Am I right when I say Bootstrap 6 must be released before October 2022 when support for `@import` is dropped? I'm also trying to figure out a way how to support both `@import` and Sass modules in v5.x but I cannot imagine that at the moment—and v5 is already on the way…

mdo:
Yeah I just don't see how we can make the change this late in the game. We're about to launch our beta, so unfortunately I think we need to keep things as they are and make further infrastructure/tooling changes in v6.

PR #31981, opened 2020-10-27
Update list-group.md

XhmikosR:
@patrickhlauke this one I meant on Slack; not sure if the div is supposed to be in the snippet or if we should remove it?

* Before: https://twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#using-data-attributes
* After: https://deploy-preview-31981--twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#using-data-attributes

This is a regression after #31806


patrickhlauke:
yeah the outer div is part of the code, so should be inside the fenced code block

patrickhlauke (approved):

XhmikosR:
It seems this affects v4 too, I'll manually apply the fix in my v4-dev-xmr branch.

PR #31992, opened 2020-10-27
Added table sizes. Such as on buttons

Mopsan:
https://deploy-preview-31992--twbs-bootstrap.netlify.app/docs/5.0/content/tables/#sizes

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -60,12 +60,14 @@
 
 .table-lg {
   > :not(caption) > * > * {
+    // stylelint-disable-next-line selector-max-universal
     padding: $table-cell-padding-y-lg $table-cell-padding-x-lg;
   }
 }
 
 .table-sm {
   > :not(caption) > * > * {
+    // stylelint-disable-next-line selector-max-universal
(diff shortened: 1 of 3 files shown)


Mopsan:
Maybe it may be better to set the variables $table-cell-padding-y-lg and $table-cell-padding-x-lg as .75rem

MartijnCuppens:
To be honest, I never really needed larger tables anywhere. This is different for the larger buttons, they come in handy in CTA block where you want to emphasize the button.

Thoughts @twbs/team?

mdo:
Agreed. Smaller makes sense for super dense tables, but I don't see the use case for larger ones just yet. Could be convinced if there's some good examples out there?

mdo:
Closing per above comment.

Tangent thought we might want to explore later: adding some CSS custom properties to component likes `.table` for their padding. That way someone could override at the `<table>` level with ease.

```html
<table class="table" style="--padding: .25rem;"
  ...
</table>
```

```scss
.table {
  --padding: .5rem;
  // ...

  td,
  th {
    padding: var(--padding);
  }
}
```

PR #32003, opened 2020-10-29
CI: switch to Node.js 14.

XhmikosR:
~~I'll try to split the patches later~~

Node.js 14 hit LTS https://nodejs.org/en/about/releases/


mdo (approved):

XhmikosR:
GitHub is stuck even though I've rebased the branch and it still shows that there are conflicts :/

PR #32004, opened 2020-10-29
docs(3.3): fix input-groups link

simon04:
Currently, https://getbootstrap.com/docs/3.3/css/#forms-example links to https://getbootstrap.com/components/#input-groups which  redirects to https://getbootstrap.com/docs/4.5/components/alerts/


XhmikosR:
This is already fixed in 3.4, and we usually don't bother with old docs. @mdo your call

simon04:
Searching for "getbootstrap 3" on https://duckduckgo.com/?q=getbootstrap+3&ia=web directed me to the 3.3 docs, https://getbootstrap.com/docs/3.3/ –

XhmikosR:
Sure, but that doesn't mean anything, really. Every release includes doc fixes which are applied in new releases. We don't really fix the old docs unless it's something important, and this is already fixed in 3.4 which is the latest 3.x release.

mdo:
Feels like a harmless fix we can push to old docs. I'm fine seeing it move ahead @XhmikosR.

XhmikosR (approved):

PR #32008, opened 2020-10-29
Update Edge's Rendering Engine on CONTRIBUTING.md

coliff:
Also updated the URL to prevent a HTTP redirect.


.github/CONTRIBUTING.md
```diff
@@ -106,7 +106,7 @@ Sometimes bugs reported to us are actually caused by bugs in the browser(s) them
 | Mozilla       | Firefox                      | Gecko            | https://bugzilla.mozilla.org/enter_bug.cgi                            | "Core" is normally the right product option to choose.   |
 | Apple         | Safari                       | WebKit           | https://bugs.webkit.org/enter_bug.cgi?product=WebKit                  | In Apple's bug reporter, choose "Safari" as the product. |
 | Google, Opera | Chrome, Chromium, Opera v15+ | Blink            | https://bugs.chromium.org/p/chromium/issues/list                      | Click the "New issue" button.                            |
-| Microsoft     | Edge                         | EdgeHTML         | https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/ |                                                          |
+| Microsoft     | Edge                         | Blink            | https://developer.microsoft.com/en-us/microsoft-edge/                 |                                                          |
```
XhmikosR, line 109: @coliff Do you know if they have a separate bug tracker nowadays?

coliff: I spent about 15 minutes looking for one today for this PR :-)   The official stance is that users should go to Help > Send Feedback from their browser. I was considering adding that as a Note in the table. what do you think?

XhmikosR: Yup, might be better to mention this in the Notes :)

coliff: cool. let's agree on the wording first. How about:

Go to "Help > Send Feedback" from the browser

XhmikosR: Yeah sounds good. I don't think many people really read the file, unfortunately 😛

mdo (approved):

PR #32011, opened 2020-10-29
polyfill.js minor tweaks

XhmikosR:
* rename function to `supportsScopeQuery`
* use a ternary in `findOne` function return value

### DIFF
--- a/js/src/dom/polyfill.js
+++ b/js/src/dom/polyfill.js
@@ -27,7 +27,7 @@ const defaultPreventedPreservedOnDispatch = (() => {
 })()
 
 const scopeSelectorRegex = /:scope\b/
-const supportScopeQuery = (() => {
+const supportsScopeQuery = (() => {
   const element = document.createElement('div')
 
   try {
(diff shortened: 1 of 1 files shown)


js/src/dom/polyfill.js
```diff
@@ -71,11 +71,7 @@ if (!supportScopeQuery) {
 
     const matches = find.call(this, selector)
 
-    if (typeof matches[0] !== 'undefined') {
-      return matches[0]
-    }
-
-    return null
+    return matches[0] ? matches[0] : null
```
acshef, line 74: The old logic checked if `matches[0]` is `undefined`, while new logic checks if `matches[0]` is falsey. *Assuming that change is okay* (and from what I can tell about return types of `find()`, I believe that's true), it's also possible to be DRYer with:
```js
return matches[0] || null // And if you're using the fanciest of ES/JS, you could even do `?? null` which more closely fits the old logic
```

XhmikosR: We usually prefer the ternary operator than the shortcut one. Also, we don't make use of the nullish operator in the codebase yet.

Johann-S (approved):

PR #31346, opened 2020-07-21
Turn off scroll anchoring for accordions

patrickhlauke:
New default behavior for scroll anchoring (rolled out in Chrome 84?) leads to unsightly/odd accordion interactions - see #31341
This rule suppresses this new behavior and reverts back to the old way.

See https://drafts.csswg.org/css-scroll-anchoring/

Closes #31341

PR #32013, opened 2020-10-30
Add dedicated accordion component

### LINKED ISSUE #25811: .accordion-flush

### LINKED ISSUE #28134: Accordion: card-header border bottom hidden by card-body.

### LINKED ISSUE #28873: position:sticky inside accordion

### LINKED ISSUE #30015: Create standalone accordion component

mdo:
This PR replaces and adds onto #30042. The goal of this is to replace our half-baked card accordion with a fully functioning accordion component, powered by the Collapse plugin. While this adds some CSS, personally I think it makes accordions more widely accessible (e.g., fewer requirements to creating the component).

According to the origin PR by @gijsroge, this does the following:

- Fixes #30015, the original feature request issues
- Fixes #28134 by moving the border-bottom from the header to a border-top on the body
- Fixes #28873 by removing the `overflow: hidden;` on the .accordion-item
- Fixes #25811 by adding an `.accordion-flush` variant (in addition to `.accordion-striped`, which may not be necessary IMO)

Also from that PR, there were some remaining todos. Design and code are largely final now with my latest changes here. Looking through https://www.w3.org/TR/wai-aria-practices-1.1/#accordion, I think we're largely in order. There's also this handy dandy [example accordion](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html) they provide.

**Preview:** https://deploy-preview-32013--twbs-bootstrap.netlify.app/docs/5.0/components/accordion/

---

Some open questions from me now moving this forward:

- [x] Do we need striping here? I think not, and may remove it shortly.
- [x] Any other variations, documentation, etc we should be considering?

/cc @twbs/css-review

### DIFF
--- a/scss/_accordion.scss
+++ b/scss/_accordion.scss
@@ -1,0 +1,124 @@
+//
+// Base styles
+//
+
+.accordion-button {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  padding: $accordion-button-padding-y $accordion-button-padding-x;
+  @include font-size($font-size-base);
+  color: $accordion-button-color;
+  background-color: $accordion-button-bg;
+  border: solid $accordion-border-color;
+  border-width: $accordion-border-width $accordion-border-width 0;
+  @include border-radius(0);
+  overflow-anchor: none;
+
+  &:not(.collapsed) {
+    color: $accordion-button-active-color;
+    background-color: $accordion-button-active-bg;
+
+    &::after {
+      background-image: escape-svg($accordion-button-active-icon);
+      transform: $accordion-icon-transform;
+    }
+  }
+
+  // Accordion icon
+  &::after {
+    flex-shrink: 0;
+    width: $accordion-icon-width;
+    height: $accordion-icon-width;
+    margin-left: auto;
+    content: "";
+    background-image: escape-svg($accordion-button-icon);
+    background-repeat: no-repeat;
+    background-size: $accordion-icon-width;
+    transform-origin: center center;
+    @include transition($accordion-icon-transition);
+  }
+
+  &:focus {
+    position: relative;
+    outline: 0;
+    box-shadow: $btn-focus-box-shadow;
+  }
+}
+
+.accordion-header {
+  margin-bottom: 0;
+}
+
+.accordion-item {
+  @include border-radius($accordion-border-radius);
+
+  &:last-of-type {
+    .accordion-button {
+      border-bottom-width: $accordion-border-width;
+
+      // Only set a border-radius on the last item if the accordion is collapsed
+      &.collapsed {
+        @include border-bottom-radius($accordion-border-radius);
+      }
+    }
+
+    .accordion-body {
+      border-width: 0 $accordion-border-width $accordion-border-width;
+      @include border-bottom-radius($accordion-border-radius);
+    }
+  }
+
+  &:first-of-type {
+    .accordion-button {
+      @include border-top-radius($accordion-border-radius);
+    }
+  }
+}
+
+.accordion-body {
+  padding: $accordion-body-padding-y $accordion-body-padding-x;
+  border: solid $accordion-border-color;
+  border-width: $accordion-border-width $accordion-border-width 0;
+}
+
+
+// Flush accordion items
+//
+// Remove borders and border-radius to keep accordion items edge-to-edge.
+
+.accordion-flush {
+  .accordion-button {
+    border-right: 0;
+    border-left: 0;
+    @include border-radius(0);
+  }
+
+  .accordion-body {
+    border-width: 0;
+  }
+
+  .accordion-item {
+    border-right-width: 0;
+    border-left-width: 0;
+    @include border-radius(0);
+
+    &:first-of-type {
+      .accordion-button {
+        border-top-width: 0;
+        @include border-top-radius(0);
+      }
+    }
+
+    &:last-of-type {
+      .accordion-button {
+        border-bottom-width: 0;
+        @include border-bottom-radius(0);
+      }
+
+      .accordion-body {
+        border-width: 0;
+      }
+    }
+  }
+}
--- a/scss/_accordion.scss
+++ b/scss/_accordion.scss
@@ -36,7 +36,6 @@
     background-image: escape-svg($accordion-button-icon);
     background-repeat: no-repeat;
     background-size: $accordion-icon-width;
-    transform-origin: center center;
     @include transition($accordion-icon-transition);
   }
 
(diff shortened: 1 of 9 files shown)


scss/_card.scss
```diff
@@ -213,30 +213,3 @@
...
   }
 }
-
-
-//
-// Accordion
-//
-
-.accordion {
-  overflow-anchor: none;
```
Hiws, line 223: I think `overflow-anchor: none;` should be added to `.accordion-button` to avoid reintroducing the issue fixed in https://github.com/twbs/bootstrap/pull/31346

XhmikosR:
@mdo this needs a proper rebase, preferably squashing the original PR's patches into one.

scss/_accordion.scss
```diff
@@ -0,0 +1,124 @@
...
+  &::after {
+    flex-shrink: 0;
+    width: $accordion-icon-width;
+    height: $accordion-icon-width;
+    margin-left: auto;
+    content: "";
+    background-image: escape-svg($accordion-button-icon);
+    background-repeat: no-repeat;
+    background-size: $accordion-icon-width;
+    transform-origin: center center;
```
MartijnCuppens, line 38: This is the default transform-origin, or is there any other reason why this is here?
```suggestion
```

MartijnCuppens, line 38: This is the default transform-origin, or is there any other reason why this is here?
```suggestion
    background-size: $accordion-icon-width;
```

scss/_accordion.scss
```diff
@@ -0,0 +1,124 @@
...
+    @include transition($accordion-icon-transition);
+  }
+
+  &:focus {
+    position: relative;
```
MartijnCuppens, line 43: What's the purpose of this rule?

scss/_accordion.scss
mdo: Leftover from me. I was debugging the original PR's problem of a flashing background resize. Was playing with ideas before I realized the transition was set to `all` and the `background-image` was being swapped mid-transition.

scss/_accordion.scss
mdo: This effectively brings the focused accordion button to the top of the stack—it's like our `z-index: 1` tricks elsewhere I think, minus the `z-index`. Thought it was a little clever, so unsure if we should add the `position` by default and toggle `z-index` instead. Thoughts?

scss/_accordion.scss
mdo: Strange, I can't accept your suggested change. See anything weird on your end? 🤔

MartijnCuppens: Hmm, nope. I think github just can't handle removing lines in comments.

scss/_accordion.scss
MartijnCuppens: I'm ok with this shorter solution, but we might need to explain its purpose in the source code, because it's not self explanatory.

scss/_variables.scss
```diff
@@ -1014,6 +1014,33 @@ $card-img-overlay-padding:          $spacer !default;
...
+
+$accordion-body-padding-y:                $accordion-padding-y !default;
+$accordion-body-padding-x:                $accordion-padding-x !default;
+
+$accordion-button-padding-y:              $accordion-padding-y !default;
+$accordion-button-padding-x:              $accordion-padding-x !default;
+$accordion-button-color:                  $accordion-color !default;
+$accordion-button-bg:                     $accordion-bg !default;
+$accordion-button-active-bg:              tint-color($component-active-bg, 90%) !default;
+$accordion-button-active-color:           $primary !default;
```
MartijnCuppens, line 1034: Color contrast between this color and the background is less than 4.5. We might need to shade this color a bit?

MartijnCuppens:
The bottom border is missing when the last element is collapsing
image

The bottom border is below the focus state here:
image

patrickhlauke:
Just gave this a quick check in terms of accessibility - of course, as it's based on our current one, it's all kosher.

One small request though: would it be possible to ditch the latin placeholder text for some generic english text? Even if it's just something that is self explanatory, like "This is the example content for the accordion panel. It is revealed when the accordion control is activated." or similar? It's not a huge deal, but it avoids having the page all of a sudden switch to absolutely incomprehensible gibberish for screen reader users, for instance.

MartijnCuppens:
Fixed the remaining bugs by adding an additional `accordion-collapse` class.

mdo:
> Fixed the remaining bugs by adding an additional `accordion-collapse` class.

I... think I love you? Thank you for the fixes!

MartijnCuppens (approved):

XhmikosR:
Saved a few more bytes by reordering the SVG attributes :)

patrickhlauke:
> Fixed the remaining bugs by adding an additional `accordion-collapse` class.

oh, i just noticed that includes also changing the latin text. thank you, top stuff! (funnily enough, I've just written up a bug for another project that also uses lorem ipsum for their demos :) )

PR #32014, opened 2020-10-30
v5: Correct displays <input type="date/time"> on iOS

Mopsan:
For detail, see https://stackoverflow.com/questions/26573346/ios-safari-messes-up-input-type-date

### DIFF
--- a/scss/_reboot.scss
+++ b/scss/_reboot.scss
@@ -563,6 +563,15 @@ legend {
   -webkit-appearance: textfield; // 2
 }
 
+// Correct element displays on iOS.
+// See https://stackoverflow.com/a/26573346
+[type="date"],
+[type="time"],
+[type="datetime-local"],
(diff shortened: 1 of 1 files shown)


mdo:
Hmm, I'm not sure we need to add this. @twbs/css-review any thoughts?

MartijnCuppens:
Indeed, we already reset the appearance for `.form-control` here: https://github.com/twbs/bootstrap/blob/6fc35e3231225559a6a4689bfa9727dcfb7fafde/scss/forms/_form-control.scss#L17

XhmikosR:
Wouldn't it make more sense to move the reset to reboot, though?

MartijnCuppens:
> Wouldn't it make more sense to move the reset to reboot, though?

I can imagine someone would prefer the default browser theming sometimes?

PR #32024, opened 2020-10-30
Fix TypeError when Bootstrap is included in `head`

### LINKED ISSUE #31202: TypeError: jQuery present, loading script in head
When using jQuery and loading JS pack from the `<head>` Bootstrap 5 throws an error:

Screenshot 2020-06-30 12 51 36

The offending line:

https://github.com/twbs/bootstrap/blob/f7cf44d3a0815e51364acaba929895dd975131ef/js/src/util/index.js#L176

should check for the presence of body.

Docs do not state that JS bundle needs to be in the body so I conclude this is a bug. Happy to submit a PR if confirmed.

- Operating system and version: macOS Catalina
- Browser and version: Chrome 83.0.4103.116

Demo site:
https://amazing-heisenberg-e74368.netlify.app/ (see console)

tidusIO:
I've added waiting for domContentLoaded event before extending jQuery. That's one possible solution to be sure, that data attributes of body can be read.

Fixes #31202

Preview: <https://deploy-preview-32024--twbs-bootstrap.netlify.app/>

### DIFF
--- a/js/src/dom/event-handler.js
+++ b/js/src/dom/event-handler.js
@@ -5,7 +5,7 @@
  * --------------------------------------------------------------------------
  */
 
-import { getjQuery } from '../util/index'
+import { getjQuery, onDOMContentLoaded } from '../util/index'
 import { defaultPreventedPreservedOnDispatch } from './polyfill'
 
 /**
@@ -14,7 +14,6 @@ import { defaultPreventedPreservedOnDispatch } from './polyfill'
  * ------------------------------------------------------------------------
  */
 
-const $ = getjQuery()
 const namespaceRegex = /[^.]*(?=\..*)\.|.*/
 const stripNameRegex = /\..*/
 const stripUidRegex = /::\d+$/
@@ -282,13 +281,18 @@ const EventHandler = {
     let defaultPrevented = false
     let evt = null
 
-    if (inNamespace && $) {
-      jQueryEvent = $.Event(event, args)
+    if (inNamespace) {
+      onDOMContentLoaded(() => {
+        const $ = getjQuery()
+        if ($) {
+          jQueryEvent = $.Event(event, args)
 
-      $(element).trigger(jQueryEvent)
-      bubbles = !jQueryEvent.isPropagationStopped()
-      nativeDispatch = !jQueryEvent.isImmediatePropagationStopped()
-      defaultPrevented = jQueryEvent.isDefaultPrevented()
+          $(element).trigger(jQueryEvent)
+          bubbles = !jQueryEvent.isPropagationStopped()
+          nativeDispatch = !jQueryEvent.isImmediatePropagationStopped()
+          defaultPrevented = jQueryEvent.isDefaultPrevented()
+        }
+      })
     }
 
     if (isNative) {
--- a/js/src/dom/event-handler.js
+++ b/js/src/dom/event-handler.js
@@ -5,7 +5,7 @@
  * --------------------------------------------------------------------------
  */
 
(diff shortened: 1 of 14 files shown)


js/src/util/index.js
```diff
@@ -180,8 +180,15 @@ const getjQuery = () => {
   return null
 }
 
+const onDOMContentLoaded = callback => {
```
Johann-S, line 183: Can you add a unit test here please ?

Johann-S (changes requested):
Hi @tidusIO,

Thanks for your work 👍 

The overall is good, it needs just a few unit tests

js/src/util/index.js
tidusIO: This function is tested now.

tidusIO:
It would be useful having some tests mocking getjQuery or jquery in window object, so that the uncovered parts are covered.

I'm struggling a bit with mocking these parts.

tidusIO:
We still have the problem in event-handler.js, when calling getjQuery().

tidusIO:
I would prefer a way like this: https://github.com/twbs/bootstrap/commit/ff1fa49b88e1c48de603f1613b04650d2f8354d0
It's much cleaner and there shouldn't be no problems. If we'd implement it with event listener, the problem could be, that the user has to wait for DOMContentLoaded at some places too.

js/src/dom/event-handler.js
```diff
@@ -282,13 +281,18 @@ const EventHandler = {
     let defaultPrevented = false
     let evt = null
 
-    if (inNamespace && $) {
-      jQueryEvent = $.Event(event, args)
+    if (inNamespace) {
+      onDOMContentLoaded(() => {
```
Johann-S, line 285: we don't need to wait for the DOM content to be loaded here because we don't register any jQuery plugin here

Johann-S:
it seems something decreased the minimum coverage required

tidusIO:
Do you know, if coverage also includes `js-test-jquery`?

Johann-S:
The coverage doesn't include the jQuery part.
I don't think it's really necessary since jQuery is a very small part of our codebase and the main goal of v5 is to work without jQuery

tidusIO:
Okay, still not familiar enough with bootstrap codebase. I simply overlooked the `/* istanbul ignore if */` statements.

Johann-S (approved):
Thanks for your hard work @tidusIO 👍

js/src/util/index.js
```diff
@@ -180,8 +180,15 @@ const getjQuery = () => {
...
+  if (document.readyState === 'loading') {
```
XhmikosR, line 184: Just to confirm, this should work with defer too?

XhmikosR:
@tidusIO apart from my comment above, can you please make a CodePen with the scripts in `head` so that we can test?

tidusIO:
Yes, I've replied your comment. I can, but probably not before Wednesday.

XhmikosR:
NVM I tested it myself and it works as expected, thanks for the PR!

rpokrovskij:
Bad decision. Jquery plugins should be loaded in developer defined order: there is convention that this order is maintained by `script` loading.   How to control this order after you broke it with domContentLoaded  ? What if every jquery plugin start doing it this way?

NefedovIv:
Rly like last comment. We works with sharepoint onprem, and have a problem there. SP have own event model, many usefull events trigered before DOMContentLoaded without access to jquery bootstrap plagins. We spand a time for only understand that plagins extend performed in postponed format.

PR #32037, opened 2020-11-01
Add .navbar-nav-scroll for vertical scrolling of navbar content

### LINKED ISSUE #23374: Cannot scroll down a collapsed navbar on mobile devices
When visiting a site using a mobile device with a **navbar with .fixed-top**, if the opened navbar-collapse area runs beyond the end on the viewport vertically (due to lots of links etc), you cannot scroll down the the links at the bottom of the navbar as you can in Bootstrap 3.

When you do try to scroll, the body behind the navbar scrolls, with the navbar fixed in place.

This can be fixed by setting `overflow-y: auto;` and adding a `max-height:Xpx` to the `.collapse.show` class and allows you to scroll the navbar-collapse area.

This happens in (the ones I tested) major browsers (Safari, Chrome, both on the latest updates). I haven't yet tried whether this fix breaks anything elsewhere.

mdo:
Fixes #23374.

TODO:

- [x] Fix example

Preview: <https://deploy-preview-32037--twbs-bootstrap.netlify.app/docs/5.0/components/navbar/#scrolling>


site/content/docs/5.0/components/navbar.md
```diff
@@ -479,6 +480,59 @@ Also note that **`.sticky-top` uses `position: sticky`, which [isn't fully suppo
 </nav>
 {{< /example >}}
 
+## Scrolling
+
+Add `.navbar-nav-scroll` to a `.navbar-nav` to enable vertical scrolling when expanded navbar content gets too tall. By default, scrolling kicks in at `75vh` (or 75% of the viewport height), but you may override that with a local CSS custom property. For example, add `style="--height: 65vh;"` to the `.navbar-nav`.
```
XhmikosR, line 485: `70%`?

site/content/docs/5.0/components/navbar.md
```diff
@@ -43,7 +44,7 @@ Here's an example of all the sub-components included in a responsive light-theme
     <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
       <span class="navbar-toggler-icon"></span>
     </button>
-    <div class="collapse navbar-collapse" id="navbarSupportedContent">
+    <div class="collapse navbar-collapse navbar-scroll" id="navbarSupportedContent">
```
ffoodd, line 47: Shouldn't it be `navbar-nav-scroll`?

site/content/docs/5.0/components/navbar.md
mdo: Are you suggesting we change something to `70%`?

XhmikosR: Isn't the default 70%?

patrickhlauke:
nitpick ... not just vertical scrolling, but horizontal scrolling as well, no?

ffoodd:
Just a suggestion here: couldn't we use the [scrolling shadow trick by Lea Verou](https://lea.verou.me/2012/04/background-attachment-local/) here?

Would help to get a visual clue of overflown things.

And maybe some `scroll-snap-*` love too? Don't have a concrete idea but there's probably something to do.

patrickhlauke:
> Just a suggestion here: couldn't we use the [scrolling shadow trick by Lea Verou](https://lea.verou.me/2012/04/background-attachment-local/) here?
> 
> Would help to get a visual clue of overflown things.
> 
> And maybe some `scroll-snap-*` love too? Don't have a concrete idea but there's probably something to do.

i'd suggest a new ticket for this. and agree, would be good to provide a visual hint that there's more that can be reached with scrolling (for platforms where the scrollbar is not shown, like mobile). maybe not a shadow, but perhaps a subtle arrow or similar if that's still possible to do with pure CSS

mdo:
Rebased and updated this PR to prevent the horizontal scrolling issue on expanded navbars. Also documented the potential issue with using `overflow-y` and clipping through `overflow-x` being automatically set by browsers. I think this is good to go and could be nice to include in both v5 and v4.6.

Can y'all take another look?

XhmikosR:
@mdo it seems there are still overflow in certain display widths

image

But it seems to affect the other snippets too. That being said, I'm having trouble visualizing the scroll part here; I don't see the ability to horizontally scroll at all.

ffoodd:
Like @XhmikosR said, even duplicating some nav items it still overflows horizontally. Am I missing something?

mdo:
> That being said, I'm having trouble visualizing the scroll part here; I don't see the ability to horizontally scroll at all.

This is for vertical scroll, not horizontal. On the expanded navbars (that is to say, the horizontal ones) the intent is that you don't see anything different than other navbars. Resetting to overflow visible might not be the right move though (maybe unset or initial?), but I was aiming for something that can be backported to v4.

XhmikosR:
TBH I'm not sure about the feature itself, but if you all think it's something useful, that's fine.

BTW the scrollbar seems to overlap here with the button (FF on Windows x64)

image

mdo:
Scrollbar only overlapped the button focus, but I've tweaked the demo to prevent that. Also updated the docs verbiage a bit, so should be good to go now.

XhmikosR (approved):

baszero:
Is this available in Bootstrap 4.5 ?

PR #32040, opened 2020-11-02
Add callouts about using light colors ideally on a dark background

patrickhlauke:
would be even better to show those buttons on a dark background, but i assume that'd require a lot more logic behind the scenes (to only show the outline buttons that are contrasty enough against white on a white background, and the others on a dark background)

### DIFF
--- a/site/content/docs/5.0/components/buttons.md
+++ b/site/content/docs/5.0/components/buttons.md
@@ -54,6 +54,10 @@ In need of a button, but not the hefty background colors they bring? Replace the
 {{< /buttons.inline >}}
 {{< /example >}}
 
+{{< callout info >}}
+Some of the button styles use a relatively light foreground color, and should only be used on a dark background in order to have sufficient contrast.
+{{< /callout >}}
+
 ## Sizes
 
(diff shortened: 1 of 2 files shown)


ffoodd:
With #30044 merged a few days ago, it might be quite trivial to handle  backgrounds (but would require to use some kind of containers to apply backgrounds). However the callout **si* a good idea :)

ffoodd (approved):

patrickhlauke:
Just realised we should probably do the same for https://v5.getbootstrap.com/docs/5.0/helpers/colored-links/ ... i'll add another commit here

mdo:
Would be very nice to add a background element around them if needed based on the contrast colors. A callout like this feels unnecessary IMO, but I'm down for whatever.

patrickhlauke:
let's do callout just now, and then somebody cleverer than me can work out how to do a conditional bit of light vs dark background

XhmikosR:
@patrickhlauke this needs a manual backport, can you do it please and ping me?

patrickhlauke:
@XhmikosR done - https://github.com/twbs/bootstrap/pull/32054

PR #32042, opened 2020-11-02
add image types to .gitattributes

kurahaupo:
Unfortunately `text=auto` doesn't always make the right guess about image files, so add them explicitly to `.gitattributes`.

(This certainly has burned me while I've been trying to track local modifications.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,16 @@
-# Enforce Unix newlines
+# Enforce POSIX newlines
 * text=auto eol=lf
 
+# Git control files should always be POSIX text files
+.git*   text
+
 # Don't diff or textually merge source maps
-*.map binary
+*.map   binary
 
 bootstrap.css linguist-vendored=false
 bootstrap.js linguist-vendored=false
+
+# Images should not be modified
+
+*.jpg   binary
+*.png   binary


kurahaupo:
Also:
* explicit disposition of `.git*` as text, in case you ever want to change the default again;
* some minor white-space tweaking to align the dispositions to column 8 (unless the filename match is longer);
* mention "POSIX" because it's a _de jure_ ISO standard, not just a _de facto_ one.

XhmikosR:
Thanks but we don't need this. git >= 2.10 is required

kurahaupo:
> Thanks but we don't need this. git >= 2.10 is required

Is there any technical reason why git ≥ 2.10 is mandated?

(And since _I_ certainly need it, so you'll have to apply a reversal on top of any other patches I send you.)

XhmikosR:
>=2.10 is required due to `* text=auto eol=lf`. git 2.10 is out for more than 4 years. It's the same way we don't support Node.js 4.

kurahaupo:
Ok, I fixed the core problem for myself by creating `$HOME/.config/git/attributes` and populating it with
```lang=conf
*	text=auto eol=lf
*.BMP	binary
*.DOC	binary
*.GIF	binary
*.JPG	binary
*.PNG	binary
*.XLS	binary
*.Z	binary
*.ar	binary
*.bmp	binary
*.bz2	binary
*.cpio	binary
*.doc	binary
*.docx	binary
*.gif	binary
*.gz	binary
*.ico	binary
*.img	binary
*.iso	binary
*.jar	binary
*.jpeg	binary
*.jpg	binary
*.o	binary
*.pax	binary
*.png	binary
*.pyc	binary
*.so	binary
*.tar	binary
*.tbz2	binary
*.tgz	binary
*.txz	binary
*.xbm	binary
*.xls	binary
*.xlsx	binary
*.xz	binary
*.zip	binary
.git*	text

```

However the requirement for git ≥ 2.10 doesn't make everything else in this PR obsolete:

* adjusting comments and whitespace is still valid
* it is still helpful to override crazy settings other users might have in `$HOME/.gitconfig` or elsewhere.

Would you like me to refactor this into separate commits?

PR #32060, opened 2020-11-04
File input: inherit padding to avoid duplicating it for sizes

ffoodd:
While backporting it to Boosted, I noticed we duplicate `padding` for `::file-selector-button` but it would benefit from `inherit`ing it from its parent input since we wouldn't have to duplicate it for sizes variant.


MartijnCuppens:
Did this at first, but removed this because incorrect inheriting with validation backgrounds: https://deploy-preview-32060--twbs-bootstrap.netlify.app/docs/5.0/forms/validation/#supported-elements

ffoodd:
Thanks, didn't spotted that!

PR #32062, opened 2020-11-05
Update sass.md

arnissolle:
Using the prefix "~" at the start of the path tells to resolve the import "like a module" from a node module path.


ydmitry:
> Using the prefix "~" at the start of the path tells to resolve the import "like a module" from a node module path.

Well, it's only if you have special "importer" for Sass. E.g. webpack sass loader adds this importer automatically

XhmikosR:
Closing as per the above comment, thanks @ydmitry

PR #32074, opened 2020-11-06
Add .gap utilities

mdo:
Following up on #32066 and the responsive `.btn-block` classes in #31995. If this is going well, I'm hoping we can add a quick commit here (or another PR) to remove `.btn-block` and replace it with `d-grid` and `gap` utilities.

Screen Shot 2020-11-06 at 12 22 54 PM

Preview: https://deploy-preview-32074--twbs-bootstrap.netlify.app/docs/5.0/utilities/spacing/#gap


site/content/docs/5.0/utilities/spacing.md
```diff
@@ -93,3 +95,17 @@ The syntax is nearly the same as the default, positive margin utilities, but wit
   margin-top: -0.25rem !important;
 }
 ```
+
+## Gap
```
XhmikosR, line 99: Maybe we should move Negative margin and Horizontal centering to an `h2`?

mdo: Opted for another organization to separate margin and padding from gap, since the two behave differently.

XhmikosR (approved):

MartijnCuppens (approved):

PR #32077, opened 2020-11-07
v4: switch to Hugo

XhmikosR:
So, apart from Hugo being multiple times faster as we already know, I also switched to using Hugo to handle the docs assets (CSS/JS) like we do on main. Which means, it will be equally simple to work on the v4-dev branch as main. 😃

TODO:

- [x] fix getting started snippets (also search for `site.`)
- [x] replace the remaining Liquid code
- [x] fix jQuery being bundled in the docs JS; not sure how to handle this without hardcoding the files we want to include, since we are using `Resources.match` - **EDIT**: worked around it, not sure if there's a better solution
- [x] make sure we use `docsref` when possible like in main
- [x] Fix migration sidebar link
- [x] Read browserslist, fonts, CSS variables etc
- [x] Deduplicate npm scripts
- [x] Fix duplicate trailing slashes in URLs; also run a full site link checker
- [x] Fix ToC CSS - **EDIT**: added a patch but it seems there are more cases that we need to fix, with nested lists:

  <details><summary>Screenshot</summary>

  | Before | After |
  | ------- | ----- |
  | image | image |
  </details>

  **EDIT2**: it was due to going from an `h2` to `h4`; switched to `h3` and now it's fine

- [x] go through the layout code one more time; fix discrepancies like social logo, classes etc
- [x] Figure out why some classes are missing: **EDIT:** this is right and previously was wrong
  * `card-img` is missing here https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/components/card/#horizontal **this seems to affect main too**
  * `card-img-top` becomes `card-img` here https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/components/card/#card-columns
  * ` class="mr-3"` is missing here: https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/components/media-object/#nesting
- [x] see if this bothers us a lot; I can't remove the newline there (although, we apply margin via CSS, so we could probably remove that if it bothers us a lot) https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/utilities/borders/#border-radius (it's the same on main too so we probably want to keep it)

  <details><summary>Screenshot</summary>

  | Before | After |
  | ------- | ----- |
  | 1-v4-dev | 2-v4-dev-hugo |
  </details>
- [x] ~~see if we are going to go with the homepage patch or if I should drop it~~ - **EDIT**: moved to #32196

- [ ] Update team members
- [x] test all the stuff multiple times

Preview: <https://deploy-preview-32077--twbs-bootstrap.netlify.app/>


coliff:
looking good. 👍  Do you plan to minify the HTML?
The build could either user 'hugo --minify' or could add the following to the hugo config.

```
minify:
  minifyOutput: true
```

https://gohugo.io/getting-started/configuration/#configure-minify

XhmikosR:
> Do you plan to minify the HTML?

Nope, because we want the output to be easily readable which is why we don't do this in main :)

coliff:
I was going to mention this in my comment, but using this is a great option worth trying:
```
minify:
  minifyOutput: true
  tdewolff:
    html:
      keepWhitespace: true
```
it means that hugo serve is always equal to the final compiled output and means its easily both reduced in file size and readable in 'view-source'.

XhmikosR:
Yeah, I know, but I'd rather keep the potential points of failure to a minimum, you know, keep it simple.

We did use HTML minification in v3.3 and it was a PITA when we need to go back to the old docs to change something manually. We might reconsider in the future, but the gains are generally small from HTML minification so they don't justify this.

PS. I do always minify the HTML in most of my projects, but this is not a common project.

XhmikosR:
All right, I tested this as good as I could. I found a few mistakes in the layouts and in the callouts and fixed those. I compared some of the generated HTML files and the diff is:

```
line: 20861(+) 38442(-) files: normal=0, non-versioned=0, modified=117, added=1, deleted=8, conflicted=0
```

Most changes are due to the sidebar not including the inactive group items anymore.

coliff:
Looking good. couple of minor quick things I noticed:

- The code syntax highlighting is not quite as good imo:
https://getbootstrap.com/docs/4.5/getting-started/javascript/#sanitizer
https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/javascript/
Maybe just need to add blue color: #006ee0 to class `.nx` ?

- Looks like an unnecessary extra line break at bottom of the browser list here:
https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/browsers-devices/#supported-browsers

XhmikosR:
>     * The code syntax highlighting is not quite as good imo:
>       https://getbootstrap.com/docs/4.5/getting-started/javascript/#sanitizer
>       https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/javascript/
>       Maybe just need to add blue color: `#006ee0` to class `.nx` ?

Yup, I noticed that also but this affects our main branch too, and I have it in my local TODO. I still need to regenerate the chroma styles in case we are missing more classes; it's been a while since I generated the styles and may be outdated.

>     * Looks like an unnecessary extra line break at bottom of the browser list here:
>       https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/browsers-devices/#supported-browsers

I noticed that also but it's the same thing on main too.

I considered both of these things lower priority but I'll try to address them for main and then here too. I also noticed a few other things while I was working on this branch, so it's a good chance to revisit some things :)

ffoodd:
@XhmikosR  Just checked the missing classes:
- for horizontal card, I think it's on purpose since `.card-img` adds uselesss `border-radius` to the right corners, and changes nothing to the card itself…
- card in columns looks good, not sure if ou fixed anything but the correct markup is applied
- same for nested media objects: I can see margins where needed.

I spent a few minutes going around, everything looks fine! Didn't notice anything.

XhmikosR:
@ffoodd it seems that the previous markup was wrong and this branch has the fixed snippet markup :)

That being said, I wonder if we shouldn't use `example` and duplicate the snippets so that they do include something that's needed without us showing it in the preview?

ffoodd:
I don't think so, since if we do not use it there's some reason. We already had a few issues about snippets and examples not corresponding, and this is indeed very confusing for users.

ffoodd:
@XhmikosR Still reviewing this but wer're missing jQuery in our examples: https://github.com/twbs/bootstrap/commit/3da46e35500c323693fcef949ff1af709fc0c4fe?w=1#diff-bbabe5324b4522d5f784e4c6608ccf9cbb6c4cacc0cdbd68c57474d5533cdc49R264

Cannot seek in detail right now, but apart from line-breaks this is the only thing I noticed. Great job!

XhmikosR:
> 
> 
> @XhmikosR Still reviewing this but we're missing jQuery in our examples: [3da46e3?w=1#diff-bbabe5324b4522d5f784e4c6608ccf9cbb6c4cacc0cdbd68c57474d5533cdc49R264](https://github.com/twbs/bootstrap/commit/3da46e35500c323693fcef949ff1af709fc0c4fe?w=1#diff-bbabe5324b4522d5f784e4c6608ccf9cbb6c4cacc0cdbd68c57474d5533cdc49R264)
> 
> Cannot seek in detail right now, but apart from line-breaks this is the only thing I noticed. Great job!

Oops, fixed.

We really need to use partials, but this PR is mostly based on the template code we use on main, so we'll get it fixed at some point there.

XhmikosR:
New diff: https://github.com/twbs/bootstrap/commit/4661cd922ae3fc2b97014b4c8caacae728f9e735?w=1

XhmikosR:
Update the gh-pages comparison branch: https://github.com/twbs/bootstrap/commit/a0b8dd8d138ead44fd0350211265889b19dea1c7?w=1

ffoodd (approved):
LGTM

.gitignore
```diff
@@ -1,16 +1,8 @@
...
-/site/.jekyll-cache
-/site/.jekyll-metadata
-/site/docs/**/dist/
 # Hugo folders
 /resources/
 
-# Ignore ruby/bundler files
-/.bundle/
-/vendor/
-/.ruby-version
```
ffoodd, line 12: :heart:

ffoodd (approved):

XhmikosR:
Thanks @ffoodd for the review, it's been really helpful catching some of the issues :)

So, basically this leaves us:

1. See if we are going to go with the homepage changes /CC @mdo
2. I need to make a final rebase to clean up the patches
3. About the team members, we can tackle it in another PR; I just noted here whatever I noticed while comparing each of the pages

MartijnCuppens (approved):
Looks good!

Incredible job, once again, @XhmikosR!

PR #32080, opened 2020-11-08
Bundle requires two scripts

techboyg5:
The bundle requires two scripts (jQuery, Bootstrap), not one.


mdo:
This isn't the right approach, but I agree this can be clarified. Opened #32106 to do that for v4 and #32105 for v5.

PR #32090, opened 2020-11-09
[Docs] Changes to navbar documentation/explanation

### LINKED ISSUE #32041: [Docs] Example for components > navbar > external content is ... odd/contrived
https://v5.getbootstrap.com/docs/5.0/components/navbar/#external-content is a strangely convoluted/contrived example - i'm failing to come up with an actual good use case for the feature in the first place, but what's shown there in the example seems bizarre. can we come up with something closer to a real-life situation where this might be of use?

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/32041


mdo (approved):

XhmikosR:
@patrickhlauke  this needs to be backported manually

PR #32095, opened 2020-11-09
JS: use more "modern" APIs

XhmikosR:
So, I had this branch around since March 2020.

Currently, with our target browsers, I think it's safe to make these changes.

Preview: <https://deploy-preview-32095--twbs-bootstrap.netlify.app/>

- [x] Test Safari 10 - **EDIT**: I tested Safari 10.1 on BrowserStack and seems to work fine.


XhmikosR:
I also have https://github.com/twbs/bootstrap/commit/1777e8e03f0041d77cdd879c972aab3e8b050d39 which it might make sense to include it here. Not sure about Safari support, though.

Johann-S (approved):
LGTM 👍

Issue #32104, opened 2020-11-09
Docs Hugo minor issues


XhmikosR:
Looking good. couple of minor quick things I noticed:

- [ ] The code syntax highlighting is not quite as good imo:
https://getbootstrap.com/docs/4.5/getting-started/javascript/#sanitizer
https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/javascript/#sanitizer
Maybe just need to add blue color: `#006ee0` to class `.nx` ?

- [x] Looks like an unnecessary extra line break at bottom of the browser list here:
https://deploy-preview-32077--twbs-bootstrap.netlify.app/docs/4.5/getting-started/browsers-devices/#supported-browsers

_Originally posted by @coliff in https://github.com/twbs/bootstrap/pull/32077#issuecomment-724113090_

XhmikosR:
I'll try to fix these issues tomorrow; the extra line break one is lower prio, though, but I'd like to update/revisit the syntax highlighting styles.

XhmikosR:
Going to close this. If anyone wants to report any discrepancies, they are more than welcome :)

https://github.com/alecthomas/chroma

PR #32117, opened 2020-11-10
Update _syntax.scss

XhmikosR:
This is a draft PR which adds any missing chroma classes. Unfortunately, it seems it's not as simple as specifying the color for `.nx` since it breaks highlighting in other places like variable declarations.

Open for suggestions and we will probably need to report any issues upstream.

PS. if we find a solution, I'll sort the selectors to make things easier in the future to see the diff.

Refs #32104.

/CC @coliff for any suggestions


XhmikosR:
All right, I think we can merge this after I sort the selectors so that we have the latest ones. That being said, the differences we see are either bugs or lack of implementation on the chroma side.

If someone could report some of the issues upstream, that would help :)

Jp-watcher (approved):

mdo (approved):

PR #32120, opened 2020-11-10
Add note that `user-select-all` is not supported on IE or Edge Legacy

### LINKED ISSUE #32092: User-Select: all is not supported in IE or Edge Legacy
The Text Selection `.user-select-all` is not supported in IE or Edge Legacy.
REF: https://caniuse.com/mdn-css_properties_user-select_all

Maybe it could mention this as a note on on the v4 docs?
https://getbootstrap.com/docs/4.5/utilities/interactions/#text-selection

The compiled CSS actually includes:
```
.user-select-all {
  -ms-user-select: all !important;
```
but it doesn't do anything. Not really a big deal since it'd make almost no difference to the file-size after gzip compression - but  Can that be excluded from the CSS? Or would this be an issue with Autoprefixer?

coliff:
Fixes: #32092 

I kept the style/formatting of the compatibility issue **bold**, so it matches what is used elsewhere for IE issues, e.g.:
https://getbootstrap.com/docs/4.5/utilities/flex/#auto-margins
https://getbootstrap.com/docs/4.5/utilities/position/#sticky-top

I haven't been able to test this change locally by building the docs (don't have Jekyll 4.x installed right now)... but as this is simple change hopefully it should be ok.

Preview: https://deploy-preview-32120--twbs-bootstrap.netlify.app/docs/4.5/utilities/interactions/

### DIFF
--- a/site/docs/4.5/utilities/interactions.md
+++ b/site/docs/4.5/utilities/interactions.md
@@ -16,4 +16,6 @@ Change how the content is selected when the user interacts with it.
 {% endcapture %}
 {% include example.html content=example %}
 
+**Note: `user-select-all` is not supported on Internet Explorer or Edge Legacy.**
+
 Customize the available classes by changing the `$user-selects` Sass list in `_variables.scss`.


XhmikosR:
Thanks!

@mdo LGTY or do you think we should use a (warning) callout?

MartijnCuppens (approved):
I'm ok with the bold warning

XhmikosR:
@coliff we can't merge this it seems; you removed your fork at some point :/

site/docs/4.5/utilities/interactions.md
```diff
@@ -16,4 +16,6 @@ Change how the content is selected when the user interacts with it.
 {% endcapture %}
 {% include example.html content=example %}
 
+**Note: `user-select-all` is not supported on Internet Explorer or Edge Legacy.**
```
mdo, line 19: I'd move this up to be part of Line 10, and make this edit:

> Note that Internet Explorer and Legacy Edge have no support for the `all` value for `user-select`, and as such, **`.user-select-all` is not supported by either browser.**

XhmikosR: @mdo can you make a PR please? ❤

Issue #29818, opened 2019-12-07
Enable `recursivelyOptimizeProperties` option


MartijnCuppens:
Lines like these can be optimised in our generated CSS by enabling  [`recursivelyOptimizeProperties`](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations)
https://github.com/twbs/bootstrap/blob/1a5b7f4a8e235b0733b2ead9a4404ffaabb9ead2/dist/css/bootstrap.css#L5035

I tried, but I don't have a clue how to enable this here:
https://github.com/twbs/bootstrap/blob/1a5b7f4a8e235b0733b2ead9a4404ffaabb9ead2/package.json#L28

Probably worth backporting since this will decrease our file size.

@XhmikosR, could you help me out with this?

XhmikosR:
Generally I don't like enabling any of the level 2 optimizations because they can break stuff. And if we go down this route, we'll end up with us specifying a ton of level 2 optimizations...

That being said did you try ` -O1 recursivelyOptimizeProperties:true`?

XhmikosR:
And please share how much we save. I'm not sure how many cases we have that we could benefit, and again, I'm not sure if we should just go with the -O2 optimizations at all. Judging from this https://goalsmashers.github.io/css-minification-benchmark/ the gain isn't big enough to justify the risk of breaking things IMHO.

MartijnCuppens:
> Generally I don't like enabling any of the level 2 optimizations because they can break stuff

This one feels pretty safe. 

> That being said did you try `-O1 recursivelyOptimizeProperties:true`?

Yup, doesn't seem to work.

XhmikosR:
It might feel safe, but then why don't we enable further O2 optimizations? I'd rather keep it simple and safe than having to maintain a huge list of potential optimizations...

mendozagioo:
@MartijnCuppens why not just change variables configuration?
For example:
`$modal-header-padding: 1rem;`
instead of:
`$modal-header-padding: $modal-header-padding-y $modal-header-padding-x !default;`
The same for another situations

XhmikosR:
```diff
 package.json | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/package.json b/package.json
index 15170d720..1ac77f3fb 100644
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "css-lint": "npm-run-all --continue-on-error --parallel css-lint-*",
[...]

PR #32121, opened 2020-11-10
Use correct value order

MartijnCuppens:
Spotted in https://github.com/twbs/bootstrap/issues/29818#issuecomment-724185027, some values aren't in the correct order.


XhmikosR:
Just to confirm, none of this applies to v4?

mdo (approved):

XhmikosR:
I'm going to add this in the 4.6.0 and check if anything applies to v4 later.

PR #30622, opened 2020-04-20
Colors rewrite

MartijnCuppens:
Purpose of this PR is to use a uniform approach on using colors:

- Use `tint-color()` to replace `lighten()`
- Use `shade-color()` to replace `darken()`
- Use `scale-color()` to replace `color-level()` (since we don't work with levels anymore)
- Get rid of `$theme-color-interval`, the `8%` levels are hard to calculate with.
- Table variants, list item variants & alerts variants used to have different colors:
  - https://twbs-bootstrap.netlify.app/docs/5.0/content/tables/#variants
  - https://twbs-bootstrap.netlify.app/docs/5.0/components/alerts/#examples
  - https://twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#contextual-classes
    vs
  - https://deploy-preview-30622--twbs-bootstrap.netlify.app/docs/5.0/content/tables/#variants
  - https://deploy-preview-30622--twbs-bootstrap.netlify.app/docs/5.0/components/alerts/#examples
  - https://deploy-preview-30622--twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#contextual-classes
    (color contrast on the light variants is still an issue, but can be tackled later on to prevent this PR from being too overwhelming)

PR #32122, opened 2020-11-11
Prepare v5.0.0-alpha3

XhmikosR:
Preview: <https://deploy-preview-32122--twbs-bootstrap.netlify.app/>


coliff:
heya. Alpha-3 Changelog page feedback (though I realise this is a draft PR so you may already have some of this in hand).

- Suggest changing 'Colors' header to 'Sass' (to match Alpha2's notes) since the spinners animation bullet point isn't color-related. Can the first 4 bullets of this section be combined to one? As the 'See this PR for more details' isn't an item on its own. Also that link could be 'See #30622' for consistency.
- Maybe don't bold all of the first sentence of the 'Buttons' bullet-point? That's not done elsewhere and stands out a bit too much.
image.

Let me know if you want me to open a PR for any of those....

XhmikosR:
You can make a PR, but I forgot to lock the PR. :) The release PRs are what's going in and we don't use it for feedback.

mdo (approved):

PR #32124, opened 2020-11-11
tooltip.js: deduplicate `_fixTitle()`

XhmikosR:


js/src/tooltip.js
```diff
@@ -578,14 +578,11 @@ class Tooltip {
...
-    if (this.element.getAttribute('title') || titleType !== 'string') {
-      this.element.setAttribute(
-        'data-original-title',
-        this.element.getAttribute('title') || ''
-      )
+    const titleAttribute = this.element.getAttribute('title')
+    const titleIsString = this.element.getAttribute('data-original-title') === 'string'
 
+    if (titleAttribute || !titleIsString) {
+      this.element.setAttribute('data-original-title', titleAttribute || '')
```
XhmikosR, line 585: Looking at this, I think the extra check might not be needed after all?

Johann-S (approved):

PR #32125, opened 2020-11-11
Update codeql.yml

XhmikosR:
Specify the branches

I'm seeing a weird behavior, i.e. fixes when there's no change in a branch. It's like CodeQL is mixing up the branches. I'm not sure if this will fix the issue, but it's what they suggest in their workflow example

@jhutchings1 if you have any suggestions let me know :)


jhutchings1:
@XhmikosR Can you point me to a pull request where things got messed up? We've been chasing a couple of issues with our alert diffing functionality, and I'm wondering if what you're seeing is an instance of that. 

cc: @rneatherway @tibbes

XhmikosR:
@jhutchings1 will do as soon as I hit it again. It seems to happen if we push or make PRs sort of at the same time to different branches.

So, you think this PR shouldn't land? I mostly made it so that we follow the upstream action more closely and to see if that would fix the issue.

XhmikosR:
@jhutchings1 https://github.com/twbs/bootstrap/runs/1400254580 This is a v4-dev branch and the "fix" refers to the main branch.

jhutchings1:
Aha, yeah, I think you're hitting one of the race conditions in our diffing functionality. We're working on it, but haven't got it all fixed just yet. The changes you proposed in the PR are absolutely fine, but they won't fix the root cause, unfortunately. Hoping to have these issues fixed up soon!

XhmikosR:
Ah, perfect, I knew we couldn't be the only ones hitting this :)

I will simplify the workflow since we only need one language and I'll wait monitor for any fixes. Thanks for the reply!

rneatherway:
> @jhutchings1 https://github.com/twbs/bootstrap/runs/1400254580 This is a v4-dev branch and the "fix" refers to the main branch.

@XhmikosR :wave: I'm the engineer working on this issue. I had a look at this specific example and what you're seeing here is some older compatibility code we have for workflows that do analysis `on: pull_request`. This code looks for a pull request for an analysed commit and if one isn't found it speculatively does the comparison against the default branch. If you then opened a pull request against the default branch everything would appear correct. We are phasing this behaviour out. Your change to the workflow restricts the `on: push` triggers and should therefore prevent you seeing this in the future for that branch.

mdo (approved):

PR #32136, opened 2020-11-12
browsers-devices.md: make heading an `h3`

XhmikosR:
Noticed it in #32077 due to the ToC being indented.


XhmikosR:
Coveralls is getting confused because there was a branch with the same name in the past (which was against main).

MartijnCuppens (approved):

PR #32144, opened 2020-11-12
media-object.md: remove `a` tag

XhmikosR:
Not sure if this is intentional or not.


mdo:
It was, yes, as a way to represent linked avatars.

XhmikosR:
Ah, all good then; I  just noticed the difference in the snippet.

PR #32145, opened 2020-11-12
Avoid invisible real file input "spilling" out of container

### LINKED ISSUE #32133: Rare behaviour of the custom file input in some specific browsers
Hello to all,

First of all, thanks very, very much for Bootstrap CSS. I find a very rare behaviour with the custom file input. Basically I place a custom file input and next to this a "simple" text input (a CodePen is provided below).

The expected behaviour when we click on the File input is to see the File picker dialog, and, when we click on the text input, we expect that that control take the user focus and allows to write on it.

In browsers like Mozilla Firefox and Microsoft Edge all works as expected, that is, the controls are separated and have their expected behaviours when clicked, however, in browsers like Google Chrome and Opera we can see a rare behaviour.

It's like if the custom file input overlap the text input (even when the console developer inspector don't show that), so, if we click at the left of the text input, what we get is the File picker dialog of the custom file input opened, instead of get the focus in the text input.

[Here is a code pen in which you can see what I try to explain](https://codepen.io/decsoft/pen/pobqeZx)

If you try the above CodePen in Firefox or Microsoft Edge everything is working as expected. If you try it, on the other hand, in Google Chrome or Opera, you can see the rare behaviour which I try to describe here.

In few words, the custom file input File picker dialog is opened even if we click outside the control, for example, in another text input control placed next to the custom file input.

Here is a GIF image with the expected behaviour in Mozilla Firefox:

And here is a GIF image with the unexpected behaviour in Google Chrome:

Somethings to maybe take in consideration:

1. Maybe the issue is related with the "z-index" CSS rule of the ".custom-file-input"? If I remove that CSS rule, then the wrong behaviour is reduced (note, did not dissapear, but it's reduced), however, we can't then click over the custom file as expected.

2. Apparently we can have the same wrong behaviour if we place another kind of element next to the custom file input: the point is that in this case may we can't see the same that if the control placed next is a text input: because in this case we expect that the text input get the focus, and, this never occur, due to the launch of the custom file input File picker dialog.

I hope that someone can help me with this issue. Please, if you need more information, just ask me and I will try to provide it as quickly as possible. Thanks very, very much in advance for your help!

patrickhlauke:
Closes https://github.com/twbs/bootstrap/issues/32133


XhmikosR:
Just confirming, this doesn't affect v5, right?

patrickhlauke:
> Just confirming, this doesn't affect v5, right?

yup, correct. v4 only

MartijnCuppens (approved):

mdo (approved):

PR #32155, opened 2020-11-14
Add equal column mixin

cfxd:
~~Add equal column mixin so that we can do `@include make-col-eq` instead of having to extend the `.col` class or use it in markup.~~

Make the `make-col` mixin default to equal columns when not providing a column size.


mdo:
I think I can dig this. Put together a little Codepen: https://codepen.io/emdeoh/pen/JjRbXYr.

Since this only changes the mixin, I don't see any problems for us. Wondering if we need to update any docs?

cfxd:
Do you need me to make edits to that codepen? It looks like what I have in mind.

I can add a quick note for the docs with a note about the default behavior.

mdo:
No codepen edits needed. Just some docs updates I think to mention the behavior. Not sure where myself just yet, so feel free to suggest!

cfxd:
Ok, thank you! Added a docs edit and rebased 👍🏻

mdo (approved):

ffoodd (approved):
LGTM, pretty nice idea 👌

Issue #28342, opened 2019-02-22
Remaining to do after the jQuery removal


XhmikosR:
- [x] Update plugin pages
  - [x] Carousel - still has a few references to `$()` (#28768)
  - [x] Collapse - still has a few references to `$()` (#28768)
  - [x] Dropdown (#28768)
  - [x] List group
  - [x] Modal (#28768)
  - [x] Navs (#29181)
  - [x] Popover (#28768)
  - [x] ScrollSpy (#29181)
  - [x] Toasts (#28779)
  - [x] Tooltip (#28779)
- [x] Other doc pages
  - [x] getting-started/browsers-devices.md (#28779)
  - [x] getting-started/javascript.md (#28768)
- [ ] Various
  - [x] Add a section explaining that it's possible to use this version with and without jQuery
  - [x] Search docs for `jquery` (not case sensitive) and `$(`
  - [x] Replace https://codepen.io/Johann-S/pen/djJYPb or add a codepen with vanilla JS (#32163)
  - [ ] Maybe explain when an element is considered visible in the scrollspy callout
  - [x] See if we really need `DOMContentLoaded` anymore in tests and docs. #28337
  - [x] Remove `If you’re building our JavaScript from source, it requires util.js.` not needed anymore (#28779)
  - [x] Remove https://twbs-bootstrap.netlify.com/docs/4.3/getting-started/browsers-devices/#android-stock-browser after #28317 or in that branch (#28779)

midzer:
I am trying to help and port "Replace https://codepen.io/Johann-S/pen/djJYPb or add a codepen with vanilla JS" item

This is what i got until now:
https://paste.headstrong.de/view/08ba7d0b

Too bad I can't test properly, because `bootstrap` object is not found (I've included https://github.com/twbs/bootstrap/blob/master-xmr-v5/dist/js/bootstrap.bundle.js)

Johann-S:
Hi @midzer, thank you for trying to tackle that, but you should wait for our first alpha release

moliata:
> Hi @midzer, thank you for trying to tackle that, but you should wait for our first alpha release

Do you have an ETA when the first alpha is going to be released?

alpadev:
Since the alpha has been released - some updated codepen without the jQuery dependency: https://codepen.io/-alpa-/pen/PoZWzNM

[...]

PR #32156, opened 2020-11-15
Update selector example codepen

alpadev:
In reference to https://github.com/twbs/bootstrap/issues/28342

As requested by @Johann-S - Updated the codepen example in the popovers.md and tooltips.md

### DIFF
--- a/site/content/docs/5.0/components/popovers.md
+++ b/site/content/docs/5.0/components/popovers.md
@@ -214,7 +214,7 @@ Note that for security reasons the `sanitize`, `sanitizeFn`, and `allowList` opt
       <td><code>selector</code></td>
       <td>string | false</td>
       <td><code>false</code></td>
-      <td>If a selector is provided, popover objects will be delegated to the specified targets. In practice, this is used to enable dynamic HTML content to have popovers added. See <a href="{{< param repo >}}/issues/4215">this</a> and <a href="https://codepen.io/Johann-S/pen/djJYPb">an informative example</a>.</td>
+      <td>If a selector is provided, popover objects will be delegated to the specified targets. In practice, this is used to enable dynamic HTML content to have popovers added. See <a href="{{< param repo >}}/issues/4215">this</a> and <a href="https://codepen.io/-alpa-/pen/PoZWzNM">an informative example</a>.</td>
     </tr>
     <tr>
       <td><code>template</code></td>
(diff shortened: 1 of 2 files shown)


XhmikosR:
@mdo: we should fork that and use our Pen.

alpadev:
Well, that's what I actually proposed to do.. 🙄

XhmikosR <notifications@github.com> schrieb am So., 15. Nov. 2020, 06:22:

alpadev:
Maybe I should mention that this example isn't supported in old IE because I used `NodeList.forEach` directly, instead of `Array.forEach`. I didn't know if you got any preferred way to do this e.g. `[].forEach.call()` vs `Array.prototype.forEach.call()`.

Also I kept `var button = null` that could be simplified to `var button`.

XhmikosR:
RE: IE, that's OK, it's not supported on v5.

Feel free to simplify the pen and I'll fork it and update this PR after you are done. :)

alpadev:
@XhmikosR done. :)

XhmikosR:
@alpadev can you check https://codepen.io/team/bootstrap/pen/e1ebd72cfda6962c984dfd274e39658f please? I made a few more needed changes.

alpadev:
@XhmikosR looks much more polished now! Sorry for not putting more effort into it.

I agree it's better with some comments on those helpers.

About the code. The minor updates I had done was to rename `el` to `element` on the toggle function, though I personally prefer `el` too. (I had some linter to complain about it recently.)

And for optimization I guess that
```js
// Disable selector checkbox, put a tooltip on it, and show the buttons panel
    $("#use-selector").disabled = true;
    new bootstrap.Tooltip($("#use-selector-label span"));
    toggle($(".buttons"), true);
```
could be within the if-block because it needs to run only once imo. The jQuery example before had it outside the if-block (that way it's disabling the checkbox, initializing the tooltip and showing the buttons on each button press), maybe there is a reason to have it that way though.

Anyways, tested it and lgtm. :+1:

XhmikosR:
I actually got rid of the toggle function and used our utility classes, please check one more time.

https://codepen.io/team/bootstrap/pen/e1ebd72cfda6962c984dfd274e39658f?editors=1010

The CodePen doesn't need to be perfect, it should just be a good demo :)

alpadev:
Makes sense. You could use it like that too,
```js
// Demo-related
function updateCodeView() {
  $('#with-selector-code').classList.toggle('d-none', !usingSelectorOption());
  $('#without-selector-code').classList.toggle('d-none', usingSelectorOption());
}
```
but that's up to you. 👍

PR #32165, opened 2020-11-16
Fix focus for dropdown items

spahno:
### What we have tried:

1. The problem is, that if you set the ``-webkit-focus-ring-color auto`` attribute, the outline style and outline color works differently in every browser for the ``<a>`` and the ``<button>`` tag. For example, the ``<a>`` is dotted in Firefox unlike the ``<button>``which is surrounded with a solid blue outline.

2. We tried to recreate the blue outline with a border and border-radius for every tag but we could not make the anchor tag have a rounded border, because a link cannot have a border-radius.

### Solution:
To find a consistent solution for every browser, we decided to add a blue outline without a border-radius as you can see below in the screenshots.

### Tested browsers:
Chrome 86.0.4240.193, Firefox 82.0.3 and Safari 14.0.

### Screenshots:
The focus, when navigating with arrow keys:
Focus with arrow keys

The focus, when clicking:
Focus on click

### Team:
@LtHustensaft
@lewe98
@spahno


MartijnCuppens (changes requested):

scss/_reboot.scss
```diff
@@ -268,6 +268,11 @@ a {
   }
 }
 
+// dropdown-item focus
+a.dropdown-item:focus {
```
ffoodd, line 272: This shouldn't be in `_reboot.scss` since it's using a component selector.

ffoodd (changes requested):

ffoodd:
I think this is not the way to proceed here: either we normalize an opinionated focus style for everything, or we simply style the focus state for dropdown items.

But this PR mixes approaches and will more likely introduce unexpected behaviours elsewhere.

patrickhlauke:
...and this is one of the cases where `:focus-visible` + polyfill could help square the circle as well without being in-your-face for mouse/touch users...

mdo:
Agreed on not the right approach here.

scss/_reboot.scss
spahno: Could you please explain to me why ``_reboot.scss`` is the wrong file? I'm a new bootstrap contributor, so I'm asking. 😅

spahno:
> I think this is not the way to proceed here: either we normalize an opinionated focus style for everything, or we simply style the focus state for dropdown items.
> 
> But this PR mixes approaches and will more likely introduce unexpected behaviours elsewhere.

Hey @ffoodd! Thanks for your reply. What would you recommend to do? Would you agree to @patrickhlauke's idea, that a polyfill would help in this case?

spahno:
> ...and this is one of the cases where `:focus-visible` + polyfill could help square the circle as well without being in-your-face for mouse/touch users...

Hey @patrickhlauke! So you're saying ``:focus-visible``would solve this issue properly? Since this pseudo-class isn't working properly for all browsers, I was looking for a polyfill and found this:

https://github.com/WICG/focus-visible

Do you think this could help?

ffoodd:
@spahno we already made an attempt to use the `:focus-visible` polyfill — but it's a complexe topic since it means adding a dependency.

spahno:
> @spahno we already made an attempt to use the `:focus-visible` polyfill — but it's a complexe topic since it means adding a dependency.

@ffoodd alright 😅 so you're basically saying it's not worth the effort? 😬🥴

ffoodd:
Sorry?

No. I said:
> it's a complex topic since it means adding a dependency

I was the one who made the effort, and am always pushing alongside @patrickhlauke to make it happen.

Please read carefully before replying, and avoid  making assumptions about our intentions or make us say what we didn't say.

That's not really helpful, is it?

spahno:
> Sorry?
> 
> No. I said:
> 
> > it's a complex topic since it means adding a dependency
> 
> I was the one who made the effort, and am always pushing alongside @patrickhlauke to make it happen.
> 
> Please read carefully before replying, and avoid making assumptions about our intentions or make us say what we didn't say.
> 
> That's not really helpful, is it?

Sorry! I didn't mean to be rude. I think we misunderstood each other 😅 I hope that you're able to resolve this issue in the future :)

PR #32173, opened 2020-11-17
Make the attribute methods bs specific

rohit2sharma95:
Ref: https://github.com/twbs/bootstrap/pull/31827#issuecomment-728864694


js/tests/unit/dom/manipulator.spec.js
```diff
@@ -73,39 +75,40 @@ describe('Manipulator', () => {
   })
 
   describe('getDataAttribute', () => {
```
Johann-S, line 77: can you add a test here with a non bs attribute please ? To check it doesn't get non bs attribute

rohit2sharma95: Added @Johann-S 👍

Johann-S (approved):
LGTM 👍 
Thanks @rohit2sharma95

XhmikosR:
Perfect, thank you both!

PR #32174, opened 2020-11-17
Position Utilities Docs: Fix the marker SVG color to match the button

coliff:
The bottom marker icon is a different color to the button color.

image

This looked like a mistake, so this PR updates the color to match the `btn-dark` color - but if it was intentional to show how this is positioned of course feel free to decline the PR. :-)


ffoodd:
It's a missed case of bumping `$dark` from `$gray-800` to `$gray-900` in #30550.

Thanks for spotting this!

ffoodd (approved):

PR #32176, opened 2020-11-17
docs: use `expanded` Sass output style for development

XhmikosR:


ffoodd (approved):
Could we use `enableSourceMap` too, only for dev? Or drop `noMap` for PostCSS?

Not sure how they would work together though, but sourcemaps are useful locally :)

XhmikosR:
> Could we use `enableSourceMap` too, only for dev? Or drop `noMap` for PostCSS?
> 
> Not sure how they would work together though, but sourcemaps are useful locally :)

I had issues with postcss so I ended up disabling sourcemaps. But feel free to experiment with and make a PR if you find a solution that works both for dev and prod :)

PR #32177, opened 2020-11-17
examples: fix `.btn-block` instances

XhmikosR:
Regression of #31995

### DIFF
--- a/site/content/docs/5.0/examples/checkout/index.html
+++ b/site/content/docs/5.0/examples/checkout/index.html
@@ -215,7 +215,7 @@ body_class: "bg-light"
 
           <hr class="my-4">
 
-          <button class="btn btn-primary btn-lg btn-block" type="submit">Continue to checkout</button>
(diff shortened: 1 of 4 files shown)


ffoodd (approved):
This fixes missed cases, however re-reading the initial PR, using `.w-100` as a simple replacement for `.btn-block` isn't mentionned in our docs nor in our migration guide.

Shouldn't it be?

XhmikosR:
I guess it could, not sure why @mdo didn't mention it, but he might have just missed it :)

XhmikosR:
BTW not sure if `w-100` should be added at the end or not. Let me know if you have a preference.

mdo (approved):
Hadn't mentioned `.w-100` because most of the situations I was thinking about were for multiple stacked buttons. Feel free to add something to the Migration docs for our last release.

XhmikosR:
I'm unsure how to reword the current text, so please have a look @mdo in another PR or push it here :)

mdo (approved):

PR #32179, opened 2020-11-17
Don't hide modal when `config.keyboard` is false

### LINKED ISSUE #30929: Bootstrap 4.5.0 modal "keyboard: false" don't work
- Operating system: Ubuntu 20.04
- Browsers: Chrome 83.0.4103.61, Firefox 76.0.1
- Bootstrap version: 4.5.0

https://jsfiddle.net/4sujahx5/

rohit2sharma95:
Closes #30929

Modal will not be closed with <kbd>esc</kbd> key when modal's configuration has `keyboard:false` and backdrop is not static. It will trigger the `_triggerBackdropTransition()` method to show the animation though. But I am not sure if modal should show animation when pressing <kbd>esc</kbd> key and modal's backdrop is not static. 🤔 

Should I add/update the unit tests if it is okay? 🙂


Johann-S:
I don't know if I totally understood what you said, but the behavior should be: if a modal have `keyboard` at `false` in that case `esc` shouldn't work that's all

rohit2sharma95:
> but the behavior should be: if a modal have `keyboard` at `false` in that case `esc` shouldn't work

Yes, <kbd>esc</kbd> is not working.
But modal is being animated even if its backdrop is not static. 🙂

rohit2sharma95:
Should I add unit tests then, if it is fine 🙂

Johann-S:
yep please add a unit test 👌

rohit2sharma95:
@Johann-S Seems that updating the previous test would be okay here 🤔

Johann-S:
LGTM 👍 

/CC @XhmikosR

XhmikosR:
@Johann-S you forgot to approve?

js/tests/unit/modal.spec.js
```diff
@@ -595,12 +595,11 @@ describe('Modal', () => {
       modal.show()
     })
 
-    it('should not close modal when escape key is pressed with keyboard = false and backdrop = static', done => {
-      fixtureEl.innerHTML = '<div class="modal" data-bs-backdrop="static" data-bs-keyboard="false"><div class="modal-dialog"></div>'
+    it('should not close modal when escape key is pressed with keyboard = false', done => {
+      fixtureEl.innerHTML = '<div class="modal"><div class="modal-dialog"></div>'
```
XhmikosR, line 599: Aren't we missing a `</div>` here BTW? It doesn't affect anything, and we don't enforce it, though.

rohit2sharma95: Updated @XhmikosR 👍

Johann-S:
I asked if you it was good to you too @XhmikosR 😉

Johann-S (approved):

XhmikosR:
Thank you both!

@rohit2sharma95 can you make a PR against v4-dev please?

rohit2sharma95:
Okay sure @XhmikosR 👍 
Thanks 🙂

PR #32180, opened 2020-11-17
refactor: use a Map instead of an Object in dom/data

### LINKED ISSUE #32123: Refactor data.js to use a `Map`
It seems [dom/data.js](https://github.com/twbs/bootstrap/blob/main/js/src/dom/data.js) is the perfect candidate for using a `Map` instead of the custom implementation.

alpadev:
Fixes #32123

Preview: <https://deploy-preview-32180--twbs-bootstrap.netlify.app/>

### DIFF
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -98,7 +98,7 @@ class Alert extends BaseComponent {
 
   static jQueryInterface(config) {
     return this.each(function () {
-      let data = Data.getData(this, DATA_KEY)
+      let data = Data.get(this, DATA_KEY)
 
       if (!data) {
         data = new Alert(this)
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -25,7 +25,7 @@ export default {
     // can be removed later when multiple key/instances are fine to be used
     if (!instanceMap.has(key) && instanceMap.size !== 0) {
       // eslint-disable-next-line no-console
-      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${[...instanceMap.keys()].join(', ')}.`)
+      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${Array.from(instanceMap.keys()).join(', ')}.`)
     }
 
     instanceMap.set(key, instance)
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -11,57 +11,44 @@
  * ------------------------------------------------------------------------
  */
 
-const mapData = (() => {
-  const storeData = {}
-  let id = 1
-  return {
-    set(element, key, data) {
-      if (typeof element.bsKey === 'undefined') {
-        element.bsKey = {
-          key,
-          id
-        }
-        id++
-      }
+const elementMap = new Map()
 
-      storeData[element.bsKey.id] = data
-    },
-    get(element, key) {
-      if (!element || typeof element.bsKey === 'undefined') {
-        return null
-      }
+export default {
+  set(element, key, instance) {
+    if (!elementMap.has(element)) {
+      elementMap.set(element, new Map())
+    }
 
-      const keyProperties = element.bsKey
-      if (keyProperties.key === key) {
-        return storeData[keyProperties.id]
-      }
+    const instanceMap = elementMap.get(element)
 
-      return null
-    },
-    delete(element, key) {
-      if (typeof element.bsKey === 'undefined') {
-        return
-      }
+    // make it clear we only want one instance per element
+    // can be removed later when multiple key/instances are fine to be used
+    if (!instanceMap.has(key) && instanceMap.size !== 0) {
+      // eslint-disable-next-line no-console
+      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${[...instanceMap.keys()].join(', ')}.`)
+    }
 
-      const keyProperties = element.bsKey
-      if (keyProperties.key === key) {
-        delete storeData[keyProperties.id]
-        delete element.bsKey
+    instanceMap.set(key, instance)
+  },
+
+  get(element, key) {
+    if (elementMap.has(element)) {
+      return elementMap.get(element).get(key) || null
+    }
+
+    return null
+  },
+
+  remove(element, key) {
+    if (elementMap.has(element)) {
+      const instanceMap = elementMap.get(element)
+
+      instanceMap.delete(key)
+
+      // free up element references if there are no instances left for an element
+      if (instanceMap.size === 0) {
+        elementMap.delete(element)
       }
     }
   }
-})()
-
-const Data = {
-  setData(instance, key, data) {
-    mapData.set(instance, key, data)
-  },
-  getData(instance, key) {
-    return mapData.get(instance, key)
-  },
-  removeData(instance, key) {
-    mapData.delete(instance, key)
-  }
 }
-
-export default Data
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -25,7 +25,7 @@ export default {
     // can be removed later when multiple key/instances are fine to be used
     if (!instanceMap.has(key) && instanceMap.size !== 0) {
       // eslint-disable-next-line no-console
-      console.error(`Notice: The design of Bootstrap disallows to use more than one instance per element. Bound instance: ${Array.from(instanceMap.keys())[0]}.`)
+      console.error(`Bootstrap doesn't allow more than one instance per element. Bound instance: ${Array.from(instanceMap.keys())[0]}.`)
       return
     }
 
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -40,15 +40,17 @@ export default {
   },
 
   remove(element, key) {
-    if (elementMap.has(element)) {
-      const instanceMap = elementMap.get(element)
+    if (!elementMap.has(element)) {
+      return
+    }
 
-      instanceMap.delete(key)
+    const instanceMap = elementMap.get(element)
 
-      // free up element references if there are no instances left for an element
-      if (instanceMap.size === 0) {
-        elementMap.delete(element)
-      }
+    instanceMap.delete(key)
+
+    // free up element references if there are no instances left for an element
+    if (instanceMap.size === 0) {
+      elementMap.delete(element)
     }
   }
 }
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -26,6 +26,7 @@ export default {
     if (!instanceMap.has(key) && instanceMap.size !== 0) {
       // eslint-disable-next-line no-console
       console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${Array.from(instanceMap.keys()).join(', ')}.`)
+      return
     }
 
     instanceMap.set(key, instance)
--- a/js/src/dom/data.js
+++ b/js/src/dom/data.js
@@ -25,7 +25,7 @@ export default {
     // can be removed later when multiple key/instances are fine to be used
     if (!instanceMap.has(key) && instanceMap.size !== 0) {
       // eslint-disable-next-line no-console
-      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${Array.from(instanceMap.keys()).join(', ')}.`)
+      console.error(`Notice: The design of Bootstrap disallows to use more than one instance per element. Bound instance: ${Array.from(instanceMap.keys())[0]}.`)
       return
     }
 
--- a/js/tests/unit/dom/data.spec.js
+++ b/js/tests/unit/dom/data.spec.js
@@ -4,128 +4,103 @@ import Data from '../../../src/dom/data'
 import { getFixture, clearFixture } from '../../helpers/fixture'
 
 describe('Data', () => {
+  const TEST_KEY = 'bs.test'
+  const UNKNOWN_KEY = 'bs.unknown'
+  const TEST_DATA = {
+    test: 'bsData'
+  }
+
   let fixtureEl
+  let div
 
   beforeAll(() => {
     fixtureEl = getFixture()
   })
 
+  beforeEach(() => {
+    fixtureEl.innerHTML = '<div></div>'
+    div = fixtureEl.querySelector('div')
+  })
+
   afterEach(() => {
+    Data.remove(div, TEST_KEY)
     clearFixture()
   })
 
-  describe('setData', () => {
-    it('should set data in an element by adding a bsKey attribute', () => {
-      fixtureEl.innerHTML = '<div></div>'
+  it('should return null for unknown elements', () => {
+    const data = { ...TEST_DATA }
 
-      const div = fixtureEl.querySelector('div')
-      const data = {
-        test: 'bsData'
-      }
+    Data.set(div, TEST_KEY, data)
 
-      Data.setData(div, 'test', data)
-      expect(div.bsKey).toBeDefined()
-    })
-
-    it('should change data if something is already stored', () => {
-      fixtureEl.innerHTML = '<div></div>'
-
-      const div = fixtureEl.querySelector('div')
-      const data = {
-        test: 'bsData'
-      }
-
-      Data.setData(div, 'test', data)
-
-      data.test = 'bsData2'
-      Data.setData(div, 'test', data)
-
-      expect(div.bsKey).toBeDefined()
-    })
+    expect(Data.get(null)).toBeNull()
+    expect(Data.get(undefined)).toBeNull()
+    expect(Data.get(document.createElement('div'), TEST_KEY)).toBeNull()
   })
 
-  describe('getData', () => {
-    it('should return stored data', () => {
-      fixtureEl.innerHTML = '<div></div>'
+  it('should return null for unknown keys', () => {
+    const data = { ...TEST_DATA }
 
-      const div = fixtureEl.querySelector('div')
-      const data = {
-        test: 'bsData'
-      }
+    Data.set(div, TEST_KEY, data)
 
-      Data.setData(div, 'test', data)
-      expect(Data.getData(div, 'test')).toEqual(data)
-    })
-
-    it('should return null on undefined element', () => {
-      expect(Data.getData(null)).toEqual(null)
-      expect(Data.getData(undefined)).toEqual(null)
-    })
-
-    it('should return null when an element have nothing stored', () => {
-      fixtureEl.innerHTML = '<div></div>'
-
-      const div = fixtureEl.querySelector('div')
-
-      expect(Data.getData(div, 'test')).toEqual(null)
-    })
-
-    it('should return null when an element have nothing stored with the provided key', () => {
-      fixtureEl.innerHTML = '<div></div>'
-
-      const div = fixtureEl.querySelector('div')
-      const data = {
-        test: 'bsData'
-      }
-
-      Data.setData(div, 'test', data)
-
-      expect(Data.getData(div, 'test2')).toEqual(null)
-    })
+    expect(Data.get(div, null)).toBeNull()
+    expect(Data.get(div, undefined)).toBeNull()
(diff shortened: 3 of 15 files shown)


alpadev:
@XhmikosR I realised something in the dom/data spec https://github.com/twbs/bootstrap/blob/main/js/tests/unit/dom/data.spec.js#L30-L44 

The test says 'should change data if something is already stored' but it expects the `bsKey` to be defined. Also if `data.test` is updated, there would be no need to call `setData` again as the object from the first call remains the same.

js/src/dom/data.js
```diff
@@ -24,7 +24,7 @@ const mapData = (() => {
         id++
       }
 
-      storeData[element.bsKey.id] = data
+      storeData.set(element.bsKey.id, data)
```
rohit2sharma95, line 27: The `element` can not be used directly here? 🤔 
Ref: https://github.com/twbs/bootstrap/issues/32123#issuecomment-730180864

Johann-S: yep the element should be used here to the data

alpadev: Uhm that's a bit contradictory to this response https://github.com/twbs/bootstrap/issues/32123#issuecomment-728855901 but well

Johann-S: haha yes sorry @alpadev but things can change 🤷

XhmikosR:
FYI this results in a nice size decrease 🙂 https://ja2r7.app.goo.gl/5pUhLMyZD1iuZgYe9

I move this to beta1 @Johann-S. Tests do seem to pass, although I notice a few of them are removed.

rohit2sharma95:
Yes, we discussed the tests here: https://github.com/alpadev/bootstrap/pull/1, though
Few of them are removed, I am not sure which tests should be added again 🙂

js/src/alert.js
```diff
@@ -49,7 +49,7 @@ class Alert {
     this._element = element
 
     if (this._element) {
-      Data.setData(element, DATA_KEY, this)
```
XhmikosR, line 52: BTW the `DATA_KEY` constants are only used in one place in each of the plugins except for popover.js and tooltip.js. The minifier already inlines those, so I guess we better keep them for consistency.

js/src/dom/data.js
```diff
@@ -11,57 +11,16 @@
...
-      }
-    }
-  }
-})()
-
-const Data = {
-  setData(instance, key, data) {
-    mapData.set(instance, key, data)
+export default {
+  setData(element, data) {
```
XhmikosR, line 17: Maybe we should rename the objects to `set`, `get` and `remove` @Johann-S? I mean, we use them as `Data.foo` so having `Data/fooData` seems a little redundant. Just thinking out loud here :)

Johann-S: yep I agree too 👍

Johann-S:
LGTM 👍 since we don't allow different instance on an element 👌

Johann-S (dismissed):

XhmikosR:
@Johann-S My only concerns are:

1. if the removed tests should all be removed and if we still cover all cases we did before, like passing non`-bs` prefixed data attributes etc
2. I suppose if we want to support multiple keys in the future it shouldn't be a lot of work, right?

All in all, I love this change; it makes the code simpler and also saves us a good amount of code. I just want to make sure we aren't missing anything and that we won't be circling around in the future 🙂

alpadev:
If it's expected to have multiple keys/instances on one element in a closer future, this might be not the ideal implementation to go with, I guess.

This basically removes the circuit of mapping a `UID` (and an `DATA_KEY`), that both been stored in an object at the elements' `bsKey`, to an instance that is saved in the `mapData`, and instead uses the 1:1 relationship of element and instance directly with the element being the key to the `mapData`s' instance.

So before it was like:
A `PluginInstance <> mapData.id <> element.bsKey.id` mapping, with the `DATA_KEY` only being some kind of validation to check if `Data.getData` or `Data.removeData` is called with the same `DATA_KEY` that has been stored in `element.bsKey` when `Data.setData` was called (the first time).

Now it's like:
A `PluginInstance <> mapData.element <> element` mapping, with the `DATA_KEY` not being used here at all.

Another thing to mention, is that `Plugin.getInstance(element)` isn't "type-safe" anymore. Before if I had a `Tooltip` bound to an element, `Tooltip.getInstance(element)` would return some Tooltip-Instance, while `Button.getInstance(element)` returned `null`. Now every Plugin.getInstance returns whatever is saved in `mapData` for that element or `null`.

XhmikosR:
Sounds like we need to rethink this. Let's keep this on hold for a few days
until we get more feedback

alpadev:
If I was to propose some implementation for the data, it should look something like this:

```js
const elementMap = new Map();

export default {
  set(element, key, instance) {
    if (!elementMap.has(element)) {
      elementMap.set(element, new Map());
    }

    const instanceMap = elementMap.get(element);

    // make it clear we only want one instance per element
    // can be replaced at a later time when multiple key/instances are allowed
    if (!instanceMap.has(key) && instanceMap.size !== 0) {
      throw new Error(
        "Sorry, we currently allow only one instance per element."
      );
    }

    instanceMap.set(key, instance);
  },

  get(element, key) {
    if (elementMap.has(element)) {
      return elementMap.get(element).get(key) || null;
    }

    return null;
  },

  remove(element, key) {
    if (elementMap.has(element)) {
      const instanceMap = elementMap.get(element);

      instanceMap.delete(key);

      // free up references if there are no instances left for an element
      if (!instanceMap.size) {
        elementMap.delete(element);
      }
    }
  }
};
```

This uses both an `element` and a `key` to be used to store some data/instance. It'd also allow for multiple `key<>interface` mappings.

Jbo666:
Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/ghei36>

________________________________

XhmikosR:
@alpadev yup, that sounds like a better approach since that seems closer to what we have right now :)

Johann-S:
@alpadev said :
> Another thing to mention, is that Plugin.getInstance(element) isn't "type-safe" anymore. Before if I had a Tooltip bound to an element, Tooltip.getInstance(element) would return some Tooltip-Instance, while Button.getInstance(element) returned null. Now every Plugin.getInstance returns whatever is saved in mapData for that element or null if nothing saved.

The old behavior should be kept here, it's important to be type safe.

And I like your other implementation (https://github.com/twbs/bootstrap/pull/32180#issuecomment-731640223)

BTW when I created `Data` object, I tried to be as close as possible to how jQuery store data when we call `.data` on a jQuery object

XhmikosR:
So, assuming we follow @alpadev's [suggestion(https://github.com/twbs/bootstrap/pull/32180#issuecomment-731640223), would this still be considered a breaking change? If so, we need to get this done in beta1, otherwise we can move it to beta2.

rohit2sharma95:
It is not a breaking change then IMO 🤔 
Because as per [the comment](https://github.com/twbs/bootstrap/pull/32180#issuecomment-731640223) of @alpadev the old behaviour would be kept here.

Johann-S:
Yep I agree with @rohit2sharma95 👍

alpadev:
Alright, then. I think, we could've reimplemented the type-safety by adding instance checks, if it was only for that reason tho. 

Then I'm reseting and updating my branch with the latest proposal.

That also means we won't get that nice improvement in bundle size (as most of it results from removing the `KEY_DATA`), I'm sorry @XhmikosR :). Maybe it makes sense to consider some kind of generic plugin base class that gets extended by each plugin. That maybe could allow for some generalization and thus some savings in that regard.

@Johann-S we usually develop solutions to a certain requirement or problem. Like you said, things may change and so eventually the implementation does. 👍

XhmikosR:
> Maybe it makes sense to consider some kind of generic plugin base class that gets extended by each plugin. That maybe could allow for some generalization and thus some savings in that regard.

@alpadev yeah, it's already in the TODO list, see #29370, although the size increases, it's not a big increase.

alpadev:
I guess, for the spec it would be better to export the map and validate on it directly but my experience in writing tests is rather basic.

js/src/dom/data.js
```diff
@@ -11,57 +11,45 @@
...
-  return {
-    set(element, key, data) {
-      if (typeof element.bsKey === 'undefined') {
-        element.bsKey = {
-          key,
-          id
-        }
-        id++
-      }
+const elementMap = new WeakMap()
```
alpadev, line 14: Used a WeakMap here. The advantage is, it holds weak references on the elements and thus shouldn't prevent them from getting garbage collected. Not sure if there are any concerns about this.

js/src/dom/data.js
```diff
@@ -11,57 +11,45 @@
...
-      if (keyProperties.key === key) {
-        delete storeData[keyProperties.id]
-        delete element.bsKey
-      }
+    // make it clear we only want one instance per element
+    // can be removed at a later time when multiple key/instances are allowed
+    if (!instanceMap.has(key) && instanceMap.size !== 0) {
+      throw new Error(
+        'Sorry, we currently allow only one instance per element.'
+      )
```
rohit2sharma95, line 29: It would be better if you use bootstrap here instead of we

js/src/dom/data.js
XhmikosR: Not super familiar with this either. I'd go with a Map personally, but maybe wait for someone else to chime in

js/src/dom/data.js
XhmikosR: Not sure if we should throw an error. As much as I like this approach, we generally try to not break people's code.

js/src/dom/data.js
rohit2sharma95: Agree with @alpadev here

> But `WeakMap` automatically also removes the key along with the information kept by that key when the key (as object) becomes unreachable.

So probably in the future when supporting multiple plugin instances it should be revisited because `dispose` method in the plugins sets the element to null. What do you think? 🙂

js/src/dom/data.js
alpadev: @XhmikosR in general I'm with you on this, but I'm not sure what's the best way to deal with this.
Throwing an error has the advantage of making tests fail in case someone was to implement/refactor a plugin while using multiple instances on an element without being aware of it.
From another perspective the old implementation had kind of a bug, as it wouldn't update the `key` properly when used with a different `key` on any additional call to `setData`. That likely to cause some implications down the line, even when used in production code (maybe without even noticing). So an error could be helpful to detect there is something wrong here.
As an alternative to an error, what do you think would be a good way? We could use console.error and return without altering the map, or do we want to show just the information and continue? We could also mimic the old behaviour, but that means we reintroduce that `key` flaw again.

js/src/dom/data.js
alpadev: @rohit2sharma95 tbh I'm not completely aware how WeakMaps internally work, but to my understanding, using an object (or element) as a key here doesn't alter the general behaviour of the garbage collection. It rather uses a weak reference to an object that can be freed automatically when the object has no other reference, instead of a strong reference that would prevent an object from being collected until that reference is explicitely removed.

In that case you mention here, when there are multiple plugin instances each referencing the same element. Setting the element to null in one instance (e.g. `this._element = null`) shouldn't have an impact on any other reference to the same element or the element itself. It's not the deleting the element but rather overwriting that single reference to it with null. 

As such the key-element of the WeakMap should stay intact, as this isn't altered by the code.

js/src/dom/data.js
XhmikosR: Just so that we are on the same boat, I don't believe that no feedback is a good approach. That being said, if we are going to throw an error, this change needs to be in beta1 which is soon to be released, after documenting it in migration notes.

Otherwise, we can't land it later, only in a major version bump.

js/src/dom/data.js
XhmikosR: I'm still hesitant to using a WeakMap here. I'd rather err on the safe side and go with a Map personally.

alpadev:
Just pushed changes in regard to the received feedback:

It now uses a `Map` instead of a `WeakMap`. Leaking should be prevented by checking for `instanceMap.size === 0` and removing the element.

Instead of throwing an error, it now uses `console.error` with the following message `Notice: The design of Bootstrap currently discourage to use more than one instance per element. Bound instance(s): ${instanceMap.keys().join(', ')}.`, and continues to set additional keys/instances. Or do you rather want to return from there?

rohit2sharma95:
For now, Bootstrap is not supporting multiple instances on a single element though but IMO it is a good addition @XhmikosR

And this feature should not break anything :thinking: @alpadev

alpadev:
@rohit2sharma95 Ideally it shouldn't break something. The signature remains unchanged. If someone uses it with multiple instances on one element, he would get a `console.error` with the information about using it inappropriate, while fixing the problem with the key.

Like I mentioned before - Old behaviour: if someone was to use multiple keys on one element, it would only use the first key and overwrite that on additional calls. Now: it uses multiple keys.

It's possible to mimic the old behaviour exactly like it has been, tho. (But imo I'd consider the old behaviour a bit flawed).

rohit2sharma95:
Agree with your comment: https://github.com/twbs/bootstrap/issues/32123#issuecomment-729107917
And thanks for the explanatory demo :smile:

XhmikosR:
So, I don't think we should console.error. We should stick to the current behavior, unless we can't find another way. Otherwise, this could be considered a breaking change AFAICT.

alpadev:
@XhmikosR Not sure if I agree on this.

I mean officially it's not a supported feature to use multiple instances on one element according to https://github.com/twbs/bootstrap/issues/32123#issuecomment-731020351. If someone is currently using it, I wouldn't consider it even a bad practise to throw an error, since it's against this intention. Neither do I consider it a breaking change - it's not breaking something that is *supported* behaviour.
I agree tho, that since it's not nice to break user code, it might be a better approach to not make things fail, but rather to give the implementor some hint that he might use it in an inappropriate way. That's whats the console.error is doing.

I'd consider this an edge case anyway, since the documentation doesn't mention it. However, using multiple instances can cause things to break (intransparently) with the current behaviour like in this demo https://codepen.io/-alpa-/pen/KKMExwM. That's why I kinda dislike to keep it the way it is.

Enabling multiple keys allows Data.set/Data.get, to work properly with multiple instances (That's what's causing that many tooltips to spawn in the demo). The behaviour with one instance on one element isn't changed here.

js/src/dom/data.js
```diff
@@ -11,57 +11,44 @@
...
+      return elementMap.get(element).get(key) || null
+    }
+
+    return null
   },
-  removeData(instance, key) {
-    mapData.delete(instance, key)
+
+  remove(element, key) {
+    if (elementMap.has(element)) {
```
Johann-S, line 43: I'm a big fan of return as early as possible so I would prefer :
```js
if (!elementMap.has(element)) {
  return;
}
```

Johann-S:
I'm not a big fan on console.logs in build files, but in this case I think it make sense to let people know why it doesn't work.

Personally I'm still against allowing people to have multiple plugin instances I think we'll have weird behavior and I don't see any valid use cases currently, that's all for my two cents 😄

Johann-S (approved):
LGTM 👍 and keep the old behavior in a more up to date way 👌 
thanks @alpadev

js/src/dom/data.js
```diff
@@ -11,57 +11,44 @@
...
-      const keyProperties = element.bsKey
-      if (keyProperties.key === key) {
-        delete storeData[keyProperties.id]
-        delete element.bsKey
-      }
+    // make it clear we only want one instance per element
+    // can be removed later when multiple key/instances are fine to be used
+    if (!instanceMap.has(key) && instanceMap.size !== 0) {
+      // eslint-disable-next-line no-console
+      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${[...instanceMap.keys()].join(', ')}.`)
```
rohit2sharma95, line 28: I get `[object Map Iterator]` instead the keys' names in the console :thinking: 
image

js/tests/unit/dom/data.spec.js
```diff
@@ -4,128 +4,103 @@ import Data from '../../../src/dom/data'
...
+    const copy = { ...data }
 
-      Data.removeData(div, 'test')
+    Data.set(div, TEST_KEY, data)
+    Data.set(div, UNKNOWN_KEY, copy)
 
-      expect(div.bsKey).toBeUndefined()
-    })
+    expect(console.error).toHaveBeenCalled()
+    expect(Data.get(div, UNKNOWN_KEY)).toBe(copy)
```
rohit2sharma95, line 104: This expectation makes sure that `UNKNOW_KEY` key is now set on the map. So there are multiple instances registered with the same DOM element. (With `TEST_KEY` and `UNKNOW_KEY`)

rohit2sharma95 (commented):
And this PR now supports multiple plugin instances on a single DOM element (But also closes the issue that was before: https://github.com/twbs/bootstrap/issues/32123#issuecomment-729107917).

There should be a `return` statement after `console.error` in the `set` method :thinking: (If it is supposed not to support multiple instances)

XhmikosR (changes requested):
There are some last comments to address

alpadev:
> I'm not a big fan on console.logs in build files, but in this case I think it make sense to let people know why it doesn't work.
> 
> Personally I'm still against allowing people to have multiple plugin instances I think we'll have weird behavior and I don't see any valid use cases currently, that's all for my two cents 😄

I didn't mean to enforce my opinion here 🙂. My intention was just to explain my standpoint and to receive some input.

As I see it we have those options:

Allow the use of multiple components per element
- Just let it happen and have the implementers figure out problems that may occure 
- Allow it but show a message (console.error/warn/log) to give them a hint **<-- this is how it is for now**

Disallow the use of multiple components
- Throw an error describing the problem
- Overwrite the first component but disposing it before (with or without a message)
- Overwrite the first component without cleaning up (with or without a message)
- Ignore the second component (with or without a message)


> There are some last comments to address

Yeah sorry.. I got my chance to try that thing called Covid and been totally mashed potato in my head for the last week or so. I'm on it.

js/src/dom/data.js
alpadev: Mh looks like that's a problem of how the code is processed.

In the source there is `[...instanceMap.keys()].join(', ')` with the spread operator taking care of `keys()` returning an iterator object and iterates that automatically. While in the composed script it is replaced with `[].concat(instanceMap.keys()).join(', ')` which concats that iterator object into an array without iterating.

Replacing it with `Array.from()` to solve that problem.

js/tests/unit/dom/data.spec.js
alpadev: That's indeed true. I'm open for any feedback in that regard or how to deal with this topic.

rohit2sharma95: Since it is now supposed that multiple instances will not be supported, for now, you can just show an error in the console and then just return from there. Although the instance of the second plugin would not be created properly, there is an error in the console to let the user know about it.

rohit2sharma95 (approved):

js/src/dom/data.js
```diff
@@ -11,57 +11,47 @@
...
-      const keyProperties = element.bsKey
-      if (keyProperties.key === key) {
-        delete storeData[keyProperties.id]
-        delete element.bsKey
-      }
+    // make it clear we only want one instance per element
+    // can be removed later when multiple key/instances are fine to be used
+    if (!instanceMap.has(key) && instanceMap.size !== 0) {
+      // eslint-disable-next-line no-console
+      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance(s): ${Array.from(instanceMap.keys()).join(', ')}.`)
```
rohit2sharma95, line 28: Now you can omit **s** from the instance(s) since there is only one instance bound.
```suggestion
      console.error(`Notice: The design of Bootstrap currently discourages to use more than one instance per element. Bound instance: ${Array.from(instanceMap.keys())[0]}.`)

```

alpadev: I think we should also change the `discourages` to `disallow` then and maybe remove `currently` as this isn't something that is likely to be addressed in the closer future.

rohit2sharma95: /CC @XhmikosR

XhmikosR (approved):

PR #32184, opened 2020-11-17
Ability to add listeners directly to objects like Modals/Collapse/etc

### LINKED ISSUE #31266: [Feature][v5] Ability to add listeners directory to objects like Modals/Collapse/etc
How about a public method to add listener(s) an object, something like:

```js
class Modal {
    //...

    on(event, callback) {
        this._element.addEventListener(`${event}${EVENT_KEY}`, callback);

        return this;
    }

    //...
}
```

This will allows us to do something like:

```js
var myModal = new bootstrap.Modal(document.getElementById('myModal'));

myModal
    .on('shown', function (e) {
        // do something...
    })
    .on('hidden', function (e) {
        // do something...
    });
```

Instead of:

```js
var myModal = new bootstrap.Modal(document.getElementById('myModal'));

myModal._element.addEventListener('show.bs.modal', function (e) {
    // do something...
});
myModal._element.addEventListener('hidden.bs.modal', function (e) {
    // do something...
});
```

arcanedev-maroc:
Hi everyone,

I count on your guidance/help for my first PR :wink: 

As i mentioned in my suggestion (#31266), this PR will allow us to add event listeners directly from an instantiated object with a new `on()` method.

I've also added a new methd `off()` to remove a registered event listener. 

### TODO:

* [ ] Alert
* [ ] Button
* [ ] Carousel
* [ ] Collapse
* [ ] Dropdown
* [ ] Popover
* [x] Modal
* [ ] Scrollspy
* [ ] Tab
* [ ] Toast
* [ ] Tooltip

I'll start with `Modal` class just to be sure that i'm on the right path.

There is a class called `EventHandler`, do i need to use it to add/remove event listeners?

Closes: #31266

### DIFF
--- a/js/src/base-component.js
+++ b/js/src/base-component.js
@@ -30,6 +30,18 @@ class BaseComponent {
     this._element = null
   }
 
+  on(event, handler) {
+    this._element.addEventListener(`${event}.${this.constructor.DATA_KEY}`, handler)
+
+    return this
+  }
+
+  off(event, handler) {
+    this._element.removeEventListener(`${event}.${this.constructor.DATA_KEY}`, handler)
+
+    return this
+  }
+
   /** Static */
 
   static getInstance(element) {
--- a/js/tests/unit/modal.spec.js
+++ b/js/tests/unit/modal.spec.js
@@ -1119,4 +1119,60 @@ describe('Modal', () => {
       expect(Modal.getInstance(div)).toEqual(null)
     })
   })
+
+  describe('on', () => {
+    it('should register an event handler', done => {
+      fixtureEl.innerHTML = '<div class="modal"><div class="modal-dialog"></div></div>'
+
+      const modalEl = fixtureEl.querySelector('.modal')
+
+      const onShow = e => {
+        expect(e).toBeDefined()
+      }
+
+      const onShown = () => {
+        expect(modalEl.getAttribute('aria-modal')).toEqual('true')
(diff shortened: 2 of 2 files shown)


Johann-S:
Thanks @arcanedev-maroc 👍 

it would be better if we pass your MR after this one : https://github.com/twbs/bootstrap/pull/29370 because the `.on` and `.off` method will be in the base component and inherited from the other components

/CC @XhmikosR

XhmikosR:
True, but #29370 isn't ready and right now there's not much benefit from it IMHO.

So, I'd say, if this PR is ready soon, we can proceed with it and we can make any further refactoring later :)

XhmikosR:
@arcanedev-maroc the base component PR has landed if you want to rebase and update your branch.

arcanedev-maroc:
Hi @XhmikosR, i've made the changes on the base component.

**Questions:**
* Do i need to use `EventHandler` to add/remove the event listener?
* Are my tests good or do i need to change them ?

septatrix:
*Personally* I think it is more reasonable to properly expose the `element` attribute instead of adding shortcuts for everything. The `on` and `off` shortcuts feel like they would fit very well when jQuery ís detected however as jQuery provides similar methods.

Regardless of my personal preferences one thing which should be possible is passing other options to `addEventListener` like once, passive etc.

mdo:
Closing as stale.

PR #32185, opened 2020-11-17
terser: use 2 passes instead of the default 1.

XhmikosR:
Saves us a few bytes.

Preview: <https://deploy-preview-32185--twbs-bootstrap.netlify.app/>


Johann-S:
nice -0.1% 👍

mdo (approved):

PR #32191, opened 2020-11-18
Correct block button breakpoint docs typo

matthew-shaw:
breakpoing --> breakpoint


ffoodd (approved):
Thanks for spotting this!

PR #32192, opened 2020-11-18
Baseof v4

NotWoods:
Based on #31608, for #28479

Adds a base layout used by most other layouts. Also adds a base "redirect" layout to avoid some duplication.


XhmikosR:
Hey, @NotWoods thanks for the PR!

Can you fix the HTML validation issues please?

Also, I see you went with the "no footer" comment; isn't there any other way? Like we don't include footer.html in baseof.html and include it where used? What do you think?

XhmikosR:
Oops @NotWoods can you make a new PR please? GitHub auto-closed this after I merged the Hugo PR.

PR #32194, opened 2020-11-18
Docs: fix homepage npm install version

luooooob:
The `npm install bootstrap` should be `npm install bootstrap@next` on the homepage . People (such as me) will mistakenly think that v5 is already the latest version, especially when I see the following sentence is `
gem install bootstrap -v 5.0.0.alpha3`...


mdo:
We'll skip this for now I think since Beta 1 is right around the corner :).

mdo:
Well, maybe we still want it. Thoughts @XhmikosR?

XhmikosR:
Yeah, let's land it because the tag is needed until we release the stable version.

I'm gonna add it to #31057.

XhmikosR (approved):

PR #32195, opened 2020-11-19
docs: Fix syntax highlighting language

XhmikosR:
Split from #32077

Preview: <https://deploy-preview-32195--twbs-bootstrap.netlify.app/docs/4.5/content/reboot/#native-font-stack>


XhmikosR:
All right, this is borked with Jekyll/Rouge, so I'll keep the patch in #32077.

image

PR #32206, opened 2020-11-19
Revert colors.yml accidental changes

XhmikosR:
I accidentally changed these to the main ones in #32077.


XhmikosR:
Actually, it seems that file has unused stuff; the values aren't used, just the names. Any reason we had those @mdo?

mdo (approved):

PR #32207, opened 2020-11-19
feat(grid): simplify gutters usage

ffoodd:
Right after #31735, I think the same kind of simplification on every gutter usage could help a lot. It simplifies quite a lot of occurrencies, for a single more complex (in our `.gx-*` utilities, needs to `/ 2`).

## To-do

- [ ] documentation?
- [ ] check what needs to be added to the migration guide

---

- Preview: https://deploy-preview-32207--twbs-bootstrap.netlify.app/docs/5.0/examples/grid/
- Main: https://getbootstrap.com/docs/5.0/examples/grid/

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -330,14 +330,14 @@ $container-max-widths: (
 // Set the number of columns and specify the width of the gutters.
 
 $grid-columns:                12 !default;
-$grid-gutter-width:           1.5rem !default;
+$grid-gutter-width:           .75rem !default;
 $grid-row-columns:            6 !default;
 
 $gutters: $spacers !default;
 
 // Container padding
 
-$container-padding-x: $grid-gutter-width / 2 !default;
+$container-padding-x: $grid-gutter-width !default;
 
 
 // Components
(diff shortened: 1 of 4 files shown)


MartijnCuppens:
There is a difference between the gutter width and the padding around containers.
```scss
$grid-gutter-width:           .75rem !default;
```
This is incorrect. The spacing (gutter) between 2 columns is `1.5rem`, not `.75rem`.

ffoodd:
So I may introduce a new, more accurate variable like `$grid-gutter-half` or something?

ffoodd:
So to sum up, this PR simple drops a few `calc()` and Sass maths by using an intermediate `$grid-gutter-half` new variable. @MartijnCuppens @mdo Does this feel useful to you?

Still needs docs and a mention in our migration guide, if so.

MartijnCuppens:
As far as I know the "gutter" distance has always been the distance between columns, so this PR doesn't really make sense to me.

ffoodd:
Well, the gutter doesn't change at all: it's only decreasing cyclomatic complexity by dividing in a single place instead of a dozen.

Less divisions in Sass, less stylelint exceptions for `calc()`, less generated code.

But if everyone agrees, let's drop it.

ffoodd:
@mdo Feel free to close this one if this doesn't sound interesting :)

PR #32211, opened 2020-11-19
v4: Move negative margin-bottom from .nav-item to .nav-link

kyletsang:
This fix prevents the tab's bottom border from showing on the actively selected tab.  This bug is only visible when using `.nav-link`s as direct children of `.nav-tabs`.

This issue can be reproduced in the second docs example under this link: https://getbootstrap.com/docs/4.5/components/navs/#javascript-behavior

Here is an image showing before/after the fix.

image

Preview: <https://deploy-preview-32211--twbs-bootstrap.netlify.app/docs/4.5/components/navs/#javascript-behavior>


ffoodd (approved):
LGTM.

@MartijnCuppens @mdo, any thoughts?

mdo (approved):

PR #32214, opened 2020-11-20
Remove `version_short`.

XhmikosR:
We no longer use it in package.json.


XhmikosR:
We need to drop this since #30980 is using this again. We can revisit later if needed.

PR #32217, opened 2020-11-20
Ability to add custom class in tooltip/popover

### LINKED ISSUE #19415: Simplify the customization of popover

rohit2sharma95:
Porting of #31834 to main.

Fixes #19415


js/src/tooltip.js
```diff
@@ -761,6 +768,16 @@ class Tooltip {
...
+  _getCustomClass() {
+    const customClass = Manipulator.getDataAttribute(this.element, 'custom-class')
+
+    if (customClass) {
+      return customClass
+    }
+
+    return typeof this.config.customClass === 'function' ? this.config.customClass() : this.config.customClass
+  }
+
```
rohit2sharma95, line 780: I have done this as per #31834 though but IMO, `customClass` in the config should have more priority than in data attributes 🤔

Johann-S (approved):
LGTM 👍 

Thanks @rohit2sharma95

PR #32225, opened 2020-11-21
Partially backport #32217

rohit2sharma95:
Partially backport #32217

### DIFF
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -286,7 +286,7 @@ class Tooltip {
       this._popper = new Popper(this.element, tip, this._getPopperConfig(attachment))
 
       $(tip).addClass(CLASS_NAME_SHOW)
-      $(tip).addClass(this._getCustomClass())
+      $(tip).addClass(this.config.customClass)
 
       // If this is a touch-enabled device we add extra
       // empty mouseover listeners to the body's immediate children;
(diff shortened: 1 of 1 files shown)


XhmikosR:
I'm confused with the backports :/

js/src/tooltip.js
```diff
@@ -734,10 +734,6 @@ class Tooltip {
     this.config.animation = initConfigAnimation
   }
 
-  _getCustomClass() {
-    return this.element.getAttribute('data-custom-class') || this.config.customClass
```
XhmikosR, line 738: How is this not needed? 😵

rohit2sharma95: Because the `customClass` will be in the configuration already, even if it is provided in the data attributes 🤔 
https://github.com/twbs/bootstrap/blob/2d8c02eb1024ef2ce8877d6484a9ea8a4d9cd67f/js/src/tooltip.js#L660-L664

rohit2sharma95:
> I'm confused with the backports :/

Not exactly backport but minor fixes 🙂

js/src/tooltip.js
rohit2sharma95: And jQuery's [`addClass()`](https://api.jquery.com/addclass/) method will handle it, either `this.config.customClass` is a function or a class name (Even it is a blank string)

XhmikosR (approved):

XhmikosR:
@rohit2sharma95 can you move the typo fixes to a new PR please?

js/src/tooltip.js
Johann-S: yep good catch 👌

Johann-S (approved):

PR #32227, opened 2020-11-21
tests/integration/index.html: minor tweaks

XhmikosR:


XhmikosR:
@rohit2sharma95 can you try this file locally on main after you run `npm t`? I'm getting the following error and I don't quite get if it's a regression or not:

```
Uncaught TypeError: can't convert undefined to object
    getDataAttributes file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:3499
    _getConfig file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:6458
    Tooltip file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:5980
    <anonymous> file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:7579
    <anonymous> file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:7578
    EventListener.handleEvent* file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:7577
    <anonymous> file:///C:/Users/xmr/Desktop/bootstrap/js/coverage/bundle.js:7583
```

rohit2sharma95:
Hi, @XhmikosR 
I checked by reverting the source code to the [previous commit](https://github.com/twbs/bootstrap/tree/fe961c192d951bde610cdcd4817f4d8548d33141) before #31827 🙂 

I was also getting the following issue:
```Uncaught TypeError: element.addEventListener is not a function```

This is because of:
https://github.com/twbs/bootstrap/blob/21737ed530b7ac37fe1ce67026df373c04282ac8/js/tests/integration/bundle.js#L4-L7

Here the Tooltip constructor is getting the Nodelist instead of a single element (thats why tooltip is also not working), because the above code becomes:
```js
window.addEventListener('load', function () {
    [].concat(document.querySelectorAll('[data-bs-toggle="tooltip"]')).map(function (tooltipNode) {
      return new Tooltip(tooltipNode);
    });
  });
```

PR #32233, opened 2020-11-22
Update bundle.js in tests/integration

rohit2sharma95:
Use `Array.prototype.concat()` instead of direct spreading nodelist.

I am not sure if it should be done in [this file](https://github.com/twbs/bootstrap/blob/main/js/tests/integration/bundle-modularity.js) also @XhmikosR 

Ref: https://github.com/twbs/bootstrap/pull/32227#issuecomment-731785066

### DIFF
--- a/js/tests/integration/bundle-modularity.js
+++ b/js/tests/integration/bundle-modularity.js
@@ -2,6 +2,6 @@ import 'popper.js'
 import Tooltip from '../../dist/tooltip'
 
 window.addEventListener('load', () => {
-  [...document.querySelectorAll('[data-bs-toggle="tooltip"]')]
+  [].concat(...document.querySelectorAll('[data-bs-toggle="tooltip"]'))
     .map(tooltipNode => new Tooltip(tooltipNode))
 })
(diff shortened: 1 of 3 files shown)


XhmikosR:
@rohit2sharma95 good catch! Yes, please change the other instance too.

It seems I caused this regression [here](https://github.com/twbs/bootstrap/commit/46c037410b8c7eaab3cf50a5cf44093aa2fd41f4#diff-bca2597346dc6857e7359f8ae54bae1cf7b1ce3dfd87cf654c85478cdf681520) and never noticed it until a couple of days ago :/

What I don't get is why this worked when I changed the script to `dist/js/bootstrap.bundle.js`.

PS. Do you have an account on our Slack? If not, feel free to join and DM me :)

rohit2sharma95:
Tooltip is working after you changed the path of the script because it is being initialized here:
https://github.com/twbs/bootstrap/blob/5a8cd9526d1e7098119b688632a974d29aba6b59/js/tests/integration/index.html#L67-L69 

And carousel seems working on my local machine 🤔

rohit2sharma95:
It is being done in #31178 though
Ref: https://github.com/twbs/bootstrap/pull/31178#discussion_r512668334

XhmikosR:
Yeah, but it shouldn't be sneaked it like that :)

I'd rather merge it separately. I'll have a look tomorrow.

PR #32238, opened 2020-11-23
CI: add a 30-min timeout for BrowserStack

XhmikosR:
Could probably decrease it even more. If merged, I will backport it.


XhmikosR:
This should help us in cases of timeouts. The default timeout is either 4 or 6 hours.

mdo (approved):

PR #32240, opened 2020-11-24
CI: specify `OPENCOLLECTIVE_HIDE: true`

XhmikosR:


ffoodd:
Does it simple drop the funding message in console?

XhmikosR:
Yeah

XhmikosR:
TBH this shouldn't be needed https://github.com/opencollective/opencollective-postinstall/blob/d6627320e81bc4effa140f04998aab9ed7a064fa/index.js#L7

But apparently nodemon is using their own implementation https://github.com/remy/nodemon/blob/f74f4a20772b5d96506386d2e35569bc3e2c1811/bin/postinstall.js#L4

XhmikosR:
This might not be needed if https://github.com/remy/nodemon/pull/1800 is merged

PR #32244, opened 2020-11-24
Tweak `list-group` to show numbers on list items when used on an `<ol>`

### LINKED ISSUE #26202: Why hide the numbers in an ordered list?
I tried looking for an explanation or another issue that addresses this, but I didn't find anything.

What's the reasoning behind hiding the numbers or default list item style for an ordered list? There's no visual difference that I can see when using an ordered list vs. unordered list, which defeats the purpose of having this useful distinction.

patrickhlauke:
Closes #26202

Per my comment there, I'm not sure which approach would be best/most palatable to take. This PR goes for the simplest/most surgical one, but I know this goes against the grain of targetting specific elements and using child selectors. Then again, the alternative becomes overly verbose...

>the simplest would be modifying the list-group css styles to special-case if they're applied to `<ol>` rather than `<ul>`...something like
>
>```
>ol.list-group > .list-group-item {
>   display: list-item;
>   list-style-position: inside;
> }
> ```
>
> but i know we've been keen to avoid modifying the styles of bootstrap's classes based on which element they've been applied to. so...do we instead propose a whole separate set of styles? `.list-group-ordered` ? which replicates all the styles of `.list-group`, but then overrides this aspect?
>
>```
>.list-group-ordered > .list-group-item {
>  display: list-item;
>  list-style-position: inside;
>}
>```
>
>or do we do a modifier hanging off of `.list-group-item`? `.list-group-item-ordered`? to be used perhaps in conjunction with `.list-group-item`, so you'd end up with lengthy constructs like `<li class="list-group-item list-group-item-ordered">` ?
>
>```
>.list-group-item-ordered {
>  display: list-item;
>  list-style-position: inside;
>}
>```
>
>basically, there's variations on how we want to get to it. depends on the philosophical approach you want to take (purity of the CSS no referencing elements/special casing based on markup? a whole new set of classes? a modifier?)

Preview: <https://deploy-preview-32244--twbs-bootstrap.netlify.app/docs/5.0/components/list-group/#basic-example>


patrickhlauke:
as expected, this fails linting because of the qualifying selector. if we feel this is acceptable, can add the extra comment that tells the linter to ignore it...

XhmikosR:
Please try and move all the lorem ipsum changes to a separate PR. :)

Also, I'm not sure if this should be backported; only if it the changed isn't considered a breaking one.

patrickhlauke:
> Please try and move all the lorem ipsum changes to a separate PR. :)

yeah that was an accident, hadn't switched branches... :)

> Also, I'm not sure if this should be backported; only if it the changed isn't considered a breaking one.

fair enough, happy for this to be v5 only

ffoodd:
If we add a new class to handle this as a variant, it could be backported. If we're using the qualifying selector, it should'nt.

Using one or another depends on whether we consider this as a design thing, or as a semantic thing:
- should `ol` always show its counter when used as a list group? Could an `ol` not have any counter? In that case, why not use an `ul`
- should `ul` be able to display counter? In that case, why not use `ol`?

I'd be in favor of enforcing semantic by applying it to `ol` (thus allow the qualifying selector, and not backport it).

XhmikosR:
Just to point out that unfortunately there are cases where people don't control the HTML (CMS etc).

ffoodd:
Regarding CMSes, most of them allows to use (un)ordered lists since AFAIK all of them are using a WYSIWYG (like TinyMCE and/or Gutenberg for WordPress, or CKEditor for Drupal). Themes, plugins and modules can mess this up, but we can't consider all of them—and moreover, requiring a class is not easier than a specific element, in any case I can think of.

patrickhlauke:
one option i did not mention in the opener (and the original comment): could also be an additional modifier `.list-group-ordered` that gets added to the list, so that you'd have to use `<ol class="list-group list-group-ordered">`

it just gets a bit verbose, but less verbose than hanging it off of individual list items i guess

XhmikosR:
The above might be the best approach, indeed. 👍

ffoodd:
I don't think so, since it'd basically allow any `ul` to get numbers, and won't solve the `ol` case since you'd still have to add this class by yourself (which isn't the expected result on the original issue).

BTW if you choose to use this approach, it'd be more relevant to create something like `.list-ordered`, which could be applied to any kind of list, wouldn't it?

patrickhlauke:
> since it'd basically allow any `ul` to get numbers

well, if applied to `<ul>` it'll show the bullet (not numbers)

so, are we saying we want an entire set of `.list-group-ordered` that replicates the entirety of what `.list-group` currently does, rather than an additional class to use together with `.list-group` ?

(this is where the various philosophies on how this should best be handled clash ... i don't particularly care either way, as i'm not a purist ... adding it directly to `ol.list-group > .list-group-item` makes sense to me, personally)

ffoodd:
I'd be for `ol.list-group > .list-group-item` too.

> well, if applied to <ul> it'll show the bullet (not numbers)
Well, in that case `.list-orederd` feels wrong. Could be something like `list-styled` (as we have `.list-unstyled` but in that case it still feels wrong to me (since we could let `list-style` alone as a basis, and add `.list-unstyled` on `.list-group` if needed.

mdo:
I'm not a fan of adding this since it would require even more work to make the numbers hanging inside the list items. If folks want a count, I feel we should encourage them to do it in other ways. As it stands now, this is an incomplete measure, and feels not very useful.

patrickhlauke:
> it would require even more work to make the numbers hanging inside the list items.

what do you mean? the styles in this PR (whichever way we go here in terms of approach) take care of that

MartijnCuppens:
> > If folks want a count, I feel we should encourage them to do it in other ways.
>
> like, how then?

I stand by Patrick. An `<ol>` feels like the most semantic approach.

MartijnCuppens (approved):

mdo:
Sorry, thought I added a screenshot to show you what I mean. The issue is with longer strings of text that wrap to new lines. Doesn't look good to me, nor does it account for other layouts inside the list group items.

Screen Shot 2020-11-28 at 8 55 29 PM

In my view, the list group is rarely a set of single-line text—our docs examples further down the page show this. It's intended to be a versatile component overall. This feels like an incomplete measure given it doesn't solve for that.

patrickhlauke:
ah, i see what you mean @mdo ... i still think though that, at least as a starting point, this SC / allowing to use `<ol>` and not actually suppressing the numbers when an author clearly intended to convey something numbered is good. it may not be super pretty, but that can be tackled in later updates if needed.

it's basically the choice of:

* allow authors to use `<ol>` now, but they might need to do some more work to make it look a bit prettier in certain situations
* prevent authors from using `<ol>` even in situations where it might be semantically correct for them to do so

I'd still lean towards the former.

mdo:
I get that. I just want to see if we can do more than provide something that feels incomplete to me. We know folks will try to do more and run into this issue, and then we'll be expected to fix it. Let's at least hold on this until after beta 1.

mdo:
Updated project to Beta 2 per last comment.

mdo:
Coming back to this, hoping to take this CodePen demo from one-off to something customizable and workable for this. https://codepen.io/emdeoh/pen/abmQarE

Any immediate thoughts?

patrickhlauke:
> Coming back to this, hoping to take this CodePen demo from one-off to something customizable and workable for this. https://codepen.io/emdeoh/pen/abmQarE
> 
> Any immediate thoughts?

quite liking the look of that. what customisation options were you thinking? variables to change the appearance of the counters (colour, font-weight, etc)?

mdo:
Opened #33068. Safe to close this I think?

patrickhlauke:
> Opened #33068. Safe to close this I think?

sure

PR #32245, opened 2020-11-24
Allow constructors to accept a CSS selector

### LINKED ISSUE #29417: Allow our constructors to accept a CSS selector
Currently in all of our plugins constructors, the first parameter is an element, we should allow our users to use CSS selector too.

X-Ref: https://github.com/twbs/bootstrap/issues/29416

rohit2sharma95:
Preview: https://deploy-preview-32245--twbs-bootstrap.netlify.app/

Closes #29417


rohit2sharma95:
IMO this PR should move after #29370

rohit2sharma95:
@XhmikosR CSS selectors should be added in the code snippets in the docs?

site/content/docs/5.0/migration.md
```diff
@@ -7,6 +7,12 @@ aliases: "/migration/"
 toc: true
 ---
 
+## v5.0.0-beta3
+
+### JavaScript
+
+- All plugins are now allowed to accept a CSS selector as the first argument. You can either pass a DOM element or any valid CSS selector to create a new instance of the component.
```
XhmikosR, line 14: Maybe we should add an example here?

And perhaps in the JavaScript page https://getbootstrap.com/docs/5.0/getting-started/javascript/ instead of adding examples in each component page?

js/src/base-component.js
```diff
@@ -17,12 +17,14 @@ const VERSION = '5.0.0-beta1'
 
 class BaseComponent {
   constructor(element) {
+    element = typeof element === 'string' ? document.querySelector(element) : element
```
XhmikosR, line 20: We also have https://github.com/twbs/bootstrap/blob/7e6d7b8f817ad1212238f15124335d37c93ebfea/js/src/util/index.js#L90

Not sure which solution is better/safer?

rohit2sharma95: `isElement` allows jQuery object as well :thinking:

site/content/docs/5.0/migration.md
rohit2sharma95: Done :+1: 
https://deploy-preview-32245--twbs-bootstrap.netlify.app/docs/5.0/getting-started/javascript/#css-selectors-in-constructors

js/src/base-component.js
XhmikosR: So, we should probably stick to `isElement`, right?

rohit2sharma95: `isElement` should not be used IMO, since Bootstrap does not allow jQuery elements in the constructors.

site/content/docs/5.0/migration.md
```diff
@@ -9,6 +9,17 @@ toc: true
 
 ## v5.0.0-beta3
 
+### JavaScript
+
+- All plugins are now allowed to accept a CSS selector as the first argument. You can either pass a DOM element or any valid CSS selector to create a new instance of the plugin.
```
mdo, line 14: Small copy edit:

```suggestion
- All plugins can now accept a CSS selector as the first argument. You can either pass a DOM element or any valid CSS selector to create a new instance of the plugin.
```

mdo:
@rohit2sharma95 @XhmikosR I see this needs docs—anything I can help with?

mdo (approved):

PR #32249, opened 2020-11-24
breadcrumb: simplify appearance, improve extensibility

### LINKED ISSUE #32053: Breadcrumbs text and default links fail contrast ratio
https://v5.getbootstrap.com/docs/5.0/components/breadcrumb/

> contrast-bootstrap-breadcrumbs

The active text/separator for breadcrumbs uses `#6c757d` on `#e9ecef` with a contrast ratio of 4:1. The default links in breadcrumbs use `#0d6efd` on `#e9ecef` with a contrast ratio of 3.8:1.

bumping the text/separator from `$gray-600` to `$gray-700` (`#495057`) shoots up the contrat to 6.9:1. For the links, however, there's very little wiggle room ... not sure how/if that can be tackled -  `#0d6efd` has a ratio of 4.5:1 only against pure white, so even lightening the breadcrumbs background a bit won't help

patrickhlauke:
Closes #32053

Preview: https://deploy-preview-32249--twbs-bootstrap.netlify.app/docs/5.0/components/breadcrumb/


XhmikosR:
Shouldn't we mention this change in migration?

patrickhlauke:
ah, good point @XhmikosR ... doh

patrickhlauke:
also, i just ripped out the padding stuff, but should it perhaps go for an explicit `padding: 0` ?

MartijnCuppens:
We could `null`ify the variables instead of removing them.

https://markdotto.com/2019/08/30/null-sass-variables/

patrickhlauke:
sigh...well, whichever, somebody make a decision :)

MartijnCuppens:
If we set
```scss
$breadcrumb-padding-y:              0 !default;
$breadcrumb-padding-x:              0 !default;
$breadcrumb-bg:                     null !default;
$breadcrumb-border-radius:          null !default;
```
we'll still be able to adjust paddings,bg & radius with Sass variables and keep the result as is (in this PR).

ffoodd:
Agreed, didn't think about this but it would even not be a breaking change I guess, since variables would still exist 👌

patrickhlauke:
Right, done ... think it'd still be a breaking change if we change the default styles like this under people's feet though, so keeping this as v5 only at the moment

ffoodd (approved):
👌

XhmikosR:
Let's wait for another approval from @mdo please :)

mdo:
Pushed some content changes and the CSS custom property mention in https://github.com/twbs/bootstrap/issues/32053#issuecomment-732385060 along with @ffoodd's recommendation of the Sass fallback.

XhmikosR:
Some notes:

1. some further characters are escaped in the snippets
2. `style="--bs-breadcrumb-divider: ;"` doesn't seem to be valid. If we confirm it is valid, I could ask upstream for a fix

XhmikosR:
@twbs/css-review so, does this still look good to merge?

MartijnCuppens (approved):
Now it does

PR #32259, opened 2020-11-25
docs: SVG consistency changes

XhmikosR:
@ffoodd I only moved `xmlns` and `fill`, but the order could be further improved later if needed.

Preview: <https://deploy-preview-32259--twbs-bootstrap.netlify.app/>


ffoodd:
Indeed, IMHO what could be done is ordering them by occurrencies (the most used the first). Might even be automated, but as long as we're only ordering attreibutes it'd probably be over-engineered.

mdo (approved):

Issue #32266, opened 2020-11-26
Linear gradients slow down (large) table rendering by a lot


stgm:
This concerns the `table` styles in BS5 alpha 3. Apparently due to the recently introduced use of `linear-gradient` the rendering of my (rather large) tables becomes very slow. My tests have been done on a Macbook Pro i5 with Intel Iris Graphics 6100.

When including a very large table in a document, after scrolling down, the rendering of the next part of the document takes as much as a second. The screen is actually white while waiting for the renderer to "catch up". This does **not** happen when the `linear-gradient` backgrounds are disabled. However, in Bootstrap 5, disabling those will also remove visible striping, for example.

**--> Demonstration**

Below is a Codepen with a large table that clearly shows this problem in Safari 14.0.1 (16610.2.11.51.8) and Chrome 87.0.4280.67 on Mac Intel. In Firefox 83.0, the problem is less noticeable but still apparent.

https://codepen.io/stgm/pen/poEoryZ

You can uncomment the style in this Codepen to override the use of the linear-gradient and appreciate the difference. It may be easier to see the problem if you enable the "Full Page View", because you can then scroll faster and see how this seems to make the problem even worse.

**--> Relevant links**

Origin of the linear-gradient idea: https://github.com/twbs/bootstrap/issues/29244
First jab at integrating these: https://github.com/twbs/bootstrap/pull/30342
Final PR containing the linear gradients for tables: https://github.com/twbs/bootstrap/pull/30466

mdo:
Hmm yeah with tables that size, I can se this being an issue. We use the linear-gradient to stack background "colors" in certain situations.

/cc @twbs/css-review

XhmikosR:
Should we close this one @MartijnCuppens since #32277 is merged?

MartijnCuppens:
> There can still be a performance issue with striped tables though.

No, this is still an issue

stgm:
The fix in #32348 does indeed solve the performance issue on my original page! Great to see.

PR #32277, opened 2020-11-29
Remove default linear gradient

MartijnCuppens:
Remove the default invisible gradient causing the performance issue in https://github.com/twbs/bootstrap/issues/32266. By removing the custom property, the linear gradient will become invalid, thus not appear by default.

There can still be a performance issue with striped tables though.

### DIFF
--- a/scss/_tables.scss
+++ b/scss/_tables.scss
@@ -4,7 +4,6 @@
 
 .table {
   --#{$variable-prefix}table-bg: #{$table-bg};
-  --#{$variable-prefix}table-accent-bg: transparent;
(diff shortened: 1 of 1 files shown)


mdo (approved):

XhmikosR:
Did you guys confirm this fixes the issue mentioned in #32266? I couldn't reproduce the issue on my setup.

MartijnCuppens:
> Did you guys confirm this fixes the issue mentioned in #32266? I couldn't reproduce the issue on my setup.

Yes, it fixes it for the default tables.

PR #32278, opened 2020-11-29
Remove accordion jump

### LINKED ISSUE #32250: Accordion slight jump when toggling items

MartijnCuppens:
[Inspired by Eric](https://www.youtube.com/watch?v=NuWF26AWxvU), I added a little accordion jump in #32013. Turns out @XhmikosR didn't like that, so he opened #32250.

Method is similar to the technique we use for the input/button groups. Because we now have an overlap, I needed a non-transparent color to prevent the colors from mixing.

I still notice a little jump in Chrome (rendering issue in Chrome, macOS?), but I'm afraid that's out of our jurisdiction (tried some hacks with `will-change`, but without result).

There's also still a little jump on the last item, but we also add an additional border. We could "fix" this by adjusting the bottom padding, but this could become problematic when the paddings are removed. Also, because in reality the accordion content won't always be exactly the same, chances are the total accordion height will vary when sections are collapsed.

Closes #32251 which closes #32250

Preview: https://deploy-preview-32278--twbs-bootstrap.netlify.app/docs/5.0/components/accordion/#example


XhmikosR:
We can move this to beta2, if someone wants to take another stab at it because indeed there's still a little jump @mdo

XhmikosR:
BTW this seems to make the border 2x compared to what it is on main.

patrickhlauke:
the jump seems much more pronounced (for me at least) at 100% zoom (keep your eye on the 3rd accordion button and how that jiggles).

100-zoom

at 150% zoom it seems gone

150-zoom

might it be something to do with rounding errors?

at least the subtle jiggle in the content of the panels themselves when they close seems to have removed completely.

XhmikosR:
I personally see the same thing regardless of zoom or browser, but I'm on 175% DPI.

mdo:
Closing for #32278.

PR #32280, opened 2020-11-29
Add toast positioning

MartijnCuppens:
Alternative for https://github.com/twbs/bootstrap/pull/28783. Uses our built-in spacing utilities doesn't require JavaScript positioning.

Also added `translate-middle-x` & `translate-middle-y` utilities for more positioning possibilities.

Preview:

* https://deploy-preview-32280--twbs-bootstrap.netlify.app/docs/5.0/components/toasts/#placement
* https://deploy-preview-32280--twbs-bootstrap.netlify.app/docs/5.0/utilities/position/#center-elements

Closes https://github.com/twbs/bootstrap/pull/28783

### DIFF
--- a/scss/_toasts.scss
+++ b/scss/_toasts.scss
@@ -1,29 +1,31 @@
 .toast {
-  max-width: $toast-max-width;
+  width: $toast-max-width;
+  max-width: 100%;
   @include font-size($toast-font-size);
   color: $toast-color;
+  pointer-events: auto;
   background-color: $toast-background-color;
   background-clip: padding-box;
   border: $toast-border-width solid $toast-border-color;
   box-shadow: $toast-box-shadow;
-  opacity: 0;
   @include border-radius($toast-border-radius);
 
-  &:not(:last-child) {
-    margin-bottom: $toast-padding-x;
-  }
-
-  &.showing {
-    opacity: 1;
-  }
-
-  &.show {
-    display: block;
-    opacity: 1;
+  &:not(.showing):not(.show) {
+    opacity: 0;
   }
 
   &.hide {
     display: none;
+  }
+}
+
+.toast-container {
+  width: max-content;
+  max-width: 100%;
+  pointer-events: none;
+
+  > :not(:last-child) {
+    margin-bottom: $toast-spacing;
   }
 }
 
--- a/site/assets/js/application.js
+++ b/site/assets/js/application.js
@@ -28,6 +28,17 @@
       new bootstrap.Popover(popover)
     })
 
+  var toastPlacement = document.getElementById('toastPlacement')
+  if (toastPlacement) {
+    document.getElementById('selectToastPlacement').addEventListener('change', function () {
+      if (!toastPlacement.dataset.originalClass) {
+        toastPlacement.dataset.originalClass = toastPlacement.className
+      }
+
+      toastPlacement.className = toastPlacement.dataset.originalClass + ' ' + this.value
+    })
+  }
+
   document.querySelectorAll('.toast')
     .forEach(function (toastNode) {
       var toast = new bootstrap.Toast(toastNode, {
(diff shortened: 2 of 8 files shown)


site/assets/js/application.js
```diff
@@ -28,13 +28,22 @@
...
+        toastPlacement.dataset.originalClass = toastPlacement.className
+      }
+
+      toastPlacement.className = toastPlacement.dataset.originalClass + ' ' + this.value
+    })
+  }
+
   document.querySelectorAll('.toast')
     .forEach(function (toastNode) {
-      var toast = new bootstrap.Toast(toastNode, {
```
XhmikosR, line 33: Can you revert this change please?

XhmikosR:
From a really quick look, I definitely dig it ❤️ 

Needs more eyes from @twbs/css-review

site/assets/js/application.js
```diff
@@ -28,13 +28,22 @@
...
     })
 
+  var toastPlacement = document.getElementById('toastPlacement')
+  if (toastPlacement) {
+    document.getElementById('selectToastPlacement').addEventListener('change', function () {
+      if (!toastPlacement.dataset.originalClass) {
+        toastPlacement.dataset.originalClass = toastPlacement.className
+      }
+
+      toastPlacement.className = toastPlacement.dataset.originalClass + ' ' + this.value
```
XhmikosR, line 38: I assume you want to overwrite any previous classes here? If not, we could just use `classList.add()`?

MartijnCuppens: Yep, the class needs to be overridden since position utilities can be added.

site/assets/js/application.js
MartijnCuppens: Done

XhmikosR:
BTW the center widgets look weird on mobile

image

MartijnCuppens:
> BTW the center widgets look weird on mobile

Fixed that

mdo (approved):
Let's do it!

XhmikosR:
@mdo we have #28783 included in beta1. I'm not sure if these are 1 to 1 equivalent, but if so, I guess this one should be merged and not #28783.

mdo:
Widths need to be ironed out I think... by default, the toasts are shrunk here. Is that behavior we want overall? Would we want to change to this behavior at all times and let folks pick their own `max-width`s?

Screen Shot 2020-12-02 at 10 24 15 AM

It's not until I add wrapping content in one that they hit their `max-width` that's set on the `.toast`.

Screen Shot 2020-12-02 at 10 22 33 AM

scss/_toasts.scss
```diff
@@ -2,31 +2,30 @@
...
 
   &.hide {
     display: none;
   }
 }
 
+.toast-container {
+  display: grid;
+  width: max-content;
+  max-width: 100%;
```
mdo, line 24: Drop these two lines for now? We can add them later if we really need them, but I think we can let our toasts stay at their `max-width`.

```suggestion
```

mdo:
Also, could this arguably wait until another release? Not sure we need to lock in something for the beta release.

MartijnCuppens:
Good point, hadn't noticed that. We'll need the `width: max-content;` to fix https://github.com/twbs/bootstrap/pull/32280#issuecomment-735437987

Just fixed the shrinking thing by using margins between the toasts instead of the using grid.

mdo (approved):
Thank you!

XhmikosR:
Branch rebased and replaced the renamed classes.

@ffoodd RTL-wise does this LGTY?

ffoodd (approved):
LGTM!

Since position relies on our utilities, that's already handled and processed by RTLCSS :+1: 

Regarding cheatsheet, I don't think so but that soemthing to question: I didn't show utilities in the cheatsheet for now. I'll add it to our remaining to-do.

PR #32285, opened 2020-11-30
Refactor components to use a utility function to define jQuery plugins

### LINKED ISSUE #32286: Refactor components to use an utility function to define jquery plugins
# Problem
Currently every plugin/component defines some boilerplate code that will set it up as a jQuery plugin.

# Solution
To make the code a little bit less repetitve, I figured it could be some improvement to move this into an utility function and use that instead.

Draft: https://github.com/twbs/bootstrap/pull/32285

alpadev:
Closes #32286

TODO:

- [x] Tighten Bundlewatch limits before merging

Preview: <https://deploy-preview-32285--twbs-bootstrap.netlify.app/>


js/src/util/index.js
```diff
@@ -203,5 +219,6 @@ export {
   noop,
   reflow,
   getjQuery,
-  onDOMContentLoaded
+  onDOMContentLoaded,
```
XhmikosR, line 222: Is the `onDOMContentLoaded` export still used?

alpadev: Not that I'm aware of. I kept it like that because there is already a spec for it and to keep it testable. Also I wasn't sure if there would be some use for it in the future.

XhmikosR:
My only concern is that this shouldn't be part of the public API since it's not used anywhere else. That being said, there must be some way we can still test it.

@rohit2sharma95 thoughts?

XhmikosR:
On the other hand, it seems we already export other stuff already in utils.

XhmikosR:
BTW thanks for the PR, it definitely reduces the code and duplication :)

js/src/util/index.js
XhmikosR: As the last resort we could rename this to `_onDOMContentLoaded`, but since we test jQuery already, testing these individually might not be needed after all.

rohit2sharma95: It seems a better approach

XhmikosR: Actually, NVM my comment. Util isn't exposed in our public API https://github.com/twbs/bootstrap/blob/main/js/index.umd.js

So, I'd say @alpadev you can go ahead with this PR and ping us when it's ready. 👍

rohit2sharma95 (approved):

PR #32289, opened 2020-11-30
Add a little more right margin for "info icon"

### LINKED ISSUE #32188: Little more right-margin for "info icon" in <select> when class is "is-invalid"
[Reference](https://v5.getbootstrap.com/docs/5.0/forms/validation/#server-side)

bharatmakvana:
Fixes #32188 can you please check it.


mdo:
This isn't the right solution—the icon can be valid, invalid, or something else entirely.

PR #32304, opened 2020-12-01
Unbreak lines.

XhmikosR:
We generally don't enforce a line width, but this is nitpicking and opinions will differ.

Some lines were really too short.


patrickhlauke:
this would WFM

mdo (approved):

PR #32309, opened 2020-12-02
tests: replace deprecated jQuery method

rohit2sharma95:
The `.on()` and `.trigger()` methods can set an event handler or generate an event for any event type, and should be used instead of the shortcut methods.

Event aliases like `.click()` will likely never be removed, but jQuery encourages the use of alternatives.

https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/


XhmikosR:
I wonder if this really makes sense? I mean, from what we know, `click()` will never be removed.

rohit2sharma95:
I found it in the blog post of jQuery: https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/
image

XhmikosR:
Yes, but it explicitly says that `click()` will likely never be removed :)

Do we have any other instances? How about v4?

rohit2sharma95:
There are no more instances for the jQuery's click method (to trigger the click event) in V5 and in V4 I can see only the native click method of the DOM elements.

XhmikosR (approved):

PR #32313, opened 2020-12-02
build-plugins: build Base component too

XhmikosR:
Not sure about this, but I noticed that currently the base plugin is inlined in each plugin's dist file.


rohit2sharma95 (approved):

XhmikosR:
We should clean up this script later. I don't see why we need so much duplication.

XhmikosR:
So, I'm not sure about this. It will add a new dependency for each component. On the other hand, it will reduce duplication in each individual plugin dist file.

If we plan to move more stuff to the Base component, maybe it makes sense to build it separately.

Oh, and we need to document this somehow.

PR #32314, opened 2020-12-02
CONTRIBUTING: mention dist files

XhmikosR:


mdo (approved):
Tweaked a a few words.

Issue #32034, opened 2020-11-01
Stack space for flow elements


thewebartisan7:
This feature is best explained here:
https://every-layout.dev/layouts/stack/

And an example can be found here:
https://jsfiddle.net/c2fbo0u3/1/

In the example I have just added two different space size, 1.5rem and large 3rem, but could be also added more.

As explained in every-layout.dev this is useful when we need to add space between elements, and in addition I added space X and Y, something like already bootstrap m-* utilities offer.

So instead of writing:

```html
<div>
<div></div>
<div class="mt-3"></div>
<div class="mt-3"></div>
</div>
```

Can be:

```html
<div class="stack-y">
<div></div>
<div></div>
<div></div>
</div>
```

Or instead of:

```html
<div class="d-flex">
<div></div>
<div class="ml-3"></div>
<div class="ml-3"></div>
</div>
```

Can be:

```html
<div class="d-flex stack-x">
<div></div>
<div></div>
<div></div>
</div>
```

Here https://jsfiddle.net/c2fbo0u3/1/ there are all examples, also creating a reverse space for space-x, and nested.

Nested could be also created without using space-nested but just adding for each nested element the same space-x/y classes. Example:

```html
<div class="stack-x stack-nested d-flex">
  <div>Stack nested X One</div>
  <div>Stack nested X Two</div>
  <div class="d-flex">
    <div>Stack nested X One</div>
    <div>Stack nested X Two</div>
    <div>Stack nested X Three</div>
  </div>
</div>
```

Can be:

```html
<div class="stack-x d-flex">
  <div>Stack nested X One</div>
  <div>Stack nested X Two</div>
  <div class="stack-x d-flex">
    <div>Stack nested X One</div>
    <div>Stack nested X Two</div>
    <div>Stack nested X Three</div>
  </div>
</div>
```

mdo:
This is very familiar to [Pylon's hstack/vstack approach](https://almonk.github.io/pylon/). Very intriguing to me also considering #31995 and not wanting to have a single component-based layout option. I'd go further than here and suggest that it be `display: flex` by default and offer inline-spacers via CSS customer properties. Example: https://codepen.io/emdeoh/pen/Bazxbrq.

thewebartisan7:
Yes I agree to be already display flex, that is how I setup mine custom at the moment. 
[...]

PR #32320, opened 2020-12-03
Stub out a .quick-grid helper class

mdo:
Idea is to enable super fast grid layouts. We could maybe back in some `gap` and `row` behavior—~~maybe providing a `--rows` like `--columns`?~~—to allow random layouts on the fly.

**Edit:** Definitely not going to include a `--rows` since it would defeat the original purpose of creating a vertical stack of elements.

Inspired by #32034.

/cc @twbs/css-review 

---

### Potential todos

- [ ] Rename to something else?
- [ ] Replace block button docs with this instead?


thewebartisan7:
I don't use css grid yet so but only flexbox.
And most of my current design that use the stack helper is added on a flexbox element.

So I am unsure of using css grid only for add space. 

The goal of this utility is not to add a grid, but space.

So maybe would be better another approuch that can be used in any type of stack.

"quick-grid" is not really appropriate name I think.

mdo:
Yeah I was debating naming this "stacks", but that implies z-index more than layout. Open to other ideas!

thewebartisan7:
I see what you mean with "stacks" (example https://css-tricks.com/how-to-stack-elements-in-css/), but that is better than "grid" which implies a grid layout.

Personally I like more your first example with flexbox and margin, named vstack/hstack, here https://codepen.io/emdeoh/pen/Bazxbrq

This should just replace margin utilities, so if you don't like stacks, consider word spacers, spacing, space...

thewebartisan7:
> Definitely not going to include a --rows since it would defeat the original purpose of creating a vertical stack of elements.

Not sure which original purpose you mean, but in all example mentioned here https://github.com/twbs/bootstrap/issues/32034, plyon included, the idea behind is to create both vertical and horizontal spacing.

mdo:
Replaced by #33986.

PR #32332, opened 2020-12-04
Move migration sidebar entry to sidebar.yml

XhmikosR:
* Non-whitespace diff: <https://github.com/twbs/bootstrap/pull/32332/files?w=1>
* Preview: <https://deploy-preview-32332--twbs-bootstrap.netlify.app/docs/5.0/getting-started/introduction/>

TODO:

- [x] Fix active styling when the current `.bd-links .btn[aria-expanded="true"]` doesn't not cover the case, like in the Migration link


mdo:
Right now this breaks the sidebar text rendering (too bold) and I don't see a Migration link anywhere.

XhmikosR:
> Right now this breaks the sidebar text rendering (too bold) and I don't see a Migration link anywhere.

Too bold? It's bold when it's active only?

But yeah, the last patch is WIP. If you revert it, it will work. I'm just not sure what our plan are, like I say in my OP:

>Decide how to handle the case with more group without subpages

XhmikosR:
I still need some help with the remaining TODO, but assuming we always want the separator to show for each group entry that doesn't have sub pages, this should be better for maintenance.

ffoodd:
@XhmikosR I simplified the active state markup-side, since the active class can be added on the link (as it's done for subnavs) without anyside-effect.

Also fixed an invalid value CSS-side, flagged by my IDE.

site/layouts/partials/docs-sidebar.html
```diff
@@ -34,8 +34,8 @@
       </li>
     {{- else }}
       <li class="my-3 mx-4 border-top"></li>
-      <li{{ if $is_active_group }} class="active"{{ end }}>
-        <a href="/docs/{{ $.Site.Params.docs_version }}/{{ $group_slug }}/" class="d-inline-flex align-items-center rounded">
+      <li>
+        <a href="/docs/{{ $.Site.Params.docs_version }}/{{ $group_slug }}/" class="d-inline-flex align-items-center rounded{{ if $is_active_group }} active{{ end }}"{{ if $is_active_group }} aria-current="page"{{ end }}>
```
XhmikosR, line 38: @patrickhlauke LGTY?

patrickhlauke: Yup

patrickhlauke (approved):

mdo (approved):

PR #32341, opened 2020-12-05
Fixed offset in scrollspy component

sehyun-hwang:
When `getBoundingClientRect` method is used and when page is refreshed to a scroll position that is not at the top but below it, clicking a nav button will result in scrolling to wrong position. I was able to fix this giving offset of `window.pageYOffset`

### DIFF
--- a/js/src/scrollspy.js
+++ b/js/src/scrollspy.js
@@ -106,7 +106,7 @@ class ScrollSpy extends BaseComponent {
 
     const offsetBase = offsetMethod === METHOD_POSITION ?
       this._getScrollTop() :
-      0
+      window.pageYOffset
 
     this._offsets = []
     this._targets = []


XhmikosR:
Thanks for the PR.

1. do we have an issue for this?
2. is the issue present in v4 too?
2. we definitely need a test added

aegggsehyun1994:
> Thanks for the PR.
> 
> 1. do we have an issue for this?
> 2. is the issue present in v4 too?
> 3. we definitely need a test added

1. I read all the issues related to scrollspy, and I haven't found any related one.

2. v4 worked fine, but I noticed that before the initialization of scrollspy the page is scrolled to a certain position. This behavior is not present on v5, and therefore the values in the offset array is offset by the scroll position of the page during initialization. You may add this behavior to v5, but user experience would be compromised. My PR will allow a scroll position to be remembered and the component to work as expected on clicking on the nav.

3. I may add a test script that checks whether the `active` class is added correct nav item. Is that fine?

XhmikosR:
@GeoSot @rohit2sharma95 can you help @sehyun-hwang with adding a test case here? ❤

rohit2sharma95:
@sehyun-hwang Can you please create a reduced test case first for the issue that you found, via [CodePen](https://codepen.io/) or [JS Bin](https://jsbin.com/). :slightly_smiling_face:

PR #32357, opened 2020-12-07
Add focus visibility on tabs with pills

### LINKED ISSUE #31427: [Accessibility] nav pills tabs : The focus on active link is not visible
On tabs with pills the focus on an **active link is not visible** (see https://v5.getbootstrap.com/docs/5.0/components/navs/) : 
TabsActiveLinkFocused

But on **link button** the **focus is visible** (see https://v5.getbootstrap.com/docs/5.0/components/buttons/) : 

activeLinkFocused

There is no focus style visible on **Firefox** browser for this active link, so we don't see the focus (we lose where we are).

To be accessible the **focus on an active link must be visible** (see success criterion 2.4.7 : https://www.w3.org/TR/WCAG21/#focus-visible).

mayralgr:
Fixes #31427 

Proposal: add the scss variables for the focus

css_nav_pills

Preview: https://deploy-preview-32357--twbs-bootstrap.netlify.app/docs/5.0/components/navs-tabs/#javascript-behavior


patrickhlauke:
I think this should specifically target only the pills nav links, as this PR in its current form affects all nav links - see, for instance, the top navigation in the header (Home, Docs, Examples ...) https://deploy-preview-32357--twbs-bootstrap.netlify.app/

Nav with hovered 'examples' showing the blue hover outline

scss/_variables.scss
```diff
@@ -854,6 +854,11 @@ $nav-link-hover-color:              null !default;
 $nav-link-transition:               color .15s ease-in-out, background-color .15s ease-in-out, border-color .15s ease-in-out !default;
 $nav-link-disabled-color:           $gray-600 !default;
 
+$nav-link-focus-width:             .25rem !default;
+$nav-link-focus-color-opacity:     .25 !default;
+$nav-link-focus-color:             rgba($component-active-bg, $nav-link-focus-color-opacity) !default;
+$nav-link-focus-box-shadow:        0 0 0 $nav-link-focus-width $nav-link-focus-color !default;
```
patrickhlauke, line 860: make it specific to `$nav-pills-link...`

patrickhlauke (changes requested):

mayralgr:
@patrickhlauke changes where made, let me know if another change is needed

mdo (changes requested):
We shouldn't be setting focus styles (the box-shadow) on hover. Will review in more depth likely after beta 2.

mayralgr:
@mdo I left it just with the focus. Let me know if it is the correct approach.

patrickhlauke:
> We shouldn't be setting focus styles (the box-shadow) on hover. Will review in more depth likely after beta 2.

well, it was just replicating what buttons do...i personally wouldn't have minded.

patrickhlauke:
beyond that though, https://deploy-preview-32357--twbs-bootstrap.netlify.app/docs/5.0/components/navs-tabs/ looks good to me

patrickhlauke (approved):

mayralgr:
@mdo is the anything else needed ? Thank you!

patrickhlauke:
This looks good to me. @mdo / @twbs/css-review ?

ESA-Security-Group:
16 gången du mailat mig nu? Var fan är du egentligen?

Location...!?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

mdo:
Sorry I let this fall off my radar.

Couple things come to mind now:

- We set `:hover, :focus` on all `.nav-link`s, so that's my bad.
- However, we don't actually set any focus styles. Seems to me if we wanted to do this right, we'd set separate hover and focus styles.
- And, we'd apply or customize those styles by nav modifier class (pills vs tabs).
- Safari, Chrome, and Firefox all have separate focus styles. In Safari, I end up seeing double focus outlines. Chrome it's a dark outline. Firefox, kinda nothing, but sometimes a thin dotted line.

Will dive into this a bit myself and see what comes to mind for v5.

Screen Shot 2021-01-12 at 10 26 48 PM
Screen Shot 2021-01-12 at 10 27 34 PM

patrickhlauke:
> Safari, Chrome, and Firefox all have separate focus styles. In Safari, I end up seeing double focus outlines. Chrome it's a dark outline. Firefox, kinda nothing, but sometimes a thin dotted line.

for the regular (non-pill) tabs, this needs to be this way because otherwise, you either define `:focus` explicitly and then it applies for mouse/touch users when they interact with the tabs as well (which would look odd), OR you use `:focus-visible` but then polyfill it ... it's the same old problem that keeps bubbling up.

could be sorted for pills I guess, as there it would look less disruptive if an explicit `:focus` is defined and applied even after mouse/touch.

mdo:
I think the doubling up I saw was because `outline` wasn't reset when `box-shadow` was added.

patrickhlauke:
> I think the doubling up I saw was because `outline` wasn't reset when `box-shadow` was added.

ah gotcha, we can fix it just for the pill ones, yes

scss/_nav.scss
```diff
@@ -81,6 +81,11 @@
 .nav-pills {
   .nav-link {
     @include border-radius($nav-pills-border-radius);
+    &:focus {
+      color: $nav-pills-link-hover-color;
+      text-decoration: if($link-hover-decoration == underline, none, null);
+      box-shadow: $nav-pills-link-focus-box-shadow;
+    }
```
patrickhlauke, line 88: add an `outline: 0` to suppress double focus styles

mayralgr:
Hi guys I did this minor change but I am having trouble with the tests, as I left it I think I need a rebase or something ( I try but still gives an error ) any help ? 😅

patrickhlauke:
i think you forgot to `git pull` changes that were made to your branch here (as a result of rebasing/merging in changes here on github) before making the latest change, so your PR suddenly included lots of reverts of things that blew the tests up. i think i managed to manually fix it

XhmikosR:
Haven't looked at it too close, but it seems like the outline results in duplicate borders for active tabs and -1px or so for inactive ones.

patrickhlauke:
> Haven't looked at it too close, but it seems like the outline results in duplicate borders for active tabs and -1px or so for inactive ones.

not sure i'm following what you're seeing. which browser? which example specifically? got a screenshot/pointer to it?

XhmikosR:
Sorry, for the late reply. I probably mixed the examples and I was looking at the `nav-tabs` example, not the `nav-pills` one.

It works, but the focus style seems to overlap with the other nav-items.

patrickhlauke:
> Sorry, for the late reply. I probably mixed the examples and I was looking at the `nav-tabs` example, not the `nav-pills` one.
> 
> It works, but the focus style seems to overlap with the other nav-items.

Yeah I think there's no clean way around this. As that overlapping focus style only shows for keyboard users as they navigate through, and momentarily (before either the link is followed or, in case of dynamic tabs, when that tab then becomes active) it's something I think is not completely deal-breaking (otherwise we have to start moving to inner shadows or something rather that box shadows)

> 
> Edit: also browsers seem to have their own focus styles already, and with this patch we will have an inconsistent style in nav-tabs?

If you go to https://getbootstrap.com/docs/5.0/components/navs-tabs/#pills with Firefox (at least on Windows), there's no visible default focus indication. See this animation where I go back and forth with `Tab` and note how on the active one nothing can be seen when it gets focus:

firefox-pills-focus

With this change, it's at least consistent across all browsers.

firefox-pills-focus2

Leaving it up to browser default in this case is not a good idea, because of this variability (while yes, Chrome for instance has very clear default focus outline).

XhmikosR:
All right, let's wait for @twbs/css-review to have a look. I'd swear there was some kind of CSS property we recently used in another RP .

patrickhlauke:
> All right, let's wait for @twbs/css-review to have a look. I'd swear there was some kind of CSS property we recently used in another RP .

ah, you might be thinking of https://developer.mozilla.org/en-US/docs/Web/CSS/isolation ... hmm, this may help yes, but not sure how the semi-transparent box shadow over the active pill would show, if at all

ffoodd:
@XhmikosR @patrickhlauke Just tried, it doesn't help there since our `box-shadow` is the same color than our `background`, with a transparency. Either we change this color (and drop transparency, but that sounds inconsistent with other styles) to make it noticeable, or we consider adding space between pills?

There might be other approaches but I'm unable to find one for now.

XhmikosR:
I'd vote to just keep it simple and consistent for the time being :)

AstroOrbis (approved):

patrickhlauke:
This has slipped a bit off our radar... @mdo and @twbs/css-review in general, what are your feelings on this? it's probably not perfect for all circumstances, but does patch a gap in accessibility we currently have.

ffoodd (approved):
LGTM, not perfect but a good patch for now 👌

Focus styles are a whole topic deserving closer attention.

GeoSot:
@twbs/css-review  may we put this on a following project or close it?

mdo:
Focus styles coming in #33125.

PR #32364, opened 2020-12-07
Wrong pattern for custom checkboxes

ffoodd:
Seems like a missed thing from moving to Hugo? Or in any backport maybe?


ffoodd:
Invalid [as per the docs](https://v4-dev--twbs-bootstrap.netlify.app/docs/4.5/components/forms/#checkboxes-and-radios), sorry!

PR #32371, opened 2020-12-07
Support .focus on form-control

kelunik:
This allows pseudo focus for custom wrappers, e.g. around https://tiptap.dev/title.

### DIFF
--- a/scss/forms/_form-control.scss
+++ b/scss/forms/_form-control.scss
@@ -31,7 +31,7 @@
   }
 
   // Customize the `:focus` state to imitate native WebKit styles.
-  &:focus {
+  &:focus, &.focus {
     color: $input-focus-color;
     background-color: $input-focus-bg;
     border-color: $input-focus-border-color;


mdo:
We don't do `.focus` on any other components or elements, so I'm inclined to punt on this. Thanks, though!

kelunik:
@mdo It's supposed to be used by consumers, not bootstrap itself. It requires copying the styles otherwise, as there's no mixin that can be used. It'd be great to have one source for the focus styles.

**Example**

```vue
<div
   :id="'form-element-' + resolvedFormValidator.formElementId"
   :class="{
      'is-invalid': hasErrors,
      'is-valid': isValid && showValid,
      'focus': hasFocus
   }"
   class="form-control crm-wysiwyg"
>
   <editor-content
      :disabled="disabled"
      :editor="editor"
   />
</div>
```

Another use case is a styleguide that wants to show default, hover, focus styles side by side.

PR #32376, opened 2020-12-08
Support Popper virtual elements

### LINKED ISSUE #32365: Support Popper v2 virtual element
Dropdowns accept a `reference` option for Popper however the type is restricted to `string|element`.

Popper itself on the other hands also allows objects which provide a `getBoundingClientRect` method.
This can be quite useful when one wants implement custom context menus, arbitrary popup location on maps/images or otherwise does not have an actual DOM element as a reference but instead only raw coordinates.

The only thing required should be to also accept object (or only object which have `getBoundingClientRect` set).

septatrix:
Fixes #32365


js/src/dropdown.js
```diff
@@ -14,6 +14,7 @@ import {
   isVisible,
   isRTL,
   noop,
+  toType,
```
XhmikosR, line 17: This is missing

septatrix: Can I add this to the export in util.js or should I perform the check manually?

alecpl:
I think `getBoundingClientRect` is available in all supported browsers, no?

XhmikosR:
> I think `getBoundingClientRect` is available in all supported browsers, no?

Yup. We already make use of it in the codebase.

js/src/dropdown.js
XhmikosR: Just make sure the PR passes any tests locally before pushing :)

septatrix: Yeah I did this as a quick draft in the github web ui but have now cloned everything locally

septatrix:
> I think `getBoundingClientRect` is available in all supported browsers, no?

Yes but if you refer to the check that is due to `getBoundingClientRect` being required for popper virtual elements which can be arbitrary objects as long as the provide this method.

alecpl:
@septatrix nevermind my comment, I've now learned what this is about ;) Sorry for noise.

septatrix:
I tidied up the commit history a bit.

I also thought about adding an example using a virtual element but I do not think it is worthwhile as it would be longer than most other examples and most most of the other options do not really have examples as well. However if you also would want to add an example I would suggest something like I drafted in [this codepen](https://codepen.io/Septatrix/pen/MWjeNLe) which uses virtual elements to provide a custom contextmenu at the cursor position. (The codepen uses some trickery to make bootstrap think the passed object is an element as it does not use a patched version.)

septatrix:
@XhmikosR should I rebase them again to squash your commits into the previous ones or do PRs get squashed when merging? I only read in the contribution guidelines that you would like to have a clean commit log :D

XhmikosR:
All good, we can squash when merging.

rohit2sharma95:
But the reference that is passed in the configuration is not being used for the reference of Popper :thinking: 
This code is still the same:
https://github.com/twbs/bootstrap/blob/4e7c9e56fc76b8cf5b34d9d7b041b3904fb6404c/js/src/dropdown.js#L164-L175

So even if you pass the reference as an object, it does not make any change in the dropdown.
@septatrix :slightly_smiling_face:

septatrix:
Fixed

js/tests/unit/dropdown.spec.js
```diff
@@ -367,6 +367,48 @@ describe('Dropdown', () => {
...
+      const dropdown = new Dropdown(btnDropdown, {
+        reference: {
+          getBoundingClientRect: () => ({
+            width: 0,
+            height: 0,
+            top: 0,
+            right: 0,
+            bottom: 0,
+            left: 0
+          })
```
rohit2sharma95, line 399: It would be better if you create a spy and then expect that to have been called after showing the dropdown.

rohit2sharma95 (commented):
If you add the `.dropdown-menu-end` class in the `.dropdown-menu`, the position of the dropdown is wrong (position is not as per the passed coordinates). But that's not the issue of Popper. Works well without `.dropdown-menu-end`.

septatrix:
> If you add the `.dropdown-menu-end` class in the `.dropdown-menu`, the position of the dropdown is wrong (position is not as per the passed coordinates). But that's not the issue of Popper. Works well without `.dropdown-menu-end`.

What exactly do you mean with wrong. It is now to the left of the passed coordinates as long as there is space for it which is what I would expect. Some bad positioning happens at e.g. lower screen edges but that also happens to normal dropdowns.

js/tests/unit/dropdown.spec.js
septatrix: I was only testing the proper handling of types in general but adding a spy is a good idea. Will update this soon.

septatrix:
> Some bad positioning happens at e.g. lower screen edges but that also happens to normal dropdowns.

Nevermind. It just changes to `left` positioning so that it is centered above the cursor so that is working as intended i guess.

js/tests/unit/dropdown.spec.js
septatrix: This turned out to be a bit tricky as the internal `update` method on popper elements in asynchronous and popper does not expose events as bootstrap does. Luckily however there is a callback for the first update which I used instead.

rohit2sharma95 (approved):
LGTM :+1: 

Maybe in the future, tooltip/popover can also use this virtual element feature to follow the mouse cursor.

js/src/dropdown.js
```diff
@@ -257,6 +259,13 @@ class Dropdown extends BaseComponent {
 
     typeCheckConfig(NAME, config, this.constructor.DefaultType)
 
+    if (typeof config.reference === 'object' && !isElement(config.reference) &&
+      typeof config.reference.getBoundingClientRect !== 'function'
+    ) {
+      // Popper virtual elements require a getBoundingClientRect method
+      throw new Error(`${NAME}: Option "reference" provided type "object" without a required "getBoundingClientRect" method.`)
```
XhmikosR, line 266: Should we change this to a `TypeError` instead of the generic `Error`?

rohit2sharma95: True, missed it :smile: 
But then it should also be TypeError :thinking: 
https://github.com/twbs/bootstrap/blob/44667d89fa1bf8de15db57fad43696c62860e288/js/src/util/index.js#L119-L123

XhmikosR: OK, let's leave it as is for now and we tackle it later.

Issue #32354, opened 2020-12-06
Popper 2.x TODO


XhmikosR:
- [x] ~~Update Popper docs links to v2 (search for `https://popper.js.org/docs/` but don't blindly replace them!)~~ Fixed by #32390
- [x] ~~Update docs after #32405~~ Fixed by #32437 (Only for Tooltip/Popover)
- [x] ~~Update the docs for `fallbackPlacements` for dropdown — https://github.com/twbs/bootstrap/issues/32506#issuecomment-746509273~~ Changes were reverted in #32843 
- [ ] Add more JS tests to cover the changes in #31178

septatrix:
#32365 should probably also be a part of this

XhmikosR:
Not really, unless a PR is submitted.

septatrix:
Okay, I will submit one if that is okay.

I also noticed that popper shows a warning in the console (at least when referencing the scripts manually and not using the minified bundle). Look inside the console of [this pen](https://codepen.io/Septatrix/pen/dypXKgx?editors=1111) when clicking the button. This is due to `.dropdown-menu` specifying a margin which poppler overwrites anyway. Do you want to track this here or create a new issue?

XhmikosR:
Please make a new issue about the second thing. We should find a way to fix this before the stable release :)

XhmikosR:
@rohit2sharma95 do we still need to cover the last TODO here?

rohit2sharma95:
Yes, I see there are some missing tests. I will update the draft PR.

PR #32390, opened 2020-12-08
docs: Fix Popper-related docs

kyletsang:
Updates the Popper v1 links to v2.

Addresses part of  #32354


kyletsang:
Also noticed this snippet in the migration guide:

> Updated Popper to v2.x:
Removed offset option from our Tooltip/Popover and Dropdown plugins; this can still be achieved using the popperConfig parameter.
The fallbackPlacement option has become fallbackPlacements.

I've fixed the docs to remove the `offset` option in Popover and rename the `fallbackPlacement` for both Popover/Tooltip.

XhmikosR (approved):
Thanks!

PR #32393, opened 2020-12-09
carousel: switch to string constants

XhmikosR:
The logic is the same, minus the `toUpperCase()` call. Should have done this in #30358.

@patrickhlauke I assume `event.pointerType` are always lowercase, right?


rohit2sharma95 (approved):

js/src/carousel.js
```diff
@@ -278,7 +276,7 @@ class Carousel extends BaseComponent {
 
   _addTouchEventListeners() {
     const start = event => {
-      if (this._pointerEvent && PointerType[event.pointerType.toUpperCase()]) {
+      if (this._pointerEvent && (event.pointerType === POINTER_TYPE_PEN || event.pointerType === POINTER_TYPE_TOUCH)) {
```
patrickhlauke, line 279: how about defining POINTER_TYPE_MOUSE and checking here simply that it's NOT the mouse?

but yes seems kosher even as it is

patrickhlauke (approved):

patrickhlauke:
and yes, the spec defines these as lowercase values https://w3c.github.io/pointerevents/#dom-pointerevent-pointertype

patrickhlauke:
sorry, wrong button press...embarassing

js/src/carousel.js
XhmikosR: I thought I'd go with the exact same checks, but we can change it later if needed :)

PR #32394, opened 2020-12-09
Remove `TRANSITION_END` from utils

XhmikosR:
This was useful in v4-dev and back when we supported older browsers.

TODO:

- [x] Drop the variable completely from plugins?

### DIFF
--- a/js/src/alert.js
+++ b/js/src/alert.js
@@ -7,7 +7,6 @@
 
 import {
   defineJQueryPlugin,
-  TRANSITION_END,
   emulateTransitionEnd,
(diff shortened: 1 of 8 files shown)


rohit2sharma95:
What if you use `'transitionend'` instead of `TRANSITION_END` in utils also? :thinking: Because now it is used everywhere else as a string.

XhmikosR:
TBH it mostly seems like unnecessary optimization right now. It made sense back when we supported more browsers.

rohit2sharma95 (approved):

PR #32397, opened 2020-12-09
Utility API, RFS option: Only generate responsive classes when needed

MartijnCuppens:
I bumped into these classes which shouldn't be generated when `responsive` isn't set `true`:
https://github.com/twbs/bootstrap/blob/b1290ff189f03fae12a74ece09b25069d6f97f1a/dist/css/bootstrap.css#L10631-L10673


XhmikosR:
@MartijnCuppens Is this a regression from the RTL PR?

mdo (approved):

mdo:
> Is this a regression from the RTL PR?

I don't believe so—this looks like a mix up from `rfs` and `responsive` options maybe.

PR #32405, opened 2020-12-09
Add Popper fallback behaviour

### LINKED ISSUE #32384: Popover overflows container (regression)
Operating system and version (Windows, macOS, Android, iOS)
- Windows 10, Build 19042 (10.0.19042)

Browser and version (Chrome, Firefox, Safari, Microsoft Edge, Opera, Android Browser)
- Google Chrome, Version 87.0.4280.88 (Official Build) (64-bit)

--

_Note: This bug seems to have been introduced [during the transition](https://github.com/twbs/bootstrap/commit/adfdf7160b5822aae12eea677e7dd3128d2569bf) from `"popper.js": "^1.16.1"` to `"@popperjs/core": "^2.5.4"`._

### Steps to reproduce the problem

1. Visit https://getbootstrap.com/docs/4.5/components/popovers/#four-directions
2. Trigger "Popover on Left" and "Popover on Right"
3. Notice both adjust their position to be fully visible inside the body when viewport is sufficiently narrow
4. Visit https://getbootstrap.com/docs/5.0/components/popovers/#four-directions
5. Trigger "Popover on Left" and "Popover on Right"
6. Notice both overflow the body when viewport is sufficiently narrow

### What is the expected behavior?

Expected behavior is for popovers to not overflow their container (like in Bootstrap 4.5.1).

### What went wrong?

It seems like there's been some change in the inline styles that are applied by Popper where we no longer perform:
`transform: translate3d(x, y, z)` 
and instead use:
`transform: translate(x, y)`.

### Any other comments?

Maybe [this change](https://github.com/popperjs/popper-core/blame/671f71e2b6990765e5ca780daaea7d1f7175beb6/src/modifiers/computeStyles.js#L106) from 10 months ago had something to do with it.

MartijnCuppens:
Closes https://github.com/twbs/bootstrap/issues/32384

Demo: https://deploy-preview-32405--twbs-bootstrap.netlify.app/docs/5.0/components/popovers/#four-directions

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -298,6 +298,12 @@ class Dropdown extends BaseComponent {
           altBoundary: this._config.flip,
           rootBoundary: this._config.boundary
         }
+      },
+      {
+        name: 'flip',
+        options: {
+          fallbackPlacements: ['top', 'right', 'bottom', 'left']
+        }
       }]
     }
 
--- a/js/src/tooltip.js
+++ b/js/src/tooltip.js
@@ -470,7 +470,8 @@ class Tooltip extends BaseComponent {
     const flipModifier = {
       name: 'flip',
       options: {
-        altBoundary: true
+        altBoundary: true,
+        fallbackPlacements: ['top', 'right', 'bottom', 'left']
       }
     }
 


mdo (approved):

XhmikosR:
@mdo this needed docs. Also makes me wonder why we needed this since @Johann-S added this as optional (null).

XhmikosR:
@MartijnCuppens can you please update the docs tomorrow?

PR #32406, opened 2020-12-09
Use manipulator methods

rohit2sharma95:
Use manipulator methods everywhere to make consistency in the codebase

### DIFF
--- a/js/src/carousel.js
+++ b/js/src/carousel.js
@@ -418,7 +418,7 @@ class Carousel extends BaseComponent {
       return
     }
 
-    const elementInterval = Manipulator.getDataAttribute(element, 'interval')
+    const elementInterval = Number.parseInt(Manipulator.getDataAttribute(element, 'interval'), 10)
 
     if (elementInterval) {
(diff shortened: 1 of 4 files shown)


js/src/carousel.js
```diff
@@ -418,7 +418,7 @@ class Carousel extends BaseComponent {
       return
     }
 
-    const elementInterval = Number.parseInt(element.getAttribute('data-bs-interval'), 10)
+    const elementInterval = Manipulator.getDataAttribute(element, 'interval')
```
XhmikosR, line 421: `Number.parentInt` should probably stay, right?

rohit2sharma95: The data is already being normalized in manipular:
```js
if (val === Number(val).toString()) {
  return Number(val)
}
```

But that does not parse the number in integer though :thinking:

rohit2sharma95: Updated :+1:

XhmikosR:
I'm not sure this is worth doing.

Let's revisit later if needed.

PR #32417, opened 2020-12-10
Update Release Drafter configuration

XhmikosR:
* add "Performance" section
* add "CI" label in "Misc" category


XhmikosR:
Perhaps we should add a "Breaking Changes" section too? We already have the [breaking change label](https://github.com/twbs/bootstrap/labels/breaking%20change). And maybe Tests and Utility API too?

rohit2sharma95 (approved):

PR #32419, opened 2020-12-10
Tweak .form-select padding

### LINKED ISSUE #32188: Little more right-margin for "info icon" in <select> when class is "is-invalid"

### LINKED ISSUE #32356: [v5] .form-select arrow not aligned perfectly
I'd like the arrow to be aligned with the validation icons. Right now the icon position as well as the padding is different.
select

mdo:
Fixes #32356 and fixes #32188. Still might need some work, and I might be able to simplify things further.

Screen Shot 2020-12-10 at 9 24 43 AM

Screen Shot 2020-12-10 at 9 24 29 AM


mdo:
Second commit here makes some slightly bigger changes to simplify the padding. Unsure if it's a possible breaking change though, so would love feedback from @twbs/css-review.

Screen Shot 2020-12-10 at 9 31 46 AM

Screen Shot 2020-12-10 at 9 31 36 AM

MartijnCuppens:
The complex `calc()` functions were added to prevent breaking changes in v4.3, but your simpler approach makes more sense to me. I'm ok with simplifying it a little at this point.

MartijnCuppens (approved):

PR #32424, opened 2020-12-10
WIP: More CSS vars

mdo:
This PR adds additional CSS custom properties to our `:root` in the hopes of improving ease of customization and extensibility.

## Approach

In some situations, I'm just adding CSS vars for the sake of adding them. The more we expose here, the more folks can do with Bootstrap. Of course there's a filesize issue to consider, but we'll manage that by keeping things pretty tight for now. Beyond that, I'm creating CSS vars for some situations and replacing their Sass variables with interpolated CSS vars (e.g., instead of `$body-bg` it's `var(--bs-body-bg)` which has a computed value of `$body-bg`).

## What's new

- Added all `$grays`
- Added `$spacers` (ideally we could make the CSS var versions powered by `--bs-spacer`, but unsure how to do that)
- Added grid column count and gutter width, and replaced their Sass values in our mixins
- Added body values (font, color, background-color) and replaced their Sass values in Reboot

Still more to think about and fine-tune here.

## All CSS variables

```css
--bs-blue: #0d6efd;
--bs-indigo: #6610f2;
--bs-purple: #6f42c1;
--bs-pink: #d63384;
--bs-red: #dc3545;
--bs-orange: #fd7e14;
--bs-yellow: #ffc107;
--bs-green: #198754;
--bs-teal: #20c997;
--bs-cyan: #0dcaf0;
--bs-white: #fff;
--bs-gray: #6c757d;
--bs-gray-dark: #343a40;
--bs-100: #f8f9fa;
--bs-200: #e9ecef;
--bs-300: #dee2e6;
--bs-400: #ced4da;
--bs-500: #adb5bd;
--bs-600: #6c757d;
--bs-700: #495057;
--bs-800: #343a40;
--bs-900: #212529;
--bs-primary: #0d6efd;
--bs-secondary: #6c757d;
--bs-success: #198754;
--bs-info: #0dcaf0;
--bs-warning: #ffc107;
--bs-danger: #dc3545;
--bs-light: #f8f9fa;
--bs-dark: #212529;
--bs-font-sans-serif: system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", "Liberation Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
--bs-font-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
--bs-gradient: linear-gradient(180deg, rgba(255, 255, 255, 0.15), rgba(255, 255, 255, 0));
--bs-body-font: var(--bs-font-sans-serif);
--bs-body-font-size: 1rem;
--bs-body-color: #212529;
--bs-body-bg: #fff;
--bs-grid-columns: 12;
--bs-grid-gutter-width: 1.5rem;
--bs-spacer: 1rem;
--bs-spacer-0: 0;
--bs-spacer-1: 0.25rem;
--bs-spacer-2: 0.5rem;
--bs-spacer-3: 1rem;
--bs-spacer-4: 1.5rem;
--bs-spacer-5: 3rem;
```

### DIFF
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -30,12 +30,13 @@
   --#{$variable-prefix}grid-gutter-width: #{$grid-gutter-width};
 
   // Spacing
-  // TODO: Ideally we'd size these values based on the CSS var instead of the Sass map, so they'd be live updating, but TBD on feasibility
+  // Ideally we'd size these values based on the CSS var instead of the Sass map, so they'd be live updating, but TBD on feasibility
   --#{$variable-prefix}spacer: #{$spacer};
   @each $size, $value in $spacers {
     --#{$variable-prefix}spacer-#{$size}: #{$value};
   }
 
+  // Breakpoints
   @each $breakpoint, $width in $grid-breakpoints {
     @if $width > 0 {
       --#{$variable-prefix}breakpoint-#{$breakpoint}: #{$width};
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -4,6 +4,10 @@
     --#{$variable-prefix}#{$color}: #{$value};
   }
 
+  @each $color, $value in $grays {
+    --#{$variable-prefix}gray-#{$color}: #{$value};
+  }
+
   @each $color, $value in $theme-colors {
     --#{$variable-prefix}#{$color}: #{$value};
   }
@@ -13,4 +17,22 @@
   --#{$variable-prefix}font-sans-serif: #{inspect($font-family-sans-serif)};
   --#{$variable-prefix}font-monospace: #{inspect($font-family-monospace)};
   --#{$variable-prefix}gradient: #{$gradient};
+
+  // Root and body
+  --#{$variable-prefix}root-font-size: #{$font-size-root};
+  --#{$variable-prefix}body-font: #{$font-family-base};
+  --#{$variable-prefix}body-font-size: #{$font-size-base};
+  --#{$variable-prefix}body-color: #{$body-color};
+  --#{$variable-prefix}body-bg: #{$body-bg};
+
+  // Layout
+  --#{$variable-prefix}grid-columns: #{$grid-columns};
+  --#{$variable-prefix}grid-gutter-width: #{$grid-gutter-width};
+
+  // Spacing
+  // TODO: Ideally we'd size these values based on the CSS var instead of the Sass map, so they'd be live updating, but TBD on feasibility
+  --#{$variable-prefix}spacer: #{$spacer};
+  @each $size, $value in $spacers {
+    --#{$variable-prefix}spacer-#{$size}: #{$value};
+  }
 }
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -2,14 +2,18 @@
   // Custom variable values only support SassScript inside `#{}`.
   @each $color, $value in $colors {
     --#{$variable-prefix}#{$color}: #{$value};
+    --#{$variable-prefix}#{$color}-rgb: #{hex-to-rgb($value)};
   }
 
   @each $color, $value in $grays {
     --#{$variable-prefix}gray-#{$color}: #{$value};
+    --#{$variable-prefix}gray-#{$color}-rgb: #{hex-to-rgb($value)};
   }
 
   @each $color, $value in $theme-colors {
+    // --#{$variable-prefix}#{$color}: rgba(#{hex-to-rgb($value)}, var(--#{$variable-prefix}color-opacity, 1));
     --#{$variable-prefix}#{$color}: #{$value};
+    --#{$variable-prefix}#{$color}-rgb: #{hex-to-rgb($value)};
   }
 
   // Use `inspect` for lists so that quoted items keep the quotes.
--- a/scss/_root.scss
+++ b/scss/_root.scss
@@ -35,4 +35,10 @@
   @each $size, $value in $spacers {
     --#{$variable-prefix}spacer-#{$size}: #{$value};
   }
+
+  @each $breakpoint, $width in $grid-breakpoints {
+    @if $width > 0 {
+      --#{$variable-prefix}breakpoint-#{$breakpoint}: #{$width};
+    }
+  }
 }
--- a/.bundlewatch.config.json
+++ b/.bundlewatch.config.json
@@ -26,11 +26,11 @@
     },
     {
       "path": "./dist/css/bootstrap.css",
-      "maxSize": "24.25 kB"
+      "maxSize": "25 kB"
     },
     {
       "path": "./dist/css/bootstrap.min.css",
-      "maxSize": "22.25 kB"
+      "maxSize": "22.5 kB"
     },
     {
       "path": "./dist/js/bootstrap.bundle.js",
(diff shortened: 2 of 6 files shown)


ffoodd:
I'm not in favor of declaring those on `:root` since it increases file size quite a lot and is not what allows customization:   using `var()` in appropriate places is what allows it.

As @MartijnCuppens did in some components, using eg. `var(--bs-gutter-y, #{$grid-gutter-width})` lets anyone customize value by defining the custom property by themselves —without requiring us to define it in a first place.

I think a component-basis would be a better approach for this.

Moreover there's no need to rush on this, since we should be able to instantiate more and more custom properties without breaking anything, thanks to `var()` fallback mechanism. IMHO that'll be a long-term project.

mdo:
> I'm not in favor of declaring those on `:root` since it increases file size quite a lot and is not what allows customization: using `var()` in appropriate places is what allows it.

Yeah, file size can be a bummer here, but there are some values I think that need/deserve to be `:root` level. I don't think all variables should be `:root` fwiw.

> As @MartijnCuppens did in some components, using eg. `var(--bs-gutter-y, #{$grid-gutter-width})` lets anyone customize value by defining the custom property by themselves —without requiring us to define it in a first place.
> 
> I think a component-basis would be a better approach for this.

I'm not ruling anything out here, mostly experimenting with what we could do and putting it into practice to get feedback.

> Moreover there's no need to rush on this, since we should be able to instantiate more and more custom properties without breaking anything, thanks to `var()` fallback mechanism. IMHO that'll be a long-term project.

In no rush here :). Not looking to ship anything around this until after the stable release :).

dennisreimann:
Nice, that's great to see! 

> I'm not ruling anything out here, mostly experimenting with what we could do and putting it into practice to get feedback.

Have outlined our use case in [this comment](https://github.com/twbs/bootstrap/issues/26596#issuecomment-706088248), which is mostly related to themeing (light/dark) and from what I see here this will already help us, thanks for that :)

It'll also help to extend this to the component level. What comes to my mind would be something like this: (pseudo-code, limited to the colors and spaces)

```css
.alert {
  /* component vars, definition for those that are used multiple times */
  --padding-vertical: var(--alert-padding-vertical, var(--bs-spacer-2));
  --padding-horizontal: var(--alert-padding-horizontal, var(--bs-spacer-3));

  /* component styles */
  padding: var(--padding-vertical) var(--padding-horizontal);
}

.alert-primary {
  /* values that are used once and offer the ability to define overrides */
  color: var(--alert-primary-color-text, #084298);
  background-color: var(--alert-primary-color-bg, #cfe2ff);
  border-color: var(--alert-primary-color-border, #b6d4fe);
}
```

This way `--alert-padding-*` and `--alert-primary-color-*` could be overrided, but would fallback to the general BS definitions. Having this would offer a good balance between file size and customizability imho.

felixfbecker:
> Moreover there's no need to rush on this, since we should be able to instantiate more and more custom properties without breaking anything, thanks to var() fallback mechanism. IMHO that'll be a long-term project.

Are you sure that it's all possible without breaking changes? Buttons and alerts use Sass functions with variables and `darken()`, I am not 100% confident that there is a solution there that is not breaking any users of those functions or variables, or at least not a "nice" one. But I may be wrong.

dennisreimann:
@felixfbecker SASS generated values would have to be swapped with static ones. Those could be generated once and get inserted for the SASS vars, ideally offering a custom prop for overriding – e.g. `color: var(--override-color, #121212);` (the hex being a previously function based value). 

That's what we've done so far when generating a custom version of Bootstrap 4, that supports custom props.

ydmitry:
> @felixfbecker SASS generated values would have to be swapped with static ones. Those could be generated once and get inserted for the SASS vars, ideally offering a custom prop for overriding – e.g. `color: var(--override-color, #121212);` (the hex being a previously function based value).
> 
> That's what we've done so far when generating a custom version of Bootstrap 4, that supports custom props.

Seems to be for backward compatibility static values are not enough. E.g. for Sass based themes already existing in apps.

To do backward compatible need to replace Sass variables to CSS variables with fallback to Sass variable

```
color: var(--bs-some-color, #{$some-color});
```

mdo:
> > Moreover there's no need to rush on this, since we should be able to instantiate more and more custom properties without breaking anything, thanks to var() fallback mechanism. IMHO that'll be a long-term project.
> 
> Are you sure that it's all possible without breaking changes? Buttons and alerts use Sass functions with variables and `darken()`, I am not 100% confident that there is a solution there that is not breaking any users of those functions or variables, or at least not a "nice" one. But I may be wrong.

@felixfbecker With us already on the beta train, we're not able to make breaking changes, hence the "no need to rush" comment. We're still figuring out our approach here, as this PR indicates.

leomp12:
Hey guys, first of all, +1 to replace all colors SASS variables with CSS vars, it'll be awesome for light/dark themes and for page builders with live preview :1st_place_medal: 

But imho it's not necessary to use CSS vars for all :thinking:
For example, always use `var(--bs-spacer-{n})` would increase bundle size and I don't think it's really useful, I think only `--bs-spacer` root var is sufficient, and output styles would still having spaces explicitly (eg.: `1rem`) set, with SASS variables for build time customization only.

About colors, [HSL](https://elad.medium.com/why-css-hsl-colors-are-better-83b1e0b6eead) is not so common, but it's really better for use with CSS vars, specially to customize and extend default Bootstrap components, something like:

```css
:root {
  --bs-primary-hs: 60, 100%;
  --bs-secondary-hs: 120, 100%;
  --bs-lighten: 15%;
  --bs-primary: hsl(var(--bs-primary-hs), 60%);
  --bs-primary-light: hsl(var(--bs-primary-hs), calc(60% + var(--bs-lighten));
}
```

Wdyt?

themeix:
Waiting for the update regarding the CSS variables support.. It would be a great addition.

scss/_root.scss
```diff
@@ -13,4 +17,22 @@
...
+  // Layout
+  --#{$variable-prefix}grid-columns: #{$grid-columns};
+  --#{$variable-prefix}grid-gutter-width: #{$grid-gutter-width};
+
+  // Spacing
+  // TODO: Ideally we'd size these values based on the CSS var instead of the Sass map, so they'd be live updating, but TBD on feasibility
+  --#{$variable-prefix}spacer: #{$spacer};
+  @each $size, $value in $spacers {
+    --#{$variable-prefix}spacer-#{$size}: #{$value};
+  }
```
jonathanhefner, line 37: Because users can put arbitrary values in `$spacers`, there may not be a definitive relationship between those values and `$spacer` that can be mimicked by the CSS variables.  But, it may still be useful to scale the variables based on their ratio:

```suggestion
  @each $size, $value in $spacers {
    --#{$variable-prefix}spacer-#{$size}: calc(var(--#{$variable-prefix}spacer) * #{$value / $spacer});
  }
```

dennisreimann:
@mdo I love to see how this is progressing, thanks for that! How far do you see custom props - especially those for colors - coming in the final v5? 

As mentioned previously, the main hurdle is the usage of color modifying SASS functions, because we cannot pipe `var(--)` definitions into those... 
For those interested how we solved it for now with the current beta2: generating a basic Bootstrap version with SASS variables set to the custom properties where possible and the replacing the generated color values with the actual custom props is a good option by now. Here's our current approach, which I see to be more and more obsoleted as the custom prop support in Bootstrap progresses: https://github.com/btcpayserver/btcpayserver-design/pull/2/files#diff-647d383dd67d8e36fb5adb51111a90806be17e2549b05478c64fb60767228355

mdo:
Closing for #34100.

Issue #32330, opened 2020-12-04
RTL: remaining tasks


ffoodd:
Mostly around cheatsheet automation.

## Questions

- [ ] Move cheatsheet from examples to main docs?
  - [ ] Allow main docs to be passed a `direction` in the front matter and have that automatically toggle LTR to RTL CSS inclusion
  - [ ] _The above items should resolve most of this, but..._ De-dupe the RTL CSS once moved
  - [ ] Convert page content to Markdown
  - [ ] And https://github.com/twbs/bootstrap/pull/30980#issuecomment-737096962 :
    - Does our doc layout need to be converted to RTL too?
    - Or do we only embed RTL content in our current LTR doc layout?
    - What about translations, if we manage to automate cheatsheet content?
- [ ] consider to show utilities in the cheatsheet?

## Need help

- [x] check and update Arabic translations in [RTL cheatsheet](https://getbootstrap.com/docs/5.0/examples/cheatsheet-rtl/) and [RTL examples](https://getbootstrap.com/docs/5.0/examples/#rtl) → in progress in #33097
- [ ] new example showing bidi content, as [asked by @arafalov (comment)](https://github.com/twbs/bootstrap/issues/32330#issuecomment-743319982)
  - That's a valid use case for `<bdi>` and `<bdo>`, I guess:  [HTML5Doctor.com content](http://html5doctor.com/element-index/#bdi) and [MDN docs for `bdi`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi) and [`bdo`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdo)
- [x] Allow JS components to work in both directions in the same page? See #33076 #32446 #32913 

## Docs improvements

- [x] maybe mention specific how-tos? → #32771
  - [x] wrapping outputs with `.ltr` / `.rtl`: https://github.com/twbs/bootstrap/issues/32666#issuecomment-753880015
  - [x] rename / alias utilities: https://github.com/twbs/bootstrap/discussions/32402#discussioncomment-194697 (probably in our "Customize / Sass" or "Utilities / API" page?)
- [x] mention in our migration that if you had a RTL page styled using v4, you'd need to rename `.ml-*` to `.me-*` and `.mr-*` to `.ms-*`—basically the opposite of LTR. → #32771
- [x] Maybe explain more explicitly the single Sass souce and the need to run RTLCSS, in a custom build? See #32816 for example.
[...]

PR #32436, opened 2020-12-11
RTL improvements

### LINKED ISSUE #32327: RTL bugs
- Windows 10
- Google Chrome 87

- [x] a weird bug in `form-switch`:
  image
  
  Notice the switch circle which isn't vertically centered.
  It only happens in unchecked unfocused state, not when checked or when it has focus.
- [ ] In this [url](https://twbs-bootstrap.netlify.app/docs/5.0/examples/cheatsheet-rtl/), please take a look at this:
  image

  I guess we should flip this chevrons like this:

  image

  Which can be done with `transform: rotate(180deg)`.  Also the `transform-origin` can be increased to `.6em` to have more space between chevron and text.
- [ ] Carousel direction is wrong in RTL when changed by keyboard or touch, however left/right icons works correct on click [See fix](https://github.com/twbs/bootstrap/issues/32327#issuecomment-739673942)

ffoodd:
Fixes #32327

- [x] Cheatsheet sidenav arrows
- [x] Carousel keyboard direction
- [ ] Missing carousel JS test to check for direction: see #32330


ffoodd:
FWIW it's only missing the JS tests, but it's tracked in #32330 too. So this PR might be considered as simple patches.

FaridAghili:
@XhmikosR @ffoodd 
Keyboard direction is fixed and works as expected in RTL.

Same should be done for touch too: [here](https://github.com/twbs/bootstrap/blob/801c79d80ec06a5ea1323653a621e7f5618b8e34/js/src/carousel.js#L254) and [here](https://github.com/twbs/bootstrap/blob/801c79d80ec06a5ea1323653a621e7f5618b8e34/js/src/carousel.js#L259)

ffoodd:
Thanks @FaridAghili, fixed ;)

XhmikosR (approved):

PR #32437, opened 2020-12-11
Tooltip/popover - change the default value for `fallbackPlacements`

rohit2sharma95:
The default value for `fallbackPlacements` has been changed already in #32405 and here are the remaining things to do. 

### What this PR does:
- Remove `null` from `fallbackPlacements` types
- Change the default value for `fallbackPlacements`
- Update docs for the tooltip/popover

### Why to remove `null` from `fallbackPlacements` types and keep array:
- The default value was previously (in v4) `'flip'` that can be achieved by passing the single value in the array, like — `[oppositePlacement]`. Keeping `null` also sets the `fallbackPlacements` to `[oppositePlacement]` (Default value in Popper)
- It's better to have **clockwise** (`['top', 'right', 'bottom', 'left']`) fallback options so that tooltip/popover can be placed to another side even if the opposite placement doesn't fit.

BTW #32384 is caused because the default value for `fallbackPlacements` is `[oppositePlacement]` in Popper and if you check the issue on a bit larger screen, it is not reproduced.

As per Popper's documentation:

> [If the fallback placement doesn't fit, **it reverts back to the original placement**.](https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements)

---

| Not enough space on right (That's why did not flip) | Enough space on right (Flipped) | After setting clockwise fallback placements in #32405 |
| --- | --- | --- |
|  |  |  |

_All SS are from v5.0.0-beta1_

This PR probably completes the third todo of #32354: _**Update docs after #32405**_

Preview: <https://deploy-preview-32437--twbs-bootstrap.netlify.app/docs/5.0/components/popovers/#example>


site/content/docs/5.0/components/popovers.md
```diff
@@ -249,9 +249,9 @@ Note that for security reasons the `sanitize`, `sanitizeFn`, and `allowList` opt
     </tr>
     <tr>
       <td><code>fallbackPlacements</code></td>
-      <td>string | array</td>
-      <td><code>'flip'</code></td>
-      <td>Allow to specify which position Popper will use on fallback. For more information refer to
+      <td>array</td>
+      <td><code>['top', 'right', 'bottom', 'left']</code></td>
+      <td>Define fallback placements by providing a list of placements in array (indexed by preferences). For more information refer to
```
XhmikosR, line 254: I don't quite get what does this mean exactly? "indexed by preferences" 😕

rohit2sharma95: For example, if you put: `['top', 'right']`, Popper will place the tooltip on top first and if the tooltip does not fit on the top then on the right side.

XhmikosR: We need to reword this maybe. @mdo: any suggestions?

patrickhlauke: "indexed by preferences" > "in order of preference" ?

site/content/docs/5.0/components/tooltips.md
```diff
@@ -254,9 +254,9 @@ Note that for security reasons the `sanitize`, `sanitizeFn`, and `allowList` opt
     </tr>
     <tr>
       <td><code>fallbackPlacements</code></td>
-      <td>null | array</td>
-      <td><code>null</code></td>
-      <td>Allow to specify which position Popper will use on fallback. For more information refer to
+      <td>array</td>
+      <td><code>['top', 'right', 'bottom', 'left']</code></td>
+      <td>Define fallback placements by providing a list of placements in array (indexed by preferences). For more information refer to
```
patrickhlauke, line 259: "indexed by preferences" > "in order of preference" ?

patrickhlauke (dismissed):

XhmikosR:
Should we mention this in migration notes @rohit2sharma95 ?

rohit2sharma95:
Mentioning it would be better because now the default fallback placement has been changed.

XhmikosR (approved):

PR #32438, opened 2020-12-11
collapse: prevent url change if `A` nested tag is clicked

joke2k:
Removing jQuery the target of the click event is not the A tag if the user clicks on a nested tag, 
ignoring the preventDefault and changing the hash in the url.

### What this PR does:
- Add `event.delegateTarget` into the check for preventDefault
- Test this behaviour

### Details:

this code:
```
<a role="button" data-bs-toggle="collapse" class="collapsed" href="#collapse">
  <span id="nested">Clickme</span>
</a>
<div id="collapse" class="collapse"></div>
```

bubbles ups the default behaviour and changes the URL appending #collapse to it because the `event.target` is not an `A` tag, this let fail the check for `event.preventDefault`.

See on bs5:

https://github.com/twbs/bootstrap/blob/764e5298dea17e15bbcbf792204a45a761bedc0a/js/src/collapse.js#L372-L376

On bs4 jQuery did the work:

https://github.com/twbs/bootstrap/blob/a716fb03f965dc0846df479e14388b1b4b93d7ce/js/src/collapse.js#L361-L365

it happened with this commit https://github.com/twbs/bootstrap/commit/69e4d4f3ac897870e8186e22b19c5adf6043f3d4#diff-3cc7f0d8580d17c90274789ffc45552c56fed69ac7404d65e88ccccb81ffff31R378


XhmikosR:
We need tests for this, can you add a test please?

joke2k:
emm... tests were already in the PR :sweat_smile:

XhmikosR:
Oops, sorry I missed them!

js/tests/unit/collapse.spec.js
```diff
@@ -374,6 +374,29 @@ describe('Collapse', () => {
   })
 
   describe('data-api', () => {
+    it('should prevent url change if click on nested elements', done => {
```
XhmikosR, line 377: Are you sure this `describe` is the right place?

joke2k: I think it is a bug of the data-api as you can see from the part of collapse.js file:
```

/**
 * Data Api implementation
 */

EventHandler.on(document, EVENT_CLICK_DATA_API, SELECTOR_DATA_TOGGLE, function (event) {
  // preventDefault only for <a> elements (which change the URL) not inside the collapsible element
  if (event.target.tagName === 'A' || (event.delegateTarget && event.delegateTarget.tagName === 'A')) {
    event.preventDefault()
  }
```

joke2k: Is It right?

rohit2sharma95 (approved):

PR #32439, opened 2020-12-11
Decouple BackDrop from modal

GeoSot:
Based on a comment was post on offcanvas merge request, I did a first approach to decouple Backdrop from Modal, for further usage and to simplify Modal code.

It misses the extra tests and a class modification, but they can be done, if you like it

Preview: https://deploy-preview-32439--twbs-bootstrap.netlify.app/docs/5.0/components/modal/

TODO:

- [x] Revert the class name change since it's a breaking one
- [x] Revert any further breaking changes like variable names, new backdrop.scss file etc


js/src/modal.js
```diff
@@ -318,81 +326,33 @@ class Modal extends BaseComponent {
...
-      if (isAnimated) {
-        const backdropTransitionDuration = getTransitionDurationFromElement(this._backdrop)
-        EventHandler.one(this._backdrop, 'transitionend', callbackRemove)
-        emulateTransitionEnd(this._backdrop, backdropTransitionDuration)
+      if (this._config.backdrop === true) {
+        this.hide()
       } else {
-        callbackRemove()
+        this._triggerBackdropTransition()
       }
```
rohit2sharma95, line 352: It will now trigger `_triggerBackdropTransition` even when the backdrop is set to false.

js/src/util/backdrop.js
```diff
@@ -0,0 +1,135 @@
...
+    this._isAppended = true
+  }
+
+  dispose() {
+    EventHandler.off(this._get(), EVENT_MOUSEDOWN)
+    if (!this._isAppended) {
+      return
+    }
+
+    this._get().parentNode.removeChild(this._get())
```
rohit2sharma95, line 120: `dispose` method is supposed to be called only when the backdrop instance exists. So can not you use `this._element` directly here :thinking:

js/src/util/backdrop.js
```diff
@@ -0,0 +1,135 @@
...
+
+    this._get().classList.add(CLASS_NAME_SHOW)
+
+    this._emulateAnimation(() => {
+      execute(callback)
+    })
+  }
+
+  hide(callback) {
+    EventHandler.off(this._get(), EVENT_MOUSEDOWN)
```
rohit2sharma95, line 64: Event listener should be removed only when the backdrop is visible (exists).

js/src/util/backdrop.js
```diff
@@ -0,0 +1,135 @@
...
+
+    this._emulateAnimation(() => {
+      this.dispose()
+      execute(callback)
+    })
+  }
+
+  // Private
+
+  _get() {
```
rohit2sharma95, line 81: Better `_get` => `_getElement` :thinking:

js/src/util/backdrop.js
```diff
@@ -0,0 +1,135 @@
...
+
+    return this._element
+  }
+
+  _getConfig(config) {
+    config = {
+      ...Default,
+      ...(typeof config === 'object' ? config : {})
+    }
+    typeCheckConfig(NAME, config, DefaultType)
```
rohit2sharma95, line 96: The user can not pass the configuration directly to the backdrop, still, need to check the type?

GeoSot: Only for consistency and bulletproof future. Your call

js/src/modal.js
GeoSot: I had the same thought. 
However:
image

rohit2sharma95: I mean you should update the check to make it specific (May be as below)
```js
if (this._config.backdrop === true) {
  this.hide()
} else if (this._config.backdrop === 'static') {
  this._triggerBackdropTransition()
}
```

js/src/util/backdrop.js
rohit2sharma95: Since it is not a standalone component, type checking can be omitted IMO.

/CC @XhmikosR to make the decision

js/src/modal.js
```diff
@@ -83,7 +83,7 @@ class Modal extends BaseComponent {
 
     this._config = this._getConfig(config)
     this._dialog = SelectorEngine.findOne(SELECTOR_DIALOG, this._element)
-    this._backdrop = null
+    this._backdrop = this._initializeBackDrop()
```
alpadev, line 86: Was there a particular reason why you chose to use a method instead of just doing `new Backdrop()`?

js/src/util/index.js
```diff
@@ -230,6 +230,12 @@ const defineJQueryPlugin = (name, plugin) => {
   })
 }
 
+const execute = callback => {
+  if (callback && typeof callback === 'function') {
```
alpadev, line 234: I think you can omit `callback &&` here.

alpadev (commented):
Just some minor things :+1:

js/src/modal.js
GeoSot: Only to separate different things, and less busy code for eyes. (maybe o future more lazy initialization?)

alpadev (approved):

js/src/util/backdrop.js
```diff
@@ -0,0 +1,135 @@
...
+
+class Backdrop {
+  constructor(config) {
+    this._config = this._getConfig(config)
+    this._isAppended = false
+    this._element = null
+  }
+
+  onClick(callback) {
+    this._clickCallback = callback
```
GeoSot, line 37: @rohit2sharma95 , @alpadev is better to keep `onClick` as method, or config property?

rohit2sharma95 (approved):

js/src/modal.js
```diff
@@ -313,81 +321,35 @@ class Modal extends BaseComponent {
...
+        this.hide()
       }
 
-      if (isAnimated) {
-        const backdropTransitionDuration = getTransitionDurationFromElement(this._backdrop)
-        EventHandler.one(this._backdrop, 'transitionend', callbackRemove)
-        emulateTransitionEnd(this._backdrop, backdropTransitionDuration)
-      } else {
-        callbackRemove()
+      if (this._config.backdrop === 'static') {
```
rohit2sharma95, line 347: You are not using `else` here. (Now both conditions will be checked, that's not needed)

PR #32440, opened 2020-12-11
Expand `visually-hidden-focusable` so it can be used on a container, …

### LINKED ISSUE #32370: .visually-hidden-focusable should also apply to :focus-within
Currently the `visually-hidden-focusable` class only shows elements which are directly focused.
It is however rather common to have a single hidden element which contains two or more anchors to skip to content, navigation, footer etc.

As it is right now this is rather cumbersome and requires manual styling/classes.
Changing the class to also show elements when the `focus-within` pseudo-class applies should not break things and simplifies this a lot.

The only cases where this would break stuff is if one was using nested focusable elements (e.g. a link inside a button - which I don't even know if something like that is allowed) **and** depends on the behaviour that focusing the element would hide and therefore unfocus both of them.

patrickhlauke:
…so the container becomes visible when focus is inside it / on one of its child elements.

Closes https://github.com/twbs/bootstrap/issues/32370

https://deploy-preview-32440--twbs-bootstrap.netlify.app/docs/5.0/helpers/visually-hidden/

### DIFF
--- a/site/content/docs/5.0/helpers/visually-hidden.md
+++ b/site/content/docs/5.0/helpers/visually-hidden.md
@@ -6,12 +6,15 @@ group: helpers
 aliases: "/docs/5.0/helpers/screen-readers/"
 ---
 
-Visually hide an element while still allowing it to be exposed to assistive technologies (such as screen readers) with `.visually-hidden`. Use `.visually-hidden-focusable` to visually hide an element by default, but to display it when it's focused (e.g. by a keyboard-only user). Can also be used as mixins.
+Visually hide an element while still allowing it to be exposed to assistive technologies (such as screen readers) with `.visually-hidden`. Use `.visually-hidden-focusable` to visually hide an element by default, but to display it when it's focused (e.g. by a keyboard-only user). `.visually-hidden-focusable` can also be applied to a container–thanks to `:focus-within`, the container will be displayed when any child element of the container receives focus.
 
 {{< example >}}
 <h2 class="visually-hidden">Title for screen readers</h2>
 <a class="visually-hidden-focusable" href="#content">Skip to main content</a>
+<div class="visually-hidden-focusable">A container with a <a href="#">focusable element</a>.</div>
 {{< /example >}}
+
+Both `visually-hidden` and `visually-hidden-focusable` can also be used as mixins.
 
 ```scss
 // Usage as a mixin
--- a/scss/helpers/_visually-hidden.scss
+++ b/scss/helpers/_visually-hidden.scss
@@ -3,6 +3,6 @@
 //
 
 .visually-hidden,
-.visually-hidden-focusable:not(:focus) {
+.visually-hidden-focusable:not(:focus):not(:focus-within) {
   @include visually-hidden();
 }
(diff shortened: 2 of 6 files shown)


site/content/docs/5.0/helpers/visually-hidden.md
```diff
@@ -6,13 +6,16 @@ group: helpers
 aliases: "/docs/5.0/helpers/screen-readers/"
 ---
 
-Visually hide an element while still allowing it to be exposed to assistive technologies (such as screen readers) with `.visually-hidden`. Use `.visually-hidden-focusable` to visually hide an element by default, but to display it when it's focused (e.g. by a keyboard-only user). Can also be used as mixins.
+Visually hide an element while still allowing it to be exposed to assistive technologies (such as screen readers) with `.visually-hidden`. Use `.visually-hidden-focusable` to visually hide an element by default, but to display it when it's focused (e.g. by a keyboard-only user). `.visually-hidden-focusable` can also be applied to a container–thank to the use of `:focus-within`, the container will be displayed when any child element of the container receives focus.
```
septatrix, line 9:
```suggestion
Visually hide an element while still allowing it to be exposed to assistive technologies (such as screen readers) with `.visually-hidden`. Use `.visually-hidden-focusable` to visually hide an element by default, but to display it when it's focused (e.g. by a keyboard-only user). `.visually-hidden-focusable` can also be applied to a container – thanks to the use of `:focus-within`, the container will be displayed when any child element of the container receives focus.
```

I also noted that you used an em-dash which is correct in this context though I am not sure if you actually use these regularly in the documentation.

septatrix (commented):
Seems good. Found a little typo ;D

site/content/docs/5.0/helpers/visually-hidden.md
patrickhlauke: it's our style not to use spaces before/after em dashes (which is grammatically correct, but just feels very antiquated). I'm personally not a fan, but ... that's what @mdo likes ;)

septatrix: As long as it is consistent I am fine with it. _thank_ → _thanks_ still holds I think (or at least sound weird however I am not a native speaker).

patrickhlauke: ah, missed that bit from your suggestion

ffoodd (approved):
Very nice :+1:

ffoodd:
Oh and also: would you mind moving our docs' `.skippy` to this pattern? I implemented this quite roughly in `site/assets/scss/_skippy.scss`:

```scss
.skippy {
  &:focus-within a {
    position: static !important;
    width: auto !important;
    height: auto !important;
    padding: $spacer / 2 !important;
    margin: $spacer / 4 !important;
    overflow: visible !important;
    clip: auto !important;
    white-space: normal !important;
  }
}
```
but we could drop this part and simply use `.visually-hidden-focusable` alongside `.skippy`.

ffoodd:
@patrickhlauke just used this new feature on our own skip links, please have a look :)

patrickhlauke:
> @patrickhlauke just used this new feature on our own skip links, please have a look :)

looking good

mdo (approved):

PR #32443, opened 2020-12-12
Restore offset option for dropdown component

### LINKED ISSUE #32418: Dropdown offset option is ignored v5
With v5 the dropdown feature doesn't work with offset property.
Maybe related to #32354

It doesn't work also in the documentation:
https://getbootstrap.com/docs/5.0/components/dropdowns/#dropdown-options

The v5 version of dropdown doesn't collect the offset:
https://github.com/twbs/bootstrap/blob/1edec546cb28baf68691f0eb9e43997901359ba5/js/src/dropdown.js#L292-L308

The v4 yes:
https://github.com/twbs/bootstrap/blob/1edec546cb28baf68691f0eb9e43997901359ba5/js/src/dropdown.js#L292-L308

joke2k:
With Popper 2.0 the dropdown offset option was removed from bootstrap but remains in the documentation and in our :heart: s.
This is an humble PR to restore them.

Fixes #32418

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -72,7 +72,7 @@ const PLACEMENT_RIGHT = isRTL ? 'left-start' : 'right-start'
 const PLACEMENT_LEFT = isRTL ? 'right-start' : 'left-start'
 
 const Default = {
-  offset: 0,
+  offset: [0, 0],
   flip: true,
   boundary: 'clippingParents',
   reference: 'toggle',
@@ -81,7 +81,7 @@ const Default = {
 }
 
 const DefaultType = {
-  offset: '(number|string|function)',
+  offset: '(array|string|function)',
   flip: 'boolean',
   boundary: '(string|element)',
   reference: '(string|element|object)',
@@ -298,6 +298,20 @@ class Dropdown extends BaseComponent {
     return this._element.closest(`.${CLASS_NAME_NAVBAR}`) !== null
   }
 
+  _getOffset() {
+    const { offset } = this._config
+
+    if (typeof offset === 'string') {
+      return offset.split(',').map(val => Number.parseInt(val, 10))
+    }
+
+    if (typeof offset === 'function') {
+      return popperData => offset(popperData, this._element)
+    }
+
+    return offset
+  }
+
   _getPopperConfig() {
     const popperConfig = {
       placement: this._getPlacement(),
@@ -312,6 +326,12 @@ class Dropdown extends BaseComponent {
         name: 'flip',
         options: {
           fallbackPlacements: ['top', 'right', 'bottom', 'left']
+        }
+      },
+      {
+        name: 'offset',
+        options: {
+          offset: this._getOffset()
         }
       }]
     }
--- a/js/tests/unit/dropdown.spec.js
+++ b/js/tests/unit/dropdown.spec.js
@@ -54,6 +54,54 @@ describe('Dropdown', () => {
       expect(dropdown.toggle).toHaveBeenCalled()
     })
 
+    it('should create offset modifier correctly when offset option is a function', done => {
+      fixtureEl.innerHTML = [
+        '<div class="dropdown">',
+        '  <button class="btn dropdown-toggle" data-bs-toggle="dropdown">Dropdown</button>',
+        '  <div class="dropdown-menu">',
+        '    <a class="dropdown-item" href="#">Secondary link</a>',
+        '  </div>',
+        '</div>'
+      ].join('')
+
+      const getOffset = jasmine.createSpy('getOffset').and.returnValue([10, 20])
+      const btnDropdown = fixtureEl.querySelector('[data-bs-toggle="dropdown"]')
+      const dropdown = new Dropdown(btnDropdown, {
+        offset: getOffset,
+        popperConfig: {
+          onFirstUpdate: state => {
+            expect(getOffset).toHaveBeenCalledWith({
+              popper: state.rects.popper,
+              reference: state.rects.reference,
+              placement: state.placement
+            }, btnDropdown)
+            done()
+          }
+        }
+      })
+      const offset = dropdown._getOffset()
+
+      expect(typeof offset).toEqual('function')
+
+      dropdown.show()
+    })
+
+    it('should create offset modifier correctly when offset option is a string into data attribute', () => {
+      fixtureEl.innerHTML = [
+        '<div class="dropdown">',
+        '  <button class="btn dropdown-toggle" data-bs-toggle="dropdown" data-bs-offset="10,20">Dropdown</button>',
+        '  <div class="dropdown-menu">',
+        '    <a class="dropdown-item" href="#">Secondary link</a>',
+        '  </div>',
+        '</div>'
+      ].join('')
+
+      const btnDropdown = fixtureEl.querySelector('[data-bs-toggle="dropdown"]')
+      const dropdown = new Dropdown(btnDropdown)
+
+      expect(dropdown._getOffset()).toEqual([10, 20])
+    })
+
     it('should allow to pass config to Popper with `popperConfig`', () => {
       fixtureEl.innerHTML = [
         '<div class="dropdown">',
--- a/site/content/docs/5.0/components/dropdowns.md
+++ b/site/content/docs/5.0/components/dropdowns.md
@@ -901,9 +901,9 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td>array | string | function</td>
       <td><code>[0, 0]</code></td>
       <td>
-        <p>Offset of the dropdown relative to its target.</p>
-        <p>When a function is used to determine the offset, it is called with an object containing the <code>popper</code> instance, the <code>refecence</code> Element and the <code>placement</code> as its first argument. The function must return an array with two numbers: <code>[<a href="https://popper.js.org/docs/v2/modifiers/offset/#skidding-1">skidding</a>, <a href="https://popper.js.org/docs/v2/modifiers/offset/#distance-1">distance</a>]</code>. The triggering element DOM node is passed as the second argument.</p>
-        <p>For more information refer to Popper.js's <a href="https://popper.js.org/docs/v2/modifiers/offset/#options">offset docs</a>.</p>
+        <p>Offset of the dropdown relative to its target. You can pass a string in data attributes with comma separated values like: <code>data-bs-offset="10,20"</code></p>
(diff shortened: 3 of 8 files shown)


XhmikosR:
My problem is that I have no idea why it was removed in the first place, and if there was some special reason...

joke2k:
> My problem is that I have no idea why it was removed in the first place, and if there was some special reason...

It think because popper changes the way to configure the offset.
I used the old tests to check. But is not fully-coverage yet

js/src/dropdown.js
```diff
@@ -289,6 +289,31 @@ class Dropdown extends BaseComponent {
...
+      return [0, 0]
+    }
+
+    if (typeof offset === 'number') {
+      offset = [offset, 0]
+    } else if (typeof offset === 'string') {
+      offset = offset.split(',')
+      if (offset.length === 1) {
+        offset = [offset[0], 0]
+      }
```
rohit2sharma95, line 305: IMO, now configuration should be updated as per the Popper. So it is better to accept `OffsetsFunction | [?number, ?number]`. But you can keep the string type so that offset can be passed from data attributes.

And as per your check:
```js
if (offset.length === 1) {
  offset = [offset[0], 0]
}
``` 
You are allowing and set only **Skidding** , because `[number, number]` => [[Skidding](https://popper.js.org/docs/v2/modifiers/offset/#skidding-1), [Distance](https://popper.js.org/docs/v2/modifiers/offset/#distance-1)]. It may or may not be the desired behavior. So you should mention it in the docs.

js/src/dropdown.js
```diff
@@ -289,6 +289,31 @@ class Dropdown extends BaseComponent {
...
+
+      offset = offset.map(val => Number.parseInt(val, 10))
+    } else if (typeof offset === 'function') {
+      offset = (popper, reference, placement) => {
+        return offset({ popper, reference, placement }, this._element)
+      }
```
rohit2sharma95, line 311: And Popper passes a single object in the offset function: https://popper.js.org/docs/v2/modifiers/offset/#options
You should use destructuring `({ placement, reference, popper })`

rohit2sharma95 (commented):
Thanks for the PR :slightly_smiling_face: @joke2k
I have left a few comments, it is better to have the `offset` option in the configuration.

js/src/dropdown.js
joke2k: You're right, the configuration could be too much "smarty".
The offset option should be a shortcut.
So using javascript is better `OffsetsFunction | [?number, ?number]` as you said, 
with data attribute is better to use the `"Number,Number"` format.
If anybody needs more can use popperConfig :D

js/src/dropdown.js
joke2k: To be honest, my intention was to pass one object as the first parameter (like was for offsets in bs4)
so it is my mistake on reading Popper docs.

rohit2sharma95:
@joke2k Can you make the update, please?

js/src/dropdown.js
```diff
@@ -72,7 +72,7 @@ const PLACEMENT_RIGHT = isRTL ? 'left-start' : 'right-start'
 const PLACEMENT_LEFT = isRTL ? 'right-start' : 'left-start'
 
 const Default = {
-  offset: 0,
+  offset: [0, 0],
```
joke2k, line 75: New default

js/src/dropdown.js
```diff
@@ -81,7 +81,7 @@ const Default = {
 }
 
 const DefaultType = {
-  offset: '(number|string|function)',
+  offset: '(array|string|function)',
```
joke2k, line 84: New DefaultType

js/src/dropdown.js
```diff
@@ -289,6 +289,24 @@ class Dropdown extends BaseComponent {
...
+
+    if (!offset) {
+      return [0, 0]
+    }
+
+    if (typeof offset === 'string') {
+      offset = offset.split(',').map(val => Number.parseInt(val, 10))
+    } else if (typeof offset === 'function') {
+      offset = ({ popper, reference, placement }) => {
+        return offset({ popper, reference, placement }, this._element)
```
rohit2sharma95, line 303: Destructuring here might be ignored because destructuring is not being used in real, thanks for considering previous comment :grin:

---
Popper may call the `offset` function multiple times, and because the `offset` function is in the closure I am getting the error: **Maximum call stack size exceeded** (You can check by creating dist files locally and passing a function to the offset in dropdowns configuration)

My suggestion:
```js
if (typeof offset === 'string') {
  return offset.split(',').map(val => Number.parseInt(val, 10))
}

if (typeof offset === 'function') {
  return popperData => {
    return offset(popperData, this._element)
  }
}
```

js/tests/unit/dropdown.spec.js
```diff
@@ -54,6 +54,42 @@ describe('Dropdown', () => {
...
+      fixtureEl.innerHTML = [
+        '<div class="dropdown">',
+        '  <button class="btn dropdown-toggle" data-bs-toggle="dropdown">Dropdown</button>',
+        '  <div class="dropdown-menu">',
+        '    <a class="dropdown-item" href="#">Secondary link</a>',
+        '  </div>',
+        '</div>'
+      ].join('')
+
+      const getOffset = () => [10, 20]
```
rohit2sharma95, line 67: It is better if you create a spy and  then expect that to have been called after showing the dropdown :slightly_smiling_face:

joke2k: yeah, I tried but I'm not confident with javascript mocking yet.
I'm working on it

js/src/dropdown.js
joke2k: I found this "strange" behaviour in js:
```
let x = () => 1
x() // print 1
x = () => x()
x() // infinite recursion
```
so I've to rewrite the variable name in this script.

js/tests/unit/dropdown.spec.js
joke2k: The funny thing is that the `shown.bs.dropdown` is not related to Popper, so in order to test that a getOffset spy function has been called showing the dropdown I've to use the `popperConfig.onFirstUpdate` attribute.

site/content/docs/5.0/components/dropdowns.md
```diff
@@ -896,6 +896,16 @@ Options can be passed via data attributes or JavaScript. For data attributes, ap
       <td><code>'dynamic'</code></td>
       <td>By default, we use Popper for dynamic positioning. Disable this with <code>static</code>.</td>
     </tr>
+    <tr>
+      <td><code>offset</code></td>
+      <td>number | string | function</td>
+      <td><code>"0,0"</code></td>
```
rohit2sharma95, line 902: Default value for offset is array @joke2k :slightly_smiling_face: 
```js
offset: [0, 0],
```

joke2k: I changed a little bit the copy

rohit2sharma95:
It is now complete for dropdown though but the same should be done for Tooltip/Popover also, IMO.
Ping @XhmikosR :slightly_smiling_face:

rohit2sharma95 (approved):

javieraviles:
If I'm not wrong, scrollspy offset is under the same use-case. In alpha v5 is not working either.

joke2k:
> If I'm not wrong, scrollspy offset is under the same use-case. In alpha v5 is not working either.

I checked into bs4 and bs5 but ScrollSpy component never used Popper for its offset.
In beta1 Scrollspy offset option works pretty well

https://jsfiddle.net/yhv62bz1/

js/src/dropdown.js
```diff
@@ -289,6 +289,20 @@ class Dropdown extends BaseComponent {
     return this._element.closest(`.${CLASS_NAME_NAVBAR}`) !== null
   }
 
+  _getOffset() {
```
XhmikosR, line 292: @rohit2sharma95 this seems to be duplicated. Might be worth checking at a later time if we it makes sense to deduplicate this. Lower priority right now :)

rohit2sharma95: Yes, should tackle it in a new PR later?

javieraviles:
> > If I'm not wrong, scrollspy offset is under the same use-case. In alpha v5 is not working either.
> 
> I checked into bs4 and bs5 but ScrollSpy component never used Popper for its offset.
> In beta1 Scrollspy offset option works pretty well
> 
> https://jsfiddle.net/yhv62bz1/

I thought the offset would not only be applied when scrolling manually to detect which button should be highlighted but also when clicking on the button to have such offset applied in the scroll. My bad

js/src/dropdown.js
joke2k: I'm trying to deduplicate this into `utils/index.js` and I was wondering if we can do the same for `_getPopperConfig`...
but the function signature becoming a little bit weird:
```js
const getPopperOffsets = (element, config) => {
  if (typeof offset === 'string') {
    return offset.split(',').map(val => Number.parseInt(val, 10))
  }

  if (typeof offset === 'function') {
    return popperData => offset(popperData, element)
  }

  return offset
}

const getPopperConfig = (element, placement, offset, modifiers, options) => {
  const defaultBsConfig = {
    placement: placement,
    modifiers: [...(modifiers || [])],
    ...(options || {})
  }

  if (offset) {
    defaultBsConfig.modifiers.push({
      name: 'offset',
      options: {
        offset: getPopperOffsets(element, offset)
      }
    })
  }

  return defaultBsConfig
}
```
the `element` is necessary because offset configured as function needs it for the second parameter.

XhmikosR: No need to do this now. We'll have plenty of time after this lands :)

rohit2sharma95:
I also notice if someone wants to pass the `offset` option (Or any other modifier for Popper) via the `popperConfig` option, **the other modifiers are replaced by the array of the modifiers passed in the popperConfig.**

So if the user passes only one modifier in the array of `popperConfig`, other modifiers are reset to the defaults used by Popper.
https://github.com/twbs/bootstrap/blob/86802f701cdff9a2fe91278fe42684a32426bc61/js/src/dropdown.js#L318-L322

So few options should be maintained individually by Bootstrap that are customized by bootstrap. For example, the default value for `fallbackPlacements` is changed by Bootstrap. And the `offset` option is one of them.

And it should be mentioned in the docs that if someone passes the configuration via `popperConfig`, it is kind of handling the Popper's configuration natively because it can override some of the default configurations by bootstrap.

/CC @XhmikosR @MartijnCuppens

mdo:
@rohit2sharma95 @XhmikosR Should we open an issue or PR to track that separately so we can merge this?

rohit2sharma95:
Yes, either the docs should be updated or the way of merging the popperConfig in the plugins.

PR #32453, opened 2020-12-13
Few updates for v5.0.0-beta1 NuGet support

supergibbs:
PopperJS 2.0 isn't on NuGet so dropped as a dependency and added a note to install independently

### DIFF
--- a/nuget/bootstrap.nuspec
+++ b/nuget/bootstrap.nuspec
@@ -3,22 +3,19 @@
   <metadata>
     <id>bootstrap</id>
     <!-- pulled from package.json -->
-    <version>4</version>
+    <version>5</version>
     <title>Bootstrap CSS</title>
     <authors>The Bootstrap Authors, Twitter Inc.</authors>
     <owners>bootstrap</owners>
     <description>The most popular front-end framework for developing responsive, mobile first projects on the web.</description>
     <releaseNotes>https://blog.getbootstrap.com/</releaseNotes>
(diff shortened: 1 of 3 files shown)


nuget/bootstrap.nuspec
```diff
@@ -3,22 +3,19 @@
...
     <!-- pulled from package.json -->
-    <version>4</version>
+    <version>5</version>
     <title>Bootstrap CSS</title>
     <authors>The Bootstrap Authors, Twitter Inc.</authors>
     <owners>bootstrap</owners>
     <description>The most popular front-end framework for developing responsive, mobile first projects on the web.</description>
     <releaseNotes>https://blog.getbootstrap.com/</releaseNotes>
-    <summary>Bootstrap framework in CSS. Includes JavaScript</summary>
+    <summary>Bootstrap framework in CSS. Includes JavaScript but requires PopperJS which must be installed independently</summary>
```
XhmikosR, line 12: Can't the users use our bundle file? That way they shouldn't need to install popper manually.

supergibbs:
True, I forgot about the bundle. Docs cover this pretty well so just reverting the note.

nuget/bootstrap.nuspec
supergibbs: True, I forgot about the bundle. Docs cover this pretty well so just reverting the note.

XhmikosR (approved):

PR #32467, opened 2020-12-14
Release v5.0.0-beta2

XhmikosR:
Preview: <https://deploy-preview-32467--twbs-bootstrap.netlify.app/>


XhmikosR:
@ffoodd can you please have a quick look at the dist CSS changes so that we make sure everything's alright?

ffoodd:
Seems good to me, I recognize every change with a single exception than I'm unable to trace back: some `.fs-*-*` utilities are removed, but I can't even find where they were coming from :thinking: 

@MartijnCuppens Might it relate to RFS?

XhmikosR:
Yeah, that must be due to #32397

ffoodd:
Indeed, missed that one— thanks :)

mdo (approved):

PR #32473, opened 2020-12-14
Default text colour of '.disabled' class could be lighter

emfluenceindia:
Hi,

Can we make the default text colour of Disabled Link (**Dropdown and Tab items**)  bit more lighter?
Currently this is **pretty close** to that of Active item's text colour.

My suggestion is instead of `$gray-600` (`#6c757d`), we may use `$gray-400` (`#ced4da`) for `.disabled` class. (`src/scss/_variables.scss`).

Please see the difference in the following screenshots

**Dropdown**

**Tabs**

Also, adding the  `.disabled` class to links might also use a different colour, but when the click is enabled (but the default behaviour is prevented), we could use the same blue colour.
However, if the click is not enabled, we might use a lighter colour as well telling that link is available but not clickable. May be we can use the same `$gray-400` for this as well!

An use case maybe a **role based restrictions**.

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -5,7 +5,7 @@
 
 // Color system
 
-$white:    #fff !default;
+$white: #fff !default;
 $gray-100: #f8f9fa !default;
 $gray-200: #e9ecef !default;
 $gray-300: #dee2e6 !default;
@@ -15,7 +15,7 @@ $gray-600: #6c757d !default;
 $gray-700: #495057 !default;
 $gray-800: #343a40 !default;
 $gray-900: #212529 !default;
-$black:    #000 !default;
+$black: #000 !default;
 
 // fusv-disable
 $grays: (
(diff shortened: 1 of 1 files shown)


MartijnCuppens:
Please revert all indention changes

mdo:
Nope, fails color contrast. https://colorable.jxnblk.com/ced4da/ffffff

patrickhlauke:
> Nope, fails color contrast. https://colorable.jxnblk.com/ced4da/ffffff

regarding contrast: WCAG exempts inactive controls https://github.com/twbs/bootstrap/issues/32474#issuecomment-744491685

patrickhlauke:
actually, as I can see that this PR is a giant mess of indentation changes that shouldn't be there...I'll close it. but the concept (making disabled controls a bit lighter) is sound in my view. but make it `$gray-500` as discussed here https://github.com/twbs/bootstrap/issues/32474

emfluenceindia:
> > Nope, fails color contrast. https://colorable.jxnblk.com/ced4da/ffffff
> 
> regarding contrast: WCAG exempts inactive controls [#32474 (comment)](https://github.com/twbs/bootstrap/issues/32474#issuecomment-744491685)

Color Contrast for $gray-500 is 2:1. Could we use this instead? Actually I am on my way home right now. May I resume tomorrow please?

emfluenceindia:
> actually, as I can see that this PR is a giant mess of indentation changes that shouldn't be there...I'll close it. but the concept (making disabled controls a bit lighter) is sound in my view. but make it `$gray-500` as discussed here #32474

Thank you so much for your accepting my suggestion and correcting me. May I ask one thing please? What kind of indentation issue is there? I used VSCODE. And the only change I made was updated the value from `$gray-600` to `$gray-400`. Do I need to use a different editor for this? Please let me know so that I can rectify my errors.

PR #32246, opened 2020-11-24
Replace Lorem Ipsum placeholder text with more representative (or at least english language) text

patrickhlauke:

PR #32479, opened 2020-12-14
Clean up lorem ipsum text in popover examples

florianlacreuse:
Following #32246 and #32463.


patrickhlauke (approved):
Good spot. And sigh, so many little cases of those critters in various places...I'll be glad when we're finally rid of them 👍

PR #32489, opened 2020-12-15
Fix popover docs example

florianlacreuse:
Use `{{< example >}}` to avoid duplicate code (and fix spacing between buttons on wrapping).

https://deploy-preview-32489--twbs-bootstrap.netlify.app/docs/5.0/components/popovers/#four-directions


ffoodd (approved):
Thanks!

PR #32501, opened 2020-12-16
.ratio missing variable prefix

ffoodd:
This is the only unprefixed custom property, and its name is very common so I think we'd better prefix it too.
Any thought? If this is approved, it'll need some updates in the docs and the migration guide.

Also, shouldn't we use a default, fallback value in the `padding-top` property?


MartijnCuppens:
Should be adjusted here too: https://getbootstrap.com/docs/5.0/helpers/ratio/#custom-ratios

> Also, shouldn't we use a default, fallback value in the `padding-top` property?


Idk, we also have a `.btn` class which can't stand on its own. Imo a `.ratio` class on its own doesn't really explain what ratio to expect, does it?

ffoodd:
Thanks for spotting docs!

I thought about a 1×1 ratio, meaning a square as a default. But not sure it makes any sense, si let's drop the idea for now.

I'll update docs tomorrow 👌

ffoodd:
Docs and migration guide updated :rocket:

XhmikosR:
There's still a conflict :P

MartijnCuppens (approved):

mdo (approved):

PR #32510, opened 2020-12-16
Reset styles when popper is calculating position

### LINKED ISSUE #32484: Dropdown menu not positioned correctly with .dropstart
When using `.dropstart`, the dropdown menu will cover most of the toggle button.  

Seems like this is caused by this style:
https://github.com/twbs/bootstrap/blob/042a4a255f3c994ee8285df9a14fe0fbd95df83c/scss/_dropdown.scss#L36-L39

If I disable this style, the menu is positioned correctly.

This can be reproduced on the docs page:
https://getbootstrap.com/docs/5.0/components/dropdowns/#dropleft

- Operating system and version: Windows
- Browser and version: All

MartijnCuppens:
Farewell style hack 👋

Fixes https://github.com/twbs/bootstrap/issues/32484

https://deploy-preview-32510--twbs-bootstrap.netlify.app/docs/5.0/components/dropdowns/#dropleft

### DIFF
--- a/js/src/dropdown.js
+++ b/js/src/dropdown.js
@@ -57,6 +57,7 @@ const CLASS_NAME_DROPUP = 'dropup'
 const CLASS_NAME_DROPEND = 'dropend'
 const CLASS_NAME_DROPSTART = 'dropstart'
 const CLASS_NAME_NAVBAR = 'navbar'
+const CLASS_NAME_POPPER_ACTIVE = 'popper-active'
 
 const SELECTOR_DATA_TOGGLE = '[data-bs-toggle="dropdown"]'
 const SELECTOR_FORM_CHILD = '.dropdown form'
@@ -176,7 +177,12 @@ class Dropdown extends BaseComponent {
         referenceElement = this._config.reference
       }
 
+      // The `.popper-active` class removes the `right` property in CSS
+      // This way popper can position the dropdown correctly
+      // We add it back after the calculations to get control over our positioning with our CSS again
+      this._menu.classList.add(CLASS_NAME_POPPER_ACTIVE)
       this._popper = Popper.createPopper(referenceElement, this._menu, this._getPopperConfig())
+      this._menu.classList.remove(CLASS_NAME_POPPER_ACTIVE)
     }
 
     // If this is a touch-enabled device we add extra
--- a/scss/_dropdown.scss
+++ b/scss/_dropdown.scss
@@ -34,7 +34,7 @@
   @include box-shadow($dropdown-box-shadow);
 
   // Reset positioning when positioned with Popper
-  &[style] {
+  &.popper-active {
     right: auto#{"/* rtl:ignore */"} !important; // stylelint-disable-line declaration-no-important
   }
 }


js/src/dropdown.js
```diff
@@ -174,7 +174,9 @@ class Dropdown extends BaseComponent {
         }
       }
 
+      this._menu.classList.add('popper-active')
```
XhmikosR, line 177: Can you please add a comment here why we add `popper-active` and we remove it after creating the popper?

MartijnCuppens: Thanks for this, we don't need to remove this class indeed. I've added a comment about why we need to add this class.

MartijnCuppens: Lol, ok, nevermind, we do need this to fix the dropleft positioning 😛

ffoodd:
I'm not sure why but the behaviour changed in RTL, `translate()` does not move horizontally anymore: 
- This PR: https://deploy-preview-32510--twbs-bootstrap.netlify.app/docs/5.0/examples/cheatsheet-rtl/#dropdowns
- `main`: https://main--twbs-bootstrap.netlify.app/docs/5.0/examples/cheatsheet-rtl/#dropdowns

I can't see the relation with this PR though, but noticed it. Might be unrelated.

js/src/dropdown.js
```diff
@@ -174,7 +175,12 @@ class Dropdown extends BaseComponent {
         }
       }
 
+      // The `.popper-active` class removes the `right` property in CSS
+      // This way popper can position the dropdown correctly
+      // We add it back after the calculations to get control over our positioning with our CSS again
+      this._menu.classList.add(CLASS_NAME_POPPER_ACTIVE)
       this._popper = Popper.createPopper(referenceElement, this._menu, this._getPopperConfig())
+      this._menu.classList.remove(CLASS_NAME_POPPER_ACTIVE)
```
septatrix, line 183: Why do we remove the class again? When the element if positioned by popper the `right` property will not have an effect anyway or am I wrong?

MartijnCuppens: That's what I would expect too, but it turns out this is needed by popper.

rohit2sharma95:
As per the comment written in the code here:
https://github.com/twbs/bootstrap/blob/122b2bdd485373c05eeab50c9f809db31117b898/scss/_dropdown.scss#L36-L40

This right style was added to reset the position when the dropdown is positioned with Popper. (This was not in V4 where Popper V1 is used)
I tried to create a raw dropdown only with native Popper. And found that I don't need to reset the position when the element is placed by Popper.

As per my research, if you [remove the right style](https://github.com/rohit2sharma95/bootstrap/pull/14) from the dropdown, it works fine (even when Popper is not used): https://bootstrap-1jtqi8bem.vercel.app/docs/5.0/components/dropdowns/

Or maybe I am missing something @twbs/css-review :slightly_smiling_face:

septatrix:
I tried to get rid of all the `top/right/bottom/left` styles which worker rather well except for the statically positioned examples (which do not use poppers dynamic positioning. It should work to only set these properties when `[data-bs-display=static]` matches or the element is inside a navbar (which are treated similar) however I do not have the time now to further experiment with this. Temporarily toggling the styles for popper however seems rather hacky (albeit appearing to work).

rohit2sharma95:
Popper does not require the elements to have positioning styles and it can not place elements correctly that have positioning styles already.
https://codepen.io/rohit2sharma95/pen/wvzyGRM

> It's not really something Popper aims to support, **the requirement is that the target popper element should not be positioned already**.

https://github.com/popperjs/popper-core/discussions/1206#discussioncomment-252521

mdo:
Is this supposed to fix the `dropdown-menu-end` alignment at https://deploy-preview-32510--twbs-bootstrap.netlify.app/docs/5.0/components/dropdowns/#menu-alignment? I still see the broken menu...

rohit2sharma95:
> I still see the broken menu...

Because of the positioning of the dropdown initially.

XhmikosR:
@rohit2sharma95 the issue mdo described above still happens. Was this supposed to be fixed by #32524?

rohit2sharma95:
#32524 was just to remove the margin from the dropdown. The wrong placement of `dropdown-menu-end` is because of the `right` style IMO

mdo:
Opened #32986 as an alternative for now.

mdo:
Closing for #32986.

PR #32511, opened 2020-12-17
modal: move common code to reusable functions

GeoSot:
Combine 2 sets of 3 different code blocks in the name of minification.

It doesn't change functionality, only de-duplicates some code


js/src/modal.js
```diff
@@ -459,64 +459,47 @@ class Modal extends BaseComponent {
...
-        })
+      const setElement = (selector, styleProp, callback) => {
+        SelectorEngine.find(selector)
+          .forEach(element => {
+            const actualValue = element.style[styleProp]
+            const calculatedValue = window.getComputedStyle(element)[styleProp]
+            Manipulator.setDataAttribute(element, styleProp, actualValue)
+            element.style[styleProp] = callback(Number.parseFloat(calculatedValue)) + 'px'
+          })
+      }
```
rohit2sharma95, line 470: It will be a little optimized way if you move this function to a private method. Because it will no be defined again and again.

XhmikosR: Yeah and probably `resetElement()` too :)

js/src/modal.js
```diff
@@ -458,65 +458,48 @@ class Modal extends BaseComponent {
...
-    if (typeof padding === 'undefined') {
-      document.body.style.paddingRight = ''
-    } else {
-      Manipulator.removeDataAttribute(document.body, 'padding-right')
-      document.body.style.paddingRight = padding
-    }
+    this._resetElement('body', 'paddingRight')
+  }
+
+  _resetElement(selector, styleProp) {
```
rohit2sharma95, line 493: Better if you name these methods like: `_resetElementAttributes` and `_setElementAttributes`

js/src/modal.js
```diff
@@ -458,65 +458,48 @@ class Modal extends BaseComponent {
     if (this._isBodyOverflowing) {
       // Note: DOMNode.style.paddingRight returns the actual value or '' if not set
       //   while $(DOMNode).css('padding-right') returns the calculated value or 0 if not set
```
XhmikosR, line 460: BTW this comment seems wrong for v5?

rohit2sharma95 (approved):
I am not sure if comments are still needed. Because now the code is self-explanatory IMO.

Issue #29760, opened 2019-11-28
Docs: add a warning about the Sass implementation


XhmikosR:
See #29743

mdo:
Less of a warning and more of a mention: #32764.

voltaek:
As requested by mdo in #32764, I've updated my PR #32512 to address this issue for the v5 docs.

As this issue is also labeled for v4, I assume that you'd like a PR for similar changes to be made to its doc. In [Getting Started > Downloads](https://getbootstrap.com/docs/4.5/getting-started/download/#source-files), it's mentioned that both LibSass and Ruby Sass are supported. Is that true for v4, or is it just LibSass?

Lastly, in the PR for v4, would you like it mentioned that LibSass is now deprecated, or should the v4 docs stay as more of a "snapshot in time" of when they were originally put together, back when LibSass was still current? I think it may be tricky to state that the recommended Sass compiler is also a deprecated one, but it would be more informative.

PR #32512, opened 2020-12-17
Docs v5: Fix Sass rounding precision and state Sass implementation

voltaek:
The switch from Node Sass to Dart Sass has ended up with all the output CSS having Dart Sass's fixed precision of 10. As mentioned in the current docs and also [here](https://github.com/twbs/bootstrap/issues/29853#issuecomment-692134133), the intended final precision is 6, not 10. As such I've adjusted the parameters of both npm minification scripts to round to a precision of 6.

I also updated the site doc's blurb about precision values used with SASS building to reflect the current precisions used.


XhmikosR (changes requested):
This was intentional, though. Please just keep the doc change to read `10` instead of `6`.

voltaek:
@XhmikosR The only discussion on the topic I could find (linked in previous comment) made it seem like the intention was to still lower the precision to 6, at least for minified CSS. Is the precision to always be left at 10, then? If so I'll adjust the doc to read such. Do you recall the reasoning why we're not utilizing the free reduction in characters, so that I can properly reword the doc explanation?

XhmikosR:
#31728

We din't want to change the default output precision, at least not yet. Please only update the docs.

voltaek:
I've reverted the minification script changes and updated the doc to reflect the current build setup.

site/content/docs/5.0/customize/sass.md
```diff
@@ -6,7 +6,7 @@ group: customize
 toc: true
 ---
 
-Utilize our source Sass files to take advantage of variables, maps, mixins, and more. In our build we've increased the Sass rounding precision to 6 (by default it's 5) to prevent issues with browser rounding.
+Utilize our source Sass files to take advantage of variables, maps, mixins, and more. In our build we've left Dart Sass's rounding precision of 10. We recommend maintaining a precision of at least 6 to prevent issues with browser rounding.
```
XhmikosR, line 9: We might need to reword this @mdo. Dart Sass does not offer the ability to customize the precision anyway.

voltaek: I wanted to keep the mention of a minimum of 6 since it's caused issues in the past, but perhaps we should specify that lowering it can only be done through further processing, such as minification? I was trying to keep it terse.

voltaek: Should it also specify that Dart Sass's precision is locked at 10?

voltaek:
I've expanded the explanation and added a mention of LibSass being deprecated

mdo:
This feels like an odd thing to put here in our customization section. I'm not entirely sure why there's anything there about the precision even now. I don't feel like this is something really worth documenting honestly?

XhmikosR:
@mdo the current sentence is wrong right now:

>In our build we've increased the Sass rounding precision to 6 (by default it's 5) to prevent issues with browser rounding.

mdo:
Yeah I'd just rip all that out. Doesn't make sense to mention it at all tbh.

XhmikosR:
Your call, feel free to edit this PR or make a new one then.

BTW we do have an issue about mentioning the Sass implementation we support. #29760

XhmikosR:
For what is worth, I personally don't mind keeping this PR as is. It explains why people shouldn't lower the precision since we have been historically bit by that.

voltaek:
The Sass implementation the framework is built with and the precision used when building and/or minifying the CSS are both required knowledge to setup a build script when customizing Bootstrap. It would be nice to not have to read through the npm scripts just to gather this knowledge as a user of the framework, and if I were a new user to the framework I think that Docs > Customize > Sass is a reasonable page to navigate to to look for such things.

Perhaps the placement within the page isn't ideal, but I think just giving it a heading of "Building" would help clarify what the paragraphs are about and further help a new user find what they're looking for quickly and where they might expect the information to live.

voltaek:
@XhmikosR Would you like me to create a similar PR for v4 that states the Sass implementation used in order to help close out #29760? If so, should that one also note that Node Sass is deprecated, or just state that its the implementation used and leave it at that (since it wasn't deprecated when v4 was created)?

voltaek:
@XhmikosR @mdo Any more changes you're looking for with this or is the current wording and additional heading sufficient?

XhmikosR:
@voltaek sorry for the delay, it's holidays + quarantine. We'll get to do it soon :)

mdo (approved):

XhmikosR:
I'm going to merge this @voltaek, can you please backport this to v4? Without dart-sass since we use Node sass there.

XhmikosR (approved):

PR #32523, opened 2020-12-18
Tooltip refactoring

### LINKED ISSUE #32372: v5: Bootstrap tooltips don't appear after numbers of focusing
- Operating System: **Windows 10**
- Browser: **Chrome version 87.0.4280.88**

Titled problem occurs even on official docs of bootstrap website at [here](https://getbootstrap.com/docs/5.0/components/tooltips). Tooltips all after some hovering are disappeared and not be accessible any more.

GeoSot:
some minor refactoring,~~plus a possible solution to focusIn glitch (as @XhmikosR [pinpointed ](https://github.com/twbs/bootstrap/issues/32372#issuecomment-740642835))~~

Preview : https://deploy-preview-32523--twbs-bootstrap.netlify.app/docs/5.0/components/tooltips/


js/src/tooltip.js
```diff
@@ -465,6 +448,30 @@ class Tooltip extends BaseComponent {
...
+  }
+
+  _enableIosTouchEventsTrick() {
+    // If this is a touch-enabled device we add extra
+    // empty mouseover listeners to the body's immediate children;
+    // only needed because of broken event delegation on iOS
+    // https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
+    if ('ontouchstart' in document.documentElement) {
+      [].concat(...document.body.children).forEach(element => {
+        EventHandler.on(element, 'mouseover', noop())
```
XhmikosR, line 470: Shouldn't this be just `noop`?

XhmikosR:
@GeoSot thank you very much for digging into this! That being said, can you please split the fix for #32372 to a separate PR?

While testing this patch I notice that indeed the #32372 issue is fixed, but there are cases the arrow momentarily shows up in the wrong direction:

js/src/tooltip.js
GeoSot: sincerely I only moved it, as it was duplicated.  Didn't change it.

XhmikosR: I know :) But I think it should be `noop`. @rohit2sharma95 WDYT?

GeoSot: I cross checked it, the one was 'noop', the other was 'noop()' (wrong). And I was wondering why the script was running with out reason the first time

GeoSot:
I've reverted the possible fix and left the others here

js/src/tooltip.js
```diff
@@ -352,12 +340,7 @@ class Tooltip extends BaseComponent {
 
     tip.classList.remove(CLASS_NAME_SHOW)
 
-    // If this is a touch-enabled device we remove the extra
-    // empty mouseover listeners we added for iOS support
-    if ('ontouchstart' in document.documentElement) {
-      [].concat(...document.body.children)
-        .forEach(element => EventHandler.off(element, 'mouseover', noop))
-    }
```
rohit2sharma95, line 360: Now the mouseover listeners are not removed, because the `_enableIosTouchEventsTrick` method only adds the empty mouseover listeners. :thinking:

js/src/tooltip.js
```diff
@@ -546,7 +553,7 @@ class Tooltip extends BaseComponent {
         EventHandler.on(this._element, eventIn, this.config.selector, event => this._enter(event))
         EventHandler.on(this._element, eventOut, this.config.selector, event => this._leave(event))
       }
-    })
+    }, this)
```
rohit2sharma95, line 556: Is this really required?

rohit2sharma95 (commented):
This PR decreases the size of the un-minified file (-80B) and increases the size of the minified file (+28B).

js/src/tooltip.js
GeoSot: Untitled


if you remove `this` scope, as you can see on the above debug screenshot, it doesn't see `this` and `this.constructor`  is undefined  . It is not supposed to happen but, as I was debugging I fell on it

js/src/tooltip.js
```diff
@@ -569,10 +576,10 @@ class Tooltip extends BaseComponent {
 
   _fixTitle() {
     const title = this._element.getAttribute('title')
-    const originalTitleType = typeof this._element.getAttribute('data-bs-original-title')
+    const originalTitleType = typeof Manipulator.getDataAttribute(this._element, 'original-title')
```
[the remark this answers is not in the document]
GeoSot: I am not sure I have understand correctly your thoughts, but it calls the below 

```js 


  getDataAttribute(element, key) {
    return normalizeData(element.getAttribute(`data-bs-${normalizeDataKey(key)}`))
  },
```

js/src/tooltip.js
GeoSot: If you compare the changed side by side, you 'll find out that I only cut it and used it for a method to avoid duplication, so I don't think I have messed something. 
With a smaaaall exception:

On line 300 was like below:


```js 

      if ('ontouchstart' in document.documentElement) {
        [].concat(...document.body.children).forEach(element => {
          EventHandler.on(element, 'mouseover', noop())
        })
      }

```
'noop' was called directly

rohit2sharma95: If you see line 355-360 in the previous version (event handlers are removed):
```js
// If this is a touch-enabled device we remove the extra
// empty mouseover listeners we added for iOS support
if ('ontouchstart' in document.documentElement) {
  [].concat(...document.body.children)
    .forEach(element => EventHandler.off(element, 'mouseover', noop))
}
```

GeoSot: Oooh sorry. I 've just noticed the difference

js/src/tooltip.js
```diff
@@ -465,6 +448,30 @@ class Tooltip extends BaseComponent {
...
+
+    return context
+  }
+
+  _enableIosTouchEventsTrick() {
+    // If this is a touch-enabled device we add extra
+    // empty mouseover listeners to the body's immediate children;
+    // only needed because of broken event delegation on iOS
+    // https://www.quirksmode.org/blog/archives/2014/02/mouse_event_bub.html
+    if ('ontouchstart' in document.documentElement) {
```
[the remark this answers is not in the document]
patrickhlauke: i'd need to do some investigating of this ... iOS has always been quirky, and apple's documentation has always been very lacking when it comes to their weird heuristics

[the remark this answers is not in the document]
GeoSot: I have reverted this change, but  I can fix it here on on another branch after patrick's feedback

js/src/tooltip.js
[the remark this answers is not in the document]
XhmikosR: Yeah, exactly :) That's not the same as `this._element.getAttribute`. And it seems we already use `Element.getAttribute()` in more places anyway.

js/src/tooltip.js
rohit2sharma95: I am not sure how you are debugging, but this is not required. Because the callback function is an arrow function that gives access to the Tooltip class. :slightly_smiling_face:

js/src/tooltip.js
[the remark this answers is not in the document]
rohit2sharma95: Since the manipulator methods normalize the value that is not required here. Only the string values are coming from the attributes that do not require normalization. :thinking: 

_But keeping Manipulator methods everywhere in the code makes consistency IMO. (That should be done in a separate PR though)_

js/src/tooltip.js
GeoSot: As i wrote before, I agree with you, but as you can see from the previous and the next screenshot the facts disagree. 

Untitled


It would be great, if you give it sometime and try it by yourself. 

If you insist to revert it, I am ok with it

js/src/tooltip.js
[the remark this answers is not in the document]
rohit2sharma95: Then, these changes can be considered in a new PR. :thinking:

js/src/tooltip.js
XhmikosR: This shouldn't be needed since we are using an arrow function thus `this` refers to the upper function.

https://stackoverflow.com/questions/19445599/accessing-this-in-a-foreach-loop-results-in-undefined

Maybe we are missing something else?

rohit2sharma95: I actually made a commit that is now removed (by force push): https://github.com/twbs/bootstrap/commit/2b02d4949a229325ae5f03c78097ba10b9d74b38

GeoSot: Yes, @rohit2sharma95 , we committed the same time, and I missed it. Sorry.

js/src/tooltip.js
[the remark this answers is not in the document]
GeoSot: I have revert them all

rohit2sharma95 (approved):

GeoSot:
@rohit2sharma95 I am glad, this request was approved. But can we ensure the 'this' scope thing? (maybe on another request)
 We all know by theory, it has to be the same scope on arrow functions, but firefox debugger can be lying?

rohit2sharma95:
True, wired behavior on firefox (Not getting the reference while hovering over `this`) :smile: 
But if you see the previous method in the call stack, you will find the context for this (in the dist file). Notice `_this5` in the image below:

image

XhmikosR:
So, basically, the only logic change is that in `toggle()` method, `context` didn't have the extra check:

```diff
-context = Data.getData(event.delegateTarget, dataKey)
+context = context || Data.getData(event.delegateTarget, dataKey)
```

I don't suppose this could cause any issues?

js/src/tooltip.js
```diff
@@ -245,83 +239,85 @@ class Tooltip extends BaseComponent {
...
     }
 
-    if (this.isWithContent() && this._isEnabled) {
-      const showEvent = EventHandler.trigger(this._element, this.constructor.Event.SHOW)
-      const shadowRoot = findShadowRoot(this._element)
-      const isInTheDom = shadowRoot === null ?
-        this._element.ownerDocument.documentElement.contains(this._element) :
-        shadowRoot.contains(this._element)
+    if (!(this.isWithContent() && this._isEnabled)) {
+      return
```
XhmikosR, line 243: This is not covered by the tests, I wonder if we could add a new test?

rohit2sharma95: This check was already there, just the way of implementation is changed:
https://github.com/twbs/bootstrap/blob/56fdcf9e3510a53e73304869fbb444227b1451eb/js/src/tooltip.js#L248

XhmikosR:
@rohit2sharma95 any thoughts about https://github.com/twbs/bootstrap/pull/32523#issuecomment-762228981?

rohit2sharma95:
The second parameter is optional in `_initializeOnDelegatedTarget` method. So still the second expression (`Data.getData(event.delegateTarget, dataKey)`) is being used for the toggle method. LGTM :+1:

PR #32524, opened 2020-12-18
Remove the initial margin from dropdown/popover in favor of Popper

### LINKED ISSUE #32379: Popper emits warning for dropdowns due to non-zero margin styles
Originated in #32354:

> I also noticed that popper shows a warning in the console (at least when referencing the scripts manually and not using the minified bundle). Look inside the console of [this pen](https://codepen.io/Septatrix/pen/dypXKgx?editors=1111) when clicking the button. This is due to `.dropdown-menu` specifying a margin which poppler overwrites anyway. Do you want to track this here or create a new issue?

The simplest fix would be to get rid of the margin as popper set to margin to 0 inline by default. If however the styling is important the margin/offset has to be passed as a config option to the popper constructor.

rohit2sharma95:
Set the default margin to 0 to override the `ul`'s margin. And add the required space (distance) in Popper's offset.

For the static dropdowns add the margin using CSS after it is shown (To make sure if it was using Popper, `data-popper-placement` has been added)

Fixes #32379


rohit2sharma95:
I am not sure about the CSS changes, maybe these changes can be optimized a little bit. :slightly_smiling_face: 
This needs to be done for Popovers as well. Ref: https://github.com/twbs/bootstrap/issues/32379#issuecomment-763860275

ffoodd (approved):
LGTM!

We may apply `margin-top: 0;` only once by adding `&.show:not([data-popper-placement]) { … }` to `.dropdown-menu` above, but I think the added selector would compensate the potential gain. Better to keep redundancy consistent and obvious.

mdo (approved):

PR #32525, opened 2020-12-18
update(utilities) : bring back left right utility classes variation

borracciaBlu:


ffoodd:
Thanks for submitting PR.

TBH I don't think there is any value to include this as-is, since it's more or less the same than adding this to any custom build. And it does not enhance any feature in our core.

Adding support for multiple class names in our utility API would be an enhancement—though I can't say if this would be useful (since it'd still require a custom build to add or rename utilities).

I seriously think the current utility API is enough to workaround those renamings. Any opposing opinion, @twbs/css-review ?

MartijnCuppens:
I got to admit, the `left`/`right` to `start`/`end` switch was an adaption at first, but I got used to it pretty fast. Since the whole CSS scene is moving more towards logical properties, I think it's just a matter of time before everyone is used to the start/end concept. Since we can only make these BCs for major versions, this was the best time we could do this.

Adding aliases doesn't feel like a good idea consistency-wise. It just increases chances of mixing 2 different approaches.

borracciaBlu:
Thanks @ffoodd , @MartijnCuppens that was a quick one.

Just a small comment for the record.

**On the current implementation**
Even if the current implementation is not ideal (i agree with that) was having the benefits of:

1 + recognise the issue on missing `r`|`l` classes and the struggle people will have on the process
2 + some how, it does facilitate the process towards `start`|`end`
3 + it was in accordance with the requirement mentioned by @ffoodd in the community [feedback]( https://github.com/twbs/bootstrap/discussions/32402#discussioncomment-218527)

**If I was implement it as I wish**
+ the flag would be true by default
+ I would use `rtl:ignore` exposing it through `@mixin generate-utility` and mark all these utility with that, to cover the case exposed by @zorn-v in the [feedback](https://github.com/twbs/bootstrap/discussions/32402#discussioncomment-222216) about having proper `left` and `right` even on `rtl`  

Thanks again for the review

ffoodd:
We recognise the needed effort to upgrade, since we made this on purpose in a major release and wrote a comprehensive list of changes to help with the upgrade.

Adding aliases would only help people to not using our feature: I don't consider this as helping people to move forward, anf more certainly the exact opposite of facilitating the change.

If one is able to come with a solution that doesn't held Bootstrap back and ease the use of old namings, we'd consider it. This PR is simply not doing this, since applying your suggestion in a custom build (without modifying our core) would give the exact same result. There's no reason to include this in our core since it does not add any value, compared to what's already possible thanks to our utility API.


And using `rtl:ignore` for those is exactly what we don't want for RTL in Bootstrap since our goal is to allow to use the same markup for both directions. Your suggestion would not improve RTL experience for Bootstrap users, compared to v4.

borracciaBlu:
Hi @ffoodd, thanks for your feedback.

Let me elaborate a little on why I do consider this helping on facilitating moving towards `start`|`end`.

I believe that because this breaking change a lot of devs will resist on embracing the new bootstrap version at all for two main reasons at leas: 1) the shift in mental model and 2) the breaking effect on existing code base and plugins.

If we were having a bridge between `left`|`right` to `start`|`end` (as this PR partially attempt to) it would be a brain 0 decision if they should upgrade at all to `v5` or not.

This would open a door to `start`|`end` syntax.

Now, I see your point when you saying that it would slow down on using `start`|`end`.  
It is true, having `right`|`left` will partially slow down on adoption at least at the beginning but i wouldn't worry too much on that because the main direction is clear as all the component are already with the `start`|`end` syntax and the docs clearly push towards `start`|`end`.

But this is because there's no perfect solution on this i'm afraid, only compromises.

It did stuck in my head `There's no reason to include this in our core`. Perhaps you right, maybe it should be a different package. Something that can be plug to restore compatibility. Something that a dev with an existing codebase can rely on.

For the `rtl:ignore` i see your point on the RTL prospective. The reason why i was thinking about that is to maintain the current mental model intact, something like `left`|`right` old way, `start`|`end` new way.

Thanks again for all those valuable considerations.

Issue #25788, opened 2018-03-05
Animated hamburger button


SkuterPL:
My next idea is create an animation for hamburger button for example change from strips to X (I could do it in BS3, but in BS4 is a svg which I can't change).

You can add choice for user which button wants to use (normal or animated)

browner12:
you can use whatever you want for the icon. simply change

```html
<span class="navbar-toggler-icon"></span>
```

to something else.

SkuterPL:
I know that and already I did it, but I think that as option it would be good for newer users of Bootstrap. They wouldn't be able to have to write extra CSS code for that.

browner12:
BS can't be everything to everyone.  But if you really want it, send a PR.

mdo:
Thanks, but we'll pass.

PR #32526, opened 2020-12-18
Animate `navbar-toggler-icon` using pseudo-elements

goulvench:
The proposed change replaces the current static SVG background in navbar toggler buttons with pseudo-elements, which animate to an x-shape when the menu is expanded. 
This change adds visual feedback when opening menus, like the accordion component, and like most menu icons. 

I know that animating the menu icon has been discussed and rejected in #25788 but since I add this to every website I build, I thought the discussion could be reopened as others might find it useful too.

Code linted and tested, works with light, dark and colored backgrounds thanks to `currentColor`, and is easy to override by using a different class on the `<span>`.


ffoodd:
Couldn't agree more since that's [what happens in Boosted](https://v5-dev--boosted.netlify.app/docs/5.0/components/navbar/#external-content)—still using inlined SVGs though.

scss/_navbar.scss
```diff
@@ -139,13 +139,49 @@
...
+  &::after {
+    position: absolute;
+    left: 0;
+    width: inherit;
+    height: inherit;
+    content: "";
+    background: currentColor;
+    transform: rotate3d(0, 0, 1, 0);
+    transform-origin: 50% 50%;
+    @include transition(.3s ease);
```
ffoodd, line 160: Specifying `transform` and `top` as transitionned properties might be better, wouldn't it?

XhmikosR:
I don't mind this but it seems the new toggler has weird padding.

As for the animation, can't/should we use another class or data attribute for toggling it?

scss/_navbar.scss
```diff
@@ -139,13 +139,49 @@
...
+  &::before,
+  &::after {
+    position: absolute;
+    left: 0;
+    width: inherit;
+    height: inherit;
+    content: "";
+    background: currentColor;
+    transform: rotate3d(0, 0, 1, 0);
+    transform-origin: 50% 50%;
```
ffoodd, line 159: Could be shortened to `50%`.

scss/_navbar.scss
```diff
@@ -139,13 +139,49 @@
...
+    @include transition(.3s ease);
+  }
+  &::before {
+    top: $navbar-toggler-padding-y * -2;
+  }
+  &::after {
+    top: $navbar-toggler-padding-y * 2;
+  }
```
ffoodd, line 167: I'd introduce a new variable for this use case, to dissociate those offsets from the button's padding.

scss/_navbar.scss
```diff
@@ -139,13 +139,49 @@
...
   display: inline-block;
-  width: 1.5em;
-  height: 1.5em;
-  vertical-align: middle;
-  background-repeat: no-repeat;
-  background-position: center;
-  background-size: 100%;
+  width: $navbar-toggler-font-size;
+  height: $border-width * 2;
+  vertical-align: super;
```
ffoodd, line 146: I'd keep `middle` as a value, and simply add `top: -#{$border-width}` to center icon.

ffoodd (changes requested):
A few things to improve maintainability and customizability, but I like it :)

ffoodd:
@XhmikosR alignment is slightly messed indeed, I suggested another way.

Also regarding `[aria-expanded]`, I personnaly prefer this—but if there's a disagreement on this, using `:not(.collapsed)` instead should do the trick :+1:

scss/_navbar.scss
XhmikosR: `top` isn't performant to transition IIRC. `transform` is though.

ffoodd: So that'd need a complex `transform` with both `rotate3d()` and `translate3d()` to drop the `top` part.

XhmikosR: Let's wait for @mdo and I'm pretty sure we'll figure something out altogether.

I still think that this should be toggleable somehow and perhaps respect the reduced motion?

ffoodd: This is handled since it's using our `transition()` mixin.

XhmikosR: OK, so this leaves the option to toggle the animation or not. The class approach would mean it would be toggleable without recompiling, but I don't think we do this anywhere else (yet).

scss/_navbar.scss
MartijnCuppens: Could also be removed since it's the default value 😃

goulvench:
Thanks people for the useful feedback! 
I've started taking your comments into account but it's trickier than I thought to make the line width and spacing customizable (I'm getting rounding issues for instance), and make it sharp-looking when toggled. 

I'll give it another shot after the Christmas rush, maybe this time without screen will help me find an even cleaner solution 😊

patrickhlauke:
My one concern with this proposed approach is that in high-contrast mode (which admittedly Bootstrap currently isn't super hot with in general), the button appears completely empty.

high-contrast-borders

Compared to the current SVG background, which is retained.

high-contrast-svg

I'd agree with showing a different SVG background when the menu is expanded, but perhaps non-animated and as an alternative SVG background instead?

XhmikosR:
@goulvench are you still willing to work on the comments above?

goulvench:
@XhmikosR Yes, I'm still alive. Been busy with paid work but I can return to this a bit. I've started working on tackling the various problems mentioned, and I'll give it another go, including mr Lauke's comment about high-contrast mode.

mdo:
Happy to leave this open or close out until you’re ready @goulvench. Let me know what you’re thinking :).

goulvench:
Thanks for the heads-up, I've been overwhelmed these last few months and things got piled upon.
I tried making the bar width customisable but it caused rounding errors (because button height is in rem). And I'm not familiar with high-contrast mode, as per Mr Lauke's remark, so I'll need to read up on that before I can move this forward. 

You can close this issue unless somebody else wants to take a stab at it, I won't have time for that before a while.

mdo:
Closing as stale for now :).

PR #32529, opened 2020-12-18
Tooltips: add missing space

XhmikosR:
@patrickhlauke could you please go through #32246 in case there are more? Also, I'll backport it since v4-dev is affected too.


patrickhlauke:
ah, oops. well spotted. had a look over #32246 and didn't spot any other such errors.

patrickhlauke (approved):

PR #32533, opened 2020-12-18
Revert "Fix for popover example showing above navbar"

XhmikosR:
Reverts twbs/bootstrap#32445

https://github.com/twbs/bootstrap/pull/32445#issuecomment-748080119

Not sure if we can fix the issue any other way.


sijusamson:
> Reverts #32445
> 
> [#32445 (comment)](https://github.com/twbs/bootstrap/pull/32445#issuecomment-748080119)
> 
> Not sure if we can fix the issue any other way.

@XhmikosR I think solution varies depending upon the people's point of view. IMHO I think modals and popover should behave almost in the same way, as popover is kind of a modal too.

mdo (approved):

XhmikosR:
Let's revert it for now and we tackle it later.

Issue #31033, opened 2020-06-14
Please add chips


davidm49090:
I know this has been nixed a few times, but the time is right to add chips to the UI toolbox. Would make life a lot easier for forms development.

https://material.io/components/chips

mdo:
Just the HTML and CSS, or JS to go with it for specific implementations?

davidm49090:
Thank you for the fast reply.

HTML and CSS would be great, but as long as it plays nicely with jQuery I would be happy.

An ideal solution would be one that would store an ID, so that you could have a contact chip that displays the person's name while holding their email address.

davidm49090:
You could make this my birthday _and_ Christmas gift this year.

leandroruel:
i believe the "close" button needs JS and looks like he works as a checkbox like buttons with toggle state
![chips](https://user-images.githubusercontent.com/2836002/84841899-17040880-b01a-11ea-8558-c7e71b2d9608.png)

klutch-kid:
Is there anything I can do to help? Or is this all under control? 
Awesome idea!

davidm49090:
> Just the HTML and CSS, or JS to go with it for specific implementations?

I know you have a lot of other components to focus on right now, but I'd love to find this under the Christmas tree.

prikeshsavla:
So
Chip Style
Icon on the front
Text label
optional close button/icon
optional checkbox state?

davidm49090:
> So
> Chip Style
> Icon on the front
> Text label
> optional close button/icon
> optional checkbox state?

Yes! I think that's it.

prikeshsavla:
I think I can work on that.

AgentSmith0:
Any updates to this?

prikeshsavla:
I couldn't do it. Can someone else take it up

jovitorrevillas:
Had a go at making what could be the design document for the chip component.

# Chip
## Design
![image](https://user-images.githubusercontent.com/22047407/186839366-d305941d-d930-4832-8e4d-d3cb2c173ab9.png)

## Properties
#### Color
![image](https://user-images.githubusercontent.com/22047407/186837746-62c90425-6b1e-4d3f-9dc8-553e712453f7.png)
```
<span class="chip bg-primary">Primary</span>
<span class="chip bg-secondary">Secondary</span>
<span class="chip bg-success">Success</span>
<span class="chip bg-danger">Danger</span>
<span class="chip bg-warning text-dark">Warning</span>
[...]

PR #32562, opened 2020-12-20
Add chips

njavid:


XhmikosR:
Thanks for the contribution but please

1. before making a PR, make sure `npm test` passes locally
2. this needs docs

patrickhlauke:
Also, it would be nice if you first filed an issue explaining what it is you'd like to add, and give an actual example/use case.

XhmikosR:
I guess this is in response to #31033.

njavid:
yes, I'm trying to solve this issue #31033.

mdo:
This is an incomplete approach, lacking docs and examples, as well as being the wrong way to write CSS in Bootstrap. We use rem units, not pixels. We also don't style by element selector generally speaking, so that'd be an issue. Alongside that, I think we'd need to consider how these show up and how folks would interact with them. As it stands now, this isn't a compelling component to be added IMO. Thanks though!

PR #32580, opened 2020-12-21
Live docs CDN

XhmikosR:
TODO:

- [ ] verify any SRI hashes still match
- [ ] ~~switch feather icons to jsDelivr too; just make sure everything still works because the current solution with unpkg is using the latest one I think~~ I'll leave this for another time

@mdo: I'm mot 100% sure we want to do this, but some of the patches might make sense like the `oss.maxcdn.com`.

Let me know and I'll verify the SRI hashes if we decide to proceed. Also, we should probably backport some of these in the v3-dev branch just so that we have everything there too.


mdo:
> * switch feather icons to jsDelivr too; just make sure everything still works because the current solution with unpkg is using the latest one I think

This one should probably be dropped for using Bootstrap Icons somehow.

XhmikosR:
> This one should probably be dropped for using Bootstrap Icons somehow.

For 4.6 and v5.0 we can do but my comment is for the old docs :)

mdo (approved):

PR #32584, opened 2020-12-21
Add floating labels sizing

### LINKED ISSUE #32130: Floating labels: label and sizing
Floating labels are really useful, thanks.

But I think they really take up a lot of space.
A behavior similar to Material Design 2.0 would be preferable with the label going over the edge:
https://mdbootstrap.com/docs/jquery/forms/inputs/#outline-inputs

Also it would be useful to have the ability to change the size (form-floating-sm).

Thanks again
Alessandro

kasraabdollahi:
- add form-floating-lg and form-floating-sm classes
- base dimensions on bootstrap standards (like display-font-sizes)
- large one dimensions are mostly calculated with 5/3.5 fraction
- small one dimensions are mostly calculated with 2.5/3.5 fraction
- 
input_fields_empty
input_fields_filled
focus

Fixes #32130


XhmikosR:
Please make sure edit from maintainers is enabled.

mdo:
Your approach copy-pastes the entire set of selectors vs using a base/modifier class approach that only modifies the required properties and values. If you want to still tackle this, please consider opening a new PR with that approach in mind. Thanks!

PR #32585, opened 2020-12-21
Throw a `TypeError` instead of the generic `Error`

rohit2sharma95:
Ref: https://github.com/twbs/bootstrap/pull/32376#discussion_r546599766


XhmikosR (dismissed):

XhmikosR:
@rohit2sharma95 actually, shouldn't this test fail right now?

https://github.com/twbs/bootstrap/blob/206b9f45f0b91fef62ac18470165d91748ade161/js/tests/unit/util/index.spec.js#L213-L215

Perhaps we should use `toThrowError`?

js/src/dropdown.js
```diff
@@ -263,7 +263,7 @@ class Dropdown extends BaseComponent {
       typeof config.reference.getBoundingClientRect !== 'function'
     ) {
       // Popper virtual elements require a getBoundingClientRect method
-      throw new Error(`${NAME}: Option "reference" provided type "object" without a required "getBoundingClientRect" method.`)
+      throw new TypeError(`${NAME}: Option "reference" provided type "object" without a required "getBoundingClientRect" method.`)
```
XhmikosR, line 266: I wonder if we should make the related tests stricter, probably makes sense.

rohit2sharma95: Updated.
Also, changed the `NAME` to upper case to make the consistency.
Should check for the same in `jQueryInterface` also? :thinking:

XhmikosR: What do you mean?

rohit2sharma95: I mean should the tests be updated here? (To check if the `error` is the instance of `TypeError`)
https://github.com/twbs/bootstrap/blob/7e6d7b8f817ad1212238f15124335d37c93ebfea/js/tests/unit/dropdown.spec.js#L1560-L1564

XhmikosR: Ah, yeah let's switch any other instances to `toThrowError`

XhmikosR (approved):

PR #32586, opened 2020-12-22
Prevent `getSelector` from returning URLs as selector

### LINKED ISSUE #32273: Dropdown-item error with external link when using Scrollspy.
Hello!
I am using Bootstrap 5.0.0-alpha3.
```html
<li><a class="dropdown-item" href="https://duckduckgo.com/">other</a></li>
```
Using <b>Scrollspy</b> I added in a dropdown a <b>dropdown-item</b> that contains a link to an external page (https://duckduckgo.com/) and I get an error in the console.

The error occurs on the following line:
https://github.com/twbs/bootstrap/blob/c63aebc86ba05f0ebb420add653b80804c6a0cff/dist/js/bootstrap.bundle.js#L96-L102

If the selector variable is equal to some external link such as "https://duckduckgo.com/", the following error occurs:

```diff
- 'https://duckduckgo.com/' is not a valid selector
```

My proposal is to enclose it in a try catch:
```js
  var selector = getSelector(element);

  try {
    return document.querySelector(selector) ? selector : null;
  } catch (_) {
    return null;
  }
```

Operating system: "windows 10";
Browser and version : {"Firefox":"83.0", "Microsoft Edge":"87.0.664.47"};

A reduced test case [here](https://jsfiddle.net/vj6x5ek2/5/).

RoflCopter24:
As discussed in issue #32273 the Dropdown and Scrollspy Modules crash with `Uncaught DOMException: Failed to execute 'querySelector' on 'Document': 'foo/bar.html' is not a valid selector` when there happens to be an actual URL in the href.

The culprit seems to be, that https://github.com/twbs/bootstrap/blob/9b7bb7b568e890a088999b023aabf4898dcd1117/js/src/util/index.js#L35-L45
does no checks if that href contains a valid selector.

__This pull request aims to prevent that case by adding conditions that only allow class selectors (`.target`) and anchors/IDs (`#target`) to be returned as selectors, thus preventing crashes caused by real URLs.__

Fixes #32273


js/src/util/index.js
```diff
@@ -36,7 +36,23 @@ const getSelector = element => {
...
+    let hrefAttr = element.getAttribute('href')
+
+    if (!hrefAttr) {
+      return null
+    }
+
+    // The only valid content that could double as a selector are ids, so everything starting with . or #
+    // If a 'real' url is used as the selector, document.querySelector will rightfully
+    // complain it is invalid. (see twbs#32273)
+    if (!hrefAttr.includes('#') && !hrefAttr.startsWith('.')) {
```
XhmikosR, line 48: How about merging this with the `!hrefAttr` check?

RoflCopter24: No problem, I thought it was more readable this way.

js/src/util/index.js
```diff
@@ -36,7 +36,19 @@ const getSelector = element => {
   let selector = element.getAttribute('data-bs-target')
 
   if (!selector || selector === '#') {
-    const hrefAttr = element.getAttribute('href')
+    let hrefAttr = element.getAttribute('href')
+
+    // The only valid content that could double as a selector are ids, so everything starting with . or #
+    // If a 'real' url is used as the selector, document.querySelector will rightfully
+    // complain it is invalid. (see twbs#32273)
+    if (!hrefAttr || (!hrefAttr.includes('#') && !hrefAttr.startsWith('.'))) {
```
rohit2sharma95, line 44: It restricts the user to use only the class and id selectors in the href attribute. This restriction was not before :thinking:

RoflCopter24: I do not see a problem there. I did not find any mention in the documentation, that anything other than IDs in the href are even supported.

Considering that that freedom for using invalid hrefs as selectors was exactly what caused problems and lead to this PR (and this is the big jump to v5), I think this would be the last opportunity to make such a change before possibly a myriad of users starts misusing the href in that way.

Besides, that's what the `data-bs-target` attribute is for, right? (which is encouraged everywhere in the docs)

XhmikosR: Actually, this would fix the issue we have. I guess it's a good solution for v5. We can revisit in v6.

XhmikosR (approved):

rohit2sharma95 (approved):
The comment on line 41 is a little bit misleading, id selector does not start with a dot (`.`).
So this comment should be removed IMO. :thinking: ﻿

mdo (approved):

PR #32599, opened 2020-12-23
Update border using grid breakpoints

ThunderBirdsX3:


MartijnCuppens:
In v5, you can easily enable this functionality with our utility API: https://getbootstrap.com/docs/5.0/utilities/api/

PR #32601, opened 2020-12-23
add link from Download to Javascript

ghost:
Hello! I've been using v5 already in a project and it looks great so far. I typically use npm and thought that paragraph should include a direct link to where you discuss Bootstrap as a module.


mdo (approved):

XhmikosR:
I'm unsure about this since the module section refers to the browser module implementation which isn't exactly the same as Node.js. We have an issue about being module compliant in Node.js.

/CC @rohit2sharma95

christinavoudouris:
@XhmikosR Does that mean importing Bootstrap in React doesn't quite work yet? It didn't work for me, and I ended up just using the CDN in the index.html file. That's actually why I submitted this PR. I was a little confused by the language, but I also wasn't sure if I just had my paths wrong or something. It was easy to just go with the CDN.

XhmikosR:
Without further info I can't know. There's the issue tracker you can try and also fill the needed template. Otherwise, you should provide the smallest test repo we can use to reproduce the issue.

Anyway, as far as this PR goes, I'm unsure if it makes any sense.

christinavoudouris:
That's fair! I see your point about the two parts in the docs being a bit different. I thought the link would provide some clarification, but perhaps there can be more edits after the Node.js issue is worked out. Thanks for looking at it. I'll take another look at importing Bootstrap in React as well.

mdo:
Removing from v5 Beta 2 until we sort it out :).

XhmikosR:
I'm going to close this since it doesn't make sense. We should expand our Node.js docs, though. Feel free to submit another PR :)

PR #32617, opened 2020-12-26
homepage-hero.svg redundant path

arthurshlain:
Redundant pixelated outline path of letter B removed.

image
image


ArthurShlain:
image
image

patrickhlauke:
for the hero image itself, it's probably not that important, but note that just removing the extra path/outline makes the B slightly thinner (animation here to show the difference)

animation showing the hero image with and without the extra outline path

mdo:
The path is intentional given the gradient inside the B—it provides a slight highlight. If you have tips for improving how it's rendered, I'm all ears! Otherwise, we'll leave it as-is :).

PR #32620, opened 2020-12-27
docs: work around Toast CSS conflict with utils

### LINKED ISSUE #32350: Toasts: `.hide` CSS can conflict with utilities
1. Close this toast https://twbs-bootstrap.netlify.app/docs/5.0/components/toasts/#color-schemes
2. Notice that its `display` is not set to `none`

Probably affects v4 too, I haven't checked. We should use `!important` I guess? - **EDIT**: might not be so simple, we need to consider the use cases.

sijusamson:
Fixes #32350.


mdo (approved):
Awesome, thanks!

mdo (approved):

XhmikosR:
My only objection is that maybe we shouldn't use this extra div in the snippet? It only affect our docs, right? If so, we should split the blocks.

sijusamson:
> My only objection is that maybe we shouldn't use this extra div in the snippet? It only affect our docs, right? If so, we should split the blocks.

I guess we have two other options to resolve it.
1) To add `!important` in `display`.
2) May be we can add `display: flex` in `toast` class itself, but not sure what all impact will it have?

mdo:
I'm fine with the extra div. I don't think we need to make any changes to that as it leaves the core component intact and encourages folks to customize everything within the toast.

Issue #30882, opened 2020-05-22
Dropdowns don't close on click in shadow dom / custom element


mohe2015:
Dropdowns don't close on click in shadow dom / custom element.

- Operating system: ArchLinux
- Browser: Firefox Nightly 78.0a1, Chromium 83.0.4103.61
- https://codepen.io/mohe2015/pen/wvKOBNz
- affects latest version v4.5.0 and also main
- maybe because of https://developers.google.com/web/fundamentals/web-components/shadowdom#events but I don't have enough knowledge of bootstrap to fix this myself
- I know that this is not the intended way to use custom elements but it may be how you use them if you want to use bootstrap but still want to use custom elements for your own things.
- I hope I didn't mess anything up.
- Any help would be really appreciated.

mohe2015:
I updated the codepen to 5.0.1 and it seems to be fixed.

PR #32625, opened 2020-12-27
Dropdown — Emit events on the `.dropdown-toggle` button

### LINKED ISSUE #31935: Dropdown Events Not Firing
OSX 10.15.7
Chrome 86.0.4240.80
Bootstrap 5.0-beta-2

Events not firing: `shown.bs.dropdown`, `hide.bs.dropdown`, `hidden.bs.dropdown`
Note: I can't get a simple `click` event either.

Test on the v5 live doc in Console. (https://v5.getbootstrap.com/docs/5.0/components/dropdowns/)

```
const d = document.getElementById('dropdownMenuButton');
d.addEventListener('show.bs.dropdown', (e) =>  console.log(e) );
d.addEventListener('shown.bs.dropdown', (e) => console.log(e) );
d.addEventListener('hide.bs.dropdown', (e) => console.log(e) );
d.addEventListener('hidden.bs.dropdown', (e) => console.log(e) );
```

rohit2sharma95:
Emit all events on the `.dropdown-toggle` button instead of `.dropdown`. Now seems that `relatedTarget` is not needed because the toggling element is available in the main event's target. @XhmikosR 

Closes #31935


rohit2sharma95:
It does not change/break the previous behavior though @XhmikosR 
The event is now triggered on the toggle button and then bubbled up to the dropdown menu container.

XhmikosR:
I haven't looked closely, but if it's not a breaking change please finish the patch so that we ship it. I guess we could think about backporting it if it's really a safe patch.

rohit2sharma95:
I believe that the `relatedTarget` property is not needed because the toggling button is present in the native event object. Since it would be a breaking change so leaving it as it is.

site/content/docs/5.0/components/dropdowns.md
```diff
@@ -958,8 +958,7 @@ Note when `boundary` is set to any value other than `'scrollParent'`, the style
 
 ### Events
 
-All dropdown events are fired at the `.dropdown-menu`'s parent element and have a `relatedTarget` property, whose value is the toggling anchor element.
-`hide.bs.dropdown` and `hidden.bs.dropdown` events have a `clickEvent` property (only when the original Event type is `click`) that contains an Event Object for the click event.
+All dropdown events are fired at the toggling element and then bubbled up. So you can also add event listeners on the `.dropdown-menu`'s parent element. `hide.bs.dropdown` and `hidden.bs.dropdown` events have a `clickEvent` property (only when the original Event type is `click`) that contains an Event Object for the click event.
```
rohit2sharma95, line 961: This `clickEvent` property is present on the `relatedTarget` property (not on the native event object). Should be mentioned IMO.

XhmikosR: @rohit2sharma95 please mention it :)

mohe2015:
Does this also fix https://github.com/twbs/bootstrap/issues/30882 ?

mdo (approved):

